Issue #305, opened 2015-11-02
Enable delay signing the assemblies


tushargupta51:
To enable strong-naming the assemblies, we need to delay sign them with the public key, later our code signing system does the actual signing. We had it enabled for 4.x

tushargupta51:
Adding "keyFile" and "delaySign" attributes in project.json is the solution

PR #306, opened 2015-11-02
delay signing the assemblies

tushargupta51:
#305


tushargupta51:
@brentschmaltz

brentschmaltz:
:shipit:

Issue #289, opened 2015-10-16
Should AuthenticationProtocolMessage constructor take a string ? 


brentschmaltz:
This blocks passing a 'json' string to the constructor and forces users to convert to JsonObject first.

tushargupta51:
@brentschmaltz I am going to investigate why we have the AuthenticationProtocolMessage.ctor that takes a string issuer address? What scenario does it support?

tushargupta51:
@brentschmaltz @RandalliLama I don't see any specific need to have a ctor to set the IssuerAddress property only. We can get rid of it. But if we change the ctor that takes a 'JObject' to the one that takes a 'json string', it will be breaking changes for asp.net guys since they are using our 'JObject' ctor that we added in Rc1.

tushargupta51:
Moving this to RC2. We will keep the constructor that takes 'JObject' so its not a breaking API change. Instead, we will modify the existing 'stirng' ctor to interpret the string as json instead of issuerAddress. and move that ctor from AuthenticationProtocolMessage to OpenIdConnectMessage.

PR #307, opened 2015-11-03
Rc2 fixes

### LINKED ISSUE #302: Update build.cmd to use latest nuget
build.cmd downloads the nuget.exe from https://nuget.org/nuget.exe. That does not download the latest version. Use https://dist.nuget.org/win-x86-commandline/latest/nuget.exe instead.

Also, update nuget.config to use v3 feeds (both myget and nuget), they are faster than v2 feeds. KRelease branch has those changes:
    <add key="DotNetRelease" value="https://www.myget.org/F/dotnet-core-rel/api/v3/index.json" />
    <add key="AspNetRelease" value="https://www.myget.org/F/aspnetcirelease/api/v3/index.json" />
    <add key="NuGet.org" value="https://api.nuget.org/v3/index.json" />

tushargupta51:
Fixes #302, #289

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/LogMessages.cs
@@ -40,7 +40,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
 
         // properties, configuration 
         internal const string IDX10105 = "IDX10105: NonceLifetime must be greater than zero. value: '{0}'";
-        internal const string IDX10106 = "IDX10106: Error in deserializing json: '{0}' into Newtonsoft.JObject.";
+        internal const string IDX10106 = "IDX10106: Error in deserializing to json: '{0}'";
 
         // protocol validation
         internal const string IDX10300 = "IDX10300: The hash claim: '{0}' in the id_token did not validate with against: '{1}', algorithm: '{2}'.";
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/LogMessages.cs
@@ -40,6 +40,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
 
         // properties, configuration 
         internal const string IDX10105 = "IDX10105: NonceLifetime must be greater than zero. value: '{0}'";
+        internal const string IDX10106 = "IDX10106: Error in deserializing json: '{0}' into Newtonsoft.JObject.";
 
         // protocol validation
         internal const string IDX10300 = "IDX10300: The hash claim: '{0}' in the id_token did not validate with against: '{1}', algorithm: '{2}'.";
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
@@ -25,10 +25,11 @@
 //
 //------------------------------------------------------------------------------
 
-using Microsoft.IdentityModel.Logging;
-using Newtonsoft.Json.Linq;
+using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
+using Microsoft.IdentityModel.Logging;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
 {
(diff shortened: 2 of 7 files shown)


tushargupta51:
@brentschmaltz

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
```diff
@@ -40,13 +41,26 @@ public class OpenIdConnectMessage : AuthenticationProtocolMessage
...
 
         /// <summary>
-        /// Initializes an instance of <see cref="OpenIdConnectMessage"/> class with a specific issuerAddress.
+        /// Initializes an instance of <see cref="OpenIdConnectMessage"/> class with a json string.
         /// </summary>
-        public OpenIdConnectMessage(string issuerAddress) : base(issuerAddress) 
+        public OpenIdConnectMessage(string json)
         {
+            if (string.IsNullOrEmpty(json))
+                LogHelper.LogArgumentNullException("json");
```
brentschmaltz: need to do "throw ..." as the api just returns a populated exception.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
```diff
@@ -40,13 +41,26 @@ public class OpenIdConnectMessage : AuthenticationProtocolMessage
...
+
+            try
+            {
+                JObject jsonObject = JObject.Parse(json);
+                SetJsonParameters(jsonObject);
+            }
+            catch
+            {
+                LogHelper.LogException<ArgumentException>(LogMessages.IDX10106, json);
```
brentschmaltz: same, need to 'throw...'

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
tushargupta51: right, forgot that. will update

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
```diff
@@ -40,13 +41,26 @@ public class OpenIdConnectMessage : AuthenticationProtocolMessage
...
-        public OpenIdConnectMessage(string issuerAddress) : base(issuerAddress) 
+        public OpenIdConnectMessage(string json)
         {
+            if (string.IsNullOrEmpty(json))
+                LogHelper.LogArgumentNullException("json");
+
+            try
+            {
+                JObject jsonObject = JObject.Parse(json);
+                SetJsonParameters(jsonObject);
```
brentschmaltz: You could choose to write: SetJsonParameters(JObject.Parse(json)); 
Loosing the local variable.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectMessageTests.cs
```diff
@@ -45,12 +45,13 @@ public void Constructors()
         {
             OpenIdConnectMessage openIdConnectMessage = new OpenIdConnectMessage();
             Assert.Equal(openIdConnectMessage.IssuerAddress, string.Empty);
-            openIdConnectMessage = new OpenIdConnectMessage("http://www.got.jwt.com");
+            openIdConnectMessage = new OpenIdConnectMessage() { IssuerAddress = "http://www.got.jwt.com" };
             Assert.Equal(openIdConnectMessage.IssuerAddress, "http://www.got.jwt.com");
-            ExpectedException expectedException = ExpectedException.ArgumentNullException("IssuerAddress");
+            ExpectedException expectedException = ExpectedException.NoExceptionExpected;
+            string json = @"{""response_mode"":""responseMode"", ""response_type"":""responseType"", ""refresh_token"":""refreshToken""}";
```
brentschmaltz: We should have a tests for 'bad' json. Maybe there is, but I don't see it here.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/LogMessages.cs
```diff
@@ -40,6 +40,7 @@ internal static class LogMessages
 
         // properties, configuration 
         internal const string IDX10105 = "IDX10105: NonceLifetime must be greater than zero. value: '{0}'";
+        internal const string IDX10106 = "IDX10106: Error in deserializing json: '{0}' into Newtonsoft.JObject.";
```
brentschmaltz: We shouldn't specify 'Newtonsoft' here. Just inform user that the 'json' was not well formed.

tushargupta51: We are specifically trying to convert the json string to Newtonsoft.JObject. What is the harm in specifying 'Newtonsoft'?

Issue #168, opened 2015-05-28
Pull request 161 should be reverted and fixed in Mono.


brentschmaltz:
This is a temporary work-around for a Mono issue.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/161

It should be reverted once Mono has been fixed.

tushargupta51:
This issue is for 4.x since the mentioned PR #161  was pushed to 4.x. I am not sure if we want to revert it  since 4.x is not supported on mono.

brentschmaltz:
Agreed, our 4.x code is working as desired.

PR #310, opened 2015-11-12
Release4x

brentschmaltz:
Merging release into master


tushargupta51:
one of the fixes in release4x has an issue opened for reverting the fix: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/168. Though I am not sure if we want to revert that change.

Issue #250, opened 2015-09-17
SymmetricSignatureProvider for coreclr has no implementation / support for HMAC


sandorfr:
As mentioned in this issue (https://github.com/aspnet/Security/issues/439), it is not possible to produce HMAC signed token atm. This is quite annoying because Hmac token are very common (azure mobile services uses them) and the RSA alternative does not currently work on OSX coreclr.

The workaround was quite easy, I just had to create a SignatureProviderFactory which returns a SymmetricSignatureProvider inspired from the main branch of this repo. It just feels really hacky for something that common.

Since there is nothing coming for that matter on the KDev branch, I would be very happy to provide a pull request adding that support.

Warm regards

sandorfr:
I also had to derive SymmetricSecurityKey to implement GetSignatureProvider which is used by the current JwtSecurityTokenHandler despite I provide an explicit SignatureProviderTokenFactory.

The current code base seems a little confusing with that SignatureProviderTokenFactory which can come from to many different places and you can't replace with your own with a single configuration.

RandalliLama:
@brentschmaltz Could you take a look at this?

tushargupta51:
@sandorfr We are planning to add Symmetric Signature provider support in KDev in RC. Do you want to send a PR for just HMAC support or complete symmetric signature provider because currently in KDev, it is just stubs.

sandorfr:
I'm gonna make a preview PR with my current work so we can discuss how I can make it more complete.

brentschmaltz:
great, love to see it.

sandorfr:
Yes I noticed that while preparing my PR, there almost nothing to be done (or I'm missing something)
https://github.com/sandorfr/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1

brentschmaltz:
@sandorfr I think your fix looks pretty good.  We don't want to create signature with SHA1, we can verify them though. Can you get the PR against this repo?

sandorfr:
@brentschmaltz I already signed a CLA for another PR should I do it again?

Regarding the SHA1, I know this is not the best choice in terms of security but is there a specific reason to prevent people to do it without overriding the signature provider is they need it?

martinbliss:
[...]

PR #313, opened 2015-11-16
adding support for symmetric signature provider

tushargupta51:
#250


src/System.IdentityModel.Tokens/SymmetricSecurityKey.cs
```diff
@@ -57,7 +57,16 @@ public override int KeySize
 
         public override SignatureProvider GetSignatureProvider(string algorithm, bool forSigning)
```
brentschmaltz: change 'forSigning' to be consistent with SignatureProviderFactory.

Issue #316, opened 2015-11-18
Re-introduce a SecurityAlgorithms/JwtAlgorithms mapping table


PinpointTownes:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/300#issuecomment-157877007

/cc @brentschmaltz

tushargupta51:
@PinpointTownes @brentschmaltz 
I think we have consensus on having an outbound mapping (which means jwt generated by this library will have short alg names).

I and @brentschmaltz felt that we do not need an inbound mapping because the signature providers do not need to distinguish between a jwt alg or a xmldsig alg. As long as it knows how to handle both algorithms, it should be good. And what it supports is reflected by `IsSupported` method.

brentschmaltz:
@PinpointTownes @tushargupta51 Some background on why the mapping was originally needed. In 4.x and previous versions of Wilson, Algorithm logic such as: IsSupportedAlgorithm, GetHashAlgorithm, were delegated to the SecurityKey in System.IdentityModel.dll. Those SecurityKeys only understood certain algorithms. For example an X509AsymmetricSecurityKey would not understand RS256. Since RS256 and http://www.w3.org/2001/04/xmldsig-more#rsa-sha256 are the same, we needed to map. 
With our new model of viewing the SecurityKey as:
1. a bunch of bits
2. a wrapper for a CryptoProvider (think constructor RsaSecurityKey(RSA))
We don't need the mapping as our CryptoProviders understand that RS256 == http://www.w3.org/2001/04/xmldsig-more#rsa-sha256 and will do the right thing.

tushargupta51:
@brentschmaltz @PinpointTownes where have we landed with this?

PinpointTownes:
Sadly, nowhere. And it's still somewhere we **absolutely need**, specially since XMLdsig and JWA identifiers have been merged in the same class (`SecurityAlgorithms`).

polita:
Because we only support JWT right now, we will consider this when we reintroduce SAML.

PinpointTownes:
Fantastic, and in the meantime... https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/400

brentschmaltz:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/447
[...]

Issue #315, opened 2015-11-18
Add missing ECDSA xmldsig identifiers to SecurityAlgorithms


PinpointTownes:
> http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256
http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha384
http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha512

https://tools.ietf.org/html/rfc6931

@brentschmaltz I can submit a PR for that :clap:

PinpointTownes:
RSASSA-PSS identifiers should also be added:

> http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1
http://www.w3.org/2007/05/xmldsig-more#sha384-rsa-MGF1
http://www.w3.org/2007/05/xmldsig-more#sha512-rsa-MGF1

brentschmaltz:
@PinpointTownes so even though we don't support those right now, I see no downside in introducing them for users who are taking advantage of custom SigningProviders.

PR #300, opened 2015-10-28
RC1 fixes.

brentschmaltz:
1. Formatting in logging.
2. Add SecurityTokenDescriptor
3. Use SigningCredentials only, remove SignatureProvider on JwtSecurityTokenHandler
4. Remove statics on JwtSecurityTokenHandler
5. OM assuming JwtSecurityToken on JwtSecurityTokenHandler
6. Test Cleanup
7. General code cleanup.
[...]

PR #317, opened 2015-11-21
Re-introduce a SecurityAlgorithms/JwtAlgorithms mapping table

PinpointTownes:
Fixes https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/316 and https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/315.

/cc @brentschmaltz

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
@@ -429,28 +429,22 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
             {
                 switch (algorithm)
                 {
-                    case SecurityAlgorithms.SHA256:
-                    case SecurityAlgorithms.ECDSA_SHA256:
-                    case SecurityAlgorithms.HMAC_SHA256:
-                    case SecurityAlgorithms.RSA_SHA256:
-                    case SecurityAlgorithms.RsaSha256Signature:
-                    case SecurityAlgorithms.PS256:
+                    case JwtAlgorithms.ECDSA_SHA256:
+                    case JwtAlgorithms.HMAC_SHA256:
+                    case JwtAlgorithms.RSASSA_PSS_SHA256:
+                    case JwtAlgorithms.RSA_SHA256:
                         return SHA256.Create();
 
-                    case SecurityAlgorithms.SHA384:
-                    case SecurityAlgorithms.ECDSA_SHA384:
-                    case SecurityAlgorithms.HMAC_SHA384:
-                    case SecurityAlgorithms.RSA_SHA384:
-                    case SecurityAlgorithms.RsaSha384Signature:
-                    case SecurityAlgorithms.PS384:
+                    case JwtAlgorithms.ECDSA_SHA384:
+                    case JwtAlgorithms.HMAC_SHA384:
+                    case JwtAlgorithms.RSASSA_PSS_SHA384:
+                    case JwtAlgorithms.RSA_SHA384:
                         return SHA384.Create();
 
-                    case SecurityAlgorithms.SHA512:
-                    case SecurityAlgorithms.RsaSha512Signature:
-                    case SecurityAlgorithms.RSA_SHA512:
-                    case SecurityAlgorithms.ECDSA_SHA512:
-                    case SecurityAlgorithms.HMAC_SHA512:
-                    case SecurityAlgorithms.PS512:
+                    case JwtAlgorithms.ECDSA_SHA512:
+                    case JwtAlgorithms.HMAC_SHA512:
+                    case JwtAlgorithms.RSASSA_PSS_SHA512:
+                    case JwtAlgorithms.RSA_SHA512:
                         return SHA512.Create();
                 }
 
--- a/src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
@@ -57,11 +57,14 @@ namespace System.IdentityModel.Tokens.Jwt
         {
             if (signingCredentials == null)
             {
-                this[JwtHeaderParameterNames.Alg] = SecurityAlgorithms.NONE;
+                this[JwtHeaderParameterNames.Alg] = JwtAlgorithms.NONE;
             }
             else
             {
-                this[JwtHeaderParameterNames.Alg] = signingCredentials.Algorithm;
+                string algorithm;
+                if (OutboundAlgorithmMap.TryGetValue(signingCredentials.Algorithm, out algorithm))
+                    this[JwtHeaderParameterNames.Alg] = algorithm;
+
                 if (!string.IsNullOrEmpty(signingCredentials.Key.KeyId))
                     this[JwtHeaderParameterNames.Kid] = signingCredentials.Key.KeyId;
             }
@@ -69,6 +72,45 @@ namespace System.IdentityModel.Tokens.Jwt
             this[JwtHeaderParameterNames.Typ] = JwtConstants.HeaderType;
             SigningCredentials = signingCredentials;
         }
+
+        /// <summary>
+        /// Gets or sets the <see cref="IDictionary{TKey, TValue}"/> used to map
+        /// the JWA identifiers with the corresponding security algorithms.
+        /// </summary>
+        public static IDictionary<string, string> InboundAlgorithmMap { get; } = new Dictionary<string, string>
+        {
+            [JwtAlgorithms.ECDSA_SHA256] = SecurityAlgorithms.EcdsaSha256Signature,
+            [JwtAlgorithms.ECDSA_SHA384] = SecurityAlgorithms.EcdsaSha384Signature,
+            [JwtAlgorithms.ECDSA_SHA512] = SecurityAlgorithms.EcdsaSha512Signature,
+
+            [JwtAlgorithms.HMAC_SHA256] = SecurityAlgorithms.HmacSha256Signature,
+            [JwtAlgorithms.HMAC_SHA384] = SecurityAlgorithms.HmacSha384Signature,
+            [JwtAlgorithms.HMAC_SHA512] = SecurityAlgorithms.HmacSha512Signature,
+
+            [JwtAlgorithms.RSA_SHA256] = SecurityAlgorithms.RsaSha256Signature,
+            [JwtAlgorithms.RSA_SHA384] = SecurityAlgorithms.RsaSha384Signature,
+            [JwtAlgorithms.RSA_SHA512] = SecurityAlgorithms.RsaSha512Signature
+        };
+
+        /// <summary>
+        /// Gets or sets the <see cref="IDictionary{TKey, TValue}"/>
+        /// used to map the security algorithms with their JWA identifier.
+        /// </summary>
+        /// <exception cref="ArgumentNullException">'value' is null.</exception>
+        public static IDictionary<string, string> OutboundAlgorithmMap { get; } = new Dictionary<string, string>
+        {
+            [SecurityAlgorithms.EcdsaSha256Signature] = JwtAlgorithms.ECDSA_SHA256,
+            [SecurityAlgorithms.EcdsaSha384Signature] = JwtAlgorithms.ECDSA_SHA384,
+            [SecurityAlgorithms.EcdsaSha512Signature] = JwtAlgorithms.ECDSA_SHA512,
+
+            [SecurityAlgorithms.HmacSha256Signature] = JwtAlgorithms.HMAC_SHA256,
+            [SecurityAlgorithms.HmacSha384Signature] = JwtAlgorithms.HMAC_SHA384,
+            [SecurityAlgorithms.HmacSha512Signature] = JwtAlgorithms.HMAC_SHA512,
+
+            [SecurityAlgorithms.RsaSha256Signature] = JwtAlgorithms.RSA_SHA256,
+            [SecurityAlgorithms.RsaSha384Signature] = JwtAlgorithms.RSA_SHA384,
+            [SecurityAlgorithms.RsaSha512Signature] = JwtAlgorithms.RSA_SHA512
+        };
 
         /// <summary>
         /// Gets the signature algorithm that was used to create the signature.
--- a/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
@@ -659,6 +659,12 @@ namespace System.IdentityModel.Tokens.Jwt
                     return jwt;
             }
 
+            var algorithm = jwt.Header.Alg;
+
+            // If the algorithm cannot be resolved, used the 'alg' claim as-is.
+            if (!string.IsNullOrEmpty(jwt.Header.Alg) && !JwtHeader.InboundAlgorithmMap.TryGetValue(jwt.Header.Alg, out algorithm))
+                algorithm = jwt.Header.Alg;
+
             byte[] signatureBytes = Base64UrlEncoder.DecodeBytes(jwt.RawSignature);
 
             // if the kid != null and the signature fails, throw SecurityTokenSignatureKeyNotFoundException
@@ -693,7 +699,7 @@ namespace System.IdentityModel.Tokens.Jwt
                 {
                     try
                     {
-                        if (ValidateSignature(encodedBytes, signatureBytes, securityKey, jwt.Header.Alg))
+                        if (ValidateSignature(encodedBytes, signatureBytes, securityKey, algorithm))
                         {
                             IdentityModelEventSource.Logger.WriteInformation(LogMessages.IDX10242, token);
                             jwt.SigningKey = securityKey;
--- a/src/System.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/System.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -58,12 +58,9 @@ namespace System.IdentityModel.Tokens
         /// </summary>
         public static readonly Dictionary<string, int> DefaultMinimumAsymmetricKeySizeInBitsForSigningMap = new Dictionary<string, int>()
         {
-            { SecurityAlgorithms.ECDSA_SHA256, 256 },
-            { SecurityAlgorithms.ECDSA_SHA384, 256 },
-            { SecurityAlgorithms.ECDSA_SHA512, 256 },
-            { SecurityAlgorithms.RSA_SHA256, 2048 },
-            { SecurityAlgorithms.RSA_SHA384, 2048 },
-            { SecurityAlgorithms.RSA_SHA512, 2048 },
+            { SecurityAlgorithms.EcdsaSha256Signature, 256 },
+            { SecurityAlgorithms.EcdsaSha384Signature, 256 },
+            { SecurityAlgorithms.EcdsaSha512Signature, 256 },
             { SecurityAlgorithms.RsaSha256Signature, 2048 },
             { SecurityAlgorithms.RsaSha384Signature, 2048 },
             { SecurityAlgorithms.RsaSha512Signature, 2048 }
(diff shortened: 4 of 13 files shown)


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -429,28 +429,22 @@ public virtual HashAlgorithm GetHashAlgorithm(string algorithm)
             {
                 switch (algorithm)
                 {
-                    case SecurityAlgorithms.SHA256:
```
PinpointTownes: FYI, I removed the `SHA256` value as it's not a valid JWA algorithm.

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
```diff
@@ -57,13 +57,21 @@ public JwtHeader(SigningCredentials signingCredentials)
...
             {
-                this[JwtHeaderParameterNames.Alg] = SecurityAlgorithms.NONE;
+                this[JwtHeaderParameterNames.Alg] = JwtAlgorithms.NONE;
             }
             else
             {
-                this[JwtHeaderParameterNames.Alg] = signingCredentials.Algorithm;
+                string algorithm;
+                // Review: should we throw an exception if no JWA identifier can be found?
+                if (OutboundAlgorithmMap.TryGetValue(signingCredentials.Algorithm, out algorithm))
```
PinpointTownes: Should we throw if the algorithm is not supported or just ignore it?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -659,6 +659,17 @@ protected virtual JwtSecurityToken ValidateSignature(string token, TokenValidati
                     return jwt;
             }
 
+            var algorithm = jwt.Header.Alg;
+
+            if (!string.IsNullOrEmpty(jwt.Header.Alg))
+            {
+                // Review: should this code path throw instead of using the "alg" as-is?
+                if (!JwtHeader.InboundAlgorithmMap.TryGetValue(jwt.Header.Alg, out algorithm))
```
PinpointTownes: Same question.

src/System.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -158,21 +152,15 @@ protected virtual HashAlgorithmName GetHashAlgorithmName(string algorithm)
 
             switch (algorithm)
             {
-                case SecurityAlgorithms.SHA256:
```
PinpointTownes: I removed SecurityAlgorithms.SHA256 (and the equivalents) because it's not really a signature algorithm, but I'm fine reverting this change if you think it's better to keep it.

PinpointTownes:
Hey, no CI builds on this repository? YOLO! :smile:

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectProtocolValidatorTests.cs
```diff
@@ -271,7 +271,7 @@ public void ValidateMessageWithIdTokenCode()
             var cHashClaim = IdentityUtilities.CreateHashClaim(validCode, "SHA256");
             var jwt = CreateValidatedIdToken();
             jwt.Payload.AddClaim(new Claim(JwtRegisteredClaimNames.Nonce, validNonce));
-            jwt.Header[JwtHeaderParameterNames.Alg] = "SHA256";
+            jwt.Header[JwtHeaderParameterNames.Alg] = JwtAlgorithms.RSA_SHA256;
```
PinpointTownes: Same remark as above, SHA256 is not a signature algorithm and is not a valid valid for `alg`.

PinpointTownes:
Question: `OpenIdConnectProtocolValidator.HashAlgorithmMap` doesn't seem to be used (except by the tests :smile:). Should we remove it?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
brentschmaltz: It is a valid Hash algorithm.

PinpointTownes: `GetHashAlgorithm` is used by `ValidateHash`, whose `algorithm` parameter corresponds to the `alg` header, which is necessarily a signature algorithm, not a digest algorithm.

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
```diff
@@ -71,6 +79,45 @@ public JwtHeader(SigningCredentials signingCredentials)
         }
 
         /// <summary>
+        /// Gets or sets the <see cref="IDictionary{TKey, TValue}"/> used to map
+        /// the JWA identifiers with the corresponding security algorithms.
+        /// </summary>
+        public static IDictionary<string, string> InboundAlgorithmMap { get; } = new Dictionary<string, string>
```
brentschmaltz: Ideally, Inbound/OutboundAlgorithmMap would be instance, however we can't instantiate for every token being processed, as this is clearly a performance issue. There is a way to provide instance, but would require a bit of internal logic.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: Since ValidateSignature is extensible, we should not throw here.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
brentschmaltz: I understand that. However you are making assumptions about what the algorithm will be used for.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -429,28 +429,22 @@ public virtual HashAlgorithm GetHashAlgorithm(string algorithm)
...
-                    case SecurityAlgorithms.ECDSA_SHA256:
-                    case SecurityAlgorithms.HMAC_SHA256:
-                    case SecurityAlgorithms.RSA_SHA256:
-                    case SecurityAlgorithms.RsaSha256Signature:
-                    case SecurityAlgorithms.PS256:
+                    case JwtAlgorithms.ECDSA_SHA256:
```
brentschmaltz: You are turning this into: JwtAsymmetricSignatureProvider. It is general purpose and needs to serve other algorithms.

src/System.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -58,12 +58,9 @@ public class AsymmetricSignatureProvider : SignatureProvider
         /// </summary>
         public static readonly Dictionary<string, int> DefaultMinimumAsymmetricKeySizeInBitsForSigningMap = new Dictionary<string, int>()
         {
-            { SecurityAlgorithms.ECDSA_SHA256, 256 },
-            { SecurityAlgorithms.ECDSA_SHA384, 256 },
-            { SecurityAlgorithms.ECDSA_SHA512, 256 },
-            { SecurityAlgorithms.RSA_SHA256, 2048 },
-            { SecurityAlgorithms.RSA_SHA384, 2048 },
-            { SecurityAlgorithms.RSA_SHA512, 2048 },
+            { SecurityAlgorithms.EcdsaSha256Signature, 256 },
```
brentschmaltz: I am concerned about reduction here. If ValidateAsymmetricSecurityKeySize can't find the algorithm, we allow it through.

src/System.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
brentschmaltz: Given we don't know how the HashAlgorithmName is going to be used, why should we choose to claim we don't support it?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -693,7 +704,7 @@ protected virtual JwtSecurityToken ValidateSignature(string token, TokenValidati
                 {
                     try
                     {
-                        if (ValidateSignature(encodedBytes, signatureBytes, securityKey, jwt.Header.Alg))
+                        if (ValidateSignature(encodedBytes, signatureBytes, securityKey, algorithm))
```
brentschmaltz: Have a look at: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L395 , https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L599

src/System.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
PinpointTownes: There's actually no reduction here, I just removed the JWA identifiers and replaced them by their XMLDsig equivalent :smile:

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
PinpointTownes: And what am I supposed to see, exactly? :smile: 

Here, the algorithm variable contains the XMLDsig name if it was found, or the `alg` header as-is if no corresponding algorithm was found.

brentschmaltz:
@PinpointTownes what do you what to do with this?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: We are passing the SigningCredentials not JwtHeader.Alg.

PinpointTownes: Sorry, I don't understand what you mean.

PinpointTownes:
@brentschmaltz when you're happy with all the changes, don't hesitate to merge it :smile:

src/System.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
PinpointTownes: This method is always used to "extract"/resolve the digest algorithm from a signature algorithm name (RS256 -> SHA256). I'm not sure it's really useful/semantically correct to include the digest algorithms.

PinpointTownes:
Rebased.

test/System.IdentityModel.Tokens.Jwt.Tests/ExtensibilityTests.cs
```diff
@@ -133,6 +133,45 @@ public void NamedKeySecurityKeyIdentifierClause_Extensibility()
         }
 #endif
 
+        [Fact(DisplayName = "ExtensibilityTests: Algorithm names can be mapped inbound and outbound (AsymmetricSignatureProvider)")]
```
PinpointTownes: FYI, these tests have been removed by @tushargupta51's latest PR. If you think it's not a good idea to re-introduce them, I'm fine.

test/System.IdentityModel.Tokens.Jwt.Tests/ExtensibilityTests.cs
```diff
@@ -133,6 +133,45 @@ public void NamedKeySecurityKeyIdentifierClause_Extensibility()
...
+        {
+            JwtSecurityTokenHandler handler = new JwtSecurityTokenHandler();
+            string newAlgorithmValue = "bobsYourUncle";
+
+            string originalAlgorithmValue = ReplaceAlgorithm(SecurityAlgorithms.HmacSha256Signature, newAlgorithmValue, JwtHeader.OutboundAlgorithmMap);
+            JwtSecurityToken jwt = handler.CreateToken(issuer: IdentityUtilities.DefaultIssuer, audience: IdentityUtilities.DefaultAudience, signingCredentials: KeyingMaterial.DefaultSymmetricSigningCreds_256_Sha2) as JwtSecurityToken;
+            ReplaceAlgorithm(SecurityAlgorithms.HmacSha256Signature, originalAlgorithmValue, JwtHeader.OutboundAlgorithmMap);
+
+            // outbound mapped algorithm is "bobsYourUncle", inbound map will not find this
+            ExpectedException expectedException = ExpectedException.SecurityTokenInvalidSignatureException(substringExpected: "IDX10640:");
```
PinpointTownes: This test wasn't passing because the expected exception was not the correct one.

marc-mueller:
@brentschmaltz What is the current state with this PR? I'm really looking forward to have it merged since we are currently struggling with the alg definition.

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
```diff
@@ -57,13 +57,20 @@ public JwtHeader(SigningCredentials signingCredentials)
...
             {
-                this[JwtHeaderParameterNames.Alg] = signingCredentials.Algorithm;
+                string algorithm;
+                if (OutboundAlgorithmMap.TryGetValue(signingCredentials.Algorithm, out algorithm))
+                {
+                    this[JwtHeaderParameterNames.Alg] = algorithm;
+                }
+
                 if (!string.IsNullOrEmpty(signingCredentials.Key.KeyId))
+                {
```
tushargupta51: we are moving towards a coding style to not use braces { } for a single line statement. https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
tushargupta51: @brentschmaltz @PinpointTownes statics have caused issues in past, thats why we moved all claim mapping maps from static to instance. Are we introducing a static map again?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
tushargupta51: I think what @brentschmaltz is saying is that we are not consistent with the value of algorithm. It should be jwt.header.alg instead of signingCredentials.algorithm

PinpointTownes: I'm fine moving the JWA -> XMLDSIG mapping to the private `ValidateSignature` method if you think it's clearer, but what's sure is that we'll need to call `GetSignatureProviderForValidating` with the XMLDSIG identifier, not with the JWA short name.

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
PinpointTownes: Fixed.

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
PinpointTownes: @tushargupta51 I'm fine opting for an instance dictionary. But in this case, it will probably have to be moved to the handler itself, since the mapping is done in `JwtHeader`'s constructor.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
PinpointTownes: @brentschmaltz your opinion?

leastprivilege:
it is great that something that was working is now broken in RC1. This whole going back and forth makes me wonder how much confidence I should have in this library (and the work here in general) altogether.

PinpointTownes:
Luckily, thanks to @marc-mueller and @jgoelten, I've been able to send a PR very early.

(TBH, I'm much more worried about the time needed to review and merge this PR than the breaking changes)

leastprivilege:
Microsoft should hire you. they def need the help, and you should make them pay ;)

brentschmaltz:
@leastprivilege not a bad idea, we always need good devs. It's easy to apply :-).
@PinpointTownes My thoughts on this are that this PR is driven by the desire to use the same SecurityTokenDescriptor across TokenTypes (SAML, JWT and soon to be CWT). Great. I don't think we need changes to the algorithms the SignatureProviders support, they are not JWTSignautreProviders but general in nature. 

I am landing on a static mapping in JwtHeader outbound ONLY. The developer is responsible for setting the algorithm. JWS specifies some algorithms, but does not require only those. A library is free to understand "http://foo.bar" if it wants to.

PinpointTownes:
@leastprivilege lol :smile: 

> @PinpointTownes My thoughts on this are that this PR is driven by the desire to use the same SecurityTokenDescriptor across TokenTypes (SAML, JWT and soon to be CWT). Great. I don't think we need changes to the algorithms the SignatureProviders support, they are not JWTSignautreProviders but general in nature.

Heh no, this PR has nothing to do with `SecurityTokenDescriptor` :smile: 

It's just meant to fix the total inconsistency between `SecurityAlgorithms` and `JwtAlgorithms`, which broke our existing code in ASOS (as reported by @marc-mueller).

> I am landing on a static mapping in JwtHeader outbound ONLY. The developer is responsible for setting the algorithm. JWS specifies some algorithms, but does not require only those. A library is free to understand "http://foo.bar" if it wants to.

The mapping must be bidirectional since we also want to convert JWA algorithms to their XMLDSig name (exposed by `SecurityAlgorithms`) when receiving a token.

But I agree with you on the fact it should be extensible, and that's why - per your suggestion - no exception is thrown if the `alg` header doesn't correspond to a known XMLDSig (the value is just used as-is and flowed up to the `SignatureProvider`, which is free to reject or accept it).

Personally, I'd move `InboundAlgorithmMap` and `OutboundAlgorithmMap` back to `JwtSecurityTokenHandler` and make then non-static, but heh, that's just my opinion :smile:

brentschmaltz:
@PinpointTownes so without the SecurityTokenDescriptor reason, I need to be reminded the reason for mapping. If you are generating tokens, you usually have provided the set of algorithms you support. Usually from .well-known. Something like: "id_token_signing_alg_values_supported":["RS256"]. In this case: 'RS256' should be the algorithm value in SigningCredentials and in the JwtHeader.
SignatureProvider is not JWT specific so there is no reason to only have algorithms specified in JWA / JWS.

PinpointTownes:
> @PinpointTownes so without the SecurityTokenDescriptor reason, I need to be reminded the reason for mapping.

lol, you really need vacations then :smile:

We already discussed the "why" part of this PR here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/300#discussion_r45131494

> In this case: 'RS256' should be the algorithm value in SigningCredentials and in the JwtHeader. 

Specifying the standard `SecurityAlgorithms.RsaSha256Signature` in the `SigningCredentials` used to work but no longer works since your "RC1 fixes" PR.

> SignatureProvider is not JWT specific so there is no reason to only have algorithms specified in JWA / JWS.

That's why I'm suggesting exactly the opposite: make `SignatureProvider` only deal with XMLDSig names (`SecurityAlgorithms`) for consistency and make the JWA -> XMLDSig resolution in the JWT security token handler.

tushargupta51:
@PinpointTownes 
> Specifying the standard SecurityAlgorithms.RsaSha256Signature in the SigningCredentials used to work but no longer works since your "RC1 fixes" PR.
In #300, you mentioned that it works on IM 5, the issue occurs when using IM 4, right? Or is it breaking IM 5 too?

> That's why I'm suggesting exactly the opposite: make SignatureProvider only deal with XMLDSig names (SecurityAlgorithms) for consistency and make the JWA -> XMLDSig resolution in the JWT security token handler.
Why not have `SignatureProvider` work both with JWA and XMLDSig algorithms?

PinpointTownes:
> In #300, you mentioned that it works on IM 5, the issue occurs when using IM 4, right? Or is it breaking IM 5 too?

No, it's working fine in IM 4, but not in IM 5 since the "RC1 fixes" PR.

It's just that IM 4 won't accept processing a JWT token using a XMLDSig (which is definitely the intended behavior!)

Here's what you get when using `SecurityAlgorithms.RsaSha256Signature`:

image

brentschmaltz:
@PinpointTownes what I am also suggesting is the responsibility of specifying the algorithm is with the token creator. The runtime should get and stay out of the way. When we first developed the JWT work, we agonized long and hard about mapping. Mainly because of ASP.Net and User.Name. The relationship between User.Name (ClaimsIdentity.Name) was resulting in issues. That mapping is needed (or perhaps easier to pass the NameClaimType in the constructor).

Other than that, mapping has not really shown the rewards we needed. In the case of IM4 not working, I am suggesting we fix IM4, and move forward with a cleaner OM.

PinpointTownes:
Okay, I give up.

I really thought @leastprivilege was exaggerating when he said he couldn't have any confidence in all the stuff you're doing here but I now realize he was right: you have absolutely NO idea what you're doing.

You think mixing XMLDSig and JWA is not bad... good for you, having XMLDSig identifiers in a JWT is so fun, and totally standard. You want to tie SigningCredentials with JWA and make the algorithm name dependent on the security token handler that creates the token? Fantastic: introduce more coupling, it's the right way to go.

IdentityModel was a great framework, you've killed it.

@marc-mueller @jgoelten sorry guys.

/ragequit

brentschmaltz:
@PinpointTownes I do believe that having a mapping in JwtHeader outbound is a good idea. I am not suggesting that XMLDsig algorithms in JwtHeader is a good idea.
I am suggesting that SignatureProviders shouldn't be JWA or XMLDsig specific but should reflect the algorithms that are supported by the runtime. 
The issue seems to be that IM4 will fault, with tokens generated by RC1. I believe outbound mapping will fix that.

brentschmaltz:
Closing for now.

Issue #320, opened 2015-11-23
Add ConfigureAwait(false) when retrieving metadata.


brentschmaltz:
Getting metatdata will deadlock in the application layer when the httpcontext is locked. We need to maintain the async behavior and not use .Result.

brentschmaltz:
Fixed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/321

PR #321, opened 2015-11-23
Add ConfigureAwait(false) when retrieving metadata. #320

brentschmaltz:
Add ConfigureAwait(false) when retrieving metadata. #320


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -152,7 +152,7 @@ public TimeSpan RefreshInterval
         /// <remarks>If the time since the last call is less than <see cref="AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
         public async Task<T> GetConfigurationAsync()
         {
-            return await GetConfigurationAsync(CancellationToken.None);
+            return await GetConfigurationAsync(CancellationToken.None).ConfigureAwait(false);
```
PinpointTownes: nit: you can remove `async` and directly return the task here, without having to use `ConfigureAwait(false);`.

tushargupta51:
@brentschmaltz do we need configureAwait in FileDocumentRetriever also?

tushargupta51:
:shipit:

brentschmaltz:
Fixed issue #321 (Configure Await)

Issue #325, opened 2015-12-02
Update RsaSecurityKey and ECDsaSecurityKey to support cross-platform scenarios


PinpointTownes:
In IdentityModel 5, `RsaSecurityKey` and `ECDsaSecurityKey` are directly tied to `RSAParameters` and `CngKey`, which prevents cross-platform and advanced scenarios like using a HSM.

These primitives should only accept `RSA` and `ECDsa` instances in their constructor parameters and determining the right implementation should be made by the classes using `RsaSecurityKey` and `ECDsaSecurityKey`, just like in IdentityModel 4.

tushargupta51:
@PinpointTownes How `RSAParameters` prevent cross plat scenarios? `RSAParameters` is available on unix/coreclr: https://github.com/dotnet/corefx/blob/c02d33b18398199f6acc17d375dab154e9a1df66/src/System.Security.Cryptography.OpenSsl/src/System/Security/Cryptography/RSAOpenSsl.cs#L44

PinpointTownes:
Actually, RsaSecurityKey's issue is not about cross-platform but about usability: not being able to flow a `RSA` instance prevents tons of advanced scenarios like using a `RSACryptoServiceProvider` tied to a HSM.

tushargupta51:
@PinpointTownes @brentschmaltz this came up in the review of PR #334 and I want to start the discussion here to agree upon a solution before I make any further changes to the PR.

What is the best place to put these constructors that take `RSA` and `ECDSA` instances, is it the `RsaSecurityKey`/`ECDsaSecurityKey` classes or `AsymmetricSignatureProvider` or both. Comparing to System.IdentityModel 4.5, `RsaSecurityKey` has the constructor that takes the `Rsa` instance as well as the `AsymmetricSignatureFormatter`/Deformatter (which is similar to SignatureProviders in Wilson) took the `Rsa` instance in the constructor.

@brentschmaltz pointed out that it does not look right that we create a SecurityKey from an `Rsa` instance which is essentially a signature provider and then use that key to create the `SignatureProvider`. Why not have a constructor on `SignatureProvider` that takes `AsymmetricAlgorithm` and try to cast it as `ECDsa` or `RSA`, if it cannot then we throw.

PinpointTownes:
[...]

Issue #324, opened 2015-12-02
Fix AsymmetricSignatureProvider to support RSA keys on CoreCLR/Linux


PinpointTownes:
Moved from https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/179#issuecomment-155972405.

_____________

On CoreCLR, `AsymmetricSignatureProvider` tries to use `RSACng` (which is Windows-specific) instead of `RSAOpenSsl`.

      System.DllNotFoundException: Unable to load DLL 'ncrypt.dll': The specified module could not be found.
       (Exception from HRESULT: 0x8007007E)
         at Interop.NCrypt.NCryptOpenStorageProvider(SafeNCryptProviderHandle& phProvider, String pszProviderName, Int32 dwFlags)
         at Internal.Cryptography.Helpers.OpenStorageProvider(CngProvider provider)
         at System.Security.Cryptography.CngKey.Import(Byte[] keyBlob, CngKeyBlobFormat format, CngProvider provider)
         at System.Security.Cryptography.RSACng.ImportParameters(RSAParameters parameters)
         at System.IdentityModel.Tokens.AsymmetricSignatureProvider.ResolveDotNetCoreAsymmetricAlgorithm(AsymmetricSecurityKey key, String algorithm, Boolean willCreateSignatures)
         at System.IdentityModel.Tokens.AsymmetricSignatureProvider..ctor(AsymmetricSecurityKey key, String algorithm, Boolean willCreateSignatures)
         at System.IdentityModel.Tokens.SignatureProviderFactory.CreateProvider(SecurityKey key, String algorithm, Boolean willCreateSignatures)
         at System.IdentityModel.Tokens.SignatureProviderFactory.CreateForSigning(SecurityKey key, String algorithm)
         at System.IdentityModel.Tokens.RsaSecurityKey.GetSignatureProvider(String algorithm, Boolean verifyOnly)
         at System.IdentityModel.Tokens.SecurityKey.GetSignatureProviderForSigning(String algorithm)
         at System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.CreateJwt(String issuer, String audience, IEnumerable`1 claims, Nullable`1 notBefore, Nullable`1 expires, Nullable`1 issuedAt, SigningCredentials signingCredentials)
[...]

PR #328, opened 2015-12-07
using openssl when running on unix/coreclr

tushargupta51:
Fixing #324. I am not changing ECDsa to use OpenSSL yet since that requires updating ECDsaSecurityKey to work on unix (#325). Will come to that later.


tushargupta51:
@brentschmaltz

PinpointTownes:
It looks fine as a temporary fix, but you should definitely consider updating `RsaSecurityKey`'s constructor to take a `RSA` instance instead of creating one in the signature provider.

PinpointTownes:
:shipit:

ericstj:
@bartonjs don't we have a cross-plat type for RSACng/RSAOpenSsl?

The current implementation will cause the JIT to hit the OpenSSL assembly, even on Windows.  That won't work in the future.

The only reason this may work today is because we've delayed the packaging changes to move these platform specific libraries to platform specific folders in the packages.  We're delaying that work because folks cannot yet represent platform-specific dependencies in nuget/project.json.  https://github.com/dotnet/cli/issues/465 https://github.com/NuGet/Home/issues/1660

bartonjs:
@ericstj We didn't 10 days ago, no.  But https://github.com/dotnet/corefx/issues/2953 was fixed 2 days ago (which @PinpointTownes celebrated).

So this particular change can be switched to using RSA.Create() now, and there'll be a couple of builds overlap between RSA.Create() existing and the OpenSsl package going back to registered as unix-only (https://github.com/dotnet/corefx/issues/6046).

Issue #323, opened 2015-12-01
Test failure in System.IdentityModel.Tokens.Jwt.Test.RoundTripTokens


brentschmaltz:
The 'iat' claim was not set implicitly, a millisecond later it was set by default leading to an error when validating claims.

PR #331, opened 2015-12-10
specify IssuedAt property

tushargupta51:
#323 was not completely fixed. The test was still failing. Couple of places were missed where 'issuedAt' needs to be added.


tushargupta51:
@brentschmaltz

brentschmaltz:
:shipit:

PR #333, opened 2015-12-18
Namespace changed, all tests passing.

brentschmaltz:
moved System.IdentityModel.Tokens -> Microsoft.IdentityModel.Tokens


test/System.IdentityModel.Tokens.Jwt.Tests/CreateAndValidateTokens.cs
```diff
@@ -25,13 +25,18 @@
...
 using System.Collections.Generic;
 using System.Globalization;
-using System.IdentityModel.Tokens.Tests;
 using System.Security.Claims;
+using Microsoft.IdentityModel.Tokens;
+using Microsoft.IdentityModel.Tokens.Tests;
 using Xunit;
 
+// since we are in the System ns, we need to map to M.IM.Tokens
+using SigningCreds = Microsoft.IdentityModel.Tokens.SigningCredentials;
```
tushargupta51: @brentschmaltz  Why is this needed?

brentschmaltz: As I hoped the comment indicated, the compiler will default to System.IdentityModel.Tokens.SigningCredentials. So we have to explicitly define the type.

test/System.IdentityModel.Tokens.Jwt.Tests/project.json
```diff
@@ -1,15 +1,19 @@
...
+        "Microsoft.IdentityModel.Tokens.Tests": "5.0.0-*"
     },
     "commands": {
         "test": "xunit.runner.aspnet"
     },
     "frameworks": {
-        "dnx451": { },
+        "dnx451": {
+          "frameworkAssemblies": {
+            "System.IdentityModel": "4.0.0.0"
```
tushargupta51: Why adding this dependency? I don't think we are using any types in this assembly.

brentschmaltz: The is testing that S.IM can be included. Which was not possible before.

tushargupta51:
:shipit:  it has cla-required label for some reason

Issue #318, opened 2015-11-21
Code/access token hashing should use ASCII encoding


PinpointTownes:
http://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken

`OpenIdConnectProtocolValidator.ValidateHash` uses UTF8 while the specs target ASCII. It's probably not a big deal, since access tokens and authorization codes are often only composed of ASCII chars, but something we may want to fix for RTM to avoid potential mismatches.

/cc @brentschmaltz

brentschmaltz:
thanks

Issue #280, opened 2015-10-09
Revert the workaround in OpenIdConnectProtocolValidatorTests


tushargupta51:
https://github.com/dotnet/corefx/issues/3667 bug is fixed in S.Secuirty.Claims but it seems that dnx packages the assembly in the runtime which takes precedence. We need to wait for an updated dnx. Once that is available, we should revert this in the test:
#if DNXCORE50
                    // Temporary until https://github.com/dotnet/corefx/issues/3667 is fixed
                    claims: new List<Claim> { new Claim(JwtRegisteredClaimNames.CHash, " ") },
#else
                    claims: new List<Claim> { new Claim(JwtRegisteredClaimNames.CHash, string.Empty) },
#endif

Issue #330, opened 2015-12-09
Revert the changes applied to AsymmetricSecurityKey


PinpointTownes:
In IdentityModel 5, `HasPrivateKey` and `HasPublicKey` are now properties (instead of methods). 

Though it's not currently an issue as `RsaSecurityKey` and `ECDsaSecurityKey` only accepts key materials (see https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/325), fixing these classes to directly accept `RSA` and `ECDsa` will likely require implementing more elaborated `HasPrivateKey` and `HasPublicKey` implementations to determine whether a private key is available or not, which may have an important CPU/IO cost (when a HSM is used, for instance).

Using methods here would be much more appropriate.

/cc @brentschmaltz

tushargupta51:
@PinpointTownes In a recent PR, I added ctors that take `RSA` or `ECDsa` instances when creating the `SecurityKey`. Though in those cases, because of the cost involved in fake signing to determine whether the key is private or public only (especially for hardware keys), we set the HasPrivateKey property to true and let it fail later at the time of signing.

tushargupta51:
@PinpointTownes @brentschmaltz 
I was debating on whether `HasPrivateKey` should be a property or a method. Based on the guidelines here (https://msdn.microsoft.com/library/ms229054(v=vs.100).aspx), I think a property sounds more appropriate because of the following:
1. The value of the property does not change between calls unless the instance holding the property changes.
2. Though you can argue it could be computationally expensive in case of hardware keys because of fake signing, the effect is not profound since it is only done once per instance and for later calls the stored result is returned, which indicates more of a property behavior.
3. The getter does not generate any potential side effect, it only sets the value of the private variable  _hasPrivateKey.

Also, I am considering removing `HasPublicKey` method from `AsymmetricSecurityKey` since that does not add much value for an asymmetric key which are supposed to almost always have the public key. Also, in current implementation, the API always tends to return true or determine the result based on HasPrivateKey attribute.

PinpointTownes:
> Though you can argue it could be computationally expensive in case of hardware keys because of fake signing, the effect is not profound since it is only done once per instance and for later calls the stored result is returned, which indicates more of a property behavior.

Well, the rule is clear: accessing a property is not supposed to be slow, even if it just impacts the first call.

[...]

PR #334, opened 2015-12-22
Rc2 fixes

tushargupta51:
1. Add ctors to RsaSecurityKey and EcDSASecurityKey that takes rsa and ecdsa instances respectively: #325 
2. SignatureProviderFactory => CryptoProviderFactory
3. Split JsonConstants and JwtConstants classes into proper classes
4. Clean up signature provider tests
5. #318 : use ascii hashing
6. #280: revert test workaround


tushargupta51:
@brentschmaltz

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -34,13 +34,13 @@ namespace Microsoft.IdentityModel.Tokens
     /// Creates <see cref="SignatureProvider"/>s by specifying a <see cref="SecurityKey"/> and algorithm.
     /// <para>Supports both <see cref="AsymmetricSecurityKey"/> and <see cref="SymmetricSecurityKey"/>.</para>
     /// </summary>
-    public class SignatureProviderFactory
+    public class CryptoProviderFactory
```
PinpointTownes: FWIW, the new name looks weird.

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
```diff
@@ -42,30 +42,20 @@ public ECDsaSecurityKey(byte[] blob, CngKeyBlobFormat blobFormat)
 
             CngKey = CngKey.Import(blob, blobFormat);
             BlobFormat = blobFormat;
+            HasPrivateKey = BlobFormat.Format == CngKeyBlobFormat.EccPrivateBlob.Format || BlobFormat.Format == CngKeyBlobFormat.GenericPrivateBlob.Format;
```
PinpointTownes: I don't think keeping this CNG (i.e Windows)-specific constructor is the right thing to do. The public contract should only take OS/platform-agnostic parameters (i.e an `ECDsa` instance).

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
```diff
@@ -42,30 +42,20 @@ public ECDsaSecurityKey(byte[] blob, CngKeyBlobFormat blobFormat)
...
+            ECDsa = ecdsa;
 
-        public override SignatureProvider GetSignatureProvider(string algorithm, bool verifyOnly)
-        {
-            if (verifyOnly)
-                return SignatureProviderFactory.CreateForVerifying(this, algorithm);
-            else
-                return SignatureProviderFactory.CreateForSigning(this, algorithm);
+            // fake signing to determine if the ecdsa instance has the private key or not is a costly operation especially in case of HSM. We return true by default in that case, it will fail later at the time of signing or decrypting.
+            HasPrivateKey = true;
```
PinpointTownes: IMHO, lying is terribad. If I decide to use these properties, I expect them to return an appropriate result, not a fake one.

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
```diff
@@ -78,12 +68,32 @@ public override SignatureProvider GetSignatureProvider(string algorithm, bool ve
         /// </summary>
         public CngKey CngKey { get; private set; }
```
PinpointTownes: Same remark, this type shouldn't expose platform-specific types.

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
```diff
@@ -27,69 +27,79 @@
...
+                    return Parameters.D.Length * 8;
+                else
+                    return 0;
             }
         }
 
+        /// <summary>
+        /// <see cref="RSAParameters"/> used to initialize the key.
+        /// </summary>
+        public RSAParameters Parameters { get; private set; }
```
PinpointTownes: Keeping this property is probably not necessary, as you can extract the RSA parameters via `Rsa.ExportParameters`.

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
```diff
@@ -27,69 +27,79 @@
...
 using System.Security.Cryptography;
+using Microsoft.IdentityModel.Logging;
 
 namespace Microsoft.IdentityModel.Tokens
 {
     public class RsaSecurityKey : AsymmetricSecurityKey
     {
-        private RSAParameters _rsaParamaeters;
-
         public RsaSecurityKey(RSAParameters rsaParameters)
```
PinpointTownes: Personally, I'd probably move that to a static helper that would internally create a `RSA` instance using the most appropriate implementation depending on the runtime environment (Windows: RSACng, Mono: RSACryptoServiceProvider, Linux: RSAOpenSsl).

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -465,6 +445,15 @@ public SignatureValidator SignatureValidator
         }
 
         /// <summary>
+        /// Gets or sets the <see cref="ReadOnlyCollection{SecurityKey}"/> that is to be used for decrypting inbound tokens.
+        /// </summary>
+        public IList<SecurityToken> TokenDecryptionKeys
```
PinpointTownes: Has this property a default value somewhere? Having a null list seems weird.

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
```diff
@@ -27,69 +27,79 @@
...
-        }
-
-        public override bool HasPrivateKey
-        {
-            get
+            else
             {
-                return !(_rsaParamaeters.D == null || _rsaParamaeters.DP == null || _rsaParamaeters.DQ == null || _rsaParamaeters.P == null || _rsaParamaeters.Q == null);
+                // fake signing to determine if the rsa instance has the private key or not is a costly operation especially in case of HSM. We return true by default in that case, it will fail later at the time of signing or decrypting.
+                HasPrivateKey = true;
```
PinpointTownes: Same remark, lying is horrible.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
tushargupta51: I will initialize with the default list.

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
tushargupta51: Another alternative is to use the approach we had in System.IdentityModel.Tokens to use an    `enum PrivateKeyStatus
        {
            AvailabilityNotDetermined,
            HasPrivateKey,
            DoesNotHavePrivateKey
        }`
I agree that its preferable to let the user know the exact status rather than setting an incorrect value. @brentschmaltz

brentschmaltz: +1, we shouldn't be asserting a private key exists. 
I don't like the idea that we try and sign something to determine if the key exists AND I don't like the enum.

I think we have to choose the lesser of two evils.

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
brentschmaltz: +1.

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
brentschmaltz: +1

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: @PinpointTownes any suggestions?

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
```diff
@@ -27,69 +27,79 @@
...
+        public override bool HasPublicKey { get; }
+
+        public override int KeySize
         {
             get
             {
-                return !(_rsaParamaeters.Exponent == null || _rsaParamaeters.Modulus == null);
+                if (Rsa != null)
+                    return Rsa.KeySize;
+                if (HasPublicKey)
```
brentschmaltz: Are public and private keys the same length?

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
```diff
@@ -27,69 +27,79 @@
...
-                throw new ArgumentException("no public or private key material found");
+                throw LogHelper.LogException<ArgumentException>("No public or private key material found");
             }
 
-            _rsaParamaeters = rsaParameters;
+            Parameters = rsaParameters;
         }
 
-        public override int KeySize
+        public RsaSecurityKey(RSA rsa)
```
brentschmaltz: With this constructor what does the RsaParameters property return?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
PinpointTownes: `SignatureProviderFactory` was fine, why changing it? :smile:

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
brentschmaltz: Let's decide if this should be IList<>, ICollection<>, IEnumerable<>. Also return ...<SecurityKey>, instead of ...<SecurityToken>.

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
PinpointTownes: IIRC, the enum was an implementation detail (i.e a private type) in System.IdentityModel 4.5. If you don't like it, it could be replaced by a nullable field returning a boolean:
- `null`: not determined.
- `true`: has a private key.
- `false`: doesn't have a private key.

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
PinpointTownes: Yes it is. But it's not cross-platform: it won't work outside Windows.

PinpointTownes: https://github.com/dotnet/corefx/tree/master/src/System.Security.Cryptography.Cng/src

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
tushargupta51: We plan to add JWE and add methods to create encryption provider later. We are trying to update the OM so that JWE can be added later without breaking stuff.

PinpointTownes: Curious: why not just adding an `EncryptionProviderFactory`?

tushargupta51: That would be yet another thing to have. Currently, SecurityKey class only has CryptoProviderFactory that it can use to create a signature provider or encryption provider. Also both of these providers share the basic attributes - key and algorithm, so it makes sense to have a common wrapper for them.

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
PinpointTownes: Nothing, which sucks, terribly.
If you really want to keep the `RsaParameters` property for conveniency, it might be better to make a simple wrapper calling `Rsa.ExportParameters`.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
PinpointTownes: > Also both of these providers share the basic attributes - key and algorithm, so it makes sense to have a common wrapper for them.

Why not. But semantically, it seems really odd, as signing and encrypting routines are really different (algorithms too, BTW. For instance, there's no encrypting equivalent for ECDsa).

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
brentschmaltz: alright then, we need to think about hiding CNG.

tushargupta51: This property is used by `AsymmetricSignatureProvider` to instantiate an `ECDsaCng` instance if one is not present already. Removing this property only makes sense if we want to remove the constructor that takes the blob and blobFormat, which IMO is useful in scenarios where users want to create they key using components and not the instance, even though that will work on Windows only platforms.

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
tushargupta51: It will be null but we can set it to Rsa.ExportParameters() or Rsa.ExportParameters(true) inside the constructor itself.

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
tushargupta51: This property is useful when the key is constructed using the `RsaParameters`. `Rsa` will be null in that case.

tushargupta51:
@brentschmaltz @PinpointTownes One main comment on the PR is about the usefulness of following 2 constructors:
1. `ECDsaSecuirytKey(byte[] blob, CngKeyBlobFormat blobFormat)`
2. `RsaSecurityKey(RsaParameters rsaParameters)`

I am going to open an issue for discussing that so that we can move forward with this PR. `CngKey` and `RsaParameters` properties are used by `AsymmetricSignatureProvider

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
tushargupta51: yes, for rsa it is, though we should only use Modulus.length for calculating the size. It must be present in RsaParams othewise ImportParameters will throw even for the private keys.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: @PinpointTownes @tushargupta5 My feeling is that as we add new crypto, we would need mods to each key, however I am will to consider this given
>  BTW. For instance, there's no encrypting equivalent for ECDsa
I would like to get this PR in as we need to get the JWE prototype going which needs this PR. Can or did we open an issue here?

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
brentschmaltz: How will we know if we should use 'true' or 'false'?

PinpointTownes: You could use `HasPrivateKey` but it wouldn't be very reliable, as it would like fail with HSM, that never allow exporting the private key.

Ideally, `RsaSecurityKey` should have no `RsaParameters` property. The developer still can export the parameters by using the algorithm instance.

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
```diff
@@ -27,69 +27,110 @@
...
 using System;
 using System.Security.Cryptography;
+using Microsoft.IdentityModel.Logging;
 
 namespace Microsoft.IdentityModel.Tokens
 {
     public class RsaSecurityKey : AsymmetricSecurityKey
     {
-        private RSAParameters _rsaParamaeters;
+        PrivateKeyStatus _privateKeyStatus = PrivateKeyStatus.AvailabilityNotDetermined;
```
PinpointTownes: nit: `private`?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
PinpointTownes: Honestly, if you think the new name is a better name, I won't fight to keep the old one, as it's just a cosmetic change :smile:

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -381,7 +402,7 @@ public override bool Verify(byte[] input, byte[] signature)
                 throw LogHelper.LogArgumentNullException("signature");
 
             if (input.Length == 0)
-                throw LogHelper.LogException<ArgumentException>(LogMessages.IDX10625, "signature");
+                throw LogHelper.LogException<ArgumentException>(LogMessages.IDX10625, "input");
```
PinpointTownes: nit: you could consider `nameof` to avoid this magic string syndrome :smile:

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
```diff
@@ -27,69 +27,110 @@
...
         }
 
-        public override int KeySize
+        public RsaSecurityKey(RSA rsa)
+        {
+            if (rsa == null)
+                throw LogHelper.LogArgumentNullException("rsa");
+
+            Rsa = rsa;
+            if (HasPrivateKey)
```
PinpointTownes: Hopefully, you'll be able to remove the `Parameters` property and avoid that in the constructor :smile:

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
```diff
@@ -27,69 +27,110 @@
...
+                Parameters = Rsa.ExportParameters(true);
+            else
+                Parameters = Rsa.ExportParameters(false);
+        }
+
+        public override bool HasPrivateKey
```
PinpointTownes: Curious: have you considered moving back to a method (due to the potential fake signing cost) or will it be part of another work item?

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
PinpointTownes: Since @brentschmaltz doesn't like the private enum, you could use a nullable boolean here :smile:

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
tushargupta51: I am going to have another PR coming soon (after we discuss the possible solutions on the respective issues: #337 #330 #336). Will handle this issue then.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -188,6 +188,12 @@ private void ResolveDotNetCoreAsymmetricAlgorithm(AsymmetricSecurityKey key, str
...
             {
+                if (rsaKey.Rsa != null)
+                {
+                    _rsa = rsaKey.Rsa;
+                    return;
+                }
+
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                     _rsa = new RSACng();
                 else
```
brentschmaltz: A couple of lines below this, (_rsa as RSA) ... we don't need that cast.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: @PinpointTownes thanks. We still have some final smoothing out here. I am not convinced we havew a good extensibility cross plat story. If you can spare some cycles and have a look once this PR is in, we would appreciate it.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
brentschmaltz: I think we should follow that advise in the future. Thanks.

brentschmaltz:
:shipit: We will need to focus on extensibility XPlat and JWE. This will flush out any other needs.

PR #338, opened 2016-01-13
disposing httpClient in HttpDocumentRetriever.cs

### LINKED ISSUE #288: Review the implementation of IDisposable.
We used to dispose just the 'hash' but we create other disposable objects.

tushargupta51:
fix #288

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
@@ -40,10 +40,10 @@ namespace Microsoft.IdentityModel.Protocols
     /// </summary>
     public class HttpDocumentRetriever : IDocumentRetriever
     {
-        private readonly HttpClient _httpClient;
-        private bool _requireHttps;
+        private HttpClient _httpClient;
+        private static readonly HttpClient _defaultHttpClient = new HttpClient();
 
-        public HttpDocumentRetriever() : this(new HttpClient())
+        public HttpDocumentRetriever()
         {
         }
 
@@ -53,17 +53,12 @@ namespace Microsoft.IdentityModel.Protocols
                 throw LogHelper.LogArgumentNullException("httpClient");
 
             _httpClient = httpClient;
-            _requireHttps = true;
         }
 
         /// <summary>
         /// Requires Https secure channel for sending requests.. This is turned ON by default for security reasons. It is RECOMMENDED that you do not allow retrieval from http addresses by default.
         /// </summary>
-        public bool RequireHttps
-        {
-            get { return _requireHttps; }
-            set { _requireHttps = value; }
-        }
+        public bool RequireHttps { get; set; } = true;
 
         public async Task<string> GetDocumentAsync(string address, CancellationToken cancel)
         {
@@ -76,7 +71,9 @@ namespace Microsoft.IdentityModel.Protocols
             try
             {
                 IdentityModelEventSource.Logger.WriteVerbose(LogMessages.IDX10805, address);
-                HttpResponseMessage response = await _httpClient.GetAsync(address, cancel).ConfigureAwait(false);
+                var httpClient = _httpClient ?? _defaultHttpClient;
+                HttpResponseMessage response = await httpClient.GetAsync(address, cancel).ConfigureAwait(false);
+
                 response.EnsureSuccessStatusCode();
                 return await response.Content.ReadAsStringAsync().ConfigureAwait(false);
             }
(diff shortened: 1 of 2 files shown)


tushargupta51:
@brentschmaltz

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -40,11 +40,12 @@ namespace Microsoft.IdentityModel.Protocols
...
     public class HttpDocumentRetriever : IDocumentRetriever
     {
-        private readonly HttpClient _httpClient;
+        private HttpClient _httpClient;
         private bool _requireHttps;
 
-        public HttpDocumentRetriever() : this(new HttpClient())
+        public HttpDocumentRetriever()
         {
+            _requireHttps = true;
```
brentschmaltz: You can set the default when declaring the variable OR use the new syntax and set it on the property. So this line and the same below can be removed.

tushargupta51: @brentschmaltz updated the declaration

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -76,7 +70,14 @@ public bool RequireHttps
             try
             {
                 IdentityModelEventSource.Logger.WriteVerbose(LogMessages.IDX10805, address);
-                HttpResponseMessage response = await _httpClient.GetAsync(address, cancel).ConfigureAwait(false);
+                HttpResponseMessage response;
+
+                if (_httpClient == null)
+                    using (HttpClient _httpClient = new HttpClient())
```
PinpointTownes: FWIW, this pattern looks weird and is rather inefficient, as `HttpClient` has been specifically designed to be cached and reused concurrently.

Instead, why not adding a static readonly `DefaultHttpClient` field and use it if no custom client has been provided? This way, you wouldn't have to dispose anything at all.

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -53,17 +52,12 @@ public HttpDocumentRetriever(HttpClient httpClient)
...
 
         /// <summary>
         /// Requires Https secure channel for sending requests.. This is turned ON by default for security reasons. It is RECOMMENDED that you do not allow retrieval from http addresses by default.
         /// </summary>
-        public bool RequireHttps
-        {
-            get { return _requireHttps; }
-            set { _requireHttps = value; }
-        }
+        public bool RequireHttps { get; set; } = true;
```
PinpointTownes: :clap: :clap:

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
tushargupta51: Agree. I will update it to use a static httpClient object instead of either creating or disposing it everytime.

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -40,11 +40,12 @@ namespace Microsoft.IdentityModel.Protocols
     /// </summary>
     public class HttpDocumentRetriever : IDocumentRetriever
     {
-        private readonly HttpClient _httpClient;
-        private bool _requireHttps;
+        private HttpClient _httpClient;
+        private static HttpClient _defaultHttpClient = new HttpClient();
```
PinpointTownes: nit: `readonly`?

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -53,17 +54,12 @@ public HttpDocumentRetriever(HttpClient httpClient)
                 throw LogHelper.LogArgumentNullException("httpClient");
 
             _httpClient = httpClient;
-            _requireHttps = true;
```
brentschmaltz: For complete correctness, we may want to think about distinguishing between our http client and this one passed in. The owner of httpClient could call dispose after calling HttpDocumentRetriever(HttpClient httpClient). I agree is it very unlikely and indicates a logic programming error. However, our logic could be: (local)httpclient = _httpClient ?? _defaultHttpClient.

tushargupta51: @brentschmaltz updated

brentschmaltz:
:shipit:

Issue #335, opened 2016-01-05
Add extensibility for different RSA platforms


brentschmaltz:
Our current implementation of AsymmetricSignatureProvider.ResolveDotNetCoreAsymmetricAlgorithm assumes RSAOPenSsl is the right thing if !Windows. Need some kind of extensibility here.

if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                    _rsa = new RSACng();
                else
                    _rsa = new RSAOpenSsl();

tushargupta51:
@brentschmaltz Currently, if a user wants to extend the `AsymmetricSignatureProvider`, he has to override the implementation in SignatureProviderFactory (or SecurityKey.GetSignatureProvider) to return the custom provider and plug in their own implementation of algorithms.

What do you mean by adding extensibility here? Making ResolveDotNetCoreAsymmetricAlgorithm  and ResolveDotNetDesktopAsymmetricAlgorithm  methods virtual (currently they are private)?

brentschmaltz:
We need a way to allow users to create the RSA type they need. I don't think a static method Rsa.Create() will work as it is not dynamic. We can have a virtual ResolveAsymmetricAlgorithm (btw the method for DOTNET5_4 and other can be the same name) OR introduce a delegate: AsymmetricAlgorithmResolver. The tension of creating a clean design is the result of  AymmetricAlgorithm cannot do much by default and we always have to upcast. 

We may need to invent that type (similar to the Formatter and Deformatters).

bartonjs:
https://github.com/dotnet/corefx/issues/2953 was fixed, so the platform RSA.Create() is now available.  While it currently does not allow for fully-arbitrary RSA implementations, it does leave the decision of "what is the right RSA for this platform?" in corefx.

tushargupta51:
@bartonjs are these changes part of rc2 packages or rc3? We are using the rc2 version of .net packages for next release.

bartonjs:
@tushargupta51 Looks like both? :smile:  I just checked in rc2-23818, and I see it there.  Lemme know if you're not encountering a path-to-success.

PR #339, opened 2016-01-13
Remove validateSignature flag and make ResolveAsymmetricAlgorithm extensible

### LINKED ISSUE #332: TokenValidationParameters.RequireSignedTokens and ValidateSignature ambiguous
Two properties on TVP: RequireSignedTokens and ValidateSignature are ambiguous.

RequireSignedTokens enforce that tokens should have a signature whereas ValidateSignature controls whether signature will be validated during validation process or not. Both are set to true by default.

I don't think there is a useful scenario where signed tokens are required but signature validation is optional. In the other case, where signed tokens are not required, we may or may not want to validate a signature if it is present and ValidateSignature flag makes sense in that case.

Also, in the current code, ValidateSignature flag takes precedence. If RequireSignedTokens is set to true and ValidateSignature is false, we skip signature validation which does not seem like the right behavior.

Do we want to keep both flags? I guess we should get rid of ValidateSignature flag.

If we want to keep both of them, we should skip signature validation only if both flags are turned off (which can again be confusing to users if they only set the ValidateSignature flag to false and expect the signature to be not validated).

tushargupta51:
Fixes #332

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -121,11 +121,7 @@ namespace Microsoft.IdentityModel.Tokens
             if (willCreateSignatures && !key.HasPrivateKey)
                 throw LogHelper.LogException<InvalidOperationException>(LogMessages.IDX10638, key);
 
-#if DOTNET5_4
-            ResolveDotNetCoreAsymmetricAlgorithm(key, algorithm, willCreateSignatures);
-#else
-            ResolveDotNetDesktopAsymmetricAlgorithm(key, algorithm, willCreateSignatures);
-#endif
+            ResolveAsymmetricAlgorithm(key, algorithm, willCreateSignatures);
         }
 
         /// <summary>
@@ -181,7 +177,7 @@ namespace Microsoft.IdentityModel.Tokens
             throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10640, algorithm);
         }
 
-        private void ResolveDotNetCoreAsymmetricAlgorithm(AsymmetricSecurityKey key, string algorithm, bool willCreateSignatures)
+        private void ResolveAsymmetricAlgorithm(AsymmetricSecurityKey key, string algorithm, bool willCreateSignatures)
         {
             _hashAlgorithm = GetHashAlgorithmName(algorithm);
             RsaSecurityKey rsaKey = key as RsaSecurityKey;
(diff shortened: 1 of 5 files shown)


tushargupta51:
@brentschmaltz

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -181,7 +177,7 @@ protected virtual HashAlgorithmName GetHashAlgorithmName(string algorithm)
             throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10640, algorithm);
         }
 
-        private void ResolveDotNetCoreAsymmetricAlgorithm(AsymmetricSecurityKey key, string algorithm, bool willCreateSignatures)
+        public virtual void ResolveAsymmetricAlgorithm(AsymmetricSecurityKey key, string algorithm, bool willCreateSignatures)
```
PinpointTownes: Hmm, I'm not sure to understand what's the point of making `ResolveAsymmetricAlgorithm` public and virtual if it still returns void and if the fields are all private :fearful:

tushargupta51: Hmm, this does not look right and it makes me think if we even need to make this virtual. There are multiple ways to allow users to provide their own Rsa/Ecdsa implementation: 
1. SecurityKey constructor
2. SecurityKey.GetSignatureProvider override
3. CryptoProviderFactory.CreateForSigning/Verfiying override

All of them will impact the custom implementation/override at different level. CryptoProviderFactory having the most broader effect.

Also, there does not seem to be a clean way to provide an override at this level. All fields are private and we cannot have this method return `AsymmetricAlgorithm` type since that is too vanilla. I am thinking its best to keep it as it was before.
@brentschmaltz thoughts?

PinpointTownes: :+1:

brentschmaltz: Guys, how about the method returns
1. an AsymmetricAlgorithm, this would result in the need to cast to something we understand such as RSA or ECDsa.
2. Invent an interface similar to Deformater / Formater that can sign or verify

tushargupta51: I agree to taking virtual off now and open an issue for this. I am not convinced if we need an extensibility point here. Lets discuss as a separate issue. I will revert back the changes.

tushargupta51: Lets discuss this on https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/335

Issue #337, opened 2016-01-11
Constructors on RsaSecurityKey and ECDsaSecurityKey classes


tushargupta51:
From a comment on the PR #334: this is about the usefulness of following 2 constructors:
1. ECDsaSecuirytKey(byte[] blob, CngKeyBlobFormat blobFormat)
2. RsaSecurityKey(RsaParameters rsaParameters)
CngKey and RsaParameters properties are used by AsymmetricSignatureProvider to create ECDsa and Rsa instances respectively when keys are constructed using components and not the crypto provider instances. Taking these properties off of the class only makes sense when we remove the constructors too, otherwise those constructors wont offer any value.

Investigate whether having these constructors on the classes are useful or not:
1. Should we have `CngKey` property on the `ECDsaSecurityKey` class?
2. Should we have `RsaParamaters` property on `RsaSecurityKey` class?

cc: @brentschmaltz @pinpointtownes

PinpointTownes:
> Should we have RsaParamaters property on RsaSecurityKey class?

Ideally, I'd say no. That said, I agree it's a really convenient way to instantiate a RSA key, so maybe it would be worth keeping it as a static factory method:

```csharp
public static RsaSecurityKey Create(RSAParameters parameters) {
    var algorithm = // Since RSA.Create() is not ready yet,
                    // determine the best implementation locally:
                    // RSACryptoServiceProvider on Windows desktop and Mono,
                    // RSACng on CoreCLR for Windows and RSAOpenSsl on CoreCLR for Linux.
    algorithm.ImportParameters(parameters);
    return new RsaSecurityKey(algorithm);
}
```

> Should we have CngKey property on the ECDsaSecurityKey class?

CngKey is totally Windows-specific (it won't work at all everywhere else), so I'm not sure it's really pertinent to have it in the public contract. Sadly, we don't have an `ECDsaParameters` equivalent to easily import an elliptic curve key (from a JWK set for instance). Even on Windows, you must hack a blob array to import a CNG key (which is probably why you've decided to close https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/62).

/cc @bartonjs

bartonjs:
[...]

PR #340, opened 2016-01-20
Adding support to Crypto framework

tushargupta51:
1. Adding support signing/verifying using json web keys
2. Removing platform dependent public api from ecdsasecuritykey
2. Cleanup securityAlgorithms/JwtAlgorithms

Fixing #337, #315, #330


tushargupta51:
@brentschmaltz

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -433,24 +433,27 @@ public virtual HashAlgorithm GetHashAlgorithm(string algorithm)
                     case SecurityAlgorithms.ECDSA_SHA256:
                     case SecurityAlgorithms.HMAC_SHA256:
                     case SecurityAlgorithms.RSA_SHA256:
+                    case SecurityAlgorithms.RSASSA_PSS_SHA256:
                     case SecurityAlgorithms.RsaSha256Signature:
-                    case SecurityAlgorithms.PS256:
+                    case SecurityAlgorithms.EcdsaSha256Signature:
```
PinpointTownes: It's **really** time to fix these inconsistent names (XMLDSig URN, JWA short names...), really :smile:

src/Microsoft.IdentityModel.Tokens/AsymmetricSecurityKey.cs
```diff
@@ -29,8 +29,7 @@ namespace Microsoft.IdentityModel.Tokens
 {
     public abstract class AsymmetricSecurityKey : SecurityKey
     {
-        public abstract bool HasPrivateKey { get; }
+        public abstract bool? HasPrivateKey { get; }
```
PinpointTownes: Heh, why returning `bool?` here? It's going to be painful to use...

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -230,9 +230,28 @@ private void ResolveDotNetCoreAsymmetricAlgorithm(AsymmetricSecurityKey key, str
...
-                    _ecdsaCng = new ECDsaCng(ecdsaKey.CngKey);
+                    _rsa = new RSAOpenSsl();
+
+                _rsa.ImportParameters(parameters);
+                _disposeRsa = true;
+                return;
+            }
+            else if (webKey.Kty == JsonWebAlgorithmsKeyTypes.EllipticCurve)
+            {
+                CreateECDsaFromJsonWebKey(webKey, willCreateSignatures);
```
PinpointTownes: This code path is CNG-specific, it will crash on other platforms. You absolutely need to ensure we're on Windows at this point.

tushargupta51: Yes, creating signature providers from ec parameters is only supported on windows currently. We will wait for relevant cross plat support to show up in .net (creating ecdsaopenssl() from a key). I can add platformnotsupported exception here.

PinpointTownes: Yes, please.

src/Microsoft.IdentityModel.Tokens/AsymmetricSecurityKey.cs
tushargupta51: This is a mistake. I wanted the public API to return bool only and have the internal implementation use bool?. I will correct this. Thanks for pointing out.

PinpointTownes: Ah, thanks :+1:

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -315,6 +345,161 @@ private void ResolveDotNetDesktopAsymmetricAlgorithm(AsymmetricSecurityKey key,
...
 #endif
 
+        private RSAParameters CreateRsaParametersFromJsonWebKey(JsonWebKey webKey, bool willCreateSignatures)
+        {
+            RSAParameters parameters;
+            if (willCreateSignatures)
+            {
+                parameters = new RSAParameters()
+                {
+                    D = Base64UrlEncoder.DecodeBytes(webKey.D),
```
PinpointTownes: Is `Base64UrlEncoder.DecodeBytes` guarded against null values?
A malformed JWKS response might miss one of these precious parameters and we should have a nice exception path for that.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -25,18 +25,18 @@
...
 
-namespace System.IdentityModel.Tokens.Jwt
+namespace Microsoft.IdentityModel.Tokens
 {
     /// <summary>
     /// Represents a Json Web Key as defined in http://tools.ietf.org/html/rfc7517.
     /// </summary>
     [JsonObject]
-    public class JsonWebKey
+    public class JsonWebKey : AsymmetricSecurityKey
```
PinpointTownes: IMHO, it feels extremely artificial as symmetric keys can also be represented as JWK keys.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -143,12 +142,6 @@ private void Copy(JsonWebKey key)
         /// </summary>
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore, NullValueHandling = NullValueHandling.Ignore, PropertyName = JsonWebKeyParameterNames.E, Required = Required.Default)]
         public string E { get; set; }
-        /// <summary>
-        /// Gets or sets the 'k' (Symmetric - Key Value)..
-        /// </summary>
-        /// Base64urlEncoding
-        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore, NullValueHandling = NullValueHandling.Ignore, PropertyName = JsonWebKeyParameterNames.K, Required = Required.Default)]
-        public string K { get; set; }
```
PinpointTownes: Hahaha, nice trick... :smile:

src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
```diff
@@ -39,18 +39,39 @@ public static class SecurityAlgorithms
...
+        public const string EcdsaSha384Signature = "http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha384";
+        public const string EcdsaSha512Signature = "http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha512";
+
+        // See https://tools.ietf.org/html/rfc6931#section-2.3.10
+        public const string RsaSsaPssSha256Signature = "http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1";
+        public const string RsaSsaPssSha384Signature = "http://www.w3.org/2007/05/xmldsig-more#sha384-rsa-MGF1";
+        public const string RsaSsaPssSha512Signature = "http://www.w3.org/2007/05/xmldsig-more#sha512-rsa-MGF1";
+
+        /// see: http://tools.ietf.org/html/rfc7518#section-3
         public const string ECDSA_SHA256 = "ES256";
```
PinpointTownes: FWIW, these constants should be in a separate JWA-specific class.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
tushargupta51: Yes, they can be represented as JWK. But it is not very common to go to a metadata endpoint to retrieve a symmetric key. This usually has value in case of asymmetric keys. I had to split out the JsonWebKey into asymmetric and symmteric parts, because keeping them together was messing the OM when trying to create signature provider from JsonWebKey. While doing so, I didn't see enough value in supporting symmetric keys as JWK, and hence just kept the asymmetric parts in it. Do you think retrieving symmetric JWK from metadata endpoint is a common scenario? In that case, I might have to consider renaming those to AsymmetricJsonWebKey and SymmetricJsonWebKey.

PinpointTownes: > Do you think retrieving symmetric JWK from metadata endpoint is a common scenario?

Nah, it's not a common scenario (it would probably have to require client authentication to avoid leaking the key material). But the real question is: why do you want to make `JsonWebKey` inherit from `SecurityKey`?

src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
tushargupta51: Signature providers supports both jwt algorithms and xmldsig algorithms, due to which we have short alg names in SecurityAlgorithms. And signature providers cannot use JwtAlgorithms to avoid circular dependency. That's why the constants in JwtAlgorithms class were redundant.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
tushargupta51: It will throw ArgumentNullException.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
tushargupta51: To enable creating signature providers directly from JsonWebKey. Currently, we have `GetSigningKeys` method on JsonWebKeySet that translates a JWK to a `SecurityKey`. But that is not complete and we are considering getting rid of it. Making JWK inherit from `SecurityKey` will allow creating signature providers from these keys directly. Also, this enables users to obtain parameters from metadata endpoint and hook in their own signature provider (one that works on unix) by overriding `GetSignatureProvider`.

PinpointTownes: Dunno how I feel about that, still seems a bit artificial to me. Have you considered introducing a `JsonWebKey` -> `SecurityKey` in the JWT package instead?

Also, though symmetric JWK are unlikely to be exposed by a metadata endpoint, developers may still want to use symmetric JWK locally (it's a cool way to serialize a key in a JSON file)

src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
PinpointTownes: But having multiple constants representing the same algorithms but with different names seems odd, hence my suggestion about opting for a single set (XMLDSig being the de-facto default in IdentityModel <5) and providing XMLDSig/JWA mappings.

Not to mention that these JWA constants don't really respect the C# naming conventions.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
brentschmaltz: @PinpointTownes where the conversion takes place is what this PR is about. As Tushar mentioned the JsonWebKeySet currently does the translation. Asp.Net uses OIDCConfiguration.SigningKeys (ICollection<SecurityKey>) when validating a token. As Tushar mentioned we are trying to provide a way for Linux users to take control. It seemed to be natural in the SignatureProvider.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
brentschmaltz: I am thinking we leave the key as is and support Symmetric. The SignatureProvider can examine and return the appropriate provider.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -314,6 +402,167 @@ private void ResolveAsymmetricAlgorithm(AsymmetricSecurityKey key, string algori
...
+
+            uint dwMagic = GetMagicValue(webKey.Crv, willCreateSignatures);
+            uint cbKey = GetKeyByteCount(webKey.Crv);
+            byte[] keyBlob;
+            if (willCreateSignatures)
+                keyBlob = new byte[3 * cbKey + 2 * Marshal.SizeOf<uint>()];
+            else
+                keyBlob = new byte[2 * cbKey + 2 * Marshal.SizeOf<uint>()];
+
+            GCHandle keyBlobHandle = GCHandle.Alloc(keyBlob, GCHandleType.Pinned);
```
brentschmaltz: We are going to need some Try / Finally logic here. this is kind-of leaky.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -282,6 +283,23 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
...
+        {
+            get
+            {
+                return _cryptoProviderFactory;
+            }
+            set
+            {
+                if (value == null)
+                    throw LogHelper.LogException<ArgumentNullException>(LogMessages.IDX10001, nameof(CryptoProviderFactory));
+                _cryptoProviderFactory = value;
```
brentschmaltz: our standard style is to have a line after a clause, in this case it is an 'if'.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -25,18 +25,18 @@
...
 
-namespace System.IdentityModel.Tokens.Jwt
+namespace Microsoft.IdentityModel.Tokens
 {
     /// <summary>
     /// Represents a Json Web Key as defined in http://tools.ietf.org/html/rfc7517.
     /// </summary>
     [JsonObject]
-    public class JsonWebKey
+    public class JsonWebKey : SecurityKey
```
brentschmaltz: CryptoProviderFactory.CreateProvider needs to handle JsonWebKeys. It currently only handles Asymmetric or Symmetric keys.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -273,5 +274,40 @@ public IList<string> X5c
         /// <remarks> value is formated as: Base64urlEncoding</remarks>
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore, NullValueHandling = NullValueHandling.Ignore, PropertyName = JsonWebKeyParameterNames.Y, Required = Required.Default)]
         public string Y { get; set; }
+
+        public override int KeySize
```
brentschmaltz: Is there a KeySize for SymmetricKeys?

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -118,10 +129,30 @@ public AsymmetricSignatureProvider(AsymmetricSecurityKey key, string algorithm,
...
+            if (willCreateSignatures && !HasPrivateKey(key))
                 throw LogHelper.LogException<InvalidOperationException>(LogMessages.IDX10638, key);
 
-            ResolveAsymmetricAlgorithm(key, algorithm, willCreateSignatures);
+            if (asymmetricAlgorithmResolver != null)
+            {
+                AsymmetricAlgorithm asymmetricAlgorithm = asymmetricAlgorithmResolver(key, algorithm, willCreateSignatures);
+#if DOTNET5_4
+                _rsa = asymmetricAlgorithm as RSA;
+                if (_rsa == null)
```
brentschmaltz: if _rsa != null, _ecdsa will always be null.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
brentschmaltz: Clone() needs to handle _cryptoProfiderFactory

src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
```diff
@@ -59,7 +60,7 @@ public CryptoProviderFactory CryptoProviderFactory
             {
                 if (value == null)
                 {
-                    throw new ArgumentNullException("value");
+                    throw LogHelper.LogArgumentNullException("value");
```
brentschmaltz: Rename GetSignatureProviderForValidating(string algorithm) -> GetSignatureProviderForVerifying(string algorithm)

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -146,30 +177,42 @@ public AsymmetricSignatureProvider(AsymmetricSecurityKey key, string algorithm,
...
             if (string.IsNullOrWhiteSpace(algorithm))
                 throw LogHelper.LogArgumentNullException("algorithm");
 
-
             switch (algorithm)
             {
-                case SecurityAlgorithms.SHA256:
-                case SecurityAlgorithms.ECDSA_SHA256:
-                case SecurityAlgorithms.RSA_SHA256:
+                case SecurityAlgorithms.Sha256:
```
brentschmaltz: Shaxxx are not signature algorithms, we should remove these.

src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
```diff
@@ -77,7 +77,15 @@ public SymmetricSignatureProvider(SymmetricSecurityKey key, string algorithm)
 
             try
             {
-                _keyedHash.Key = key.Key;
+                SymmetricSecurityKey symmetricSecurityKey = key as SymmetricSecurityKey;
```
brentschmaltz: Need error checking on SecurityKey type, make sure key material (JsonWebKey.K) is not null. 
We can pass key into GetKeyedHashAlgorithm and new up the HMACxxx with the key.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -314,6 +403,194 @@ private void ResolveAsymmetricAlgorithm(AsymmetricSecurityKey key, string algori
...
+            if (webKey.Crv == null)
+                throw LogHelper.LogArgumentNullException(nameof(webKey.Crv));
+
+            if (webKey.X == null)
+                throw LogHelper.LogArgumentNullException(nameof(webKey.X));
+
+            if (webKey.Y == null)
+                throw LogHelper.LogArgumentNullException(nameof(webKey.Y));
+
+            GCHandle keyBlobHandle = new GCHandle();
```
brentschmaltz: Do we need to 'new GCHandle()' here? Why not just 'null'; keyBlobHandle is set with GCHandle.Alloc(...) below.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -314,6 +403,194 @@ private void ResolveAsymmetricAlgorithm(AsymmetricSecurityKey key, string algori
...
+        }
+
+        /// <summary>
+        /// Returns the size of key in bytes
+        /// </summary>
+        /// <param name="curveId">represents ecdsa curve -P256, P384, P512</param>
+        /// <returns>size of the key in bytes</returns>
+        private uint GetKeyByteCount(string curveId)
+        {
+            if (string.IsNullOrEmpty(curveId))
```
brentschmaltz: Since this method is private we don't need this 'null' check.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -314,6 +403,194 @@ private void ResolveAsymmetricAlgorithm(AsymmetricSecurityKey key, string algori
...
+                default:
+                    throw LogHelper.LogException<ArgumentException>(LogMessages.IDX10645, curveId);
+            }
+            return (uint)magicNumber;
+        }
+
+        /// <summary>
+        /// Magic numbers identifying ECDSA blob types
+        /// </summary>
+        internal enum KeyBlobMagicNumber : int
```
brentschmaltz: This enum should be uint.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -314,6 +403,194 @@ private void ResolveAsymmetricAlgorithm(AsymmetricSecurityKey key, string algori
...
+        private uint GetMagicValue(string curveId, bool willCreateSignatures)
+        {
+            if (string.IsNullOrEmpty(curveId))
+                throw LogHelper.LogArgumentNullException(nameof(curveId));
+
+            KeyBlobMagicNumber magicNumber;
+            switch (curveId)
+            {
+                case JsonWebKeyECTypes.P256:
+                    if (willCreateSignatures)
```
brentschmaltz: If enum is uint, then you can do away with local variable (magicNumber) and just return.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -314,6 +403,194 @@ private void ResolveAsymmetricAlgorithm(AsymmetricSecurityKey key, string algori
...
+            if (webKey.Y == null)
+                throw LogHelper.LogArgumentNullException(nameof(webKey.Y));
+
+            GCHandle keyBlobHandle = new GCHandle();
+            try
+            {
+                uint dwMagic = GetMagicValue(webKey.Crv, willCreateSignatures);
+                uint cbKey = GetKeyByteCount(webKey.Crv);
+                byte[] keyBlob;
+                if (willCreateSignatures)
```
brentschmaltz: We need to open a bug to investigate consolidating all the writes using Marshal. I ran some quick tests and see consolidating the bytes using Buffer.BlockCopy is about 25 - 30% faster:

Time for: '100000' loops using CreateECDsaUsingBlockCopy: '00:00:00.1730127'
Time for: '100000' loops using CreateECDsaUsingMarshalByByte: '00:00:00.2140152'

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -314,6 +403,194 @@ private void ResolveAsymmetricAlgorithm(AsymmetricSecurityKey key, string algori
...
+                }
+                else
+                {
+                    Marshal.Copy(keyBlobPtr, keyBlob, 0, keyBlob.Length);
+                    using (CngKey cngKey = CngKey.Import(keyBlob, CngKeyBlobFormat.EccPublicBlob))
+                    {
+                        _ecdsa = new ECDsaCng(cngKey);
+                    }
+                }
+                keyBlobHandle.Free();
```
brentschmaltz: The finally clause will also Free. See: https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.gchandle.free(v=vs.100).aspx

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
tushargupta51: Why remove this if we want to caution future developers that is not checked for null?

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
tushargupta51: I will open the bug and we can fix this post Rc2 I think.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
tushargupta51: I will open the bug and we can fix this post Rc2.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
tushargupta51: I changed the enum to uint but it still gives the error then an explicit conversion is required.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
tushargupta51: I forgot to take this off when added the .free() in finally. It is removed now.

brentschmaltz:
:shipit:

PR #341, opened 2016-01-26
Skip C_Hash validation if c_hash is not present.

ajadex:
The actual [specification] (http://openid.net/specs/openid-connect-core-1_0.html#CodeValidation) states: The value of c_hash in the ID Token MUST match the value produced in the previous step **if** c_hash is present in the ID Token.

This validation was not taking into account the condition **IF c_hash is present** and is throwing an exception regardless.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocol.Extensions/OpenIdConnectProtocolValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocol.Extensions/OpenIdConnectProtocolValidator.cs
@@ -247,7 +247,7 @@ namespace Microsoft.IdentityModel.Protocols
 
             if (!jwt.Payload.ContainsKey(JwtRegisteredClaimNames.CHash))
             {
-                throw new OpenIdConnectProtocolInvalidCHashException(string.Format(CultureInfo.InvariantCulture, ErrorMessages.IDX10308, jwt));
+                return;
             }
 
             HashAlgorithm hashAlgorithm = null;


ajadex:
Makes sense. The same applies for when is a CODE Flow in which the id_token and the code token don't arrive together? Also, shouldn't this assumption (valid) be commented in the code or mentioned in the documentation. 

I'm using MitreID as the OpenIdConnect server and it's not serving the C_Hash in the id_token. Your input makes me wonder if I should get back to them and let them know about this missing part.

brentschmaltz:
@ajadex I would encourage you to ask MitreID about why the don't have c_hash claim in the id_token. It is possible to have a code arrive, redeem that for an access_token + id_token. In this case the id_token has an 'at_hash' claim that is validated same as 'c_hash'. These are important as these artifacts flow through the user-agent and can be replaced. Signing of the individual pieces doesn't guard against swapping.

Have a look at:
http://openid.net/specs/openid-connect-core-1_0.html#CodeValidation

ajadex:
Thanks. Closing the PR.

brockallen:
@ajadex would you mind following up here once you find out about MitreID? I'm curious about the answer.

ajadex:
@brentschmaltz, @brockallen, MitreID hasn't implemented the **Hybrid flow** with response types `code id_token` yet. They only return `at_hash` if `token` is in the response type which doesn't appear to compromise security since in a pure `code` flow scenario the artifacts never flow through the user-agent because the redemption of `code` for the  ID and Access Tokens happens on the server.

Section 3.1 (Code Flow) of the OpenId Connect Specification makes no mention of `c_hash` for code validation and the `at_hash` ID Token claim is optional. (See 3.1.3.6.  ID Token, http://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken).

leastprivilege:
As you can see here

https://openid.net/certification/

MitreID is not certified for implicit and hybrid flow. So there implementation is probably not complete / correct (yet).

leastprivilege:
..and not using c_hash and at_hash - you are neglecting the security benefits of hybrid flow.

ajadex:
@leastprivilege you are absolutely right, but those security concerns are not present since I'm only using the **code** flow, not the **hybrid** one. My take is that for `response_type=code` these hash validations need to be relaxed.

leastprivilege:
code has no id_token - and hence no c_hash. So yes - in that case there is no hash validation needed.

But you could argue that code id_token is the superior and more secure flow anyways. I wouldn't mind not supporting raw code at all.

ajadex:
`code id_token` is superior in terms of performance and simplicity but I wouldn't argue it's a more secure flow. To the contrary, the only flow that can deliver the `access_token` directly to the client without the compromise of passing through the client's user agent is the `code` flow (therefore the lack of hash checks).

This is stated in the OAuth specification [http://tools.ietf.org/html/rfc6749#section-1.3.1](http://tools.ietf.org/html/rfc6749#section-1.3.1). You can find there the security benefits of the `code` flow over `hybrid` and `implicit`.

leastprivilege:
you got that wrong. There is no access token in `code id_token`.
`code id_token` has all the properties of `code` + the validation of the code itself via c_hash.
`code` on its own has some potential issues - e.g. the cut'n paste attack.

So wherever you would use `code` - you can actually step up security by changing to hybrid flow.

http://nat.sakimura.org/2016/01/25/cut-and-pasted-code-attack-in-oauth-2-0-rfc6749/

ajadex:
Interesting. Glad you pointed those issues with the `code` only approach.

PR #344, opened 2016-02-17
updating to use dotnet cli

tushargupta51:


tushargupta51:
@brentschmaltz

brentschmaltz:
:shipit: I sent some questions over email. But if it builds, we can do some work again :-)

PR #345, opened 2016-02-18
set default built configuration to Release

tushargupta51:


tushargupta51:
@brentschmaltz

brentschmaltz:
:shipit:

PR #346, opened 2016-02-19
fix build failure

tushargupta51:
using latest dotnet cli version and reverting "Release" to "Debug" until https://github.com/aspnet/Universe/issues/389 is addressed.


tushargupta51:
@brentschmaltz

brentschmaltz:
:shipit:

PR #348, opened 2016-02-22
adding rc2-* to package versions

tushargupta51:


tushargupta51:
@brentschmaltz

brentschmaltz:
:shipit:

Issue #304, opened 2015-11-02
Change #defines for new targets


brentschmaltz:
from: https://github.com/aspnet/Announcements/issues/98

#if DNX451
        // whatever 
#endif 
...
#if DNXCORE50
        // whatever 
#endif 

Change that to:

#if NET451
        // whatever 
#endif 
...
#if DOTNET5_4
        // whatever 
#endif

brentschmaltz:
Moved to RC2 to revisit as Tushar pointed out that since these are tests, it is OK for now.

tushargupta51:
dnxcore50 changes to netstandardapp1.5

PR #351, opened 2016-03-01
updating to use TFM

tushargupta51:
@brentschmaltz


brentschmaltz:
@tushargupta51 will asp.net be able to consume these assemblies?

tushargupta51:
@brentschmaltz only after they update to new TFM.

brentschmaltz:
:shipit:

tushargupta51:
fixes #304

PR #352, opened 2016-03-11
updating build scripts to use new korebuild and adding a hack to fix …

tushargupta51:
…build


tushargupta51:
@brentschmaltz

brentschmaltz:
:shipit:

PR #357, opened 2016-04-01
Fixing Rsa Upleveling and cleanup

### LINKED ISSUE #353: Revert temporary update to project.json
Revert this line: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/352/files#diff-aa5cb96b60346374809bbffafe4dd2b2R16

There is a bug (https://github.com/NuGet/Home/issues/2191) in nuget.common where it pulls in old .net packages which fails the build. This line is a workaround to that, it forced to pull in the latest version of System.Diagnostics.Process package.

tushargupta51:
1. #353 Removing this workaround since nuget packages are being.
2. #358 : upleveling of rsacsp.
3. Updating testBuild.cmd to run tests for both net451 and netstandardapp1.5

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
+++ b/src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
@@ -39,8 +39,12 @@ namespace Microsoft.IdentityModel.Tokens
     {
         private const int PROV_RSA_AES = 24;    // CryptoApi provider type for an RSA provider supporting sha-256 digital signatures
 
+        // CryptoApi provider type for an RSA provider only supporting sha1 digital signatures
+        private const int PROV_RSA_FULL = 1;
+        private const int PROV_RSA_SCHANNEL = 12;
+
         private bool _disposed;
-        
+
         // Only dispose of the RsaCryptoServiceProvider object if we created a new instance that supports SHA-256,
         // otherwise do not disposed of the referenced RsaCryptoServiceProvider
         private bool _disposeRsa;
@@ -52,24 +56,20 @@ namespace Microsoft.IdentityModel.Tokens
                 throw LogHelper.LogArgumentNullException("rsa");
 
             //
-            // If the provider does not understand SHA256, 
-            // replace it with one that does.
-            //
-            if (rsa.CspKeyContainerInfo.ProviderType != PROV_RSA_AES)
+            // Level up the provider type only if:
+            // 1. it is PROV_RSA_FULL or PROV_RSA_SCHANNEL which denote CSPs that only understand Sha1 algorithms
+            // 2. it is not associated with a hardware key
+            if ((rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_FULL || rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_SCHANNEL) && !rsa.CspKeyContainerInfo.HardwareDevice)
             {
                 CspParameters csp = new CspParameters();
                 csp.ProviderType = PROV_RSA_AES;
                 csp.KeyContainerName = rsa.CspKeyContainerInfo.KeyContainerName;
                 csp.KeyNumber = (int)rsa.CspKeyContainerInfo.KeyNumber;
                 if (rsa.CspKeyContainerInfo.MachineKeyStore)
-                {
                     csp.Flags = CspProviderFlags.UseMachineKeyStore;
-                }
 
-                //
                 // If UseExistingKey is not specified, the CLR will generate a key for a non-existent group.
                 // With this flag, a CryptographicException is thrown instead.
-                //
                 csp.Flags |= CspProviderFlags.UseExistingKey;
 
                 _rsa = new RSACryptoServiceProvider(csp);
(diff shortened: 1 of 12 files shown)


brentschmaltz:
:shipit:

tushargupta51:
@brentschmaltz @polita I have included the fix for rsa csp proxy in this branch itself (i know name of the branch is not entirely correct)

test/Microsoft.IdentityModel.Tokens.Tests/RsaCryptoServiceProviderProxyTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                rsaCsp.SignData(input, _hashAlgorithm);
+#endif
+                ee.ProcessNoException();
+            }
+            catch (Exception ex)
+            {
+                ee.ProcessException(ex);
+            }
+        }
+        private void SignData(RSACryptoServiceProviderProxy rsaCspProxy, ExpectedException ee)
```
brentschmaltz: nit: add a line here.

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
```diff
@@ -52,24 +56,20 @@ public RSACryptoServiceProviderProxy(RSACryptoServiceProvider rsa)
...
 
             //
-            // If the provider does not understand SHA256, 
-            // replace it with one that does.
-            //
-            if (rsa.CspKeyContainerInfo.ProviderType != PROV_RSA_AES)
+            // Level up the provider type only if:
+            // 1. it is PROV_RSA_FULL or PROV_RSA_SCHANNEL which denote CSPs that only understand Sha1 algorithms
+            // 2. it is not associated with a hardware key
+            if ((rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_FULL || rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_SCHANNEL) && !rsa.CspKeyContainerInfo.HardwareDevice)
```
brentschmaltz: Can we get in a test for hardware key? If not we should open an issue.

tushargupta51: I will open an issue.

brentschmaltz:
Have a look at how we handle X509SecurityKey.PrivateKey and X509SecurityKey.PublicKey. PrivateKey calls GetRSAPrivateKey(_certificate) when applicable. PublicKey does not. This forces X509.KeySize to call the extension.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fixNugetWorkaround/src/Microsoft.IdentityModel.Tokens/X509SecurityKey.cs#L63

AsymmetricSignatureProvider.ResolveAsymmetricAlgorithm also calls the extension.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fixNugetWorkaround/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs#L267

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
brentschmaltz: In Desktop, we only 'levelup' if the algorithm is RSASha2, here don't check the algorithm.

tushargupta51: because we don't support Sha1 in Wilson. For sha256, sha384, sha512 (the algorithms that we support in wilson), we need to uplevel. And in desktop WCF, we don't support Sha384 and Sha512, so uplevel only for sha256

brentschmaltz: we need to accept Sha1, but not create.

tushargupta51:
@brentschmaltz The difference is that GetRSAPrivateKey(_certificate) returns RSA object whereas Publickey property if of type `PublicKey`. I am ok to change the property type to AsymmetricAlgorithm (like PrivateKey). Couple of places where publicKey is used, we always use it as, x509Key.PublicKey.Key (which is the associated RSA object). I am not sure if there is any value in exposing publicKey as `PublicKey` object instead of `AsymmetricAlgorithm`. If we agree to make that change, I will update the code.

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
tushargupta51: We agreed earlier to not support Sha1 even for validation, I have sent you an email. Currently, when creating signature providers (either for signing or validation), we throw "Algorithm not supported" exception for sha1.

brentschmaltz:
PublicKey doesn't seem offer much value other than .Key which is an AsymmetricAlgorithm. I think it would smooth out the OM to have PublicKey return an AsymmetricAlgorithm.

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
brentschmaltz: I agree, we did make the decision that we only will NOT work with IdentityProviders that sign with RsaSha1. Users who want that support need to use extensibility.

brentschmaltz:
:shipit:

Issue #363, opened 2016-04-04
Rename JwtSecurityTokenHandler.CreateToken to CreateJwtSecurityToken


brentschmaltz:
This would provide consistency with JwtSecurityTokenHandler.CreateJwtSecurityToken(SecurityTokenDescritptor ...).

Issue #361, opened 2016-04-04
Reintroduce ClaimsIdentity.Actor support via SecurityTokenDescriptor


PinpointTownes:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/327#issuecomment-162992208

brentschmaltz:
@PinpointTownes excellent timing :-) as I was about to see if you had an opinion on if we should scrap the OM that uses IEnumerable<Claim> and go back to ClaimsIdentity on SecurityTokenDescriptor and CreateJwt(....). It does force users to create a ClaimsIdentity when all they want to pass are claims.


(proposing a rename below)
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/363

PinpointTownes:
Sounds like a good idea :+1:

brentschmaltz:
@PinpointTownes It would affect a lot of tests, would you be OK if we make this in the next milestone as we are trying to get RC2 out.

PinpointTownes:
I'm fine :smile: (even if breaking changes like this one should be added very early to avoid breaking too many people with the next release).

brentschmaltz:
I agree, early is better. I'll try and get it in, we have other teams taking a dependency so we need to get done soon.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/360 seems more important.

JohnGalt1717:
As it stands because of this and other issues you've got rogue nuget packages floating around that show up to automatically update any .NET 4.5+ solution but if you do, there is no way of getting the same functionality in alternative ways because of stuff that you've not included that falsely reports it should be there. (i.e. try creating openidconnect login support in mvc.net right now, you can't with these.

Given this is about the 10th such instance of .net core "upgrades" completely destroying .net 4.x full fat projects because of your nuget packages and your lack of testing and thought on .net full you guys need to get in a meeting immediately and fix the way you're doing this stuff:

1. If you're not providing the same functionality as the previous version AND not providing an alternative then you MUST NOT claim compatibility in your nuget package with .NET full.
2. If case #1 is true, then you SHOULD be creating a separate nuget package.

[...]

Issue #360, opened 2016-04-03
JwtSecurityTokenHandler.CreateJwt no longer sets JwtSecurityToken.SigningCredentials


PinpointTownes:
_Moved from https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/232._

______

The recent changes introduced by @brentschmaltz [in this commit](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/6cefb7cf902abdfb594fb785f30bec8e5af4059b) broke `CreateJwt`, which is no longer able to flow the `SigningCredentials` taken as a parameter to the `JwtSecurityToken` it returns due to the (weird) lossy serialization + deserialization process it's now using:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L371

This is obviously a serious regression since calling `WriteToken` to serialize the JWT now results in a compact JWT that lacks the signature.

/cc @brentschmaltz @tushargupta51

brentschmaltz:
@PinpointTownes thanks, will look into this.

PinpointTownes:
> Will put back the original behavior.

Thanks! :clap:

PinpointTownes:
@brentschmaltz people want your fix :trollface: 

![image](https://cloud.githubusercontent.com/assets/6998306/14323305/4e261590-fc20-11e5-92b0-02c1dc2cfebb.png)

brentschmaltz:
@PinpointTownes It was committed today. Should be in tonight's myget.release.

sunsided:
Which package would that be? Because force-refreshing my packages from `aspnetcidev` doesn't seem to do the trick.

capesean:
It's not in the `aspnetcidev` feed yet, it's only in the `azureadwebstacknightly`:
https://www.myget.org/gallery/azureadwebstacknightly

See the version for: `System.IdentityModel.Tokens.Jwt`

You'll need to add that feed to your nuget config if you want to pull that package from there before it gets into aspnetcidev.

sunsided:
Ah, well that both explains it and makes sense at the same time!

PinpointTownes:
[...]

PR #364, opened 2016-04-05
SecurityTokenDescriptor uses ClaimsIdentity instead of IEnumerable<Claim>

brentschmaltz:
Includes Issues:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/363
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/361
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/360

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
+++ b/src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
@@ -26,7 +26,6 @@
 //------------------------------------------------------------------------------
 
 using System;
-using System.Collections.Generic;
 using System.Security.Claims;
 
 namespace Microsoft.IdentityModel.Tokens
@@ -34,8 +33,6 @@ namespace Microsoft.IdentityModel.Tokens
     public class SecurityTokenDescriptor
     {
         public string Audience { get; set; }
-
-        public IEnumerable<Claim> Claims { get; set; }
 
         public DateTime? Expires { get; set; }
 
@@ -46,5 +43,7 @@ namespace Microsoft.IdentityModel.Tokens
         public DateTime? NotBefore { get; set; }
 
         public SigningCredentials SigningCredentials { get; set; }
+
+        public ClaimsIdentity Subject { get; set; }
     }
 }
--- a/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
@@ -386,6 +386,26 @@ namespace System.IdentityModel.Tokens.Jwt
             return CreateJwtSecurityTokenPrivate(issuer, audience, subject, notBefore, expires, issuedAt, signingCredentials);
         }
 
+        /// <summary>
+        /// Creates a Json Web Token (JWT).
+        /// </summary>
+        /// <param name="tokenDescriptor"> a <see cref="SecurityTokenDescriptor"/> that contains details of contents of the token.</param>
+        /// <remarks><see cref="SecurityTokenDescriptor.SigningCredentials"/> is used to sign <see cref="JwtSecurityToken.RawData"/>.</remarks>
+        public override SecurityToken CreateToken(SecurityTokenDescriptor tokenDescriptor)
+        {
+            if (tokenDescriptor == null)
+                throw LogHelper.LogArgumentNullException("tokenDescriptor");
+
+            return CreateJwtSecurityTokenPrivate(
+                tokenDescriptor.Issuer,
+                tokenDescriptor.Audience,
+                tokenDescriptor.Subject,
+                tokenDescriptor.NotBefore,
+                tokenDescriptor.Expires,
+                tokenDescriptor.IssuedAt,
+                tokenDescriptor.SigningCredentials);
+        }
+
         private JwtSecurityToken CreateJwtSecurityTokenPrivate(string issuer, string audience, ClaimsIdentity subject, DateTime? notBefore, DateTime? expires, DateTime? issuedAt, SigningCredentials signingCredentials)
         {
             if (SetDefaultTimesOnTokenCreation)
@@ -405,7 +425,7 @@ namespace System.IdentityModel.Tokens.Jwt
             JwtPayload payload = new JwtPayload(issuer, audience, (subject == null ? null : OutboundClaimTypeTransform(subject.Claims)), notBefore, expires, issuedAt);
             JwtHeader header = new JwtHeader(signingCredentials);
 
-            if (subject != null && subject.Actor != null)
+            if (subject?.Actor != null)
             {
                 payload.AddClaim(new Claim(JwtRegisteredClaimNames.Actort, this.CreateActorValue(subject.Actor)));
             }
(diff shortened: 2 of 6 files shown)


src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
```diff
@@ -46,5 +43,7 @@ public class SecurityTokenDescriptor
         public DateTime? NotBefore { get; set; }
 
         public SigningCredentials SigningCredentials { get; set; }
+
+        public ClaimsIdentity Subject { get; set; }
```
PinpointTownes: :clap:

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -387,9 +402,14 @@ public string CreateEncodedJwt(string issuer, string audience, IEnumerable<Claim
             }
 
             IdentityModelEventSource.Logger.WriteVerbose(LogMessages.IDX10721, (audience ?? "null"), (issuer ?? "null"));
-            JwtPayload payload = new JwtPayload(issuer, audience, (claims == null ? null : OutboundClaimTypeTransform(claims)), notBefore, expires, issuedAt);
+            JwtPayload payload = new JwtPayload(issuer, audience, (subject == null ? null : OutboundClaimTypeTransform(subject.Claims)), notBefore, expires, issuedAt);
             JwtHeader header = new JwtHeader(signingCredentials);
 
+            if (subject != null && subject.Actor != null)
```
PinpointTownes: You should add more C#6 love :smile: 

```
if (subject?.Actor != null)
```

PinpointTownes:
LGTM :+1: 

Quick remark: I can't see where `JwtSecurityTokenHandler` is overriding [`SecurityTokenHandler.CreateToken(SecurityTokenDescriptor)`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Tokens/SecurityTokenHandler.cs#L50). Is this an intentional omission?

brentschmaltz:
@PinpointTownes We did intentional leave out STH.CreateToken(SecurityTokenDescriptor). We were planning on bringing it back when SAML shows up. But now that I think about it, might as well add it now.

PinpointTownes:
> But now that I think about it, might as well add it now.

Yeah, specially since it would be trivial to implement.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -351,27 +376,37 @@ public JwtSecurityToken CreateJwtSecurityToken(SecurityTokenDescriptor tokenDesc
...
+        /// </summary>
+        /// <param name="tokenDescriptor"> a <see cref="SecurityTokenDescriptor"/> that contains details of contents of the token.</param>
+        /// <remarks><see cref="SecurityTokenDescriptor.SigningCredentials"/> is used to sign <see cref="JwtSecurityToken.RawData"/>.</remarks>
+        public override SecurityToken CreateToken(SecurityTokenDescriptor tokenDescriptor)
         {
-            return new JwtSecurityToken(CreateEncodedJwt(issuer, audience, claims, notBefore, expires, issuedAt, signingCredentials));
+            if (tokenDescriptor == null)
+                throw LogHelper.LogArgumentNullException("tokenDescriptor");
+
+            return CreateJwtSecurityTokenPrivate(
```
PinpointTownes: Note: you could simplify this method by simply calling `CreateJwtSecurityToken(tokenDescriptor)`.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -351,27 +376,37 @@ public JwtSecurityToken CreateJwtSecurityToken(SecurityTokenDescriptor tokenDesc
...
+        /// <summary>
+        /// Creates a Json Web Token (JWT).
+        /// </summary>
+        /// <param name="tokenDescriptor"> a <see cref="SecurityTokenDescriptor"/> that contains details of contents of the token.</param>
+        /// <remarks><see cref="SecurityTokenDescriptor.SigningCredentials"/> is used to sign <see cref="JwtSecurityToken.RawData"/>.</remarks>
+        public override SecurityToken CreateToken(SecurityTokenDescriptor tokenDescriptor)
         {
-            return new JwtSecurityToken(CreateEncodedJwt(issuer, audience, claims, notBefore, expires, issuedAt, signingCredentials));
+            if (tokenDescriptor == null)
+                throw LogHelper.LogArgumentNullException("tokenDescriptor");
```
PinpointTownes: nit: more `nameof` happiness to avoid the magic string? :smile:

PinpointTownes:
:shipit:

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -351,27 +376,37 @@ public JwtSecurityToken CreateJwtSecurityToken(SecurityTokenDescriptor tokenDesc
...
+            return CreateJwtSecurityTokenPrivate(issuer, audience, subject, notBefore, expires, issuedAt, signingCredentials);
         }
 
-        private JwtSecurityToken CreateJwt(string issuer, string audience, IEnumerable<Claim> claims, DateTime? notBefore, DateTime? expires, DateTime? issuedAt, SigningCredentials signingCredentials)
+        /// <summary>
+        /// Creates a Json Web Token (JWT).
+        /// </summary>
+        /// <param name="tokenDescriptor"> a <see cref="SecurityTokenDescriptor"/> that contains details of contents of the token.</param>
+        /// <remarks><see cref="SecurityTokenDescriptor.SigningCredentials"/> is used to sign <see cref="JwtSecurityToken.RawData"/>.</remarks>
+        public override SecurityToken CreateToken(SecurityTokenDescriptor tokenDescriptor)
```
tushargupta51: Why have both `CreateJwtSecurityToken` and `CreateToken` methods? It's confusing. I think we should just have `CreateToken` method, since that is the one on the base class. Users deriving this class will have 2 overrides available for same purpose.

PinpointTownes: AFAIK, you've decided to introduce `CreateJwtSecurityToken` to make directly dealing with `JwtSecurityTokenHandler` a bit easier. Unlike `CreateToken` (defined by the base blass), `CreateJwtSecurityToken` directly returns a `JwtSecurityToken`, which helps avoid casts.

To reduce the risk of confusion, you could seal `JwtSecurityToken.CreateToken`.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: I could, but it just adds another call on the stack. BUT less likely to break in the future.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: thanks, will do.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: Not a big fan of sealed. You are correct, the idea is to avoid the need for casting OR JwtSecurityToken.RawData OR the two step CreateJwtSecurityToken, WriteToken. These are remnants of XML.

Issue #378, opened 2016-04-28
JsonWebKey.KeySize throws ArgumentNullException if N, X, K is null.


brentschmaltz:
If any of these are null, just return 0.

brentschmaltz:
Bundled in this fix.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/382

PR #382, opened 2016-05-04
Json parsing

brentschmaltz:


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
```diff
@@ -40,57 +40,33 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
...
+            return new OpenIdConnectConfiguration(json);
         }
 
         /// <summary>
         /// Serializes the <see cref="OpenIdConnectConfiguration"/> object to a json string.
         /// </summary>
         /// <param name="configuration"><see cref="OpenIdConnectConfiguration"/> object to serialize.</param>
         /// <returns>json string representing the configuration object.</returns>
+        /// <exception cref="ArgumentNullException">if 'configuration' is null or empty.</exception>
+        /// <exception cref="ArgumentException">if 'json' fails to deserialize.</exception>
```
tushargupta51: The comment does not apply to this method.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
```diff
@@ -107,72 +83,35 @@ public OpenIdConnectConfiguration()
         /// Initializes an new instance of <see cref="OpenIdConnectConfiguration"/> from a json string.
         /// </summary>
         /// <param name="json">a json string containing the metadata</param>
-        /// <exception cref="ArgumentNullException">if 'json' is null or whitespace.</exception>
+        /// <exception cref="ArgumentNullException">if 'json' is null or empty.</exception>
         public OpenIdConnectConfiguration(string json)
+            : this()
```
tushargupta51: Why chaining the constructor?

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -155,21 +138,7 @@ private void Copy(JsonWebKey key)
...
-            set
-            {
-                if (value == null)
-                    throw LogHelper.LogException<ArgumentNullException>(LogMessages.IDX10001, "KeyOps");
-
-                foreach (string keyOp in value)
-                    _keyops.Add(keyOp);
-            }
-        }
+        public IList<string> KeyOps { get; private set; } = new List<string>();
```
tushargupta51: There is a change of behavior here, just want to make sure its intentional. Earlier, setting KeyOps would add values to existing array (always additive), now it replaces the whole list. The new behavior seems right to me, always additive does not make sense.

brentschmaltz: Yep, glad you noticed. This is the correct behavior always return a valid collection and have users modify it.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
brentschmaltz: Will fix.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
brentschmaltz: I will remove this.

tushargupta51:
:shipit:

brentschmaltz:
also fixed
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/378

Issue #336, opened 2016-01-06
Reintroduce SecurityKey.IsSupportedAlgorithm


PinpointTownes:
Unlike `System.IdentityModel` 4.5, the vNext equivalent doesn't have this precious API. To work around this limitation in ASOS, I had to create my own extension, but it has an obvious downside: it's tied to specific `SecurityKey` implementations, and defeats the whole purpose of having a base class that everyone can implement to support his own keys.

```csharp
internal static bool IsSupportedAlgorithm(this SecurityKey key, string algorithm) {
    if (key == null) {
        throw new ArgumentNullException(nameof(key));
    }

    if (string.IsNullOrEmpty(algorithm)) {
        throw new ArgumentNullException(nameof(algorithm));
    }

    // Note: SecurityKey currently doesn't support IsSupportedAlgorithm.
    // To work around this limitation, this static extension tries to
    // determine whether the given security key supports the specified
    // algorithm via CryptoConfig when available or using a pre-defined table.
    var symmetricSecurityKey = key as SymmetricSecurityKey;
    if (symmetricSecurityKey != null) {
#if DNX451
        try {
            var instance = CryptoConfig.CreateFromName(algorithm);
            if (instance is SymmetricAlgorithm || instance is KeyedHashAlgorithm) {
                return true;
            }
        }

        // Ignore the InvalidOperationException thrown by CryptoConfig.
        catch (InvalidOperationException) { }
#endif
        switch (algorithm) {
            case SecurityAlgorithms.HmacSha256Signature:
            case SecurityAlgorithms.HMAC_SHA256:
            case SecurityAlgorithms.HMAC_SHA384:
            case SecurityAlgorithms.HMAC_SHA512:
                return true;

            case SecurityAlgorithms.Aes128Encryption:
            case SecurityAlgorithms.Aes128KeyWrap:
                return symmetricSecurityKey.KeySize >= 128 &&
                       symmetricSecurityKey.KeySize <= 256;

            case SecurityAlgorithms.Aes192Encryption:
            case SecurityAlgorithms.Aes192KeyWrap:
                return symmetricSecurityKey.KeySize >= 192 &&
                       symmetricSecurityKey.KeySize <= 256;

            case SecurityAlgorithms.Aes256Encryption:
            case SecurityAlgorithms.Aes256KeyWrap:
                return symmetricSecurityKey.KeySize == 256;

            default:
                return false;
        }
    }

[...]

PR #386, opened 2016-05-05
Moving IsSupportedAlgorithm api from SignatureProvider to SecurityKey

tushargupta51:
#336

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
+++ b/src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
@@ -25,6 +25,7 @@
 //
 //------------------------------------------------------------------------------
 
+using System;
 using System.Security.Cryptography;
 using Microsoft.IdentityModel.Logging;
 
@@ -87,6 +88,26 @@ namespace Microsoft.IdentityModel.Tokens
                 return CryptoProviderFactory.CreateForVerifying(this, algorithm);
             else
                 return CryptoProviderFactory.CreateForSigning(this, algorithm);
+        }
+
+        public override bool IsSupportedAlgorithm(string algorithm)
+        {
+            if (string.IsNullOrEmpty(algorithm))
+                return false;
+
+            switch (algorithm)
+            {
+                case SecurityAlgorithms.EcdsaSha256:
+                case SecurityAlgorithms.EcdsaSha384:
+                case SecurityAlgorithms.EcdsaSha512:
+                case SecurityAlgorithms.EcdsaSha256Signature:
+                case SecurityAlgorithms.EcdsaSha384Signature:
+                case SecurityAlgorithms.EcdsaSha512Signature:
+                    return true;
+
+                default:
+                    return false;
+            }
         }
     }
 }
--- a/src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
+++ b/src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
@@ -309,5 +309,63 @@ namespace Microsoft.IdentityModel.Tokens
             else
                 return CryptoProviderFactory.CreateForSigning(this, algorithm);
         }
+
+        public override bool IsSupportedAlgorithm(string algorithm)
+        {
+            if (string.IsNullOrEmpty(algorithm))
+                return false;
+
+            if (Kty == JsonWebAlgorithmsKeyTypes.RSA)
+            {
+                switch (algorithm)
+                {
+                    case SecurityAlgorithms.RsaSha256:
+                    case SecurityAlgorithms.RsaSha384:
+                    case SecurityAlgorithms.RsaSha512:
+                    case SecurityAlgorithms.RsaSha256Signature:
+                    case SecurityAlgorithms.RsaSha384Signature:
+                    case SecurityAlgorithms.RsaSha512Signature:
+                        return true;
+
+                    default:
+                        return false;
+                }
+            }
+            else if (Kty == JsonWebAlgorithmsKeyTypes.EllipticCurve)
+            {
+                switch (algorithm)
+                {
+                    case SecurityAlgorithms.EcdsaSha256:
+                    case SecurityAlgorithms.EcdsaSha384:
+                    case SecurityAlgorithms.EcdsaSha512:
+                    case SecurityAlgorithms.EcdsaSha256Signature:
+                    case SecurityAlgorithms.EcdsaSha384Signature:
+                    case SecurityAlgorithms.EcdsaSha512Signature:
+                        return true;
+
+                    default:
+                        return false;
+                }
+            }
+            else if (Kty == JsonWebAlgorithmsKeyTypes.Octet)
+            {
+                switch (algorithm)
+                {
+                    case SecurityAlgorithms.HmacSha256Signature:
+                    case SecurityAlgorithms.HmacSha384Signature:
+                    case SecurityAlgorithms.HmacSha512Signature:
+                    case SecurityAlgorithms.HmacSha256:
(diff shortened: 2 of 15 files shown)


test/Microsoft.IdentityModel.Tokens.Tests/ECDsaSecurityKeyTests.cs
```diff
@@ -61,5 +61,14 @@ public void Defaults()
         {
             // there are no defaults.
         }
+
+        [Fact(DisplayName = "EcdsaSecurityKeyTests: IsSupportedAlgorithm")]
```
brentschmaltz: I stopped setting DisplayName as it makes it harder to find the fully qualified method when debugging.
We should use either TheoryData or Inline so that all test cases will get executed. This test could fail on first one, then the others will not get run.

test/Microsoft.IdentityModel.Tokens.Tests/JsonWebKeyTests.cs
```diff
@@ -171,5 +171,16 @@ private bool IsDefaultJsonWebKey(JsonWebKey jsonWebKey)
 
             return true;
         }
+
+        [Fact(DisplayName = "RsaSecurityKeyTests: IsSupportedAlgorithm")]
+        public void IsSupportedAlgorithm()
```
brentschmaltz: Same comments about DisplayName and TheoryData.

brentschmaltz:
:shipit:

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
```diff
@@ -88,5 +89,25 @@ public override SignatureProvider GetSignatureProvider(string algorithm, bool ve
             else
                 return CryptoProviderFactory.CreateForSigning(this, algorithm);
         }
+
+        public override bool IsSupportedAlgorithm(string algorithm)
+        {
+            if (string.IsNullOrEmpty(algorithm))
```
PinpointTownes: Is there any reason to tolerate null/empty strings? Throwing would probably be more appropriate.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -309,5 +309,63 @@ public override SignatureProvider GetSignatureProvider(string algorithm, bool ve
...
+                {
+                    case SecurityAlgorithms.HmacSha256Signature:
+                    case SecurityAlgorithms.HmacSha384Signature:
+                    case SecurityAlgorithms.HmacSha512Signature:
+                    case SecurityAlgorithms.HmacSha256:
+                    case SecurityAlgorithms.HmacSha384:
+                    case SecurityAlgorithms.HmacSha512:
+                        return true;
+
+                    default:
```
PinpointTownes: No story for encryption/key wrap algorithms?

src/Microsoft.IdentityModel.Tokens/SymmetricSecurityKey.cs
```diff
@@ -65,6 +65,26 @@ public override SignatureProvider GetSignatureProvider(string algorithm, bool ve
...
+            {
+                case SecurityAlgorithms.HmacSha256Signature:
+                case SecurityAlgorithms.HmacSha384Signature:
+                case SecurityAlgorithms.HmacSha512Signature:
+                case SecurityAlgorithms.HmacSha256:
+                case SecurityAlgorithms.HmacSha384:
+                case SecurityAlgorithms.HmacSha512:
+                    return true;
+
+                default:
```
PinpointTownes: Same remark: encryption/key wrap algos need more love.

src/Microsoft.IdentityModel.Tokens/X509SecurityKey.cs
```diff
@@ -127,6 +127,26 @@ public override SignatureProvider GetSignatureProvider(string algorithm, bool ve
...
+
+            switch (algorithm)
+            {
+                case SecurityAlgorithms.RsaSha256:
+                case SecurityAlgorithms.RsaSha384:
+                case SecurityAlgorithms.RsaSha512:
+                case SecurityAlgorithms.RsaSha256Signature:
+                case SecurityAlgorithms.RsaSha384Signature:
+                case SecurityAlgorithms.RsaSha512Signature:
+                    return _certificate.PublicKey != null;
```
PinpointTownes: Heh, this check doesn't seem rock-solid. It will return a wrong value for DSA/ECDsa certificates.

src/Microsoft.IdentityModel.Tokens/X509SecurityKey.cs
```diff
@@ -127,6 +127,26 @@ public override SignatureProvider GetSignatureProvider(string algorithm, bool ve
...
+
+                default:
```
PinpointTownes: Why not adding the ECDsa algorithms?

tushargupta51: X509SecurityKey.PublicKey/PrivateKey does not support ecdsa keys yet (it always reads the certificate key as Rsa). Whats the benefit of returning true for ecdsa here? It will fail later on when creating the signature provider.

PinpointTownes: Ah fine then. I had not realized ECDsa certificates were not supported by Wilson ATM.

It doesn't change anything to the fact it will return `true` on .NET Desktop if you try to use an ECDsa certificate and specify an RSA algorithm, tho.

PinpointTownes:
Hem, we're seeing weird errors with the RC2 bits. Was this PR included in the upcoming release?

tushargupta51:
@PinpointTownes Yes, this PR is part of Rc2 bits. Version: 305061149

PinpointTownes:
Marvelous. Now, we're blocked by an incomplete IsSupportedAlgorithm. I assume there's no point to address the missing algorithms before RC2?

tushargupta51:
@PinpointTownes How is this blocking you? The set of supported algorithms is same as before (what was in SignatureProviders). The encryption/key wrap algorithms are not supported by default as signature providers do not know how to create a crypto provider for them.

PinpointTownes:
It's blocking us because we rely on IsSupportedAlgorithm to determine if a particular key is compatible with key wraps/symmetric algorithms. With the introduction of this method in RC2, our static helper (that handles correctly all the supported algorithms) is now ignored, which will cause much pain until we rename it. 

I'm not sure why this method should be limited to what the signature provider supports. These are two different classes, and a key can be used for something else than signing.

brentschmaltz:
@PinpointTownes 
Good points as it is going to be difficult to extend algorithm support without deriving.
Extensiblility for the algorithms a key can exist as a static default list on the key OR we could use a delegate to answer the question.

Also, if the AsymmetricAlgorithmResolver is set, we shouldn't check here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs#L126

PinpointTownes:
@brentschmaltz so what's the plan? Are you going to add the precious missing algorithms in RC2 or is this something we need to work around ourselves?

brentschmaltz:
@PinpointTownes my current thought is to role back the change, make the fix that we check for AsymmetricAlgorithmResolver before faulting and rethink it. We are always going to have the situation where users need to extend algorithm support. The model we put in requires deriving both Provider and Key. I am thinking all the key votes on is does the algorithm match my class: Symmetric or Asymmetric. Even then, with all the specs it will be hard to have a comprehensive solution over time. A natural extensibility model is needed here.

Issue #374, opened 2016-04-26
Clean up Exception Message for 'ArgumentException' related Exceptions


brentschmaltz:
For example ArgumentOutOfRangeException reads:

'System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
Parameter name: IDX10603: The algorithm: 'HS256' cannot have less than: '128' bits. KeySize: '56'.

I would like it to read (something like):
'System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
Parameter name: key.KeySize
The algorithm: 'HS256' key cannot be used with a key.KeySize less than: '128' bits. key.KeySize is: '56'.

Our ArgumentNullException has similar issues, basically we need to special case the 'ArgumentException' derived types.

PR #405, opened 2016-05-20
Adding ArgumentOutOfRangeException messages that include the argument…

polita:
… name

@tushargupta51 @brentschmaltz @ChenYan98 
This addresses #374 to clean up ArgumentOutOfRangeException messages for users. I didn't pull out the error numbers because I assume those are used to diagnose and debug when users report issues. ArgumentNullException messages didn't change because they already had clear messages.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
@@ -134,7 +134,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
             {
                 if (value <= TimeSpan.Zero)
                 {
-                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10105, value);
+                    throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("value", LogMessages.IDX10105, value);
                 }
 
                 _nonceLifetime = value;
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -124,7 +124,7 @@ namespace Microsoft.IdentityModel.Protocols
             set
             {
                 if (value < MinimumAutomaticRefreshInterval)
-                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10107, MinimumAutomaticRefreshInterval, value);
+                    throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("value", LogMessages.IDX10107, MinimumAutomaticRefreshInterval, value);
 
                 _automaticRefreshInterval = value;
             }
@@ -139,7 +139,7 @@ namespace Microsoft.IdentityModel.Protocols
             set
             {
                 if (value < MinimumRefreshInterval)
-                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10106, MinimumRefreshInterval, value);
+                    throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("value", LogMessages.IDX10106, MinimumRefreshInterval, value);
 
                 _refreshInterval = value;
             }
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -141,7 +141,7 @@ namespace Microsoft.IdentityModel.Tokens
                 {
                     _ecdsa = asymmetricAlgorithm as ECDsa;
                     if (_ecdsa == null)
-                        throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("key", LogMessages.IDX10641, key);
+                        throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>(nameof(key), LogMessages.IDX10641, key);
                 }
 #else
                 _rsaCryptoServiceProvider = asymmetricAlgorithm as RSACryptoServiceProvider;
@@ -149,7 +149,7 @@ namespace Microsoft.IdentityModel.Tokens
                 {
                     _ecdsa = asymmetricAlgorithm as ECDsaCng;
                     if (_ecdsa == null)
-                        throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("key", LogMessages.IDX10641, key);
+                        throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>(nameof(key), LogMessages.IDX10641, key);
                 }
 #endif
             }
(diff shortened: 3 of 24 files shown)


src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -141,15 +141,15 @@ public AsymmetricSignatureProvider(SecurityKey key, string algorithm, bool willC
                 {
                     _ecdsa = asymmetricAlgorithm as ECDsa;
                     if (_ecdsa == null)
-                        throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10641, key);
+                        throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("key", LogMessages.IDX10641, key);
```
brentschmaltz: think about using nameof(key) to catch when we change parameter names.

polita:
@brentschmaltz The new error message for the example in the original #374 issue description now reads:
Exception displayed to user: 
System.ArgumentOutOfRangeException: IDX10603: The algorithm: 'http://www.w3.org/2001/04/xmldsig-more#hmac-sha256' cannot have less than: '128' bits. KeySize is: '56'.
Parameter name: key.KeySize

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -134,7 +134,7 @@ public TimeSpan NonceLifetime
             {
                 if (value <= TimeSpan.Zero)
                 {
-                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10105, value);
+                    throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("value", LogMessages.IDX10105, value);
```
tushargupta51: I think argument name should be the name of property: NonceLifetime instead of "value". I think the message will read as:

System.ArgumentOutOfRangeException: IDX10603: IDX10105: NonceLifetime must be greater than zero. value: '<some value'"
Parameter name: NonceLifetime

We include the argument name in the message itself. So, the argument name in such cases will be redundant.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -124,7 +124,7 @@ public TimeSpan AutomaticRefreshInterval
             set
             {
                 if (value < MinimumAutomaticRefreshInterval)
-                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10107, MinimumAutomaticRefreshInterval, value);
+                    throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("value", LogMessages.IDX10107, MinimumAutomaticRefreshInterval, value);
```
tushargupta51: Same as above: "value" -> AutomaticRefreshInterval

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -139,7 +139,7 @@ public TimeSpan RefreshInterval
             set
             {
                 if (value < MinimumRefreshInterval)
-                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10106, MinimumRefreshInterval, value);
+                    throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("value", LogMessages.IDX10106, MinimumRefreshInterval, value);
```
tushargupta51: "value" -> RefreshInterval

src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
```diff
@@ -112,7 +112,7 @@ public int MinimumSymmetricKeySizeInBits
             set
             {
                 if (value < DefaultMinimumSymmetricKeySizeInBits)
-                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10628, DefaultMinimumSymmetricKeySizeInBits);
+                    throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("value", LogMessages.IDX10628, DefaultMinimumSymmetricKeySizeInBits);
```
tushargupta51: "value" => MinimumSymmetricKeySizeInBits

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -240,7 +240,7 @@ public TimeSpan ClockSkew
             set
             {
                 if (value < TimeSpan.Zero)
-                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10100, value);
+                    throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("value", LogMessages.IDX10100, value);
```
tushargupta51: "value" => ClockSkew

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -391,7 +391,7 @@ public string NameClaimType
             set
             {
                 if (string.IsNullOrWhiteSpace(value))
-                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10102);
+                    throw LogHelper.LogArgumentException<ArgumentOutOfRangeException>("value", LogMessages.IDX10102);
```
tushargupta51: "value" => NameClaimType

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
polita: In this case, because it's a property, you'll get "Parameter name: value" along with the specialized error message. This is generally consistent with the parameter name for .NET properties.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
polita: The special message and the stack will help people find the property name.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
polita: The special message and the stack will help people find the property name.

src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
polita: The special message and the stack will help people find the property name.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
polita: The special message and the stack will help people find the property name.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
polita: Same as above.

brentschmaltz:
:shipit:

Issue #401, opened 2016-05-19
Default Automatic MetadataRefresh should be 1 day instead of 5 days.


brentschmaltz:
See: 
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs#L48

PR #408, opened 2016-05-23
Update default automatic MetadataRefresh to 1 day.

polita:
For #401. @brentschmaltz , please take a look.


brentschmaltz:
:shipit:

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -45,7 +45,7 @@ public class ConfigurationManager<T> : IConfigurationManager<T> where T : class
         /// <summary>
         /// 5 days is the default time interval that afterwards, <see cref="GetConfigurationAsync()"/> will obtain new configuration.
```
PinpointTownes: Nice, the XML documentation is now incorrect... maybe it was worth updating it? :sweat_smile:

polita: Thanks, I'll fix it.

polita: See PR #410

PR #410, opened 2016-05-24
Update comment for updated Default Automatic Refresh interval.

polita:
@PinpointTownes noticed that the comment wasn't updated for the new refresh interval; fixing that issue.


PinpointTownes:
:shipit:

Issue #373, opened 2016-04-26
OpenIdConnectRequestType, OpenIdConnectResponseTypes should be named consistently


brentschmaltz:
OpenIdConnectResponseTypes, OpenIdConnectResponseModes, OpenIdConnectScopes all end with an  s
OpenIdConnectRequestType does not.

PR #411, opened 2016-05-25
Change OpenIdConnectRequestType to OpenIdConnectRequestTypes

ChenYan98:
Issue: #373


PinpointTownes:
FWIW, this change doesn't comply with the enum naming guidance: https://msdn.microsoft.com/en-us/library/ms229040.aspx#Anchor_2

polita:
@PinpointTownes is correct. We should change the other enums to be singular (unless they are flags) rather than changing this to be plural.

ChenYan98:
Fixed in #417 PR.

PR #417, opened 2016-05-26
Update OpenIdConnectRequestType

ChenYan98:
#373

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectRequestType.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectRequestType.cs
@@ -30,13 +30,24 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
     /// <summary>
     /// RequestTypes for OpenIdConnect.
     /// </summary>
-    /// <remarks>Can be used to determine the message type.</remarks>
+    /// <remarks>Can be used to determine the message type by consumers of an <see cref="OpenIdConnectMessage"/>.
+    /// For example: <see cref="OpenIdConnectMessage.CreateAuthenticationRequestUrl"/> sets <see cref="OpenIdConnectMessage.RequestType"/>
+    /// to <see cref="OpenIdConnectRequestType.Authentication"/>.</remarks>
     public enum OpenIdConnectRequestType
     {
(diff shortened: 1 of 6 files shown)


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectRequestType.cs
```diff
@@ -34,9 +34,9 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
     public enum OpenIdConnectRequestType
     {
 #pragma warning disable 1591
```
polita: Can we pull out this disabling of the compiler warning and correctly xml-doc this member?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectResponseMode.cs
```diff
@@ -30,7 +30,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
     /// <summary>
     /// Response modes for OpenIdConnect.
     /// </summary>
-    public static class OpenIdConnectResponseModes
+    public static class OpenIdConnectResponseMode
     {
 #pragma warning disable 1591
```
polita: Same as above.

brentschmaltz: Will need to add comments as it is public.

ChenYan98: Added

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectRequestType.cs
ChenYan98: Added comments as it is public.

polita: Can we remove the pragmas then?

brentschmaltz:
:shipit: You will need to rebase before merging.

polita:
+1

Issue #404, opened 2016-05-20
ValidateTokenReplay happens too early in JwtSecurityTokenHandler.ValidateToken


bgever:
Refer to the following line in the `JwtSecurityTokenHandler.ValidateToken` method:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L562

The call to `ValidateTokenReplay` happens presumably at this line because the `expires` variable is available. However, there are more validations after this call.

*These validations, including the issuer, should happen before the replay is checked.*

The reason is the `jti` claim, which says per the [RFC7519 spec](https://tools.ietf.org/html/rfc7519#section-4.1.7) (emphasis mine):

> 4.1.7.  "jti" (JWT ID) Claim
> 
>    The "jti" (JWT ID) claim provides a unique identifier for the JWT.
>    The identifier value MUST be assigned in a manner that ensures that
>    there is a negligible probability that the same value will be
>    accidentally assigned to a different data object; **if the application
>    uses multiple issuers, collisions MUST be prevented among values
>    produced by different issuers as well.**  The "jti" claim can be used
>    to prevent the JWT from being replayed.  The "jti" value is a case-
>    sensitive string.  Use of this claim is OPTIONAL.

So when the cache delegate as part of the ValidateTokenReplay call would check the JWT, it would use the `jti` claim to look up duplicates. However, this should be from the **context of the issuer**, which currently isn't verified at that stage.

Therefore, I would recommend to change the location of the ValidateTokenReplay to move it down until all validations have completed, after [line 613](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L613).

brentschmaltz:
@bgever this makes sense. Thanks for the input. Marked for RTM.

bgever:
Great, thanks for adding to the RTM milestone, @brentschmaltz.

PR #418, opened 2016-05-27
Adding extensibility for IsSupportedAlgorithm method

tushargupta51:
@brentschmaltz @polita @pinpointtownes

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
@@ -420,6 +420,14 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         /// <returns><see cref="HashAlgorithm"/> corresponding to the input string 'algorithm'</returns>
         public virtual HashAlgorithm GetHashAlgorithm(string algorithm)
         {
+
+            if (CryptoProviderFactory.CustomCryptoProvider != null)
+            {
+                var hashalg = CryptoProviderFactory.CustomCryptoProvider.ResolveHashAlgorithmFromSignatureAlgorithm(algorithm);
+                if (hashalg != null)
+                    return hashalg;
+            }
+
             if (algorithm == null)
             {
                 algorithm = SecurityAlgorithms.RsaSha256;
@@ -429,7 +437,6 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
             {
                 switch (algorithm)
                 {
-                    case SecurityAlgorithms.Sha256:
                     case SecurityAlgorithms.EcdsaSha256:
                     case SecurityAlgorithms.HmacSha256:
                     case SecurityAlgorithms.RsaSha256:
@@ -438,7 +445,6 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
                     case SecurityAlgorithms.EcdsaSha256Signature:
                         return SHA256.Create();
 
-                    case SecurityAlgorithms.Sha384:
                     case SecurityAlgorithms.EcdsaSha384:
                     case SecurityAlgorithms.HmacSha384:
                     case SecurityAlgorithms.RsaSha384:
@@ -447,16 +453,14 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
                     case SecurityAlgorithms.EcdsaSha384Signature:
                         return SHA384.Create();
 
-                    case SecurityAlgorithms.Sha512:
-                    case SecurityAlgorithms.RsaSha512:
                     case SecurityAlgorithms.EcdsaSha512:
                     case SecurityAlgorithms.HmacSha512:
+                    case SecurityAlgorithms.RsaSha512:
                     case SecurityAlgorithms.RsaSsaPssSha512:
                     case SecurityAlgorithms.RsaSha512Signature:
                     case SecurityAlgorithms.EcdsaSha512Signature:
                         return SHA512.Create();
                 }
-
             }
             catch (Exception ex)
             {
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -29,7 +29,6 @@ using System;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography;
-using System.Security.Cryptography.X509Certificates;
 using Microsoft.IdentityModel.Logging;
 
 namespace Microsoft.IdentityModel.Tokens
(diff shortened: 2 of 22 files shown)


PinpointTownes:
Curious: any particular to not reuse `AsymmetricAlgorithmResolver`? (we'd also need a `SymmetricAlgorithmResolver` btw)

Isn't your `IsSupportedAlgorithm` semantically similar to `var isSupported = AsymmetricAlgorithmResolver(algorithm) != null`?

tushargupta51:
@PinpointTownes Both delegates provide different knobs. `IsSupportedAlgorithm` allows users to control algorithms names that runtime understands. For example, they can add a new alg name that creates an `RSA` instance. They don't have to worry about returning an `RSA` instance, just hook up the `IsSupportedAlgorithm` and return true for that string.

If users hook up `AsymmetricAlgorithmResolver` to support `AsymmetricAlgoritm` that runtime does not create by default, they don't need to override the `IsSupportedAlgorithm`.

ChenYan98:
Issue: #404

PinpointTownes:
@tushargupta51 sure, but the two are closely linked: if `IsSupportedAlgorithm` returns `true`, then `GetAsymmetricAlgorithm` must return a non-null value.

If you use a different extensibility hook for `IsSupportedAlgorithm` and `GetAsymmetricAlgorithm`, then there's a risk that a user will set up a `IsSupportedAlgorithm` delegate returning `true` but won't provide an `AsymmetricAlgorithmResolver` delegate. In this case, `GetAsymmetricAlgorithm` might return a null value, even if `IsSupportedAlgorithm` returns `true`.

My suggestion was to adopt something similar to IM 4.5, that relies on `CryptoHelper.GetAlgorithmFromConfig`:

```csharp
public override bool IsSupportedAlgorithm(string algorithm) {
    if (string.IsNullOrEmpty(algorithm)) {
        throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgument(algorithm, SR.GetString(SR.EmptyOrNullArgumentString, "algorithm"));
    }

    object algorithmObject = null;

    try {
        algorithmObject = CryptoHelper.GetAlgorithmFromConfig(algorithm);
    }
    catch (InvalidOperationException) {
        algorithmObject = null;
    }

    if (algorithmObject != null) {
        SignatureDescription signatureDescription = algorithmObject as SignatureDescription;
        if (signatureDescription != null)
            return true;
        AsymmetricAlgorithm asymmetricAlgorithm = algorithmObject as AsymmetricAlgorithm;
        if (asymmetricAlgorithm != null)
            return true;
        return false;
    }

    switch (algorithm) {
    case SecurityAlgorithms.RsaV15KeyWrap:
    case SecurityAlgorithms.RsaOaepKeyWrap:
    case SecurityAlgorithms.RsaSha1Signature:
        return true;
    default:
        return false;
    }
}
```

tushargupta51:
@PinpointTownes I wanted to stay away from calling the `AsymmetricAlgorithmResolver` in `IsSupportedAlgorithm` method because of its possible time complexity to create actual algorithms. In case of CryptoHelper, since there is an underlying map between string alg and alg object, it might not be so bad.

But I realized that people will have to follow these steps if they want to add custom algorithms:
1. Create a CPF and hook up CPF.AsymmetricAlgorithmResolver.
2. Set SecurityKey.CPF to the one created above.
3. Set CPF.IsSupportedAlgorithm to return true for the new alg.

Although without doing 3, users can still call `key.GetSignatureProvider` and it will return the custom `AsymmetricAlgorithm` object, it will be confusing and error prone if they rely on results of `key.IsSupportedAlgorithm` before calling `key.GetSignatureProvider`. 

Keeping these in mind, I updated the code acc to your suggestion. `IsSupportedAlgorithm` will call the delegate and returns true if delegate returns a non-null object. If it is null, it fall back to default checks and compare input alg against default algorithms. This removes step 3 in the process above.

I also added a `SymmetricAlgorithmResolver` delegate.

PinpointTownes:
Fantastic, exactly what I had in mind :clap: 

One question though: is there any particular reason to limit `SymmetricAlgorithmResolver` to `KeyedHashAlgorithm` instances? Supporting MAC algorithms is great, but probably not enough: one won't be able to return an encryption algorithm (e.g AES/TripleDES). Why not just returning `SymmetricAlgorithm`?

tushargupta51:
@PinpointTownes I initially had `SymmetricAlgorithm` but later dropped it since that does not support returning hash algorithms like MD5 or SHA256. I had to choose between `HashAlgorithm` and `SymmetricAlgorithm` as the return type since they do not have a common parent. Also, I chose `KeyedHashAlgorithm` instead of `HashAlgorithm` because we are doing key based hashing.

PinpointTownes:
Shouldn't we just have a separate `HashAlgorithmResolver`?

tushargupta51:
@PinpointTownes I see that the name of the delegate is misleading. It should be `HashAlgorithmResolver` or 'KeyedHashAlgorithmResolver`. The intention here is to allow users to hook up their own keyed hash algorithm. Plugging in a `SymmetricAlgorithm` won't help with sign/verify operations as it does not support those.

PinpointTownes:
> I see that the name of the delegate is misleading. It should be HashAlgorithmResolver or 'KeyedHashAlgorithmResolver.

Yep. That said, you'll also need a `SymmetricAlgorithmResolver` returning a `SymmetricAlgorithm` at some point (when implementing JWE or when adding back SAML), so maybe it's not harmful to have the 3 hooks in place right now.

tushargupta51:
@PinpointTownes I agree we need those delegates when we have the encryption support in. Though, it makes more sense to add those hooks when we add the encryption feature.

brentschmaltz:
@PinpointTownes as you said
>My suggestion was to adopt something similar to IM 4.5, that relies on >CryptoHelper.GetAlgorithmFromConfig:

It would be ideal if extensibility was opaque to IdentityModel. CryptoConfig is absent from CoreFx so we need something internal to IM. CryptoFactory may be a nature place for such extensibility. Although it would need to be 'global' to the app.

thoughts?

tushargupta51:
@PinpointTownes The problem with the delegate resolvers is that it's not easy to add a support for just one algorithm. Since we skip the default algorithm resolution logic if a delegate is set by the user and if the delagate returns null, we throw.

We were evaluating if its better to have something similar to `CryptoConfig.AddAlgorithm(Type, String[])` instead of delagates. We will add static methods on `CryptoProviderFactory`:
1. `AddAsymmetricAlgorithm(String[], AsymmetricAlgorithm)`
2. `AddKeyedHashAlgorithm(String[], KeyedHashAlgorithm)`

And these methods will maintain a mapping internally. The benefit here is that users do not have to ensure that the default behavior (creating RSA, ECDSA) is also present in the delegate resolvers.

PinpointTownes:
Sure, a dictionary is a good option too (I'm pretty sure I had suggested it before the resolver delegate was introduced). That said, why making that completely static?

tushargupta51:
I don't see the use case where users will need to set the mapping from a string alg to an alg obj per CPF. Setting it once and using the same object type in future whenever the new "alg" string is found seems like the right thing to do.

brentschmaltz:
@tushargupta51 @PinpointTownes sounds like we agreement that this type of extensibility makes sense. Another reason for the static is OpenIdConnectProtocolValidator could reach out for custom algorithms across OS's. When validating the c_hash or at_hash we need to map the 'alg' in the header to a HashAlg. Currently, OIDCPV knows nothing about a CryptoProviderFactory instance. 
A good instance solution would start with the static defaults and them be configurable per instance something like the JwtSecurityTokenHandler claim mapping (that everyone loves :-)

PinpointTownes:
> I don't see the use case where users will need to set the mapping from a string alg to an alg obj per CPF. 

Well, the same reasoning could be applied to the signature provider. IMHO, it's probably better to be consistent here.

> Another reason for the static is OpenIdConnectProtocolValidator could reach out for custom algorithms across OS's.

Sure, but nothing prevents you from adding a `CryptoProviderFactory` property in this class. If none is explicitly provided, the default `CryptoProviderFactory.Default` static instance is used as a fallback. The resolvers/mappings properties don't have to be static for that.

tushargupta51:
Closing this PR as the changes here are subset of #444 .

PR #424, opened 2016-06-01
moved build files to root from .build

brentschmaltz:
added clean
removed obsolete xml from buildConfiguration.xml
consistency between src and test nodes
group statements
old and new versions for netfxappversion
old and new versions for coreFx files


build.ps1
```diff
@@ -1,5 +1,239 @@
...
+    Invoke-WebRequest $dotnetUrl -OutFile "$dotnetInstallDir\dotnet-install.ps1"
+
+    & "$dotnetInstallDir\dotnet-install.ps1" -Channel $dotnetChannel -Version $dotnetVersion -Architecture x64 -InstallDir $dotnetInstallDir -Verbose
+    if($LASTEXITCODE -ne 0)
+    {
+        throw "Failed to install dotnet"
+    }
+}
+
+if ($updateCoreFxVersion -eq "YES")
```
tushargupta51: We should not update the coreFx version in our local builds. This ends up showing bunch of updated files in "git status" and I don't think we gain anything with specific version for our local builds. This should live in Jenkins.

Also, how can you pass this flag when using build.cmd?

buildConfiguration.xml
```diff
@@ -0,0 +1,63 @@
...
+        <build>yes</build>
+        <dependency>Microsoft.IdentityModel.Tokens</dependency>
+      </project>
+      <project name="System.IdentityModel.Tokens.Saml">
+        <build>yes</build>
+        <dependency>Microsoft.IdentityModel.Tokens</dependency>
+      </project>
+      <project name="Microsoft.IdentityModel.Protocols">
+        <build>yes</build>
+        <dependency>System.IdentityModel.Tokens.Jwt</dependency>
```
tushargupta51: Why do we need this?

buildConfiguration.xml
```diff
@@ -0,0 +1,63 @@
...
+      </project>
+      <project name="Microsoft.IdentityModel.Tokens">
+        <build>yes</build>
+      </project>
+      <project name="System.IdentityModel.Tokens.Jwt">
+        <build>yes</build>
+        <dependency>Microsoft.IdentityModel.Tokens</dependency>
+      </project>
+      <project name="System.IdentityModel.Tokens.Saml">
+        <build>yes</build>
```
tushargupta51: We don't use the "build" property. We reply on the global "$build" parameter. We either build none or all.

build.ps1
brentschmaltz: We can change the mapping in a future commit.
User the flag -updateCoreFxVersion NO

buildConfiguration.xml
brentschmaltz: It separates test from src. Currently there is a 1:1 mapping, but this may not always be the case.

buildConfiguration.xml
brentschmaltz: True, i left it in with the promise that in the future we make use of it.

buildConfiguration.xml
tushargupta51: Not sure I understand. How does having the "dependency" node separate "src" and "test" projects?

build.ps1
tushargupta51: I don't see why we want to include it now. If you really want it, the default should be "No".

brentschmaltz: I agree, i'll set the default to "no"

buildConfiguration.xml
polita: Let's take it out until we actually use it. If you want to leave it in but commented out as a reminder, that's fine, but otherwise it's confusing.

buildConfiguration.xml
polita: Doesn't "dependency" tell the build that the dependency must be built before this assembly?

polita:
It looks like install.ps1 went away. Does build.ps1 now install the CLI as well?

buildConfiguration.xml
brentschmaltz: right you are, I'll remove it.

buildConfiguration.xml
brentschmaltz: @polita install.ps1 is pulled down dynamically during a build step.

tushargupta51:
@polita Yes, build.ps1 installs the "install.ps1" script as well, that is what .net folks recommended.

@brentschmaltz :shipit: after rebase

Issue #372, opened 2016-04-22
JwtSecurityTokenHandler should not set CryptoProviderFactory.Default.


brentschmaltz:
This will lead to following tokens using the instance provider.
See: JwtSecurityTokenHandler ~ line 518
CryptoProviderFactory.Default = validationParameters.CryptoProviderFactory;

PR #425, opened 2016-06-01
Fix setting CPF.Default, clean up exceptions

tushargupta51:
@brentschmaltz @polita 

#372

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -146,6 +146,9 @@ namespace Microsoft.IdentityModel.Tokens
             if (asymmetricAlgorithmResolver != null)
             {
                 AsymmetricAlgorithm asymmetricAlgorithm = asymmetricAlgorithmResolver(key, algorithm, willCreateSignatures);
+                if (asymmetricAlgorithm == null)
+                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10646, nameof(key), algorithm);
+
 #if NETSTANDARD1_4
                 _rsa = asymmetricAlgorithm as RSA;
                 if (_rsa == null)
@@ -670,9 +673,6 @@ namespace Microsoft.IdentityModel.Tokens
 
             if (_disposed)
                 throw LogHelper.LogException<ObjectDisposedException>(GetType().ToString());
-
-            if (_hashAlgorithm == null)
-                throw LogHelper.LogException<InvalidOperationException>(LogMessages.IDX10621, "signature");
 
 #if NETSTANDARD1_4
             if (_rsa != null)
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -147,7 +147,7 @@ namespace Microsoft.IdentityModel.Tokens
             {
                 AsymmetricAlgorithm asymmetricAlgorithm = asymmetricAlgorithmResolver(key, algorithm, willCreateSignatures);
                 if (asymmetricAlgorithm == null)
-                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10646, nameof(key), algorithm);
+                    throw LogHelper.LogException<InvalidOperationException>(LogMessages.IDX10646, nameof(key), algorithm);
 
 #if NETSTANDARD1_4
                 _rsa = asymmetricAlgorithm as RSA;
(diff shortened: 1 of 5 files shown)


src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -146,6 +146,9 @@ public AsymmetricSignatureProvider(SecurityKey key, string algorithm, bool willC
             if (asymmetricAlgorithmResolver != null)
             {
                 AsymmetricAlgorithm asymmetricAlgorithm = asymmetricAlgorithmResolver(key, algorithm, willCreateSignatures);
+                if (asymmetricAlgorithm == null)
+                    throw LogHelper.LogException<ArgumentOutOfRangeException>(LogMessages.IDX10646, nameof(key), algorithm);
```
brentschmaltz: I don't think we want to use nameof(key) here. Perhaps: key.ToString()

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
```diff
@@ -475,7 +475,7 @@ public void ValidateTokens()
                     ValidIssuer = IdentityUtilities.DefaultIssuer,
                 };
 
-            TestUtilities.ValidateTokenReplay(securityToken: jwt, tokenValidator: tokenHandler, validationParameters: validationParameters);
+            TestUtilities.ValidateTokenReplay(jwt, tokenHandler, validationParameters);
```
brentschmaltz: nice, thanks :-)

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
tushargupta51: @brentschmaltz  I am confused about the exception type that should be thrown here. `ArgumentOutOfRange` does not seem the right one. But I couldn't find any existing exception that can fit best.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -90,6 +89,7 @@ internal static class LogMessages
         internal const string IDX10643 = "IDX10643: Comparing the signature created over the input with the token signature: '{0}'.";
         internal const string IDX10644 = "IDX10644: Crypto operation not supported.";
         internal const string IDX10645 = "IDX10645: Elliptical Curve not supported: '{0}'";
+        internal const string IDX10646 = "IDX10646: The delagate for resolving asymmetric algorithm returned null. Key: {0}, Alg: {1}.";
```
brentschmaltz: misspelling: delagate -> delegate

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
brentschmaltz: ArgumentOutOfRangeException is fine. The error message text will help out devs.
On second thought, just pass 'key' instead of 'key.ToString()'. That is our standard.

src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
```diff
@@ -168,7 +168,7 @@ public override byte[] Sign(byte[] input)
                 throw LogHelper.LogException<ObjectDisposedException>(GetType().ToString());
 
             if (_keyedHash == null)
-                throw LogHelper.LogException<InvalidOperationException>(LogMessages.IDX10623);
+                throw LogHelper.LogException<ArgumentNullException>(LogMessages.IDX10623);
```
polita: Is _keyedHash a parameter? Should this really be an ArgumentNullException? _keyedHash looks like a member var, not an argument to the method.

src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
```diff
@@ -205,7 +205,7 @@ public override bool Verify(byte[] input, byte[] signature)
                 throw LogHelper.LogException<ObjectDisposedException>(typeof(SymmetricSignatureProvider).ToString());
 
             if (_keyedHash == null)
-                throw LogHelper.LogException<ArgumentException>(LogMessages.IDX10623);
+                throw LogHelper.LogException<ArgumentNullException>(LogMessages.IDX10623);
```
polita: This probably shouldn't be an ArgumentNullException because there isn't a null method argument. InvalidOperationException?

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
PinpointTownes: `InvalidOperationException`?

brentschmaltz: or InvalidOperationException. Either is fine. Point is the message will help the dev understand what the error is about.

Issue #136, opened 2015-04-21
NullReferenceException in JwtPayload when a claim has a null value. (A null value should be acceptable)


StephaneBischoff:
If a claim has a null value, then a NullReferenceException is raised in JwtPayload.

I believe that the line at fault is:
JwtPayload.cs -  line 375 ->  keyValuePair.Value.GetType()

brentschmaltz:
@StephaneBischoff can you add the stacktrace? The .Net Claim.Value cannot be 'null', so we are trying to figure out what to do in this case. JSON allows 'null' values, but .Net doesn't.

brentschmaltz:
We should at least fix the null ref and drop the claim. It will still be in the jwtpayload, but not in the ClaimsIdentity.

Beanow:
+1 to fix this bug, blocking issue for us currently.

@brentschmaltz I disagree that the values should be dropped though. It suggests that the claim was missing, but this is not the case. The 3rd party explicitly indicated that a certain claim has an empty value. If we're talking dirty fixes I would suggest for example using ClaimValueTypes.HexBinary or any of the other binary encodings with value string.Empty. Essentially this is a valid representation of null.

```cs
if(keyValuePair.Value == null){
    claims.Add(new Claim(keyValuePair.Key, string.Empty, ClaimValueTypes.HexBinary, issuer, issuer));
}
```

On the other hand, getting this fixed at all takes priority in my opinion. Just putting it out there.

tushargupta51:
@Beanow Thanks for your input. We are looking into this. Are you using 4.x or 5.x version?

Beanow:
4.x, the latest stable using nuget.
And for debugging I've looked at the master branch's head.

brentschmaltz:
@Beanow We are thinking about introducing a new ClaimValueType in this library to represent the null do you think that would work for you?

Beanow:
I think that all hinges on how closely it behaves like null when implementing controller logic. Does it equal to null? Will it serialize to null when you're making json strings? Will the new C# operators treat it as null (`claim ?? default`)?
Then, will it conflict with whatever reason there originally was to not have nulls in owin?

tushargupta51:
@Beanow You mentioned that using `string.Empty and ClaimValuetypes.HexBinary` is a valid representation of null. I was trying to find some evidence to back it up. Does it serialize to null when converting to json strings?

Beanow:
[...]

PR #428, opened 2016-06-02
adding support for claims with null values

tushargupta51:
#136 

@brentschmaltz @polita

### DIFF
--- a/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
@@ -314,9 +314,11 @@ namespace System.IdentityModel.Tokens.Jwt
                 // there is some code redundancy here that was not factored as this is a high use method. Each identity received from the host will pass through here.
                 foreach (KeyValuePair<string, object> keyValuePair in this)
                 {
-                    // do not add claims with null value
                     if (keyValuePair.Value == null)
+                    {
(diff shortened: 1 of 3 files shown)


src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -534,11 +536,19 @@ internal static object GetClaimValueUsingValueType(Claim claim)
                 return JArray.Parse(claim.Value);
             }
 
+            if (claim.ValueType == JsonClaimValueTypes.JsonNull)
```
polita: This whole block is looking rife for a switch statement, but we don't need to change it right now.

polita:
LGTM

test/System.IdentityModel.Tokens.Jwt.Tests/JwtPayloadTest.cs
```diff
@@ -138,9 +138,14 @@ public void FirstClassProperties()
         [Fact]
         public void TestClaimWithNullValue()
         {
-            JwtPayload jwtPayload = new JwtPayload();
-            jwtPayload.Add("testClaim", null);
-            List<Claim> claims = jwtPayload.Claims as List<Claim>;   // this should not throw
+            JwtPayload payload = new JwtPayload();
```
brentschmaltz: I would like to see test that round-trips to and from JSON.

brentschmaltz:
:shipit:

Issue #426, opened 2016-06-02
Double space in error message IDX10214


bgever:
The error message [IDX10214](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b2f7484d3bf6786e355922b113d27ca06d0cc2ba/src/common/ErrorMessages.cs#L59) contains a double space, inconsistent with error message IDX10205. Occurs in both the 4.x and 5.x branches.

Noticed this when creating unit tests checking both these messages.

Will submit a PR shortly.

brentschmaltz:
@bgever thanks, will wait for your PR

bgever:
Note, the PR is for 4.x (dev branch), the same issue is also in 5.x, do I need to make another PR to the KDev branch?

tushargupta51:
@bgever Yes, please.

bgever:
Submitted PR #437 for 5.x.

tushargupta51:
Thanks @bgever .

bgever:
Thanks for accepting the PRs. :smile:

PR #429, opened 2016-06-03
Remove double space in error message IDX10214

bgever:
To be consistent with error message IDX10205, as described in issue #426.


brentschmaltz:
@bgever did you mean to fix this in KDev?
That is our release for 5.x (support for asp.net)

Issue #270, opened 2015-10-08
Add assembly versions to logs.


brentschmaltz:
We need a way to log the version of the assemblies that are running. Some runtimes do that with each message. Since our logger is separate from the assemblies we need to capture the dependant assemblies versions. We could do it with the first message or in each message.

brentschmaltz:
I have completed some work addressing #205 so taking ownership.

polita:
Can we log the assembly version in JWTSecurityTokenHandler and ConfigurationManager with a static ctor.

PR #430, opened 2016-06-03
Add assembly version to logs

polita:
@tushargupta51 @brentschmaltz 
Addressed #270 by adding static ctors to ConfigurationManager and JwtSecurityTokenHandler. Verified that it emits to logs in Verbose mode.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -74,6 +74,14 @@ namespace Microsoft.IdentityModel.Protocols
         private T _currentConfiguration;
 
         /// <summary>
+        /// Static initializer for a new object. Static initializers run before the first instance of the type is created.
+        /// </summary>
+        static ConfigurationManager()
+        {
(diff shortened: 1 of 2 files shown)


tushargupta51:
@polita It might be good to have a test for this. You can add something along the lines of this:

        [Fact]
        public void TestAssemblyVersionLog()
        {
            SampleListener listener = new SampleListener();
            // since null parameters exceptions are logged at Verbose level
            IdentityModelEventSource.Logger.LogLevel = EventLevel.Verbose;
            listener.EnableEvents(IdentityModelEventSource.Logger, EventLevel.Verbose);
            // call configuration manager/jwtsecuritytoken handler ctor here
            Assert.Contains(<assemblyVersion>, listener.TraceBuffer);
        }

tushargupta51:
:shipit: as we can't have a reliable test for this since we can not guarantee when the static ctor will run.

polita:
The static ctors run only on first creation of the class. Since the classes can get created by other tests, we won't be able to deterministically build a test that runs on first creation. Reordering or parallelizing test runs will break the tests.
Also, we would need to keep updating the test for updated versions of the assembly if we want to check against the literal string, which seems like a big tax.

PR #432, opened 2016-06-06
Use HashAlgorithmMap in OpenIdConnectProtocolValidator

ChenYan98:


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -702,5 +710,61 @@ protected virtual void ValidateState(OpenIdConnectProtocolValidationContext vali
...
+        private void InitializeHashAlgorithmMap()
+        {
+            ISet<string> securityAlgorithms256 = new HashSet<string>();
+            securityAlgorithms256.Add(SecurityAlgorithms.EcdsaSha256);
+            securityAlgorithms256.Add(SecurityAlgorithms.HmacSha256);
+            securityAlgorithms256.Add(SecurityAlgorithms.RsaSha256);
+            securityAlgorithms256.Add(SecurityAlgorithms.RsaSsaPssSha256);
+            securityAlgorithms256.Add(SecurityAlgorithms.RsaSha256Signature);
+            securityAlgorithms256.Add(SecurityAlgorithms.EcdsaSha256Signature);
+            _hashAlgorithmMap.Add(securityAlgorithms256, SHA256.Create());
```
tushargupta51: Where have we initialized `_hashAlgorithm`?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -702,5 +710,61 @@ protected virtual void ValidateState(OpenIdConnectProtocolValidationContext vali
...
+            securityAlgorithms512.Add(SecurityAlgorithms.EcdsaSha512Signature);
+            _hashAlgorithmMap.Add(securityAlgorithms512, SHA512.Create());
+        }
+
+        /// <summary>
+        /// This method is used in derived class which customizes the specific HashAlgorithmMap.
+        /// </summary>
+        /// <param name="hashAlgorithmMap">Customized HashAlgorithmMap.</param>
+        /// <param name="isAppend">Appends the customized HashAlgorithmMap to the default HashAlgorithmMap or not.</param>
+        protected void CustomizeHashAlgorithmMap(IDictionary<ISet<string>, HashAlgorithm> hashAlgorithmMap, bool isAppend = false)
```
tushargupta51: The intention here is to allow users to modify the default hashAlgorithm map. Why do we need a method for that? Users can directly modify the map as `HashAlgorithm` property is public. Not sure I understand why this is meant for derived classes.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -48,24 +48,14 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
...
-                { SecurityAlgorithms.RsaSha256, "SHA256" },
-                { SecurityAlgorithms.HmacSha256, "SHA256" },
-                { SecurityAlgorithms.EcdsaSha384, "SHA384" },
-                { SecurityAlgorithms.RsaSha384, "SHA384" },
-                { SecurityAlgorithms.HmacSha384, "SHA384" },
-                { SecurityAlgorithms.EcdsaSha512, "SHA512" },
-                { SecurityAlgorithms.RsaSha512, "SHA512" },
-                { SecurityAlgorithms.HmacSha512, "SHA512" },
-          };
+        private IDictionary<ISet<string>, HashAlgorithm> _hashAlgorithmMap;
```
tushargupta51: We moved away from the pattern where we had local privates for public properties. Should we just have the `HashAlgorithm` property?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -334,6 +326,38 @@ public virtual void ValidateUserInfoResponse(OpenIdConnectProtocolValidationCont
...
+        /// <param name="disposing">true, if called from Dispose(), false, if invoked inside a finalizer.</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!this._disposed)
+            {
+                if (disposing)
+                {
+                    foreach (var item in _hashAlgorithmMap)
+                    {
+                        if (item.Value != null)
```
tushargupta51: Have a look at this: https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md. We decided to follow these guidelines long time back. In this case, no { } if there is only one line inside if.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -334,6 +326,38 @@ public virtual void ValidateUserInfoResponse(OpenIdConnectProtocolValidationCont
...
+        /// Calls <see cref="Dispose(bool)"/> and <see cref="GC.SuppressFinalize"/>
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Calls <see cref="HashAlgorithm.Dispose()"/> to release this managed resources.
```
tushargupta51: "to release *the* managed resources". I noticed we have this typo in other places too.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
ChenYan98: A derived class will use this method when it wants to customize its own hashAlgorithm map. That is why this method has been set as "protected".

brentschmaltz:
We are going to take a different path.

PR #433, opened 2016-06-06
updated newtonsoft to 9.0.1-beta1

brentschmaltz:
aspnet wants us to upgrade to this version of newtonsoft


tushargupta51:
:shipit:

Issue #395, opened 2016-05-11
Reintroduce GetAsymmetricAlgorithm/GetSymmetricAlgorithm


PinpointTownes:
The base `AsymmetricSecurityKey` and `SymmetricSecurityKey` classes should expose `GetAsymmetricAlgorithm`/`GetSymmetricAlgorithm` to allow retrieving the underlying algorithm without having to cast the security key to known types.

@brentschmaltz any chance we could bring back these APIs for RTM?

brentschmaltz:
@PinpointTownes my current thought is to role back the change we did in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/386, make the fix that we check for AsymmetricAlgorithmResolver before faulting and rethink it. We are always going to have the situation where users need to extend algorithm support. The model we put in requires deriving both Provider and Key. I am thinking all the key votes on is does the algorithm match my class: Symmetric or Asymmetric. Even then, with all the specs it will be hard to have a comprehensive solution over time. A natural extensibility model is needed here.

This kind of fits with this bug.
What are your thoughts on rolling back 386.

PinpointTownes:
TBH, without seeing a prototype, it's hard to say :smile: 

That said, assuming `AsymmetricAlgorithmResolver` would work as .NET Desktop's `CryptoConfig`, then yeah, this sounds like the right approach to enable crypto agility on the long term.

Concerning `SecurityKey.IsSupportedAlgorithm`, no real preference. I believe the signature should stay as-is, so I'm not sure it's worth reverting it ATM. From my POV, it just needs to support the missing algorithms. Introducing a `CryptoConfig`-like for crypto agility would be a great bonus for RTM.

tushargupta51:
@PinpointTownes Unfortunately, we cant make any changes to RC2, the bar is too high. We will get it fixed in nightlies though.

For `IsSupportedAlgorithm`, I am thinking of adding a delegate: `CryptoProviderFactory.IsSuuportedAlgorithmDelegate` (maybe a different name). Then `SecurityKey.IsSupportedAlgorithm` will look like this:

         public override bool IsSupportedAlgorithm(string algorithm)
        {
            if (string.IsNullOrEmpty(algorithm))
                return false;

            if (CryptoProviderFactory.IsSupportedAlgorithmDelegate)
                return CryptoProviderFactory.IsSupportedAlgorithmDelegate(algorithm);

[...]

PR #444, opened 2016-06-10
Replaced delgates with interface on CryptoProviderFactory

brentschmaltz:
Added CryptoProviderFactory property to OIDCProtocolValidator
Removed Algorithm resolvers from signature providers
Users will set ICryptoProvider on CryptoProviderFactory for extensibility.
Fixed Disposing incorrectly with ECDsa keys


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -414,59 +426,50 @@ protected virtual void ValidateIdToken(OpenIdConnectProtocolValidationContext va
...
+        public CryptoProviderFactory CryptoProviderFactory
+        {
+            get
+            {
+                return _cryptoProviderFactory;
+            }
+            set
+            {
+                if (value == null)
+                    throw LogHelper.LogArgumentNullException(nameof(value));
```
polita: "value"

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
```diff
@@ -103,22 +103,7 @@ public override int KeySize
         /// <returns>true if this supports the algorithm; otherwise, false.</returns>
         public override bool IsSupportedAlgorithm(string algorithm)
```
tushargupta51: Should this be virtual? The method is more like `GetSignaureProvider` now, all it does is call into CPF. It will provide multiple ways of overriding the same thing.

src/Microsoft.IdentityModel.Tokens/SignatureProvider.cs
```diff
@@ -41,7 +41,7 @@ public abstract class SignatureProvider : IDisposable
         /// <param name="key">The <see cref="SecurityKey"/> that will be used for signature operations.</param>
         /// <param name="algorithm">The signature algorithm to apply.</param>
         /// <exception cref="ArgumentNullException">'key' is null.</exception>
-        protected SignatureProvider(SecurityKey key, string algorithm)
+        public SignatureProvider(SecurityKey key, string algorithm)
```
tushargupta51: This should not be public. We can actually get rid of this ctor. All it does it null key check and couple of assignments. We, anyways, do the check in derived classes, so this is not really useful.

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
brentschmaltz: I agree, we will remove this method and have users call: SecurityKey.CryptoProviderFactory.IsAlgorithmSupported(algorithm, this);

src/Microsoft.IdentityModel.Tokens/SignatureProvider.cs
brentschmaltz: I will move it back to protected.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
brentschmaltz: sure.

PinpointTownes: @polita curious: what's wrong with `nameof`?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -476,14 +479,25 @@ public virtual HashAlgorithm GetHashAlgorithm(string algorithm)
...
-                if (!string.Equals(expectedValue, hashString, StringComparison.Ordinal))
-                {
-                    throw LogHelper.LogException<OpenIdConnectProtocolException>(LogMessages.IDX10300, expectedValue, hashItem, algorithm);
-                }
+                hashAlgorithm = GetHashAlgorithm(algorithm);
+                CheckHash(hashAlgorithm, expectedValue, hashItem, algorithm);
+            }
+            finally
+            {
+                CryptoProviderFactory.ReleaseHashAlgorithm(hashAlgorithm);
```
PinpointTownes: Will `ReleaseHashAlgorithm` handle null values?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -51,15 +42,137 @@ public class CryptoProviderFactory
...
+        public virtual bool IsSupportedAlgorithm(string algorithm, SecurityKey key)
+        {
+            if (CustomCryptoProvider != null && CustomCryptoProvider.IsSupportedAlgorithm(algorithm, key))
+                return true;
+
+            if (key as RsaSecurityKey != null)
+                return IsRsaAlgorithmSupported(algorithm);
+
+            if (key as X509SecurityKey != null)
+                return IsRsaAlgorithmSupported(algorithm);
```
PinpointTownes: I know DSA/ECDsa certificates are not supported ATM by IdentityModel, but this method will return `true` for any RSA-based algorithm even if the certificate is not a RSA certificate. Please fix that.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -113,9 +290,21 @@ private SignatureProvider CreateProvider(SecurityKey key, string algorithm, bool
...
+            {
+                SignatureProvider signatureProvider = CustomCryptoProvider.Create(algorithm, key, willCreateSignatures) as SignatureProvider;
+                if (signatureProvider == null)
+                    throw LogHelper.LogException<InvalidOperationException>(LogMessages.IDX10646, key, algorithm, typeof(SignatureProvider));
+
+                return signatureProvider;
+            }
+
+            if (!IsSupportedAlgorithm(algorithm, key))
+                throw LogHelper.LogException<ArgumentException>(LogMessages.IDX10634, (algorithm ?? "null"), key);
```
PinpointTownes: nit: `algorithm` cannot be null at this point.

src/Microsoft.IdentityModel.Tokens/ICryptoProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+//------------------------------------------------------------------------------
+
+using System.Security.Cryptography;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Crypto operations
+    /// </summary>
+    public interface ICryptoProvider
```
PinpointTownes: IMHO, this "loosely-typed" contract seems really weird. Why is this even needed? Why isn't `CryptoProviderFactory` directly responsible of creating algorithm instances?

src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
```diff
@@ -48,13 +48,6 @@ public abstract class SecurityKey
         public string KeyId { get; set; }
 
         /// <summary>
-        /// This must be overridden to specify whether this SecurityKey supports the algorithm.
-        /// </summary>
-        /// <param name="algorithm">The crypto algorithm to use.</param>
-        /// <returns>true if this supports the algorithm; otherwise, false.</returns>
-        public abstract bool IsSupportedAlgorithm(string algorithm);
```
PinpointTownes: What's the reasoning behind removing this method?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -51,15 +42,137 @@ public class CryptoProviderFactory
         /// </summary>
         public static CryptoProviderFactory Default;
```
PinpointTownes: Why not making this a property?

PinpointTownes:
Where are `GetAsymmetricAlgorithm` and `GetSymmetricAlgorithm`? I can't believe I'll still have to cast  `SecurityKey` to known types to get the algorithm instance I need with IdentityModel RTM...

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
brentschmaltz: Good point, it should just return. I'll make sure.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: OK.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: thanks. that code was borrowed from above.
Will fix.

src/Microsoft.IdentityModel.Tokens/ICryptoProvider.cs
brentschmaltz: It gives one a way to plug in a new crypto provider OR to return a specific algorithm without deriving and overridding. Before this would have been handled by CryptoConfig. By default, we don't create FIPS compliant algorithms. If you need them, you can register a provider here that will return a specific version of a crypto operator.

PinpointTownes: But isn't that the whole point of `CryptoProviderFactory`: providing a hook to return a custom algorithm?

AFAICT, you basically moved `SecurityKey.IsSupportedAlgorithm()` to `CryptoProviderFactory.IsSupportedAlgorithm()` and `CryptoProviderFactory.IsSupportedAlgorithm()` to `ICryptoProvider.IsSupportedAlgorithm()`. What was wrong with the first separation?

PinpointTownes: The naming is also a bit weird: despite what its name might suggest, `CryptoProviderFactory` doesn't create instances of `ICryptoProvider` but delegates everything to `ICryptoProvider`, which is the real factory.

src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
brentschmaltz: @pinpointtownes
We thought it was redundant, you can call SecurityKey.CryptoProviderFactory.IsSupportedAlgorithm(algorithm, this).
This set of changes addresses a couple of issues. 
1. OpenIdConnectProtocolVerifier is disjoint from TokenValidation. You could plug in a custom algorithm but OIDCPV would still fail.
2. It was possible for a key to report YES but no CryptoSupport was actually available.
3. Plugging in a new OR overriding an existing required deriving.
4. Our delegates were scoped to AsymmetricAlgorithms and we would need one for each new type of crypto operations. In addition, the AsymmetricAlgorithm would need to be of a certain type or we would fail.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: We can do that.
I will make the change.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
polita: Nothing wrong with it, but since the name of the implicit "value" param will never change names, seems more performant to avoid the lookup and just provide the string "value". nameof provides consistency across parameter renames that we don't need on getters/setters.

src/Microsoft.IdentityModel.Tokens/ICryptoProvider.cs
brentschmaltz: ICryptoProvider allows for hooking a specific algorithm without deriving. The CryptoProvider is given first priority for obtaining a crypto provider. It allows overriding a single operation.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
PinpointTownes: More performant? Hum, not really ; `nameof` is a purely compile-time sugar. The IL code emitted by the compiler is exactly the same as when you use a magic string.

> The nameof expression is a constant. In all cases, nameof(...) is evaluated at compile-time to produce a string. Its argument is not evaluated at runtime, and is considered unreachable code (however it does not emit an "unreachable code" warning).

[nameof operator: spec v5](https://roslyn.codeplex.com/discussions/570551)

brentschmaltz:
@PinpointTownes we don't really serve up AsymmetricAlgorithms or SymmetricAlgorithms. We serve up SignatureProviders and HashAlgorithms. We choose to add this abstraction on top of these "algorithms" as it becomes confusing what is really supported across all the OS's and .Net versions.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
polita: Fair enough; thanks.

PinpointTownes:
> @PinpointTownes we don't really serve up AsymmetricAlgorithms or SymmetricAlgorithms. We serve up SignatureProviders and HashAlgorithms.

Grrrrr...

Here's an exercise for you: try to port this snippet to IM 5.

```csharp
if (!(credentials.SigningKey is AsymmetricSecurityKey) ||
     !credentials.SigningKey.IsSupportedAlgorithm(SecurityAlgorithms.RsaSha256Signature)) {
    Options.Logger.LogInformation("An unsupported signing key was ignored and excluded " +
                                  "from the key set: {Type}. Only asymmetric security keys " +
                                  "supporting RS256, RS384 or RS512 can be exposed " +
                                  "via the JWKS endpoint.", credentials.SigningKey.GetType().Name);

    continue;
}

// Resolve the underlying algorithm from the security key.
var algorithm = (RSA) ((AsymmetricSecurityKey) credentials.SigningKey)
    .GetAsymmetricAlgorithm(
        algorithm: SecurityAlgorithms.RsaSha256Signature,
        privateKey: false);
Debug.Assert(algorithm != null, "The instance returned by GetAsymmetricAlgorithm() shouldn't be null.");

// Export the RSA public key to create a new JSON Web Key
// exposing the exponent and the modulus parameters.
var parameters = algorithm.ExportParameters(includePrivateParameters: false);
Debug.Assert(parameters.Exponent != null, "A null exponent was returned by RSA.ExportParameters()");
Debug.Assert(parameters.Modulus != null, "A null modulus was returned by RSA.ExportParameters()");
```

brentschmaltz:
@PinpointTownes It looks like you are creating metadata for the discovery endpoint. ECDsa would expose public data in a different way and would return null when cast to RSA even though it is an AsymmetricSecurityKey. Is it reasonable to handle each key individually?
Or better have a method AsymmetricSecurityKey that serves up publicInfo.

PinpointTownes:
>  It looks like you are creating metadata for the discovery endpoint. ECDsa would expose public data in a different way and would return null when cast to RSA even though it is an AsymmetricSecurityKey.

But this can't happen, thanks to the `!credentials.SigningKey.IsSupportedAlgorithm(SecurityAlgorithms.RsaSha256Signature` check.

FWIW, here's the pattern I'm forced to use with IM 5:

```csharp
// Ignore the key if it's not supported.
if (!(credentials.Key is AsymmetricSecurityKey) &&
     !credentials.Key.IsSupportedAlgorithm(SecurityAlgorithms.RsaSha256Signature)) {
    Logger.LogInformation("An unsupported signing key was ignored and excluded " +
                          "from the key set: {Type}. Only asymmetric security keys " +
                          "supporting RS256, RS384 or RS512 can be exposed " +
                          "via the JWKS endpoint.", credentials.Key.GetType().Name);

    continue;
}

RSA algorithm = null;

// Note: IdentityModel 5 doesn't expose a method allowing to retrieve the underlying algorithm
// from a generic asymmetric security key. To work around this limitation, try to cast
// the security key to the built-in IdentityModel types to extract the required RSA instance.
// See https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/395
var x509SecurityKey = credentials.Key as X509SecurityKey;
if (x509SecurityKey != null) {
    algorithm = (RSA) x509SecurityKey.PublicKey;
}

var rsaSecurityKey = credentials.Key as RsaSecurityKey;
if (rsaSecurityKey != null) {
    algorithm = rsaSecurityKey.Rsa;

    // If no RSA instance can be found, create one using
    // the RSA parameters attached to the security key.
    if (algorithm == null) {
        algorithm = RSA.Create();
        algorithm.ImportParameters(rsaSecurityKey.Parameters);
    }
}

// Export the RSA public key to create a new JSON Web Key
// exposing the exponent and the modulus parameters.
var parameters = algorithm.ExportParameters(includePrivateParameters: false);
Debug.Assert(parameters.Exponent != null, "A null exponent was returned by RSA.ExportParameters()");
Debug.Assert(parameters.Modulus != null, "A null modulus was returned by RSA.ExportParameters()");
```

brentschmaltz:
@PinpointTownes I agree our OM requires you to check for a specific key type. So, if I understand your ask, is it:
if an AsymmetricSecurityKey supports RsaSha256Signature, it should serve up an RSA object and you shouldn't have to fish it out?

PinpointTownes:
> if an AsymmetricSecurityKey supports RsaSha256Signature, it should serve up an RSA object and you shouldn't have to fish it out?

Yeeeeeeeeeeeeeeees!

brentschmaltz:
OK, got it.
We are at the finish line for version. Can we open an issue?

PinpointTownes:
For `GetAsymmetricAlgorithm`/`GetSymmetricAlgorithm`? There's already one: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/395

PinpointTownes:
Note that having `GetAsymmetricAlgorithm`/`GetSymmetricAlgorithm` is not just for convenience. It's also needed for people to use custom keys, than can't be supported with casts.

brentschmaltz:
You still have to cast to RSA, what I am thinking is you should be able to get to the public asymmetric bits for metadata without casting. Something like: ToXml or ToJson.

We don't have the runway for this version though.

PinpointTownes:
> what I am thinking is you should be able to get to the public asymmetric bits for metadata without casting. Something like: ToXml or ToJson.

No need for something complicated. Being able to retrieve the underlying algorithm would be enough to cover most cases and can be easily added in this PR.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
brentschmaltz: I checked, yes it checks for nulls.

Issue #301, opened 2015-10-29
ConfigureAwait?


danielwertheim:
Unfortunately I'm in the position where I need to make a blocking call (`foo.Result`) and in the chain underneath, the `ConfigurationManager.GetAsync` method is used with `await` and `ConfigureAwait(false)` but I do experience some deadlocks (which goes away when not using `ConfigurationManager` but other async retrieving components).

Could this be, since you don't use `ConfigureAwait(false)` e.g. here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/Microsoft.IdentityModel.Protocol.Extensions/Configuration/ConfigurationManager.cs#L170

brentschmaltz:
@danielwertheim we put the ConfigureAwait(false) on HttpDocumentRetriever https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs#L88

Are you are saying every 'await' in the chain needs .ConfigureAwait(false) ?

danielwertheim:
I was trying both the GenericDocumentRetriever´and the HttpDocumentRetriver. Btw...

Blocking call? Why?

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs#L79

brentschmaltz:
We chose to block to check the Response code.

danielwertheim:
Yes, but there's no need to do a blocking call there since you are in an async context. Just await it. Get the response. Check the status on it and continue the flow.

```
var response = await _httpClient.GetAsync(address, cancel).ConfigureAwait(false);
response.EnsureSuccessStatusCode();
```

I usually have an extension method for the `ConfigureAwait....`

```
    public static class TaskExtensions
    {
        public static ConfiguredTaskAwaitable<T> ForAwait<T>(this Task<T> t)
        {
            return t.ConfigureAwait(false);
        }
    }
```

Can also make that one return the `Task<T>` instead if you want.

Then

```
var response = await _httpClient.GetAsync(address, cancel).ForAwait();
response.EnsureSuccessStatusCode();
```

brentschmaltz:
[...]

PR #448, opened 2016-06-13
Add ConfigureAwait for await to avoid deadlock

ChenYan98:
#301


brentschmaltz:
:shipit:

tushargupta51:
@ChenYan98 Have you compared 4.x and 5.x to make sure we have covered all the places where we should add the configureAwait?

brentschmaltz:
We had just missed those two.

:shipit:

ChenYan98:
@tushargupta51 @brentschmaltz Actually, I found we do missed this one on 5.x: await _refreshLock.WaitAsync(cancel); but I am not sure if it is good time to fix it now. Could you make decision?

brentschmaltz:
We need separate PR for KDev. I will merge this into 4.x

PR #457, opened 2016-06-27
Update url and version for cli

brentschmaltz:
cli downloads url and version needs updating


brentschmaltz:
@tushargupta51 have a look

tushargupta51:
:shipit:

Issue #454, opened 2016-06-20
Introduce new property "X5t" in X509SecurityKey


ChenYan98:
The Kid we set in the in the constructor is the certificate.Thumbprint. That Kid is not the correct value to use if a user wants to set x5t in the JwtHeader. It should be Base64UrlEncoded(cert.Thumbprint); We can leave Kid alone and introduce a Property: X509SecurityKey.X5t { get; }. We can decide if we want to automatically add the x5t in the JwtHeader.

PR #460, opened 2016-07-01
Introduce new property "X5t" in X509SecurityKey

ChenYan98:


ChenYan98:
#454

PinpointTownes:
Hum, this change breaks the existing layering.

`x5t` is a JWK-specific property and I'm not sure it's a good idea to add it to `X509SecurityKey`, which is part of `Microsoft.IdentityModel.Tokens`. Maybe it would be better to simply expose a `byte[]` `Thumbprint` property and make the base64url-encoding in the JWT handler.

brentschmaltz:
I understand the layering, but following that will result in a base64-url encoding each time we want to match a key to an x5t. I don't think there is anything wrong with offering this property.

src/Microsoft.IdentityModel.Tokens/X509SecurityKey.cs
```diff
@@ -54,6 +54,7 @@ public X509SecurityKey(X509Certificate2 certificate)
 
             _certificate = certificate;
             KeyId = certificate.Thumbprint;
+            X5t = Base64UrlEncoder.Encode(certificate.Thumbprint);
```
brentschmaltz: I think this should be Base64UrlEnoder.Encode(certificate.GetCertHash()) according to https://tools.ietf.org/html/rfc7517#section-4.8

Issue #422, opened 2016-05-31
SecurityToken exceptions should include details as properties


Tratcher:
E.g. SecurityTokenInvalidAudienceException should have an InvalidAudiance property containing the invalid value. Similarly for SecurityTokenInvalidIssuerException, SecurityTokenInvalidLifetimeException, SecurityTokenNotYetValidException, SecurityTokenExpiredException, and maybe SecurityTokenInvalidSignatureException and SecurityTokenSignatureKeyNotFoundException.

RE: https://github.com/aspnet/Security/pull/845

polita:
@Tratcher We've moved this to the backlog because we don't think we can get this done for RTM.

polita:
@Tratcher This info is in the exception message for logging and debugging. We don't think this needs to be specific properties in the exception types.

Tratcher:
@polita @brentschmaltz please reconsider, the current messages are not suitable for our scenario (sending to a remote client). We want to generate our own messages with the relevant details.

brentschmaltz:
@Tratcher can you give us an example for say: InvalidAudienceException. What is bad about the current message, what your new message would look like and what would the new InvalidAudienceException properties return.

Tratcher:
The issue is not the current messages per say, the issue is that we are crossing a trust boundary and cannot pass through unknown data. Your messages are "for logging and debugging", where ours are for remote users. Even if we validated all of your messages now, we can't vet them every time you make changes.  We just want the raw data and we'll make sure what we send to the client is appropriate.

Prior discussion for reference: https://github.com/aspnet/Security/pull/845#issuecomment-221709584

polita:
Done with #479.

Tratcher:
Milestone Backlog? What release was this in?

brentschmaltz:
@Tratcher 5.1.0

PR #479, opened 2016-07-25
New methods in LogHelper

polita:
Move to LogHelper.LogExceptionMessage methods

The existing LogHelper methods instantiate the exception in them, which is convenient,
but it makes it impossible to set properties on typed exceptions to give more info
to the developer. As part of moving toward more informative exceptions with properties,
methods will now create the exception locally and then call LogHelper methods to log
the exception. Existing LogHelper methods were public so they can't be removed,
but internally we'll stop using them and instead follow a pattern of instantiating the
exception locally to the calling code.

This is the first part of the change to make more specific exceptions with properties covered by Issue #422. Part two will follow, but the change was big, so I wanted to get it in in better-sized chunks.


Tratcher:
oooo, this looks promising.
cc: @PinpointTownes

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
```diff
@@ -57,7 +58,7 @@ public TextWriterEventListener()
             }
             catch (Exception ex)
             {
-                LogHelper.LogException<InvalidOperationException>(ex, LogMessages.MIML11001);
+                LogHelper.LogExceptionMessage(new InvalidOperationException(LogMessages.MIML11001, ex));
                 throw ex;
```
ilmax: just curious, shouldn't this be just 
```csharp
LogHelper.LogExceptionMessage(new InvalidOperationException(LogMessages.MIML11001, ex));
throw
```

Doesn't `throw ex` changes the StackTrace?

PinpointTownes:
> oooo, this looks promising.
cc: @PinpointTownes

Agreed. The good point is that nothing critical is exposed (e.g the expected signature).

That said, I'm not sure why `LogHelper.LogArgumentNullException` and its friends are being replaced by much longer methods. AFAICT, there's no point doing that, as `LogArgumentNullException` already takes the parameter name.

polita:
@PinpointTownes:

> The existing LogHelper methods instantiate the exception in them, which is convenient,
 but it makes it impossible to set properties on typed exceptions to give more info
 to the developer.

If I could remove the Log*Exception methods that Activator.CreateInstance to create the exceptions, I would, but they're public. I generally feel that calling Activator.CreateInstance to create an exception instance isn't a good pattern.

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
polita: Good point. Updating- thanks.

PinpointTownes:
I get your point about the fact you can't easily set properties when you don't instantiate them but my remark was specific to `ArgumentNullException`/`ArgumentException`/`InvalidOperationException`.

Your changes added much noise in the "parameters validation" regions. I'm still unsure why you couldn't update `LogArgumentNullException` to replace reflection by a good old `new ArgumentNullException(argument)`.

polita:
For LogArgumentNullException, yes, you're right- we could change that to just create the exception. For LogArgumentNullException (and all it's ridiculous number of overrides), the method is still generic, meaning you can type it to a derived type of ArgumentException, and then not be able to set any of the specific properties on the type. I went for consistency in the model across all exception types.

I think that the code that finds the exception case should new up the exception. It removes a level of indirection and ensures the caller has control over exactly what the created exception looks like.

polita:
@brentschmaltz We now have properties on SecurityTokenExpiredException, SecurityTokenInvalidAudienceException, SecurityTokenInvalidIssuerException, SecurityTokenInvalidLifetimeException, SecurityTokenInvalidSigningKeyException, and SecurityTokenNotYetValidException. I've also added test code to verify the values for the properties anytime we create those exceptions. Can you take a look?

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
```diff
@@ -83,7 +84,7 @@ public TextWriterEventListener(string filePath)
         public TextWriterEventListener(StreamWriter streamWriter)
         {
             if (streamWriter == null)
-                throw LogHelper.LogArgumentNullException("streamWriter");
+                throw LogHelper.LogExceptionMessage(new ArgumentNullException("streamWriter", String.Format(CultureInfo.InvariantCulture, LogMessages.IDX10000, "streamWriter")));
```
brentschmaltz: While not architecturally pure, having LogHelper.LogArgumentNullException("...") create a new ArgumentNullException would reduce code change.

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenExpiredException.cs
```diff
@@ -38,6 +38,11 @@ namespace Microsoft.IdentityModel.Tokens
     public class SecurityTokenExpiredException : SecurityTokenValidationException
     {
         /// <summary>
+        /// Gets or sets the Expires value that created the validation exception.
+        /// </summary>
+        public DateTime? Expires { get; set; }
```
brentschmaltz: When would this ever be null? Why DateTime? and not DateTime.

src/Microsoft.IdentityModel.Tokens/X509SecurityKey.cs
```diff
@@ -147,5 +147,27 @@ public X509Certificate2 Certificate
...
+        /// Returns a bool indicating if this key is equivalent to another key.
+        /// </summary>
+        /// <return>true if the keys are equal; otherwise, false.</return>
+        public override bool Equals(object obj)
+        {
+            X509SecurityKey other = obj as X509SecurityKey;
+            if (other == null)
+                return false;
+
+            return other.Certificate.ToString() == this._certificate.ToString();
```
brentschmaltz: X509Certificate2.ToString() is a pretty heavy call. Would checking the Thumbprint suffice? Why do we introduce this method?

polita:
@brentschmaltz: I took all of your suggestions. The X509SecurityKey Equals override is used only in unit tests to compare the values of the SecurityTokenInvalidSigningKeyException.SigningKey value, but I changed it to compare Thumbnail anyway, as you suggested. Good catch on the nullable NotBefore and Expires.

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
```diff
@@ -69,7 +69,7 @@ public TextWriterEventListener()
         public TextWriterEventListener(string filePath)
         {
             if (string.IsNullOrEmpty(filePath))
-                throw LogHelper.LogArgumentNullException("filePath");
+                throw LogHelper.LogArgumentNullException(filePath);
```
brentschmaltz: did you want nameof(filePath) here?

src/Microsoft.IdentityModel.Protocols/AuthenticationProtocolMessage.cs
```diff
@@ -146,7 +145,7 @@ public string IssuerAddress
             set
             {
                 if (value == null)
-                    throw LogHelper.LogExceptionMessage(new ArgumentNullException("value", String.Format(CultureInfo.InvariantCulture, LogMessages.IDX10000, "value"))); 
+                    throw LogHelper.LogArgumentNullException(value);
```
brentschmaltz: "value"

brentschmaltz:
Couple of typo's, fix then :shipit:

PinpointTownes:
> Also, through a series of bad rebasing decisions, I ended up with changes from other PRs in this PR. It should merge in smoothly. I have learned my lesson.

That's because you're using `git merge` instead of a proper rebasing (`git rebase`). The good news is that it's never too late to fix that. Simply rebase your PR on top of the current `dev` branch and all these unwanted commits should disappear.

```
git fetch AzureAD
git rebase AzureAD/dev
git push -f
```

polita:
@PinpointTownes Thanks for the tip. Next time, I'm going to totally have the guts to try it. :) From now on, only rebases, no merges for me.

Issue #486, opened 2016-08-17
JsonWebKeyECTypes.P512 doesn't correspond to a standardized named curve


PinpointTownes:
`P512` is likely a typo, as there's no such curve in the DSS spec: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf

Instead, it should be named `P521` and the constant value should be `P-521`. Since the existing values are invalid, I'd recommend removing them in the next version.

https://tools.ietf.org/html/rfc7518#section-6.2.1.1

/cc @brentschmaltz @polita

polita:
Fixed.

PR #503, opened 2016-09-26
Update JsonWebKeyECTypes.P512 to P521 without breaking changes

polita:
#486


PinpointTownes:
FYI, a quick GitHub search shows that ASOS is the only public project using `JsonWebKeyECTypes`: https://github.com/search?q=JsonWebKeyECTypes&type=Code.

Since treating `P512` like `P521` is very likely to produce unexpected results (i.e unverifiable signature), removing the invalid constant might be a better approach, even if it's a binary/source breaking change.

brentschmaltz:
:shipit:

polita:
@PinpointTownes Unfortunately, we can't make any breaking changes in minor or patch versions. As @brentschmaltz says, we can remove later.

Issue #477, opened 2016-07-21
AsymmetricSignatureProvider abusively disposes RSACryptoServiceProvider instances it doesn't own on .NET Desktop


PinpointTownes:
On .NET Desktop, signing or verifying a token more than once when the security key is a `RsaSecurityKey` results in an exception if the underlying algorithm instance is a `RSACryptoServiceProvider`:

```
System.ObjectDisposedException: Le handle sécurisé a été fermé
   à System.Security.Cryptography.Utils._GetKeyParameter(SafeKeyHandle hKey, UInt32 paramID)
   à System.Security.Cryptography.RSACryptoServiceProvider.get_KeySize()
   à Microsoft.IdentityModel.Tokens.RsaSecurityKey.get_KeySize()
   à Microsoft.IdentityModel.Tokens.AsymmetricSignatureProvider.ValidateAsymmetricSecurityKeySize(SecurityKey key, String algorithm, Boolean willCreateSignatures)
   à Microsoft.IdentityModel.Tokens.AsymmetricSignatureProvider..ctor(SecurityKey key, String algorithm, Boolean willCreateSignatures)
   à Microsoft.IdentityModel.Tokens.CryptoProviderFactory.CreateProvider(SecurityKey key, String algorithm, Boolean willCreateSignatures)
   à Microsoft.IdentityModel.Tokens.CryptoProviderFactory.CreateForSigning(SecurityKey key, String algorithm)
   à System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.CreateEncodedSignature(String input, SigningCredentials signingCredentials)
   à System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.WriteToken(SecurityToken token)
 `AsymmetricSignatureProvider.Dispose()`
```

This bug is likely caused by `Dispose()`, that disposes the underlying `RSACryptoServiceProvider` instance without ensuring it owns it: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs#L691

(it's likely the same as the one an ASOS user experienced last month when trying to use a CSP HSM)

polita:
[...]

PR #518, opened 2016-10-14
control when _rsa is disposed

brentschmaltz:
Fixes incorrect disposing of RSA
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/477


brentschmaltz:
@PinpointTownes thanks.

PR #529, opened 2016-11-09
Add telemetry to the authorization and signout message redirect urls,…

polita:
… along with a way to turn it off.

Also added tests for the telemetry parameters and the switch to turn it off.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocol.Extensions/OpenIdConnectMessage.cs
+++ b/src/Microsoft.IdentityModel.Protocol.Extensions/OpenIdConnectMessage.cs
@@ -19,6 +19,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
+using System.Reflection;
 using System.Security.Cryptography.X509Certificates;
 
 namespace Microsoft.IdentityModel.Protocols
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Protocol.Extensions/OpenIdConnectMessage.cs
```diff
@@ -137,12 +139,30 @@ public string CreateAuthenticationRequestUrl()
...
             openIdConnectMessage.RequestType = OpenIdConnectRequestType.LogoutRequest;
             return openIdConnectMessage.BuildRedirectUrl();
         }
 
         /// <summary>
+        /// Adds telemetry values to the message parameters.
+        /// </summary>
+        private void EnsureTelemetryValues()
+        {
+            if (DisableTelemetryParameters)
```
polita: Yes, if you were guaranteed that Clone was always called before EnsureTelemetryValues. My concern is that, in some future iteration of the code, someone changes the flow to not call Clone first, and then they expect that setting DisableTelemetryValues to true, and then to false, would remove them again. I can see your point though; if you think we shouldn't set up the code to ever remove the parameters, I'm fine with changing it.

src/Microsoft.IdentityModel.Protocol.Extensions/OpenIdConnectMessage.cs
```diff
@@ -137,12 +139,25 @@ public string CreateAuthenticationRequestUrl()
         /// <returns>The uri to use for a redirect.</returns>
         public string CreateLogoutRequestUrl()
         {
+            EnsureTelemetryValues();
```
brentschmaltz: You may want to consider adding telemetry on the cloned message. Otherwise turning it off, would result in the second message having the parameters.

polita:
@brentschmaltz The PR is updated with your suggestions. Thanks...

polita:
@brentschmaltz Can you CR?

polita:
Updated with CR feedback from 5x change.

src/Microsoft.IdentityModel.Protocol.Extensions/OpenIdConnectMessage.cs
```diff
@@ -239,6 +262,17 @@ public string DomainHint
...
         /// <summary>
+        /// Gets or sets whether parameters for the library and version are sent on the query string for this <see cref="OpenIdConnectMessage"/> instance. 
+        /// This value is set to the value of EnableTelemetryParametersByDefault at message creation time.
+        /// </summary>
+        public bool EnableTelemetryParameters { get; set; }
+
+        /// <summary>
+        /// Gets or sets whether parameters for the library and version are sent on the query string for all instances of <see cref="OpenIdConnectMessage"/>. 
+        /// </summary>
+        public static bool EnableTelemetryParametersByDefault = true;
```
brentschmaltz: Could you make this a property also :-)

polita:
@brentschmaltz Updated with the same property pattern from 5x using pre-C# 6.0 syntax.

brentschmaltz:
:shipit:

PR #531, opened 2016-11-11
Add telemetry values to auth request and logout URIs, along with a sw…

polita:
…itch to turn telemetry off

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
@@ -31,6 +31,7 @@ using System.Collections.Specialized;
 using Microsoft.IdentityModel.Logging;
 using Newtonsoft.Json.Linq;
 using System.Globalization;
+using System.Reflection;
 
 namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
 {
@@ -39,6 +40,8 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
     /// </summary>
     public class OpenIdConnectMessage : AuthenticationProtocolMessage
     {
+        private const string _skuTelemetryValue = "ID_NET";
+
         /// <summary>
         /// Initializes a new instance of the <see cref="OpenIdConnectMessage"/> class.
         /// </summary>
@@ -82,6 +85,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
             IssuerAddress = other.IssuerAddress;
             RequestType = other.RequestType;
             TokenEndpoint = other.TokenEndpoint;
+            EnableTelemetryParameters = other.EnableTelemetryParameters;
         }
 
         /// <summary>
@@ -169,6 +173,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         {
             OpenIdConnectMessage openIdConnectMessage = Clone();
             openIdConnectMessage.RequestType = OpenIdConnectRequestType.Authentication;
+            EnsureTelemetryValues(openIdConnectMessage);
             return openIdConnectMessage.BuildRedirectUrl();
         }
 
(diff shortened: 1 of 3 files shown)


brentschmaltz:
There is some feedback from the community that OFF by default seems to be preferred.
https://github.com/aspnet/Security/issues/1035 (I accidentally clicked +1 on the original issue opened by pinpointtownes)

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectMessageTests.cs
```diff
@@ -165,15 +165,14 @@ public void GetSets()
             OpenIdConnectMessage message = new OpenIdConnectMessage();
             Type type = typeof(OpenIdConnectMessage);
             PropertyInfo[] properties = type.GetProperties();
-            if (properties.Length != 47)
+            if (properties.Length != 48)
                 Assert.True(true, "Number of public fields has changed from 47 to: " + properties.Length + ", adjust tests");
```
Eilon: Change msg to 48 as well 😄

PinpointTownes:
> I accidentally clicked +1 on the original issue opened by pinpointtownes

Are you sure it was accidental? :trollface:

Eilon:
@brentschmaltz BTW you can un-thumbs-up by clicking the 👍 icon at the bottom of the issue again.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectMessageTests.cs
polita: Thank you. Fixed.

brentschmaltz:
@PinpointTownes :-) always a pleasure.

polita:
@brentschmaltz Can you CR?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
```diff
@@ -280,6 +304,20 @@ public string DomainHint
         }
 
         /// <summary>
+        /// Gets or sets whether parameters for the library and version are sent on the query string for this <see cref="OpenIdConnectMessage"/> instance. 
+        /// This value is set to the value of EnableTelemetryParametersByDefault at message creation time.
+        /// </summary>
+        public bool EnableTelemetryParameters
```
brentschmaltz: The only thing I can think of is if we did

{get;set;} = EnableTelemetryParametersByDefault;

would remove the need for a few sets in some methods.

PinpointTownes:
I don't wanna be evil, but note that thanks to the "lowest applicable version" policy used by NuGet, folks using the OIDC middleware won't get telemetry until a new OIDC package is released, unless they explicitly specify the latest IdentityModel version in their project.json :smile: 

... so maybe a static property is not necessary...? :trollface:

polita:
@PinpointTownes Yes, it's true that you'd have to explicitly download the new package. But since people might be pulling down a new package for other reasons than the awesome new telemetry feature, giving them a single line of code way to turn it off is still desirable. 

At least, I thought for sure you'd be a proponent of that... :)

polita:
@brentschmaltz Thanks- updated the CR and squashed.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
brentschmaltz: @polita I don't see this in current PR. I was thinking this should be on the Property EnableTelemetryParameters

polita:
@brentschmaltz Updated with the change to use auto-property initializers.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
```diff
@@ -280,6 +298,17 @@ public string DomainHint
...
         /// <summary>
+        /// Gets or sets whether parameters for the library and version are sent on the query string for this <see cref="OpenIdConnectMessage"/> instance. 
+        /// This value is set to the value of EnableTelemetryParametersByDefault at message creation time.
+        /// </summary>
+        public bool EnableTelemetryParameters { get; set; } = EnableTelemetryParametersByDefault;
+
+        /// <summary>
+        /// Gets or sets whether parameters for the library and version are sent on the query string for all instances of <see cref="OpenIdConnectMessage"/>. 
+        /// </summary>
+        public static bool EnableTelemetryParametersByDefault = true;
```
PinpointTownes: Out of curiosity, why not a property? :open_mouth:

polita: What would a property getter/setter get you that this doesn't?

PinpointTownes: The immense satisfaction of respecting the best practices? :trollface: 

Exposing public fields is generally discouraged (and I'm pretty sure there's even an FxCop rule for that)

Tratcher: Property implementations can be changed later without breaking callers. Fields can be changed to properties later but it requires recompiling the caller.

brentschmaltz: @Tratcher are you thinking something like, we set the property from an external source (config, etc)? Where the property could have some smarts?

polita: Updated to make it a property. Thanks for the feedback.

brentschmaltz:
Well that was easy :-)
:shipit:

PR #533, opened 2016-11-17
Rsa key wrap

ChenYan98:


src/Microsoft.IdentityModel.Tokens/Encryption/RsaKeyWrapProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+//------------------------------------------------------------------------------
+
+using System;
+using System.Security.Cryptography;
+using System.Globalization;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    public class RsaKeyWrapProvider
```
brentschmaltz: We need comments on classes and all properties

src/Microsoft.IdentityModel.Tokens/Encryption/RsaKeyWrapProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            if (key == null)
+                return false;
+
+            if (string.IsNullOrEmpty(algorithm))
+                return false;
+
+            if (algorithm.Equals(SecurityAlgorithms.RsaPKCS1, StringComparison.Ordinal) || algorithm.Equals(SecurityAlgorithms.RsaOAEP, StringComparison.Ordinal))
+            {
+                if (key is SymmetricSecurityKey)
```
brentschmaltz: This doesn't seem right. RSA keywrap requires RSA key.

brentschmaltz:
@ChenYan98 this should be committed into dev.  you will need to rebase RsaKeyWrap against dev, then resubmit the PR

Issue #527, opened 2016-11-07
OpenIdConnectProtocolValidator should have 'RequireSub' true by default.


brentschmaltz:
see: http://openid.net/specs/openid-connect-core-1_0.html#IDToken
sub claim is required.

ChenYan98:
#538

Issue #528, opened 2016-11-07
Ensure OpenIdConnectProtocolValidator follows spec for hybrid flow or deviations are recorded on wiki.


Tratcher:
When doing a code+id_token hybrid flow there are some additional validations required between the id_token from the authorization endpoint and the one from the token endpoint.
http://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken2
1) the sub claim must match.
2) the iss claim must match.

keegan-caruso:
@keegan-caruso To move issue to Identity Web

PR #538, opened 2016-11-29
Set "RequireSub" to true by default

ChenYan98:
#527

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
@@ -83,12 +83,6 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         public static readonly TimeSpan DefaultNonceLifetime = TimeSpan.FromMinutes(60);
 
         /// <summary>
-        /// Default for the RequreSub value.
-        /// </summary>
-        /// <remarks>default: true.</remarks>
-        public static bool RequireSubByDefault = true;
-
-        /// <summary>
         /// Creates a new instance of <see cref="OpenIdConnectProtocolValidator"/>,
         /// </summary>
         public OpenIdConnectProtocolValidator()
@@ -99,7 +93,6 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
             RequireAzp = false;
             RequireNonce = true;
             RequireState = true;
-            RequireSub = RequireSubByDefault;
             RequireTimeStampInNonce = true;
             RequireStateValidation = true;
 
@@ -206,6 +199,13 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         /// </summary>
         [DefaultValue(true)]
         public bool RequireSub { get; set; } = RequireSubByDefault;
+
+
+        /// <summary>
+        /// Gets or sets a value for default RequreSub.
+        /// </summary>
+        /// <remarks>default: true.</remarks>
+        public static bool RequireSubByDefault { get; set; } = true;
 
         /// <summary>
         /// Gets or set logic to control if a nonce is prefixed with a timestamp.
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
@@ -83,6 +83,12 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         public static readonly TimeSpan DefaultNonceLifetime = TimeSpan.FromMinutes(60);
 
         /// <summary>
+        /// Default for the RequreSub value.
+        /// </summary>
+        /// <remarks>default: true.</remarks>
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -83,6 +83,12 @@ public class OpenIdConnectProtocolValidator
         public static readonly TimeSpan DefaultNonceLifetime = TimeSpan.FromMinutes(60);
 
         /// <summary>
+        /// Default for the RequreSub value.
+        /// </summary>
+        /// <remarks>default: true.</remarks>
+        public static bool RequireSubByDefault = true;
```
brentschmaltz: Please make this a property.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -93,7 +99,7 @@ public OpenIdConnectProtocolValidator()
             RequireAzp = false;
             RequireNonce = true;
             RequireState = true;
-            RequireSub = false;
+            RequireSub = RequireSubByDefault;
```
brentschmaltz: I don't believe this is needed since you are setting it at the property with } = RequireSubByDefault

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
ChenYan98: Modified

brentschmaltz:
:shipit: please rebase before merging.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -198,8 +197,15 @@ public TimeSpan NonceLifetime
         /// <summary>
         /// Gets or sets a value indicating if a 'sub' claim is required.
         /// </summary>
-        [DefaultValue(false)]
-        public bool RequireSub { get; set; }
+        [DefaultValue(true)]
```
polita: The DefaultValueAttribute is just for designers to use to set a default value in the UI, which doesn't apply to Wilson. Why do we set the DefaultValueAttribute?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -198,8 +197,15 @@ public TimeSpan NonceLifetime
...
+        public bool RequireSub { get; set; } = RequireSubByDefault;
```
polita: Is changing this from optional to required going to break compatibility with other IDPs that might not provide the subject in some cases? Can we confidently do this without breaking anyone's existing code?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
ChenYan98: Just following the existing format...

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
ChenYan98: According to the spec, the subject is required. There are some cases that might not provide the subject, and that is why we give the user the option to change RequireSubByDefault to false. @brentschmaltz , can you confirm this?

brentschmaltz: @polita @ChenYan98 This may be breaking to some users. The specification requires a 'sub' claim, we relaxed that in our first versions as some IDP's were not including the sub claim. Our old design would have required users to derive from OpenIdConnectProtocolValidator to choose a different value.  This design, allows a simpler implementation just deploy with setting RequireSubByDefault = false.

brentschmaltz: @polita Yes, if users deploy x.1.1 and their IDP doesn't supply the sub, things will start to fail. However, their IDP is not spec compliant. For their users to be affected, they would have had to deploy directly to production without testing.

brentschmaltz: In addition, https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/528 an additional protocol requirement, cannot be satisfied unless the 'sub' claim exists.

polita: @brentschmaltz I'm hesitant on this one because it feels like a potential breaking behavioral change, but if you're feeling confident that we're not going to break people in an unexpected way, I trust your judgement. :)

polita:
:shipit:

Issue #488, opened 2016-08-17
CryptoProviderFactory.IsEcdsaAlgorithmSupported() doesn't ensure algorithms are compatible with the curves used by ECDSA keys


PinpointTownes:
`IsEcdsaAlgorithmSupported` returns `true` if the algorithm corresponds to `ES256`, `ES384` or `ES512` but doesn't ensure the curve type of the ECDSA key is compatible with the specified algorithm.

`ES256` -> `P-256` (= `ECCurve.NamedCurves.nistP256`)
`ES384` -> `P-384` (= `ECCurve.NamedCurves.nistP384`)
`ES512` -> `P-521` (= `ECCurve.NamedCurves.nistP521`)

/cc @brentschmaltz @polita

brentschmaltz:
moving to 5.1.4

PinpointTownes:
@mafurman why was this thread closed? I don't see in the PR diff any chance impacting the APIs mentioned in this ticket.

brentschmaltz:
@PinpointTownes @mafurman Yes, we need a check between 'algorithm' and 'CRV'. I would argue this is not a IsEcdsaAlgorithmSupported function. As this describes what the runtime is capable of.
The check you are looking for is matching CRV with Algorithm. I think this is covered with the combination of GetKeySize and ValidateECDSAKeySize.

PinpointTownes:
Let me remind you what the requirement is, in case you'd have missed it (well, this ticket was opened almost 2 years ago so...):

I need `CryptoProviderFactory.IsSupportedAlgorithm(string algorithm, SecurityKey key)` to return `false` if `key` is an ECDSA key AND `algorithm` an ECDSA algorithm that uses a curve different than the one associated with `key`.

brentschmaltz:
We will add tests to make sure that contract is enforced.

PR #542, opened 2016-12-02
Chenyan98/488

ChenYan98:
Leave a comment

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -88,6 +88,19 @@ namespace Microsoft.IdentityModel.Tokens
         };
 
         /// <summary>
+        /// Mapping from algorithm to <see cref="ECDsa"/>.KeySize.
+        /// </summary>
+        private static readonly Dictionary<string, int> DefaultECDsaKeySizeInBitsMap = new Dictionary<string, int>()
+        {
+            { SecurityAlgorithms.EcdsaSha256, 256 },
+            { SecurityAlgorithms.EcdsaSha384, 384 },
+            { SecurityAlgorithms.EcdsaSha512, 521 },
+            { SecurityAlgorithms.EcdsaSha256Signature, 256 },
+            { SecurityAlgorithms.EcdsaSha384Signature, 384 },
+            { SecurityAlgorithms.EcdsaSha512Signature, 521 }
+        };
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="AsymmetricSignatureProvider"/> class used to create and verify signatures.
         /// </summary>
         /// <param name="key">The <see cref="SecurityKey"/> that will be used for signature operations.<see cref="SecurityKey"/></param>
@@ -254,7 +267,14 @@ namespace Microsoft.IdentityModel.Tokens
                 if (ecdsaKey.ECDsa != null)
                 {
                     _ecdsa = ecdsaKey.ECDsa;
-                    return;
+                    if (_ecdsa != null && ValidateECDSAKeySize(_ecdsa.KeySize, algorithm))
+                    {
+                        return;
+                    }
+                    else
+                    {
+                        throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException("key.KeySize", String.Format(CultureInfo.InvariantCulture, LogMessages.IDX10671, key, DefaultECDsaKeySizeInBitsMap[algorithm], key.KeySize)));
+                    }
                 }
             }
 
@@ -276,8 +296,16 @@ namespace Microsoft.IdentityModel.Tokens
                 if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                     throw new PlatformNotSupportedException();
 
-                CreateECDsaFromJsonWebKey(webKey, willCreateSignatures);
-                return;
+                _ecdsa = CreateECDsaFromJsonWebKey(webKey, algorithm, willCreateSignatures);
+                _disposeEcdsa = true;
+                if (_ecdsa != null)
+                {
+                    return;
+                }
+                else
+                {
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(String.Format(CultureInfo.InvariantCulture, LogMessages.IDX10672, key)));
+                }
             }
 
             throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(key), String.Format(CultureInfo.InvariantCulture, LogMessages.IDX10641, key)));
@@ -357,8 +385,15 @@ namespace Microsoft.IdentityModel.Tokens
                 if (ecdsaKey.ECDsa != null)
                 {
                     _ecdsa = ecdsaKey.ECDsa as ECDsaCng;
-                    _ecdsa.HashAlgorithm = new CngAlgorithm(_hashAlgorithm);
-                    return;
+                    if (_ecdsa != null && ValidateECDSAKeySize(_ecdsa.KeySize, algorithm))
+                    {
+                        _ecdsa.HashAlgorithm = new CngAlgorithm(_hashAlgorithm);
+                        return;
+                    }
+                    else
+                    {
+                        throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException("key.KeySize", String.Format(CultureInfo.InvariantCulture, LogMessages.IDX10671, key, DefaultECDsaKeySizeInBitsMap[algorithm], key.KeySize)));
+                    }
                 }
             }
 
@@ -372,8 +407,16 @@ namespace Microsoft.IdentityModel.Tokens
             }
             else if (webKey.Kty == JsonWebAlgorithmsKeyTypes.EllipticCurve)
             {
-                CreateECDsaFromJsonWebKey(webKey, willCreateSignatures);
-                return;
+                _ecdsa = CreateECDsaFromJsonWebKey(webKey, algorithm, willCreateSignatures);
+                _disposeEcdsa = true;
+                if (_ecdsa != null)
+                {
+                    return;
+                }
+                else
+                {
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(String.Format(CultureInfo.InvariantCulture, LogMessages.IDX10672, key)));
+                }
             }
 
             throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(key), String.Format(CultureInfo.InvariantCulture, LogMessages.IDX10641, key)));
@@ -417,7 +460,7 @@ namespace Microsoft.IdentityModel.Tokens
             return parameters;
         }
 
-        private void CreateECDsaFromJsonWebKey(JsonWebKey webKey, bool willCreateSignatures)
+        private ECDsaCng CreateECDsaFromJsonWebKey(JsonWebKey webKey, string algorithm, bool willCreateSignatures)
         {
             if (webKey == null)
                 throw LogHelper.LogArgumentNullException(nameof(webKey));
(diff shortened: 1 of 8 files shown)


ChenYan98:
#488

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -177,33 +178,51 @@ public virtual bool IsSupportedAlgorithm(string algorithm, SecurityKey key)
...
-        private bool IsEcdsaAlgorithmSupported(string algorithm)
+        private bool IsEcdsaAlgorithmSupported(string algorithm, ECDsa key)
         {
             switch (algorithm)
             {
                 case SecurityAlgorithms.EcdsaSha256:
-                case SecurityAlgorithms.EcdsaSha384:
-                case SecurityAlgorithms.EcdsaSha512:
                 case SecurityAlgorithms.EcdsaSha256Signature:
+                    if (key.KeySize == 256)
```
PinpointTownes: I am **definitely not a crypto expert**, but I doubt the key size alone is a very strong indicator for this kind of check, as nothing prevents me from using another curve than the ones expected here (respectively P-256, P-384 and P-521): if the key size is 256/384/521, it will return `true` even if the curve doesn't correspond.

Consider comparing the curve OID instead of the key size here.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -469,17 +471,15 @@ private void CreateECDsaFromJsonWebKey(JsonWebKey webKey, bool willCreateSignatu
...
                     }
                 }
                 else
                 {
                     Marshal.Copy(keyBlobPtr, keyBlob, 0, keyBlob.Length);
                     using (CngKey cngKey = CngKey.Import(keyBlob, CngKeyBlobFormat.EccPublicBlob))
                     {
-                        _ecdsa = new ECDsaCng(cngKey);
-                        _disposeEcdsa = true;
+                        return new ECDsaCng(cngKey);
```
brentschmaltz: @ChenYan98 please check if we have a test for different 'crv' and key sizes. For example a curve specified with 'P-521' and a key size of '256' should fail.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: @ChenYan98 in addition to the key size, we should query the ECDsa object for the cure.

ChenYan98: @PinpointTownes @brentschmaltz  First of all, I agree with you that we need check key size and curve for ECDsa object. But in our current API, there is no way to check curve. So I only to check key size for now. We can add logic to check ECDsa curve once the API support it.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
ChenYan98: We have test cases for different key size. But like I mentioned above, we cannot set ECDsa curve manually in current API.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
PinpointTownes: It's definitely possible, as I use a similar approach in ASOS to determine the best algo to expose via the JWKS endpoint:

```csharp
var parameters = algorithm.ExportParameters(includePrivateParameters: false);
var isP256 = parameters.Curve.IsNamed && parameters.Curve.Oid.FriendlyName == ECCurve.NamedCurves.nistP256.Oid.FriendlyName;
```

You'll need to target .NET Standard 1.6/.NET 4.6.3 for that, tho'.

brentschmaltz: @PinpointTownes right, we would need a new target of 1.6 for obtaining the the curve data. We have to support 1.4 for our .Net 4.6.1 users. So for our 1.4 target, we can't check the more than the size.

PinpointTownes: @brentschmaltz what's wrong with adding a new TFM?

brentschmaltz: @PinpointTownes nothing, we can do that. Just hadn't planned on that for our December release. Point is we need to support 1.4, so for that TFM we can only check the key.

PinpointTownes: The cool thing with adding a 1.6 TFM is that it would allow us to make the ECDSA support truly cross-platform (currently, it directly deals with CNG blobs, which can't work on Linux/MacOS: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs#L418-L555 :sweat_smile:)

brentschmaltz: @PinpointTownes that adds some value. The thing is every TFM has a tax associated with it. More ifdef's, test failures, backcompat, etc. For this issue of failing early, it doesn't seem to be enough of a reason to add 1.6 TFM for this issue. For sure we will need 2.0 so it comes down to a ROI.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -149,8 +149,9 @@ private bool IsSupportedKeyWrapAlgorithm(string algorithm, SecurityKey key)
         /// </summary>
         /// <param name="algorithm">the algorithm to check.</param>
         /// <param name="key">the <see cref="SecurityKey"/>.</param>
+        /// <param name="isPrivateKey">Whether the <see cref="SecurityKey"/> is private key then set this to true.</param>
         /// <returns>true if 'algorithm, key' pair is supported.</returns>
-        public virtual bool IsSupportedAlgorithm(string algorithm, SecurityKey key)
+        public virtual bool IsSupportedAlgorithm(string algorithm, SecurityKey key, bool isPrivateKey)
```
polita: Is this a breaking change? Do we need isPrivateKey?

ChenYan98: This is not a breaking change. It only let the failure happens early. The reason we need this parameter is,  'PrivateKey' is used in AsymmetricAlgorithm for signing, and 'PublicKey' is used for verification. Since I have to check the key size, I have to create a ECDsa instance with the correct parameters in JsonWebkey.

polita: You're changing the parameter list for a public virtual method on a class that's public and pre-existing. This is definitely a breaking change. You can't change the signature on a method that we've already shipped that's public virtual without breaking people who used the old signature. Please either figure out how to do this without isPrivateKey or you'll need to add a new method with that parameter.

ChenYan98: Yes, you are right. This is a breaking change. I neglect "Public" keyword by mistake...

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -88,6 +89,19 @@ public class AsymmetricSignatureProvider : SignatureProvider
         };
 
         /// <summary>
+        /// Mapping from algorithm to <see cref="ECDsa"/>.KeySize.
+        /// </summary>
+        public static readonly Dictionary<string, int> DefaultECDsaKeySizeInBitsMap = new Dictionary<string, int>()
```
PinpointTownes: I'm not sure to understand what's the point of having such a field public or making it configurable (the key size can't be changed, it's fixed). What will happen if I mess up with the values?

ChenYan98: Yes, you are right. I will change this map to 'private'.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -124,6 +138,7 @@ public AsymmetricSignatureProvider(SecurityKey key, string algorithm, bool willC
 
             _minimumAsymmetricKeySizeInBitsForSigningMap = new Dictionary<string, int>(DefaultMinimumAsymmetricKeySizeInBitsForSigningMap);
             _minimumAsymmetricKeySizeInBitsForVerifyingMap = new Dictionary<string, int>(DefaultMinimumAsymmetricKeySizeInBitsForVerifyingMap);
+            _ecdsaKeySizeInBitsMap = new Dictionary<string, int>(DefaultECDsaKeySizeInBitsMap);
```
polita: Do you need to make a copy of the original map? Are you altering it in some way? It looks like you're just looking for a key in it, not changing it, so you should be able to use the original without copying it.

ChenYan98: Yes, you are right since I change DefaultECDsaKeySizeInBitsMap from 'public' to 'private'.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -372,8 +407,16 @@ private void ResolveAsymmetricAlgorithm(SecurityKey key, string algorithm, bool
             }
             else if (webKey.Kty == JsonWebAlgorithmsKeyTypes.EllipticCurve)
             {
-                CreateECDsaFromJsonWebKey(webKey, willCreateSignatures);
-                return;
+                _ecdsa = CreateECDsaFromJsonWebKey(webKey, willCreateSignatures);
+                _disposeEcdsa = true;
+                if (_ecdsa != null && validateECDSAKeySize(_ecdsa, algorithm))
```
brentschmaltz: It would be better if CreateECDasFromJsonWebKey calls validateECDSAKeySize before 'new ECDsaCng(cngKey)'  otherwise we will leak memory when we throw here.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -607,6 +648,14 @@ public override byte[] Sign(byte[] input)
             throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogMessages.IDX10644));
         }
 
+        private bool validateECDSAKeySize(ECDsa key, string algorithm)
```
brentschmaltz: Method should have a capital 'v'.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
ChenYan98: I agree and modified.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
ChenYan98: Modified

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -648,7 +662,15 @@ public override byte[] Sign(byte[] input)
             throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogMessages.IDX10644));
         }
 
-        private bool validateECDSAKeySize(ECDsa key, string algorithm)
+        private bool ValidateECDSAKeySize(ECDsa key, string algorithm)
```
brentschmaltz: If you passed the key.KeySize, you could remove ValidateCngKeySize.

ChenYan98: Greate idea. Modified.

brentschmaltz:
:shipit:

Issue #509, opened 2016-10-03
Re-enable SignatureProviderTests.AsymmetricSignatureProviders_Verify test for ECDsa384 on non-.NET core


polita:
The test for ECDsaSha384 used to return a CryptographicException on desktop because it wasn't supported. A recent update to 4.6.2 apparently added support because it stopped throwing. I commented out the line in the test that checks for it because it will continue to throw on older .NET installs but not throw on newly-serviced installs of 4.6.2.

We need to figure out how to handle this and re-enable the test.

PR #543, opened 2016-12-06
Update tests to expect exception below 4.6.2 and no exception above 4…

polita:
….6.2 and on Core

#509


brentschmaltz:
:shipit:

PR #544, opened 2016-12-08
Fail build if unit tests fail

polita:
This will keep the build from succeeding when unit tests fail.


brentschmaltz:
nice: :shipit:

PR #560, opened 2017-01-12
Add support for frontchannel_logout_session_supported and frontchanne…

ChenYan98:
…l_logout_supported for 5.x


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
```diff
@@ -161,6 +161,18 @@ public OpenIdConnectConfiguration(string json)
         public string EndSessionEndpoint { get; set; }
 
         /// <summary>
+        /// Gets or sets the frontchannel_logout_session_supported.
```
brentschmaltz: Chen, can you do us a favor and modify the comment for EndSessionEndpoint and add underscores to "end session endpoint" => "end_session_endpoint".  then merge into dev.

PR #574, opened 2017-02-12
fix: Incorrect handling of date/time claims

### LINKED ISSUE #490: Incorrect handling of 'exp', 'nbf' and 'iat' claim values
**System.IdentityModel.Tokens.Jwt: v5.0.0**

I'm building an .NET Core app for an existing large ecosystem.
.NET Core side upon receiving a JWT which works everywhere else (in the ecosystem and decoded by https://jwt.io decoder) fails with the following exception:

```
An unhandled exception occurred while processing the request.

SecurityTokenNoExpirationException: IDX10225: Lifetime validation failed. The token is missing an Expiration Time.
Tokentype: 'System.IdentityModel.Tokens.Jwt.JwtSecurityToken'.
```

however the underlying issue is this:

```
IDX10700: Error found while parsing date time. The 'nbf' claim has value '1471924298.67746' which is could not be parsed to an integer.
```

[The offending code](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L732):

```
secondsAfterBaseTime = Convert.ToInt64(dateValue, CultureInfo.InvariantCulture);
```

The implementation is contrary to the spec: https://tools.ietf.org/html/rfc7519#section-2:

> NumericDate
>       A JSON numeric value representing the number of seconds from
>       1970-01-01T00:00:00Z UTC until the specified UTC date/time,
>       ignoring leap seconds.  This is equivalent to the IEEE Std 1003.1,
>       2013 Edition [POSIX.1] definition "Seconds Since the Epoch", in
>       which each day is accounted for by exactly 86400 seconds, **other
>       than that non-integer values can be represented**.  See RFC 3339
>       [RFC3339] for details regarding date/times in general and UTC in
>       particular.

The correct behavior must process datetime represented by a `double` instead of an `long`.

RussKie:
Fixes #490
The current implementation treats date/time claims (i.e. iat, nbf, exp) as integers. This is contrary to the spec
https://tools.ietf.org/html/rfc7519#section-2 which states that "_non-integer values can be represented_".

Fix converts `iat`, `nbf`, `exp` and `auth_time` claim representations to double


src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -488,7 +488,7 @@ public void AddClaims(IEnumerable<Claim> claims)
             }
         }
 
-        internal static object GetClaimValueUsingValueType(Claim claim)
+        private static object GetClaimValueUsingValueType(Claim claim)
```
RussKie: non of these internals seems to be referenced anywhere, so they are converted to private

test/System.IdentityModel.Tokens.Jwt.Tests/References.cs
```diff
@@ -469,9 +469,9 @@ public static class EncodedJwts
...
-        public static string LiveJwt        =       @"eyJhbGciOiJIUzI1NiIsImtpZCI6IjAiLCJ0eXAiOiJKV1QifQ.eyJ2ZXIiOjEsImlzcyI6InVybjp3aW5kb3dzOmxpdmVpZCIsImV4cCI6MTM2ODY0ODg2MywidWlkIjoiMzgwZTE3YzMxNGU2ZmMyODA0NzA3MjI5NTc3MjEwZmIiLCJhdWQiOiJ3d3cuc3JpLWRldjEwMC5jb20iLCJ1cm46bWljcm9zb2Z0OmFwcHVyaSI6Im1zLWFwcDovL1MtMS0xNS0yLTM2MzczOTQzNzAtMjIzMTgyMTkzNi01NjUwMTU1MS0xNTE0NjEzNDgyLTQ1NjgzNjc4LTM1NzUyNjE4NTItMjMzNTgyNjkwIiwidXJuOm1pY3Jvc29mdDphcHBpZCI6IjAwMDAwMDAwNEMwRTdBNUMifQ.I-sE7t6IJUho1TfgaLilNuzro-pWOMgg33rQ351GcoM";
```
RussKie: removed as I haven't found any references to this one

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -598,33 +598,32 @@ internal string GetStandardClaim(string claimType)
             return null;
         }
 
-        internal int? GetIntClaim(string claimType)
+        private double? GetDoubleClaim(string claimType)
```
brentschmaltz: why change these to private?

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
brentschmaltz: no need to do that

RussKie: I'll revert these.
What's the reason they are marked as internals? The internals aren't used outside the class nor the assembly is decorated with `InternalsVisibleToAttribute`

brentschmaltz: I am concerned this is a breaking change for users. Is there a specific issue that this is causing?

RussKie: As I said above I changed to private since I was unable to find any references to these methods outside JwtPayload class. Coupled with the fact the assembly doesn't share its internals with any other assembly I can't see how it is a breaking change for anyone. 
Also if someone would be accessing this method via reflection calls, `BindingFlags` only have "Public" and "NonPublic" flags, i.e. there is no difference between "internal" and "private" modifiers as far as they go. 

I am happy to be corrected or proven wrong.

RussKie:
I've reinstated the "internals" back

brentschmaltz:
@RussKie I am concerned about back-compat on this

RussKie:
What exactly is the concern?

PinpointTownes:
This is clearly a breaking change (both binary and source), so this should/must wait until the next major version.

brentschmaltz:
@PinpointTownes thanks for chiming in. @RussKie yes this is clearly breaking as users will have runtime  and compile time errors.

In situations such as this, which will cause a breaking change, I like to understand how big is the problem we are solving? What is the impact to customers of not returning a double? Will an attack occur?

RussKie:
@brentschmaltz this aligns MS's implementation with the [spec](https://tools.ietf.org/html/rfc7519#section-2)
For more details please refer to the original bug description https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/490

PinpointTownes:
It's important to note that this bug could be fixed without changing the public API, which would allow fixing it immediately without having to wait until the next major release (and you could even backport it to IM 4.5).

Short-term fix: treat `exp`/`nbf`/`iat`/`auth_time` internally as `double` values, but keep exposing them as `int` (add a tiny check to ensure it doesn't overflow).
Long-term fix: update the public API to expose `long` (to prevent the year 2038 bug) or `decimal`/`double` if you think it's worth supporting decimal values.

RussKie:
What is the best way to go forward?

brentschmaltz:
@RussKie @PinpointTownes the short term fix would work and would be non breaking. We could also, truncate before converting to int.  2038 is a long way away.

brentschmaltz:
@PinpointTownes @RussKie we truncated to int as suggested.

PR #575, opened 2017-02-15
Fix $clean handling

RussKie:
`$clean` like any other input parameter is set to "YES". Presumably the opposite value is expected to be "NO" however the current implementation does not test for "YES" or "NO".


brentschmaltz:
thanks

PR #582, opened 2017-02-28
Upgraded to use netstandard1.6 and make solution compilable and testable in VS2015

dissoupo:
- open Wilson.sln in VS2015
- compile
- open Test Explorer, run all 

You may need to install the latest [.NET Core 1.0.3 SDK](https://www.microsoft.com/net/download/core)

### DIFF
--- a/src/Microsoft.IdentityModel.Logging/project.json
+++ b/src/Microsoft.IdentityModel.Logging/project.json
@@ -3,7 +3,7 @@
   "description": "Includes Event Source based logging support.",
   "dependencies": { },
   "frameworks": {
-    "netstandard1.4": {
+    "netstandard1.6": {
       "dependencies": {
         "System.Diagnostics.Tracing": "4.3.0-*",
         "System.Globalization": "4.3.0-*",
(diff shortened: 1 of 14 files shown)


brentschmaltz:
@dissoupo We can't just upgrade to 1.6, we will need to add an additional target.
When we do add another target, we can fix up some of the crypto issue. So we should take some time and be comprehensive. @PinpointTownes  For example https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/488 will be easy to fix in 1.6.

brentschmaltz:
@dissoupo I created issue: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/583

PinpointTownes:
> @dissoupo We can't just upgrade to 1.6, we will need to add an additional target.

Yes, please :smile:

brentschmaltz:
@PinpointTownes we will try and get this in 5.2.0 milestone, currently tagged for April 8th release.

brentschmaltz:
@dissoupo I am closing this we now have a net452 and netcore 2.0 targets that works with XUnit.

PR #592, opened 2017-03-26
Fix newline formatting

Eilon:
For text to show up on its own line after a list, there has to be a blank line. (Previously the text showed up *within* the list.)


Eilon:
THANK YOU

PR #606, opened 2017-04-24
Update build script for .NetCore/Linux

mbearup:


brentschmaltz:
@mbearup We finally got around to building a linux script for our 5.2.0 release. We are merging it into dev.

PR #608, opened 2017-04-27
add getToken function and test

lovemaths:


brentschmaltz:
looks fine. rebase then push to WsFed.

PR #610, opened 2017-05-04
add tests for WsFedMessage and WsFedRetriever

lovemaths:


src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
```diff
@@ -204,6 +204,9 @@ public virtual string GetToken()
                 }
             }
 
+            if (token == null)
+                throw XmlUtil.LogReadException(LogMessages.IDX10902);
```
brentschmaltz: Throw a WsFederationException.

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
```diff
@@ -110,6 +126,9 @@ protected virtual void ReadEntityDescriptor(WsFederationConfiguration configurat
         /// <param name="reader">xmlreader</param>
         protected virtual void ReadKeyDescriptorForSigning(WsFederationConfiguration configuration, XmlReader reader)
         {
+            if (configuration == null)
+                throw LogHelper.LogArgumentNullException(nameof(configuration));
+
             XmlUtil.CheckReaderOnEntry(reader, WsFederationConstants.Elements.KeyDescriptor, WsFederationConstants.Namespaces.MetadataNamespace);
 
             if (!IsKeyDescriptorForSigning(reader))
```
brentschmaltz: We should get rid of the method IsKeyDescriptorForSigning. We know we are on a KeyDescriptor element.
If the Attribute "use" is missing, log a warning. If the attribute equals anything but 'signing' throw.

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
```diff
@@ -154,20 +176,40 @@ protected virtual void ReadSecurityTokenServiceTypeRoleDescriptor(WsFederationCo
...
             reader.ReadStartElement();
 
+            // Warn if there is no KeyDescriptor element for Signing
+            // Throw if there is no SecurityTokenEndpoint element
+            var hasKeyDescriptorForSigning = false;
+            var hasSecurityTokenEndpoint = false;
+
             while (reader.IsStartElement())
             {
                 if (IsKeyDescriptorForSigning(reader))
```
brentschmaltz: Do the same check you do below.
if (reader.IsStartElement(WsFederationConstants.Elements.KeyDescriptor, ns)

We don't need that methog Is...

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
```diff
@@ -154,20 +176,40 @@ protected virtual void ReadSecurityTokenServiceTypeRoleDescriptor(WsFederationCo
...
                 else
+                {
                     throw XmlUtil.LogReadException(LogMessages.IDX13003, reader.Name);
+                }
             }
 
             // Process </RoleDescriptorr>
             reader.ReadEndElement();
+
+            if (!hasKeyDescriptorForSigning)
```
brentschmaltz: We don't need these flags, we can check the configuration object.
In both cases just log warnings.
I know we discussed throwing if no token endpoint. But let's just log.

PR #612, opened 2017-05-07
Update README.md

theunrepentantgeek:
Don't need to prefix with `#` when using `=====` for a markdown heading.


brentschmaltz:
Thanks!

Issue #585, opened 2017-03-01
ArgumentException with wrong arguments order


joaope:
There are some ArgumentException being constructed with the arguments in the wrong order:

Example:

`throw LogHelper.LogExceptionMessage(new ArgumentException(nameof(algorithm), string.Format(CultureInfo.InvariantCulture, LogMessages.IDX10652, algorithm)));`

Should be:

`throw LogHelper.LogExceptionMessage(new ArgumentException( string.Format(CultureInfo.InvariantCulture, LogMessages.IDX10652, algorithm), nameof(algorithm)));`

This obviously leads to the exception's message and parameter name to be wrong and, consequently, precious debugging time being wasted.

Found 7 occurrences of these. I've fixed them [here](https://github.com/joaope/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/08311d896d4369071c6578c977e2ed273d1a9074) just not sure if I'm free to open a PR with it.

brentschmaltz:
@joaope thanks. Will get this in the April release. Most likely in a nightly before than. We will do a scrub and see what else we can find.

brentschmaltz:
@lovemaths just fix the order of the params so the error is correct

PR #614, opened 2017-05-09
Issue #585 ArgumentException with wrong arguments order

lovemaths:


src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -81,7 +81,7 @@ public HttpDocumentRetriever(HttpClient httpClient)
                 throw LogHelper.LogArgumentNullException("address");
 
             if (!Utility.IsHttps(address) && RequireHttps)
-                throw LogHelper.LogExceptionMessage(new ArgumentException(LogMessages.IDX10108, address));
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogMessages.IDX10108, nameof(address)));
```
brentschmaltz: The idea here was that IDX10108 should have a parameter that has the 'address' value. Include the value for 'address' in the error string.

lovemaths: fixed

brentschmaltz:
:shipit:

PR #620, opened 2017-05-25
Allow base64 padding characters in tokens

george-ayris:
Hello!

I've been testing out the JwtBearer middleware and was having some problems with tokens that contain [base64 padding characters](https://en.wikipedia.org/wiki/Base64#Output_Padding). 

This PR has an example test, which demonstrates the issue in the JWS header, but it also applies to the payload and signature, which have been updated in the `JsonCompactSerializationRegex`. I have not update d the `JweCompactSerializationRegex`, but would guess that this also applies here.

Hope this is useful.


brentschmaltz:
@george-ayris JWT's shouldn't have any base64 padding characters. Did we miss something in the spec?

george-ayris:
@brentschmaltz apologies, second time I searched the specs I found it right away. Sorry to waste your time and thanks for your help.

brentschmaltz:
@george-ayris no worries. Much better to open an issue than to remain silent if you have a question.
Thanks for engaging.

PR #631, opened 2017-06-02
Downgrade to support .Net 4.5 (instead of 4.5.1)

JefWight:
This change simply re-targets the projects to depend on .Net 4.5 instead of 4.5.1.


brentschmaltz:
closing cause we solved this another way

Issue #593, opened 2017-03-29
Support JWE with "zip" in header


brentschmaltz:
If a jwt header indicates that payload is compressed, customers will have to take ownership of  ValidateSignature. It would be cleaner to separate the two.

Should have support for standard algorithms: Deflate, zip, etc.

brentschmaltz:
@mafurman @lovemaths had a hard time getting this into our JwtSecurityTokenHandler, can you make sure this will work with our JsonWebTokenHandler.

PR #643, opened 2017-06-29
Issue #593 Support JWE with "zip" in header

lovemaths:
Support decompression only.


src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
```diff
@@ -0,0 +1,245 @@
...
+                return output.ToArray();
+            }
+        }
+
+        /// <summary>
+        /// Decompress the value using DEFLATE algorithm.
+        /// </summary>
+        /// <param name="value">The bytes to decompress</param>
+        /// <returns>Decompression string</returns>
+        public static string DecompressWithDeflate(byte[] value)
```
brentschmaltz: We need to validate parameters, check for null.
Are there any other possible values that don't make sense?

src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
```diff
@@ -0,0 +1,245 @@
...
+    /// Utility function for compression and decompression.
+    /// </summary>
+    public class CompressionUtils
+    {
+        /// <summary>
+        /// Compress the value using DEFLATE algorithm.
+        /// </summary>
+        /// <param name="value">The string to compress</param>
+        /// <returns>Compression bytes</returns>
+        public static byte[] CompressWithDeflate(string value)
```
brentschmaltz: Need to check parameters on all public and protected methods.

src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
```diff
@@ -0,0 +1,245 @@
...
+        {
+            using (MemoryStream output = new MemoryStream())
```
brentschmaltz: use 'var' where possible.

src/Microsoft.IdentityModel.Tokens/ICompressionProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Custom compression provider interface.
+    /// </summary>
+    public interface ICompressionProvider
```
brentschmaltz: Instead of an Interface can you consider an abstract base class. The reason is an interface breaks when new methods are added. All our other providers are abstract base classes.

src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
lovemaths: Only need to validate null, now validated.

src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
lovemaths: checked.

src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
lovemaths: changed to var

src/Microsoft.IdentityModel.Tokens/ICompressionProvider.cs
lovemaths: changed to abstract base class

brentschmaltz:
We put this into JsonWebToken. No plans to add to JwtSecurityToken

PR #655, opened 2017-07-24
Saml

ChenYan98:


src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSerializer.cs
```diff
@@ -451,10 +522,11 @@ protected virtual SamlAudienceRestrictionCondition ReadAudienceRestrictionCondit
         /// Read the saml:AuthenticationStatement.
         /// </summary>
         /// <param name="reader">XmlReader positioned at a saml:AuthenticationStatement.</param>
+        /// <param name="statements">A <see cref="ICollection{SamlStatement}"/>(s) regarding the subject.</param>
         /// <returns>SamlAuthenticationStatement</returns>
         /// <exception cref="ArgumentNullException">The input parameter 'reader' is null.
         /// or the statement contains a unknown child element.</exception>
-        protected virtual SamlAuthenticationStatement ReadAuthenticationStatement(XmlDictionaryReader reader)
+        protected virtual SamlAuthenticationStatement ReadAuthenticationStatement(XmlDictionaryReader reader, ICollection<SamlStatement> statements)
```
brentschmaltz: Passing in ICollection here is not right. ReadAuthenticationStatement should just read the statement, nothing else.
Consolidating Statements into ClaimsPrincipals and ClaimsIdentities, should not be introduced at this layer.

PR #670, opened 2017-08-09
add transformFactory to token handlers

lovemaths:


brentschmaltz:
@lovemaths you will need to rebase against dev. Sorry for the extra work.

lovemaths:
@brentschmaltz Rebased.

brentschmaltz:
@lovemaths we decided not to do this now, but move classes into the right place so we can eventually use cryptoproviderfactory as the extensibility point.

PR #671, opened 2017-08-10
Fixed the bug in IdentityComparer

mafurman:
localcontext is now properly merged with the original context before the AreEqual() method terminates.


brentschmaltz:
:shipit:

PR #683, opened 2017-08-18
Add CrossVersionTokenValidation test project

ChenYan98:


src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -553,6 +553,8 @@ protected virtual SamlSubject CreateSubject(SecurityTokenDescriptor tokenDescrip
                 //    }
             }
 
+            samlSubject.ConfirmationMethods.Add(SamlConstants.BearerConfirmationMethod);
```
brentschmaltz: please remove the commented code.
thanks.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -1006,9 +1008,9 @@ protected virtual void ValidateConditions(SamlSecurityToken securityToken, Token
                 {
                     SamlAudienceRestrictionCondition audienceRestriction = condition as SamlAudienceRestrictionCondition;
                     if (validationParameters.AudienceValidator != null)
-                        validationParameters.AudienceValidator(audienceRestriction.Audiences, securityToken, validationParameters);
+                        validationParameters.AudienceValidator(audienceRestriction.Audiences.ToDictionary(x => x.ToString()).Keys, securityToken, validationParameters);
```
brentschmaltz: Best to use x.OriginalString() as that leaves the value unchanged.

test/CrossVersionTokenValidation.Tests/SamlCrossVersionTokenValidationTests.cs
```diff
@@ -0,0 +1,789 @@
...
+            {
+                context.Diffs.Add(Environment.NewLine + "obj1 is null");
+                return false;
+            }
+
+            if (objs2 == null)
+            {
+                context.Diffs.Add(Environment.NewLine + "obj2 is null");
+                return false;
+            }
```
brentschmaltz: You could break out these three checks into a separate function. then each method could use them.

buildConfiguration.xml
```diff
@@ -14,6 +14,7 @@
       <project name="Microsoft.IdentityModel.Protocols.OpenIdConnect" />
     </src>
     <test>
+	  <project name="CrossVersionTokenValidation.Tests" test="yes"/>
```
ChenYan98: fixed

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
ChenYan98: fixed

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
ChenYan98: Fixed

test/CrossVersionTokenValidation.Tests/SamlCrossVersionTokenValidationTests.cs
ChenYan98: If I understand right, then it is the same thing like add ContinueCheckingEquality. If it is, then fixed.

brentschmaltz:
:shipit:

PR #692, opened 2017-08-25
Fixed the RsaSecurityKey validation test

### LINKED ISSUE #686: Add a function that compares RsaSecurityKeys to the IdentityComparer class
Currently, RsaSecurityKeys are compared using the generic CompareAllPublicProperties method. However, this causes the 'Parameters' property to be compared improperly (due to the base type being a ValueType), resulting in two equal RsaParameters being considered unequal.

A function that specifically compares RsaSecurityKeys needs to be added to the IdentityComparer._equalityDict in order to resolve this issue.

mafurman:
Fixes #686.

Instead of adding in a function that compares RsaSecurityKeys, I added the AreRSAParametersEqual() function to the _equalityDict. 

I also modified the BaseType == typeof(System.ValueType) condition so that it also checks whether or not the specific type is located in the dictionary. If it is (such as in the case of RsaParameters), we know that we need to call AreEqual() and not treat the property in the same way that we would treat other properties whose BaseType is a ValueType.

I feel like this makes more sense since the function for RsaSecurityKeys would essentially be CompareAllPublicProperties but with a special case for RSAParameters.

### DIFF
--- a/test/Microsoft.IdentityModel.Tests/IdentityComparer.cs
+++ b/test/Microsoft.IdentityModel.Tests/IdentityComparer.cs
@@ -89,6 +89,7 @@ namespace Microsoft.IdentityModel.Tests
                 { typeof(OpenIdConnectMessage).ToString(), CompareAllPublicProperties },
                 { typeof(Reference).ToString(), CompareAllPublicProperties },
                 { typeof(RsaSecurityKey).ToString(), CompareAllPublicProperties },
+                { typeof(RSAParameters).ToString(), AreRsaParametersEqual },
                 { typeof(SamlAction).ToString(), CompareAllPublicProperties },
                 { typeof(SamlAudienceRestrictionCondition).ToString(), CompareAllPublicProperties },
                 { typeof(SamlAssertion).ToString(), CompareAllPublicProperties},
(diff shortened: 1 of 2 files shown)


test/Microsoft.IdentityModel.Tests/IdentityComparer.cs
```diff
@@ -86,6 +86,7 @@ public class IdentityComparer
                 { typeof(OpenIdConnectMessage).ToString(), CompareAllPublicProperties },
                 { typeof(Reference).ToString(), CompareAllPublicProperties },
                 { typeof(RsaSecurityKey).ToString(), CompareAllPublicProperties },
+                { typeof(RSAParameters).ToString(), AreRsaParametersEqual },
```
brentschmaltz: With this here, could we remove: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/rsa/test/Microsoft.IdentityModel.Tests/IdentityComparer.cs#L652

mafurman: I think you're right! This doesn't seem to be necessary anymore; I just removed it.

brentschmaltz: Can you add a test to IdentityComparerTests to cover RSAParameters.
Somewhere in here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/test/Microsoft.IdentityModel.Tokens.Tests/IdentityComparerTests.cs#L370

mafurman: Just added in a separate test for RSAParameters.

test/Microsoft.IdentityModel.Tests/IdentityComparer.cs
```diff
@@ -655,42 +656,60 @@ public static bool AreSecurityKeysEqual(SecurityKey securityKey1, SecurityKey se
...
             var localContext = new CompareContext(context);
             if (!ContinueCheckingEquality(rsaParameters1, rsaParameters2, localContext))
                 return context.Merge(localContext);
 
             if (!AreBytesEqual(rsaParameters1.D, rsaParameters2.D, context))
+            {
+                localContext.Diffs.Add("D:");
                 localContext.Diffs.Add("!AreBytesEqual(rsaParameters1.D, rsaParameters2.D)");
-
+            }
```
brentschmaltz: nit: normally we add a line after a closing brace.

mafurman: Added in the spaces.

brentschmaltz:
:shipit:

PR #694, opened 2017-08-25
Added different telemetry values for Core and Desktop

### LINKED ISSUE #596: Telemetry values should be different for 4.x 5.x AND Core vs. Desktop
Currently they are all the same.

mafurman:
Fixes #596.


brentschmaltz:
@mafurman I am thinking this should have been on AuthenticationProtocolMessage so WsFederationMessage would have support. Think about what we can do without breaking changes.

brentschmaltz:
@mafurman I thought about this some more. I think we stuck with different telemetry code for different protocols: wsFed and OIDC.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
```diff
@@ -39,17 +39,25 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
...
+#elif NETSTANDARD1_4
+            SkuTelemetryValue = "ID_NETSTANDARD1_4";
+#endif
+        }
 
         /// <summary>
         /// Initializes an instance of <see cref="OpenIdConnectMessage"/> class with a json string.
         /// </summary>
-        public OpenIdConnectMessage(string json)
+        public OpenIdConnectMessage(string json): this()
```
brentschmaltz: normally we put :this() on a new line, indented.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
```diff
@@ -535,6 +543,11 @@ public string Sid
         }
 
         /// <summary>
+        /// Gets 'skuTelemetryValue'.
```
brentschmaltz: Comment should read as an English statement. Something like: Gets the string value that is used when sending telemetry data as part of an OpenIdConnectRequest message.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
```diff
@@ -39,17 +39,25 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
     /// </summary>
     public class OpenIdConnectMessage : AuthenticationProtocolMessage
     {
-        private const string _skuTelemetryValue = "ID_NET";
```
brentschmaltz: remove this extra line

mafurman: Done!

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
mafurman: Done!

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
mafurman: Done! Though I'm not entirely sure what the syntax should look like since I can't find an example of assigning a default value to a property along with #if statements.

brentschmaltz:
:shipit:

PR #696, opened 2017-08-26
Stopped unset properties from being serialized and added in conditional property serialization for collections in JsonWebKey

### LINKED ISSUE #579: Do not serialize some properties on JsonWebKey
When trying to serialize an instance of JsonWebKey there are a few properties that are showing up on the wire even though they are not set.

```
{
  "key_ops": [],
  "kid": "DevelopmentKey.5efea89c-6b7f-4ab3-9579-14b1134a4636",
  "kty": "RSA",
  "e": "AQAB",
  "n": "",
  "use": "sig",
  "x5c": [],
  "keySize": 2048,
  "hasPrivateKey": false,
  "keyId": null,
  "cryptoProviderFactory": {
      "customCryptoProvider": null
}
```

It would be good if you could hide these properties by using JsonIgnore http://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonIgnoreAttribute.htm

(KeySize, HasPrivateKey, KeyId, CryptoProviderFactory)

It would also be good if you can add a setter to the collection properties so that they can be set to null and not send over the wire when empty.

This issue is related to https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/578

mafurman:
Fixes #579.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
+++ b/src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
@@ -27,6 +27,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography;
 using Microsoft.IdentityModel.Logging;
@@ -248,6 +249,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// <summary>
         /// Gets the key size of <see cref="JsonWebKey"/>.
         /// </summary>
+        [JsonIgnore]
         public override int KeySize
         {
             get
@@ -267,6 +269,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// Gets a bool indicating if a private key exists.
         /// </summary>
         /// <return>true if it has a private key; otherwise, false.</return>
+        [JsonIgnore]
         public bool HasPrivateKey
         {
             get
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -280,6 +283,24 @@ public bool HasPrivateKey
             }
         }
 
+        /// <summary>
+        /// Gets a bool that determines if the 'key_ops' (Key Operations) property should be serialized.
+        /// </summary>
+        /// <return>true if 'key_ops' (Key Operations) is not empty; otherwise, false.</return>
+        public bool ShouldSerializeKeyOps()
```
brentschmaltz: Where is this used?

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -280,6 +283,24 @@ public bool HasPrivateKey
...
+        {
+            return KeyOps.Any();
+        }
+
+        /// <summary>
+        /// Gets a bool that determines if the 'x5c' collection (X.509 Certificate Chain) property should be serialized.
+        ///</summary>
+        /// <return>true if 'x5c' collection (X.509 Certificate Chain) is not empty; otherwise, false.</return>
+        public bool ShouldSerializeX5c()
```
brentschmaltz: Where is this used?

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -280,6 +283,24 @@ public bool HasPrivateKey
...
+        {
+            return X5c.Any();
```
brentschmaltz: Don't need Linq for this. X5c is an IList. If method is needed, just use X5c.Count > 0

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
mafurman: I believe this is called automatically when serialization occurs. I just followed the instructions here:
https://www.newtonsoft.com/json/help/html/ConditionalProperties.htm

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
mafurman: See the comment above.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
mafurman: Just made this change for both methods.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
brentschmaltz: oh, so does this puts the serialization under user control?

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
brentschmaltz: Can you add a to the comments that JSON.Net uses this.

mafurman: Just added this in for both methods.

brentschmaltz:
:shipit:

PR #700, opened 2017-08-28
Deleted the XmlEncryptionException class

### LINKED ISSUE #699: Remove XmlEncryptionException
The 5.2.0 release will not support Encryption for XML.
So we should just delete this file.

mafurman:
Fixes #699 

Also removed an unnecessary 'using' statement from JsonWebKey.cs.


brentschmaltz:
:shipit:

Issue #690, opened 2017-08-25
Add perl script to check LogMessages


brentschmaltz:
The script should
1. Ensure usage specifies the correct number of parameters
2. LogMessages are being used and call out any unused messages
3. Ensure ID is the prefix of the string

PR #720, opened 2017-09-11
Added in a perl script to check LogMessages

mafurman:
Addresses  #690.

This script checks for:
1) Unused strings in LogMessages.cs
2) Mismatching string IDs and string value prefixes
3) Duplicate IDs across LogMessages.cs files

There is currently some unused code in this script that may need to be modified in order to support checking for incorrect argument count.


brentschmaltz:
:shipit:

PR #721, opened 2017-09-11
Made sure the IDs of each message match up with their string prefixes

### LINKED ISSUE #712: Make sure the ID of each message is the same as the prefix of its string value in LogMessages.cs
Some of the messages used in LogMessages.cs have variable names that are inconsistent with their string values.

For example in Microsoft.IdentityModel.Xml:

`internal const string IDX21403 = "IDX21402: Unable to write XML. One of the values in Reference.Transforms is null or empty.";`

mafurman:
Fixes #712.


brentschmaltz:
:shipit: we will need to add the perl script to our test scripts. We can do that as a separate item.

Issue #691, opened 2017-08-25
Review LogMessages across assemblies to ensure partitioning of error space


brentschmaltz:
In 5.2.0 we have added a number of new assemblies. Each of these has a LogMessage file.
Each of these should have it's own range for error id's.
Some messages are duplicated.

For example: Saml and Saml2 both define: IDX10400.

brentschmaltz:
Once this is done we will be in a position to provide FAQ's and help on error messages.

brentschmaltz:
Partition id's as per xls that we worked on.

PR #730, opened 2017-09-14
Make all the LogMessage error ids unique

mafurman:
Addresses #691.


brentschmaltz:
@mafurman  Can you a comment just below the copyright statement in each LogMessages.cs as per the xls (I added an entry for m.im.tokens.xml).

Something like:

//  System.IdentityModel.Tokens.Jwt 
//  Range: 12000 - 12999

We can then have the perl script check the range.

brentschmaltz:
:shipit:

PR #760, opened 2017-09-25
Remove hardcoded strings from SamlSecurityTokenHandler

ChenYan98:


brentschmaltz:
just put the id in the right range then :shipit:

PR #763, opened 2017-09-27
Added support for RSAKeyValue

### LINKED ISSUE #719: SamlSerializer support for more complex KeyInfo inside Signature
Doing some testing with the new WsFederation stuff (#703). I get the following exception testing against one of my Issuers (that is saml1). It appears the SamlSerailizer doesn't like my more complex KeyInfo element. 

>  IDX11122: Exception thrown while reading 'Assertion' for SamlSecurityToken. Inner exception: 'Microsoft.IdentityModel.Xml.XmlReadException: IDX21025: Unable to read XML. Expecting XmlReader to be at EndElement: 'Signature'. Found XmlNode 'type.name': 'Element.KeyInfo'.

Here is a test with an example token I threw together showing the error:

```csharp
        [Fact]
        public void Read_Token_With_ComplexKeyInfo()
        {
            var token = @"
<saml:Assertion MajorVersion=""1"" MinorVersion=""1"" AssertionID=""O7B1iY1MKGlAbG7n3QUOE_d.nLJ"" IssueInstant=""2017-09-10T14:19:26.835Z"" Issuer=""https://m12test-sso.auto.aws.infor.com/wsfed"" xmlns:saml=""urn:oasis:names:tc:SAML:1.0:assertion"">
<saml:Conditions NotBefore=""2017-09-10T14:14:26.835Z"" NotOnOrAfter=""2017-09-10T22:19:26.835Z"">
    <saml:AudienceRestrictionCondition>
        <saml:Audience>https://ccdev.pcm.infor.com/</saml:Audience>
    </saml:AudienceRestrictionCondition>
</saml:Conditions>
<saml:AuthenticationStatement AuthenticationInstant=""2017-09-10T14:19:26.835Z"" AuthenticationMethod=""urn:oasis:names:tc:SAML:2.0:ac:classes:unspecified"">
    <saml:Subject>
        <saml:NameIdentifier Format=""urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress"">e65fb001-5b83-495a-85c5-a6abfd0c63c8</saml:NameIdentifier>
    </saml:Subject>
</saml:AuthenticationStatement>
<saml:AttributeStatement>
    <saml:Subject><saml:NameIdentifier Format=""urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress"">e65fb001-5b83-495a-85c5-a6abfd0c63c8</saml:NameIdentifier></saml:Subject>
    <saml:Attribute AttributeName=""SecurityRole"" AttributeNamespace=""http://schemas.infor.com/claims/"">
        <saml:AttributeValue>UserAdmin</saml:AttributeValue>
    </saml:Attribute>
    <saml:Attribute AttributeName=""upn"" AttributeNamespace=""http://schemas.xmlsoap.org/ws/2005/05/identity/claims"">
        <saml:AttributeValue>myemail@mail.com</saml:AttributeValue>
    </saml:Attribute>
</saml:AttributeStatement>
<ds:Signature xmlns:ds=""http://www.w3.org/2000/09/xmldsig#"">
    <ds:SignedInfo>
        <ds:CanonicalizationMethod Algorithm=""http://www.w3.org/2001/10/xml-exc-c14n#"" />
        <ds:SignatureMethod Algorithm=""http://www.w3.org/2000/09/xmldsig#rsa-sha1"" />
        <ds:Reference URI=""#O7B1iY1MKGlAbG7n3QUOE_d.nLJ"">
            <ds:Transforms>
                <ds:Transform Algorithm=""http://www.w3.org/2000/09/xmldsig#enveloped-signature"" />
                <ds:Transform Algorithm=""http://www.w3.org/2001/10/xml-exc-c14n#"" />
            </ds:Transforms>
            <ds:DigestMethod Algorithm=""http://www.w3.org/2000/09/xmldsig#sha1"" />
            <ds:DigestValue>G9YRrRupdzdumNqXWyxmw8bFeUE=</ds:DigestValue>
        </ds:Reference>
    </ds:SignedInfo>
    <ds:SignatureValue>
        dummy
    </ds:SignatureValue>
    <ds:KeyInfo>
        <ds:X509Data>
            <ds:X509Certificate>
                dummy
            </ds:X509Certificate>
        </ds:X509Data>
        <ds:KeyValue>
            <ds:RSAKeyValue>
                <ds:Modulus>
                    dummy
                </ds:Modulus>
                <ds:Exponent>AQAB</ds:Exponent>
            </ds:RSAKeyValue>
        </ds:KeyValue>
    </ds:KeyInfo>
</ds:Signature>
</saml:Assertion>
";

            var serializer = new SamlSerializerPublic();

            var xmlReader = XmlUtilities.CreateDictionaryReader(token);
            var result = serializer.ReadAssertionPublic(xmlReader);
        }
```

mafurman:
Fixes #719.

KeyInfo now has an RSAKeyValue property. The ReadKeyInfo() method of DSigSerializer has also been modified to support this change.

### DIFF
--- a/src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
+++ b/src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
@@ -153,13 +153,29 @@ namespace Microsoft.IdentityModel.Xml
                     //else if (reader.IsStartElement(XmlSignatureConstants.Elements.KeyName, XmlSignatureConstants.Namespace))
                     //{
                     //}
-                    // TODO <KeyValue>
-                    //else if (reader.IsStartElement(XmlSignatureConstants.Elements.KeyValue, XmlSignatureConstants.Namespace))
-                    //{
-                    //}
+                    // <KeyValue>
+                    else if (reader.IsStartElement(XmlSignatureConstants.Elements.KeyValue, XmlSignatureConstants.Namespace))
+                    {
+                        reader.ReadStartElement(XmlSignatureConstants.Elements.KeyValue, XmlSignatureConstants.Namespace);
+                            if (reader.IsStartElement(XmlSignatureConstants.Elements.RSAKeyValue, XmlSignatureConstants.Namespace))
+                            {
+                                // Multiple RSAKeyValues were found
+                                if (keyInfo.RSAKeyValue != null)
+                                    throw XmlUtil.LogReadException(LogMessages.IDX30015, XmlSignatureConstants.Elements.RSAKeyValue);
+
+                                ReadRSAKeyValue(reader, keyInfo);
+                            }
+                            else
+                            {
+                                // Skip the element since it is not an <RSAKeyValue>
+                                Logger.WriteWarning(LogMessages.IDX30300, reader.ReadOuterXml());
+                            }
+                        // </KeyValue>
+                        reader.ReadEndElement();
+                    }
                     else
                     {
-                        // Skip the element since it is not one of  <RetrievalMethod>, <X509Data>
+                        // Skip the element since it is not one of  <RetrievalMethod>, <X509Data>, <KeyValue>
                         Logger.WriteWarning(LogMessages.IDX30300, reader.ReadOuterXml());
                     }
                 }
@@ -215,6 +231,29 @@ namespace Microsoft.IdentityModel.Xml
          }
 
         /// <summary>
+        /// Reads the "RSAKeyValue" element conforming to https://www.w3.org/TR/2001/PR-xmldsig-core-20010820/#sec-RSAKeyValue.
+        /// </summary>
+        /// <param name="reader">A <see cref="XmlReader"/> positioned on a <see cref="XmlSignatureConstants.Elements.RSAKeyValue"/> element.</param>
+        /// <param name="keyInfo">The <see cref="KeyInfo"/> to hold the RSAKeyValue.</param>
+        private void ReadRSAKeyValue(XmlReader reader, KeyInfo keyInfo)
+        {
+            reader.ReadStartElement(XmlSignatureConstants.Elements.RSAKeyValue, XmlSignatureConstants.Namespace);
+
+            if (!reader.IsStartElement(XmlSignatureConstants.Elements.Modulus, XmlSignatureConstants.Namespace))
+                throw XmlUtil.LogReadException(LogMessages.IDX30011, XmlSignatureConstants.Elements.Modulus, reader.LocalName);
+
+            string modulus = reader.ReadElementContentAsString(XmlSignatureConstants.Elements.Modulus, XmlSignatureConstants.Namespace);
+
+            if (!reader.IsStartElement(XmlSignatureConstants.Elements.Exponent, XmlSignatureConstants.Namespace))
+                throw XmlUtil.LogReadException(LogMessages.IDX30011, XmlSignatureConstants.Elements.Exponent, reader.LocalName);
+
+            string exponent = reader.ReadElementContentAsString(XmlSignatureConstants.Elements.Exponent, XmlSignatureConstants.Namespace);
+
+            keyInfo.RSAKeyValue = new RSAKeyValue(modulus, exponent);
+
+            reader.ReadEndElement();
+        }
+        /// <summary>
         /// Reads the "X509SKI" element conforming to https://www.w3.org/TR/2001/PR-xmldsig-core-20010820/#sec-X509Data.
         /// </summary>
         /// <param name="reader">The <see cref="XmlReader"/> currently pointing at the <see cref="XmlSignatureConstants.Elements.X509SKI"/> element.</param>
@@ -258,7 +297,7 @@ namespace Microsoft.IdentityModel.Xml
                 reader.MoveToContent();
                 var signatureValue = reader.ReadElementContentAsString().Trim();
                 KeyInfo keyInfo = null;
-                if (reader.IsStartElement(XmlSignatureConstants.Elements.KeyInfo))
+                if (reader.IsStartElement(XmlSignatureConstants.Elements.KeyInfo, XmlSignatureConstants.Namespace))
                     keyInfo = ReadKeyInfo(reader);
 
                 // </Signature>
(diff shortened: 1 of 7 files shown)


src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -212,6 +228,29 @@ private void ReadIssuerSerial(XmlReader reader, KeyInfo keyInfo)
          }
 
         /// <summary>
+        /// Reads the "RSAKeyValue" element conforming to https://www.w3.org/TR/2001/PR-xmldsig-core-20010820/#sec-RSAKeyValue.
+        /// </summary>
+        /// <param name="reader">A <see cref="XmlReader"/> positioned on a <see cref="XmlSignatureConstants.Elements.RSAKeyValue"/> element.</param>
+        /// <param name="keyInfo">The <see cref="KeyInfo"/> to hold the RSAKeyValue.</param>
+        private void ReadRSAKeyValue(XmlReader reader, KeyInfo keyInfo)
```
brentschmaltz: It would be better to return a RSAKeyKeyValue

src/Microsoft.IdentityModel.Xml/KeyInfo.cs
```diff
@@ -106,6 +111,15 @@ public string RetrievalMethodUri
         }
 
         /// <summary>
+        /// Gets or sets the RSAValue.
+        /// </summary>
+        public RSAKeyValue RSAValue
```
brentschmaltz: Please change name to RSAKeyValue

src/Microsoft.IdentityModel.Xml/KeyInfo.cs
```diff
@@ -66,6 +67,10 @@ public KeyInfo(SecurityKey key)
                 CertificateData = Convert.ToBase64String(x509Key.Certificate.RawData);
                 Kid = x509Key.Certificate.Thumbprint;
             }
+            else if (key is RsaSecurityKey rsaKey)
```
brentschmaltz: There is a possibility that rsaKey.Parameters is null, when the constructor that takes an RSA as a parameter is used.
In this case you will have to use rsaKey.RSA.ExportParameters(false) to obtain the parameters.

src/Microsoft.IdentityModel.Xml/RSAKeyValue.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    /// <summary>
+    /// The RSAKeyValue found inside of the KeyValue element.
+    /// </summary>
+    public class RSAKeyValue
+    {
+        /// <summary>
+        /// The modulus of the RSAKeyValue.
+        /// </summary>
+        public byte[] modulus;
```
brentschmaltz: These should be readonly properties.
Use this pattern: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-6#read-only-auto-properties

test/Microsoft.IdentityModel.Tests/TestSets.cs
```diff
@@ -415,6 +415,36 @@ public static KeyInfoTestSet SingleSubjectName
...
+                                </X509Data>
+                                <KeyValue>
+                                    <RSAKeyValue>
+                                            <Modulus>rCz8Sn3GGXmikH2MdTeGY1D711EORX/lVXpr+ecGgqfUWF8MPB07XkYuJ54DAuYT318+2XrzMjOtqkT94VkXmxv6dFGhG8YZ8vNMPd4tdj9c0lpvWQdqXtL1TlFRpD/P6UMEigfN0c9oWDg9U7Ilymgei0UXtf1gtcQbc5sSQU0S4vr9YJp2gLFIGK11Iqg4XSGdcI0QWLLkkC6cBukhVnd6BCYbLjTYy3fNs4DzNdemJlxGl8sLexFytBF6YApvSdus3nFXaMCtBGx16HzkK9ne3lobAwL2o79bP4imEGqg+ibvyNmbrwFGnQrBc1jTF9LyQX9q+louxVfHs6ZiVw==</Modulus>
```
brentschmaltz: tabbed once to many

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
mafurman: This is currently not the way that the other read methods in DSigSerializer work.
I created a separate issue to address this problem:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/766

src/Microsoft.IdentityModel.Xml/KeyInfo.cs
mafurman: Done!

RSAParameters is a struct, so it can't actually be null. However, I think that checking whether it equals the default value accomplishes the same purpose.

src/Microsoft.IdentityModel.Xml/KeyInfo.cs
mafurman: Done!

src/Microsoft.IdentityModel.Xml/RSAKeyValue.cs
mafurman: Done!

test/Microsoft.IdentityModel.Tests/TestSets.cs
mafurman: Fixed!

src/Microsoft.IdentityModel.Xml/RSAKeyValue.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    /// <summary>
+    /// The RSAKeyValue found inside of the KeyValue element.
+    /// </summary>
+    public class RSAKeyValue
+    {
+        /// <summary>
+        /// The modulus of the RSAKeyValue.
+        /// </summary>
+        public byte[] Modulus { get; }
```
brentschmaltz: JsonWebKey.N property is a string. Consider having these properties as a string, then we don't have to convert when reading or writing.

mafurman: Just made this change! I think storing these properties as strings rather than byte arrays will make the reading and writing process cleaner.

brentschmaltz:
:shipit:

PR #764, opened 2017-09-27
Add checking for DecisionType

ChenYan98:


src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlAuthorizationDecisionStatement.cs
```diff
@@ -36,6 +37,10 @@ namespace Microsoft.IdentityModel.Tokens.Saml
     /// </summary>
     public class SamlAuthorizationDecisionStatement : SamlSubjectStatement
     {
+        private const string Permit = "Permit";
```
brentschmaltz: Remove these and add a test.

brentschmaltz:
:shipit:

PR #765, opened 2017-09-27
Add missing Newtonsoft.Json dependency

ChenYan98:


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Microsoft.IdentityModel.Protocols.OpenIdConnect.csproj
```diff
@@ -15,6 +15,7 @@
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.IdentityModel.Protocols\Microsoft.IdentityModel.Protocols.csproj" />
     <ProjectReference Include="..\System.IdentityModel.Tokens.Jwt\System.IdentityModel.Tokens.Jwt.csproj" />
+    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
```
brentschmaltz: We should bump this up one level.
Inside <root>\build\dependencies.props there is a parameter <JsonNetVersion> use that for defining the Newtonsoft version in M.IM.Protocols.OpenIdConnect.csproj, M.IM.Tests.csproj, M.IM.Tokens.csproj, M.IM.Tokens.Tests.

ChenYan98:
Switch JsonNetVersion from 10.0.1 to 9.0.1. By using version 10, this library will get exception when call JsonExtensions.DeserializeJwtPayload() which use JsonConvert.DeserializeObject by default.

brentschmaltz:
:shipit:  opened https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/776

Issue #772, opened 2017-09-28
DSigSerializer will throw on Empty KeyInfo element


brentschmaltz:
If the KeyInfo xml is empty, the Serializer will try and read the empty element.

brentschmaltz:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/773

PR #773, opened 2017-09-28
ensure we can read keyinfo that we write

brentschmaltz:
fix issue with empty element

Fixes: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/772


mafurman:
:shipit:

Issue #768, opened 2017-09-28
Add support to DSigSerializer to write RSAKeyValue


brentschmaltz:
DSigSerializer can read a RSAKeyValue and needs to write an RSAKeyValue.
Add support in DSigSerializer.WriteKeyInfo

Issue #766, opened 2017-09-27
The DSigSerializer read methods are not consistent with those of WsFederationMetadataSerializer


mafurman:
Currently, the DSigSerializer read methods (ex. ReadIssuerSerial()) return void, but those of WsFederationMetadataSerializer all return the type that was read. The methods in DSigSerializer should be changed to behave in a similar way.

brentschmaltz:
We should fix this for 5.2.0 release

PR #783, opened 2017-10-04
Modified the KeyInfo class so that it better represents the structure of the DSig KeyInfo element

### LINKED ISSUE #777: KeyInfo class should be modified so it can represent additional elements available in DSig.
Currently the KeyInfo class is unstructured and could represent the structure of the DSig KeyInfo element.

Specifically:
1. Create a X509Data class
2. Create an IssuerSerial class with natural properties
3. KeyInfo should expose a collection of X509Data.
4. X509Data should have properties: IssuerSerial, SubjectName, Collection of X509Certificate, SKI, CRL (all strings),
5. Add KeyName to KeyInfo - we can use this when looking for kid's.

RSAKeyValue is a FirstClass property.

For now we will Ignore: PGPData, MgmtData,

mafurman:
Fixes #777. 

Also addresses #768.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSubject.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSubject.cs
@@ -246,13 +246,7 @@ namespace Microsoft.IdentityModel.Tokens.Saml
                 return false;
 
             if (subject1.KeyInfo != null && subject2.KeyInfo != null)
-                if (string.Compare(subject1.KeyInfo.CertificateData, subject2.KeyInfo.CertificateData, StringComparison.OrdinalIgnoreCase) != 0 ||
-                string.Compare(subject1.KeyInfo.IssuerName, subject2.KeyInfo.IssuerName, StringComparison.OrdinalIgnoreCase) != 0 ||
-                string.Compare(subject1.KeyInfo.Kid, subject2.KeyInfo.Kid, StringComparison.OrdinalIgnoreCase) != 0 ||
-                string.Compare(subject1.KeyInfo.RetrievalMethodUri, subject2.KeyInfo.RetrievalMethodUri, StringComparison.OrdinalIgnoreCase) != 0 ||
-                string.Compare(subject1.KeyInfo.SerialNumber, subject2.KeyInfo.SerialNumber, StringComparison.OrdinalIgnoreCase) != 0 ||
-                string.Compare(subject1.KeyInfo.SKI, subject2.KeyInfo.SKI, StringComparison.OrdinalIgnoreCase) != 0 ||
-                string.Compare(subject1.KeyInfo.SubjectName, subject2.KeyInfo.SubjectName, StringComparison.OrdinalIgnoreCase) != 0)
+                if (!subject1.KeyInfo.Equals(subject2.KeyInfo))
                     return false;
                 else if (subject1.KeyInfo == null | subject2.KeyInfo == null)
                     return false;
@@ -284,20 +278,10 @@ namespace Microsoft.IdentityModel.Tokens.Saml
             hashCode ^= (subject.ConfirmationData == null) ? defaultHash : subject.ConfirmationData.GetHashCode();
 
             if (subject.KeyInfo != null)
-            {
-                hashCode ^= (subject.KeyInfo.CertificateData == null) ? defaultHash : subject.KeyInfo.CertificateData.GetHashCode();
-                hashCode ^= (subject.KeyInfo.IssuerName == null) ? defaultHash : subject.KeyInfo.IssuerName.GetHashCode();
-                hashCode ^= (subject.KeyInfo.Kid == null) ? defaultHash : subject.KeyInfo.Kid.GetHashCode();
-                hashCode ^= (subject.KeyInfo.RetrievalMethodUri == null) ? defaultHash : subject.KeyInfo.RetrievalMethodUri.GetHashCode();
-                hashCode ^= (subject.KeyInfo.SerialNumber == null) ? defaultHash : subject.KeyInfo.SerialNumber.GetHashCode();
-                hashCode ^= (subject.KeyInfo.SKI == null) ? defaultHash : subject.KeyInfo.SKI.GetHashCode();
-                hashCode ^= (subject.KeyInfo.SubjectName == null) ? defaultHash : subject.KeyInfo.SubjectName.GetHashCode();
-            }
+                hashCode ^= subject.KeyInfo.GetHashCode();
 
             foreach (var method in subject.ConfirmationMethods)
-            {
                 hashCode ^= method.GetHashCode();
-            }
 
             return hashCode.GetHashCode();
         }
--- a/src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
+++ b/src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
@@ -28,7 +28,6 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Security.Cryptography.X509Certificates;
 using System.Text;
 using System.Xml;
 using Microsoft.IdentityModel.Tokens;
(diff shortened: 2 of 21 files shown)


src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSubject.cs
```diff
@@ -285,15 +279,8 @@ public override int GetHashCode(SamlSubject subject)
 
             if (subject.KeyInfo != null)
             {
```
brentschmaltz: You can remove the '{ }' if you want.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -140,6 +145,9 @@ public virtual KeyInfo ReadKeyInfo(XmlReader reader)
                             }
                         }
 
+                        var X509Data = new X509Data(issuerSerial, subjectName, SKI, certificateData, CRL);
+                        keyInfo.X509Data.Add(X509Data);
+
                         // </X509Data>
                         reader.ReadEndElement();
```
brentschmaltz: We need to check if X509Data is an empty element. If so we should throw as that indicates an error.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -101,37 +101,42 @@ public virtual KeyInfo ReadKeyInfo(XmlReader reader)
                     // <X509Data>
                     if (reader.IsStartElement(XmlSignatureConstants.Elements.X509Data, XmlSignatureConstants.Namespace))
```
brentschmaltz: Should factor into a method that returns an X509Data, it can be private such as ReadIssuerSerial.

src/Microsoft.IdentityModel.Xml/X509Data.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        /// <summary>
+        /// Initializes an instance of <see cref="X509Data"/>.
+        /// </summary>
+        /// <param name="issuerSerial">the <see cref="IssuerSerial"/>to populate the X509Data.</param>
+        /// <param name="subjectName">the subject name of the X509Data.</param>
+        /// <param name="SKI">the SKI of the X509Data.</param>
+        /// <param name="certificates">the list of X509Certificates belonging to the X509Data.</param>
+        /// <param name="CRL">the CRL belonging to the X509Data.</param>
+        public X509Data(IssuerSerial issuerSerial, string subjectName, string SKI, List<string> certificates, string CRL)
```
brentschmaltz: I would suggest just one the constructor: X509Data(X509Certificate2 certificate).
The other properties can be set using Properties.

Make sure to check for null.

src/Microsoft.IdentityModel.Xml/X509Data.cs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        /// <summary>
+        /// Get or sets the 'X509Certificate'(s) value that is a part of 'X509Data'.
+        /// </summary>
+        public List<string> Certificates
+        {
+          get;
+          set; 
+        } = new List<string>();
```
brentschmaltz: Return an ICollection<string>.
Change comment.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -101,37 +101,42 @@ public virtual KeyInfo ReadKeyInfo(XmlReader reader)
...
                     {
+                        var certificateData = new List<string>();
```
brentschmaltz: if you create an X509Data you don't need these local variables. You can assign as you go.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -101,37 +101,42 @@ public virtual KeyInfo ReadKeyInfo(XmlReader reader)
...
                         while (reader.IsStartElement())
                         {
                             if (reader.IsStartElement(XmlSignatureConstants.Elements.X509Certificate, XmlSignatureConstants.Namespace))
                             {
-                                // multiple certs
-                                if (keyInfo.CertificateData != null)
-                                    throw XmlUtil.LogReadException(LogMessages.IDX30015, XmlSignatureConstants.Elements.X509Certificate);
-
-                                ReadCertificate(reader, keyInfo);
+                                certificateData.Add(reader.ReadElementContentAsString());
```
brentschmaltz: Leave a space after an if clause.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
mafurman: Done!

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
mafurman: Done!

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
mafurman: Done!

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
mafurman: Done!

src/Microsoft.IdentityModel.Xml/X509Data.cs
mafurman: Done! I added multiple constructors.

src/Microsoft.IdentityModel.Xml/X509Data.cs
mafurman: Done!

mafurman:
Also addresses  #766.

src/Microsoft.IdentityModel.Xml/X509Data.cs
```diff
@@ -0,0 +1,150 @@
...
+        public string CRL
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// Checks if an 'X509Data' object is empty.
+        /// </summary>
+        public bool IsEmpty()
```
brentschmaltz: I don't think we need this. When reading, throw if the X509Data is an EmptyElement

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -101,58 +100,22 @@ public virtual KeyInfo ReadKeyInfo(XmlReader reader)
...
-                    //else if (reader.IsStartElement(XmlSignatureConstants.Elements.KeyName, XmlSignatureConstants.Namespace))
-                    //{
-                    //}
+
+                    // <KeyName>
+                    else if (reader.IsStartElement(XmlSignatureConstants.Elements.KeyName, XmlSignatureConstants.Namespace))
+                    {
+                        keyInfo.KeyName = reader.ReadElementContentAsString(XmlSignatureConstants.Elements.KeyName, XmlSignatureConstants.Namespace);
+                    }
+
```
brentschmaltz: We don't need a new line after the else if, only at the end of a clause.
Sorry if that wasn't clear.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -196,46 +161,93 @@ public virtual KeyInfo ReadKeyInfo(XmlReader reader)
...
+                else
+                {
+                    // Skip the element since it is not one of  <X509Certificate>, <X509IssuerSerial>, <X509SKI>, <X509SubjectName>, <X509CRL>
+                    Logger.WriteWarning(LogMessages.IDX30300, reader.ReadOuterXml());
+                }
+            }
+
+            // </X509Data>
+            reader.ReadEndElement();
```
brentschmaltz: If we were on an empty element, this will throw.
In that case, we should throw and then we don't need the check below.

src/Microsoft.IdentityModel.Xml/X509Data.cs
```diff
@@ -0,0 +1,150 @@
...
+        public X509Data()
+        {
+        }
+
+        /// <summary>
+        /// Initializes an instance of <see cref="X509Data"/>.
+        /// </summary>
+        public X509Data(X509Certificate2 certificate)
+        {
+            if (certificate != null)
```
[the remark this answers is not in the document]
mafurman: Done!

src/Microsoft.IdentityModel.Xml/X509Data.cs
```diff
@@ -0,0 +1,150 @@
...
+                Certificates.Add(Convert.ToBase64String(certificate.RawData));
+        }
+
+
+        /// <summary>
+        /// Initializes an instance of <see cref="X509Data"/>.
+        /// </summary>
+        public X509Data(IEnumerable<X509Certificate2> certificates)
+        {
+            if (certificates != null)
```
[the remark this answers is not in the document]
mafurman: Done!

src/Microsoft.IdentityModel.Xml/X509Data.cs
mafurman: That makes sense!

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
mafurman: Done!

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
mafurman: Done!

brentschmaltz:
fix the nameof, squash, rebase and :shipit:

src/Microsoft.IdentityModel.Xml/KeyInfo.cs
```diff
@@ -127,30 +110,38 @@ public RSAKeyValue RSAKeyValue
...
-        /// Get or sets the SerialNumber that is part of a 'X509IssuerSerial'
+        /// Get or sets the 'X509Data' value.
         /// </summary>
-        public string SerialNumber
+        public List<X509Data> X509Data
         {
             get;
             set;
-        }
+        } = new List<X509Data>();
```
brentschmaltz: needs to be ICollection<X509Data>

mafurman: Fixed!

PR #789, opened 2017-10-12
Regex should use CultureInvariant. Fix for https://github.com/AzureAD/azure…

brentschmaltz:
…-activedirectory-identitymodel-extensions-for-dotnet/issues/788


brentschmaltz:
@phkelley can you review?

phkelley:
@brentschmaltz This looks good, but there are two more calls to `Regex.IsMatch` in the method `JwtSecurityTokenHandler.CanReadToken` that we should cover as well.

phkelley:
One other thing - you were asking about how to see the performance difference between `Regex.IsMatch` the static method, and calling `IsMatch` on an instance of the `Regex` class. The easiest way to do this is to set `Regex.CacheSize` to 0 in your test environment.

brentschmaltz:
@phkelley thanks, will add the fix to CanReadtoken.

Issue #787, opened 2017-10-11
Add tests to make sure empty elements are handled correctly in WsFederation Serializer


brentschmaltz:
The WsFed serializer in a number of locations assumes that the XML is not an empty element such as in here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs#L129

This would fail if element is empty.

brentschmaltz:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/79be3ea38fc5f5e18c4f4e4a90857162f456ebb8

PR #792, opened 2017-10-12
Issue #787 Add tests for empty elements

lovemaths:


test/Microsoft.IdentityModel.Tests/ReferenceMetadata.cs
```diff
@@ -725,6 +737,25 @@ public static string AADCommonMetadataFormated
             }
         }
 
+        public static string AADCommonMetadataWithEmptyElement
+        {
+            get
+            {
+                return
+                @"<EntityDescriptor xmlns=""urn:oasis:names:tc:SAML:2.0:metadata"" ID=""_6c4f3672-45c2-47a6-9515-afda95224009"" entityID=""https://sts.windows.net/268da1a1-9db4-48b9-b1fe-683250ba90cc/"">
```
brentschmaltz: What I actually meant was that the %lt;EntityDescriptor> was empty.
I think this would fail if that was the case: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs#L135
Similar elements would also fail.

PR #801, opened 2017-10-18
Exception is no longer thrown when a large 'exp' value is encountered

### LINKED ISSUE #795: Large 'exp' values are not handled.
Customer reported:

Normally the “exp” claim in a valid token looks like this “1492151232” so I would agree that “1507680819080” is a little ridiculously oversized. Our service is getting thousands of these and correctly returning 401s.

This should be handled gracefully.

mafurman:
Fixes #795.


brentschmaltz:
:shipit:

PR #803, opened 2017-10-19
Made M.IM.Tokens.LogMessages internal

### LINKED ISSUE #771: Microsoft.IdentityModel.Tokens.LogMessages should not be public
This causes a problem for other libraries that have the same logging messaging structure.
In 5.1.4 it was internal, we should change it back.

see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/Microsoft.IdentityModel.Tokens/LogMessages.cs#L33

mafurman:
Fixes #771.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
@@ -33,7 +33,7 @@ namespace Microsoft.IdentityModel.Tokens
     /// <summary>
     /// Log messages and codes
     /// </summary>
-    public static class LogMessages
+    internal static class LogMessages
     {
 #pragma warning disable 1591
         // general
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -28,12 +28,26 @@
 // Microsoft.IdentityModel.Tokens
 // Range: 10000 - 10999
 
+using System.Runtime.CompilerServices;
```
brentschmaltz: @mafurman while there is no harm putting this here, we usually put it in AssemblyInfo.cs.
Can you move it there?

see: https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Clients.ActiveDirectory/Properties/AssemblyInfo.cs#L53

mafurman: Just made the change!

brentschmaltz:
:shipit:

PR #806, opened 2017-10-26
Added in empty element checks and tests for SamlSerializer, Saml2Serializer, and DSigSerializer

### LINKED ISSUE #785: Review xml reading to ensure Empty elements are handled correctly
There is a pattern in the code where a check is made for an expected element. 
Many places in the code are written as: reader.ReadEndElement(), which will throw. We need to make sure those errors do not bubble up to the application layer.

A set of tests is needed.

mafurman:
Fixes #785.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthorizationDecisionStatement.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthorizationDecisionStatement.cs
@@ -129,7 +129,7 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
                 if (value == null)
                     throw LogArgumentNullException(nameof(value));
 
-                if (value.IsAbsoluteUri)
+                if (!value.IsAbsoluteUri)
                     throw LogExceptionMessage(new ArgumentException(FormatInvariant(LogMessages.IDX13300, nameof(value), value)));
 
                 _resource = value;
--- a/test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlSerializerTests.cs
+++ b/test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlSerializerTests.cs
@@ -106,6 +106,13 @@ namespace Microsoft.IdentityModel.Tokens.Saml.Tests
                         ActionTestSet = ReferenceSaml.SamlActionValid,
                         SamlSerializer = new SamlSerializerPublic(),
                         TestId = nameof(ReferenceSaml.SamlActionValid)
+                    },
+                    new SamlTheoryData
+                    {
+                        ExpectedException = new ExpectedException(typeof(SamlSecurityTokenReadException), "IDX11137:"),
+                        ActionTestSet = ReferenceSaml.SamlActionEmpty,
+                        SamlSerializer = new SamlSerializerPublic(),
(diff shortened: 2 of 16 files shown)


test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectMessageTests.cs
```diff
@@ -506,7 +506,7 @@ public void OidcCreateLogoutRequestUrl(string testId, OpenIdConnectMessage messa
 #if NET452
             if(!message.SkuTelemetryValue.Equals("ID_NET451"))
                 context.Diffs.Add($"{message.SkuTelemetryValue} != ID_NET451");
-#elif NETCOREAPP2_0
+#elif NETCOREAPP2_0
```
brentschmaltz: just revert this file

test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlSerializerTests.cs
```diff
@@ -106,6 +106,13 @@ public static TheoryData<SamlTheoryData> ReadActionTheoryData
                         ActionTestSet = ReferenceSaml.SamlActionValid,
                         SamlSerializer = new SamlSerializerPublic(),
                         TestId = nameof(ReferenceSaml.SamlActionValid)
+                    },
+                    new SamlTheoryData
```
brentschmaltz: Add XML element as property then we can get rid of testsets.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectMessageTests.cs
mafurman: Done!

test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlSerializerTests.cs
mafurman: I did this for Saml2SerializerTests, but it's somewhat more complicated to do this for SamlSerializerTests because a lot of the test sets rely on information created in other test sets (ex.  a SamlAssertionTestSet relies on Conditions created in a SamlConditionTestSet). Would it make sense to extract the information from the test sets and store them as separate properties in ReferenceSaml.cs?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthorizationDecisionStatement.cs
```diff
@@ -129,7 +129,7 @@ public Uri Resource
                 if (value == null)
                     throw LogArgumentNullException(nameof(value));
 
-                if (value.IsAbsoluteUri)
+                if (!value.IsAbsoluteUri)
```
brentschmaltz: nice catch.

brentschmaltz:
:shipit: - rebase etc...

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthorizationDecisionStatement.cs
mafurman: Thanks! :blush:

PR #809, opened 2017-10-28
Added delegate functionality for ReadToken() into JwtSecurityTokenHan…

### LINKED ISSUE #798: Add delegate for ReadToken similar to SignatureValidator
Customers that have legacy code or who want to apply different configuration for tokens could use this delegate to read the token once, obtain the correct policy and proceed with validation.

mafurman:
Fixes #798.


src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -96,6 +96,12 @@ namespace Microsoft.IdentityModel.Tokens
     public delegate SecurityToken SignatureValidator(string token, TokenValidationParameters validationParameters);
 
     /// <summary>
+    /// Definition for TokenReader.
+    /// </summary>
+    /// <param name="token">A securityToken with a signature.</param>
+    public delegate SecurityToken TokenReader(string token);
```
brentschmaltz: Even though it is not needed, we always pass TokenValidationParameters to facilitate extensibility.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -96,6 +96,12 @@ namespace Microsoft.IdentityModel.Tokens
...
+
+    /// <summary>
```
brentschmaltz: In the ctor: TokenValidationParameters(TokenValidationParameters other), we need to set TokenReader to other.TokenReader

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -931,7 +931,20 @@ protected virtual JwtSecurityToken ValidateSignature(string token, TokenValidati
...
+            if (validationParameters.TokenReader != null)
+            {
+                var securityToken = validationParameters.TokenReader(token);
+                if (securityToken == null)
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10507, token)));
+
+                jwtToken = securityToken as JwtSecurityToken;
+                if (jwtToken == null)
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10508, typeof(JwtSecurityToken), securityToken.GetType(), token)));
+            } else
```
brentschmaltz: In this case we should put the else on a separate line and then use { }. Even though the else has only one clause.

test/Microsoft.IdentityModel.Tests/ValidationDelegates.cs
```diff
@@ -29,6 +29,8 @@
 using System.Collections.Generic;
 using System.IdentityModel.Tokens.Jwt;
 using Microsoft.IdentityModel.Tokens;
+using Microsoft.IdentityModel.Tokens.Saml;
+using Microsoft.IdentityModel.Tokens.Saml2;
```
brentschmaltz: we don't need M.IM.Tokens or M.IM.Tokens.Saml usings.
Another way to return a different token, is to define a CustomSecurityToken like in: CustomTypes.cs. If we use CustomSecurityToken, then we can remove the M.IM.Tokens.Saml2 using.

In fact, this CustomSecurityToken should be moved to DerivedTypes.cs.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
mafurman: Got it!

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
mafurman: Sorry for missing this. Just fixed it.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
mafurman: Fixed it! I'll try to be more careful next time!

test/Microsoft.IdentityModel.Tests/ValidationDelegates.cs
mafurman: Moved the CustomSecurityToken into DerivedTypes.cs and removed the unnecessary using statements.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -948,7 +948,22 @@ protected virtual JwtSecurityToken ValidateSignature(string token, TokenValidati
...
+            if (validationParameters.TokenReader != null)
+            {
+                var securityToken = validationParameters.TokenReader(token, validationParameters);
+                if (securityToken == null)
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10510, token)));
+
+                jwtToken = securityToken as JwtSecurityToken;
+                if (jwtToken == null)
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10509, typeof(JwtSecurityToken), securityToken.GetType(), token)));
+            } else
```
brentschmaltz: please put else on separate line.

brentschmaltz:
one small comment then :shipit:

PR #811, opened 2017-11-01
Add full debug info for FxCop

ChenYan98:


src/Microsoft.IdentityModel.Logging/Microsoft.IdentityModel.Logging.csproj
```diff
@@ -12,6 +12,21 @@
     <PackageTags>.NET;Windows;Authentication;Identity;Logging</PackageTags>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net45|AnyCPU'">
```
brentschmaltz: Given that net45, net451 and netstandard1_4 all set the same values. Could we have one PropertyGroup based off of 'Configuration == Debug' ?

PR #813, opened 2017-11-01
Solve crypto issue

lovemaths:


brentschmaltz:
We used ideas from here when creating the 461 and core 2.0 targets.

PR #814, opened 2017-11-03
Added in code that makes sure a JWT's signing certificate is currently valid and has not expired yet

### LINKED ISSUE #805: JWT's signing certificate's expiry date is not checked once loaded (Version 5.1.4).
# To reproduce
* Create public/private certificate pair.
* Install the public cert.
* Configure your settings like so:
```c#
public static void ConfigureAuth(IServiceCollection services, IList<X509SecurityKey> certs)
{
    services
        .AddAuthorization(opt => ....
        .AddAuthentication(opt => ....
        .AddJwtBearer(opt =>
        {
            opt.RequireHttpsMetadata = true;
            opt.IncludeErrorDetails = true;

            opt.TokenValidationParameters.RequireExpirationTime = true;
            opt.TokenValidationParameters.RequireSignedTokens = true;
            opt.TokenValidationParameters.ValidateAudience = false;
            opt.TokenValidationParameters.ValidateActor = false;
            opt.TokenValidationParameters.ValidateIssuerSigningKey = true;
            opt.TokenValidationParameters.ValidateLifetime = true;
            opt.TokenValidationParameters.IssuerSigningKeys = certs;
            opt.TokenValidationParameters.ValidateIssuer = false;
        });
}
```
* Create a JWT that's signed with the private key.
* Throw it at the web-site in the HTTP `authorization` header - all is good
* Change the computer's dated to something *past* the public cert's expiry date (without restarting the process).
* Throw the same JWT at the web-site again - all is.....  Still good?

Note that the .NET framework will correctly flag the same public certificate as "invalid" once it's expired.  i.e. the following code will *not* load an expired cert:

```c#
var certs = store.Certificates.Find(X509FindType.FindByThumbprint, thumbprint, true);
```

which was what I was expecting.  What I *wasn't* expecting was that once loaded and handed to the framework, that it would continue to be used once it had expired.

There's a work around, using something like:

```c#
...
opt.TokenValidationParameters.IssuerSigningKeyValidator = KeyValidator;
...

private static bool KeyValidator(SecurityKey securityKey, SecurityToken securityToken, TokenValidationParameters validationParameters)
{
    if (securityKey is X509SecurityKey key)
    {
        return (DateTime.Now >= key.Certificate.NotBefore) && (DateTime.Now <= key.Certificate.NotAfter);
    }

    return true;
}
```
but according to the documentation (emphasis mine):

> If set, this delegate will be called to validate the Tokens.SecurityKey that signed the token, **instead of normal processing**.

and while it [doesn't look like much is happening in the way of validation by default](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/Microsoft.IdentityModel.Tokens/Validators.cs#L166), I haven't had a chance to go through the code thoroughly and so may be missing something (plus it could change in the future and I'd be missing out!).

mafurman:
Fixes #805.


brentschmaltz:
:shipit:

PR #820, opened 2017-11-06
Replaced all instances of IMEventSource.Logger.WriteInformation with LogHelper.LogInformation

### LINKED ISSUE #817: Ensure that all calls to the EventSource use LogHelper
Some parts of the code are calling: IdentityModelEventSource.Logger.WriteInformation
Replace this with the appropriate LogHelper function.

mafurman:
Fixes #817.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
+++ b/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
@@ -30,8 +30,8 @@ using System.Collections.Generic;
 using System.IO;
 using System.Text;
 using System.Xml;
+using Microsoft.IdentityModel.Logging;
 using Microsoft.IdentityModel.Xml;
-using static Microsoft.IdentityModel.Logging.IdentityModelEventSource;
 using static Microsoft.IdentityModel.Logging.LogHelper;
 
 namespace Microsoft.IdentityModel.Protocols.WsFederation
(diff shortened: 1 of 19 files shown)


src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
```diff
@@ -28,6 +28,7 @@
 using System;
 using System.Security.Cryptography.X509Certificates;
 using System.Xml;
+using Microsoft.IdentityModel.Logging;
 using Microsoft.IdentityModel.Tokens;
 using Microsoft.IdentityModel.Xml;
 using static Microsoft.IdentityModel.Logging.IdentityModelEventSource;
```
brentschmaltz: Check if you still need any of these 'using static'

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
```diff
@@ -100,7 +101,7 @@ public WsFederationMessage(WsFederationMessage wsFederationMessage)
         {
             if (wsFederationMessage == null)
             {
-                Logger.WriteWarning(FormatInvariant(LogMessages.IDX22000, "wsfederationMessage"));
```
brentschmaltz: Search for 'Logger.Write...' then

mafurman: Done! All of those should be replaced now.

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
mafurman: Got rid of all unnecessary 'using' statements in the files that were modified.

Issue #818, opened 2017-11-06
Ensure that all exceptions that are thrown through LogHelper.


brentschmaltz:
If an ArgumentNullException is thrown without using LogHelper, then the error will not show up in logs.
These type of param checks should use LogHelper.

PR #821, opened 2017-11-07
Made sure that all ArgumentNullExceptions are thrown through the use of LogHelper

mafurman:
Addresses #818.


brentschmaltz:
:shipit:

PR #825, opened 2017-11-08
Saml2SecurityTokenHandler no longer throws a NullReferenceException when the Signature is missing and RequiredSignedTokens is false

### LINKED ISSUE #824: Microsoft.IdentityModel.Tokens.Saml2.Saml2SecurityTokenHandler.ValidateSignature throws on missing signature
`Microsoft.IdentityModel.Tokens.Saml.dll!Microsoft.IdentityModel.Tokens.Saml2.Saml2SecurityTokenHandler.ValidateSignature(string token, Microsoft.IdentityModel.Tokens.TokenValidationParameters validationParameters)` throws a `NullReferenceException` if the Saml2 assertion contains no signature, even though `TokenValidationParameters.RequireSignedTokens` is `false`.

I hit this when debugging against the 5.02.0.41106 nightly build from myget. I'll try to copy over a test case to repro it and fix it.

Regarding the scenario: It is fairly common when using the SAML2 protocol that the assertion XML node does not contain a signature. Instead the entire response message is signed. So handling an assertion without a signature is a must to be able implement SAML2P.

mafurman:
Fixes #824.


brentschmaltz:
:shipit:

PR #828, opened 2017-11-10
update license to rtw

MSagar07:


brentschmaltz:
we don't plan on releasing any additional 3.x products

PR #830, opened 2017-11-11
Saml2Serializer no longer requires NameID to be an absolute URI

### LINKED ISSUE #815: NameID validation rules in Saml2Serializer
In [Saml2Serializer.ReadNameIdTyp](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/82ecdbd6454a1900e963c315fe54ff3c4656b7e7/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs#L1018) some validation is done on the name identifier. There's two potential issues with this.

1. The validation is only done if `nameIdentifier.Format` is explicitly set to `urn:oasis:names:tc:SAML:2.0:nameid-format:entity`. This is correct per section 8.3.6. in the spec. But in some cases, most notably when reading an `<Issuer>` element an unspecified value should be treated as entity format (section 2.2.5). So if the element being read is the issuer, the validation should run also on a non-specified format.

2. SAML2 implementations frequently break this rule, see Sustainsys/Saml2#558 for some discussion about this. Getting an upstream Idp to change it's EntityId to conform to the standard is sometimes not possible. And I cannot really see a security reason to do the validation. Some kind of configurable opt-out of the validation would be great to have.

Actually the missing validation if the `nameIdentifier.Format` isn't set makes it possible to bypass the validation as per point 2, but to me it feels more like exploiting a bug than a proper design.

mafurman:
Fixes #815.

Also added in similar tests for SamlSerializer (even though the absolute URI check was not being performed).

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
@@ -1020,9 +1020,6 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
                 if (nameIdentifier.Format != null &&
                     StringComparer.Ordinal.Equals(nameIdentifier.Format.OriginalString, Saml2Constants.NameIdentifierFormats.Entity.OriginalString))
                 {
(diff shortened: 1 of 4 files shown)


ChenYan98:
:shipit:

PR #837, opened 2017-11-16
Remove filePatterns

ChenYan98:
In this change, we only verify the specific dll in each folder.

### DIFF
--- a/build/VerifySigning.ps1
+++ b/build/VerifySigning.ps1
@@ -7,7 +7,7 @@ if(($root -eq $null) -or ($root -eq [System.String]::Empty))
 
 $srcPath = $root + "\src"
 
-Write-Host ">>> Searching for sn tool..."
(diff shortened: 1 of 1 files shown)


build/VerifySigning.ps1
```diff
@@ -48,39 +47,36 @@ foreach ($project in $projects)
...
                     }
+                }
 
-                    if ($unSigned)
-                    {
-                        Write-Host "$file is unsigned."
-                    }
+                if ($unSigned)
+                {
+                    Write-Host "$file is unsigned."
```
brentschmaltz: does writing this fault?

ChenYan98: Yes, if one of the assembly is unsigned, it will.

brentschmaltz:
:shipit:

PR #841, opened 2017-11-28
add support for JWE token to OpenIdConnectProtocolValidator

van-vothanh:
use the `alg` value from the header of the decrypted jwt for `c_hash` and `at_hash` validations


brentschmaltz:
@van-vothanh I am closing this now as we shouldn't be accepting JWE in the protocol verifier.

We need a model for web server runtimes to handle JWE and pass clear text.

scottbrady91:
>We shouldn't be accepting JWE in the protocol verifier

@brentschmaltz So, if I receive an identity token as a JWE (compact serialization), you're expecting `OpenIdConnectProtocolValidator` to receive the inner token?

brentschmaltz:
@scottbrady91 I think that makes sense, as the JWE is just a wrapper of a JWS. The JWS has the payload that needs validation. It's also the alg of the JWS that defines the hash.

Issue #844, opened 2017-11-30
Nuget.org site should show License Info.


brentschmaltz:
See: https://www.nuget.org/packages/Microsoft.IdentityModel.Clients.ActiveDirectory/
Under 'Info', there should be a license Info link.

Issue #843, opened 2017-11-30
The author for nuget packages should be just Microsoft.


brentschmaltz:
Currently the author is "Microsoft AzureAD"
It should be just "Microsoft"

PR #845, opened 2017-12-01
Changed author for nuget packages to Microsoft and added in the package license url

mafurman:
Addresses #844 and #843.


brentschmaltz:
:shipit:

Issue #834, opened 2017-11-14
JSON Web Token validity verification passed on .NET Core but failed on .NET 4.6.1


petrovic-ivan:
I opened this issue on stackoverflow because I am not able properly to verify is JWT token valid or not. I am using JWKS downloaded from URL and try to validate JWT. I am using same code for both .Net 4.6.1 and .Net Core. In .Net Core validation passed but in .Net 4.6.1 validation failed. 

I am expecting validation to pass becasue I used valid JWT. 

I saw that `AsymmetricSignatureProvider` handle differently .NET Core and .NET 4.6.1. Can you please redirect me to some instructions how to handle those two versions?

Here is opened issue on [Stackoverflow ](https://stackoverflow.com/questions/47294960/json-web-token-validity-verification-passed-on-net-core-but-failed-on-net-4-6)with more details.

brentschmaltz:
@petrovic-ivan we will get to the bottom of this. It would be helpful if you could indicate what exception is occurring. If you could include the token and keys.
@lovemaths please help petrovic.

lovemaths:
@petrovic-ivan I tried the code provided in stackoverflow, it seems to work for .NET 4.6.1. What are the version of Microsoft.IdentityModel.Tokens and System.IdentityModel.Tokens.Jwt you are using? I am using 5.2.0-preview. What was the exception you got?

petrovic-ivan:
@lovemaths Thanks for digging into. I am using Microsoft.IdentityModel.Tokens with assembly version 5.1.4 (this come from master branch). Exception which I have is `SecurityTokenInvalidSignatureException`. I will try code with 5.2.0-preview version and share feedback., also I am trying to share some JWT and JWKS which I am using and elaborate more this issues.

petrovic-ivan:
Here is repo where I added my [examople-identitymodel-issue](https://github.com/petrovic-ivan/example-identitymodel-issue) app. I added README file which will explains how to reproduce issue. It demonstrates how same code with same JWT, JWKS and same version of System.IdentityModel.Tokens.Jwt works differently. Please let me know if I am doing something wrong.

petrovic-ivan:
@lovemaths  Do you have any news since I provided source code? Please note that JWT which I provided has exp date. Please let me know if you had issue to reproduce.

lovemaths:
[...]

PR #846, opened 2017-12-01
Fix Issue #834 JSON web token validity verification failed on .NET 4.6.1

lovemaths:


src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
```diff
@@ -165,5 +190,39 @@ public override int KeySize
...
+            return rsaParameters;
+        }
+
+        private byte[] RemoveLeadingZeroInBase64UrlDecodedBytes(byte[] bytes)
+        {
+            if (bytes == null)
+                return bytes;
+
+            var n = bytes.Length;
+            if (n > 0 && bytes[0].Equals(0x00))
```
brentschmaltz: What happens if n==1?

lovemaths: all rsa parameters are positive integers, if n==1, the byte cannot be 0, so this is fine. I changed the condition to n>1 in the add fix anyway.

brentschmaltz:
:shipit:

Issue #850, opened 2017-12-01
Writing xml (Saml, Saml2, Dsig) should have namespaces.


brentschmaltz:
When xml is written it should have the namespace in the first element.
Saml, Saml2 and Dsig should be added.

brentschmaltz:
@lovemaths can you have at look at this and get us an idea how much work this would be. Writing tokens is not a P0 scenario for this release, but if this is a reasonable amount of work we could consider it.

brentschmaltz:
@mafurman I noticed that SamlConstants and Saml2Contants have a different value for "saml" prefix.
Saml1 - > PreferredPrefix, Saml2 Prefix. Lets make is PreferredPrefix.

PR #854, opened 2017-12-13
Made sure that SamlSerializer, Saml2Serializer, DSigSerializer, and WsFederationMetadataSerializer all use the correct prefixes when writing XML

mafurman:
Addresses #850.


brentschmaltz:
:shipit:

Issue #855, opened 2017-12-19
Saml2SecurityTokenHandler should not allow Serializer to be set to null


brentschmaltz:
need to add a null check.

PR #857, opened 2017-12-19
Added in a null check in the setter of the Serializer property in Saml2SecurityTokenHandler

mafurman:
Addresses #855.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
@@ -69,14 +69,7 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
         public Saml2Serializer Serializer
         {
             get { return _serializer; }
-
-            set
-            {
-                if (value == null)
-                    throw new ArgumentNullException(nameof(value));
-
-                _serializer = value;
-            }
+            set { _serializer = value ?? throw LogHelper.LogArgumentNullException(nameof(value)); }
         }
 
         /// <summary>
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
@@ -49,6 +49,7 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
     {
         private const string Actor = "Actor";
         private int _defaultTokenLifetimeInMinutes = DefaultTokenLifetimeInMinutes;
+        private Saml2Serializer _serializer = new Saml2Serializer();
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -64,11 +65,19 @@ public Saml2SecurityTokenHandler()
...
         /// Gets or set the <see cref="Saml2Serializer"/> that will be used to read and write a <see cref="Saml2SecurityToken"/>.
         /// </summary>
+        /// <exception cref="ArgumentNullException">'value' is null.</exception>
         public Saml2Serializer Serializer
         {
-            get;
-            set;
-        } = new Saml2Serializer();
+            get { return _serializer; }
+
```
brentschmaltz: remove this line

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -64,11 +65,19 @@ public Saml2SecurityTokenHandler()
...
+            set
+            {
+                if (value == null)
```
brentschmaltz: use 'null condition operators' pattern, always use LogHelper to throw exceptions.
_serializer = value ?? LogHelper.LogArgumentNullException(nameof(value))

test/Microsoft.IdentityModel.Tokens.Tests/IdentityComparerTests.cs
```diff
@@ -572,11 +572,11 @@ public void CompareSaml2SecurityTokenHandlers()
         {
             TestUtilities.WriteHeader($"{this}.CompareSaml2SecurityTokenHandlers", true);
             var context = new CompareContext($"{this}.CompareSaml2SecurityTokenHandlers");
-            var saml2SecurityTokenHandler1 = new Saml2SecurityTokenHandler { Serializer = null };
+            var saml2SecurityTokenHandler1 = new Saml2SecurityTokenHandler { Serializer = new Saml2Serializer() };
             var saml2SecurityTokenHandler2 = new Saml2SecurityTokenHandler();
             IdentityComparer.AreEqual(saml2SecurityTokenHandler1, saml2SecurityTokenHandler2, context);
 
-            Assert.True(context.Diffs.Count(s => s == "Serializer:") == 1);
+            Assert.True(context.Diffs.Count(s => s == "Serializer:") == 0);
```
brentschmaltz: Do we need this check?

mafurman: No, I don't think we do. I wanted to keep some sort of test for Saml2SecurityTokenHandler, but it doesn't have any other public properties (aside from Serializer) that are checked by IdentityComparer. 

I'll go ahead and remove it entirely.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
mafurman: Done! Some properties in other classes put a space between the 'get' and 'set' so I wasn't sure what would be considered the correct style.

Issue #862, opened 2018-01-02
Add a perl script to ensure that any RegEx usage uses CultureInvariant unless an exception is recorded


brentschmaltz:
We had to push a fix 5.1.5 as without this, big servers were gobbling up large amount of memory.

see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/3d27674c6eec8f55e3500973082f818ef331129e#diff-93fdf196bfa840222d2ca889af286336R105

Add a perlscript to our build process that checks to make sure we use: RegexOptions.CultureInvariant

brentschmaltz:
We are going to use code analyzers.

PR #870, opened 2018-01-25
Created a custom Roslyn static code analyzer that checks for proper usage of RegexOptions.CultureInvariant.

mafurman:
Addresses #862.

Created a custom Roslyn static code analyzer that ensures Regex usage always includes the RegexOptions.CultureInvariant option. There are two main cases that are caught by this analyzer:

1) A warning is displayed for any Regex that is instantiated without the RegexOptions.CultureInvariant option present.  
2) An error is shown if a static method from the Regex class (such as Regex.Match, Regex.IsMatch, or Regex.Matches) is called without the RegexOptions.CultureInvariant option present. This error is not shown for instance methods since we cannot be sure what options the particular instance of a Regex was created with.

This project has a separate build script from the main repository, and this separate build is called from the main build.ps1 PowerShell script (this ensures the latest version of the tool is used during the build process).


brentschmaltz:
@mafurman lets pull this into 5.2.3

mafurman:
We're no longer using Regex, so this PR is no longer relevant.

Issue #882, opened 2018-02-28
ES512/P-521 point parameter size is 66 bytes, not 64


omsmith:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/d771b5c3ef22b7ff065e8fad1a63d6a2937b7d7f/src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs#L436-L439

Found when working on migrating our library from 4.0.1 and our ES512 integration tests failed (we had implemented our own `EcDsaSecurityKey`).

PR #883, opened 2018-02-28
Correct the P521 keyByteCount in JsonWebKey

omsmith:
The points on P-521 are 66 bytes, not 64. This is likely a leftover
from being incorrectly referred to as P-512.

Fixes: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/882


brentschmaltz:
@omsmith thanks

PR #890, opened 2018-03-14
read 'InclusiveNamespace' with and without namespace prefix

brentschmaltz:
When writing the element, write using correct namespace, was using a dsig namespace. That is incorrect.
Remove unused namespaces
Add two tests, one with and one without namespace


mafurman:
:shipit:

Issue #901, opened 2018-03-31
JsonWebKeySet fail to deserialize in some cases 


hanabi1224:
When JsonConvert.DefaultSettings.ObjectCreationHandling is set to ObjectCreationHandling.Replace, JsonWebKeySet cannot deserialize Keys property thus causes IDX10500  : Signature validation failed. No security keys were provided to validate the signature.

There's also bug in OpenIdConnectConfigurationRetriever line 100, openIdConnectConfiguration.JsonWebKeySet = JsonConvert.DeserializeObject<JsonWebKeySet>(keys);
When the same default JsonConvert setting is set, it fails to deserialize keys

brentschmaltz:
@mafurman please have a look at this.

PR #902, opened 2018-03-31
Fix bug in WebKeySet deserialization

hanabi1224:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/901 
https://github.com/JamesNK/Newtonsoft.Json/issues/703

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
+++ b/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
@@ -67,10 +67,20 @@ namespace Microsoft.IdentityModel.Tokens
         /// Initializes an new instance of <see cref="JsonWebKeySet"/> from a json string.
         /// </summary>
         /// <param name="json">a json string containing values.</param>
+        /// <exception cref="ArgumentNullException">If 'json' is null or empty.</exception>
+        /// <exception cref="ArgumentException">If 'json' fails to deserialize.</exception>
+        public JsonWebKeySet(string json) : this(json, null)
+        {
+        }
+
+        /// <summary>
+        /// Initializes an new instance of <see cref="JsonWebKeySet"/> from a json string.
+        /// </summary>
+        /// <param name="json">a json string containing values.</param>
         /// <param name="jsonSerializerSettings">jsonSerializerSettings</param>
         /// <exception cref="ArgumentNullException">If 'json' is null or empty.</exception>
         /// <exception cref="ArgumentException">If 'json' fails to deserialize.</exception>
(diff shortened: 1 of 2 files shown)


brentschmaltz:
@hanabi1224 Thanks for the fix. I am concerned about the static in the test as sometimes we have issues with parallel tests. How about adding an overload to JsonWebKeySet(string, JsonSerializerSettings).

hanabi1224:
@brentschmaltz  Do you mean overloading the constructor of JsonWebKeySet or JsonWekKeySetDataSet in test method?

BTW, if this static setting causes flakiness to other test cases, it might indicate there're similar bugs in other places.

brentschmaltz:
@hanabi1224 Overloading the JsonWebKeySet. In general, statics have given us some pain when running tests in parallel. With the overload to JsonWebKeySet, we could test it out without affecting other tests.

hanabi1224:
@brentschmaltz  Can you please elaborate? I don't quite understand, is it a constructor only for tests?  And what if we want to test  OpenIdConnectConfigurationRetriever, it's directly calling openIdConnectConfiguration.JsonWebKeySet = JsonConvert.DeserializeObject(keys);   how would it help by overloading JsonWebKeySet constructor?

brentschmaltz:
@hanabi1224  If we don't have the instance control and we rely on the static, the entire application and all code in the call graph would be affected if someone wants to Deserialize a JsonWebKeySet using a specific setting. This insulates the call graph to just the one call. Calls to OpenIdConnectConfigurationRetriever is another issue. In fact, we may want to think about how other types perform serialization.

hanabi1224:
@brentschmaltz  I think this bug is essentially caused by improper usage of Json.Net (setter should always be supplied for JsonProperty) and has nothing to do with instance control of JsonConvertSettings, that's why I'm not convinced to change JsonWebKeySet constructor, your proposal sounds like a different way of fixing the problem, keep improper usage of Json.Net and always use a good JsonConvertSetting that happen to work.

brentschmaltz:
@hanabi1224 no, not at all. We should have the private setter, I thank you for bringing that to our attention.
I am working on the presumption that it may be the case a user wants to use a specific JsonSerializerSettings for an object as that gives finer grained control. Something I think would add benefit for a library.
As for testing, if you can't test a library, you really don't have a library. Which means it has to be developed so it can be tested.
I have found global static properties cause test issues and result in serializing tests.

hanabi1224:
@brentschmaltz  I have updated PR as u suggested, can you plz take another look?

hanabi1224:
@brentschmaltz  PR updated

brentschmaltz:
@hanabi1224 thanks!
We'll pull it in.

PR #903, opened 2018-04-02
Changed P-521 point parameter size to 66 bytes

mafurman:
Addresses #882.


omsmith:
Is there a problem with #883?

brentschmaltz:
@omsmith i just merged #883, we didn't notice the PR before we did this work. Thanks.

Issue #898, opened 2018-03-28
Saml2SecurityTokenHandler calls ValidateTokenReplay with the NotBefore time as the token expiration time instead of NotOnOrAfter


mdw211:
```csharp
public class Saml2SecurityTokenHandler : SecurityTokenHandler
{
    public override ClaimsPrincipal ValidateToken(string token, TokenValidationParameters validationParameters, out SecurityToken validatedToken)
    {
        ...
        ValidateTokenReplay(samlToken.Assertion.Conditions.NotBefore, token, validationParameters);
```

brentschmaltz:
@mdw211 nice catch.
We will fix this in 5.2.2.

PR #905, opened 2018-04-03
Made sure ValidateTokenReplay() uses 'NotOnOrAfter' as the token expiration time.

mafurman:
Addresses #898.
Previously 'NotBefore' time was used.


brentschmaltz:
we must be missing a test here, probably don't need a new one. there should be a test can have this added

mafurman:
@brentschmaltz I'll look for an existing test where a test case for this issue can be added.

brentschmaltz:
:shipit:

Issue #885, opened 2018-03-06
JsonWebKey KeyId is null resulting in JwtHeader with no kid


kevin-meyer:
I'm creating a `JsonWebKey` from json and using it in the `SigningCredentials` for a `JwtSecurityToken`.  This results in a `JwtHeader` with no `kid` defined (it looks to the `signingCredentials.Key.KeyId`).

Shouldn't `JsonWebKey` keep the base `KeyId` in sync with the `Kid`?

brentschmaltz:
@kevin-meyer yes that is a bug. To get around it, as you probably figured out, you will need to set JsonWebKey.Kid.

PR #907, opened 2018-04-03
JsonWebKey 'kid' and 'KeyId' values are now kept in sync.

mafurman:
Addresses #885.

The SecurityKey.KeyId property has been changed to virtual so that JsonWebKey can override it. JsonWebKey now uses the same backing field for its 'KeyId' and 'kid' properties, ensuring that when one changes the other changes as well (ensuring the two properties stay in sync).


brentschmaltz:
:shipit:

PR #910, opened 2018-04-09
Fixed errors found by FxCop.

mafurman:
FxCop is still displaying certain warnings that we want to suppress; a separate task has been created to track that issue.


brentschmaltz:
nice :shipit:

PR #912, opened 2018-04-09
Fixed a bug in the AsymmetricSignatureProvider_Constructor() test

mafurman:
The AsymmetricSignatureProvider_Constructor() test was failing since one of the test cases was using a JsonWebKey that wasn't formatted properly.


brentschmaltz:
@mafurman I had fixed this in the work for caching providers as tests were breaking.

PR #916, opened 2018-04-10
Added the 'x5t' claim to the JwtHeader when the Security Key used is an X509SecurityKey.

### LINKED ISSUE #908: Add 'x5t' claim to JWTHeader when SecurityKey is of type X509SecurityKey
In 4.x we set 'x5t'.
In 5.x we set 'kid'

Some runtimes are expecting one or the other and are broken by this change from 4.x to 5.x.
So we will add both.

mafurman:
Fixes #908.

### DIFF
--- a/src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
@@ -94,6 +94,10 @@ namespace System.IdentityModel.Tokens.Jwt
 
                 if (!string.IsNullOrEmpty(signingCredentials.Key.KeyId))
                     Kid = signingCredentials.Key.KeyId;
+
+                var x509SecurityKey = signingCredentials.Key as X509SecurityKey;
+                if (x509SecurityKey != null && x509SecurityKey.Certificate != null)
+                    X5t = Base64UrlEncoder.Encode(x509SecurityKey.Certificate.GetCertHash());
             }
 
             Typ = JwtConstants.HeaderType;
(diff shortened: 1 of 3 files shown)


brentschmaltz:
@leastprivilege thanks for the heads up. Will think about this.

leastprivilege:
I would rather prefer you fix your old libraries to use kid. And we can get rid of x5t altogether.

brentschmaltz:
@leastprivilege are you asking for a new version of 4x?

brentschmaltz:
@leastprivilege i agree that x5t is a carryover from legacy. 
So if I understand, if 4.x would use the 'kid' you would be fine.

You may want to think about
header["x5t"] = Base64Url.Encode(cert.GetCertHash()));
rather than
header.Add("x5t", Base64Url.Encode(cert.GetCertHash()));

brentschmaltz:
@leastprivilege thanks again for point this out. This would have been a runtime breaking change for users.
We have decided against taking this change.

brentschmaltz:
@leastprivilege this is causing problems with users that are expecting the x5t, some IdentityProviders are expecting the x5t in the ClientCredentials flow.
We will need to add it back.

leastprivilege:
Some?

Then why not add it manually in this case like we do - instead of adding it by default?

brentschmaltz:
@leastprivilege well to be honest, anyone calling into AAD using OAuth ClientCredential flow. AAD doesn't consider the kid, only the x5t.
I am thinking of reintroducing X509SigningCredentials, ctor(X509Certificate2). The JwtHeader will recognize that and add the x5t in this case.
Would that work for you.

I you have a chance, would love your opinion on our new model for JSonWebToken in Microsoft.IdentityModel.Tokens.Jwt. We are moving towards simplifying on consuming and returning JObject. It's early, we want to start small and build up the correct API's.

brentschmaltz:
We re-introduced X509SigningCredentials that adds x5t

Issue #919, opened 2018-04-11
Modify JwtHeader() to not call this


brentschmaltz:
A side effect is that the serialization on the claims can change and be different from the 'json' string that was passed in.

This combination of methods could result in different json
public static JwtHeader Deserialize(string jsonString)
public virtual string Base64UrlEncode()

PR #920, opened 2018-04-11
The empty JwtHeader() constructor no longer calls into one of the other constructors.

mafurman:
Addresses #919.

### DIFF
--- a/test/System.IdentityModel.Tokens.Jwt.Tests/CreateAndValidateTokens.cs
+++ b/test/System.IdentityModel.Tokens.Jwt.Tests/CreateAndValidateTokens.cs
@@ -408,6 +408,48 @@ namespace System.IdentityModel.Tokens.Jwt.Tests
             return theoryData;
         }
 
+        [Theory, MemberData(nameof(SerializeDeserializeJwtTokensTheoryData))]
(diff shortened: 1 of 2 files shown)


test/System.IdentityModel.Tokens.Jwt.Tests/CreateAndValidateTokens.cs
```diff
@@ -408,6 +408,48 @@ public static TheoryData<JwtTheoryData> RoundTripTokensTheoryData()
...
+                theoryData.ExpectedException.ProcessException(ex, context);
+            }
+
+            TestUtilities.AssertFailIfErrors(context);
+        }
+
+        public static TheoryData<JwtTheoryData> SerializeDeserializeJwtTokensTheoryData()
+        {
+            var theoryData = new TheoryData<JwtTheoryData>();
+            var handler = new JwtSecurityTokenHandler();
```
brentschmaltz: I don't think you need the handler.
You can also save some code and write: return new TheoryData<JwtTheoryData>{ new JwtTheoryData{ ...} };

mafurman: You're right. Just made these changes.

Issue #933, opened 2018-05-23
[JWT] MaximumTokenSizeInBytes usage is surprising


ycrumeyrolle:
I have a well-formed JWS : [base64url-encoded header].[base64url-encoded payload].[base64url-encoded signature].
This token is about 1.7MB in its compact serialization form. 
When I try to validate this token, I have an error `IDX12709` saying that my token is malformed and should a JWE or a JWE... The error message is not very useful. When I look at the source code, I see many usage of the same error message : 
 - The token has 2 '.' but does not match the JWS regex
 - The token has 4 '.' but does not match the JWE regex
 - The token has something else than 2 or 4 '.'
 - The token cannot be read:
    - The token is empty
    - The token length * 2 > MaximumTokenSizeInBytes
    - Others validations already done previously

My token length is 1.7MB:
token.length (1.7MB) * 2 > MaximumTokenSizeInBytes (2MB) == True

I found my problem! MaximumTokenSizeInBytes should be renamed HalfMaximumTokenSizeInBytes and the workaround is just to double this value. :roll_eyes: 

This is strange because there is an error for the token too big (`IDX10209`) but this error is raised when token length >MaximumTokenSizeInBytes

My points are: 
 - The error message is unclear
 -  Two validations for the same data (token.length) with the same parameter (MaximumTokenSizeInBytes) produce two different behaviours
 - I expect that my token length can be of 2MB if I set the maximum to 2MB

brentschmaltz:
@ycrumeyrolle sorry for the hassle, will fix in 5.2.3

ycrumeyrolle:
No problem, there is a workaround by setting to an upper value.

I was wondering why there is a size limit? Is it for security reasons?

brockallen:
>  This token is about 1.7MB in its compact serialization form.

Why so big? Doesn't your web server reject the HTTP request because the headers are too large? IIRC, IIS or Kestrel rejects after 16K of headers.

ycrumeyrolle:
The JWT is not an access token used sent within the http headers but as the http payload.
The payload is a document, and the JWT is  used as a standard structure to exchange data. 
This is for a POC and I am looking for the drawback of this.

mafurman:
@ycrumeyrolle The reason why we use 'token.length * 2' is because each character in the token string is 2 bytes. We know there's some places in the code where we don't double the length, and we're planning on fixing this inconsistency.

[...]

PR #938, opened 2018-05-31
Made sure that token length is never doubled when compared to MaximumTokenSizeInBytes across all SecurityTokenHandlers

mafurman:
Addresses #933.


src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -183,7 +183,7 @@ protected virtual void AddActorToAttributes(ICollection<SamlAttribute> attribute
         /// <returns>'true' if the string has a start element equal <see cref="SamlConstants.Elements.Assertion"/>.</returns>
         public override bool CanReadToken(string securityToken)
         {
-            if (string.IsNullOrWhiteSpace(securityToken) || securityToken.Length > MaximumTokenSizeInBytes)
+            if (string.IsNullOrWhiteSpace(securityToken) || securityToken.Length * 2 > MaximumTokenSizeInBytes)
```
CBaud: Instead of multiplying by the magic number '2', I would recommend using 'sizeof(char)'.

Issue #941, opened 2018-06-04
All the ReleaseProvider methods in CryptoProviderFactory should throw an ArgumentNullException if the provider passed into them is null


mafurman:
Currently only ReleaseSignatureProvider does this.

ReleaseHashAlgorithm, ReleaseKeyWrapProvider, and ReleaseRsaKeyWrapProvider should be modified to behave in a consistent manner.

Issue #932, opened 2018-05-22
Extensibility for Release is broken for CustomCryptoProvider


brentschmaltz:
CryptoProvider.CreateSignatureProvider checks if a CustomCryptoProvider is set, but releasing doesn't.
This is unbalanced and breaks extensibility.
We should fix this for all Create / Release patterns.

PR #943, opened 2018-06-05
Added in support for CustomCryptoProvider in the release methods of CryptoProviderFactory

mafurman:
Addresses #932 and #941.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
+++ b/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
@@ -26,6 +26,7 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Security.Cryptography;
 using Microsoft.IdentityModel.Logging;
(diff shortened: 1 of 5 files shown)


src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -40,6 +41,7 @@ public class CryptoProviderFactory
     {
         private CryptoProviderCache _cryptoProviderCache = new InMemoryCryptoProviderCache();
         private static CryptoProviderFactory _default;
+        private static IDictionary<string, string> _typeToAlgorithmMap = new Dictionary<string, string>();
```
brentschmaltz: Please use a ConcurrentDictionary
https://msdn.microsoft.com/en-us/library/dd287191(v=vs.110).aspx

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -507,6 +511,194 @@ public static TheoryData<SignatureProviderTheoryData> ReleaseSignatureProvidersT
...
+                };
+
+                return theoryData;
+            }
+        }
+
+        [Theory, MemberData(nameof(ReleaseHashAlgorithmsTheoryData))]
+        public void ReleaseHashAlgorithms(CryptoProviderFactoryTheoryData theoryData)
+        {
+            IdentityModelEventSource.ShowPII = true;
```
brentschmaltz: the constructor to TheoryDataBase has code: IdentityModelEventSource.ShowPII = true;

Issue #949, opened 2018-06-14
System.FormatException at LogHelper.FormatInvariant when called from IdentityModelEventSource


aakash1906:
Added Appinsight listener to IdentityModelEventSource and set loglevel to verbose first
At GetConfigurationAsync of OpenId, there was an exception at string.Format inside LogHelper.FormatInvariant inside IdentityModelEventSource- **PrepareMessage**.

aakash1906:
Raised a pull request to fix the same.

brentschmaltz:
@aakash1906 thanks for the fix. 
We are pushing into dev, can you package up the fix in dev?
We use master as a place holder.

aakash1906:
Have raised a pull request to dev with that fix. Please review

brentschmaltz:
I like the fix, loggers should never fault.
Thanks.

aakash1906:
Thanks for closing this.!

PR #953, opened 2018-06-15
Supporting fix for #949

aakash1906:


aakash1906:
supporting for fix #949

Dark-Bond:
After setting IdentityModelEventSource.ShowPII = true;, I also tried to force a resolution to the StringFormatter class.  Now, I'm getting this:

https://imgur.com/a/8x9Okp9

Dark-Bond:
It would appear that whatever process attempts to instantiate **System.FormattableString**, that process will blow up Azure,  In the case of the **Microsoft.IdentityLibrary**, we have the issue in:

`The type initializer for 'Microsoft.IdentityModel.Protocols.HttpDocumentRetriever' threw an exception.
`
with an innner exception of:

`Could not load type 'System.FormattableString' from assembly 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.
`
When I tried to instantiate a **System.FormattableString** in the method that called **configManager.GetConfigurationAsync()**, I took the exception in that method instead of the Microsoft.Identity library.

Issue #958, opened 2018-06-16
Issue with LogHelper.FormatInvariant


aakash1906:
Though the logging issue of #949 is addressed, still, the Root cause, is at this method. LogHelper.FormatInvariant
When any string contains formats like "{asdf" etc., this will cause exception.

PR #959, opened 2018-06-16
Patch 3

aakash1906:
Fix for issue #958 

Sample Test: addressing the issue.,and supporting code fix:

var format = "{ 0} asdf {1} {asdf} {aasdf {{asdf2} {pr} asdf}  {1} asdf3}  asd}f asd{f {{{{proc}o}pond}} {a{1}p}  {{{{asdf}}  {{9{asdf}}";
format = SafeStringFormat(null, format, "", "replaced");
//Output: { 0} asdf replaced {asdf} {aasdf {{asdf2} {pr} asdf}  replaced asdf3}  asd}f asd{f {{{{proc}o}pond}} {areplacedp}  {{{{asdf}}  {{9{asdf}}


brentschmaltz:
@aakash1906 in general we shy away from using Regex. 
It has been known to bring down servers.
When we do use it, we do so sparingly and always compiled with culture invariant.

aakash1906:
@brentschmaltz So, what do you suggest now. 
Intention is mainly only to avoid exception. Thats why, have put on few lines of code, to detect Format, IndexOutOfRange Exceptions, and tune the input before calling string.Format 
(Because, I was able to get some scenarios where inputs from DocumentRetriever or something contained strings with Curly braces causing FormatException when LogHelper.FormatInvariant is directly used)

Now, shall I change it in a way, to just have a simple try catch around it, with a motto that logging shouldn't cause exceptions?

mafurman:
@aakash1906 We can't use this Regex code in our logging since we are a server component, and logging needs to be fast. Users of our library may need to validate many tokens in a short period of time, so milliseconds matter and we can't have logging slow this process down.

We understand that avoiding format exceptions while logging is important, and we will add in a rule that will check for mismatched braces in our log messages. I've created an issue to track this: #989.

brentschmaltz:
@aakash1906 we took action on your action and introduced a script at build time that checks for parameter validation.

see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/VerifyResourceUsage.pl

PR #970, opened 2018-06-26
Overcome 'The Year 2038 problem'

fomenta:
Overcome 'The Year 2038 problem' as described in link below:

https://en.wikipedia.org/wiki/Year_2038_problem
The Year 2038 problem relates to representing time in many digital systems as number of seconds passed since 1 January 1970 and storing it as a signed 32-bit integer.

If a date is greater than year 2038, token failed with the following error:

System.OverflowException: 'Value was either too large or too small for Int32.'

With this fix, I can use dates greater than 2038 without a problem. I tried one in year 2130 and it worked just fine.

### DIFF
--- a/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
@@ -25,13 +25,13 @@
 //
 //------------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


brentschmaltz:
We applied the ideas here when we created microsoft.identitymodel.jsonwebtokens.
Since this is where we will move forward, we won't make this change.

Issue #839, opened 2017-11-24
Create a JwtHeader with both Signing and Encrypting Credential


leastprivilege:
Does not seem to be supported. why?

brentschmaltz:
@leastprivilege I am thinking of apis that work with JObject, SigningCredentials, EncryptingCrentials. JObject becomes a JWS using SigningCredentials for definition of how to sign. JWE is created using EncryptingCredentials. I am experimenting with a set of such API's. It seems cleaner and is quite a bit faster.

brentschmaltz:
We will be adding this to M.IM.Tokens.JsonWebTokenHandler in 5.2.4

PR #976, opened 2018-07-10
Added in JWE support for M.IM.JsonWebTokens

### LINKED ISSUE #962: Add JWE support to JsonWebSecurityTokenHandler
Essentially make the following public:

private string CreateJsonWebToken(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials)

mafurman:
Fixes #962, #981, and #839.


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -178,19 +186,24 @@ public JObject Header
         /// Gets the 'value' of the 'JWT ID' claim { jti, ''value' }.
         /// </summary>
         /// <remarks>If the 'JWT ID' claim is not found, an empty string is returned.</remarks>
-        public override string Id => Payload.Value<string>(JwtRegisteredClaimNames.Jti) ?? String.Empty;
+        public override string Id => Payload?.Value<string>(JwtRegisteredClaimNames.Jti) ?? String.Empty;
+
+        /// <summary>
+        /// Gets the <see cref="JsonWebToken"/> associated with this instance.
+        /// </summary>
+        public JsonWebToken InnerToken { get; internal set; }
```
brentschmaltz: It would be best if we didn't need this property.
Try and expose just Header, Payload

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -178,19 +186,24 @@ public JObject Header
...
 
         /// <summary>
         /// Gets the 'value' of the 'iat' claim { iat, 'value' } converted to a <see cref="DateTime"/> assuming 'value' is seconds since UnixEpoch (UTC 1970-01-01T0:0:0Z).
         /// </summary>
         /// <remarks>If the 'exp' claim is not found, then <see cref="DateTime.MinValue"/> is returned.</remarks>
```
brentschmaltz: Review comments to ensure they match the property. 
This one specifies 'exp', but this is the 'iat' claim.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -178,19 +186,24 @@ public JObject Header
...
-        public DateTime IssuedAt => GetDateTime(JwtRegisteredClaimNames.Iat);
+        public DateTime IssuedAt => Payload == null ? DateTime.MinValue : GetDateTime(JwtRegisteredClaimNames.Iat);
 
         /// <summary>
         /// Gets the 'value' of the 'issuer' claim { iss, 'value' }.
         /// </summary>
         /// <remarks>If the 'issuer' claim is not found, an empty string is returned.</remarks>
```
brentschmaltz: This is the 'iss' claim, not 'issuer' claim.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -136,7 +136,7 @@ public string CreateToken(JObject payload, SigningCredentials signingCredentials
         /// <param name="signingCredentials">Defines the security key and algorithm that will be used to sign the JWT.</param>
         /// <param name="encryptingCredentials">Defines the security key and algorithm that will be used to encrypt the JWT.</param>
         /// <returns>A JWT in compact serialization format.</returns>
-        private string CreateJsonWebToken(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials)
+        public string CreateJsonWebToken(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials)
```
brentschmaltz: Change name to CreateToken, with the EncryptingCredentials parameter the overload should be ok.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -189,6 +189,84 @@ private string CreateJsonWebToken(JObject payload, SigningCredentials signingCre
...
+            if (validationParameters == null)
+                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
+
+            if (string.IsNullOrEmpty(jwtToken.Enc))
+                throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(TokenLogMessages.IDX10612)));
+
+            var keys = GetContentEncryptionKeys(jwtToken, validationParameters);
+
+            // keep track of exceptions thrown, keys that were tried
+            StringBuilder exceptionStrings = new StringBuilder();
```
brentschmaltz: use var

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -189,6 +189,84 @@ private string CreateJsonWebToken(JObject payload, SigningCredentials signingCre
...
+            throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10609, jwtToken.EncodedToken)));
+        }
+
+        private string DecryptToken(JsonWebToken jwtToken, CryptoProviderFactory cryptoProviderFactory, SecurityKey key)
+        {
+            var decryptionProvider = cryptoProviderFactory.CreateAuthenticatedEncryptionProvider(key, jwtToken.Enc);
+            if (decryptionProvider == null)
+                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10610, key, jwtToken.Enc)));
+
+            var tokenParts = jwtToken.EncodedToken.Split('.');
```
brentschmaltz: we need to do a controlled split here, with a maximum number of expected '.' (dots).
review what ValidateToken does.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -103,5 +104,18 @@ public static byte[] GenerateKeyBytes(int sizeInBits)
...
             return key;
         }
+
+        /// <summary>
+        /// Gets all decryption keys.
+        /// </summary>
+        public static IEnumerable<SecurityKey> GetAllDecryptionKeys(TokenValidationParameters validationParameters)
+        {
+            if (validationParameters.TokenDecryptionKey != null)
+                yield return validationParameters.TokenDecryptionKey;
```
brentschmaltz: 'yield' can cause an issue with the memory pressure.
It would be best if we could avoid this call.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: Will think about this.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: Went through all the comments and made the appropriate changes.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: I fixed this, as well as other occurrences of the same thing.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Got it. 👍

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: I went through the new code that was added and switched to 'var' wherever possible.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Makes sense. 👍

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: This has been changed so that this method returns a list of all the decryption keys instead of using yield.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -127,9 +126,12 @@ public virtual IEnumerable<Claim> Claims
         {
             get
             {
-                List<Claim> claims = new List<Claim>();
+                var claims = new List<Claim>();
                 string issuer = this.Issuer ?? ClaimsIdentity.DefaultIssuer;
 
+                if (Payload == null)
+                    return claims;
```
CBaud: >claims [](start = 27, length = 6)

Allocating a new List can be avoided in this scenario by moving the IF check above where 'claims' is initialized, and instead returning Enumerable.Empty<Claim>()

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -175,22 +183,27 @@ public JObject Header
...
+        /// </summary>
+        public JsonWebToken InnerToken { get; internal set; }
 
         /// <summary>
         /// Gets the 'value' of the 'iat' claim { iat, 'value' } converted to a <see cref="DateTime"/> assuming 'value' is seconds since UnixEpoch (UTC 1970-01-01T0:0:0Z).
         /// </summary>
-        /// <remarks>If the 'exp' claim is not found, then <see cref="DateTime.MinValue"/> is returned.</remarks>
-        public DateTime IssuedAt => GetDateTime(JwtRegisteredClaimNames.Iat);
+        /// <remarks>If the 'iat' claim is not found, then <see cref="DateTime.MinValue"/> is returned.</remarks>
+        public DateTime IssuedAt => Payload == null ? DateTime.MinValue : GetDateTime(JwtRegisteredClaimNames.Iat);
```
CBaud: >DateTime.MinValue [](start = 54, length = 17)

Nullable<T> seems like a more appropriate contract for optional claims like 'iat', 'exp', 'nbf', etc. Otherwise, this requires consumers of this library to know the internal details, and check for special cases like this.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -319,7 +334,7 @@ private void AddClaimsFromJToken(List<Claim> claims, string claimType, JToken jt
 
         private void AddDefaultClaimFromJToken(List<Claim> claims, string claimType, JToken jtoken, string issuer)
         {
-            JValue jvalue = jtoken as JValue;
+            var jvalue = jtoken as JValue;
             if (jvalue != null)
```
CBaud: These two lines can be consolidated into one. For example: if (jtoken is JValue jvalue)

CBaud:
>             get;

Have you considered refactoring this to always return a valid JObject reference? In my opinion, a JObject with no properties seems like a better way to represent an empty payload.

---
Refers to: src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs:219 in fc908d4. [](commit_id = fc908d40004a849b70f015b8f7200404b0b73b6b, deletion_comment = False)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz: I agree, it would be a breaking change. Given that this api has just been released, it may be a good time to consider changing this. We may not have another opportunity.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: Will make this change. Thanks!

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: Good suggestion. Thanks! 👍

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: I agree that allowing these types to be nullable is more intuitive for a user of the library. However, the 'exp' (ValidTo) and 'nbf' (ValidFrom) claims are represented as DateTime values on the SecurityToken base class. As a result, it's not possible to make them nullable on the JsonWebToken class.

mafurman:
@CBaud That might be a better approach. I will think about making this change. :)

@brentschmaltz What do you think about using an empty JObject to represent an empty payload?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -116,7 +116,7 @@ public IEnumerable<string> Audiences
                         return value.ToObject<List<string>>();
```
CBaud: In order to address issue #981, we also need to check if 'value' is a null reference. The JObject.GetValue(string) method returns null if the property name does not exist.

CBaud: One approach would be to use the IS keyword here. For example, the 'if (Payload != null)' could be replaced by 'if (Payload?.GetValue(JwtRegisteredClaimNames.Aud) is JToken value)'

mafurman: Thanks for pointing this out. :) I just made the change.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -99,35 +99,48 @@ public JsonWebToken(JObject header, JObject payload)
...
                 }
 
-                return new List<string>();
+                return Enumerable.Empty<string>();
             }
         }
 
+        /// <summary>
+        /// Gets the original raw data of this instance when it was created.
```
brentschmaltz: Gets the AuthenticationTag from the original...

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -405,7 +472,7 @@ private DateTime GetDateTime(string key)
             if (!Payload.TryGetValue(key, out jToken))
```
brentschmaltz: use inline variable

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -239,16 +285,16 @@ public override SecurityKey SigningKey
         public string Typ => Header.Value<string>(JwtHeaderParameterNames.Typ) ?? String.Empty;
 
         /// <summary>
-        /// Gets the 'value' of the 'notbefore' claim { nbf, 'value' } converted to a <see cref="DateTime"/> assuming 'value' is seconds since UnixEpoch (UTC 1970-01-01T0:0:0Z).
+        /// Gets the 'value' of the 'nbf' claim { nbf, 'value' } converted to a <see cref="DateTime"/> assuming 'value' is seconds since UnixEpoch (UTC 1970-01-01T0:0:0Z).
         /// </summary>
-        /// <remarks>If the 'notbefore' claim is not found, then <see cref="DateTime.MinValue"/> is returned.</remarks>
-        public override DateTime ValidFrom => GetDateTime(JwtRegisteredClaimNames.Nbf);
+        /// <remarks>If the 'nbf' claim is not found, then <see cref="DateTime.MinValue"/> is returned.</remarks>
+        public override DateTime ValidFrom => !Payload.HasValues ? DateTime.MinValue : GetDateTime(JwtRegisteredClaimNames.Nbf);
```
brentschmaltz: GetDateTime handles null.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -103,5 +104,21 @@ public static byte[] GenerateKeyBytes(int sizeInBits)
 
             return key;
         }
+
+        /// <summary>
+        /// Gets all decryption keys.
+        /// </summary>
+        public static IEnumerable<SecurityKey> GetAllDecryptionKeys(TokenValidationParameters validationParameters)
+        {
+            var decryptionKeys = new List<SecurityKey>();
```
brentschmaltz: Use a Collection

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -58,28 +60,36 @@ public static TheoryData<JwtTheoryData> SegmentTheoryData()
...
-                theoryData.JwtSecurityTokenHandler.ValidateToken(jwtFromJwtHandler, theoryData.ValidationParameters, out SecurityToken validatedToken);
-                theoryData.JsonWebTokenHandler.ValidateToken(jwtFromJsonHandler, theoryData.ValidationParameters);
+                theoryData.JwtSecurityTokenHandler.ValidateToken(jweFromJwtHandler, theoryData.ValidationParameters, out SecurityToken validatedToken);
+                theoryData.JsonWebTokenHandler.ValidateToken(jweFromJsonHandler, theoryData.ValidationParameters);
 
                 theoryData.ExpectedException.ProcessNoException(context);
-                var jwtTokenFromBuilder = new JsonWebToken(jwtFromJwtHandler);
-                var jwtTokenFromHandler = new JsonWebToken(jwtFromJsonHandler);
-                IdentityComparer.AreEqual(jwtTokenFromBuilder, jwtTokenFromHandler, context);
+                var jweTokenFromJwtHandler = new JsonWebToken(jweFromJwtHandler);
```
brentschmaltz: Can we add a test to compare at the Claims level.

mafurman: Added this in for all the tests.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Done.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: Fixed this.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -136,7 +136,7 @@ public string CreateToken(JObject payload, SigningCredentials signingCredentials
         /// <param name="signingCredentials">Defines the security key and algorithm that will be used to sign the JWT.</param>
         /// <param name="encryptingCredentials">Defines the security key and algorithm that will be used to encrypt the JWT.</param>
         /// <returns>A JWT in compact serialization format.</returns>
-        private string CreateJsonWebToken(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials)
+        public string CreateToken(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials)
         {
             if (payload == null)
```
brentschmaltz: null check on payload, signingCredentials, encryptingCredentials.
then call private CreateToken.
Move check for cached header in private CreateToken, no need to pass as parameter.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -126,7 +126,7 @@ public bool CanValidateToken
         /// <returns>A JWS in Compact Serialization Format.</returns>
         public string CreateToken(JObject payload, SigningCredentials signingCredentials)
         {
-            return CreateJsonWebToken(payload, signingCredentials, null);
+            return CreateToken(payload, signingCredentials, null);
```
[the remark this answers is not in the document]

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs

PR #990, opened 2018-08-09
The NotBefore and NotAfter properties of X509Certificates are now checked against local time instead of UTC

### LINKED ISSUE #967: NotBefore/NotAfter and timezones
I've been having a problem where a newly issued certificate is not accepted, because the NotBefore time has not yet been reached.  It seems to be connected with timezones.  According to [the NotBefore documentation](https://docs.microsoft.com/en-gb/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.notbefore?view=netcore-2.1), NotBefore is in local time.  [This file](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/Microsoft.IdentityModel.Tokens/Validators.cs#L213) seems to be responsible for checking that the NotBefore date has been reached, and it is comparing it with UtcNow.

Oddly, [X509SecurityToken.ValidFrom](https://msdn.microsoft.com/en-us/library/system.identitymodel.tokens.x509securitytoken.validfrom(v=vs.110).aspx) is specified as being in UTC, so the other check in the same file seems to be correct.

(I should add, I've checked that my computer's clock has the right time, which would be the obvious cause of problems like this!  Thank you for all your work on the framework, and I hope this report is helpful.)

mafurman:
Fixes #967.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validators.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validators.cs
@@ -28,6 +28,7 @@
 using Microsoft.IdentityModel.Logging;
 using System.Collections.Generic;
 using System;
+using System.Security.Cryptography.X509Certificates;
 
 namespace Microsoft.IdentityModel.Tokens
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -207,16 +208,19 @@ public static void ValidateIssuerSecurityKey(SecurityKey securityKey, SecurityTo
                 throw LogHelper.LogArgumentNullException(nameof(securityToken));
 
             X509SecurityKey x509SecurityKey = securityKey as X509SecurityKey;
-            if (x509SecurityKey != null)
+            if (x509SecurityKey?.Certificate is X509Certificate2 cert)
             {
-                var cert = x509SecurityKey.Certificate;
-                DateTime utcNow = DateTime.UtcNow;
+                DateTime localTime = DateTime.Now;
```
CBaud: The root cause of this issue is a comparison between local and universal time. Rather than making these comparisons all local, have you considered making them all universal? This would keep the code robust against various time concerns, such as differences in time zones, daylight saving, etc. I would recommend bringing back the original implementation with an update to call the ToUniversalTime() method on the NotBefore property of the certificate and comparing that value against UtcNow.

brentschmaltz: thanks Chris.
we found out that X509Certificate2 returns localtime rather than utc time.
we don't have any control over that.

brentschmaltz:
:shipit:

src/Microsoft.IdentityModel.Tokens/Validators.cs
CBaud: It can be easily converted into a UTC kind of DateTime using the DateTime.ToUniversalTime() method. For example: cert.NotBefore.ToUniversalTime()

mafurman: Just made this change. @CBaud Thanks for explaining why this is a good idea!

mafurman:
I removed the cert.NotBefore and cert.NotAfter null checks since they are DateTime values and can never be null.

brentschmaltz:
:shipit:

Issue #963, opened 2018-06-18
Design 'TokenCreation' Policy that guides token creation by SecurityTokenHandlers


brentschmaltz:
Api's on this object would allow users to set standard items in tokens that don't change per token, modify claims, etc. A SecruityTokenHandler would have a ctor that takes such an object.
1. Expiration, NotBefore, IssuedAt parameters
2. Issuer
3. JWTHeader
4. ClaimsMapping
5. SecurityKeys, Algorithms

brentschmaltz:
After discussion, lets forget about 5 as users can manage using the same keys per issuer.
Since we cache the JWTHeader, let's not pass it.
Defer ClaimsMapping until we work with ClaimsIdentity and Claims.

brentschmaltz:
We decided to have a new base class TokenHandler that has properties:
MaximumTokenSize, TokenLifetimeInMinutes, SetDefaultTimesOnTokenCreation

PR #991, opened 2018-08-10
Added in a new abstract TokenHandler class that contains properties shared across all token handlers

mafurman:
Addresses #963.

This PR also adds default time setting functionality (during token creation) to the JsonWebTokenHandler.

Exceptions had to be added to the ApiBreakingChangeTest because it does not take base class methods/properties into account.

Also fixes a small typo mentioned in this issue: #999

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -34,33 +34,14 @@ using System.Text;
 using System.Text.RegularExpressions;
 using TokenLogMessages = Microsoft.IdentityModel.Tokens.LogMessages;
 
-
 namespace Microsoft.IdentityModel.JsonWebTokens
 {
     /// <summary>
     /// A <see cref="SecurityTokenHandler"/> designed for creating and validating Json Web Tokens. 
     /// See: http://tools.ietf.org/html/rfc7519 and http://www.rfc-editor.org/info/rfc7515.
     /// </summary>
-    public class JsonWebTokenHandler 
+    public class JsonWebTokenHandler : TokenHandler
     {
-        private int _maximumTokenSizeInBytes = TokenValidationParameters.DefaultMaximumTokenSizeInBytes;
-
-        /// <summary>
-        /// Gets and sets the maximum token size in bytes that will be processed.
-        /// </summary>
-        /// <exception cref="ArgumentOutOfRangeException">'value' less than 1.</exception>
-        internal virtual int MaximumTokenSizeInBytes
-        {
-            get { return _maximumTokenSizeInBytes; }
-            set
-            {
-                if (value < 1)
-                    throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(value), LogHelper.FormatInvariant(TokenLogMessages.IDX10101, value)));
-
-                _maximumTokenSizeInBytes = value;
-            }
-        }
-
         /// <summary>
         /// Gets the type of the <see cref="JsonWebToken"/>.
         /// </summary>
@@ -82,7 +63,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// </remarks>
         /// <returns>
         /// <para>'false' if the token is null or whitespace.</para>
-        /// <para>'false' if token.Length is greater than <see cref="SecurityTokenHandler.MaximumTokenSizeInBytes"/>.</para>
+        /// <para>'false' if token.Length is greater than <see cref="TokenHandler.MaximumTokenSizeInBytes"/>.</para>
         /// <para>'true' if the token is in JSON compact serialization format.</para>
         /// </returns>
         public bool CanReadToken(string token)
@@ -171,9 +152,23 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 JsonWebTokenManager.KeyToHeaderCache.TryAdd(JsonWebTokenManager.GetHeaderCacheKey(signingCredentials), rawHeader);
             }
 
+            if (SetDefaultTimesOnTokenCreation)
+            {
+                var now = DateTime.UtcNow;
+                if (!payload.TryGetValue(JwtRegisteredClaimNames.Exp, out _))
+                    payload.Add(JwtRegisteredClaimNames.Exp, EpochTime.GetIntDate(now + TimeSpan.FromMinutes(TokenLifetimeInMinutes)).ToString());
+
+                if (!payload.TryGetValue(JwtRegisteredClaimNames.Iat, out _))
+                    payload.Add(JwtRegisteredClaimNames.Iat, EpochTime.GetIntDate(now).ToString());
+
+                if (!payload.TryGetValue(JwtRegisteredClaimNames.Nbf, out _))
+                    payload.Add(JwtRegisteredClaimNames.Nbf, EpochTime.GetIntDate(now).ToString());
+            }
+       
             var rawPayload = Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(payload.ToString(Newtonsoft.Json.Formatting.None)));
             var message = rawHeader + "." + rawPayload;
             var rawSignature = JwtTokenUtilities.CreateEncodedSignature(message, signingCredentials);
+
             if (encryptingCredentials != null)
                 return EncryptToken(message + "." + rawSignature, encryptingCredentials);
             else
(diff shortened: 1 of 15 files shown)


mafurman:
@CBaud That is one of the reasons why it was made into an abstract class. It also allows the logic for the MaximumTokenSizeInBytes and TokenLifetimeInMinutes properties to be shared across all classes that extend TokenHandler. Making this into an interface would mean that each of those classes would need to implement the getters/setters for these properties, which would result in more code duplication.

That being said, I do see how making it into an interface may make more sense... it's definitely something to consider.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -155,6 +148,23 @@ private string CreateJsonWebToken(JObject payload, SigningCredentials signingCre
                 JsonWebTokenManager.KeyToHeaderCache.TryAdd(JsonWebTokenManager.GetHeaderCacheKey(signingCredentials), rawHeader);
             }
 
+            var expires = JwtTokenUtilities.GetDateTime(JwtRegisteredClaimNames.Exp, payload);
```
brentschmaltz: We don't care about the value, just did the user set it.
We could write: 
var expires = payload.TryGetValue(JwtRegisteredClaimNames.Exp, out JToken token);

This returns a bool.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -103,5 +105,47 @@ public static byte[] GenerateKeyBytes(int sizeInBits)
...
+        /// <summary>
+        /// Gets the DateTime using the number of seconds from 1970-01-01T0:0:0Z (UTC)
+        /// </summary>
+        /// <param name="key">Claim in the payload that should map to an integer, float, or string.</param>
+        /// <param name="payload">The payload that contains the desired claim value.</param>
+        /// <remarks>If the claim is not found, the function returns: DateTime.MinValue
+        /// </remarks>
+        /// <exception cref="FormatException">If the value of the claim cannot be parsed into a long.</exception>
+        /// <returns>The DateTime representation of a claim.</returns>
+        public static DateTime GetDateTime(string key, JObject payload)
```
brentschmaltz: Can we make this 'Internal'?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -103,5 +105,47 @@ public static byte[] GenerateKeyBytes(int sizeInBits)
...
+        {
+            JToken jToken;
```
brentschmaltz: Can 'inline' using 
if (!payload.TryGetValue(key, out JToken jToken))

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -103,5 +105,47 @@ public static byte[] GenerateKeyBytes(int sizeInBits)
...
+            if (!payload.TryGetValue(key, out jToken))
+                return DateTime.MinValue;
+
+            long dateValue = ParseTimeValue(jToken, key);
```
brentschmaltz: could lose the dateValue variable and inline it into 
....Convert.ToDouble(ParseTimeValue(jToken, key),

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <remarks>See: <see cref="TokenLifetimeInMinutes"/> for configuration.</remarks>
+        [DefaultValue(true)]
+        public bool SetDefaultTimesOnTokenCreation { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets the token lifetime in minutes.
+        /// </summary>
+        /// <remarks>Used during token creation to set the default expiration ('exp'). </remarks>
+        /// <exception cref="ArgumentOutOfRangeException">'value' less than 1.</exception>
+        public int TokenLifetimeInMinutes
```
brentschmaltz: Could write this a bit cleaner (also MaximumTokenSizeInBytes).

get => _defaultTokenLifetimeInMinutes;
set => _defaultTokenLifetimeInMinutes = (value < 1) ? throw... : value;

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -119,6 +118,8 @@ public JwtSecurityTokenHandler()
             _outboundClaimTypeMap = new Dictionary<string, string>(DefaultOutboundClaimTypeMap);
             _inboundClaimFilter = new HashSet<string>(DefaultInboundClaimFilter);
             _outboundAlgorithmMap = new Dictionary<string, string>(DefaultOutboundAlgorithmMap);
+
+            TokenLifetimeInMinutes = DefaultTokenLifetimeInMinutes;
```
brentschmaltz: TokenLifetimeInMinutes is set in the base

If we move this value to TokenHandler
public static readonly int DefaultTokenLifetimeInMinutes = 60;

then we can set _defaultTokenLifetimeInMinutes = DefaultTokenLifetimeInMinutes.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -147,6 +147,32 @@ public void RoundTripToken()
             TestUtilities.AssertFailIfErrors(context);
         }
 
+        // Test checks to make sure that default times are correctly added to the token
+        // upon token creation.
+        [Fact]
+        public void SetDefaultTimesOnTokenCreation()
```
brentschmaltz: Do we have tests for SAML, SAML2, JwtSecurityToken?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
CBaud: Also, if we truly don't care about the value then I would recommend using a [Discard](https://docs.microsoft.com/en-us/dotnet/csharp/discards). For example:
var expires = payload.TryGetValue(JwtRegisteredClaimNames.Exp, out _);

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -32,7 +32,6 @@
 using System.Collections.Generic;
 using System.Text;
 using System.Text.RegularExpressions;
-using static Microsoft.IdentityModel.Logging.LogHelper;
 using TokenLogMessages = Microsoft.IdentityModel.Tokens.LogMessages;
```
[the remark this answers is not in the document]

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -43,26 +42,20 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// See: http://tools.ietf.org/html/rfc7519 and http://www.rfc-editor.org/info/rfc7515.
     /// Currently only includes support for tokens in JWS format.
     /// </summary>
-    public class JsonWebTokenHandler 
+    public class JsonWebTokenHandler : TokenHandler
     {
-        private int _maximumTokenSizeInBytes = TokenValidationParameters.DefaultMaximumTokenSizeInBytes;
+        /// <summary>
+        /// Default lifetime of tokens created. When creating tokens, if 'expires' and 'notbefore' are both null, then a default will be set to: expires = DateTime.UtcNow, notbefore = DateTime.UtcNow + TimeSpan.FromMinutes(TokenLifetimeInMinutes).
```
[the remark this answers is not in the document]

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: This is a really good suggestion, thanks!  :) 
I've made the change.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Done.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Done.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Done.

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
mafurman: Good suggestion, thanks! 👍

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
mafurman: Done. I initially thought that the different token handlers may not want to share the same default value, but I agree that it makes sense for this property to be located on the base TokenHandler class.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
mafurman: I don't think so. 
I added in a similar test for the other three handlers.

mafurman:
Also fixes this null pointer exception bug in Saml2SecurityTokenHandler: #1003

Issue #989, opened 2018-08-09
Add a rule that will check for mismatched braces inside of exception messages


mafurman:
Mismatched braces inside of exception messages will cause an exception to be thrown at the time the message is being formatted.

We should add a rule that will run through all of our exception messages and ensure that no mismatched braces are present.

PR #993, opened 2018-08-14
Got rid of unused log messages and added in logic for checking mismatched braces in strings

mafurman:
Addresses #989.

Changes made:

- VerifyResourceUsage.pl now ensures that all log message strings are properly formatted. (specifically that all opening braces have a corresponding closing brace)
- The VerifyResourceUsage.pl task on VSTS will now fail if it finds any errors (previously it just printed the errors but still passed).
- Unused log messages have been removed and any mismatches between log message names and their corresponding string values have been fixed.


brentschmaltz:
:shipit:

Nice

PR #997, opened 2018-08-15
Fixed a race condition in CryptoProviderFactory.CreateSignatureProvider

### LINKED ISSUE #996: Fix the race condition in CryptoProviderFactory.CreateSignatureProvider()
In a multi-threading scenario, it's currently possible to create two signature providers but only have one of them added to the CryptoProviderCache. 

We should introduce a double lock to ensure that this cannot happen.

mafurman:
Fixes #996.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
+++ b/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
@@ -41,6 +41,7 @@ namespace Microsoft.IdentityModel.Tokens
     {
         private static CryptoProviderFactory _default;
         private static ConcurrentDictionary<string, string> _typeToAlgorithmMap = new ConcurrentDictionary<string, string>();
+        private static object _cacheLock = new object();
 
         /// <summary>
         /// Returns the default <see cref="CryptoProviderFactory"/> instance.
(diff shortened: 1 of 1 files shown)


src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -41,6 +41,7 @@ public class CryptoProviderFactory
     {
         private static CryptoProviderFactory _default;
         private static ConcurrentDictionary<string, string> _typeToAlgorithmMap = new ConcurrentDictionary<string, string>();
+        private static object _cacheLock = new object();
```
CBaud: Another approach would be to move the lock into the CryptoProviderCache class, and expose a new method to provide this capability. ConcurrentDictionary demonstrates a good pattern for this scenario via the [GetOrAdd()](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=netframework-4.5#System_Collections_Concurrent_ConcurrentDictionary_2_GetOrAdd__0_System_Func__0__1__) function. It either returns a known value or uses the value specified by the caller instead. It includes two overloads, one with an explicit object reference and another that takes a Func<> for the cases where the object is expensive to create.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -383,22 +384,34 @@ private SignatureProvider CreateSignatureProvider(SecurityKey key, string algori
...
-            else
-                signatureProvider = new SymmetricSignatureProvider(key, algorithm, willCreateSignatures);
+                    if (createAsymmetric)
+                        signatureProvider = new AsymmetricSignatureProvider(key, algorithm, willCreateSignatures);
+                    else
+                        signatureProvider = new SymmetricSignatureProvider(key, algorithm, willCreateSignatures);
 
-            if (CacheSignatureProviders)
-                CryptoProviderCache.TryAdd(signatureProvider);
+                    if (CacheSignatureProviders)
```
CBaud: This check is redundant.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: @CBaud what you point out is a better model because we are locking from the outside of the cache. As soon as something other than SignatureProviders are cached OR the cache is shared between types, we will see a problem.

Not sure if we should address this in 5.2.5.
For sure as soon as we cache other crypto operators, such as KeyWrapProvider, AuthenticatedEncryptionProviders, we would smart to change this design.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -383,22 +384,34 @@ private SignatureProvider CreateSignatureProvider(SecurityKey key, string algori
...
+                        CryptoProviderCache.TryAdd(signatureProvider);
+                }
+            } else
+            {
+                if (createAsymmetric)
```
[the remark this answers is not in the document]
brentschmaltz: @mafurman I would like to see the 'else' on a separate line.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
mafurman: Thanks for noticing this! Just fixed this.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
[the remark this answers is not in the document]
mafurman: Just made both of these changes. 👍

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
mafurman: I just added an issue for moving the lock into the CryptoProviderCache: #1004
It's currently set for the 5.2.6 release.

Issue #995, opened 2018-08-15
Add support for creating EncryptedCredentials using X509Certificate2


GeoK:
**Note:**
- If case when only a certificate is provided, key wrap encryption algorithm and data encryption algorithm will be set by default to RsaOaepKeyWrap and A128CBC-HS256, respectively.

See also: #734

Issue #734, opened 2017-09-15
Add support for Saml2 EncryptedAssertion


brentschmaltz:
Saml2 tokens can have encrypted assertions, we didn't get them into 5.2.0.

Veikedo:
Up
ps. Saml1 token also can be encrypted

Extant-1:
+1

This feature would be very much welcome to avoid having to write a custom TokenValidator

brentschmaltz:
@Veikedo @Extant-1 I marked this for 5.2.3 (June 2018). It may be a stretch, I'll cost it out.

brentschmaltz:
Moved to 5.2.4

brentschmaltz:
No need for this now, removing until needed.

PR #1000, opened 2018-08-16
Add X509EncryptedCredentials class

GeoK:
If only a certificate is provided, key wrap encryption algorithm
and data encryption algorithm will be set by default to RsaOaepKeyWrap
and A128CBC-HS256, respectively.

* Add new ctor to EncryptingCredentials to allow users to pass
only a 'shared' symmetric key which will be used to encrypt data, but
it will not be serialized to a SAML token.

* Add internal const strings DefaultAsymmetricKeyWrapAlgorithm and
DefaultSymmetricAlgorithm to indicate the default algorithms used for
key wrap and data encryption

* Add protected ctor to EncryptingCredentials to check if a certificate
passed to X509EncryptedCredentials is null. Provides cleaner stack
trace in case of an exception caused by a null cert.

* Refactor EncryptingCredentials. Move null/empty checks to setters
and provide clearer comments

* Add tests for X509EncryiptingCredentials and EncryptingCredentials
classes

Resolves: #995
See also: #734

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
+++ b/src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
@@ -25,67 +25,102 @@
 //
 //------------------------------------------------------------------------------
 
+using System;
+using System.Security.Cryptography.X509Certificates;
 using Microsoft.IdentityModel.Logging;
 
 namespace Microsoft.IdentityModel.Tokens
 {
     /// <summary>
-    /// A wrapper class for properties that are used for token encryption.
+    /// A class for properties that are used for token encryption.
     /// </summary>
     public class EncryptingCredentials
     {
+        private string _alg;
+        private string _enc;
+        private SecurityKey _key;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="EncryptingCredentials"/> class.
         /// </summary>
-        /// <param name="key"><see cref="SecurityKey"/></param>
-        /// <param name="alg">The key encryption algorithm to apply.</param>
-        /// <param name="enc">The encryption algorithm to apply.</param>
-        public EncryptingCredentials(SecurityKey key, string alg, string enc)
+        /// <param name="certificate"><see cref="X509Certificate2"/>.</param>
+        /// <param name="alg">A key wrap algorithm to use when encrypting a session key.</param>
+        /// <param name="enc">Data encryption algorithm to apply.</param>
+        /// <exception cref="ArgumentNullException">if 'certificate' is null.</exception>
+        /// <exception cref="ArgumentNullException">if 'alg' is null or empty.</exception>
+        /// <exception cref="ArgumentNullException">if 'enc' is null or empty.</exception>
+        protected EncryptingCredentials(X509Certificate2 certificate, string alg, string enc)
         {
-            if (key == null)
-                throw LogHelper.LogArgumentNullException(nameof(key));
+            if (certificate == null)
+                throw LogHelper.LogArgumentNullException(nameof(certificate));
 
-            if (string.IsNullOrWhiteSpace(alg))
-                throw LogHelper.LogArgumentNullException(nameof(alg));
-
-            if (string.IsNullOrWhiteSpace(enc))
-                throw LogHelper.LogArgumentNullException(nameof(enc));
-
+            Key = new X509SecurityKey(certificate);
             Alg = alg;
             Enc = enc;
-            Key = key;
         }
 
         /// <summary>
-        /// Gets the algorithm which used for token encryption.
+        /// Initializes a new instance of the <see cref="EncryptingCredentials"/> class.
+        /// </summary>
+        /// <param name="key"><see cref="SecurityKey"/> to use when encrypting a session key.</param>
+        /// <param name="alg">A key wrap algorithm to use when encrypting a session key.</param>
+        /// <param name="enc">Data encryption algorithm to apply.</param>
+        /// <exception cref="ArgumentNullException">if 'key' is null.</exception>
+        /// <exception cref="ArgumentNullException">if 'alg' is null or empty.</exception>
+        /// <exception cref="ArgumentNullException">if 'enc' is null or empty.</exception>
+        public EncryptingCredentials(SecurityKey key, string alg, string enc)
+        {
+            Key = key;
+            Alg = alg;
+            Enc = enc;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EncryptingCredentials"/> class.
+        /// </summary>
+        /// <remarks> Used in scenarios when a key represents a 'shared' symmetric key.
+        /// For example, SAML 2.0 Assertion will be encrypted using a provided symmetric key
+        /// which won't be serialized to a SAML token.
+        /// </remarks>
+        /// <param name="key"><see cref="SymmetricSecurityKey"/> to apply.</param>
+        /// <param name="enc">Data encryption algorithm to apply.</param>
+        /// <exception cref="ArgumentException">If the <see cref="SecurityKey"/> is not a <see cref="SymmetricSecurityKey"/>.</exception>
+        /// <exception cref="ArgumentNullException">if 'enc' is null or empty.</exception>
+        public EncryptingCredentials(SymmetricSecurityKey key, string enc)
+            : this(key, SecurityAlgorithms.None, enc)
+        {
+        }
+
+        /// <summary>
+        /// Gets the key wrap algorithm used for session key encryption.
         /// </summary>
         public string Alg
         {
-            get;
-            private set;
+            get => _alg;
+            private set => _alg = string.IsNullOrEmpty(value) ? throw LogHelper.LogArgumentNullException("alg") : value;
         }
 
         /// <summary>
-        /// Gets the algorithm which used for token encryption.
+        /// Gets the data encryption algorithm.
         /// </summary>
         public string Enc
         {
-            get;
-            private set;
+            get => _enc;
+            private set => _enc = string.IsNullOrEmpty(value) ? throw LogHelper.LogArgumentNullException("enc") : value;
         }
 
         /// <summary>
-        /// Users can override the default <see cref="CryptoProviderFactory"/> with this property. This factory will be used for creating encryition providers.
+        /// Users can override the default <see cref="CryptoProviderFactory"/> with this property. This factory will be used for creating encryption providers.
         /// </summary>
         public CryptoProviderFactory CryptoProviderFactory { get; set; }
 
         /// <summary>
-        /// Gets the <see cref="SecurityKey"/> which used for signature valdiation.
+        /// Gets the <see cref="SecurityKey"/> used for encryption.
         /// </summary>
         public SecurityKey Key
         {
-            get;
-            private set;
+            get => _key;
+            private set => _key = value ?? throw LogHelper.LogArgumentNullException("key");
         }
     }
 }
--- a/src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
+++ b/src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
@@ -113,6 +113,9 @@ namespace Microsoft.IdentityModel.Tokens
         public const string Aes192CbcHmacSha384 = "A192CBC-HS384";
         public const string Aes256CbcHmacSha512 = "A256CBC-HS512";
 
+        internal const string DefaultAsymmetricKeyWrapAlgorithm = RsaOaepKeyWrap;
+        internal const string DefaultSymmetricEncryptionAlgorithm = Aes128CbcHmacSha256;
+
 #pragma warning restore 1591
     }
 }
--- a/src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
+++ b/src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
@@ -1,0 +1,76 @@
+﻿//------------------------------------------------------------------------------
+//
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+using System;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// An <see cref="X509EncryptingCredentials"/> designed to construct <see cref="EncryptingCredentials"/> based on a x509 certificate.
+    /// </summary>
+    public class X509EncryptingCredentials : EncryptingCredentials
+    {
+        /// <summary>
+        /// Designed to construct <see cref="EncryptingCredentials"/> based on a x509 certificate.
+        /// </summary>
+        /// <param name="certificate">A <see cref="X509Certificate2"/></param>
+        /// <remarks>
+        /// <see cref="SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm"/> will be used as the key wrap algorithm
+        /// <see cref="SecurityAlgorithms.DefaultSymmetricEncryptionAlgorithm"/> will be used as the data encryption algorithm
+        /// </remarks>
+        /// <exception cref="ArgumentNullException">if 'certificate' is null.</exception>
+        public X509EncryptingCredentials(X509Certificate2 certificate)
+            : this(certificate, SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm, SecurityAlgorithms.DefaultSymmetricEncryptionAlgorithm)
+        {
+        }
+
+        /// <summary>
+        /// Designed to construct <see cref="EncryptingCredentials"/> based on the x509 certificate, a key wrap algorithm, and data encryption algorithm.
+        /// </summary>
+        /// <param name="certificate">A <see cref="X509Certificate2"/></param>
+        /// <param name="keyWrapAlgorithm">A key wrap algorithm</param>
+        /// <param name="dataEncryptionAlgorithm">Data encryption algorithm</param>
+        /// <exception cref="ArgumentNullException">if 'certificate' is null.</exception>
+        /// <exception cref="ArgumentNullException">if 'keyWrapAlgorithm' is null or empty.</exception>
+        /// <exception cref="ArgumentNullException">if 'dataEncryptionAlgorithm' is null or empty.</exception>
+        public X509EncryptingCredentials(X509Certificate2 certificate, string keyWrapAlgorithm, string dataEncryptionAlgorithm)
+            : base(certificate, keyWrapAlgorithm, dataEncryptionAlgorithm)
+        {
+            Certificate = certificate;
+        }
+
+        /// <summary>
+        /// Gets the <see cref="X509Certificate2"/> used by this instance.
+        /// </summary>
+        public X509Certificate2 Certificate
+        {
+            get;
+            private set;
+        }
+    }
+}
--- a/test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
+++ b/test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
@@ -1,0 +1,174 @@
+﻿//------------------------------------------------------------------------------
+//
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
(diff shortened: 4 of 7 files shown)


src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
```diff
@@ -25,67 +25,109 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EncryptingCredentials"/> class.
+        /// </summary>
+        /// <remarks> Used in scenarios when a key represents a 'shared' symmetric key.
+        /// For example, SAML 2.0 Assertion will be encrypted using a provided symmetric key
+        /// which won't be serialized to a SAML token.
+        /// </remarks>
+        /// <param name="key"><see cref="SecurityKey"/></param>
+        /// <param name="enc">Data encryption algorithm to apply when encrypting plaintext.</param>
+        /// <exception cref="ArgumentException">If the <see cref="SecurityKey"/> is not <see cref="SymmetricSecurityKey"/>.</exception>
```
mafurman: I would put an "a" right before SymmetricSecurityKey for readability.

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
```diff
@@ -25,67 +25,109 @@
...
-            Alg = alg;
+            //explicitly setting Alg to None
+            Alg = SecurityAlgorithms.None;
             Enc = enc;
-            Key = key;
         }
 
         /// <summary>
-        /// Gets the algorithm which used for token encryption.
+        /// Gets the key wrap encryption algorithm used for a session key encryption.
```
mafurman: Get rid of the "a" before session key in this comment.

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
```diff
@@ -25,67 +25,109 @@
...
         /// </summary>
         public string Alg
         {
-            get;
-            private set;
+            get => _alg;
+            private set => _alg = string.IsNullOrEmpty(value) ? throw LogHelper.LogArgumentNullException("alg") : value;
         }
```
mafurman: Maybe pass in nameof(value) instead of "alg" to LogArgumentNullException()?

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
```diff
@@ -25,67 +25,109 @@
...
         /// <summary>
-        /// Users can override the default <see cref="CryptoProviderFactory"/> with this property. This factory will be used for creating encryition providers.
+        /// Users can override the default <see cref="CryptoProviderFactory"/> with this property. This factory will be used for creating encryption providers.
         /// </summary>
         public CryptoProviderFactory CryptoProviderFactory { get; set; }
 
         /// <summary>
-        /// Gets the <see cref="SecurityKey"/> which used for signature valdiation.
+        /// Gets the <see cref="SecurityKey"/> which used for signature validation.
         /// </summary>
```
mafurman: Get rid of the "which" in this comment.

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -0,0 +1,77 @@
...
+//
+//------------------------------------------------------------------------------
+
+using System;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// An <see cref="X509EncryptingCredentials"/> designed for to construct an <see cref="EncryptingCredentials"/> based on a x509 certificate.
```
mafurman: Either change this to "designed for constructing EncryptingCredentials" or "designed to construct EncryptingCredentials"

Maybe add in <see cref="X509Certificate2"/>?

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    public class X509EncryptingCredentials : EncryptingCredentials
+    {
+        /// <summary>
+        /// Constructs an <see cref="EncryptingCredentials"/> based on a x509 certificate.
```
mafurman: Get rid of "an"

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// </summary>
+        /// <param name="certificate">A <see cref="X509Certificate2"/></param>
+        /// <remarks>
+        /// <see cref="SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm"/> algorithm will be used by default as a key wrap encryption algorithm
+        /// <see cref="SecurityAlgorithms.DefaultSymmetricAlgorithm"/> algorithm will be used by default as a data encryption algorithm
```
mafurman: Change to "The algorithm that will be used"

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// Constructs an <see cref="EncryptingCredentials"/> based on the x509 certificate, key wrapping algorithm, and data encryption algorithm.
+        /// </summary>
+        /// <param name="certificate">A <see cref="X509Certificate2"/></param>
+        /// <param name="keyWrapEncryptionAlgorithm">A key wrapping algorithm</param>
+        /// <param name="dataEncryptionAlgorithm">A data encryption algorithm</param>
+        /// <exception cref="ArgumentNullException">if 'certificate' is null.</exception>
+        /// <exception cref="ArgumentNullException">if 'keyWrapEncryptionAlgorithm' is null or empty.</exception>
+        /// <exception cref="ArgumentNullException">if 'dataEncryptionAlgorithm' is null or empty.</exception>
+        public X509EncryptingCredentials(X509Certificate2 certificate, string keyWrapEncryptionAlgorithm, string dataEncryptionAlgorithm)
+            : base(new X509SecurityKey(certificate), keyWrapEncryptionAlgorithm, dataEncryptionAlgorithm)
```
mafurman: Change this to use the constructor that takes in a certificate and not a key

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
```diff
@@ -0,0 +1,146 @@
...
+//
+//------------------------------------------------------------------------------
+
+using System;
+using Microsoft.IdentityModel.Tests;
+using Xunit;
+
+namespace Microsoft.IdentityModel.Tokens.Tests
+{
+    public class EncryptingCredentialsTests
```
mafurman: I would consider refactoring this test to use TheoryData if possible. I think the OpenIdConnectMessageTests.Constructors() test is similar to what you would need to do here.

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
GeoK: Not sure if it makes sense as the resulting log message would look something like this:
*IDX10000: The parameter **value** cannot be a 'null' or an empty object.*

Using reflection, we can get names of parameters from a constructor but I don't believe that is a good solution.

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
GeoK: Good catch!

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
```diff
@@ -25,67 +25,109 @@
...
+        /// <param name="enc">Data encryption algorithm to apply when encrypting plaintext.</param>
+        /// <exception cref="ArgumentException">If the <see cref="SecurityKey"/> is not <see cref="SymmetricSecurityKey"/>.</exception>
+        /// <exception cref="ArgumentNullException">if 'enc' is null or empty.</exception>
+        public EncryptingCredentials(SecurityKey key, string enc)
+        {
+            Key = key;
 
-            if (string.IsNullOrWhiteSpace(enc))
-                throw LogHelper.LogArgumentNullException(nameof(enc));
+            if (key.GetType() != typeof(SymmetricSecurityKey))
```
CBaud: Instead of checking for SymmetricSecurityKey explicitly, it would be good to enforce this contract by allowing any class that derives from SymmetricSecurityKey to be accepted. This could be accomplished using the 'is' keyword or by calling the Type.IsAssignableFrom() method.

Bonus points if we can enforce this contract using an interface instead. The current implementation is not extensible to classes that do not inherit from SymmetricSecurityKey. (e.g. a cloud-based symmetric key from a service like Key Vault) An interface could similarly be used to determine if the key derives from SymmetricSecurityKey (if that class implments the specified interface) and would also make this class extensible to any other security key that implements the specified contract.

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
GeoK: Will do - it's much cleaner :)

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
GeoK: Great input. I decided to use *is* as *Type.IsAssignableFrom()* is not supported in netstandard1.4
Good point w.r.t. an interface, will think about that.

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            TestUtilities.AssertFailIfErrors(context);
+        }
+
+        public static TheoryData<EncryptingCredentialsTheoryData> ConstructorATheoryData()
+        {
+            return new TheoryData<EncryptingCredentialsTheoryData>
+            {
+                new EncryptingCredentialsTheoryData
+                {
```
brentschmaltz: We don't need this case.

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
```diff
@@ -0,0 +1,198 @@
...
+                    Key = null,
+                    Alg = null,
+                    Enc = null,
+                    ExpectedException = ExpectedException.ArgumentNullException("IDX10000:"),
+                    TestId = "null key"
```
brentschmaltz: Best not to have spaces ' a test name ', underscores '_', periods '.' in a TestId. This makes is harder to set a conditional break point when the test fails. Try double clicking on a TestId.

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
```diff
@@ -0,0 +1,198 @@
...
+                },
+                new EncryptingCredentialsTheoryData
+                {
+                    Key = null,
+                    Alg = SecurityAlgorithms.RsaOaepKeyWrap,
+                    Enc = SecurityAlgorithms.Aes128CbcHmacSha256,
+                    ExpectedException = ExpectedException.ArgumentNullException("IDX10000:"),
```
brentschmaltz: It is possible to check the exception contains "enc", this catches typo's in the error message.

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            TestUtilities.AssertFailIfErrors(context);
+        }
+
+        //Used in scenarios when a key represents a 'shared' symmetric key
+        //public EncryptingCredentials(SecurityKey key, string enc)
+        [Theory, MemberData(nameof(ConstructorBTheoryData))]
+        public void ConstructorB(EncryptingCredentialsTheoryData theoryData)
+        {
+            TestUtilities.WriteHeader($"{this}.ConstructorB", theoryData);
```
brentschmaltz: TestUtilities.WriteHeader returns a CompareContext, so you can write:
var context = TestUtilities.WriteHeader($"{this}.ConstructorB", theoryData);

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            var context = new CompareContext($"{this}.ConstructorB, {theoryData.TestId}");
+            try
+            {
+                var encryptingCredentials = new EncryptingCredentials(theoryData.Key, theoryData.Enc);
+                //Algorithm value should be 'None'
+                IdentityComparer.AreEqual(encryptingCredentials.Alg, SecurityAlgorithms.None, context);
+                theoryData.ExpectedException.ProcessNoException();
```
brentschmaltz: Pass context to ProcessNoException.

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            }
+            catch (Exception exception)
+            {
+                theoryData.ExpectedException.ProcessException(exception);
```
brentschmaltz: Pass context to ProcessException.
theoryData.ExpectedException.ProcessException(exception, context);

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
```diff
@@ -0,0 +1,198 @@
...
+
+namespace Microsoft.IdentityModel.Tokens.Tests
+{
+    public class EncryptingCredentialsTests
+    {
+        //public EncryptingCredentials(SecurityKey key, string alg, string enc)
+        [Theory, MemberData(nameof(ConstructorATheoryData))]
+        public void ConstructorA(EncryptingCredentialsTheoryData theoryData)
+        {
+            TestUtilities.WriteHeader($"{this}.ConstructorA", theoryData);
```
brentschmaltz: See comments below.

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <param name="certificate">A <see cref="X509Certificate2"/></param>
+        /// <remarks>
+        /// <see cref="SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm"/> The algorithm that will be used by default as a key wrap encryption algorithm
+        /// <see cref="SecurityAlgorithms.DefaultSymmetricAlgorithm"/> The algorithm that will be used by default as a data encryption algorithm
+        /// </remarks>
+        /// <exception cref="ArgumentNullException">if 'certificate' is null.</exception>
+        public X509EncryptingCredentials(X509Certificate2 certificate)
+            : this(certificate, SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm, SecurityAlgorithms.DefaultSymmetricAlgorithm)
+        {
+
```
brentschmaltz: can remove this line.

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <exception cref="ArgumentNullException">if 'keyWrapEncryptionAlgorithm' is null or empty.</exception>
+        /// <exception cref="ArgumentNullException">if 'dataEncryptionAlgorithm' is null or empty.</exception>
+        public X509EncryptingCredentials(X509Certificate2 certificate, string keyWrapEncryptionAlgorithm, string dataEncryptionAlgorithm)
+            : base(certificate, keyWrapEncryptionAlgorithm, dataEncryptionAlgorithm)
+        {
+            Certificate = certificate;
+        }
+
+        /// <summary>
+        /// Gets a <see cref="X509Certificate2"/> used by this instance.
```
brentschmaltz: Gets 'the', instead Gets 'a'

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    public class X509EncryptingCredentials : EncryptingCredentials
+    {
+        /// <summary>
+        /// Designed to construct <see cref="EncryptingCredentials"/> based on a x509 certificate.
+        /// </summary>
+        /// <param name="certificate">A <see cref="X509Certificate2"/></param>
+        /// <remarks>
+        /// <see cref="SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm"/> The algorithm that will be used by default as a key wrap encryption algorithm
+        /// <see cref="SecurityAlgorithms.DefaultSymmetricAlgorithm"/> The algorithm that will be used by default as a data encryption algorithm
```
brentschmaltz: suggest:
/// <see cref="SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm"/> will be used as the key wrap algorithm
/// <see cref="SecurityAlgorithms.DefaultSymmetricAlgorithm"/> will be used by as the data encryption algorithm

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <summary>
+        /// Designed to construct <see cref="EncryptingCredentials"/> based on the x509 certificate, key wrapping algorithm, and data encryption algorithm.
+        /// </summary>
+        /// <param name="certificate">A <see cref="X509Certificate2"/></param>
+        /// <param name="keyWrapEncryptionAlgorithm">A key wrapping algorithm</param>
+        /// <param name="dataEncryptionAlgorithm">A data encryption algorithm</param>
+        /// <exception cref="ArgumentNullException">if 'certificate' is null.</exception>
+        /// <exception cref="ArgumentNullException">if 'keyWrapEncryptionAlgorithm' is null or empty.</exception>
+        /// <exception cref="ArgumentNullException">if 'dataEncryptionAlgorithm' is null or empty.</exception>
+        public X509EncryptingCredentials(X509Certificate2 certificate, string keyWrapEncryptionAlgorithm, string dataEncryptionAlgorithm)
```
brentschmaltz: suggest:
keyWrapEncryptionAlgorithm -> keyWrapAlgorithm

KeyWrap always implies some sort of encryption.

src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
```diff
@@ -113,6 +113,9 @@ public static class SecurityAlgorithms
         public const string Aes192CbcHmacSha384 = "A192CBC-HS384";
         public const string Aes256CbcHmacSha512 = "A256CBC-HS512";
 
+        internal const string DefaultAsymmetricKeyWrapAlgorithm = RsaOaepKeyWrap;
+        internal const string DefaultSymmetricAlgorithm = Aes128CbcHmacSha256;
```
brentschmaltz: suggest:
DefaultSymmetricAlgorithm -> DefaultSymmetricEncryptionAlgorithm

there are other SymmetricAlgorithms such as Signing.

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
```diff
@@ -25,67 +25,109 @@
...
+
         /// <summary>
         /// Initializes a new instance of the <see cref="EncryptingCredentials"/> class.
         /// </summary>
-        /// <param name="key"><see cref="SecurityKey"/></param>
-        /// <param name="alg">The key encryption algorithm to apply.</param>
-        /// <param name="enc">The encryption algorithm to apply.</param>
-        public EncryptingCredentials(SecurityKey key, string alg, string enc)
+        /// <param name="certificate"><see cref="X509Certificate2"/>.</param>
+        /// <param name="alg">A key wrap encryption algorithm to apply when encrypting a session key.</param>
```
brentschmaltz: suggest:
A key wrap encryption algorithm -> A key wrap algorithm

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
```diff
@@ -25,67 +25,109 @@
...
+        /// <param name="enc">Data encryption algorithm to apply when encrypting plaintext.</param>
```
brentschmaltz: suggest:
Data encryption algorithm to apply when encrypting plaintext ->
The Data encryption algorithm to apply

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
```diff
@@ -25,67 +25,109 @@
...
-                throw LogHelper.LogArgumentNullException(nameof(enc));
+            Key = new X509SecurityKey(certificate);
+            Alg = alg;
+            Enc = enc;
+        }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EncryptingCredentials"/> class.
+        /// </summary>
+        /// <param name="key"><see cref="SecurityKey"/> to apply when encrypting a session key.</param>
```
brentschmaltz: suggest:
<see cref="SecurityKey"/> to apply when encrypting a session key ->
<see cref="SecurityKey"/> to use when encrypting a session key

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
GeoK: I'll remove it from the test case.

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
GeoK: Got it - I'll replace TestId values.

test/Microsoft.IdentityModel.Tokens.Tests/EncryptingCredentialsTests.cs
GeoK: I'm going to use values in the following format as expected substrings:
*"IDX10000: The parameter enc"*

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
GeoK: it makes sense - I'll rename.

src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
GeoK: it makes sense - I'll rename.

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
GeoK: Done.

GeoK:
@brentschmaltz Thank you for the suggestions, I addressed them with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/7578fe64d6bb4efd9f7fa0b5c074872fcee9ad4c

brentschmaltz:
:shipit:

PR #1007, opened 2018-08-23
Support JWE with "zip" in header

### LINKED ISSUE #593: Support JWE with "zip" in header

mafurman:
Fixes #593.

Supported for JWEs containing a 'zip' header parameter has now been added to the JsonWebTokenHandler.


src/Microsoft.IdentityModel.Tokens/CompressionAlgorithms.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    /// <summary>
+    /// Constants for compression algorithms.
+    /// </summary>
+    public class CompressionAlgorithms
+    {
+#pragma warning disable 1591
+        // See: https://tools.ietf.org/html/rfc1951
+        public const string Deflate = "DEF";
+    }
```
CBaud: Should warning 1591 be restored here?

src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
```diff
@@ -0,0 +1,119 @@
...
+        public static CompressionProviderFactory Default
+        {
+            get => _default;
+            set => _default = value ?? throw LogHelper.LogArgumentNullException(nameof(Default));
+        }
+
+        /// <summary>
+        /// Extensibility point for custom compression support application wide.
+        /// </summary>
+        public CompressionProvider CustomCompressionProvider { get; set; }
```
CBaud: If you'd like this to be extensible, then it would be good to create an ICompressionProvider interface and declare this property as that type instead.

src/Microsoft.IdentityModel.Tokens/DeflateCompressionProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// </summary>
+        /// <param name="value">the bytes to decompress.</param>
+        /// <returns>the decompressed string.</returns>
+        public override string Decompress(byte[] value)
+        {
+            if (value == null)
+                throw LogHelper.LogArgumentNullException(nameof(value));
+
+            using (var inputStream = new MemoryStream(value))
+            {
```
CBaud: [Minor] Isn't the coding style of this library to not use braces when it encapsulates a single line of code. You can chain these together using that pattern. For example:

using (var inputStream = ...)
using (var deflateStream = ...)
using (reader = ...)
    return reader.ReadToEnd();

src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            if (CustomCompressionProvider != null && CustomCompressionProvider.IsSupportedAlgorithm(algorithm))
+                return true;
+
+            return IsSupportedCompressionAlgorithm(algorithm);
+        }
+
+        private bool IsSupportedCompressionAlgorithm(string algorithm)
+        {
+            return CompressionAlgorithms.Deflate.Equals(algorithm);
```
CBaud: It would be nice to specify a string comparison algorithm here. For example, StringComparison.Ordinal or StringComparison.OrdinalIgnoreCase depending on whether or not you care about case sensitivity. If you don't, Ordinal has significantly better performance.

src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        /// <summary>
+        /// Returns a <see cref="CompressionProvider"/> for a specific algorithm.
+        /// </summary>
+        /// <param name="algorithm">the decompression algorithm.</param>
+        /// <returns>a <see cref="CompressionProvider"/>.</returns>
+        public CompressionProvider CreateCompressionProvider(string algorithm)
```
CBaud: If we decide to create an ICompressionProvider interface, it would be great to use that to specify the return type for this function. This would help to decouple the contract of this method from the underlying implementation.

src/Microsoft.IdentityModel.Tokens/DeflateCompressionProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    /// <summary>
+    /// A compression provider that supports compression and decompression using the <see cref="CompressionAlgorithms.Deflate"/> algorithm.
+    /// </summary>
+    public class DeflateCompressionProvider : CompressionProvider
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DeflateCompressionProvider"/> class used to compress and decompress used the <see cref="CompressionAlgorithms.Deflate"/> algorithm.
+        /// </summary>
+        public DeflateCompressionProvider()
```
CBaud: I would encourage you to review the [DeflateStream](https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.deflatestream) documentation to fully understand the default options it is using. I think it could be useful to provide a property or constructor input parameter that enables the consumer of this library to specify the desired CompressionLevel. This would enable them to decide if they care more about speed or optimization of the compression being performed.

src/Microsoft.IdentityModel.Tokens/CompressionAlgorithms.cs
mafurman: Yes, it should. Thanks for noticing this!

src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
mafurman: Done. I created an ICompressionProvider interface.

src/Microsoft.IdentityModel.Tokens/DeflateCompressionProvider.cs
mafurman: We don't use braces when using single line 'if' statements, but I don't think this rule applies to 'using' statements.

src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
mafurman: I've changed this line to use StringComparison.Ordinal.

src/Microsoft.IdentityModel.Tokens/CompressionProviderFactory.cs
mafurman: Done.

src/Microsoft.IdentityModel.Tokens/DeflateCompressionProvider.cs
mafurman: This is a good idea! 👍 It looks like by default CompressionLevel is set to Optimal when the CompressionMode is Compress. I added in another constructor that allows a user to specify the desired compression level if they would like to.

mafurman:
@CBaud @GeoK @brentschmaltz What do you think about changing the Compress() and Decompress() methods on the ICompressionProvider interface to take in byte[] and return byte[]?

This is what they look like currently:

        /// <summary>
        /// Decompress.
        /// </summary>
        /// <param name="value">the value to decompress.</param>
        string Decompress(byte[] value);

        /// <summary>
        /// Compress.
        /// </summary>
        /// <param name="value">the value to decompress.</param>
        byte[] Compress(string value);

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -134,10 +134,35 @@ public string CreateToken(JObject payload, SigningCredentials signingCredentials
...
+            if (payload == null)
+                throw LogHelper.LogArgumentNullException(nameof(payload));
+
+            if (signingCredentials == null)
+                throw LogHelper.LogArgumentNullException(nameof(signingCredentials));
+
+            if (encryptingCredentials == null)
+                throw LogHelper.LogArgumentNullException(nameof(encryptingCredentials));
+
+            if (compressionAlgorithm == null)
```
brentschmaltz: check: if (string.IsNullOrEmpty(...))

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -170,11 +195,61 @@ private string CreateTokenPrivate(JObject payload, SigningCredentials signingCre
...
+            var compressionProvider = CompressionProviderFactory.Default.CreateCompressionProvider(compressionAlgorithm);
+        
+            return compressionProvider.Compress(Encoding.UTF8.GetBytes(token)) ?? throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10680, compressionAlgorithm)));
+        }
+
+        /// <summary>
+        /// Decompress JWT token bytes.
+        /// </summary>
+        /// <param name="tokenBytes"></param>
+        /// <param name="decompressionAlgorithm"></param>
```
brentschmaltz: I would suggest to use 'algorithm', this is similar to our SignatureProviders.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -170,11 +195,61 @@ private string CreateTokenPrivate(JObject payload, SigningCredentials signingCre
...
+        /// <exception cref="ArgumentNullException">if 'tokenBytes' is null.</exception>
+        /// <exception cref="ArgumentNullException">if 'decompressionAlgorithm' is null.</exception>
+        /// <exception cref="NotSupportedException">if the decompression algorithm is not supported.</exception>
+        /// <returns>Decompressed JWT token</returns>
+        protected string DecompressToken(byte[] tokenBytes, string decompressionAlgorithm)
```
brentschmaltz: Would it make sense to have this return byte[], similar to the provider

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -184,6 +259,7 @@ private string CreateTokenPrivate(JObject payload, SigningCredentials signingCre
         /// <exception cref="ArgumentNullException">if 'jwtToken' is null.</exception>
         /// <exception cref="ArgumentNullException">if 'validationParameters' is null.</exception>
         /// <exception cref="SecurityTokenException">if 'jwtToken.Enc' is null or empty.</exception>
+        /// <exception cref="SecurityTokenException">if decompression failed.</exception>
```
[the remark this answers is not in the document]
mafurman: Good suggestion. 👍  I've added those two exception types in.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: I've changed the methods to be private.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Makes sense. I've changed all instances of 'compressionAlgorithm' and 'decompressionAlgorithm' to just be 'algorithm'.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs

PR #1008, opened 2018-08-25
Got rid of any public references to JObject in Microsoft.IdentityModel.JsonWebTokens

mafurman:


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -74,16 +73,16 @@ public JsonWebToken(string jwtEncodedString)
...
+        /// <param name="payload">A string containing JSON which represents the claims contained in the JWT. Each claim is a JSON object of the form { Name, Value }.</param>
         /// <exception cref="ArgumentNullException">'header' is null.</exception>
         /// <exception cref="ArgumentNullException">'payload' is null.</exception>
-        public JsonWebToken(JObject header, JObject payload)
+        public JsonWebToken(string header, string payload)
         {
-            Header = header ?? throw LogHelper.LogArgumentNullException(nameof(header));
-            Payload = payload ?? throw LogHelper.LogArgumentNullException(nameof(payload));
+            Header = JObject.Parse(header) ?? throw LogHelper.LogArgumentNullException(nameof(header));
+            Payload = JObject.Parse(payload) ?? throw LogHelper.LogArgumentNullException(nameof(payload));
```
CBaud: These null coalescing operators are likely dead code now. The JObject.Parse() method throws a JsonReaderException exception if the input contains invalid json. Otherwise, I would expect that it always returns a valid JObject reference.

mafurman: You're right. 👍 This isn't the right place for the null checks. I've changed this around so that the null checks are actually done on the header and payload parameters.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -74,16 +73,22 @@ public JsonWebToken(string jwtEncodedString)
...
         {
-            Header = header ?? throw LogHelper.LogArgumentNullException(nameof(header));
-            Payload = payload ?? throw LogHelper.LogArgumentNullException(nameof(payload));
+            if (string.IsNullOrEmpty(header))
+                throw LogHelper.LogArgumentNullException(nameof(header));
+
+            if (string.IsNullOrEmpty(payload))
+                throw LogHelper.LogArgumentNullException(nameof(payload));
+
+            Header = JObject.Parse(header);
```
GeoK: Would it make sense to capture an exception here and log & re-throw _our_ exception, in case that provided json string is invalid?  _JObject.Parse_ throws _JsonReaderException_ when an error occurs during parsing.
Same applies for _Payload_.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -151,7 +156,7 @@ public virtual IEnumerable<Claim> Claims
                 {
                     if (entry.Value == null)
                     {
-                        claims.Add(new Claim(entry.Key, string.Empty, JsonClaimValueTypes.JsonNull, issuer, issuer));
+                        claims.Add(new Claim(entry.Key, String.Empty, JsonClaimValueTypes.JsonNull, issuer, issuer));
```
brentschmaltz: this should be lower case S 'string' not 'String'

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -460,5 +465,23 @@ private static string GetClaimValueType(object obj)
...
             return objType.ToString();
         }
+
+        /// <summary>
+        /// Gets the 'value' corresponding to the provided key from the JWT payload { key, 'value' }.
+        /// </summary>
+        /// <remarks>If the key has no corresponding value, returns null. </remarks>   
+        public T GetPayloadValue<T>(string key)
+        {
+            return Payload.Value<T>(key);
```
brentschmaltz: need null check on 'key'

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -460,5 +465,23 @@ private static string GetClaimValueType(object obj)
...
+        }
+
+        /// <summary>
+        /// Gets the 'value' corresponding to the provided key from the JWT header { key, 'value' }.
+        /// </summary>
+        /// <remarks>If the key has no corresponding value, returns null. </remarks>   
+        public T GetHeaderValue<T>(string key)
+        {
+            return Header.Value<T>(key);
```
brentschmaltz: need null check on 'key'

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz: @GeoK I agree, that would be cleaner.
We throw ArgumentException IDX12723 and IDX12724 when parsing fails in JwtSecurityToken.
Are you asking for a new exception OR stick with ArgumentException?

mafurman: Good suggestion. I decided to throw an ArgumentException just to be consistent with what JwtSecurityToken does.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: Changed all occurrences of String.Empty to string.Empty in this file.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: Done. Not sure if it would be better to incline this check with the return statement or not...

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs

brentschmaltz:
:shipit:

PR #1015, opened 2018-09-12
Address NuGet compliance issues

GeoK:
Addressed _copyright_ as that is the only check that would fail validation.


brentschmaltz:
:shipit:

PR #1016, opened 2018-09-12
Fix hanging build

GeoK:
Avoid leaving MSBuild/VBCSCompiler processes around after a build.
These processes may leave Start-Process waiting and hang a build.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -99,17 +99,19 @@ CreateArtifactsRoot($artifactsRoot);
 pushd
 Set-Location $root
 Write-Host ""
-Write-Host ">>> Start-Process -wait -NoNewWindow $msbuildexe /restore:True /verbosity:m /p:Configuration=$buildType $slnFile"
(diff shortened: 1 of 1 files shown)


brentschmaltz:
:shipit: Lets give it a try

build.ps1
```diff
@@ -99,17 +99,19 @@ CreateArtifactsRoot($artifactsRoot);
 pushd
 Set-Location $root
 Write-Host ""
-Write-Host ">>> Start-Process -wait -NoNewWindow $msbuildexe /restore:True /verbosity:m /p:Configuration=$buildType $slnFile"
+Write-Host ">>> Start-Process -wait -NoNewWindow $msbuildexe /restore:True /p:UseSharedCompilation=false /nr:false /verbosity:m /p:Configuration=$buildType $slnFile"
```
brentschmaltz: what is /nr:false ?

GeoK: it turns-off node reuse. VSTS sets the same flag for Wilson build.

>Enable or disable the re-use of MSBuild nodes. You can specify the following values:
>- True. Nodes remain after the build finishes so that subsequent builds can use them (default).
>- False. Nodes don't remain after the build completes.

[source](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2017#switches)

PR #1017, opened 2018-09-12
Update strongNameByPass scripts

GeoK:
- Remove unused entries from strongNameByPass scripts.
- Add new entries to strongNameByPass scripts as some assemblies
are now delay signed.

Following assemblies are now delay signed:
* Microsoft.IdentityModel.Tokens.Tests
* Microsoft.IdentityModel.JsonWebTokens.Tests
* Microsoft.IdentityModel.Tests
* System.IdentityModel.Jwt.Tests


brentschmaltz:
:shipit: I can't see the changes as they are binary, but I'll try it out after you push it.

PR #1025, opened 2018-09-26
Saml2EncryptedAssertion

### LINKED ISSUE #734: Add support for Saml2 EncryptedAssertion

GeoK:
### Key Point:
AES-GCM is still not supported by System.Security.Cryptography. Therefore using IdentityModel extensions project to Encrypt/Decrypt a Saml2 Assertion, using an AES-GCM algorithm, will result with an exception. One should override CryptoProviderFactory and AuthenticationEncryptionProvider (similar to the ones developed for test) in order to use existing logic to Encrypt/Decrypt a Saml2 Assertion (using an AES-GCM algorithm). 
### Summary:
Saml2 tokens can have encrypted assertions. The goal of this PR is to enable basic support for encrypting and decrypting a Saml2Assertion, based on the latest xml enc standard, using modern authenticated encryption mechanism ([AES-GCM](http://www.w3.org/2009/xmlenc11#aes128-gcm)) and key wrap (transport) mechanism ([RSA-OAEP-MGF1P](http://www.w3.org/TR/xmlenc-core/#sec-RSA-OAEP)), and without introducing any breaking changes to the current project infrastructure.

The following scenarios are supported:
1) Encryption using a pre-shared symmetric key (session key)  to encrypt an assertion: Session key will not be serialized and Relying Party must be able to locally determine the decryption key.
	
2) New session key will be created to encrypt an assertion and session key will be wrapped using a provided asymmetric public-key (X.509 certificate).
	
### Saml2EncryptedAssertion support (basic profile):

* When an assertion is to be signed and encrypted, the following rules apply. A relying party MUST perform signature validation and decryption in the reverse order that signing and encryption were performed. When a signed <Assertion> element is encrypted, the signature MUST first be calculated and placed within the <Assertion> element before the element is encrypted. 
* The following authenticated encryption algorithms are supported: [AES128-GCM](http://www.w3.org/2009/xmlenc11#aes128-gcm), [AES192-GCM](http://www.w3.org/2009/xmlenc11#aes192-gcm), and [AES256-GCM](http://www.w3.org/2009/xmlenc11#aes256-gcm)
* The following key-wrap algorithms is supported: [RSA-OAEP-MGF1P](http://www.w3.org/2001/04/xmlenc#rsa-oaep-mgf1p)
* EncryptionMethod element is treated as required as there is no support for users to provide an encryption algorithm.
* If an encrypted key is NOT included in the XML instance, then the relying party must be able to locally determine the decryption key.
* EncryptedKey element (if available) must be next to associated EncryptedData element, within the enclosing Saml parent element. EncryptedKey element embedded in EncryptedData element is not currently supported.
* If the encrypted key is included in the XML instance, then it SHOULD be referenced within the associated <xenc:EncryptedData> element. When an <xenc:EncryptedKey> element is used, the <ds:KeyInfo> element within <xenc:EncryptedData> SHOULD reference the <xenc:EncryptedKey> element using a <ds:RetrievalMethod> element of Type http://www.w3.org/2001/04/xmlenc#EncryptedKey. In addition, an <xenc:EncryptedKey> element SHOULD contain an <xenc:ReferenceList> element containing a <xenc:DataReference> that references the corresponding <xenc:EncryptedData> element(s) that the key was used to encrypt. 
* CarriedKeyName (multicast) is not supported.
* Encoding attribute of EncryptedType is always set to base64 and MimeType attribute is ignored.

 ### Implementation notes:

* Current Saml2 project infrastructure is leverage to seamlessly include support for Saml2EncryptedAssertion, without introducing any breaking changes. 
* Saml2Assertion is enriched with a public bool property Encrypted which indicates if the assertion is Encrypted or not. In case when the assertion is encrypted, public string property EncryptedAssertion will be populated with raw assertion data. Any GET operation on encrypted Saml2Assertion will result with a corresponding exception being throw.
* In order to encrypt a Saml2Assertion, one needs to set EncryptingCredentials on SecurityTokenDescriptor, while creating a Saml2SecurityToken. In case when EncryptingCredentials key is an asymmetric security key, an assertion will be encrypted using a new session key (and encryption algorithm) that will be wrapped using a provided asymmetric key and an algorithm. In case when EncryptingCredentials key is a symmetric security key, an assertion will be encrypted using a pre-shared symmetric key (session key) and a provided encryption algorithm.
* In order to decrypt a Saml2EncryptedAssertion, one needs to set TokenDecryptionKey or TokenDecryptionKeyResolver delegate on TokenValidationParamaters when calling ValidateToken method.  
* New classes are added under Microsoft.IdentityModel.Xml project to support XML encryption data types used for Saml2EncryptedAssertion encryption operations.
* KeyInfo class was not fully refactored to support KeyInfo clauses as that would lead to a breaking change.
* Saml2EncryptedAssertion related logic is mostly concentrated in Saml2Serializer class.

Resolves: #734


src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -181,6 +184,7 @@ internal static class LogMessages
         public const string IDX10685 = "IDX10685: Unable to Sign, Internal SignFunction is not available.";
         public const string IDX10686 = "IDX10686: Unable to Verify, Internal VerifyFunction is not available.";
         public const string IDX10687 = "IDX10687: Unable to create a AsymmetricAdapter. For NET45 or NET451 only types: '{0}' or '{1}' are supported. RSA is of type: '{2}'..";
+        public const string IDX10688 = "IDX10688: The algorithm '{0}' is not currently supported. In order to use the algorithm '{0}' one should override CryptoProviderFactory and AuthenticatedEncryptionProvider to provide support.";
```
henrik-me: >In order to use the algorithm '{0}' one should override CryptoProviderFactory and AuthenticatedEncryptionProvider to provide support. [](start = 98, length = 133)

suggestion, replace "in order...
with:
It's possible to provide your own implementation of the algorithm by overriding CryptoProviderFactory and AuthenticatedEncryptionProvider.

You can potentially add an aka.ms link which points to the test implementation you have. #Closed

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
```diff
@@ -65,18 +65,29 @@ public AuthenticatedEncryptionProvider(SecurityKey key, string algorithm)
...
-            Key = key;
-            Algorithm = algorithm;
+                _authenticatedkeys = GetAlgorithmParameters(key, algorithm);
+                _hmacAlgorithm = GetHmacAlgorithm(algorithm);
+                _symmetricSignatureProvider = key.CryptoProviderFactory.CreateForSigning(_authenticatedkeys.HmacKey, _hmacAlgorithm) as SymmetricSignatureProvider;
+                if (_symmetricSignatureProvider == null)
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10649, Algorithm)));
+
+                Key = key;
+                Algorithm = algorithm;
```
henrik-me, line 85: suggestion: move setting key/algorithm to after if/else. This means you can make the above just an if (!isAes...)
objective: reduce nesting for readability.

I supose this algo can be considered "SupportedIfYouBringYourOwnImplementation" #Closed

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -500,6 +525,35 @@ internal static bool IsSaml2Assertion(XmlReader reader)
...
+        /// <returns>'true' if assertion is encrypted, 'false' otherwise.</returns>
+        internal static bool IsSaml2EncryptedAssertion(string assertion)
+        {
+            using (var stringReader = new StringReader(assertion))
+            {
+                var settings = new XmlReaderSettings { DtdProcessing = DtdProcessing.Prohibit };
+#if NET45 || NET451
+                settings.XmlResolver = null;
+#endif
+                using (var reader = XmlReader.Create(stringReader, settings))
```
henrik-me: >XmlReader.Create(stringReader, settings)) [](start = 36, length = 41)

Suggestion: The library could have a utility function to create this one ensuring that everyone creates the XmlReader with the appropriate (secure) XmlReaderSettings #Closed

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -500,6 +525,35 @@ internal static bool IsSaml2Assertion(XmlReader reader)
...
+                {
+                    try
+                    {
+                        reader.MoveToContent();
+                        return reader.IsStartElement(Saml2Constants.Elements.EncryptedAssertion, Saml2Constants.Namespace);
+                    }
+                    catch (Exception)
```
henrik-me: is it possible to catch only specific exceptions? Ideally some system exceptions should not be swallowed. #Pending

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
```diff
@@ -82,6 +88,200 @@ internal static bool CanCreateValidUri(string uriString, UriKind uriKind)
...
+                encryptedAssertion.EncryptedData.KeyInfo.RetrievalMethodUri = encryptingCredentials.Key.KeyId;
+
+                // Create EncryptedKey
+                encryptedAssertion.EncryptedKey = CreateEncryptedKeyHelper((AsymmetricSecurityKey)encryptingCredentials.Key, sessionKey, encryptingCredentials.Alg, encryptingCredentials);
+                // Add reference from EncrypedKey to EncryptedData
+                encryptedAssertion.EncryptedKey.AddReference(new DataReference(encryptedAssertion.EncryptedData.Id));
+
+                // Set Digest method for EncryptedKey
+                if (SecurityAlgorithms.RsaOaepMgf1pKeyWrap.Equals(encryptingCredentials.Alg, StringComparison.Ordinal))
+                    encryptedAssertion.EncryptedKey.EncryptionMethod.DigestMethod = SecurityAlgorithms.Sha1Digest;
```
henrik-me: will thinks just work if the digest method is not set? otherwise perhaps you will have to throw an exception?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
```diff
@@ -82,6 +88,200 @@ internal static bool CanCreateValidUri(string uriString, UriKind uriKind)
...
+
+                // Set X509CertificateData
+                if (encryptingCredentials.Key is X509SecurityKey)
```
henrik-me: > if (encryptingCredentials.Key is X509SecurityKey) [](start = 15, length = 50)

same though as above, what if this is not true, will the method still behave as expected? what about dependencies upstream? consider if an exception should be thrown.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
```diff
@@ -302,6 +516,28 @@ public virtual Saml2Assertion ReadAssertion(XmlReader reader)
...
+        }   
+
+        internal Saml2EncryptedAssertion ReadEncryptedAssertion(string assertion)
+        {
+            using (var stringReader = new StringReader(assertion))
+            {
+                var settings = new XmlReaderSettings { DtdProcessing = DtdProcessing.Prohibit };
+#if NET45 || NET451
+                settings.XmlResolver = null;
+#endif
```
henrik-me: as mentioned, a utility method for this and perhaps even for the variations of readers. #Closed

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
```diff
@@ -1516,7 +1752,7 @@ protected virtual void WriteAdvice(XmlWriter writer, Saml2Advice advice)
         }
 
         /// <summary>
-        /// Writes the &lt;Assertion> element.
+        /// Writes the &lt;Assertion> or &lt;EncryptedAssertion> element.
```
henrik-me, line 1550: >> [](start = 36, length = 1)

consider : gt; ? ... same towards the end of the line..

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
```diff
@@ -2250,6 +2518,89 @@ protected virtual void WriteSubjectLocality(XmlWriter writer, Saml2SubjectLocali
             writer.WriteEndElement();
         }
 
+        private EncryptedData CreateEncryptedDataHelper(SymmetricSecurityKey key, string algorithm, byte[] assertionData, EncryptingCredentials encryptingCredentials)
```
henrik-me: >private E [](start = 8, length = 9)

it seems to me like the functions for creating the helper which essentially returns Encrypted data (I'm a bit confused about that), should be in another class? Or the methods names should simply not be called Helper #Closed

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
GeoK: Good point. I'll create the link and edit the log message #Closed

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
GeoK: I'll put a comment on top of AES-GCM algo definitions to improve clarity and set expectations.
Also, I'll refactor the constructor to increase current readability. 
When we add AES-GCM support it might be better to return to *if (isAesGcm)* to separate support for AES-GCM from other authenticated encryption algorithms. #Closed

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
GeoK: agree - same settings are created more than once throughout the project. #Closed

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
GeoK: I'll think about this. The goal of this method is just to return True if first/start element of provided string is EncryptedAssertion and return False otherwise. I don't really care at that point if the string is valid XML or not.

However, I think that it definitely adds value to include exception message to log information.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
GeoK: Other algorithms may not have DigestMethod (e.g. rsa-1_5) under EncryptionMethod element. For the current implementation we are using default rsa-oaep-mgf1p parameters.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
GeoK: Just being consistent - this pattern (&lt *NODE* >) is used across the library.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
GeoK: I'll think about refactoring these. #Closed

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
GeoK: agree - same settings are created more than once throughout the project. #Closed

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
GeoK: this element is optional, but it's might be useful when when users are providing custom KeyResolving logic using TokenDecryptionKeyResolver delegate.
Current implementation is relying on using TokenDecryptionKey (set on TokenValidationParams) to decrypt an assertion, if TokenDecryptionKeyResolver delegate is not set.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
GeoK: addressed with: b9c796d459d35bb7a2d1d0f2c727d51ee2a582bc
 #Closed

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
GeoK: addressed with: c9daa02e155a336edf2c4c8cfff624ccf82277a1 and c80fbb9ae2e1d68898fe6843df0881e1a315e3a2 #Closed

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
GeoK: addressed with: bfaa792c407991dd698f5922841e4bfa03884662 #Closed

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
GeoK: addressed with: 6434db7e54ff310012f06c57624bc5eaf5b09237
 #Closed

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
GeoK: addressed with: bfaa792c407991dd698f5922841e4bfa03884662 #Closed

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
GeoK: addressed with: ac218685781cbc736b39914f3afec69f16063937
 #Closed

src/Microsoft.IdentityModel.Xml/EncryptionMethod.cs
```diff
@@ -0,0 +1,122 @@
...
+
+                if (reader.IsEmptyElement)
+                {
+                    reader.Read();
+                    return;
+                }
+
+                reader.ReadStartElement(XmlEncryptionConstants.Elements.EncryptionMethod, XmlEncryptionConstants.Namespace);
+
+                while (reader.IsStartElement())
```
henrik-me, line 104: >while (reader.IsStartElement()) [](start = 16, length = 31)

what is reader.IsSTartElement supposed to do? Reading this code I would be afraid that it would either potentially just have one element to read, however if it never reads anything this will be an endless loop

src/Microsoft.IdentityModel.Xml/XMLEncryptionConstants.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// Definitions for namespace, attributes and elements as defined in http://www.w3.org/TR/xmlenc-core1/
+    /// </summary>
+    public static class XmlEncryptionConstants
+    {
+#pragma warning disable 1591
+        public const string Namespace = "http://www.w3.org/2001/04/xmlenc#";
+
+        public const string Prefix = "xenc";
+
+        public static class Attributes
```
henrik-me, line 20: does this library use embedding of types inside of other types? If not I would have suggested moving to separate files and separate by namespace instead of using embedded types, thoughts?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -414,23 +414,23 @@ public static TheoryData<CreateTokenTheoryData> RoundTripJWEKeyWrappingTheoryDat
                         ValidationParameters = Default.TokenValidationParameters(KeyingMaterial.RsaSecurityKey_2048, Default.SymmetricSigningKey256),
                         Payload = Default.PayloadString,
                         SigningCredentials = Default.SymmetricSigningCredentials,
-                        EncryptingCredentials = new EncryptingCredentials(KeyingMaterial.RsaSecurityKey_2048, SecurityAlgorithms.RsaOaepKeyWrap, SecurityAlgorithms.Aes128CbcHmacSha256)
```
henrik-me, line 1711: >RsaOaepKeyWrap [](start = 129, length = 14)

don't you still need the old test?

test/Microsoft.IdentityModel.Tests/KeyingMaterial.cs
```diff
@@ -225,6 +225,11 @@ public static RsaSecurityKey RsaSecurityKey2
         public static byte[] DefaultSymmetricKeyBytes_128 = Convert.FromBase64String(DefaultSymmetricKeyEncoded_128);
         public static SymmetricSecurityKey DefaultSymmetricSecurityKey_128 = new SymmetricSecurityKey(DefaultSymmetricKeyBytes_128) { KeyId = "DefaultSymmetricSecurityKey_128" };
 
+        public static string DefaultSymmetricKeyEncoded_192 = "hf+ctyfSxzcVMn5B5D+u4It4/HHjnh/E";
+        public static byte[] DefaultSymmetricKeyBytes_192 = Convert.FromBase64String(DefaultSymmetricKeyEncoded_192);
+        public static SymmetricSecurityKey DefaultSymmetricSecurityKey_192 = new SymmetricSecurityKey(DefaultSymmetricKeyBytes_192) { KeyId = "DefaultSymmetricSecurityKey_192" };
+        
+
```
henrik-me: nit: one empty too many

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Helpers/AesGcmAuthenticatedEncryptionProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+//------------------------------------------------------------------------------
+
+using Microsoft.IdentityModel.Logging;
+using System;
+using System.Security.Cryptography;
+
+namespace Microsoft.IdentityModel.Tokens.Saml.Tests
+{
+    // Helper AuthenticatedEncryptionProvider class made to mimic AES-GCM
+    // remove when AES-GCM is released and supported
```
henrik-me, line 37: >// [](start = 4, length = 2)

it seems like you are not using xml comments, I suggest using xml comments instead of this. Also in test code

src/Microsoft.IdentityModel.Xml/XmlUtil.cs
```diff
@@ -430,5 +431,34 @@ internal static string[] TokenizeInclusiveNamespacesPrefixList(string inclusiveN
                 return result;
             }
         }
+
+#if NET45 || NET451
+        private static readonly XmlReaderSettings SafeSettings = new XmlReaderSettings { XmlResolver = null, DtdProcessing = DtdProcessing.Prohibit };
```
henrik-me: suggestion: SafeXmlReaderSettings

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
```diff
@@ -2250,6 +2466,172 @@ protected virtual void WriteSubjectLocality(XmlWriter writer, Saml2SubjectLocali
             writer.WriteEndElement();
         }
 
+        private EncryptedData CreateEncryptedData(byte[] assertionData, EncryptingCredentials encryptingCredentials, out SymmetricSecurityKey sessionKey)
```
henrik-me: >EncryptedData [](start = 16, length = 13)

suggestion: instead of an out parameter you can also return a tuple (if the platforms this runs on support value tuples that might be even better.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
```diff
@@ -2250,6 +2466,172 @@ protected virtual void WriteSubjectLocality(XmlWriter writer, Saml2SubjectLocali
...
+            if (encryptingCredentials == null)
+                throw LogExceptionMessage(new Saml2SecurityTokenEncryptedAssertionEncryptionException(LogMessages.IDX13624));
+
+            if (sessionKey == null)
+                throw LogExceptionMessage(new Saml2SecurityTokenEncryptedAssertionEncryptionException(LogMessages.IDX13629));
+
+            // AsymmetricSecurityKey is provided:
+            // New session key will be created to encrypt an assertion
+            // Session key will be wrapped with provided AsymmetricSecurityKey
+            if (encryptingCredentials.Key is AsymmetricSecurityKey)
```
henrik-me: >if [](start = 12, length = 2)

consider inverting all your if's moving all the short functions on top (e.g. Symmetric does nothing, and not assymetric does nothing. this will give an easy overview of the various return statements from the function.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
henrik-me: another reason for me asking is that it's a best practice to only catch exceptions you know what are. as mentioned certain system exceptions should ideally never be captured, it's better from a security perspective to let the code fail with an exception.

---
In reply to: [221345997](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1025#discussion_r221345997) [](ancestors = 221345997)

src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
```diff
@@ -39,6 +39,12 @@ public static class SecurityAlgorithms
         public const string Aes256Encryption = "http://www.w3.org/2001/04/xmlenc#aes256-cbc";
         public const string DesEncryption = "http://www.w3.org/2001/04/xmlenc#des-cbc";
 
+        // Currently not supported. It's possible to provide your own implementation of the algorithm by overriding CryptoProviderFactory and AuthenticatedEncryptionProvider.
+        // See: https://www.w3.org/TR/xmlenc-core1/#sec-AES-GCM
```
henrik-me, line 42: consider adding these as xml comments to all the consts. that way a developer will actually see these comments in the tooltip.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -180,7 +183,8 @@ internal static class LogMessages
         public const string IDX10684 = "IDX10684: Unable to create a AsymmetricAdapter, Algorithm: '{0}', Key: '{1}'.";
         public const string IDX10685 = "IDX10685: Unable to Sign, Internal SignFunction is not available.";
         public const string IDX10686 = "IDX10686: Unable to Verify, Internal VerifyFunction is not available.";
-        public const string IDX10687 = "IDX10687: Unable to create a AsymmetricAdapter. For NET45 or NET451 only types: '{0}' or '{1}' are supported. RSA is of type: '{2}'..";
+        public const string IDX10687 = "IDX10687: Unable to create a AsymmetricAdapter. For NET45 or NET451 only types: '{0}' or '{1}' are supported. RSA is of type: '{2}'.";
+        public const string IDX10688 = "IDX10688: The algorithm '{0}' is not currently supported. It's possible to provide your own implementation of the algorithm by overriding CryptoProviderFactory and AuthenticatedEncryptionProvider. Example: aka.ms/identitymodel-aesgcm-support";
```
henrik-me: >aka.ms/identitymodel-aesgcm-support [](start = 246, length = 35)

did you add the wiki page/sample for this and enable the aka.ms link?

henrik-me (changes requested):
:clock1:

src/Microsoft.IdentityModel.Xml/EncryptionMethod.cs
GeoK: IsStartElement checks if xml reader is pointing to a start tag or empty element tag.
In case where there is nothing to read, reader will be pointing to an EndElement and exit from the while loop.

This logic is implemented because EncryptionMethod supports any element as a child. For now, we care only about DigestMethod and we are ignoring (but logging) other elements that may appear.

That said, it might be better to use reader.Skip() instead of reader.Read() on ln. 109 as DigestMethod by xenc xsd schema might have child elements (it's usually an empty element)

src/Microsoft.IdentityModel.Xml/XMLEncryptionConstants.cs
GeoK: This is more of a utility class. While coding it's easy to find a specific constant i.e. an Attribute by typing XmlEncCostants.Attribute. 

Not really sure if I understand - Are you suggesting that it would be a good idea to create a class hierarchy based on an xenc xsd schema e.g. EncryptionMethod contains only _Algorithm_ as an attribute and _KeySize_, _OAEPparams_ as elements.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
GeoK: that's a good point. We decided to support wrong RsaOaepKeyWrap identifier so we should have tests for that one too.

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Helpers/AesGcmAuthenticatedEncryptionProvider.cs
GeoK: Are we compiling doc for test code as well?
Is using xml comments in test projects needed by some process or it serves for readability purposes?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
GeoK: I was thinking tuples while refactoring this method, and it seemed to me that using out param is a bit cleaner, considering that it works on each platform and that CreateEncryptedData is a private method.
Is there a reason why tuples (not ValueTuples) are preferable over *out* param, for this specific case?

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
GeoK: aka.ms links points to the code developed to support AES-GCM tests. Do you suggest that code sample doesn't give enough information/guidance to a developer and that I should add a wiki page as well?

test/Microsoft.IdentityModel.Tests/KeyingMaterial.cs
GeoK: addressed with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1025/commits/86c08099b5a6a16659279f788edf99625bbb7151

src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
GeoK: addressed with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1025/commits/c8e9d097cbc2e9cd1d54d0624ac67e4c6530458f

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
GeoK: addressed with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1025/commits/a562057cfa787c2e5212cd69a9b0deb22be013aa

src/Microsoft.IdentityModel.Xml/XmlUtil.cs
GeoK: addressed with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1025/commits/81f395c2111a78b7228cf65458fc81711d6ddb79

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
GeoK: addressed with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1025/commits/0b0abbf7933be1aa88a9cf0bdf3e472a69e6ea8a

src/Microsoft.IdentityModel.Xml/XMLEncryptionConstants.cs
GeoK: addressed with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/3ebd2b74bc0cae863ea82afcba6448e64021dc2e

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Assertion.cs
```diff
@@ -51,101 +59,179 @@ public Saml2Assertion(Saml2NameIdentifier issuer)
...
         /// advice or do not wish to make use of it. [Saml2Core, 2.3.3]
         /// </summary>
+        /// <exception cref="Saml2SecurityTokenEncryptedAssertionException"> If this assertion is encrypted.</exception>
         public Saml2Advice Advice
         {
-            get;
-            set;
+            get
+            {
+                if (Encrypted)
```
brentschmaltz, line 96: We could just return null here, and not throw.
The use case is a user created a SAML token with just encrypted assertion, that would be the only value that is not null.

This would allow us to remove the type Saml2SecurityTokenEncryptedAssertionException.

616b2f:
Looks great! Any clue in which version this would be awailable?

GeoK:
> 
> 
> Looks great! Any clue in which version this would be awailable?

The plan is to make this feature available in the next major release, probably 6.x.
That being said, this is feature is currently on hold and requires more work.

616b2f:
Thank you for your prompt reply. Looking forward to 6.x :)

thuannguy:
I would love to have this feature. Encryption is mandatory for my customers :)

brentschmaltz:
@thuannguy yep, we need this feature.
I will see if we can get a timeline established.

brentschmaltz:
We may want to include aes-256cbc as AzureAD supports this.

keegan-caruso:
@GeoK - Can we close this for now? This will require rework to target dev.

Issue #1006, opened 2018-08-23
SAML2 Subjectconfirmation data lifetime check should be optional


Tarig0:
Currently if there is a subject confirmation node in the SAML2 token it will check the lifetime even if there is no lifetime information in the subject data node. This will in turn throw an error since the expire and notbefore arguments will be null.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e9fe530ed023a1566c51567b4f91b88c59b35509/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs#L279-L288

brentschmaltz:
@Tarig0	will get this in 5.3.1

PR #1027, opened 2018-09-27
Make LifetimeValidation optional -ConfirmationData

GeoK:
Lifetime validation for SubjectConfirmationData will be invoked only
if lifetime information (NotBefore or NotOnOrAfter) is present under
SubjectConfirmationData.

NotBefore and NotOnOrAfter are optional elements under
SubjectConfirmationData element ([saml-core-2.0 ](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf)- 2.4.1.2).

Resolves: #1006


GeoK:
We shouldn't enforce Lifetime Validation for SubjectConfirmationData.

PR #1040, opened 2018-10-28
Fix incorrect param count in log message

AndersAbel:
The IDX30100 log message has too many parameters in the format string, causes a formatting error when trying to write the log message.


henrik-me:
Hi @AndersAbel, as far as I can see the code is used here:
**\Microsoft.IdentityModel.Xml\DsigSerializer.cs**
`throw XmlUtil.LogReadException(LogMessages.IDX30100, XmlSignatureConstants.Elements.Transform, algorithm, SecurityAlgorithms.ExclusiveC14n, SecurityAlgorithms.ExclusiveC14nWithComments);
`
which seems to be 4 params.
Do you have PII enabled or disabled?
CC: @mafurman

AndersAbel:
Yes, that's 4 params. And the original string has 5 place holders (0-4). :)

AndersAbel:
@mafurman @henrik-me Could you please look at this again?

PR #1047, opened 2018-11-09
Replace ApiChangeTest with ApiCompat tool

GeoK:
* ApiCompat is a command line tool that verifies that two sets of APIs are compatible, in other words that they don't contain any breaking changes from one to the other, for each target framework.
* ApiCompat is consumed as a standalone package from [arcade](https://github.com/dotnet/arcade).
* ApiCompat versions >= _1.0.0-beta.18557.3_ support [OptionalRules](https://github.com/dotnet/arcade/pull/1262), like [ParameterNamesCannotChange](https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.ApiCompat/Rules/Compat/ParameterNamesCannotChange.cs).
* To disable ApiCompat, set _RunApiCompat_ to **false** in _apiCompat.props_ file.
* To disable enforcement of optional rules, set ApiCompatEnforceOptionalRules to **false** in _apiCompat.props_ file.
* _generateContractAssemblies_ script is added to help moving implementation assemblies to _contractAssemblies_ directory when a new version is released.
* When `-generateContractAssemblies` flag is set, a new set of contractAssemblies will be created after a successful build.


henrik-me:
"Concern" in keeping the repo lightweight.
I do understand why you have the dll's and the text files in here, however IMO, it would be better not having the need for all these bits in the repo.... I suppose right now the only other option would be to grab the nuget package and extract the files from there. I suppose that is not supported by the tool.

GeoK:
> 
> 
> "Concern" in keeping the repo lightweight.
> I do understand why you have the dll's and the text files in here, however IMO, it would be better not having the need for all these bits in the repo.... I suppose right now the only other option would be to grab the nuget package and extract the files from there. I suppose that is not supported by the tool.

That was the initial idea, but we wanted to have a uniform approach for this project, special builds, and other non-public projects which are not available on nuget.

NuGet.Config
```diff
@@ -2,5 +2,6 @@
 <configuration>
   <packageSources>
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
```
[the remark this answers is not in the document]
henrik-me: Discussed offline. Approving PR expecting the component is registered.

henrik-me:
:shipit:

build.ps1
```diff
@@ -5,7 +5,8 @@ param(
     [string]$root=$PSScriptRoot,
     [string]$runTests="YES",
     [string]$failBuildOnTest="YES",
-    [string]$slnFile="wilson.sln")
+    [string]$slnFile="wilson.sln",
+    [switch]$generateContractAssemblies)
```
brentschmaltz: should we have a default here?

GeoK: The value is false by default when the parameter is not present. I'm checking if the switch is present with`
$generateContractAssemblies.IsPresent
` which _returns true if the parameter was specified on the command line, false otherwise._

brentschmaltz: ok

brentschmaltz: :shipit:

Issue #1042, opened 2018-11-02
Add overload to Saml2SecurityTokenHandler to consume a token from an XmlReader


AndersAbel:
When working with the SAML2P protocol, the token (assertion in Saml2 terms) is embedded inside a larger piece of XML, the SamlResponse - or inside an artifact resolution response message. Code supporting Saml2P will typically already have an open XmlReader to the data. Add overloads to be able to hand in an XmlReader instead of a string to the token validation API.

This was discussed in #759, but is really an own separate task that is for SAML2P support rather than metadata support. So it should have it's own issue.

AndersAbel:
Have started looking at this and found out that the `SaveSigninToken` flag is a challenge, as it expects the entire token to be present in string format.

I first considered just not supporting `SaveSigninToken` when using an Xml Reader, but I do see some use cases ahead where I will need that, so this must be solved.

My first idea was to just fall back to extract the OuterXml of the Assertion node in a Saml Response message and store it. However, that might not work if the namespace declarations are placed on the enclosing Response node. This is a well-known problem with signatures over part of XML Documents and that is why a canonicalization transform is always used.

I'm considering trying to get the output of the canonicalization transform (based on the recorded TokenStream) and write that out to a string representation that can be saved. It should keep the semantics of the assertion and make it possible to validate the standalone string representation.

Any thoughts @brentschmaltz @GeoK?

brentschmaltz:
@AndersAbel I have had issues with OuterXml and xml that contains '/r/n'. I have found that normalization will replace to '/n' and then signatures start failing. See https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1029

I believe you will need the canonicalized string representation of the SAML token.

brentschmaltz:
Will also add to SamlSecurityTokenHandler

brentschmaltz:
@AndersAbel I think we can get the canonical stream as we have collected all the nodes.

brentschmaltz:
Added with:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1181

PR #1048, opened 2018-11-11
WIP: Read Saml tokens from XmlReader

AndersAbel:
Work in progress to solve #1042 , just opening this to get an early possibility for feedback.

Looks like this is going to work. Quite a few overloads were added. And I will need to extract a string representation (canonicalized) of the token during the reading, as it is required for TokenValidationParameters.IssuerSigningKeyResolver and SaveSigninToken.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
@@ -201,17 +201,53 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
                 throw LogExceptionMessage(new ArgumentException(FormatInvariant(TokenLogMessages.IDX10209, token.Length, MaximumTokenSizeInBytes)));
 
             var samlToken = ValidateSignature(token, validationParameters);
-            ValidateConditions(samlToken, validationParameters);
-            ValidateSubject(samlToken, validationParameters);
-            var issuer = ValidateIssuer(samlToken.Issuer, samlToken, validationParameters);
-            ValidateTokenReplay(samlToken.Assertion.Conditions.NotOnOrAfter, token, validationParameters);
-            ValidateIssuerSecurityKey(samlToken.SigningKey, samlToken, validationParameters);
-            validatedToken = samlToken;
-            var identity = CreateClaimsIdentity(samlToken, issuer, validationParameters);
+
+            ValidateToken(token, validationParameters, out validatedToken, samlToken, out ClaimsIdentity identity);
+
             if (validationParameters.SaveSigninToken)
                 identity.BootstrapContext = token;
 
             LogHelper.LogInformation(TokenLogMessages.IDX10241, token);
+
+            return new ClaimsPrincipal(identity);
+        }
+
+        private void ValidateToken(string replayIdentifier, TokenValidationParameters validationParameters, out SecurityToken validatedToken, Saml2SecurityToken samlToken, out ClaimsIdentity identity)
+        {
+            ValidateConditions(samlToken, validationParameters);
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -261,6 +297,38 @@ protected virtual void ValidateTokenReplay(DateTime? expirationTime, string secu
...
+
+            if (validationParameters == null)
+                throw LogArgumentNullException(nameof(validationParameters));
+
+            if (validationParameters.SignatureValidator != null)
+                throw new NotImplementedException();
+
+            var samlToken = ReadSaml2Token(tokenReader);
+
+            return ValidateSignature(null, samlToken, validationParameters);
```
brentschmaltz: Why pass null, this changes the model for Saml2 token validation.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -201,17 +201,53 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
                 throw LogExceptionMessage(new ArgumentException(FormatInvariant(TokenLogMessages.IDX10209, token.Length, MaximumTokenSizeInBytes)));
 
             var samlToken = ValidateSignature(token, validationParameters);
+
+            ValidateToken(token, validationParameters, out validatedToken, samlToken, out ClaimsIdentity identity);
```
brentschmaltz: why are two out params needed?

In general we need the same model for all token handling.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -261,6 +297,38 @@ protected virtual void ValidateTokenReplay(DateTime? expirationTime, string secu
...
+        /// <para>If the 'token' signature is validated, then the <see cref="Saml2SecurityToken.SigningKey"/> will be set to the key that signed the 'token'. It is the responsibility of <see cref="TokenValidationParameters.SignatureValidator"/> to set the <see cref="Saml2SecurityToken.SigningKey"/></para></remarks>
+        protected virtual Saml2SecurityToken ValidateSignature(XmlReader tokenReader, TokenValidationParameters validationParameters)
+        {
+            if (tokenReader == null)
+                throw LogArgumentNullException(nameof(tokenReader));
+
+            if (validationParameters == null)
+                throw LogArgumentNullException(nameof(validationParameters));
+
+            if (validationParameters.SignatureValidator != null)
```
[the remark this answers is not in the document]
AndersAbel: Yes, the SAML2 API that reads off an XmlReader would be different. The existing API accepting a string is still kept. But supporting the SignatureValidator when reading off an XmlReader would require the token to be written out to a string, just to later read it back from a string. Which IMHO wouldn't make sense.

AndersAbel:
This is getting far too complex, so I'm closing this.

I'm now looking into adding the features I need by subclassing the Saml2SecurityTokenHandler instead. I will probably get back with a PR that splits the ValidateToken method into two parts, to allow me to call the second part after doing the first part in my own override.

brentschmaltz:
@AndersAbel I am working on getting in the Saml2SecurityTokenHandler to accept an xmlreader.
One of the issues I am solving is that if the Saml token inside a WSFedMessage has a "/r/n", the signature will fail.
Essentially, our current code is broken. We need changes to our WsFederation message.
It's actually broken back to the originally release in WIF in 2008.

We we have this change in place, it should be natural for a Samlp serializer, similar to SamlSerializer.

Issue #1038, opened 2018-10-24
JwtSecurityTokenHandler Validate Token fails if not using Windows in 5.3.0


RichardD012:
When creating a new `JwtSecurityTokenHandler()` and passing in a `RSACryptoServiceProvider` in OSX, an exception is thrown when calling `ValidateToken`.

The error is: `System.PlatformNotSupportedException: 'CspKeyContainerInfo' requires Windows Cryptographic API (CAPI), which is not available on this platform.`

```
Exceptions caught:
 'System.PlatformNotSupportedException: 'CspKeyContainerInfo' requires Windows Cryptographic API (CAPI), which is not available on this platform.
   at System.Security.Cryptography.RSACryptoServiceProvider.get_CspKeyContainerInfo()
   at Microsoft.IdentityModel.Tokens.RSACryptoServiceProviderProxy..ctor(RSACryptoServiceProvider rsa) in C:\agent2\_work\56\s\src\Microsoft.IdentityModel.Tokens\RsaCryptoServiceProviderProxy.cs:line 74
   at Microsoft.IdentityModel.Tokens.AsymmetricAdapter.Initialize(RSA rsa, String algorithm) in C:\agent2\_work\56\s\src\Microsoft.IdentityModel.Tokens\AsymmetricAdapter.cs:line 305
   at Microsoft.IdentityModel.Tokens.AsymmetricAdapter..ctor(SecurityKey key, String algorithm, HashAlgorithm hashAlgorithm, HashAlgorithmName hashAlgorithmName, Boolean requirePrivateKey) in C:\agent2\_work\56\s\src\Microsoft.IdentityModel.Tokens\AsymmetricAdapter.cs:line 84
```

The stack-trace refers to `RsaCryptoServiceProviderProxy.cs` and looking at the commit history, it looks like it added Net Standard 2.0 Support.  The parameter references the `CspKeyContainerInfo` in the constructor [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/7692d12e49a947f68a44cd3abc040d0c241376e6/src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs#L80).  

This does not work in OSX as, as the exception states, CspKeyContainerInfo isn't supported.  So when it tries to access the parameter, this exception is thrown.  The code used to call this previously worked in 5.2.4.

RichardD012:
[...]

PR #1050, opened 2018-11-15
Make RSACryptoServiceProviderProxy Windows-specific (#1038)

GeoK:
Fixes issue that users are facing on non-Windows platforms when using RSACrpytoServiceProvider.
* There is NO support for CspParameters and CspKeyContainerInfo on non-Windows platforms which makes RSACryptoServiceProviderProxy class Windows-specific.
* NETSTANDARD1_4 doesn't not support RSACryptoServiceProvider so there is no need for the proxy.
* Disabled warning _CS0162 Unreachable code detected_ on non-Windows platforms as it's treated as an error during build.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
@@ -296,7 +296,8 @@ namespace Microsoft.IdentityModel.Tokens
             // X509Certificate2.GetPrivateKey OR X509Certificate2.GetPublicKey.Key
             // These calls return an AsymmetricAlgorithm which doesn't have API's to do much and need to be cast.
             // RSACryptoServiceProvider is wrapped to support SHA2
-#if NET45 || NET451 || NET461 || NETSTANDARD2_0
+            // RSACryptoServiceProviderProxy is only supported on Windows platform
+#if WINDOWS && (NET45 || NET451 || NET461 || NETSTANDARD2_0)
             _useRSAOeapPadding = algorithm.Equals(SecurityAlgorithms.RsaOAEP, StringComparison.Ordinal)
                               || algorithm.Equals(SecurityAlgorithms.RsaOaepKeyWrap, StringComparison.Ordinal);
 
@@ -338,7 +339,7 @@ namespace Microsoft.IdentityModel.Tokens
             SignatureFunction = SignWithRsa;
             VerifyFunction = VerifyWithRsa;
 #endif
-            }
+        }
 
         private RSA RSA { get; set; }
 
--- a/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
@@ -181,6 +181,7 @@ namespace Microsoft.IdentityModel.Tokens
         public const string IDX10685 = "IDX10685: Unable to Sign, Internal SignFunction is not available.";
         public const string IDX10686 = "IDX10686: Unable to Verify, Internal VerifyFunction is not available.";
         public const string IDX10687 = "IDX10687: Unable to create a AsymmetricAdapter. For NET45 or NET451 only types: '{0}' or '{1}' are supported. RSA is of type: '{2}'..";
+        public const string IDX10688 = "IDX10688: RSACryptoServiceProviderProxy creation is only supported on Windows.";
 
         // security keys
         public const string IDX10700 = "IDX10700: Invalid RsaParameters: '{0}'. Both modulus and exponent should be present";
--- a/src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
+++ b/src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
@@ -37,6 +37,9 @@ namespace Microsoft.IdentityModel.Tokens
     /// The purpose of this class is to ensure that we obtain an RsaCryptoServiceProvider that supports SHA-256 signatures.
     /// If the original RsaCryptoServiceProvider doesn't support SHA-256, we create a new one using the same KeyContainer.
     /// </summary>
+    /// <remarks>
+    /// There is no support for <see cref="CspParameters"/> and <see cref="CspKeyContainerInfo"/> on non-Windows platforms which makes <see cref="RSACryptoServiceProviderProxy"/> a Windows-specific class.
+    /// </remarks>
     public class RSACryptoServiceProviderProxy : RSA
     {
         // CryptoApi provider type for an RSA provider supporting sha-256 digital signatures
@@ -64,13 +67,18 @@ namespace Microsoft.IdentityModel.Tokens
         /// </summary>
         public override string KeyExchangeAlgorithm => _rsa.KeyExchangeAlgorithm;
 
+#pragma warning disable CS0162 // Disable warning on non-Windows platforms - CS0162: Unreachable code detected
         /// <summary>
         /// Initializes an new instance of <see cref="RSACryptoServiceProviderProxy"/>.
         /// </summary>
         /// <param name="rsa"><see cref="RSACryptoServiceProvider"/></param>
         /// <exception cref="ArgumentNullException">if <paramref name="rsa"/> is null.</exception>
+        /// <exception cref="PlatformNotSupportedException">RSACryptoServiceProviderProxy creation is only supported on Windows.</exception>
         public RSACryptoServiceProviderProxy(RSACryptoServiceProvider rsa)
         {
+#if !WINDOWS
+            throw new PlatformNotSupportedException(LogMessages.IDX10688);
+#endif
             if (rsa == null)
                 throw LogHelper.LogArgumentNullException(nameof(rsa));
 
@@ -101,6 +109,7 @@ namespace Microsoft.IdentityModel.Tokens
                 _rsa = rsa;
             }
         }
+#pragma warning restore CS0162
 
         /// <summary>
         /// Decrypts data with the System.Security.Cryptography.RSA algorithm.


src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -296,7 +296,8 @@ private void Initialize(RSA rsa, string algorithm)
             // X509Certificate2.GetPrivateKey OR X509Certificate2.GetPublicKey.Key
             // These calls return an AsymmetricAlgorithm which doesn't have API's to do much and need to be cast.
             // RSACryptoServiceProvider is wrapped to support SHA2
-#if NET45 || NET451 || NET461 || NETSTANDARD2_0
+            // RSACryptoServiceProviderProxy is only supported on Windows platform
+#if WINDOWS && (NET45 || NET451 || NET461 || NETSTANDARD2_0)
```
henrik-me: >WINDOWS [](start = 4, length = 7)

consider throwing a platform exception if !WINDOWS or alternatively do nothing (except logging that RSA is not supported).

henrik-me:
> #if NET461 || NETSTANDARD1_4 || NETSTANDARD2_0

I thought RSA wasn't supported on netstandard1.4 ?

---
Refers to: src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs:333 in 366aa0c. [](commit_id = 366aa0cfc9fdb1d44e5778223d3681c51c2d2df7, deletion_comment = False)

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -181,6 +181,7 @@ internal static class LogMessages
         public const string IDX10685 = "IDX10685: Unable to Sign, Internal SignFunction is not available.";
         public const string IDX10686 = "IDX10686: Unable to Verify, Internal VerifyFunction is not available.";
         public const string IDX10687 = "IDX10687: Unable to create a AsymmetricAdapter. For NET45 or NET451 only types: '{0}' or '{1}' are supported. RSA is of type: '{2}'..";
+        public const string IDX10688 = "IDX10688: RSACryptoServiceProviderProxy creation is only supported on Windows platform.";
```
henrik-me: >platform [](start = 118, length = 8)

pls. remove this (platform) word. (otherwise the sentence needs to end with ".. the Windows platform."

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
```diff
@@ -36,6 +36,7 @@ namespace Microsoft.IdentityModel.Tokens
     /// <summary>
     /// The purpose of this class is to ensure that we obtain an RsaCryptoServiceProvider that supports SHA-256 signatures.
     /// If the original RsaCryptoServiceProvider doesn't support SHA-256, we create a new one using the same KeyContainer.
+    /// There is NO support for <see cref="CspParameters"/> and <see cref="CspKeyContainerInfo"/> on non-Windows platforms which makes <see cref="RSACryptoServiceProviderProxy"/> Windows-specific class.
```
henrik-me: >// There is NO  [](start = 5, length = 15)

perhaps this should be in a remark section and not in the summary?

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
```diff
@@ -64,13 +65,18 @@ public class RSACryptoServiceProviderProxy : RSA
...
         /// <summary>
         /// Initializes an new instance of <see cref="RSACryptoServiceProviderProxy"/>.
         /// </summary>
         /// <param name="rsa"><see cref="RSACryptoServiceProvider"/></param>
         /// <exception cref="ArgumentNullException">if <paramref name="rsa"/> is null.</exception>
+        /// <exception cref="PlatformNotSupportedException">RSACryptoServiceProviderProxy creation is only supported on Windows platform.</exception>
         public RSACryptoServiceProviderProxy(RSACryptoServiceProvider rsa)
         {
+#if !WINDOWS
+            throw new PlatformNotSupportedException(LogMessages.IDX10688);
```
henrik-me: >PlatformNotSupportedException [](start = 22, length = 29)

this was what I would have considered you should throw in the adapter class...

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
GeoK: There is no reason to throw here.

The purpose of RSACryptoServiceProviderProxy is to _enable_ SHA2 signatures in cases when  provided RSACryptoServiceProvider understands only SHA1 - which is the case only on Windows.
On .NET Core, RSACryptoServiceProvider understands SHA2, so no proxy is needed.

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
GeoK: I thought about this too. Decided to include it in the summary as it may be useful to show this information in an intellisense tooltip (content under remarks section is not shown in the tooltip).

On second thought, a summary section _should_ contain only brief information about a type or member and this one is already quite long. I'll move this sentence under remarks section.

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
GeoK: We don't have any code path in Wilson that would instantiate this class on non-Windows platforms. As RSACryptoServiceProviderProxy is a public class, my desire here was to explicitly state that this class is Windows-only.

GeoK:
> 
> 
> > #if NET461 || NETSTANDARD1_4 || NETSTANDARD2_0
> 
> I thought RSA wasn't supported on netstandard1.4 ?
> 
> Refers to: src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs:333 in [366aa0c](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/366aa0cfc9fdb1d44e5778223d3681c51c2d2df7). [](commit_id = [366aa0c](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/366aa0cfc9fdb1d44e5778223d3681c51c2d2df7), deletion_comment = False)

RSACryptoServiceProvider is not supported on netstandard1.4.

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
henrik-me: no need to capitalize NO here imo.

---
In reply to: [233948740](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1050#discussion_r233948740) [](ancestors = 233948740)

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
```diff
@@ -37,6 +37,9 @@ namespace Microsoft.IdentityModel.Tokens
     /// The purpose of this class is to ensure that we obtain an RsaCryptoServiceProvider that supports SHA-256 signatures.
     /// If the original RsaCryptoServiceProvider doesn't support SHA-256, we create a new one using the same KeyContainer.
     /// </summary>
+    /// <remarks>
+    /// There is no support for <see cref="CspParameters"/> and <see cref="CspKeyContainerInfo"/> on non-Windows platforms which makes <see cref="RSACryptoServiceProviderProxy"/> Windows-specific class.
```
henrik-me: >  [](start = 178, length = 1)

missing a

henrik-me:
> #if NET45 || NET451 || NET461 || NETSTANDARD2_0

should this then have been saying if WINDOWS?

---
Refers to: src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs:28 in 30e7b0f. [](commit_id = 30e7b0fe16d3269ce6a941369f196a163f331f04, deletion_comment = False)

GeoK:
> 
> 
> > #if NET45 || NET451 || NET461 || NETSTANDARD2_0
> 
> should this then have been saying if WINDOWS? ... seems like you are throwing a platform exception instead.
> 
> Refers to: src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs:28 in [30e7b0f](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/30e7b0fe16d3269ce6a941369f196a163f331f04). [](commit_id = [30e7b0f](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/30e7b0fe16d3269ce6a941369f196a163f331f04), deletion_comment = False)

@henrik-me: #if WINDOWS would fail a build as there is no RsaCryptoServiceProvider class in NETSTANDARD1_4.

Changing this line
`#if NET45 || NET451 || NET461 || NETSTANDARD2_0)`
to
`#if WINDOWS && (NET45 || NET451 || NET461 || NETSTANDARD2_0)` requires additional refactoring of AsymmetricAdapter class.

I tried to minimize the number of changes, but I'm open to further refactor AsymmetricAdapter and  RsaCryptoServiceProvider in case that it's not preferable to throw a platform exception.

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
GeoK: @brentschmaltz:
#if WINDOWS would fail a build as there is no RsaCryptoServiceProvider class in NETSTANDARD1_4.

Changing this line
`#if NET45 || NET451 || NET461 || NETSTANDARD2_0)`
to
`#if WINDOWS && (NET45 || NET451 || NET461 || NETSTANDARD2_0)` requires additional refactoring of AsymmetricAdapter class.

I tried to minimize the number of changes, but I'm open to further refactor AsymmetricAdapter and RsaCryptoServiceProvider in case that it's not preferable to throw a platform exception.

brentschmaltz: About how much work would it be?

brentschmaltz: too much work, just leave it as it is.

brentschmaltz:
:shipit:

Issue #1055, opened 2018-11-19
Make the methods on the JsonWebTokenHandler virtual


mafurman:
It should be possible to extend the JsonWebTokenHandler and override its methods.

PR #1057, opened 2018-11-19
Included TryGetValue() methods in the JsonWebToken and also made the public methods of JsonWebTokenHandler virtual.

### LINKED ISSUE #1056: Add TryGetPayloadValue() and TryGetHeaderValue() to JsonWebToken

mafurman:
Fixes #1056 and #1055.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -31,6 +31,7 @@ using System.Linq;
 using System.Security.Claims;
 using Microsoft.IdentityModel.Logging;
 using Microsoft.IdentityModel.Tokens;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.IdentityModel.JsonWebTokens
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -499,6 +499,23 @@ public T GetPayloadValue<T>(string key)
...
+        public bool TryGetPayloadValue<T>(string key, out T value)
+        {
+            if (string.IsNullOrEmpty(key))
+            {
+                value = default(T);
+                return false;
+            }
+
+            var result = Payload.TryGetValue(key, out var jTokenValue);
+            value = jTokenValue == null ? default(T) : jTokenValue.ToObject<T>();
```
brentschmaltz: This will need to be in a try/catch as jTokenValue.ToObject<T> could throw.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,5 +527,22 @@ public T GetHeaderValue<T>(string key)
...
+        public bool TryGetHeaderValue<T>(string key, out T value)
+        {
+            if (string.IsNullOrEmpty(key))
+            {
+                value = default(T);
+                return false;
+            }
+
+            var result = Header.TryGetValue(key, out var jTokenValue);
+            value = jTokenValue == null ? default(T) : jTokenValue.ToObject<T>();
```
brentschmaltz: This will need to be in a try/catch as jTokenValue.ToObject<T> could throw.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -508,7 +549,47 @@ public T GetHeaderValue<T>(string key)
...
+        public bool TryGetHeaderValue<T>(string key, out T value)
+        {
+            if (string.IsNullOrEmpty(key))
+            {
+                value = default(T);
+                return false;
+            }
+
+            var result = Header.TryGetValue(key, out var jTokenValue);
+            try
```
brentschmaltz: if 'result' is false, then do we need to execute the try  - catch?
we are making the assumption that all call graphs will set jTokenValue to null, can we assume this?
we could write:
```
if (!Header.TryGetValue ...)
{ 
	value = default(T);
	return false;
}

try
{
	value = jTokenValue.ToObject<T>();
}

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: Fixed.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: Good suggestion! 👍 You're right; the 'result' variable is unnecessary here. I've made similar changes to the other methods as well.

brentschmaltz:
:shipit:

Issue #1052, opened 2018-11-15
Support JsonWebKey.CreateECDsa() on non-Windows platforms


GeoK:
* CreateECDsa() is currently supported in Wilson for all targets, but only on Windows.
* Netstandard2.0 supports ECDsa.ImportParameters  (net core 2.0+ and .net framework 4.7+, [but not 4.6.1](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecdsa.importparameters?view=netframework-4.7.2#applies-to)).

See also: #581, #918, #1038

PR #1058, opened 2018-11-19
Support JsonWebKey.CreateECDsa() on non-Windows platforms

GeoK:
### CreateECDsa() - .NET target + Runtime + Platform coverage

| Target/OS | Windows | Unix |
| ------------- | ------------- | ------------- |
| _Runtime + IdentityModel .Net Target_ | - | - |
| NET45  | Supported before and after the change  | N/A |
| NET451  | Supported before and after the change  | N/A |
| NET461  | Supported before and after the change  | N/A |
| NET461/2 + netstandard1.4  | Supported before and after the change  | N/A |
| >=NET47 + netstandard1.4  | Supported before and after the change  | N/A |
| NET461/2 + netstandard2.0  | Supported before and after the change  | N/A |
| >=NET47 + netstandard2.0  | Supported before and after the change  | N/A |
| Core1.0 + netstandard1.4  | Supported before and after the change  | Wasn't supported before and it's still not supported |
| Core2.0 + netstandard1.4  | Supported before and after the change  | Wasn't supported before and it's still not supported |
| Core2.0 + netstandard2.0  | Supported before and after the change  | Wasn't supported before. **Supported** after the change |

---

### CreateECDsa() - Code Paths

| Target/OS | Windows | Unix |
| ------------- | ------------- | ------------- |
| _Runtime + IdentityModel .Net Target_ | - | - |
| NET45  | legacy code utilizing CngKey/ECDsCng  | N/A |
| NET451  | legacy code utilizing CngKey/ECDsCng  | N/A |
| NET461  | legacy code utilizing CngKey/ECDsCng  | N/A |
| NET461/2 + netstandard1.4  | if  execution runtime supports CngKey operations -> legacy code utilizing CngKey/ECDsCng | N/A |
| >=NET47 + netstandard1.4  | if  execution runtime supports CngKey operations -> legacy code utilizing CngKey/ECDsCng  | N/A |
| NET461/2 + netstandard2.0  | NET461/462 doesn't support ECParameters -> if  execution runtime supports CngKey operations -> legacy code utilizing CngKey/ECDsCng (_slightly slower execution path than before, because of the support checks_)  | N/A |
| >=NET47 + netstandard2.0  | New feature (code utilizing ECParameters)  | N/A |
| Core1.0 + netstandard1.4  | if  execution runtime supports CngKey operations -> legacy code utilizing CngKey/ECDsCng  | Wasn't supported before and it's still not supported |
| Core2.0 + netstandard1.4  | if  execution runtime supports CngKey operations -> legacy code utilizing CngKey/ECDsCng  | Wasn't supported before and it's still not supported |
| Core2.0 + netstandard2.0  |  New feature (code utilizing ECParameters)  |  New feature (code utilizing ECParameters) |

------

* Created ECDsaAdapter class to abstract the decisions listed above.
* Removed 'algorithm' parameter from CreateECDsa method signature as it's not being used in the method.
* Cleaned-up an unused log message and improved an existing log message.
* Removed #if WINDOWS constructs 
* Catch, and rethrow, exceptions in the ECDsaCng/CngKey flow to achieve similarity with the ECParameters flow.
* Fixed a case when GCHandle might throw if not initialized

------

Tests:

* Added ECDsaAdapter unit tests
* Expanded ExpectedException class to allow setting ignoreInnerException property to 'true'. Previously it was hardcoded to 'false'.
* Slightly refactored KeyingMaterial so that ECDsaTests could run on Unix ('dotnet test test/Microsoft.IdentityModel.Tokens.Tests/Microsoft.IdentityModel.Tokens.Tests.csproj --filter "AsymmetricSignAndVerify|CreateECDsa"). Windows CNG is not supported on Unix.

Resolves: #1052


src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -168,8 +168,7 @@ internal static class LogMessages
         public const string IDX10672 = "IDX10672: GetKeyedHashAlgorithm returned null, key: {0}, algorithm {1}.";
         public const string IDX10673 = "IDX10673: CryptoProviderFactory.GetHashAlgorithm returned null, factory: {0}, algorithm: {1}.";
         public const string IDX10674 = "IDX10674: JsonWebKeyConverter does not support SecurityKey of type: {0}";
-        public const string IDX10675 = "IDX10675: The value of '{0}' must be '{1}' bits, but was {2}.";
-        public const string IDX10676 = "IDX10676: ECDsa creation is currently only supported for the Windows operating system.";
```
henrik-me: >IDX10676 [](start = 40, length = 8)

I suppose removing this could be a breaking change. Would it make sense somehow to mark it obsolete? or otherwise indicate that it's not used anymore?

GeoK: LogMessages is an internal class so this should not be a breaking change.

henrik-me: people could have taken a dependency on the exception identifier

GeoK: No exception with the log identifier IDX10676  is thrown anymore, so nothing should break even if someone is taking a dependency on it.

On the other side, reusing old, deleted, log identifiers might be problematic. For that reason, we'll keep a deleted log message commented out, so we don't accidentally reuse one of the old identifiers when creating a new one.
```csharp
//public const string IDX10676 = "IDX10676:"

File: test/Microsoft.IdentityModel.Tests/ExpectedException.cs [hunk 2]
```diff
@@ -143,15 +143,15 @@ public void ProcessException(Exception exception, List<string> errors = null)
...
-                    HandleError("exception.InnerException != null && expectedException.InnerTypeExpected == null.\nexception.InnerException: " + exception.InnerException, errors);
+                    HandleError("exception.InnerException != null && expectedException.InnerTypeExpected == null && IgnoreInnerException == false.\nexception.InnerException: " + exception.InnerException, errors);
                     return;
                 }
 
-                if (exception.InnerException == null && InnerTypeExpected != null)
+                if (exception.InnerException == null && InnerTypeExpected != null && !IgnoreInnerException)
                 {
-                    HandleError("exception.InnerException == null, expectedException.InnerTypeExpected != null.\nexpectedException.InnerTypeExpected: " + InnerTypeExpected, errors);
+                    HandleError("exception.InnerException == null && expectedException.InnerTypeExpected != null && IgnoreInnerException == false.\nexpectedException.InnerTypeExpected: " + InnerTypeExpected, errors);
```
henrik-me: >IgnoreInnerException == false [](start = 116, length = 29)

why this construct vs ! ... the ! is used in most places in the code. Seems inconsistent to me (and not needed)

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
GeoK: I see your point. Our tooling reports if a log message is no longer used anywhere, so I guess that the decision to remove unused log messages is made before the tooling was setup.

The reason why we decided to comment-out a log message, instead of deleting, is to be able to setup our tools to search for duplicate log ids, even if a log message is commented-out. With that in place, we'll be sure that we are not using reusing old/deleted log identifiers.
Log message text is removed only to indicate that a particular log message considered as 'deleted'.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -333,6 +330,30 @@ internal ECDsa CreateECDsa(string algorithm, bool usePrivateKey)
             if (Y == null)
                 throw LogHelper.LogArgumentNullException(nameof(Y));
 
+#if !WINDOWS && NETSTANDARD2_0
```
brentschmaltz: Do ECParameters exist in Core 2.0 on Windows?
If so could this code work on Windows?

brentschmaltz: The issue seems to be that .net 4.6.1 does not support ECParameters so our target 4.6.1 would start failing on Windows, hence Windows needs to be excluded.

brentschmaltz:
:shipit:

GeoK:
@henrik-me, @brentschmaltz:
Could you please review the additional changes I made?
I updated the PR comment to reflect the additional changes.

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
```diff
@@ -13,7 +13,28 @@
   </PropertyGroup>
   
   <PropertyGroup Condition=" '$(TargetFramework)' == 'net45' OR '$(TargetFramework)' == 'net451' OR '$(TargetFramework)' == 'net461' ">
-    <DefineConstants>DESKTOP</DefineConstants>
+    <DefineConstants>$(DefineConstants);DESKTOP</DefineConstants>
+  </PropertyGroup>
+
+  <!-- feature flags -->
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'net45' ">
+    <DefineConstants>$(DefineConstants);HAVE_CREATE_ECDSA;</DefineConstants>
```
henrik-me: >HAVE_CREATE_ECDSA [](start = 40, length = 17)

nit: consider SUPPORTS_ECDSA

test/Microsoft.IdentityModel.Tests/KeyingMaterial.cs
```diff
@@ -449,7 +450,22 @@ static KeyingMaterial()
...
+            var Ecdsa384_Public = ECDsa.Create(Ecdsa384.ExportParameters(false));
+            var Ecdsa521_Public = ECDsa.Create(Ecdsa521.ExportParameters(false));
+
+            Ecdsa256Key = new ECDsaSecurityKey(Ecdsa256) { KeyId = "ECDsa256Key" };
+            Ecdsa384Key = new ECDsaSecurityKey(Ecdsa384) { KeyId = "ECDsa384Key" };
+            Ecdsa521Key = new ECDsaSecurityKey(Ecdsa521) { KeyId = "ECDsa521Key" };
+            Ecdsa256Key_Public = new ECDsaSecurityKey(Ecdsa256_Public) { KeyId = "ECDsa256Key_Public" };
+            Ecdsa384Key_Public = new ECDsaSecurityKey(Ecdsa384_Public) { KeyId = "ECDsa384Key_Public" };
+            Ecdsa521Key_Public = new ECDsaSecurityKey(Ecdsa521_Public) { KeyId = "ECDsa521Key_Public" };
+#endif
```
henrik-me: consider if all of this should be SUPPORTS_ECDSA

henrik-me:
done, only minor comments. :shipit:

---
In reply to: [442676188](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1058#issuecomment-442676188) [](ancestors = 442676188)

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
GeoK: It might be confusing as we do support ECDsa (e.g. ECDsaSecurityKey). We don't support creating of ECDsa using ECParameters in netstandard1.4 i.e. creating an ECDsa object using parameters of a JsonWebKey.

henrik-me: then consider calling it SUPPORTS_ECParameters ?

GeoK: I agree, it looks more intuitive. I'll make the change.

henrik-me:
:shipit:

---
In reply to: [442714616](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1058#issuecomment-442714616) [](ancestors = 442714616,442676188)

src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs
```diff
@@ -0,0 +1,401 @@
...
+        /// </summary>
+        /// <returns>True if <see cref="ECParameters"/> structure is supported, false otherwise.</returns>
+        private bool SupportsECParameters()
+        {
+            try
+            {
+                LoadECParametersType();
+                return true;
+            }
+            catch (TypeLoadException)
```
brentschmaltz: Would there be any harm in catching "Exception" ?

src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs
```diff
@@ -0,0 +1,401 @@
...
+        /// <returns>True if operations using <see cref="CngKey"/> are supported on user's runtime platform, false otherwise.</returns>
+        [MethodImpl(MethodImplOptions.NoOptimization)]
+        private bool SupportsCNGKey()
+        {
+            try
+            {
+                _ = CngKeyBlobFormat.EccPrivateBlob;
+                return true;
+            }
+            catch (PlatformNotSupportedException)
```
brentschmaltz: Would there be any harm in catching "Exception" ?

GeoK: I guess not. Catching any exception will make the code more resilient, as discussed. Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1058/commits/2ad287831df091cb053b771d931553565efec723

src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs
GeoK: I guess not. Catching any exception will make the code more resilient, as discussed. Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1058/commits/2ad287831df091cb053b771d931553565efec723

PR #1062, opened 2018-11-27
Make RSACryptoServiceProviderProxy available only on Desktop targets

GeoK:
* RSACryptoServiceProviderProxy is no longer present in netstandard2.0 target, hence the change in ApiCompatBaseline file for netstandard2.0.
* `#if WINDOWS` construct in JsonWebKey class is incorrect as it's a compile-time construct. Support for ECDsa on non-Windows platforms will be provided in #1058.


brentschmaltz:
:shipit:

Issue #1026, opened 2018-09-27
[SAML2] Null pointer error if using the IssuerSigningKeyResolver


Tarig0:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs#L336

If you use a custom issuersigningkeyresolver there is no way for the token.Signingkey to be set

brentschmaltz:
@Tarig0 sorry to miss this.
We will get this into 5.3.1.

PR #1065, opened 2018-11-30
Fix null pointer error when using the IssuerSigningKeyResolver

GeoK:
* Error occured when an IssuerSigningKeyResolver delegate was used and SigningKey was not set.
* Now providing Signature.KeyInfo.Id if available, otherwise null.
* Third parameter of IssuerSigningKeyResolver delegate is 'kid'.
`<param name="kid">A key identifier. It may be null.</param>`

Resolves: #1026


brentschmaltz:
:shipit:

PR #1068, opened 2018-12-04
Fix IdentityModel build and tests on Unix

GeoK:
* Added _Microsoft.DotNet.XUnitExtensions_ package to provide ability to specify that a test may only be run on the specified platforms  
  *  To use XUnitExtensions, decorate a test method with a platform attribute, e.g. `[PlatformSpecific(TestPlatforms.Windows)]` ([Supported platform attributes](https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.XUnitExtensions/src/TestPlatforms.cs))
  * When running tests, use _--filter_ option to set one (or combination) of the following categories
     * nonwindowstests
     * nonlinuxtests 
     * nonosxtests 
  * _Example_: Run only tests which run on Windows `--filter category!=nonwindowstests`
  * XUnitExtensions package is supported only by .net core >=2.0 target
  * Increased xunit version 2.3.1 -> 2.4.0 (XUnitExtensions requirement)
  * Added third party notice (XUnitExtensions)
  * Excluded two tests from running on OSX. _SignatureTampering_ test is very slow (~6 min), only on OSX. _SignatureTruncation_ test throws an exception when Verify method is called, again only on OSX.

* Upgraded MicrosoftNETTestSdkVersion (_vstest_) 15.5.0 ->15.9.0
  * It was not possible to run able to run "dotnet test" on a solution prior to 15.9.0 version, if solution included projects that do not include Microsoft.NET.Sdk.Test
  * vstest 15.9.0 ships with .net core 2.1.500
  * Since 15.9.0 version, it is possible to specify which projects are _test_ projects by specifying `<IsTestProject>true</IsTestProject>`
----
* Renamed _Microsoft.IdentityModel.Tests_ project to _Microsoft.IdentityModel.TestUtils_
* Renamed .sln file from _Linux_ to _Unix_ and added in references of projects that are supported on Unix
* Added  `--filter category!=nonwindowstests` to build.ps1 and test.ps1 scripts
* Added script which runs tests on Unix
* Removed _#if WINDOWS_ constructs
* Replaced backslash with forward slash to allow running apiCompat on Unix
* Moved RsaCng tests under NET461 test target
* Moved ECDsaCng tests under NET461 test target
* Replaced ECDsa.Cng with ECDsa.Create for .net core 2.0 target
* Refactored _SignatureProviderTests.cs_ as _signatureProviderVerify.Verify_ method throws an exception on OSX
* Changed encoding of _ClaimSets.cs_ file to UTF8  (caused tests to fail on Linux)
* Fixed WsFederation tests


test/Microsoft.IdentityModel.Tests/Microsoft.IdentityModel.Tests.csproj
```diff
@@ -13,6 +13,7 @@
     <SignAssembly>true</SignAssembly>
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TestTargets);net45</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netcoreapp2.0</TargetFrameworks>
+    <IsTestProject>false</IsTestProject>
```
henrik-me: >false [](start = 19, length = 5)

?

GeoK: This project doesn't contain any test methods.
It's a utility test project which contains test stubs, mocks, reference data, utilities, etc.

henrik-me: imo, then the name of the project is misleading.

brentschmaltz: @henrik-me 	I agree. Perhaps we should rename it to m.im.testutilities.
Can we open a new issue?

GeoK: I don't mind renaming it as part of this PR. _TestUtilities_ looks fine to me.

GeoK: On second thought, we have a static [class](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/test/Microsoft.IdentityModel.Tests/TestUtilities.cs#L51) named _TestUtilities_, which is used in many places.
Introducing the _Microsoft.IdentityModel.TestUtilities_ namespace would clash with a method call from the _TestUtilities_ class e.g. `TestUtilities.AssertFailIfErrors()` and doing something like `TestUtilities.TestUtilities.XYZ()` looks ugly to me.

I would suggest renaming the project to _m.im.TestUtils_ or just _m.im.Utilities_.

brentschmaltz:
:shipit:

PR #1069, opened 2018-12-05
Improved VerifyResourceUsage.pl so that it now checks log message usage across all directories

mafurman:
The following changes have been made to this script:

- Log message usage is now checked across all directories rather than only the directory a particular LogMessages.cs file is located in.

- Public log message IDs are now also checked (previously only internal ones were processed).

- Commented out log message IDs are checked for reuse.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
@@ -37,7 +37,7 @@ namespace Microsoft.IdentityModel.Tokens
     {
 #pragma warning disable 1591
         // general
-        public const string IDX10000 = "IDX10000: The parameter '{0}' cannot be a 'null' or an empty object.";
+        // public const string IDX10000 = "IDX10000:";
 
         // properties, configuration 
         public const string IDX10101 = "IDX10101: MaximumTokenSizeInBytes must be greater than zero. value: '{0}'";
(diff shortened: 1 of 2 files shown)


src/VerifyResourceUsage.pl
```diff
@@ -1,13 +1,13 @@
...
 ######################################################################
-# run like this:
+# Run like this:
 #    perl VerifyResourceUsage.pl
-# to discover 
+# To discover 
 #    1) what LogMessages.cs strings are unused
 #    2) whether any of the string IDs and string value prefixes don't match
 #    3) whether any LogMessages.cs files contain duplicate IDs
+#    4) whether any commented out IDs have been accidentally reused again
```
henrik-me: >again [](start = 68, length = 5)

nit: again might not be required in the sentence

src/VerifyResourceUsage.pl
```diff
@@ -341,45 +245,87 @@ sub FindDirs {
...
         }
     }
-
     PrintPerf("\nPERF: done reading and processing source files\n");
-    {
-        PrintPerf("\nDone processing files\n");
+}
+
+# go through all the directories again to discover unused, mismatching, and/or duplicate log message ids
+foreach my $directory (keys(%dirhash))
```
henrik-me: would it make sense to also print the dirhash?

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -37,7 +37,7 @@ internal static class LogMessages
     {
 #pragma warning disable 1591
         // general
-        public const string IDX10000 = "IDX10000: The parameter '{0}' cannot be a 'null' or an empty object.";
```
henrik-me: >: The parameter '{0}' cannot be a 'null' or an empty object." [](start = 48, length = 61)

what is the reason that the original meaning is being removed from the string when not used? I've seen this being done before.

src/VerifyResourceUsage.pl
```diff
@@ -1,13 +1,13 @@
-#!/usr/bin/perl
 ######################################################################
-# verify_resource_usage.pl
+# VerifyResourceUsage.pl
```
GeoK: Do you think there is value in calling VerifyResourceUsage.pl from our build.ps1 script?
It would be nice to have these checks as part of our build script, imo.

brentschmaltz:
@mafurman Would it be possible to introduce a test to show this is working?
for a couple of key scenarios:
1. the id doesn't match the prefix of the string { IDX00001 = "IDX0002:" }
2. duplicate id's in different files
3. unused id's.

src/VerifyResourceUsage.pl
mafurman: True; fixed this!

src/VerifyResourceUsage.pl
mafurman: This script prints out the directory a log message file belongs to when it processes it, so all the keys in the dirhash are eventually printed. Printing out the entire dirhash would result in all the corresponding log messages being printed out as well, which I think just ends up being too much information.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
mafurman: We decided to remove the original meaning since it makes it more obvious that the log message is no longer being used (rather than being commented out for some other reason).

src/VerifyResourceUsage.pl
mafurman: I created an issue for this: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1094

mafurman:
@brentschmaltz I think it would be a good idea to add a test project/file that this script can be run on to ensure it works properly. I created an issue for this: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1095

PR #1073, opened 2018-12-10
Moved TokenContext and TokenValidationResult into M.IM.Tokens

mafurman:
- TokenValidationResult has been moved to M.IM.Tokens from M.IM.JsonWebTokens.
- The version has been bumped to 5.4.0 as moving TokenValidationResult to a different namespace is a breaking change.


brentschmaltz:
:shipt:

PR #1075, opened 2018-12-11
JsonWebTokenHandler.ValidateToken now uses the IsValid property of TokenValidationResult

mafurman:
Other changes:

- JsonWebTokenHandler.ValidateToken() no longer throws an exception if the token cannot be validated, but rather returns a TokenValidationResult with the Exception property set and the IsValid property set to 'false'.

- Added in tests for the IsValid property of TokenValidationResult.

- In the case of a JWE, the ClaimsIdentity of the InnerToken is now returned rather than the ClaimsIdentity corresponding to the outer token.

- Fixed the JsonWebTokenHandlerTests that were not calling 'TestUtilities.AssertFailIfErrors'.


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -746,37 +748,48 @@ public virtual SecurityToken ReadToken(string token)
         /// <returns>A <see cref="TokenValidationResult"/></returns>
         public virtual TokenValidationResult ValidateToken(string token, TokenValidationParameters validationParameters)
         {
-            if (string.IsNullOrEmpty(token))
-                throw LogHelper.LogArgumentNullException(nameof(token));
+            try
+            {
+                if (string.IsNullOrEmpty(token))
+                    throw LogHelper.LogArgumentNullException(nameof(token));
```
brentschmaltz: no reason to throw here. you could just return a new TokenValidationResult.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -621,14 +629,22 @@ public void JWEDecompressionTest(JWEDecompressionTheoryData theoryData)
                 var validationResult = handler.ValidateToken(theoryData.JWECompressionString, theoryData.ValidationParameters);
                 var validatedToken = validationResult.SecurityToken as JsonWebToken;
 
-                Assert.NotEmpty(validatedToken.Claims);
-
-                theoryData.ExpectedException.ProcessNoException(context);
+                if (validationResult.IsValid)
+                {
+                    Assert.NotEmpty(validatedToken.Claims);
```
brentschmaltz: Instead of this assert, add a string to context.Diffs().

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -118,6 +121,7 @@ public static TheoryData<CreateTokenTheoryData> CreateJWETheoryData
                     new CreateTokenTheoryData
                     {
                         First = true,
+                        IsValid = true,
```
brentschmaltz: If the default value of CreateTokenTheoryData.IsValid was true, you could remove a lot of these sets.

PR #1077, opened 2018-12-14
Support JsonWebKeys with the Elliptic Curve key type

GeoK:
* Refactored JsonWebKeySet.GetSigningKeys() method and added support for JWKs with 'kty' = 'EC'.
* Refactored OpenIdConnect E2E tests (Fact -> Theory).
* Added E2E OpenIdConnect tests where jwks metadata contains EC JWKs with P-256, P-384, and P-521 curves.
* Added test where EC JWKs obtained from the metadata is not valid (unsupported curve).
* Added test where a token is signed using a key which is not part of the metadata.


src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -123,68 +123,100 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
         /// </summary>
         public IList<SecurityKey> GetSigningKeys()
         {
-            List<SecurityKey> keys = new List<SecurityKey>();
-            for (int i = 0; i < Keys.Count; i++)
-            {
-                JsonWebKey webKey = Keys[i];
+            var keys = new List<SecurityKey>();
```
henrik-me: >keys [](start = 16, length = 4)

IMO, this should be renamed. Too close to the class level Keys. e.g. signingKeys

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -123,68 +123,100 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
...
 
-                if (!StringComparer.Ordinal.Equals(webKey.Kty, JsonWebAlgorithmsKeyTypes.RSA))
+            foreach (var webKey in Keys)
+            {
+                // skip if "use" (Public Key Use) parameter is not empty or "sig"
+                if (!(string.IsNullOrWhiteSpace(webKey.Use) || webKey.Use.Equals(JsonWebKeyUseNames.Sig, StringComparison.Ordinal)))
+                {
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10808, webKey.Use));
                     continue;
+                }
```
henrik-me: great optimization, is this relevant for anything which is required to be const speed?

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -123,68 +123,100 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
...
+                    {
+                        KeyId = jsonWebKey.Kid
+                    };
+
+                    keys.Add(key);
+                }
+
+                return keys;
+            }
+            catch (Exception ex)
```
henrik-me: Wondering if this is too generic to catch? Can this be made more precise?

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/End2EndTests.cs
```diff
@@ -32,47 +32,116 @@
...
+                        ValidIssuer = configuration.Issuer,
+                    };
+
+                tokenHandler.ValidateToken(jwtToken.RawData, validationParameters, out SecurityToken securityToken);
 
-            SecurityToken securityToken = null;
-            tokenHandler.ValidateToken(jwtToken.RawData, validationParameters, out securityToken);
+                theoryData.ExpectedException.ProcessNoException(context);
+            }
+            catch (Exception ex)
```
henrik-me: very generic. shouldn't it be a more specific exception?  (related to the product question)

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: I like _signingKeys_. I'm going to rename it.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: No, we are just filtering out signing keys from the metadata received from a discovery endpoint.
There might be some keys used for encryption ("use": "enc") or some not defined by the standard, and we do not want these.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: I believe it could, but I'm not sure what is the benefit of catching specific exceptions in this scenario.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/End2EndTests.cs
GeoK: This is how our testing framework works. We'll catch any exception, but the exception will be compared with an expected exception specified in theory data.
e.g. ` ExpectedException = ExpectedException.SecurityTokenSignatureKeyNotFoundException()`, for this scenario.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
brentschmaltz: Can we use StringComparison.OrdinalCaseInsensitive?

GeoK: We can relax this comparison but the parameter values are case-sensitive [by the standard](https://tools.ietf.org/html/rfc7517#section-4.2).

SaltyDH:
Would love to see this merged in, more and more services are generating ECD keys.

GeoK:
Closing this pull request in favor of #1143
Reason: Too many conflicts. We want this change in 5.5.0 and this PR is based on our dev branch.

PR #1079, opened 2018-12-15
Added a new API to JsonWebTokenHandler that takes in a SecurityTokenDescriptor

mafurman:
- The '_ClaimsIdentity Subject_' property on SecurityTokenDescriptor has been marked as obsolete.
- Also added a new '_Dictionary<string, object> Payload_' property to SecurityTokenDescriptor that will eventually replace the '_ClaimsIdentity Subject_' property.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -25,14 +25,16 @@
 //
 //------------------------------------------------------------------------------
 
-using Microsoft.IdentityModel.Logging;
-using Microsoft.IdentityModel.Tokens;
-using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Security.Claims;
 using System.Text;
 using System.Text.RegularExpressions;
+using Microsoft.IdentityModel.Logging;
+using Microsoft.IdentityModel.Tokens;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using TokenLogMessages = Microsoft.IdentityModel.Tokens.LogMessages;
 
 namespace Microsoft.IdentityModel.JsonWebTokens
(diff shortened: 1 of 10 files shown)


src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
```diff
@@ -65,6 +71,11 @@ public class SecurityTokenDescriptor
         /// </summary>
         public DateTime? NotBefore { get; set; }
 
+        /// <summary>
+        /// Gets or sets the <see cref="Dictionary{TKey, TValue}"/> which represents the JWT token payload.
+        /// </summary>
+        public Dictionary<string, object> Payload { get; set; }
```
CBaud: Setting the type of the Payload property to be an IDictionary<string, object> would help to decouple the contract of the SecurityTokenDescriptor type from the underlying implementation.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -117,6 +118,25 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
+        /// Creates a JWS(Json Web Signature).
+        /// </summary>
+        /// <param name="tokenDescriptor">A <see cref="SecurityTokenDescriptor"/> that contains details of contents of the token.</param>
+        /// <returns>A JWS in Compact Serialization Format.</returns>
+        public virtual string CreateToken(SecurityTokenDescriptor tokenDescriptor)
+        {
+            if (tokenDescriptor == null)
+                throw LogHelper.LogArgumentNullException(nameof(tokenDescriptor));
+
+            if (tokenDescriptor.Payload == null || tokenDescriptor.Payload.Count == 0)
```
CBaud: I would recommend using the Any() linq expression (e.g. || !tokenDescriptor.Payload.Any()) because it will enumerate at most one item in the collection. The Count property is cheap if the collection has already been evaluated, but will otherwise require all items in the collection to be enumerated which could be expensive if the collection is large.

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
brentschmaltz: @CBaud 	Thanks for chiming in. Yes IDictionary...

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: @mafurman 	@CBaud 	+1. Yes count requires the underlying structure to sometimes do a lot of work.

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
mafurman: Yes, this makes sense.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: I didn't realize this. Good suggestion! 👍

build/common.props
```diff
@@ -20,6 +20,7 @@
     <PackageLicense>https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/LICENSE.txt</PackageLicense>
     <SignAssembly>true</SignAssembly>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <WarningsNotAsErrors>0618</WarningsNotAsErrors>
```
brentschmaltz: this is OK in test, but for src use individual.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -141,6 +141,70 @@ public static TheoryData<CreateTokenTheoryData> CreateJWETheoryData
...
+            }
+            catch (Exception ex)
+            {
+                theoryData.ExpectedException.ProcessException(ex, context);
+            }
+
+            TestUtilities.AssertFailIfErrors(context);
+        }
+
+
```
brentschmaltz: extra line

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -117,6 +119,25 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
         }
 
+        /// <summary>
+        /// Creates a JWS(Json Web Signature).
+        /// </summary>
+        /// <param name="tokenDescriptor">A <see cref="SecurityTokenDescriptor"/> that contains details of contents of the token.</param>
+        /// <returns>A JWS in Compact Serialization Format.</returns>
+        public virtual string CreateToken(SecurityTokenDescriptor tokenDescriptor)
+        {
+            if (tokenDescriptor == null)
```
brentschmaltz: Can we added tests for SecurityTokenDescriptor properties.

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
```diff
@@ -40,6 +41,11 @@ public class SecurityTokenDescriptor
         /// </summary>
         public string Audience { get; set; }
 
+        /// <summary>
+        /// Defines the compression algorithm that will be used to compress the JWT token payload.
```
brentschmaltz: indenting...

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
```diff
@@ -40,6 +41,11 @@ public class SecurityTokenDescriptor
         /// </summary>
         public string Audience { get; set; }
 
+        /// <summary>
```
brentschmaltz: We should take into account the properties such as Audience when creating the payload.
Give priority to SecurityTokenDescriptor is a claim is found of the same type.
May want to log 'information' that a duplicate exists.

Continue accounting for 'SetExpirationTimeOnTokenCreation' flag.

build/common.props
mafurman: Makes sense!

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
mafurman: Woops. Fixed this.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Done!

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
mafurman: I can't see any indenting issues in Visual Studio...

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
mafurman: Done. I also added in tests to make sure that the properties on SecurityTokenDescriptor are being used as expected.

src/Microsoft.IdentityModel.JsonWebTokens/LogMessages.cs
```diff
@@ -51,10 +51,11 @@ internal static class LogMessages
         internal const string IDX14107 = "IDX14107: Token string does not match the token formats: JWE (header.encryptedKey.iv.ciphertext.tag) or JWS (header.payload.signature)";
         internal const string IDX14111 = "IDX14111: JWT: '{0}' must have three segments (JWS) or five segments (JWE).";
         internal const string IDX14112 = "IDX14112: Only a single 'Actor' is supported. Found second claim of type: '{0}', value: '{1}'";
-
+        internal const string IDX14113 = "IDX14113: A duplicate value for 'SecurityTokenDescriptor.{0}' exists in 'SecurityTokenDescriptor.Payload'.";
```
brentschmaltz: Can you add some text to say the value from the SecurityTokenDescriptor was used.

mafurman: Done. 👍

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -129,13 +129,13 @@ public virtual string CreateToken(SecurityTokenDescriptor tokenDescriptor)
             if (tokenDescriptor == null)
                 throw LogHelper.LogArgumentNullException(nameof(tokenDescriptor));
 
-            if (tokenDescriptor.Payload == null || !tokenDescriptor.Payload.Any())
-                throw LogHelper.LogArgumentNullException(nameof(tokenDescriptor.Payload));
+            if (tokenDescriptor.Claims == null || !tokenDescriptor.Claims.Any())
```
brentschmaltz: We don't need to throw, we should log at the warning level.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -129,13 +129,13 @@ public virtual string CreateToken(SecurityTokenDescriptor tokenDescriptor)
...
+                throw LogHelper.LogArgumentNullException(nameof(tokenDescriptor.Claims));
 
             if (tokenDescriptor.SigningCredentials == null)
                 throw LogHelper.LogArgumentNullException(nameof(tokenDescriptor.SigningCredentials));
 
-            var payload = new Dictionary<string, object>(tokenDescriptor.Payload);
+            var payload = new Dictionary<string, object>(tokenDescriptor.Claims);
```
brentschmaltz: Can we open a bug to 'remember' what the actual audience should be and then we don't have to copy the dictionary.

PR #1084, opened 2018-12-19
Changed the X509SecurityKey.KeyId property so that it's equal to the X5t property

mafurman:
Other changes:

- JsonWebTokenHandler now adds the 'x5t' parameter to the header by default if the security key being used is an X509SecurityKey.
- The IdentityComparer can now be used to compare boolean values.
- Fixed some of JsonWebTokenHandler tests that had incorrect TheoryData.


brentschmaltz:
:shipit:

PR #1102, opened 2019-01-24
Add PropertyBag to CallContext

GeoK:


src/Microsoft.IdentityModel.Tokens/CallContext.cs
```diff
@@ -69,5 +70,10 @@ public CallContext(Guid activityId)
         /// The collection of logs associated with a request. Use <see cref="CaptureLogs"/> to control capture.
         /// </summary>
         public ICollection<string> Logs { get; private set; } = new Collection<string>();
+
+        /// <summary>
+        /// Gets or sets an <see cref="IDictionary{String, Object}"/> that enables custom extensibility scenarios.
+        /// </summary>
+        public IDictionary<string, object> AdditionalProperties { get; set; }
```
brentschmaltz: Can we name it PropertyBag so it is the same as TokenValidationParameters?

GeoK: @brentschmaltz Sure. I'll rename the property and merge to the dev branch, if there are no other suggestions.

brentschmaltz: :shipit:

PR #1110, opened 2019-02-06
Updated build.ps1 so that VerifyResourceUsage.pl is run at build time

mafurman:
Our local build will now fail if VerifyResourceUsage.pl fails. 

This will make it easier for us to ensure that we have no duplicate, reused, or improperly formatted log message ids.


brentschmaltz:
:shipit:

PR #1113, opened 2019-02-08
Add JsonExtensions project and Newtonsoft adapter

GeoK:
work in progress


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -248,6 +248,11 @@ public virtual IEnumerable<Claim> Claims
         /// <remarks>If the 'iss' claim is not found, an empty string is returned.</remarks>   
         public override string Issuer => Payload.Value<string>(JwtRegisteredClaimNames.Iss) ?? string.Empty;
 
+        /// <summary>
+        /// 
+        /// </summary>
+        public virtual IJsonConvertible JsonAdapter { get; set; }
```
brentschmaltz: To keep the API general, we may want to use more general name such IConvertible. We may want to think if this fits the Factory pattern, where the factory knows how to create object T, from the JSON retrieved.

CBaud: Consider [System.Converter<Tinput, TOutput>](https://docs.microsoft.com/en-us/dotnet/api/system.converter-2). It represents a method that converts an object from one type to another type.

---
In reply to: [255131555](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1113#discussion_r255131555) [](ancestors = 255131555)

CBaud: Also note that [IConvertible](https://docs.microsoft.com/en-us/dotnet/api/system.iconvertible) already exists in .NET and may cause conflicts for consumers of this library. Consider something unique like `IAdapter` or `IConverter`.

---
In reply to: [255136705](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1113#discussion_r255136705) [](ancestors = 255136705,255131555)

Issue #1125, opened 2019-03-21
JwtRegisteredClaimNames invalid references to RFC7519


duki994:
### Describe the bug
Invalid JwtRegisteredClaimNames and invalid references to RFC7519 standard documentation.

#### Impersonalization (impersonation) login
I wanted to use default actor claim validation but it was failing since I implemented [OAuth Token Exchange Draft](https://tools.ietf.org/html/draft-ietf-oauth-token-exchange-16)

#### Finding unintended bug
I traced this a while, and found out that logic was different so had to implement my own case.
While tracing I found many claim types pointing to [RFC7519 Section 4](http://tools.ietf.org/html/rfc7519#section-4), but they don't exist there.

### To Reproduce
Open JwtRegisteredClaimNames (Go To Definition) class in IDE or editor of choice and 
compare to [RFC7519 Section 4](http://tools.ietf.org/html/rfc7519#section-4)

You will see wrong references and non existing claim types that refer to RFC7519 Section 4.
### Expected behavior
To be in sync with standard

### Screenshots
![screenshot](https://user-images.githubusercontent.com/9515864/54754570-15b01a00-4be4-11e9-952b-b5878f825111.png)

### Additional context
Versions: All up to latest 5.4.0 from [System.IdentityModel.Tokens.Jwt NuGet package](https://www.nuget.org/packages/System.IdentityModel.Tokens.Jwt)

brentschmaltz:
@duki994 yes, actort is not a in the spec.
If we move it now, that would be breaking.

What would you like us to do?

duki994:
@brentschmaltz
In that case just remove link towards standard from comment to not mislead developers.

I think that also "typ" claim is for JOSE header, and that is section 5 of RFC7519.

I can make a pull request, and then you and team review it :)

PR #1126, opened 2019-03-23
Sync JWT claim names with current RFC7519 standard and OpenID Connect 1.0 standard

duki994:
Work on issue #1125 

@brentschmaltz 
Changing link references in comments is easiest way to keep up with standard.
Especially since Actort claims needs to be there for backwards compatibility.

Esentially, only code docs are changed

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtRegisteredClaimNames.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtRegisteredClaimNames.cs
@@ -35,7 +35,6 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     public struct JwtRegisteredClaimNames
     {
         /// <summary>
-        /// http://tools.ietf.org/html/rfc7519#section-4
         /// </summary>
         public const string Actort = "actort";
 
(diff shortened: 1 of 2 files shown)


duki994:
Please leave any suggestions for doc comments about claims that don't exist in standard or are there for backwards compatibility with other authentication and authorization system.

brentschmaltz:
@duki994 I am stumped on what comment to leave for the errors. 
Best I can think of is: "Common claims found used by IdentityProviders, but not currently registered".

src/Microsoft.IdentityModel.JsonWebTokens/JwtRegisteredClaimNames.cs
```diff
@@ -135,7 +133,6 @@ public struct JwtRegisteredClaimNames
         public const string Nbf = "nbf";
 
         /// <summary>
-        /// http://tools.ietf.org/html/rfc7519#section-4
         /// </summary>
         public const string Prn = "prn";
```
mafurman: Could you please include this as the summary comment for the claims that aren't registered?

"Common claim type that's not currently registered."

duki994: Yeah. No problem. Will include those

brentschmaltz: @duki994 what type of guitar do you play? Do you have a favorite?

brentschmaltz:
@duki994 thanks for your contribution :-)

PR #1127, opened 2019-03-28
 Modified the M.IM.Protocols.OpenIdConnect assembly to use M.IM.JsonWebTokens instead of S.IM.Tokens.Jwt

mafurman:
Rather than incorporating these changes directly, we may also want to consider creating a new OpenIdConnect assembly that uses M.IM.JsonWebTokens instead of S.IM.Tokens.Jwt.

Addresses #1160.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -40,7 +40,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// <summary>
     /// A <see cref="SecurityToken"/> designed for representing a JSON Web Token (JWT). 
     /// </summary>
-    public class JsonWebToken : SecurityToken
+    public class JsonWebToken : SecurityToken, IJsonWebToken
     {
         /// <summary>
         /// Initializes a new instance of <see cref="JsonWebToken"/> from a string in JWS or JWE Compact serialized format.
@@ -146,7 +146,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// <summary>
         /// Gets a <see cref="IEnumerable{Claim}"/><see cref="Claim"/> for each JSON { name, value }.
         /// </summary>
-        public virtual IEnumerable<Claim> Claims
+        public IEnumerable<Claim> Claims
         {
             get
             {
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
@@ -71,12 +71,12 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         /// <summary>
         /// This id_token is assumed to have audience, issuer, lifetime and signature validated.
         /// </summary>
-        [Obsolete("The 'ValidatedIdToken' property is obsolete. Please use 'ValidatedJsonWebToken' instead.")]
+        [Obsolete("The 'ValidatedIdToken' property is obsolete. Please use 'ValidatedJwtToken' instead.")]
         public JwtSecurityToken ValidatedIdToken { get; set; }
 
         /// <summary>
         /// This JWT security token is assumed to have audience, issuer, lifetime and signature validated.
         /// </summary>
-        public JsonWebToken ValidatedJsonWebToken { get; set; }
+        public IJsonWebToken ValidatedJsonWebToken { get; set; }
     }
 }
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
@@ -25,8 +25,7 @@
 //
 //------------------------------------------------------------------------------
 
-using System;
-using System.IdentityModel.Tokens.Jwt;
+using Microsoft.IdentityModel.Tokens;
 
 namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
 {
@@ -69,6 +68,6 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         /// <summary>
         /// This id_token is assumed to have audience, issuer, lifetime and signature validated.
         /// </summary>
-        public JwtSecurityToken ValidatedIdToken { get; set; }
+        public SecurityToken ValidatedIdToken { get; set; }
     }
 }
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
@@ -25,6 +25,9 @@
 //
 //------------------------------------------------------------------------------
 
+using System;
+using System.IdentityModel.Tokens.Jwt;
+using Microsoft.IdentityModel.JsonWebTokens;
 using Microsoft.IdentityModel.Tokens;
 
 namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
@@ -68,6 +71,12 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         /// <summary>
         /// This id_token is assumed to have audience, issuer, lifetime and signature validated.
         /// </summary>
-        public SecurityToken ValidatedIdToken { get; set; }
+        [Obsolete("The 'ValidatedIdToken' property is obsolete. Please use 'ValidatedJsonWebToken' instead.")]
+        public JwtSecurityToken ValidatedIdToken { get; set; }
+
+        /// <summary>
+        /// This JWT security token is assumed to have audience, issuer, lifetime and signature validated.
+        /// </summary>
+        public JsonWebToken ValidatedJsonWebToken { get; set; }
     }
 }
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
@@ -42,11 +42,11 @@ using JwtRegisteredClaimNames = Microsoft.IdentityModel.JsonWebTokens.JwtRegiste
 namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
 {
     /// <summary>
-    /// Delegate for validating additional claims in 'id_token' (of type <see cref="JsonWebToken"/>).
+    /// Delegate for validating additional claims in 'id_token' (of type <see cref="IJsonWebToken"/>).
     /// </summary>
-    /// <param name="idToken"><see cref="JsonWebToken"/> to validate</param>
+    /// <param name="idToken"><see cref="IJsonWebToken"/> to validate</param>
     /// <param name="context"><see cref="OpenIdConnectProtocolValidationContext"/> used for validation</param>
-    public delegate void JsonWebTokenValidator(JsonWebToken idToken, OpenIdConnectProtocolValidationContext context);
+    public delegate void JsonWebTokenValidator(IJsonWebToken idToken, OpenIdConnectProtocolValidationContext context);
 
     /// <summary>
     /// Delegate for validating additional claims in 'id_token' (of type <see cref="JwtSecurityToken"/>).
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
@@ -441,7 +441,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
                 if (idToken.IssuedAt.Equals(DateTime.MinValue))
                     throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, JwtRegisteredClaimNames.Iat.ToLowerInvariant(), idToken)));
 
-                if (idToken.Issuer == null)
+                if (idToken.Issuer.Equals(string.Empty))
                     throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, JwtRegisteredClaimNames.Iss.ToLowerInvariant(), idToken)));
 
                 // sub is required in OpenID spec; but we don't want to block valid idTokens provided by some identity providers
@@ -665,8 +665,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
             }
             else // validationContext.ValidatedIdToken != null
             {
-                object cHashClaim;
-                if (!validationContext.ValidatedIdToken.Payload.TryGetValue(JwtRegisteredClaimNames.CHash, out cHashClaim))
+                if (!validationContext.ValidatedIdToken.Payload.TryGetValue(JwtRegisteredClaimNames.CHash, out var cHashClaim))
                     throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidCHashException(LogHelper.FormatInvariant(LogMessages.IDX21307, validationContext.ValidatedIdToken)));
                 chash = cHashClaim as string;
                 if (chash == null)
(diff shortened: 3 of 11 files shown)


henrik-me:
Eventhough the type is disappearing, I'm thinking we should see if it's possible to add a piece of migration help that will provide a compiler error with meaningful details (e.g. by replacing the jwtSecurityToken type with one of type object). For more details on how to write migration aid pls. take a look here: [https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/dev3x/src/Microsoft.Identity.Client/MigrationAid/MigrationAid.cs](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/dev3x/src/Microsoft.Identity.Client/MigrationAid/MigrationAid.cs)

henrik-me (dismissed):
:clock1:

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
```diff
@@ -69,6 +68,6 @@ public class OpenIdConnectProtocolValidationContext
         /// <summary>
         /// This id_token is assumed to have audience, issuer, lifetime and signature validated.
         /// </summary>
-        public JwtSecurityToken ValidatedIdToken { get; set; }
+        public SecurityToken ValidatedIdToken { get; set; }
```
GeoK: I'm wondering if ValidateIdToken should be JsonWebToken.
We are treating ValidateIdToken as JsonWebToken:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/607218ba549e4a00b8f8a078c25a1e76e7ce019d/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs#L296

Same comment applies to IdTokenValidator.

mafurman: That's a good point. Since it's always guaranteed to be a JsonWebToken that might make sense.

I was thinking that SecurityToken might be more future-proof as we wouldn't need to update the property type if we decided to use a different token type in the future. However, I'm not sure how likely this scenario is.

We actually do something similar in TokenValidationResult:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/7c7c7e13968e6656e67eb43938d4da136336049c/src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs#L58-L61

We always cast SecurityToken to a JsonWebToken since it's only ever used by the JsonWebTokenHandler.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -29,22 +29,31 @@
...
     /// <param name="context"><see cref="OpenIdConnectProtocolValidationContext"/> used for validation</param>
-    public delegate void IdTokenValidator(SecurityToken idToken, OpenIdConnectProtocolValidationContext context);
+    public delegate void JsonWebTokenValidator(JsonWebToken idToken, OpenIdConnectProtocolValidationContext context);
+
+    /// <summary>
+    /// Delegate for validating additional claims in 'id_token' (of type <see cref="JwtSecurityToken"/>).
+    /// </summary>
+    /// <param name="idToken"><see cref="JwtSecurityToken"/> to validate</param>
+    /// <param name="context"><see cref="OpenIdConnectProtocolValidationContext"/> used for validation</param>
+    public delegate void IdTokenValidator(JwtSecurityToken idToken, OpenIdConnectProtocolValidationContext context);
```
GeoK: Consider marking this one as obsolete too.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
```diff
@@ -69,6 +71,12 @@ public class OpenIdConnectProtocolValidationContext
         /// <summary>
         /// This id_token is assumed to have audience, issuer, lifetime and signature validated.
         /// </summary>
+        [Obsolete("The 'ValidatedIdToken' property is obsolete. Please use 'ValidatedJsonWebToken' instead.")]
```
GeoK: Perhaps we can explain here that ValidatedIdToken is related to IdTokenValidator and that ValidatedJsonWebToken/JsonWebTokenValidator combination should be used going forward.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
mafurman: Good suggestion! I just added this in.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
mafurman: Good idea. :) I added this explanation on the IdTokenValidator obsolete attribute, but I can put it in this obsolete attribute as well.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
brentschmaltz: @GeoK @mafurman could we have a more generalized calling pattern that receives'json'. The validator inspects the oidc reponse message received against the 'json' which represents a validated idtoken. We assume the json was validated.
I am wondering if the OpenIdConnectProtocolValidator could perform all validation functions including signature.

brentschmaltz: This would simplify the calling pattern for runtimes such as asp.net.

GeoK: @brentschmaltz If I'm understanding this correctly, a method would receive `OpenIDConnectMessage` (authorization response), TokenValidationParameters, and other properties, and it would validate an Id Token from an OIDC message and OpenIdConnectProtocol, in the same call. 

I agree that the calling pattern would be simplified, but I'm not sure if this is a right place for such a change as the class name is OpenIdConnect**Protocol**Validator.

We will probably have to account for TokenValidatedEvent which ASP.NET has, that is invoked after IdToken validation and before OIDC protocol validation.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectProtocolValidatorJsonWebTokenTests.cs
```diff
@@ -0,0 +1,1700 @@
...
+        }
+
+
+        [Theory, MemberData(nameof(ValidateAuthenticationResponseTheoryData))]
+        public void ValidateAuthenticationResponse(OidcProtocolValidatorJsonWebTokenTheoryData theoryData)
+        {
+            TestUtilities.WriteHeader($"{this}.ValidateAuthenticationResponse", theoryData);
+            try
+            {
+                theoryData.ProtocolValidator.ValidateAuthenticationResponse(theoryData.ValidationContext);
```
GeoK: I don't think that we have a test for ValidateAuthenticationResponse method which will test the whole method, including Validate State/IdToken/Nonce.. method calls.
It looks like we didn't have such a test earlier, but it would be valuable to have it, imo.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
brentschmaltz: @GeoK about the TokenValidedEvent, we may want to think about when it is called. It the token has arrived in an invalid protocol message, should we consider it a valid id_token?

This is an issue we should bring up with asp.net, users may be making decisions on this token being valid.

Let's back off validation for a moment. 
Could we validate the protocol with just json?

brentschmaltz: If you look at issue https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1131
It is coded up:
```
OnTokenValidated = context =>
{
    var claims = context.SecurityToken.Claims;
    return Task.CompletedTask;
}
```

GeoK: @brentschmaltz So we would add a string (json) representation of an id token, instead of a JsonWebToken, to OpenIdConnectProtocolValidationContext class?
That design would be more future-proof. The drawback is that we would have to parse the json string which introduces some overhead, compared to our current design. I'm not saying that we shouldn't do it, just trying to point out pros/cons.

GeoK: > 
> 
> If you look at issue #1131
> It is coded up:
> 
> ```
> OnTokenValidated = context =>
> {
>     var claims = context.SecurityToken.Claims;
>     return Task.CompletedTask;
> }
> ```

I don't think that this is a production code. My guess is that a client was debugging and trying to figure out if it could get the claims from an access token.

brentschmaltz: @GeoK We could use a string or define an interface IJsonWebToken. That would be OK as OIDC requires a JWT.

brentschmaltz: @GeoK My thoughts about OnTokenValidated are centered on how / when should we report the jwt in the OIDC response as valid? A jwt arriving over 'Beraer' as an access_token can be valid by checking iss, aud, signature etc. But a jwt arriving as part of a OIDC response should trusted be used if the 'nonce' is not valid.

Are we sending the wrong message here.

GeoK: @brentschmaltz  - This event represents an extensibility point that allows customers to put some custom logic after the id token was validated. Only in a case when a customer explicitly populates context.Result (HandleRequestResult) in their custom logic, OIDC protocol will **_not_** be validated.
Otherwise, `ValidateAuthenticationResponse` method will be called.

Perhaps OnTokenValidated event should be invoked _after_  OIDC protocol validation.

I'm wondering if there are users who use this event to prepare/set the nonce to `OpenIdConnectProtocolValidationContext`. They would be broken if OnTokenValidated event is not invoked before OIDC protocol validation.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -358,9 +401,91 @@ protected virtual void ValidateIdToken(OpenIdConnectProtocolValidationContext va
...
+                if (!idToken.Audiences.Any())
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, JwtRegisteredClaimNames.Aud.ToLowerInvariant(), idToken)));
+
+                if (idToken.ValidTo.Equals(DateTime.MinValue))
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, JwtRegisteredClaimNames.Exp.ToLowerInvariant(), idToken)));
+
+                if (idToken.IssuedAt.Equals(DateTime.MinValue))
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, JwtRegisteredClaimNames.Iat.ToLowerInvariant(), idToken)));
+
+                if (idToken.Issuer == null)
```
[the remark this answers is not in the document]
mafurman: Good catch! This was left over from the JwtSecurityToken implementation and the test that was supposed to check this exception was being thrown actually wasn't written properly.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -532,23 +657,31 @@ protected virtual void ValidateCHash(OpenIdConnectProtocolValidationContext vali
...
+                if (!validationContext.ValidatedJsonWebToken.TryGetPayloadValue(JwtRegisteredClaimNames.CHash, out  chash))
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidCHashException(LogHelper.FormatInvariant(LogMessages.IDX21307, validationContext.ValidatedJsonWebToken)));
             }
-
-            var chash = cHashClaim as string;
-            if (chash == null)
+            else // validationContext.ValidatedIdToken != null
             {
-                throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidCHashException(LogHelper.FormatInvariant(LogMessages.IDX21306, validationContext.ValidatedIdToken)));
+                object cHashClaim;
```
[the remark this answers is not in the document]
mafurman: 👍

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -582,17 +716,31 @@ protected virtual void ValidateAtHash(OpenIdConnectProtocolValidationContext val
...
+                if (!validationContext.ValidatedJsonWebToken.TryGetPayloadValue(JwtRegisteredClaimNames.AtHash, out atHash))
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidAtHashException(LogHelper.FormatInvariant(LogMessages.IDX21312, validationContext.ValidatedJsonWebToken)));
 
-            var atHash = atHashClaim as string;
-            if (atHash == null)
-                throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidAtHashException(LogHelper.FormatInvariant(LogMessages.IDX21311, validationContext.ValidatedIdToken)));
+            }
+            else // validationContext.ValidatedIdToken != null
+            {
+                object atHashClaim;
```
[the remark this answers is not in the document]
mafurman: 👍

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectProtocolValidatorTests.cs
```diff
@@ -226,7 +226,7 @@ public static TheoryData<OidcProtocolValidatorTheoryData> ValidateAuthentication
                         }
                     }
                 });
-
+
```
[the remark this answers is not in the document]
mafurman: Good idea. I didn't do this earlier as it was being treated as a warning and not an error in the tests, but it is a lot cleaner this way.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectProtocolValidatorJsonWebTokenTests.cs
mafurman: I added two tests (one using JwtSecurityToken and the other using JsonWebToken) that test this entire method.

mafurman:
I just added the ISecurityToken and IJsonWebToken interfaces. Assuming we want to keep the TryGetPayloadValue() and similar methods on the IJsonWebToken interface, I also plan on adding in tests to ensure those are implemented properly by JwtSecurityToken.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -154,7 +154,7 @@ public IEnumerable<string> Audiences
         /// <summary>
         /// Gets a <see cref="IEnumerable{Claim}"/><see cref="Claim"/> for each JSON { name, value }.
         /// </summary>
-        public virtual IEnumerable<Claim> Claims
```
henrik-me: >virtual [](start = 15, length = 7)

This is technically a breaking change in the public API, while small someone might have made an override of this method. This will cause some version update requirements etc... alternatively can the virtual stay?

henrik-me:
I'm seeing a few changes here which at a minumum calls out for a minor version change (already part of the plan), however I also see a few which would following the semantic versioning, would require us to update the major version. I suggest we stick to changes that will not require a major version bump.

mafurman:
Closing this PR as this change will likely require creation of a new assembly; also at this point we're better off creating a new PR rather than trying to rebase this one.

PR #1133, opened 2019-04-19
Skip to a content node before writing

GeoK:
Once `RequestedSecurityToken` element/node is found, it's written/extracted into a token.
Current node might not be a content node, so the reader should skip ahead to the next content node.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/1300b19eae77801a05906bc061fe52e158b513e8/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs#L286-L301


src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
```diff
@@ -285,6 +285,9 @@ public virtual string GetTokenUsingXmlReader()
 
                                 // <RequestedSecurityToken>
                                 xmlReader.ReadStartElement();
+                                // skip to the next content node
+                                xmlReader.MoveToContent();
```
henrik-me: Shouldn't there be some sort of detection on whether it's the right already (the PR description eludes at that but doesn't fully explain)? At least  consider updating the comment to say why this is needed, right now the comment is just saying what the method name already implies (thus the comment can be removed).

GeoK: `MoveToContent()` checks/detects if the current node is a content node, so no additional checks are needed.
I'll expand the comment to describe the rationale behind the change.

henrik-me: signing off, however this comment should be improved (I was also missing a link to a work item describing the reason for this issue)

---
In reply to: [277361859](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1133#discussion_r277361859) [](ancestors = 277361859)

PR #1137, opened 2019-04-23
Skip to a content node before writing

GeoK:
Cherry-picked from the dev5x branch.

Once `RequestedSecurityToken` element/node is found, it's written/extracted into a token.
Current node might not be a content node, so the reader should skip ahead to the next content node.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/1300b19eae77801a05906bc061fe52e158b513e8/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs#L286-L301


GeoK:
Verbally approved by @henrik-me and @brentschmaltz.

PR #1138, opened 2019-04-25
Made NS 2.0 adjustments

davidfowl:
Reduce the netstandard 2.0 dependency graph

PS: I'm not sure why **System.Security.Cryptography.Cng** was marked as PrivateAssets="All", that seems broken. I didn't run tests but I installed the packages build to verify the graph was indeed smaller. Hopefully your CI runs tests to verify I didn't break anything 😄

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
+++ b/src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
@@ -39,7 +39,7 @@
     <ProjectReference Include="..\Microsoft.IdentityModel.Logging\Microsoft.IdentityModel.Logging.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.4' Or '$(TargetFramework)' == 'netstandard2.0' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.4'">
     <PackageReference Include="Microsoft.CSharp" Version="4.3.0" />
     <PackageReference Include="System.Collections" Version="4.3.0-*" />
     <PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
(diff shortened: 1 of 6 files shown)


src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
```diff
@@ -62,6 +62,11 @@
     <PackageReference Include="System.Threading" Version="4.3.0-*" />
     <PackageReference Include="System.Xml.XmlDocument" Version="4.3.0" />
   </ItemGroup>
+  
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' ">
+    <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
```
GeoK: Not really sure why this package reference is needed. Could you please explain?

GeoK:
Hi @davidfowl, thank you for the PR.
Our build complains that there is a package downgrade for System.Security.Cryptography.Cng (4.5.0 to 4.3.0) in a couple of our test projects.

As these references are actually not needed, please feel free to remove them.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/72f03a9c62215688b759ea180263db8b498280ea/test/Microsoft.IdentityModel.Tokens.Tests/Microsoft.IdentityModel.Tokens.Tests.csproj#L38-L41

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/72f03a9c62215688b759ea180263db8b498280ea/test/Microsoft.IdentityModel.TestUtils/Microsoft.IdentityModel.TestUtils.csproj#L34-L36

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
davidfowl: Are you using dynamic anywhere?

GeoK: Oh, got it. Thank you.

davidfowl:
I’ll send an update

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
davidfowl: You should just remove the use of dynamic though.

davidfowl:
Updated a couple of tests. I can't run build.cmd. I tried to run some of the tests but some ran and some failed with weird binding errors. There's no PR checks enabled on this repo so it makes it hard to know if it passes the build.

GeoK:
Thank you for the update. I'm guessing that strong name validation fails as the assemblies are delay signed. We have two scripts inside the build folder to bypass strong name validation.

We've just enabled PR validation.

davidfowl:
@GeoK is this good to go?

brentschmaltz:
:shipit:  @davidfowl what errors are you getting from build.cmd?

PR #1141, opened 2019-04-25
JsonWebTokenHandler.CreateToken(SecurityTokenDescriptor) no longer copies SecurityTokenDescriptor.Claims into a new Dictionary

### LINKED ISSUE #1082: Keep track of what the actual Audience, Issuer, etc. are when creating a token using a SecurityTokenDescriptor in JsonWebTokenHandler
The CreateToken(SecurityTokenDescriptor tokenDescriptor) method is currently copying the dictionary of claims in order to update its values with the properties that were set on SecurityTokenDescriptor. 

This behavior should be modified because copying could potentially take a long time.

mafurman:
Fixes #1082.

Previously we were copying SecurityTokenDescriptor.Claims into a new Dictionary, updating that copy with the other properties set on the SecurityTokenDescriptor (e.g. SecurityTokenDescriptor.Audience), and then converting that dictionary into a JObject.

Now we create a JObject from SecurityTokenDescriptor.Claims directly and update it with the other properties set on the SecurityTokenDescriptor (so creating a Dictionary copy is not necessary).


brentschmaltz:
:shipit:

PR #1143, opened 2019-04-29
Support JsonWebKeys with the Elliptic Curve key type

GeoK:
* Refactored JsonWebKeySet.GetSigningKeys() method and added support for JWKs with 'kty' = 'EC'.
* Unresolved JsonWebKeys will be skipped by default. `SkipUnresolvedJsonWebKeys` flag controls whether unresolved JsonWebKeys will be included in the resulting collection of GetSigningKeys method.
* Adding only the first certificate from the x5c certificate chain as a SecurityKey.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
+++ b/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
@@ -76,7 +76,7 @@ namespace Microsoft.IdentityModel.Tokens
             {
                 ECDsaAdapter = new ECDsaAdapter();
             }
-            catch (PlatformNotSupportedException ex)
+            catch (Exception ex)
             {
                 LogHelper.LogExceptionMessage(ex);
             }
@@ -162,6 +162,10 @@ namespace Microsoft.IdentityModel.Tokens
                 if (!(string.IsNullOrWhiteSpace(webKey.Use) || webKey.Use.Equals(JsonWebKeyUseNames.Sig, StringComparison.Ordinal)))
                 {
                     LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10808, webKey.KeyId ?? "" , webKey.Use));
+
+                    if (!SkipUnresolvedJsonWebKeys)
+                        signingKeys.Add(webKey);
+
                     continue;
                 }
 
--- a/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
+++ b/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
@@ -161,7 +161,7 @@ namespace Microsoft.IdentityModel.Tokens
                 // https://tools.ietf.org/html/rfc7517#section-4.2
                 if (!(string.IsNullOrWhiteSpace(webKey.Use) || webKey.Use.Equals(JsonWebKeyUseNames.Sig, StringComparison.Ordinal)))
                 {
-                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10808, webKey.KeyId ?? "" , webKey.Use));
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10808, webKey.KeyId ?? "null" , webKey.Use ?? "null"));
 
                     if (!SkipUnresolvedJsonWebKeys)
                         signingKeys.Add(webKey);
@@ -187,7 +187,7 @@ namespace Microsoft.IdentityModel.Tokens
 
                     if (!isResolved)
                     {
-                        LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10810, webKey.KeyId ?? ""));
+                        LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10810, webKey.KeyId ?? "null"));
 
                         if(!SkipUnresolvedJsonWebKeys)
                             signingKeys.Add(webKey);
@@ -200,7 +200,7 @@ namespace Microsoft.IdentityModel.Tokens
                 else
                 {
                     // kty is not 'EC' or 'RSA'
-                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10809, webKey.Kty ?? ""));
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10809, webKey.Kty ?? "null"));
 
                     if (!SkipUnresolvedJsonWebKeys)
                         signingKeys.Add(webKey);
--- a/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
+++ b/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
@@ -42,6 +42,8 @@ namespace Microsoft.IdentityModel.Tokens
     [JsonObject]
     public class JsonWebKeySet
     {
+        private readonly bool _skipUnresolvedJsonWebKeys = SkipUnresolvedJsonWebKeys;
+
         /// <summary>
         /// Returns a new instance of <see cref="JsonWebKeySet"/>.
         /// </summary>
@@ -123,15 +125,15 @@ namespace Microsoft.IdentityModel.Tokens
         }
 
         /// <summary>
-        /// This adapter abstracts the <see cref="ECDsa"/> differences between versions of .Net targets.
-        /// </summary>
-        internal static ECDsaAdapter ECDsaAdapter;
-
-        /// <summary>
         /// When deserializing from JSON any properties that are not defined will be placed here.
         /// </summary>
         [JsonExtensionData]
         public virtual IDictionary<string, object> AdditionalData { get; } = new Dictionary<string, object>();
+
+        /// <summary>
+        /// This adapter abstracts the <see cref="ECDsa"/> differences between versions of .Net targets.
+        /// </summary>
+        internal static ECDsaAdapter ECDsaAdapter;
 
         /// <summary>
         /// Gets the <see cref="IList{JsonWebKey}"/>.
@@ -163,7 +165,7 @@ namespace Microsoft.IdentityModel.Tokens
                 {
                     LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10808, webKey.KeyId ?? "null" , webKey.Use ?? "null"));
 
-                    if (!SkipUnresolvedJsonWebKeys)
+                    if (!_skipUnresolvedJsonWebKeys)
                         signingKeys.Add(webKey);
 
                     continue;
@@ -189,7 +191,7 @@ namespace Microsoft.IdentityModel.Tokens
                     {
                         LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10810, webKey.KeyId ?? "null"));
 
-                        if(!SkipUnresolvedJsonWebKeys)
+                        if(!_skipUnresolvedJsonWebKeys)
                             signingKeys.Add(webKey);
                     }
                 }
@@ -202,7 +204,7 @@ namespace Microsoft.IdentityModel.Tokens
                     // kty is not 'EC' or 'RSA'
                     LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10809, webKey.Kty ?? "null"));
 
-                    if (!SkipUnresolvedJsonWebKeys)
+                    if (!_skipUnresolvedJsonWebKeys)
                         signingKeys.Add(webKey);
                 }
             }
@@ -227,7 +229,7 @@ namespace Microsoft.IdentityModel.Tokens
             {
                 LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10802, jsonWebKey.X5c[0], ex), ex));
 
-                if (!SkipUnresolvedJsonWebKeys)
+                if (!_skipUnresolvedJsonWebKeys)
                     signingKeys.Add(jsonWebKey);
             }
         }
@@ -253,7 +255,7 @@ namespace Microsoft.IdentityModel.Tokens
             {
                 LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10801, jsonWebKey.E, jsonWebKey.N, ex), ex));
 
-                if (!SkipUnresolvedJsonWebKeys)
+                if (!_skipUnresolvedJsonWebKeys)
                     signingKeys.Add(jsonWebKey);
             }
         }
@@ -264,7 +266,7 @@ namespace Microsoft.IdentityModel.Tokens
             if (ECDsaAdapter == null)
             {
                 LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10690));
-                if (!SkipUnresolvedJsonWebKeys)
+                if (!_skipUnresolvedJsonWebKeys)
                     signingKeys.Add(jsonWebKey);
 
                 return;
@@ -284,7 +286,7 @@ namespace Microsoft.IdentityModel.Tokens
             {
                 LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10807, ex), ex));
 
-                if (!SkipUnresolvedJsonWebKeys)
+                if (!_skipUnresolvedJsonWebKeys)
                     signingKeys.Add(jsonWebKey);
             }
         }
--- a/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
+++ b/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
@@ -63,7 +63,6 @@ namespace Microsoft.IdentityModel.Tokens
         {
         }
 
-#pragma warning disable CS0618 // Type or member is obsolete
         /// <summary>
         /// Initializes an new instance of <see cref="JsonWebKeySet"/> from a json string.
         /// </summary>
@@ -73,7 +72,6 @@ namespace Microsoft.IdentityModel.Tokens
         public JsonWebKeySet(string json) : this(json, null)
         {
         }
-#pragma warning restore CS0618 // Type or member is obsolete
 
         /// <summary>
         /// Initializes an new instance of <see cref="JsonWebKeySet"/> from a json string.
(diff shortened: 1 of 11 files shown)


src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -122,68 +120,100 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
...
-                            {
-                                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10802, webKey.X5c[0]), ex));
-                            }
-                            catch (FormatException fex)
-                            {
-                                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10802, webKey.X5c[0]), fex));
-                            }
-                        }
-                    }
+                        signingKeys.AddRange(CreateX509SecurityKeys(webKey));
```
brentschmaltz: you could consider passing the list to CreateX509SecurityKeys and add them directly.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -122,68 +120,100 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
...
+            return signingKeys;
+        }
+
+        private IList<SecurityKey> CreateX509SecurityKeys(JsonWebKey jsonWebKey)
+        {
+            try
+            {
+                var keys = new List<SecurityKey>();
+
+                foreach (var certString in jsonWebKey.X5c)
```
brentschmaltz: It seems from the spec, that only the first key is used to perform any signing operations. The other keys are used to supply chain verification.

I am not sure we want to add them all.

Have a look at this: https://tools.ietf.org/html/rfc7517#section-4.7

GeoK: I agree. Opened a new issue to track this: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1146

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -122,68 +120,100 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
...
+                        signingKeys.Add(CreateRsaSecurityKey(webKey));
+                }
+                else if (webKey.Kty.Equals(JsonWebAlgorithmsKeyTypes.EllipticCurve, StringComparison.Ordinal))
+                {
+                    signingKeys.Add(CreateECDsaSecurityKey(webKey));
+                }
+                else
+                {
+                    //kty is not 'EC' or 'RSA'
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10809, webKey.Kty));
```
brentschmaltz: Why not just add the key as a jsonwebkey?

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -122,68 +120,100 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
...
+            try
+            {
+                var rsaParams = new RSAParameters
+                {
+                    Exponent = Base64UrlEncoder.DecodeBytes(jsonWebKey.E),
+                    Modulus = Base64UrlEncoder.DecodeBytes(jsonWebKey.N),
+                };
 
-            return keys;
+                return new RsaSecurityKey(rsaParams)
```
brentschmaltz: I opened #1148 to add additional ctor RsaSecurityKey(E, N)

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -122,68 +120,100 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
...
+                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10801, jsonWebKey.E, jsonWebKey.N), ex));
+            }
+        }
+
+        private SecurityKey CreateECDsaSecurityKey(JsonWebKey jsonWebKey)
+        {
+            try
+            {
+                var ecdsaAdapter = new ECDsaAdapter();
+                var ecdsa = ecdsaAdapter.CreateECDsa(jsonWebKey, false);
```
brentschmaltz: ECDsaAdapter.CreateECDsa is not static as we need to determine what is available on the machine. Could we create one adapter an reuse it?

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -122,68 +120,100 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
...
+                    };
+
+                    keys.Add(key);
                 }
+
+                return keys;
             }
+            catch (Exception ex)
+            {
+                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10802, jsonWebKey.X5c[0]), ex));
```
brentschmaltz: I'm wondering about throwing here. Suppose there is one bad key and a bunch of good ones.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: I was following the existing behavior where we skip anything that's not an RSA key.
Adding a key that we can't resolve, as a JsonWebKey, would add extensibility and allow users to provide their logic during signature validation.
I'll make the change to add the keys that we don't know how to resolve as JsonWebKey(s).

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: It will save some cycles - I'll make the change.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: Is this meant to improve performance when we have more than one EC key present in JsonWebKeySet? If so, I agree.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: This is also in-line with the existing behavior. I'll think about the alternative solutions.

GeoK: @brentschmaltz  As this feature is planned for 5.5.0 release, we should keep the existing behavior imo.
I'm thinking of adding a flag that would control skipping invalid signing keys (or throwing). 

Users could easily opt-out by flipping the flag and we won't throw an exception whenever we see an invalid signing key.

GeoK: addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1143/commits/5f40af68917573f4fae7a80c0bb542f6ad275b5e

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1143/commits/5f40af68917573f4fae7a80c0bb542f6ad275b5e

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1143/commits/5f40af68917573f4fae7a80c0bb542f6ad275b5e

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1143/commits/5f40af68917573f4fae7a80c0bb542f6ad275b5e

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -117,73 +116,151 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
...
-                        }
-                        catch (CryptographicException ex)
-                        {
-                            throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10801, webKey.E, webKey.N), ex));
-                        }
-                        catch (FormatException ex)
-                        {
-                            throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10801, webKey.E, webKey.N), ex));
-                        }
+                        AddRsaSecurityKey(signingKeys, webKey);
```
brentschmaltz: If the method was TryAddRsaSecurityKey(...), we would not need the flag.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -117,73 +116,151 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
...
+                        LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10810, webKey.KeyId ?? ""));
+                        signingKeys.Add(webKey);
+                    }
+                }
+                else if (webKey.Kty != null && webKey.Kty.Equals(JsonWebAlgorithmsKeyTypes.EllipticCurve, StringComparison.Ordinal))
+                {
+                    try
+                    {
+                        if (ecdsaAdapter == null)
+                            ecdsaAdapter = new ECDsaAdapter();
```
brentschmaltz: could this be static to the class rather than instance?

GeoK: ECDsa is not supported in our NETSTANDARD1.4 target, when running on platforms other than Windows. We'll thrown PlatformNotSupportedException when users try to initialize the adapter and it would be a breaking change for our netstandard1.4 mac/linux users.

That being said, we can initialize the adapter in a static constructor and swallow the exception (log it, perhaps).

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -63,7 +64,24 @@ public JsonWebKeySet()
...
+        /// <remarks>
+        /// As ECDsa Adapter is not supported on some platforms, PlatformNotSupported exception will be swallowed and logged.
+        /// </remarks>
+        static JsonWebKeySet()
+        {
+            try
+            {
+                ECDsaAdapter = new ECDsaAdapter();
+            }
+            catch (PlatformNotSupportedException ex)
```
brentschmaltz: If we catch all exceptions here then we are better protected. Otherwise, even if the keys being translated never have a ECD key, we could bring as server down.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -117,73 +139,150 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore, NullValueHandling = NullValueHandling.Ignore, PropertyName = JsonWebKeySetParameterNames.Keys, Required = Required.Default)]
         public IList<JsonWebKey> Keys { get; private set; } = new List<JsonWebKey>();
 
+        /// <summary>
+        /// Flag that controls whether unresolved JsonWebKeys will be included in the resulting collection of <see cref="GetSigningKeys"/> method.
+        /// </summary>
+        [DefaultValue(true)]
+        public static bool SkipUnresolvedJsonWebKeys { get; set; } = true;
```
brentschmaltz: I opened this to capture perhaps have a way to set this at runtime.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1165

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -117,73 +139,150 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
...
-            {
-                JsonWebKey webKey = Keys[i];
+            var signingKeys = new List<SecurityKey>();
 
-                if (!StringComparer.Ordinal.Equals(webKey.Kty, JsonWebAlgorithmsKeyTypes.RSA))
+            foreach (var webKey in Keys)
+            {
+                // skip if "use" (Public Key Use) parameter is not empty or "sig"
+                // https://tools.ietf.org/html/rfc7517#section-4.2
+                if (!(string.IsNullOrWhiteSpace(webKey.Use) || webKey.Use.Equals(JsonWebKeyUseNames.Sig, StringComparison.Ordinal)))
```
brentschmaltz: This could be breaking for some users. Should we apply the SkipUnresolvedJsonWebKeys flag?

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: The only exception that we could throw is _PlatformNotSupportedException_, but I understand your concern.
For more robustness, I'll add another catch-all clause that logs an exception.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: Do we still want to keep the flag for this PR?

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: We had this check before the change.
I guess that we could include these keys as well when the flag is set to false.

GeoK: addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1143/commits/3f39974831857fc0e2abe7afc477ba17caa1c359.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1143/commits/3f39974831857fc0e2abe7afc477ba17caa1c359

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1143/commits/3f39974831857fc0e2abe7afc477ba17caa1c359

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1143/commits/80717334871e2c108fa390f73b5aec1257b126ff

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -117,73 +139,154 @@ public JsonWebKeySet(string json, JsonSerializerSettings jsonSerializerSettings)
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore, NullValueHandling = NullValueHandling.Ignore, PropertyName = JsonWebKeySetParameterNames.Keys, Required = Required.Default)]
         public IList<JsonWebKey> Keys { get; private set; } = new List<JsonWebKey>();
 
+        /// <summary>
+        /// Flag that controls whether unresolved JsonWebKeys will be included in the resulting collection of <see cref="GetSigningKeys"/> method.
+        /// </summary>
+        [DefaultValue(true)]
+        public static bool SkipUnresolvedJsonWebKeys { get; set; } = true;
```
brentschmaltz: Can we add the pattern of a Default static then an instance variable that is set to the default. Similar to here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/ce23b8c4002d168d6ab2f9d971a9993e5cba99cc/src/Microsoft.IdentityModel.Tokens/TokenHandler.cs#L47

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/End2EndTests.cs
```diff
@@ -32,47 +32,113 @@
...
 
-            SecurityToken securityToken = null;
-            tokenHandler.ValidateToken(jwtToken.RawData, validationParameters, out securityToken);
+        public static TheoryData<OpenIdConnectTheoryData> OpenIdConnectTheoryData()
+        {
+            return new TheoryData<OpenIdConnectTheoryData>() {
+                new OpenIdConnectTheoryData
+                {
+                    OpenIdConnectMetadataFileName = OpenIdConfigData.OpenIdConnectMetadataFileEnd2End,
+                    SigningCredentials =
```
brentschmaltz: nit: SigningCredentials has different formatting.

test/Microsoft.IdentityModel.Tokens.Tests/JsonWebKeySetTests.cs
```diff
@@ -121,6 +128,367 @@ public void GetSets()
...
+                    TestId = "InvalidX5c",
+                });
+
+                jsonWebKeySet = new JsonWebKeySet(DataSets.JsonWebKeySetOneInvalidEcOneValidEcString);
+                theoryData.Add(new JsonWebKeySetTheoryData
+                {
+                    JsonWebKeySet = jsonWebKeySet,
+                    SkipUnresolvedJsonWebKeys = true,
+                    ExpectedSigningKeys = new List<SecurityKey>(),
+                    ExpectedException = ExpectedException.NoExceptionExpected,
```
brentschmaltz: nit: TheoryDataBase sets this to NoExceptionExpected by default.

GeoK: I was trying to be explicit in the previous iterations, but this is completely redundant as we are not throwing anymore. I'll remove it.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
GeoK: I'll make this change to maintain consistency.

GeoK: addressed with f27d3344907beb84e704cac8674543090a41b68f

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/End2EndTests.cs
GeoK: addressed with f27d3344907beb84e704cac8674543090a41b68f

test/Microsoft.IdentityModel.Tokens.Tests/JsonWebKeySetTests.cs
GeoK: addressed with d239fd43fc891e23202bd4532de76a4db05c9239

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -41,14 +42,16 @@ namespace Microsoft.IdentityModel.Tokens
     [JsonObject]
     public class JsonWebKeySet
     {
+        private readonly bool _skipUnresolvedJsonWebKeys = SkipUnresolvedJsonWebKeys;
```
brentschmaltz: I didn't quite explain myself here. What I was suggesting is having the property:
public static bool DefaultSkipUnresolvedJsonWebKeys = true;

Then the instance is

public bool SkipUnresolvedJsonWebKeys {get;set;} = DefaultSkipUnresolvedJsonWebKeys;

GeoK: Got it. Addressed with 35fdd53974abd1f8330db12ab30b4898b60d94e7

Issue #1144, opened 2019-04-30
Make the public JwtHeader default constructor internal


mafurman:
The 'alg' (algorithm) header parameter must be present (according to the [JWS RFC](https://tools.ietf.org/html/rfc7515#section-4.1.1)), so we should not allow an empty JwtHeader object to be created.

However, this constructor should not be completely removed as it needs to be available for use by JsonConvert.DeserializeObject.

This is basically the same as issue #1092, which we are planning on addressing for the 5.5.0 release. However, as removing a constructor is a breaking change, the 5.5.0 fix will need to address this problem differently.

brentschmaltz:
@henrik-me this would be a breaking change, no real good reason to do it.

PR #1145, opened 2019-04-30
JwtHeader default constructor now sets the 'Alg' property to SecurityAlgorithms.None

### LINKED ISSUE #1092: JwtHeader default constructor doesn't set the Alg = None
The default constructor for `JwtHeader`, which is called from `JwtSecurityTokenHandler.CreateJwtSecurityTokenPrivate` here:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/45dfc2866d1d9dca1c25a25b152d48e46348a52c/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L501

Does not set the `JwtHeader.Alg` value to "none", as is done when calling the other constructors on `JwtHeader`, such as this one:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/45dfc2866d1d9dca1c25a25b152d48e46348a52c/src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs#L85

I assume this is not intentional to leave alg blank? The missing `alg` in the header is causing issues with our OpenId connect server, returning a JWT that has an empty header.

mafurman:
Fixes #1092.

The 'Alg' header claim is required according to the [JWS RFC](https://tools.ietf.org/html/rfc7515#section-4.1.1), so we should not be allowing the creation of an empty JWT header.

Suggested fix for the 6x release: #1144.

### DIFF
--- a/src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
@@ -45,6 +45,11 @@ namespace System.IdentityModel.Tokens.Jwt
         public JwtHeader()
             : base(StringComparer.Ordinal)
         {
+            // Since this constructor is used by JsonConvert.DeserializeObject, it is possible that the serialization order of the JwtHeader
+            // will be different from the original JSON that was used for deserialization. This can occur because setting the 'Alg' property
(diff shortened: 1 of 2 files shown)


brentschmaltz:
:shipit:

PR #1151, opened 2019-05-01
Added control for maximum number of transforms in a reference

brentschmaltz:
Transforms can be cpu intensive. By limiting them to 10 or less we reduce the possibility of a DOC attack.

### DIFF
--- a/src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
+++ b/src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
@@ -26,6 +26,7 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.ComponentModel;
 using System.IO;
 using System.Text;
 using System.Xml;
@@ -36,11 +37,12 @@ using static Microsoft.IdentityModel.Logging.LogHelper;
 namespace Microsoft.IdentityModel.Xml
 {
     /// <summary>
-    /// Reads and writes XML associated with XML DSig https://www.w3.org/TR/2001/PR-xmldsig-core-20010820
+    /// Reads and writes XML conforming to https://www.w3.org/TR/2001/PR-xmldsig-core-20010820
     /// </summary>
     public class DSigSerializer
     {
         private static DSigSerializer _default;
+        private int _maximumReferenceTransforms = 5;
         private TransformFactory _transformFactory = TransformFactory.Default;
 
         /// <summary>
@@ -48,18 +50,12 @@ namespace Microsoft.IdentityModel.Xml
         /// </summary>
         public static DSigSerializer Default
         {
-            get
-            {
-                return _default;
-            }
-            set
-            {
-                _default = value ?? throw LogArgumentNullException(nameof(value));
-            }
+            get => _default;
+            set => _default = value ?? throw LogArgumentNullException(nameof(value));
         }
 
         /// <summary>
-        /// Static constructor that initializes the default <see cref="CryptoProviderFactory"/>.
+        /// Static constructor that initializes the default <see cref="DSigSerializer"/>.
         /// </summary>
         static DSigSerializer()
         {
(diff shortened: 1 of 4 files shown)


AndersAbel:
I think you can be even more strict. In real world examples, I've never seen more than two: Enveloped transform and the canonicalization.

There's also an automatic white-list of allowed transforms through those that are implemented. Which is good. There are transforms (I'm looking at you XPath) that are bad to allow.

From the SAML2 spec:
> ## 5.4.4 Transforms
>Signatures in SAML messages SHOULD NOT contain transforms other than the enveloped signature transform (with the identifier http://www.w3.org/2000/09/xmldsig#enveloped-signature) or the exclusive canonicalization transforms (with the identifier http://www.w3.org/2001/10/xml-exc-c14n# or http://www.w3.org/2001/10/xml-exc-c14n#WithComments). Verifiers of signatures MAY reject signatures that contain other transform algorithms as invalid. If they do not, verifiers MUST ensure that no content of the SAML message is excluded from the signature. This can be accomplished by establishing out-of-band agreement as to what transforms are acceptable, or by applying the transforms manually to the content and reverifying the result as consisting of the same SAML message.

brentschmaltz:
@AndersAbel do you think 10 is too many? TransformFactory only supports Enveloped, c14n, c14nWithComments. 
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/72f03a9c62215688b759ea180263db8b498280ea/src/Microsoft.IdentityModel.Xml/TransformFactory.cs#L36

test/Microsoft.IdentityModel.Xml.Tests/DSigSerializerTests.cs
```diff
@@ -45,17 +45,23 @@ public class DSigSerializerTests
         [Fact]
         public void GetSets()
         {
-            var dsigSerializer = new DSigSerializer();
+
+            var serializer = new DSigSerializer();
```
GeoK: nit: looks like the serializer is not needed here. Extra new line.

brentschmaltz: Thanks, some weird spacing in that file also.

AndersAbel:
> @AndersAbel do you think 10 is too many? TransformFactory only supports Enveloped, c14n, c14nWithComments.

Would there be any reasonable use case where any of those transforms is applied more than once? And would there be any reasonable use case to combine c14n and c14nWithComments? To keep some room for unexpected stuff (there are some really strange implementations out there) some margin is probably good. But I'd say 3 as the limit, not 10. That reduces the DoS attack surface.

brentschmaltz:
@AndersAbel These are good points. I would go for '5' max.
I opened https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1154.

I marked it for 6.x.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -73,6 +69,18 @@ public DSigSerializer()
         {
         }
 
+        /// <summary>
+        /// Gets or sets the maximum number of Transforms that are allowed on a Reference
```
mafurman: nit: missing a period at the end of the comment.

brentschmaltz: thanks

AndersAbel:
> @AndersAbel These are good points. Can we agree on '5' max as a default?
> I opened #1154 for 6.x.

Yes, 5 works, should reduce DoS risk.

brentschmaltz:
@AndersAbel thanks. On a side note, we are thinking of adding some sort of 'AppContext' similar to what is available here: https://docs.microsoft.com/en-us/dotnet/api/system.appcontext?view=netcore-2.2

So users can control runtime settings. Would that be of value to you?

AndersAbel:
> @AndersAbel thanks. On a side note, we are thinking of adding some sort of 'AppContext' similar to what is available here: https://docs.microsoft.com/en-us/dotnet/api/system.appcontext?view=netcore-2.2
> 
> So users can control runtime settings. Would that be of value to you?

No, I don't think so. As a library author, I wouldn't touch anything that alters global behaviour - that's up to the application developer to handle. And I would be a bit concerned if there were ways for the application developer to alter behaviour without my library being able to detect so.

Issue #1147, opened 2019-04-30
Add the 'Deflate' to the processing pipeline for JwtSecurityTokenHandler.


brentschmaltz:
If an IdentityProvider will produce a JWE using deflate, then users of Asp.net will fail.
We added deflate to JsonWebTokenHandler, we need it in JwtSecurityTokenHandler also.

PR #1156, opened 2019-05-01
Added support for JWE decompression in JwtSecurityTokenHandler

mafurman:
Addresses #1147.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -86,6 +86,34 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         }
 
         /// <summary>
+        /// Decompress JWT token bytes.
+        /// </summary>
+        /// <param name="tokenBytes"></param>
+        /// <param name="algorithm"></param>
+        /// <exception cref="ArgumentNullException">if <paramref name="tokenBytes"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="algorithm"/> is null.</exception>
+        /// <exception cref="NotSupportedException">if the decompression <paramref name="algorithm"/> is not supported.</exception>
+        /// <exception cref="SecurityTokenDecompressionFailedException">if decompression using <paramref name="algorithm"/> fails.</exception>
+        /// <returns>Decompressed JWT token</returns>
+        internal static string DecompressToken(byte[] tokenBytes, string algorithm)
+        {
+            if (tokenBytes == null)
+                throw LogHelper.LogArgumentNullException(nameof(tokenBytes));
+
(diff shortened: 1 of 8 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -85,6 +85,33 @@ public static string CreateEncodedSignature(string input, SigningCredentials sig
...
+        /// <summary>
+        /// Decompress JWT token bytes.
+        /// </summary>
+        /// <param name="tokenBytes"></param>
+        /// <param name="algorithm"></param>
+        /// <exception cref="ArgumentNullException">if 'tokenBytes' is null.</exception>
+        /// <exception cref="ArgumentNullException">if 'algorithm' is null.</exception>
+        /// <exception cref="NotSupportedException">if the decompression algorithm is not supported.</exception>
+        /// <returns>Decompressed JWT token</returns>
+        public static string DecompressToken(byte[] tokenBytes, string algorithm)
```
brentschmaltz: If we make this public, then users have no easy way to add a new CompressionProviderFactory.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -85,6 +85,33 @@ public static string CreateEncodedSignature(string input, SigningCredentials sig
...
         }
 
+        /// <summary>
+        /// Decompress JWT token bytes.
+        /// </summary>
+        /// <param name="tokenBytes"></param>
+        /// <param name="algorithm"></param>
+        /// <exception cref="ArgumentNullException">if 'tokenBytes' is null.</exception>
+        /// <exception cref="ArgumentNullException">if 'algorithm' is null.</exception>
+        /// <exception cref="NotSupportedException">if the decompression algorithm is not supported.</exception>
```
brentschmaltz: Missing the InvalidOperationException.
I think there is a better choice, the application will not really know what to do here.
Maybe: SecurityTokenDecompressionFailedException (this would be new).

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Good idea. I can see how this doesn't give users any choice but to use the default compression provider. I've made this internal.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Good catch! :) We actually already have a SecurityTokenDecompressionFailedException. I've made the change to throw this exception instead of InvalidOperationException.

brentschmaltz:
:shipit:

Issue #1112, opened 2019-02-07
Improve logging when a HttpClient.SendAsync fails.


brentschmaltz:
Currently we call response.EnsureSuccessCode();
We should be logging the entire Response using: HttpResponseMessage.Content.ReadAsStringAsync().

Ensure that our current logic captures the entire error from: HttpContext.
We need a way to surface the status code back to the application.

PR #1157, opened 2019-05-02
ConfigurationManager no longer throws if _configRetriever fails but _currentConfiguration is not null

### LINKED ISSUE #1111: ConfigurationManager should only throw if _currentConfiguration is null if the _configRetriever fails
The current code in ConfigurationManager throws too early.
We should not throw here:  https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs#L202

Instead only throw if after the catch _currentConfiguration is null, that is only throw if there is no configuration to return.

mafurman:
Fixes #1111 and #1112.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -199,7 +199,10 @@ namespace Microsoft.IdentityModel.Protocols
                     catch (Exception ex)
                     {
                         _syncAfter = DateTimeUtil.Add(now.UtcDateTime, _automaticRefreshInterval < _refreshInterval ? _automaticRefreshInterval : _refreshInterval);
-                        throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20803, (_metadataAddress ?? "null")), ex));
+                        if (_currentConfiguration == null) // Throw an exception if there's no configuration to return.
+                            throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20803, (_metadataAddress ?? "null")), ex));
+                        else
+                            LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20806, (_metadataAddress ?? "null")), ex));
                     }
                 }
 
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
@@ -82,19 +82,25 @@ namespace Microsoft.IdentityModel.Protocols
             if (!Utility.IsHttps(address) && RequireHttps)
                 throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX20108, address), nameof(address)));
 
+            Exception unsuccessfulHttpResponseException;
             try
             {
                 LogHelper.LogVerbose(LogMessages.IDX20805, address);
                 var httpClient = _httpClient ?? _defaultHttpClient;
-                HttpResponseMessage response = await httpClient.GetAsync(address, cancel).ConfigureAwait(false);
+                var response = await httpClient.GetAsync(address, cancel).ConfigureAwait(false);
+                var responseContent = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
 
-                response.EnsureSuccessStatusCode();
-                return await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+                if (response.IsSuccessStatusCode)
+                    return responseContent;
+
+                 unsuccessfulHttpResponseException = new IOException(LogHelper.FormatInvariant(LogMessages.IDX20807, address, response, responseContent));
             }
             catch (Exception ex)
             {
                 throw LogHelper.LogExceptionMessage(new IOException(LogHelper.FormatInvariant(LogMessages.IDX20804, address), ex));
             }
+
+            throw LogHelper.LogExceptionMessage(unsuccessfulHttpResponseException);
         }
     }
 }
--- a/src/Microsoft.IdentityModel.Protocols/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Protocols/LogMessages.cs
@@ -48,6 +48,8 @@ namespace Microsoft.IdentityModel.Protocols
         internal const string IDX20803 = "IDX20803: Unable to obtain configuration from: '{0}'.";
         internal const string IDX20804 = "IDX20804: Unable to retrieve document from: '{0}'.";
         internal const string IDX20805 = "IDX20805: Obtaining information from metadata endpoint: '{0}'.";
+        internal const string IDX20806 = "IDX20806: Unable to obtain an updated configuration from: '{0}'. Returning the current configuration.";
+        internal const string IDX20807 = "IDX20807: Unable to retrieve document from: '{0}'. HttpResponseMessage: '{1}', HttpResponseMessage.Content: '{2}'.";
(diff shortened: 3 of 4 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -199,7 +199,10 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                     catch (Exception ex)
                     {
                         _syncAfter = DateTimeUtil.Add(now.UtcDateTime, _automaticRefreshInterval < _refreshInterval ? _automaticRefreshInterval : _refreshInterval);
-                        throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20803, (_metadataAddress ?? "null")), ex));
+                        if (_currentConfiguration == null) // Throw an exception if there's no configuration to return.
+                            throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20803, (_metadataAddress ?? "null")), ex));
+                        else
+                            LogHelper.LogWarning(LogHelper.FormatInvariant(LogMessages.IDX20806, (_metadataAddress ?? "null"), ex));
```
brentschmaltz: Can we bump this up to error?
How about LogHelper.LogException with the same message, but don't throw?

henrik-me: Suggestion: Can this only happen if there is a network connection issue or how can it happen that we don't have the configuration? If we have additional information from e.g. the http response we could possibly add that? (e.g. time-out vs. 500, vs. not found, etc.)? Perhaps also depending on the response we can decide if it's a warning or an error?

---
In reply to: [280398515](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1157#discussion_r280398515) [](ancestors = 280398515)

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -199,7 +199,10 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                     catch (Exception ex)
                     {
                         _syncAfter = DateTimeUtil.Add(now.UtcDateTime, _automaticRefreshInterval < _refreshInterval ? _automaticRefreshInterval : _refreshInterval);
-                        throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20803, (_metadataAddress ?? "null")), ex));
+                        if (_currentConfiguration == null) // Throw an exception if there's no configuration to return.
+                            throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20803, (_metadataAddress ?? "null")), ex));
```
henrik-me: >throw LogHelper. [](start = 28, length = 16)

Consider not throwing and logging here, this is happening further down with a similar message.

henrik-me:
>                     throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20803, (_metadataAddress ?? "null"))));

perhaps this should be the only place where _config == null should throw?

---
Refers to: src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs:214 in b828af1. [](commit_id = b828af1b65f066f29d8c2694a715a2513290d489, deletion_comment = False)

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
brentschmaltz: I would like to continue to throw here as we could provide details that may not be available later.
I am suggesting we throw / inform users to understand metadata retrieval failed.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
brentschmaltz: @henrik-me It depends on the retriever that was used. I have had conversations with users where they need details. For example, if an IdentityProvider accidentally pushes bad metadata {bad formatting for example}, that would cause an error.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
mafurman: I was considering only throwing further down, but this will cause us to lose information about the original exception that was thrown.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
mafurman: @brentschmaltz It looks like the HttpDocumentRetriever that's used by default in the OpenIdConnectConfigurationRetriever ends up throwing if the response does not have a success status code. So that exception could have some extra information as to why retrieving metadata failed.

I'm not sure how we would go about providing additional information about the HTTP response from the ConfigurationManager, as this information would somehow need to be bubbled up from the IDocumentRetriever that's used by the configuration retriever...

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
henrik-me: That information can be stored for later usage

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
henrik-me: If that information at that point in time is logged then perhaps the logs already have enough information for the developer, even though it may not be used to potentially improve this error message. I think that is ok, however I suggest reviewing what we log in the HttpDocumentRetriever.

mafurman:
I updated this PR so that a more specific exception type is thrown, and so that we log error instead of warning when we can't obtain an updated configuration (but the current configuration is not null).

I also added in a test case to check that the correct exception is thrown when we can't connect to the provided metadata endpoint. I think that the sequence of inner exceptions we provide does give enough information to understand why obtaining the configuration failed, but I can also work on providing more  information from the HTTP response in the exception thrown if we agree that's necessary.

Though I do believe that throwing a more specific exception (rather than InvalidOperationException) may be a better choice, I'm not sure if this is a change we want to make as it is a breaking change.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
GeoK: This discussion is closely related to https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1112.
Maria is correct, `EnsureSuccessStatusCode `method throws `HttpRequestException` that doesn't contain an http status code.

I believe that there is value in bubbling-up an exception that has an http status code property on it, which allows developers to react to specific events (e.g. 429 for throttling).

Tratcher: https://github.com/dotnet/corefx/issues/26477. Please don't invent a new exception type, we can address this in the framework instead.

brentschmaltz: @Tratcher As noted above, HttpDocumentRetriever calls EnsureSuccessStatusCode(), which throws. I have had customers indicate there is additional info as stated in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1112. How do we serve that up?

mafurman: I removed the new exception type and modified this PR to also address #1112.

brentschmaltz:
:shipit:

Issue #972, opened 2018-07-02
ValidateIssuerSigningKey shouldn't be called if the signing key is null and TokenValidationParameters.RequireSignedTokens is false


holytshirt:
When we upgrade from  netcoreapp2.0 to netcoreapp2.1 our JWT authentication for unsigned tokens fails with
```
18:23:57 INF] Failed to validate the token.
System.ArgumentNullException: IDX10000: The parameter 'securityKey' cannot be a 'null' or an empty object.
Parameter name: securityKey
   at Microsoft.IdentityModel.Tokens.Validators.ValidateIssuerSecurityKey(SecurityKey securityKey, SecurityToken securityToken, TokenValidationParameters validationParameters)
   at System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.ValidateIssuerSecurityKey(SecurityKey key, JwtSecurityToken securityToken, TokenValidationParameters validationParameters)
   at System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.ValidateTokenPayload(JwtSecurityToken jwtToken, TokenValidationParameters validationParameters)
   at System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.ValidateToken(String token, TokenValidationParameters validationParameters, SecurityToken& validatedToken)
   at Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerHandler.HandleAuthenticateAsync()
[18:23:57 INF] Bearer was not authenticated. Failure message: IDX10000: The parameter 'securityKey' cannot be a 'null' or an empty object.
Parameter name: securityKey
```
Our config looks like this which works on  netcoreapp2.0 and fails on netcoreapp2.1
```csharp
services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme).AddJwtBearer(options =>
            {
                options.Audience = _config.JWTAudience;
                
                X509Certificate2 x509Certificate2 = new X509Certificate2(Convert.FromBase64String(_config.Key));

                options.TokenValidationParameters = new TokenValidationParameters
                {
                    ValidateIssuerSigningKey = true,
                    ValidateIssuer = true,
                    ValidateAudience = true,
                    ValidIssuer = _config.JWTIssuer,
                    IssuerSigningKey = new X509SecurityKey(x509Certificate2),
                    RequireSignedTokens = !_config.AllowUnsignedTokens
                };
            });
```

[...]

PR #1158, opened 2019-05-03
ValidateIssuerSigningKey shouldn't be always called

GeoK:
ValidateIssuerSigningKey shouldn't be called if the signing key is null
and TokenValidationParameters.RequireSignedTokens is false.

* Supported in SamlSecurityTokenHandler, Saml2SecurityTokenHandler,
JsonWebTokenHandler, and JwtSecurityTokenHandler.

* Added 4 test combinations for each handler:
  - RequiredSignedTokens is false and the key is null
  - RequiredSignedTokens is false and the key is not null
  - RequiredSignedTokens is true and the key is null (throws)
  - RequiredSignedTokens is true and the key is not null

* Removed a few redundant assignments.

Resolves: #972


test/Microsoft.IdentityModel.TestUtils/ReferenceTokens.cs
```diff
@@ -218,6 +218,10 @@ public static string Saml2Token_Formated
...
+    
+
```
mafurman: nit: extra new line here

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2SecurityTokenHandlerTests.cs
```diff
@@ -690,10 +690,9 @@ public void ValidateToken(Saml2TheoryData theoryData)
         {
             TestUtilities.WriteHeader($"{this}.ValidateToken", theoryData);
             var context = new CompareContext($"{this}.ValidateToken, {theoryData}");
-            ClaimsPrincipal retVal = null;
             try
             {
-                retVal = theoryData.Handler.ValidateToken(theoryData.Token, theoryData.ValidationParameters, out SecurityToken validatedToken);
+                var retVal = theoryData.Handler.ValidateToken(theoryData.Token, theoryData.ValidationParameters, out SecurityToken validatedToken);
```
mafurman: Can we just call handler.ValidateToken() without storing the return value in a variable? (also relevant in the SamlSecurityTokenHandlerTests)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -834,14 +834,16 @@ private TokenValidationResult ValidateTokenPayload(JsonWebToken jsonWebToken, To
 
             Validators.ValidateLifetime(notBefore, expires, jsonWebToken, validationParameters);
             Validators.ValidateAudience(jsonWebToken.Audiences, jsonWebToken, validationParameters);
-            var issuer = Validators.ValidateIssuer(jsonWebToken.Issuer, jsonWebToken, validationParameters);
```
brentschmaltz: This indicates we have a difference between JsonWebTokenHandler and JwtSecurityTokenHandler.
The value returned from ValidateIssuer is set as the Claim.Issuer in the JwtSecurityTokenHandler.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1164

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -1128,7 +1128,10 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
             ValidateConditions(samlToken, validationParameters);
             var issuer = ValidateIssuer(samlToken.Issuer, samlToken, validationParameters);
             ValidateTokenReplay(samlToken.Assertion.Conditions.NotOnOrAfter, token, validationParameters);
-            ValidateIssuerSecurityKey(samlToken.SigningKey, samlToken, validationParameters);
+            if (validationParameters.RequireSignedTokens || samlToken.SigningKey != null)
```
brentschmaltz: Can we put this logic into Validators.ValidateIssuerSecurityKey?

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2SecurityTokenHandlerTests.cs
GeoK: Thanks, you are right. I removed a few redundant initialization, but sometimes when didn't need to initialize at all :)

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
GeoK: I was following the original issue description: `Our handlers should not call ValidateIssuerSigningKey...`
ValidateIssuerSecurityKey is a public static method so I assumed that it's safer only to change our handlers.

I can centralize this change by putting the logic in ValidateIssuerSecurityKey, if we are fine with altering the ValidateIssuerSecurityKey method.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
GeoK: Good catch!

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
brentschmaltz: That title wasn't accurate.
The issue is we are throwing when we shouldn't.
The pattern we locked on is: SecurityTokenHandler-> Virtual -> Validators.
In this way we will have the same results across token types.
It is this line here that is causing us grief: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/eac57a0465f9d76d1c6a840c207e20084cb7de07/src/Microsoft.IdentityModel.Tokens/Validators.cs#L203

The fix looks good. Should be in validators.
Saml2SecurityTokenHandler is fine, but SamlSecurityTokenHandler should just call into the Validator.

test/Microsoft.IdentityModel.TestUtils/ReferenceTokens.cs
GeoK: addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1158/commits/9e94ffea506bcab20e4cede543525783e149715c.

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2SecurityTokenHandlerTests.cs
GeoK: addressed with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1158/commits/9e94ffea506bcab20e4cede543525783e149715c.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
GeoK: addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1158/commits/9e94ffea506bcab20e4cede543525783e149715c.

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -200,6 +200,12 @@ public static void ValidateIssuerSecurityKey(SecurityKey securityKey, SecurityTo
                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSigningKeyException(LogHelper.FormatInvariant(LogMessages.IDX10232, securityKey)){ SigningKey = securityKey });
             }
 
+            if (validationParameters.RequireSignedTokens == false && securityKey == null)
```
brentschmaltz: We could write: if (!validationParameters.RequireSignedTokens && securityKey == null)

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -200,6 +200,12 @@ public static void ValidateIssuerSecurityKey(SecurityKey securityKey, SecurityTo
...
+            {
+                LogHelper.LogInformation(LogMessages.IDX10252);
+                return;
+            }
+
             if (securityKey == null)
```
brentschmaltz: We need a exception here that explains the securityKey is null but validationParameters.RequireSignedTokens == true. This indicates an error, but I have to admit, it would be hard to get here.

src/Microsoft.IdentityModel.Tokens/Validators.cs
GeoK: addressed with 822afd1165f5f6323218bda6534bdd3f97cc4d46

src/Microsoft.IdentityModel.Tokens/Validators.cs
GeoK: addressed with 822afd1165f5f6323218bda6534bdd3f97cc4d46

Issue #1117, opened 2019-02-26
Validating a JWT using the PS256 algorithm?


underscoreHao:
Hello,

The last two days I've been trying to verify a JWT that has been encoded using the PS256 algorithm. I have an implementation for RS256 that's working correctly, but despite my best efforts I couldn't validate a PS256 JWT. I looked and debugged through the code of **Microsoft.IdentityModel.Tokens** and it seems that although PS256 is in the [SecurityAlgorithms.cs](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs), an actual implementation of the validation is missing. Is this on the roadmap? Am I doing something wrong? Perhaps I'm not understanding the difference between RSA256 and PS256 well enough.

Any kind of information on this will be appreciated.

**EDIT:** I'm supplying an example token

    // Encoded
    eyJhbGciOiJQUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImtpZDEyMzQifQ.eyJpc3MiOiJmb28uYmFyLnRlc3Rpc3N1ZXIiLCJleHAiOjE1NTEyMDEwNjgsImF0X2hhc2giOiJqaFl3c1pyTnZ0dFNYQnR6QVMtWlNnIn0.yJePyxdJWyydG4HM97oQag6ulGKa5Afw-LHYYEXz7lVy8v0IJD0mSO9WtowlWJIeD2Vvthuj71XUfHsgz0LD9rK0VBucJbd_OiIXpbwPUqBcdj82DNLFXDJfCJnUC-Rv8QP7OUVBvLjvBQ6WYMrx1Qnq8xP6qeL_ohKwRmo6EDhZRkYBz9gFhfha1ZlKcnyR73nXdShwy7OmmyiRvVWPBf_GgSsfz8FNNoKySW1MA4tRa7cl3zPlyCnWyLaZ3kcQsmTqarHG--YXSDF5ozZ_Sx6TkunCxrOYzOFNcPyeIWqI84cemM6TgMBw9jhzMCk7Y4Fzxe5KEYJH4GlGA4s4zg
    
    // Header
    {
        "alg": "PS256",
        "typ": "JWT",
        "kid": "kid1234"
    }
    // Payload
    {
        "iss": "foo.bar.testissuer",
        "exp": 1551201068,
[...]

PR #1166, opened 2019-05-10
 Add RSA-PSS support

GeoK:
* RSA-PSS support matrix is available [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/Supported-Algorithms):
  - Rsa-Pss is not available on .NET 4.5 and .NET 4.5.1 targets
  - Rsa-Pss is available on .NET 4.6.1 target but only byt Windows CNG
  - Rsa-Pss is avaliable on .NET Core on Windows
  - Rsa-Pss is available on .NET Core on Unix on SDK versions >= 2.1
* Added new log messages
* Added tests

Resolves: #1117

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
@@ -290,10 +290,30 @@ namespace Microsoft.IdentityModel.Tokens
 
 #if NET461 || NETSTANDARD1_4 || NETSTANDARD2_0
         private HashAlgorithmName HashAlgorithmName { get; set; }
+
+        private RSASignaturePadding RSASignaturePadding { get; set; }
 #endif
 
         private void Initialize(RSA rsa, string algorithm)
         {
+
+#if NET461 || NETSTANDARD1_4 || NETSTANDARD2_0
+            if (algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha256) ||
+                algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha256Signature) ||
+                algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha384) ||
+                algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha384Signature) ||
+                algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha512) ||
+                algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha512Signature))
+            {
+                RSASignaturePadding = RSASignaturePadding.Pss;
+            }
+            else
+            {
+                // default RSASignaturePadding for other supported RSA algorithms is Pkcs1
+                RSASignaturePadding = RSASignaturePadding.Pkcs1;
+            }
+#endif
+
             // This case is the result of a calling
             // X509Certificate2.GetPrivateKey OR X509Certificate2.GetPublicKey.Key
             // These calls return an AsymmetricAlgorithm which doesn't have API's to do much and need to be cast.
@@ -366,7 +386,7 @@ namespace Microsoft.IdentityModel.Tokens
 #if NET461 || NETSTANDARD1_4 || NETSTANDARD2_0
         private byte[] SignWithRsa(byte[] bytes)
         {
-            return RSA.SignHash(HashAlgorithm.ComputeHash(bytes), HashAlgorithmName, RSASignaturePadding.Pkcs1);
+            return RSA.SignHash(HashAlgorithm.ComputeHash(bytes), HashAlgorithmName, RSASignaturePadding);
         }
 #endif
 
@@ -394,7 +414,7 @@ namespace Microsoft.IdentityModel.Tokens
 #if NET461 || NETSTANDARD1_4 || NETSTANDARD2_0
         private bool VerifyWithRsa(byte[] bytes, byte[] signature)
         {
-            return RSA.VerifyHash(HashAlgorithm.ComputeHash(bytes), signature, HashAlgorithmName, RSASignaturePadding.Pkcs1);
+            return RSA.VerifyHash(HashAlgorithm.ComputeHash(bytes), signature, HashAlgorithmName, RSASignaturePadding);
         }
 #endif
 
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -197,7 +197,6 @@ namespace Microsoft.IdentityModel.Tokens
                 case SecurityAlgorithms.RsaSha256Signature:
                 case SecurityAlgorithms.RsaSsaPssSha256:
                 case SecurityAlgorithms.RsaSsaPssSha256Signature:
-
                     return HashAlgorithmName.SHA256;
 
                 case SecurityAlgorithms.EcdsaSha384:
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -57,12 +57,12 @@ namespace Microsoft.IdentityModel.Tokens
             { SecurityAlgorithms.RsaSha256Signature, 2048 },
             { SecurityAlgorithms.RsaSha384Signature, 2048 },
             { SecurityAlgorithms.RsaSha512Signature, 2048 },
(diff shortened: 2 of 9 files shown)


src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -183,18 +195,25 @@ protected virtual HashAlgorithmName GetHashAlgorithmName(string algorithm)
                 case SecurityAlgorithms.EcdsaSha256Signature:
                 case SecurityAlgorithms.RsaSha256:
                 case SecurityAlgorithms.RsaSha256Signature:
+                case SecurityAlgorithms.RsaSsaPssSha256:
+                case SecurityAlgorithms.RsaSsaPssSha256Signature:
+
```
brentschmaltz: nit: extra line

src/Microsoft.IdentityModel.Tokens/SupportedAlgorithms.cs
```diff
@@ -170,11 +171,51 @@ internal static bool IsSupportedRsaAlgorithm(string algorithm)
...
+#if NET45 || NET451
+            // RSA-PSS is not available on .NET 4.5 and .NET 4.5.1
+            LogHelper.LogExceptionMessage(new PlatformNotSupportedException(LogMessages.IDX10692));
+            return false;
+#elif NET461 || NETSTANDARD2_0
+            // RSACryptoServiceProvider doesn't support RSA-PSS
+            if (key is RsaSecurityKey rsa && rsa.Rsa is RSACryptoServiceProvider)
+            {
+                LogHelper.LogExceptionMessage(new PlatformNotSupportedException(LogMessages.IDX10693));
+                return false;
```
brentschmaltz: Is querying for support an Exception? Perhaps information or warning?

src/Microsoft.IdentityModel.Tokens/SupportedAlgorithms.cs
```diff
@@ -170,11 +171,51 @@ internal static bool IsSupportedRsaAlgorithm(string algorithm)
...
+            }
+            else if (key is X509SecurityKey x509 && x509.PrivateKey is RSACryptoServiceProvider)
```
brentschmaltz: Public and Private keys will return the same RSA type. We only need to check the Public key.

src/Microsoft.IdentityModel.Tokens/SupportedAlgorithms.cs
```diff
@@ -170,11 +171,51 @@ internal static bool IsSupportedRsaAlgorithm(string algorithm)
...
+                case SecurityAlgorithms.RsaSsaPssSha512Signature:
+                    return IsSupportedRsaPss(key);
             }
 
             return false;
         }
 
+        private static bool IsSupportedRsaPss(SecurityKey key)
+        {
+#if NET45 || NET451
```
brentschmaltz: If the application is targeting NET451, but NET48 runtime is installed, RsaPSS will actually be supported. We need to expand our work 'lightup' work here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/ce23b8c4002d168d6ab2f9d971a9993e5cba99cc/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs#L515

test/Microsoft.IdentityModel.Tokens.Tests/AsymmetricSignatureTests.cs
```diff
@@ -115,6 +137,15 @@ public static TheoryData<SignatureProviderTheoryData> SignVerifyTheoryData
                     theoryData);
 
 #if NET461 || NETCOREAPP2_0
+
```
mafurman: nit: extra new line

src/Microsoft.IdentityModel.Tokens/SupportedAlgorithms.cs
GeoK: It should probably be an information. I wasn't looking at it from the querying perspective which was a mistake.

src/Microsoft.IdentityModel.Tokens/SupportedAlgorithms.cs
GeoK: I assumed that users running .NET 4.8 would use our .NET 4.6.1 target, but I see use cases where this could be important e.g. Katana supports only .NET 4.5 target.

Tratcher: Katana's target doesn't matter, nuget resolves based on the application target.

brentschmaltz: @GeoK we should open a new issue for this.

GeoK: Oh, that's right. Thanks @Tratcher.
@brentschmaltz In what scenarios would users targeting e.g. .NET 4.6.1 have to bind to our .NET 4.5 target?

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -290,10 +290,30 @@ internal byte[] Encrypt(byte[] data)
 
 #if NET461 || NETSTANDARD1_4 || NETSTANDARD2_0
```
henrik-me: > NET461 || NETSTANDARD1_4 || NETSTANDARD2_0 [](start = 3, length = 43)

nit/suggestion: should this have been a feature flag instead? Instead of having to write all of this again and again? - even more when adding a new target all the same places will have to be modified

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -129,7 +129,7 @@ internal static class LogMessages
         public const string IDX10628 = "IDX10628: Cannot set the MinimumSymmetricKeySizeInBits to less than '{0}'.";
         public const string IDX10630 = "IDX10630: The '{0}' for signing cannot be smaller than '{1}' bits. KeySize: '{2}'.";
         public const string IDX10631 = "IDX10631: The '{0}' for verifying cannot be smaller than '{1}' bits. KeySize: '{2}'.";
-        public const string IDX10634 = "IDX10634: Unable to create the SignatureProvider.\nAlgorithm: '{0}', SecurityKey: '{1}'\n is not supported.";
+        public const string IDX10634 = "IDX10634: Unable to create the SignatureProvider.\nAlgorithm: '{0}', SecurityKey: '{1}'\n is not supported. The list of supported algorithms is available here: https://aka.ms/IdentityModel/supported-algorithms";
```
henrik-me: >The list of supported algorithms is available here: https://aka.ms/IdentityModel/supported-algorithms" [](start = 148, length = 102)

nice!

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
GeoK: addressed with daa3d717fd9d56f3d246d0f5c329005b88f7872b

src/Microsoft.IdentityModel.Tokens/SupportedAlgorithms.cs
GeoK: addressed with daa3d717fd9d56f3d246d0f5c329005b88f7872b

src/Microsoft.IdentityModel.Tokens/SupportedAlgorithms.cs
GeoK: addressed with daa3d717fd9d56f3d246d0f5c329005b88f7872b

test/Microsoft.IdentityModel.Tokens.Tests/AsymmetricSignatureTests.cs
GeoK: addressed with daa3d717fd9d56f3d246d0f5c329005b88f7872b

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
GeoK: I like the feature flag approach.
Brent implied [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1166#discussion_r284053291) that we will do some refactoring so we might leverage the feature flag approach there.

brentschmaltz:
:shipit: I think we can do without the 'runtime' check and ask customers to use the 4.6.1 target.

PR #1174, opened 2019-05-16
Updated the JsonWebTokenHandlerTests.CreateJWEUsingSecurityTokenDescriptor tests

mafurman:
Some of our tests are failing intermittently because we sometimes generate timestamps on token creation.

Values for the 'exp', 'iat', and 'nbf' claims have been provided for test cases that may fail because of this issue. A test case that was present in the CreateJWS tests (but not CreateJWE) was also added in.

The dev5x branch will also need these changes to be pulled in (along with a similar fix to the CreateJWSUsingSecurityTokenDescriptor tests that is present in dev but not in dev5x).

### DIFF
--- a/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
+++ b/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
@@ -260,7 +260,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
                         {
                             { JwtRegisteredClaimNames.Aud, Default.Audience },
                             { JwtRegisteredClaimNames.Exp, EpochTime.GetIntDate(Default.Expires).ToString() },
-                            { JwtRegisteredClaimNames.Iat, EpochTime.GetIntDate(Default.NotBefore).ToString() },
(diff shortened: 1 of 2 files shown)


test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -258,11 +258,17 @@ public static TheoryData<CreateTokenTheoryData> CreateJWEUsingSecurityTokenDescr
                         TestId = "TokenDescriptorClaimsNull",
                         Payload = new JObject()
                         {
-                            { JwtRegisteredClaimNames.Aud, Default.Audience }
+                            { JwtRegisteredClaimNames.Aud, Default.Audience },
+                            { JwtRegisteredClaimNames.Exp, EpochTime.GetIntDate(Default.Expires).ToString() },
+                            { JwtRegisteredClaimNames.Iat, EpochTime.GetIntDate(Default.NotBefore).ToString() },
```
brentschmaltz: nit: can we have a Default.IssuedAt?

mafurman: I didn't notice this earlier, but we do have a property called Default.IssueInstant. I've switched to using this value for the 'iat' claim.

Issue #1105, opened 2019-01-28
Saml2SecurityTokenHandler.ValidateConditions does not check for missing audience


explunit:
`Saml2SecurityTokenHandler.ValidateConditions` simply iterates over the inbound `samlToken.Assertion.Conditions.AudienceRestrictions`. It does not check for the case where TokenValidationParameters.ValidateAudience is true but no AudienceRestrictions were found in the inbound token.

explunit:
To ask it a different way: this was checked in the old codebase: https://github.com/Microsoft/referencesource/blob/60a4f8b853f60a424e36c7bf60f9b5b5f1973ed1/System.IdentityModel/System/IdentityModel/Tokens/SamlSecurityTokenHandler.cs#L939-L942

But not the new? Why?

brentschmaltz:
@explunit we need to add this flag: enforceAudienceRestriction.
This should be similar to TokenValidationParameters.RequireSignedTokens

How about RequireAudience, default true.

Issue #1104, opened 2019-01-28
Saml2SecurityTokenHandler.ValidateAudience is never called


explunit:
`Saml2SecurityTokenHandler.ValidateAudience` is never called. Thus it's not possible to override this logic like it is with other `protected` methods in the same class, e.g. `ValidateIssuer`.

I am aware of the ability to provide an `TokenValidationParameters.AudienceValidator` delegate, but the class should be consistent across methods. For example, it's possible to provide an `IssuerValidator` **or** to override `ValidateIssuer`

brentschmaltz:
@explunit it is a bit confusing, sorry for the this. We'll try and clean it up for 5.4.1.

We call Validators.ValidateAudience in ValidateConditions.
It may be better to the virtual ValidateAudience from ValidateToken, which would call
Validators.ValidateAudience.

We will put in the other fix where 'audience' is required.

mafurman:
This should be fixed in the SamlSecurityTokenHandler as well.

PR #1179, opened 2019-05-17
Added the 'RequireAudience' flag to TokenValidationParameters

mafurman:
Also ensured Saml2SecurityTokenHandler and SamlSecurityTokenHandler are calling their virtual validate methods (rather than directly calling Validators.Validate).

Addresses #1105 and  #1104.


henrik-me:
>         //}

remove?

---
Refers to: src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs:277 in 47c73ce. [](commit_id = 47c73cedda86b0254812cb636604d85b1d83e4d2, deletion_comment = False)

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -509,6 +511,12 @@ public string RoleClaimType
             set;
```
henrik-me: nit: should app properties use 1 line instead this?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -896,10 +909,13 @@ protected virtual void ValidateConditions(Saml2SecurityToken samlToken, TokenVal
             if (samlToken.Assertion == null)
                 throw LogArgumentNullException(nameof(samlToken.Assertion));
 
+            if (validationParameters.RequireAudience && samlToken.Assertion.Conditions == null)
```
henrik-me: >samlToken.Assertion.Conditions == null [](start = 56, length = 38)

for saml tokens you are also looking for conditions.count == 0, should the same be done here? (will be caught later, however for consistency)?

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -915,18 +915,32 @@ protected virtual void ValidateConditions(SamlSecurityToken securityToken, Token
...
                 return;
+            }
 
-            Validators.ValidateLifetime(securityToken.Assertion.Conditions.NotBefore, securityToken.Assertion.Conditions.NotOnOrAfter, securityToken, validationParameters);
+            ValidateLifetime(securityToken.Assertion.Conditions.NotBefore, securityToken.Assertion.Conditions.NotOnOrAfter, securityToken, validationParameters);
 
+            var foundAudienceRestriction = false;
             if (securityToken.Assertion.Conditions.Conditions.ElementAt(0) is SamlAudienceRestrictionCondition)
             {
                 foreach (var condition in securityToken.Assertion.Conditions.Conditions)
```
henrik-me: this code is not changed, wondering why this needs to be different from how the loop is setup for SAML2

henrik-me:
approving. my comments are mostly to educate me

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
brentschmaltz: because Saml and Saml2 had to be different

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
brentschmaltz: Its related to we check lifetime before audience.

test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlSecurityTokenHandlerTests.cs
```diff
@@ -757,6 +757,37 @@ public static TheoryData<SamlTheoryData> ValidateTokenTheoryData
...
+                            IssuerSigningKey = KeyingMaterial.DefaultAADSigningKey,
+                            ValidAudience = "spn:fe78e0b4-6fe7-47e6-812c-fb75cee266a4",
+                            ValidateLifetime = false,
+                            ValidateIssuer = false,
+                            RequireSignedTokens = false
+                        }
+                    },
+                    new SamlTheoryData
+                    {
+
```
brentschmaltz: nit: extra space

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
henrik-me: makes sense, however the type is not necessarily "validated" it's assumed it's of type Saml2AudienceRestrictionCondition ? (only the first element type is validated). probably missing something here.

brentschmaltz: @henrik-me I think you are correct, the logic requires the first condition to be a AudienceRestriction.
The first clause with ElementAt(0) should be removed.

mafurman: I removed the clause requiring the first condition to be an AudienceRestrictionCondition from the SamlSecurityTokenHandler.  I also added some extra null checks for the Conditions property (for both Saml1 and Saml2) as if it's null we will currently throw a null reference exception. Based on this [spec](http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf) it seems that the Conditions element is optional, so we probably shouldn't be throwing.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
mafurman: @brentschmaltz If `Conditions.Count() == 0` we'll end up returning early in the case of SamlSecurityTokenHandler, so we'll never call ValidateLifetime. However, in the case of Saml2SecurityTokenHandler we will. Shouldn't we be consistent about this?

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
mafurman: We seem to be inconsistent about this, though I'm not sure which format is preferred...

test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlSecurityTokenHandlerTests.cs
mafurman: Fixed!

brentschmaltz:
:shipit:

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
mafurman: I got rid of the Conditions.Count() == 0 in the SamlSecurityTokenHandler so that we're consistent.  Even if Conditions.Count() == 0 it's possible we might still need to call ValidateLifetime (since NotBefore and NotOnOrAfter are attributes on the Condition).

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
henrik-me: not sure either... but seemed like "default" properties seemed to follow the "collapsed" guidelines. 

---
In reply to: [285785710](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1179#discussion_r285785710) [](ancestors = 285785710)

src/Microsoft.IdentityModel.Tokens.Saml/Saml/LogMessages.cs
```diff
@@ -38,6 +38,7 @@ internal static class LogMessages
 #pragma warning disable 1591
         // SecurityTokenHandler messages
         internal const string IDX11400 = "IDX11400: The '{0}', can only process SecurityTokens of type: '{1}'. The SecurityToken received is of type: '{2}'.";
+        internal const string IDX11401 = "IDX11401: Unable to validate token. TokenValidationParameters.RequireAudience is true but no AudienceRestrictions were found in the inbound token.";
```
henrik-me: >. [](start = 103, length = 1)

nit: extra space?

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -1126,7 +1131,10 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
             var samlToken = ValidateSignature(token, validationParameters);
             ValidateConditions(samlToken, validationParameters);
             var issuer = ValidateIssuer(samlToken.Issuer, samlToken, validationParameters);
-            ValidateTokenReplay(samlToken.Assertion.Conditions.NotOnOrAfter, token, validationParameters);
+
+            if (samlToken.Assertion.Conditions != null)
```
henrik-me: >samlToken.Assertion [](start = 16, length = 19)

could this be null?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -204,7 +204,9 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
             ValidateConditions(samlToken, validationParameters);
             ValidateSubject(samlToken, validationParameters);
             var issuer = ValidateIssuer(samlToken.Issuer, samlToken, validationParameters);
-            ValidateTokenReplay(samlToken.Assertion.Conditions.NotOnOrAfter, token, validationParameters);
+            if (samlToken.Assertion.Conditions != null)
```
henrik-me: >samlToken.Assertion. [](start = 16, length = 20)

null check?

brentschmaltz: Saml2SecurityToken must have a non-null Assertion.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
brentschmaltz: From a derived SamlToken yes. I opened this issue to fix that.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1185

brentschmaltz:
Good job in noticing the skipped validation of time.
:shipit:

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
mafurman: I created an issue to track this so we can address it in the future: #1186

PR #1180, opened 2019-05-18
Ensured that JsonWebTokenHandler makes use of the issuer returned by the IssuerValidator

### LINKED ISSUE #1164: JsonWebTokenHandler does not use the issuer returned from ValidateIssuer when creating claims.
JsonWebTokenHandler will produce slightly different claims if the IssuerValidator returns a value other than the 'iss' value than the JwtSecurityTokenHandler.

They should return the same claims when mapping is turned off.

mafurman:
Fixes #1164.


test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -435,6 +457,24 @@ public static TheoryData<CreateTokenTheoryData> CreateJWSTheoryData
                             ValidIssuer = Default.Issuer
                         }
                     },
+                      new CreateTokenTheoryData
```
brentschmaltz: nit: spacing

mafurman: Fixed!

Issue #1012, opened 2018-08-28
Feature request: Allow nested Signatures


Tarig0:
Looks like the only reason why signatures, that include another signature in its calculation, are not supported is because the XML token stream will take out ALL signatures and not just the first instance of it.

I think, with the assumption that the first signature is the one being verified, an element remove flag could be added and if true include any other signature elements.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/6e7a53e241e4566998d3bf365f03acd0da699a31/src/Microsoft.IdentityModel.Xml/XmlTokenStreamWriter.cs#L158-L160

Tarig0:
Looks like this was removed at 
commit 8cd95c84f1e833fcdfbcb54b56bd8dadb858e1c1

brentschmaltz:
@Tarig0 yes, we deliberately only support a single signature as we were focused on Saml tokens. I agree that xmldsig has the ability for multiple signatures. Can you share why you need this support?

Tarig0:
Did my last comment clear up what I meant?

brentschmaltz:
@Tarig0 it was simply a work scoping tactic to allow a single signature. The code was developed to extend to multiple.
What is your time frame?

I think this fit with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/759

AndersAbel:
@Tarig0 I'm afraid nested signatures are more complex than merely taking the first. There is a horrible class attacks on SAML2 implemenations that use XML Signature Wrapping in various creative ways. The short-cut the team has taken so far with only one signature and hard coding it to one reference which is hard coded to be the entire document is safe. But as soon as those assumptions are no longer valid, I think that steps must be taken to ensure that the right signature (exact location in XML doc must follow SAML2 spec) is validated and that it contain exactly one reference, which points to the right element (once again according to the SAML2 spec). The filtering out of the signature itself is done by the enveloped signature transform, which then should be used to ensure that the right signature block is excluded.

Sorry if I'm a bit pessimistic, but in my work with SAML2P I've found out that XML signatures are **hard**. Not just to implement, but even to use by calling the APIs in a secure way.

Tarig0:
The assumptions that I have is.

1. The reference is for a node that the signature is a direct child of.
2. The xmlreader that is wrapped will be windowed to said node.

Little more fleshed out

[...]

PR #1181, opened 2019-05-19
Add processing of XmlReader to Saml(2)SecurityTokenHandler

brentschmaltz:
This PR is designed to enhance the support of reading SAML (1 and 2) tokens with a xmlreader.

SAML assertions now have the canonical string available when checking for replay.

The EnvelopedSignatureReader has added support for reading xml that contains multiple objects each with its own signature while the outer element is signed. Each object can only have a single signature, but each signature is distinct. The outer element's signature is over the entire contents where embedded signatures are not significant.

To achieve this a new constructor is used to pass an interface that will read the inner signed elements.

There was some cleanup of the EnvelopedSignautreWriter where Stream.Null was used where possible. 

Care was takes to determine if a reader is able to canonicalize and if so, make use of that fact. 

Issues:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1042
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1012


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
```diff
@@ -292,7 +295,10 @@ public virtual Saml2Assertion ReadAssertion(XmlReader reader)
                 }
 
                 // attach signedXml for validation of signature
```
GeoK: nit: seem like one this comment doesn't belong.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
```diff
@@ -199,12 +199,15 @@ protected virtual Saml2Advice ReadAdvice(XmlDictionaryReader reader)
         /// <returns>A <see cref="Saml2Assertion"/> instance.</returns>
         public virtual Saml2Assertion ReadAssertion(XmlReader reader)
         {
+            if (reader == null)
```
GeoK: For consistency with SamlSerializer, `if (reader == null)` check could be removed and `XmlUtil.CheckReaderOnEntry(reader, Saml2Constants.Elements.Assertion, Saml2Constants.Namespace);` line could be placed on the top of this method.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlAssertion.cs
```diff
@@ -152,6 +157,46 @@ public string MinorVersion
         /// </summary>
         public Signature Signature { get; set; }
 
+        /// <summary>
+        /// Gets the canonicalized (ExclusiveC14n) representation without comments.
+        /// </summary>
+        public string CanonicalString
```
GeoK: Would it make sense to convert this property to a method, allowing users to specify includeComments and inclusiveNamespacesPrefixList arguments?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Assertion.cs
```diff
@@ -84,6 +89,46 @@ public Saml2Conditions Conditions
             set;
         }
 
+        /// <summary>
+        /// Gets the canonicalized (ExclusiveC14n) representation without comments.
+        /// </summary>
+        public string CanonicalString
```
GeoK: Would it make sense to convert this property to a method, allowing users to specify includeComments and inclusiveNamespacesPrefixList arguments?

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2AssertionTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        [Fact]
+        public void CanonicalString()
+        {
+            var context = new CompareContext($"{this}.CanonicalString");
+
+            var assertion = new Saml2Assertion(new Saml2NameIdentifier("nameIdentifier"));
+            var canonicalString = assertion.CanonicalString;
+            var canonicalString2 = assertion.CanonicalString;
+
+            IdentityComparer.AreStringsEqual(canonicalString, canonicalString2, context);
```
GeoK: not sure what is the intention here.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlAssertion.cs
```diff
@@ -152,6 +157,46 @@ public string MinorVersion
         /// </summary>
         public Signature Signature { get; set; }
 
+        /// <summary>
+        /// Gets the canonicalized (ExclusiveC14n) representation without comments.
```
GeoK: Perhaps we should document that null will be returned in a case that we can't get the canonical string (we swallow all exceptions).

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Assertion.cs
```diff
@@ -84,6 +89,46 @@ public Saml2Conditions Conditions
             set;
         }
 
+        /// <summary>
+        /// Gets the canonicalized (ExclusiveC14n) representation without comments.
```
GeoK: Perhaps we should document that null will be returned in a case that we can't get the canonical string (we swallow all exceptions).

src/Microsoft.IdentityModel.Xml/XmlTokenStream.cs
```diff
@@ -116,8 +119,10 @@ public void WriteTo(XmlWriter writer)
             if (writer == null)
                 throw LogArgumentNullException(nameof(writer));
 
-            var streamWriter = new XmlTokenStreamWriter(_entries, _excludedElement, _excludedElementNamespace);
-            streamWriter.WriteTo(writer);
+            var streamWriter = new XmlTokenStreamWriter(this);
+            streamWriter.WriteTo(writer, _excludedElement, _excludedElementNamespace);
         }
+
+        internal ReadOnlyCollection<XmlToken> XmlTokens => _xmlTokens.AsReadOnly();
```
GeoK: Can we return return non-readonly collection here? Perhaps it is perfectly valid to leave it as readonly by design, but I wanted to clarify.
Thinking from the perspective of EnvelopedSignatureWriter, the only way to manipulate XmlToken position or content of an XmlToken within the stream is to create a new stream out of the existing one.

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2AssertionTests.cs
brentschmaltz: I wanted to make sure it worked without any signature.

src/Microsoft.IdentityModel.Xml/XmlTokenStream.cs
brentschmaltz: I think we have to leave it readonly, otherwise we could get different results.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
brentschmaltz: We wanted to be explicit about not supporting encrypted assertions. For that we need the check below.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlAssertion.cs
brentschmaltz: Probably not, the comment is a bit misleading.

If the Assertion was created by the EnvelopedSignatureReader, the XmlTokenStream has the correct nodes. However, i don't believe we are remembering the 'includeComments and inclusiveNamespacesPrefixList when we capture the XmlTokenStream. I believe we may need to do that.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Assertion.cs
brentschmaltz: Same as above.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlAssertion.cs
brentschmaltz: Perhaps we shouldn't return anything if the XmlTokenStream is not set.
We may want to make this internal for this release.
It is used in two places:
1. When performing OnBehalfOf
2. For Replaying of tokens, we want the same string.

src/Microsoft.IdentityModel.Xml/EnvelopedSignatureReader.cs
```diff
@@ -57,6 +61,24 @@ public EnvelopedSignatureReader(XmlReader reader)
...
+        /// <param name="xmlElementReader"> specified to read inner objects.</param>
+        /// <remarks>If a &lt;Signature> element is found, the <see cref="Signature"/> will be set.</remarks>
+        /// <exception cref="ArgumentNullException">if <paramref name="reader"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="xmlElementReader"/> is null.</exception>
+        public EnvelopedSignatureReader(XmlReader reader, IXmlElementReader xmlElementReader)
+        {
+            if (reader == null)
+                throw LogArgumentNullException(nameof(reader));
+
+            _xmlElementReader = xmlElementReader ?? throw LogArgumentNullException(nameof(xmlElementReader));
```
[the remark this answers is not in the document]
brentschmaltz: We can do that, I normally don't but I don't see any reason not to.

src/Microsoft.IdentityModel.Xml/EnvelopedSignatureReader.cs
```diff
@@ -26,7 +26,10 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
```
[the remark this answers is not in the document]
brentschmaltz: Thanks.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
brentschmaltz: removed, thanks.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Assertion.cs
brentschmaltz: See above comment with SamlAssertion.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlAssertion.cs
brentschmaltz: We decided to make this internal for now. It is only available when the EnvelopedSignatureReader is involved. We pass the canonicalized string for replay detection and for the bootstrap token for OBO flows.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlAssertion.cs
brentschmaltz: Made this internal.

PR #1187, opened 2019-05-22
Upgrade System.Net.Http to 4.3.2

GeoK:
Our tooling reports a security vulnerability: 

https://github.com/aspnet/Announcements/issues/239
https://github.com/dotnet/corefx/issues/34428

The library is not affected, but we shouldn't use the System.Net.Http 4.3.0 package.
I used this opportunity to also upgrade version of the netcoreapp testing target, simplify and cleanup test-related properties, and centralize common properties.

---

* Explicitly upgraded System.Net.Http from 4.3.0 to 4.3.2.
* Upgraded Netstandard.Library to 2.0.3 as 1.6.1 includes System.Net.Http 4.3.0 as a transitive dependency.
* Removed redundant references and centralized package reference versions to dependencies.props and dependenciesTest.props files.
* Upgraded DotNetCoreAppRuntimeVersion from version 2.0.0 to 2.2.3.
* Upgraded netcoreapp testing target from version 2.0 to 2.2.
* Used common.props and commontTests.props to centralize common logic.
* Added TestOnlyCoreTargets property to hold our core test targets list.
* Added NET_CORE constant to replace NETCOREAPPx symbols and eliminate source code changes when new netcoreapp testing target is used.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Microsoft.IdentityModel.Protocols.csproj
+++ b/src/Microsoft.IdentityModel.Protocols/Microsoft.IdentityModel.Protocols.csproj
@@ -7,8 +7,6 @@
     <Description>Provides base protocol support for OpenIdConnect and WsFederation.</Description>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageId>Microsoft.IdentityModel.Protocols</PackageId>
-    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(SrcTargets)</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">$(SrcStandardTargets)</TargetFrameworks>
     <PackageTags>.NET;Windows;Authentication;Identity;OpenIdConnect;WsFederation</PackageTags>
   </PropertyGroup>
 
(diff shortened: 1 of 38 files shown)


src/Microsoft.IdentityModel.Protocols/Microsoft.IdentityModel.Protocols.csproj
```diff
@@ -25,7 +25,7 @@
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.4'">
     <PackageReference Include="System.Collections.Specialized" Version="4.3.0-*" />
     <PackageReference Include="System.Diagnostics.Contracts" Version="4.3.0-*" />
```
henrik-me: should these also use updated latest version?

GeoK: These are not affected by the vulnerability and are already at the latest version (4.3.0).

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests.csproj
```diff
@@ -30,9 +27,9 @@
     <PackageReference Include="Newtonsoft.Json" Version="$(JsonNetVersion)" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.2' ">
     <PackageReference Include="System.Diagnostics.Process" Version="$(CoreFxVersion)" />
-    <PackageReference Include="System.Net.Http" Version="$(CoreFxVersion)" />
+    <PackageReference Include="System.Net.Http" Version="4.3.2" />
```
brentschmaltz: I have considered having a version for each PackageReference to make sure all assemblies are in sync. It seemed like overkill, because most were the same version, by maybe not.

In this case we would have: 
SystemNetHttpVerson = 4.3.2 (in dependencies and dependenciesTest)

Then in each csproj:
Version="$(SystemNetHttpVersion)"

What do you think?

GeoK: @brentschmaltz - I agree with the suggestion. Changes are included (I forced-pushed the commit).

Issue #1169, opened 2019-05-14
JsonWebToken, Getting values from header and payload needs to consider when types don't match


brentschmaltz:
If a JsonWebToken has the json { "number", 12345 }. TryGetPayloadValue("number", out string number) will return true and number == "12345" a string.
It will be tricky when the value is 'nil' and the type passed in is not nullable.

brentschmaltz:
If a JsonWebToken has the json { "number", 12345 }. TryGetPayloadValue("number", out string number) will return true and number == "12345" a string.
It will be trick when the value is 'nil' and the type passed in is not nullable.

brentschmaltz:
We should review how we return all values.

mafurman:
If a user passes in a Claim as T, we should return a populated Claim object. If they pass in IList<Claim> or ICollection<Claim> we should follow the same behavior as the Claims property (only go one deep).

mafurman:
We should also add "Claim GetClaim(string jsonKeyName)" method that will allow the user to query for a specific Claim directly. If the type of the Claim is JArray or JObject we should return that as the type and return the value as a string.

This should return the same value that passing Claim as T returns (more explanation above).

mafurman:
Since we agreed that this issue only relates to the payload (values returned by the Claims property), I created a separate issue to implement similar functionality for header claims: #1188.

PR #1189, opened 2019-05-23
Added GetClaim() and TryGetClaim() methods to JsonWebToken

mafurman:
Addresses #1169.

Changes made:

- Added GetClaim() and TryGetClaim() methods to JsonWebToken (as well as associated tests).
- Added logic to retrieve a Claim from the GetPayloadValue() and TryGetPayloadValue() methods.
- Added in tests comparing the GetClaim()/TryGetClaim() versus the GetPayloadValue<Claim>()/TryGetPayloadValue<Claim>() methods.
- Some minor formatting changes.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -69,7 +69,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             {
                 var tokenParts = jwtEncodedString.Split('.');
                 Decode(tokenParts, jwtEncodedString);
-            } else
+            }
+            else
                 throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14100, jwtEncodedString)));
         }
 
@@ -91,7 +92,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             try
             {
                 Header = JObject.Parse(header);
-            } 
+            }
             catch (Exception ex)
             {
                 throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14301, header), ex));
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -488,15 +489,48 @@ private static string GetClaimValueType(object obj)
...
+        /// </summary>
+        /// <remarks>If the key has no corresponding value, this method will throw.</remarks>   
+        public Claim GetClaim(string key)
+        {
+            string issuer = Issuer ?? ClaimsIdentity.DefaultIssuer;
+
+            if (!Payload.TryGetValue(key, out var jTokenValue))
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14304, key)));
+
+            if (jTokenValue == null)
```
GeoK: I'm not sure that this will ever be null.
Perhaps we should check for null with `if (jTokenValue.Type == JTokenType.Null)`.

There is no test where the claim below will returned:
`return new Claim(key, string.Empty, JsonClaimValueTypes.JsonNull, issuer, issuer);`

mafurman: @GeoK You're right!

I wrote the tests for this incorrectly; they should have been checking to make sure that the value of the 'nill' claim was an empty string, but instead the expected value was "null." 

This was the value my tests were checking against (which is incorrect):
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fbc717d431cd90e51eb50e325e513a512601a6a4/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs#L54

I am wondering though, does it make sense to return an empty string if the value is null?

GeoK: @mafurman - I don't think that it really matters as long as the type is `JsonClaimValueTypes.JsonNull`. 
null might be a better option, but a Claim can't be created with a value set as null.

brentschmaltz: @mafurman would it make sense to have the value == string.empty when the type is JsonClaimValueTypes.JsonNull?

mafurman: @brentschmaltz I think that having value == string.Empty is the best option considering that's consistent with the behavior of JwtSecurityToken.

brentschmaltz: 👍 :shipit:

Issue #1083, opened 2018-12-19
Exception in EnvelopedSignatureWriter when Signature element is first child of element


Tragedian:
In trying to create a SAML Federation Metadata document, there appears to be a bug/unsupported scenario in the `EnvelopedSignatureWriter` when the Signature element is the first child:

```
var credentials = new SigningCredentials(...);
var settings = new XmlWriterSettings
{
    Encoding = new UTF8Encoding(false)
};
var buffer = new MemoryStream();
var writer = new EnvelopedSignatureWriter(XmlWriter.Create(buffer, settings), credentials, "_id");

writer.WriteStartElement("EntityDescriptor", "urn:oasis:names:tc:SAML:2.0:metadata");

writer.WriteSignature();

writer.WriteEndElement();
```

When the writer tries to emit the signature after the root element is written, the following exception is thrown:

```
System.Xml.XmlException: Name cannot begin with the '<' character, hexadecimal value 0x3C. Line 1, position 18.
   at System.Xml.XmlExceptionHelper.ThrowXmlException(XmlDictionaryReader reader, XmlException exception)
   at System.Xml.XmlUTF8TextReader.VerifyNCName(String s)
   at System.Xml.XmlUTF8TextReader.ReadQualifiedName(PrefixHandle prefix, StringHandle localName)
   at System.Xml.XmlUTF8TextReader.ReadAttributes()
   at System.Xml.XmlUTF8TextReader.ReadStartElement()
   at System.Xml.XmlUTF8TextReader.Read()
   at System.Xml.XmlBaseReader.MoveToContent()
   at Microsoft.IdentityModel.Xml.EnvelopedSignatureWriter.OnEndRootElement() in C:\agent2\_work\56\s\src\Microsoft.IdentityModel.Xml\EnvelopedSignatureWriter.cs:line 139
```

This appears to be because the inner writer is still positioned inside the unfinished start element when the `EnvelopedSignatureWriter` captures the signature position in its internal buffer.

I would expect the behaviour to be the same as calling `WriteStartElement`, `WriteElementString` or `WriteValue`, which will "finish" the started element before proceeding to start writing the next node as content of the element.

As a workaround, you can call `WriteComment` to finish the started element, but that does emit an unwanted comment before the signature.

Tragedian:
[...]

PR #1195, opened 2019-05-29
Support Signature as an inner element with EnvelopedSignatureWriter

GeoK:
* Use InnerWriter to write canonical bytes (without signature).
* Use InternalWriter to write non-canonical xml (including signature).
* When WriteSignature() method is called, write only a placeholder signature element using InternalWriter.
* After writing the last (end) element, create XmlTokenStream out of the InternalWriter's stream and write that XmlTokenStream into the original writer. When writing the XmlTokenStream, replace signature [placeholder] element with the real signature (using DSigSerializer). This behavior is followed only if WriteSignature method was invoked explicitly. Otherwise, the default behavior is to generate the signature at the end and to write it directly to the original stream.
* Improve EnvelopedSignatureReader and XmlTokenStreamReader to allow reading an xml where an outer signature is written right after an assertion.
* Add RoundTripWsMetadata test with a signature placed at different positions.
* Add RountTripSamlp* tests to showcase writing an xml with multiple signatures and later reading it at with an EnvelopedSignatureReader and validating both inner and outer signature.

Resolves: #1083

### DIFF
--- a/src/Microsoft.IdentityModel.Xml/DelegatingXmlDictionaryWriter.cs
+++ b/src/Microsoft.IdentityModel.Xml/DelegatingXmlDictionaryWriter.cs
@@ -37,6 +37,7 @@ namespace Microsoft.IdentityModel.Xml
     public class DelegatingXmlDictionaryWriter : XmlDictionaryWriter
     {
         private XmlDictionaryWriter _innerWriter;
+        private XmlDictionaryWriter _internalWriter;
         private XmlDictionaryWriter _tracingWriter;
 
         /// <summary>
@@ -66,6 +67,16 @@ namespace Microsoft.IdentityModel.Xml
             set => _innerWriter = value ?? throw LogArgumentNullException(nameof(value));
         }
 
+        /// <summary>
+        /// Gets or sets the InternalWriter.
+        /// </summary>
+        /// <exception cref="ArgumentNullException"> if 'value' is null.</exception>
+        internal XmlDictionaryWriter InternalWriter
+        {
+            get => _internalWriter;
+            set => _internalWriter = value ?? throw LogArgumentNullException(nameof(value));
+        }
+
 #if NET45 || NET451
         /// <summary>
         /// Closes the underlying stream.
@@ -73,8 +84,8 @@ namespace Microsoft.IdentityModel.Xml
         public override void Close()
         {
             UseInnerWriter.Close();
-            if (TracingWriter != null)
-                TracingWriter.Close();
+            TracingWriter?.Close();
+            InternalWriter?.Close();
         }
 #endif
 
@@ -84,8 +95,8 @@ namespace Microsoft.IdentityModel.Xml
         public override void Flush()
         {
             UseInnerWriter.Flush();
-            if (TracingWriter != null)
-                TracingWriter.Flush();
+            TracingWriter?.Flush();
+            InternalWriter?.Flush();
         }
 
         /// <summary>
@@ -97,8 +108,8 @@ namespace Microsoft.IdentityModel.Xml
         public override void WriteBase64(byte[] buffer, int index, int count)
         {
             UseInnerWriter.WriteBase64(buffer, index, count);
-            if (TracingWriter != null)
-                TracingWriter.WriteBase64(buffer, index, count);
+            TracingWriter?.WriteBase64(buffer, index, count);
+            InternalWriter?.WriteBase64(buffer, index, count);
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


test/Microsoft.IdentityModel.Xml.Tests/EnvelopedSignatureWriterTests.cs
```diff
@@ -135,6 +138,277 @@ public static TheoryData<EnvelopedSignatureTheoryData> RoundTripSaml2TheoryData
...
+                esw.WriteAttributeString("Destination", "https://tnia.eidentita.cz/fpsts/processRequest.aspx");
+                esw.WriteAttributeString("ID", "id-uAOhNLe7abGB6WGPk");
+                esw.WriteAttributeString("InResponseTo", "ida5714d006fcc430c92aacf34ab30b166");
+                esw.WriteAttributeString("IssueInstant", "2019-04-08T10:30:49Z");
+                esw.WriteAttributeString("Version", "2.0");
+                esw.WriteStartElement("ns1", "Issuer");
+                esw.WriteAttributeString("Format", "urn:oasis:names:tc:SAML:2.0:nameid-format:entity");
+                esw.WriteString("https://mojeid.regtest.nic.cz/saml/idp.xml");
+                esw.WriteEndElement();
+                esw.WriteSignature();
```
brentschmaltz: If the SamlToken is written before this WriteSignature, the signatures will not validate.
We will need some logic to 'know' which signature is significant.

Message: Microsoft.IdentityModel.TestUtils.TestException : Microsoft.IdentityModel.Tokens.Xml.Tests.EnvelopedSignatureWriterTests.RoundTripSamlP


exception != null, expectedException.TypeExpected == null.
exception: Microsoft.IdentityModel.Xml.XmlValidationException: IDX30201: The 'Reference' did not validate: '#id-uAOhNLe7abGB6WGPk'.

GeoK: @brentschmaltz - Writing the signature explicitly (using WriteSignature) or using the default behavior (adding the signature as the last element) didn't change the digest and signature values in my tests.
I guess that the problem lies somewhere in EnvelopeSignatureReader/XmlTokenStreamReader.

When adding a signature after an assertion, an outer signature doesn't get recorded and therefore remains included in digested canonical bytes, leading to the mismatch in digest values.

src/Microsoft.IdentityModel.Xml/DelegatingXmlDictionaryWriter.cs
```diff
@@ -66,15 +67,25 @@ protected XmlDictionaryWriter InnerWriter
...
         }
 
+        /// <summary>
+        /// Gets or sets the InternalWriter.
+        /// </summary>
+        /// <exception cref="ArgumentNullException"> if 'value' is null.</exception>
+        internal XmlDictionaryWriter InternalWriter
+        {
+            get => _internalWriter;
+            set => _internalWriter = value ?? throw LogArgumentNullException(nameof(value));
```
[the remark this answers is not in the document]
henrik-me: Assuming the stack trace will still say which property was affected and thus this will be implicit? (meaning I think this is ok, but perhaps it makes more sense to do as you suggest)

test/Microsoft.IdentityModel.Xml.Tests/EnvelopedSignatureWriterTests.cs
brentschmaltz: Yeah essentially each envelope signature writer needs to recognize it's 'signature'. One way may be to use a freshly generated Guid to mark the spot rather than a shared value.

brentschmaltz: To obtain this exception, I moved the writing of the Saml2Assertion before the call to esw.WriteSignature as in:
````
esw.WriteEndElement();
Saml2Serializer samlSerializer = new Saml2Serializer(); 
Saml2Assertion assertion = CreateAssertion(samlpTokenSigningCredentials); 
samlSerializer.WriteAssertion(esw, assertion); 
esw.WriteSignature();
````

src/Microsoft.IdentityModel.Xml/DelegatingXmlDictionaryWriter.cs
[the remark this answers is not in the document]
GeoK: Correct, the stack trace will indicate the affected property. Looking at the [framework design guidance](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types#argumentexception-argumentnullexception-and-argumentoutofrangeexception) the current approach looks fine:
>  DO use value for the name of the implicit value parameter of property setters.

test/Microsoft.IdentityModel.Xml.Tests/EnvelopedSignatureWriterTests.cs
GeoK: Addressed this issue by slightly altering the logic in EnvelopedSignatureReader and XmlTokenStreamReader classes.

Issue #1132, opened 2019-04-16
JwtSecurityTokenHandler.ValidateToken is failing with IDX10511; exception message needs to be improved


shitada:
JwtSecurityTokenHandler.ValidateToken is failed as IDX10511 even though Graph API call is succeeded by using the token.
But the validation of the token of Azure AD Graph is succeeded. 
Are there any improvements in code to resolve the problem? Or does the problem relies on Azure AD side?

using Microsoft.IdentityModel.Clients.ActiveDirectory;
using Microsoft.IdentityModel.Logging;
using Microsoft.IdentityModel.Protocols;
using Microsoft.IdentityModel.Protocols.OpenIdConnect;
using Microsoft.IdentityModel.Tokens;
using Newtonsoft.Json.Linq;
using System;
using System.Collections.Generic;
using System.IdentityModel.Tokens.Jwt;
using System.Linq;
using System.Net.Http;
using System.Text;
using System.Threading.Tasks;

namespace ValidateToken
{
    class Program
    {
        static void Main(string[] args)
        {
            //aquire token
            var tenantId = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx";
            var clientId = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx";
            var clientSecret = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx";
            var aadInstance = "https://login.microsoftonline.com/";
            var graphResourceID = "https://graph.microsoft.com";
            //var graphResourceID = "https://graph.windows.net";

            UserCredential userCredential = new UserCredential("xxxxx@xxxxx.onmicrosoft.com");
            UserIdentifier userIdentifier = new UserIdentifier("xxxxx@xxxxx.onmicrosoft.com", UserIdentifierType.RequiredDisplayableId);
            ClientCredential clientcred = new ClientCredential(clientId, clientSecret);
            AuthenticationContext authenticationContext = new AuthenticationContext(aadInstance + tenantId);
            AuthenticationResult authenticationResult = authenticationContext.AcquireTokenAsync(graphResourceID, clientId, new Uri("http://localhost"), new PlatformParameters(PromptBehavior.Always)).Result;

            Console.WriteLine(authenticationResult.AccessToken);

            //call Graph API, the request is succeeded.
            using (var client = new HttpClient())
            {
[...]

PR #1196, opened 2019-05-29
Added an extra parameter for 'exceptionStrings' in IDX10501

mafurman:
Addresses #1132.

The 'IDX10501' exception message has been modified so that any exceptions that were previously caught during signature validation are now included in it.

There are cases where 'exceptionStrings' is empty, in which case this modification doesn't actually end up providing any extra information for the user. It might make sense to have a separate log message without this parameter for that case.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
@@ -89,7 +89,7 @@ namespace Microsoft.IdentityModel.Tokens
 
         // 10500 - SignatureValidation
         public const string IDX10500 = "IDX10500: Signature validation failed. No security keys were provided to validate the signature.";
-        public const string IDX10501 = "IDX10501: Signature validation failed. Unable to match keys: \nkid: '{0}', \ntoken: '{1}'.";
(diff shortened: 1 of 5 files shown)


src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -89,7 +89,7 @@ internal static class LogMessages
 
         // 10500 - SignatureValidation
         public const string IDX10500 = "IDX10500: Signature validation failed. No security keys were provided to validate the signature.";
-        public const string IDX10501 = "IDX10501: Signature validation failed. Unable to match keys: \nkid: '{0}', \ntoken: '{1}'.";
+        public const string IDX10501 = "IDX10501: Signature validation failed. Unable to match keys: \nkid: '{0}'.\nExceptions caught:\n '{1}'. \ntoken: '{2}'.";
```
brentschmaltz: nit: (was there before) Could we change - Unable to match keys: -> Unable to match key:  (singular).

brentschmaltz:
:shipit: feel free to change the grammar issue.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
mafurman: That's a good idea! Makes the message a lot clearer. 

Just made this fix.

brentschmaltz: thanks! :shipit:

Issue #1129, opened 2019-04-04
Turn on SignatureProvider caching on by default.


brentschmaltz:
Currently this feature is off by default. Users have reported up to 4x improvement in performance and large memory pressure reductions.

brentschmaltz:
fixed with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1198

PR #1198, opened 2019-05-30
Set the CacheSignatureProviders property to true by default

mafurman:
Addresses #1129.

### DIFF
--- a/test/System.IdentityModel.Tokens.Jwt.Tests/CreateAndValidateTokens.cs
+++ b/test/System.IdentityModel.Tokens.Jwt.Tests/CreateAndValidateTokens.cs
@@ -345,7 +345,7 @@ namespace System.IdentityModel.Tokens.Jwt.Tests
             var theoryData = new TheoryData<JwtTheoryData>();
             var handler = new JwtSecurityTokenHandler();
             var asymmetricSecurityTokenDescriptor = Default.AsymmetricSignSecurityTokenDescriptor(null);
-            var cryptorProviderFactory = new CryptoProviderFactory { CacheSignatureProviders = true };
+            var cryptorProviderFactory = new CryptoProviderFactory();
             asymmetricSecurityTokenDescriptor.SigningCredentials.CryptoProviderFactory = cryptorProviderFactory;
             var asymmetricTokenValidationParameters = Default.AsymmetricSignTokenValidationParameters;
             asymmetricTokenValidationParameters.CryptoProviderFactory = cryptorProviderFactory;
--- a/runTests.ps1
+++ b/runTests.ps1
@@ -52,8 +52,8 @@ foreach ($testProject in $testProjects)
         Write-Host ">>> Set-Location $root\test\$name"
         pushd
         Set-Location $root\test\$name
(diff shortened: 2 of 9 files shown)


henrik-me:
will customers run into issues by this change? any risk for regressions of existing customer code? It can sometimes be considered a breaking change (semantically) when changing a default value. It does seem like a good idea, however it should be evaluated vs. regression risks. Also this change of behavior should be documented imo in the reference docs and likely also in the wiki (I suggest starting a wiki page with changes from 5.4 to 5.5. where something like this is mentioned)

test/System.IdentityModel.Tokens.Jwt.Tests/CreateAndValidateTokens.cs
```diff
@@ -345,7 +345,7 @@ public static TheoryData<JwtTheoryData> RoundTripTokensUsingCacheTheoryData()
             var theoryData = new TheoryData<JwtTheoryData>();
             var handler = new JwtSecurityTokenHandler();
             var asymmetricSecurityTokenDescriptor = Default.AsymmetricSignSecurityTokenDescriptor(null);
-            var cryptorProviderFactory = new CryptoProviderFactory { CacheSignatureProviders = true };
+            var cryptorProviderFactory = new CryptoProviderFactory();
```
brentschmaltz: Can we add a couple of tests that turn caching off and make sure new signature providers are created and released?

mafurman: Good idea. I added a few test cases for this, but let me know if we need more comprehensive tests.

kmcclellan:
This change introduced a bug for my team when I upgraded. I feel that caching should remain opt-in in cases where we can expect disposable resources to be involved (on an `AsymmetricSecurityKey`?). Consider this code:

```c#
private void GenerateSignature()
{
    using (var algorithm = ECDsa.Create(ECCurve.NamedCurves.nistP256))
    {
        var signingKey = new ECDsaSecurityKey(algorithm)
        {
            KeyId = "TEST"
        };

        // This is the key I want to use (not expecting caching)
        Console.WriteLine($"Key: {signingKey.GetHashCode()}");

        var provider = signingKey.CryptoProviderFactory.CreateForSigning(signingKey, SecurityAlgorithms.EcdsaSha256);

        // This key may be for a disposed algorithm!!
        Console.WriteLine($"Provider key: {provider.Key.GetHashCode()}");

        // Throws an exception on subsequent calls
        provider.Sign(new byte[8]);
    }
}
```

Exception:
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Security.Cryptography.ECDsaImplementation.ECDsaOpenSsl.SignHash(Byte[] hash)
   at Microsoft.IdentityModel.Tokens.AsymmetricAdapter.SignWithECDsa(Byte[] bytes)
   at Microsoft.IdentityModel.Tokens.AsymmetricSignatureProvider.Sign(Byte[] input)
```

This only happens on certain platforms (those using `ECDsaOpenSsl` - e.g. Linux), but the safety of caching disposed resources should not be trusted in general. It can be resolved by manually passing a `CryptoProviderFactory` to the `ECDsaSecurityKey` with caching disabled.

GeoK:
@kmcclellan - I'm sorry about a bug caused by this change. Could you please create a new issue so we can track it appropriately?

PR #1206, opened 2019-06-20
JsonWebKey and Related fixes.

brentschmaltz:
Moved converting JsonWebKey to a usable SecurityKey to single location
Convert JsonWebKey before checking key details such as KeySize
Maintain a singleton ECDAdapter
When creating cacheKey use internalId if keyid is null or empty
Use default CryptoProviderFactory on SecurityKey rather than create new one

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -136,6 +136,11 @@ namespace Microsoft.IdentityModel.Tokens
             _cryptoProviderFactory = key.CryptoProviderFactory;
             _minimumAsymmetricKeySizeInBitsForSigningMap = new Dictionary<string, int>(DefaultMinimumAsymmetricKeySizeInBitsForSigningMap);
             _minimumAsymmetricKeySizeInBitsForVerifyingMap = new Dictionary<string, int>(DefaultMinimumAsymmetricKeySizeInBitsForVerifyingMap);
+
+            var jsonWebKey = key as JsonWebKey;
+            if (jsonWebKey != null)
+                JsonWebKeyConverter.TryConvertToSecurityKey(jsonWebKey, out SecurityKey _);
+
             if (willCreateSignatures && FoundPrivateKey(key) == PrivateKeyStatus.DoesNotExist)
                 throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10638, key)));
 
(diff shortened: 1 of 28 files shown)


src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs
```diff
@@ -33,7 +33,7 @@ namespace Microsoft.IdentityModel.Tokens
 {
     /// <summary>
     /// Converts a <see cref="SecurityKey"/> into a <see cref="JsonWebKey"/>
-    /// Supports: <see cref="RsaSecurityKey"/>, <see cref="X509SecurityKey"/> and <see cref=" SymmetricSecurityKey"/>.
+    /// Supports: converting to a <see cref="JsonWebKey"/> from one of: <see cref="RsaSecurityKey"/>, <see cref="X509SecurityKey"/>, and <see cref=" SymmetricSecurityKey"/>.
```
GeoK: It looks like we are missing a method that coverts ECDsaSecurityKey into a JsonWebKey.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -136,14 +136,19 @@ public AsymmetricSignatureProvider(SecurityKey key, string algorithm, bool willC
             _cryptoProviderFactory = key.CryptoProviderFactory;
             _minimumAsymmetricKeySizeInBitsForSigningMap = new Dictionary<string, int>(DefaultMinimumAsymmetricKeySizeInBitsForSigningMap);
             _minimumAsymmetricKeySizeInBitsForVerifyingMap = new Dictionary<string, int>(DefaultMinimumAsymmetricKeySizeInBitsForVerifyingMap);
+
+            var jsonWebKey = key as JsonWebKey;
```
GeoK: Perhaps it's a bit cleaner to move this code to ResolveAsymmetricAdapter method.

src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs
brentschmaltz: Yep, we don't have that yet. vNext

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
GeoK: Actually, I'm not sure that we really need this code.
If a key is JsonWebKey, it will be converted by AsymmetricAdapter.

GeoK: Nevermind, we need the converted key down the stack when key size validation is done.

src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs
GeoK: Issue: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1208

PR #1216, opened 2019-07-08
JsonWebTokenHandler.CreateToken() now adds iat, nbf, and exp as longs not strings

### LINKED ISSUE #1207: JsonWebTokenHandler.CreateToken adds iat, nbf and exp as strings should be int
see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L154 we should not be using ToString();

mafurman:
Fixes #1207.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -151,7 +151,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 if (payload.ContainsKey(JwtRegisteredClaimNames.Exp))
                     LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, nameof(tokenDescriptor.Expires)));
 
-                payload[JwtRegisteredClaimNames.Exp] = EpochTime.GetIntDate(tokenDescriptor.Expires.Value).ToString();
+                payload[JwtRegisteredClaimNames.Exp] = EpochTime.GetIntDate(tokenDescriptor.Expires.Value);
             }
 
             if (tokenDescriptor.Issuer != null)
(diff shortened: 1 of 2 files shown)


henrik-me:
I assume this should not go into v5? This is only relevant for v6+ right?

henrik-me:
Seems relevant to figure out who will be broken from this change. Thoughts on how to prevent the impact for customers depending on the current behavior?

mafurman:
@henrik-me The original GitHub work item states that the issue is for dev5x, but I do think that you're right. I can reissue this pull request so that it merges into the dev branch.

This actually might not be a breaking change. 

The JsonWebToken class has properties on it that allow users to retrieve the iat, nbf, and exp claims as DateTime objects (and this behavior remains unchanged). 

Even if users were using the `GetPayloadValue<T>/TryGetPayloadValue<T>() `methods directly and expecting a string, the way that the methods are written as of now allow users to retrieve numeric values (and other types) as strings.

henrik-me:
thanks for pointing out the details. that's good. Still believe this should be in v6 and not (necessarily v5), considering where the ask for this change comes from.

---
In reply to: [509425717](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1216#issuecomment-509425717) [](ancestors = 509425717)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -252,13 +252,13 @@ private string CreateTokenPrivate(JObject payload, SigningCredentials signingCre
             {
                 var now = DateTime.UtcNow;
```
brentschmaltz: We could be slightly better performance by calculating long now = EpochTime.GetIntDate(DateTime.UtcNow).
We could also add a method EpochTime.UtcNow that is tuned for perf.

mafurman: That's a good point! I changed this so that EpochTime.GetIntDate(DateTime.UtcNow) is only called once.

What do you mean by 'tuned for perf'?

mafurman:
The PR is now set to merge into the 'dev' branch rather than 'dev5x.'

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: :shipit:

brentschmaltz: @mafurman have a look at: brentsch/epochtime for an idea of how we can improve perf.

Issue #1155, opened 2019-05-01
Add support for creating Saml and Saml2 tokens with TokenDescriptor.Claims


henrik-me:
We introduced SecurityTokenDescriptor.Claims as a property for SecurityTokenHandlers to create tokens.
Only JsonWebTokens is supported with this feature.
We should add support for Saml and Saml2 tokens.

**Acceptance Criteria**: Tests shows that tokens created with the same subject are the same as Claims

(Moved from [Azure DevOps](https://identitydivision.visualstudio.com/DevEx/_workitems/edit/601674))

PR #1223, opened 2019-07-18
Added methods in JsonWebTokenHandler that allow custom JWT header claims to be added

### LINKED ISSUE #1210: Allow creation of a customized header using JsonWebTokenHandler
Some customers have expressed the need for a facility that allows setting custom attributes in the header part, during token creation.

mafurman:
Fixes #1210 and #1237.


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -530,6 +701,9 @@ private string EncryptTokenPrivate(string innerJwt, EncryptingCredentials encryp
 
                 header.Add(JwtHeaderParameterNames.Typ, JwtConstants.HeaderType);
 
+                if (additionalHeaderClaims != null)
```
GeoK: What happens if the algorithm is not "dir" (the _else_ part)?
It looks like we won't add additional header claims in that case.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -227,12 +289,56 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
+            if (additionalHeaderClaims != null && (additionalHeaderClaims.ContainsKey(JwtHeaderParameterNames.Alg)
+                || additionalHeaderClaims.ContainsKey(JwtHeaderParameterNames.Kid)
+                || additionalHeaderClaims.ContainsKey(JwtHeaderParameterNames.Typ)
+                || additionalHeaderClaims.ContainsKey(JwtHeaderParameterNames.X5t)
+                || additionalHeaderClaims.ContainsKey(JwtHeaderParameterNames.Enc)
+                || additionalHeaderClaims.ContainsKey(JwtHeaderParameterNames.Zip)))
+                throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX14116, nameof(additionalHeaderClaims))));
+
+            // if 'encryptingCredentials' != null it means additionalClaims should be added to the outer JWT header.
+            if (encryptingCredentials == null && additionalHeaderClaims != null || !JsonWebTokenManager.KeyToHeaderCache.TryGetValue(JsonWebTokenManager.GetHeaderCacheKey(signingCredentials), out string rawHeader))
```
GeoK: I don't think that we should ever cache headers when _additionalHeaderClaims_ is not null.
A header cache key is formed out of a security key type, key id, and an algorithm.

Image that the first call has a key=="1" and alg == "RS256" and some additional header claims.
As header cache is empty at that point we will cache the header, together with the additional claims.

If we have the second call with the same key id and alg ("1", "RS256") but without additional header claims, we will actually use the header cached in the first call (with additional claims).

GeoK (changes requested):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Good catch! I somehow missed the fact that we're creating the JWE header in two different places depending on the algorithm used. 

I fixed this and also added in some tests for both the direct encryption and key wrapping scenarios.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: You're right; this was an oversight on my part. Fixed this so that we never cache a header that had additional claims added to it.

Also added in tests for this.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -469,7 +580,35 @@ public string EncryptToken(string innerJwt, EncryptingCredentials encryptingCred
...
                 throw LogHelper.LogArgumentNullException(nameof(encryptingCredentials));
 
-            return EncryptTokenPrivate(innerJwt, encryptingCredentials, null);
+            return EncryptTokenPrivate(innerJwt, encryptingCredentials, null, null);
+        }
+
+        /// <summary>
+        /// Encrypts a JWS.
+        /// </summary>
+        /// <param name="innerJwt">A 'JSON Web Token' (JWT) in JWS Compact Serialization Format.</param>
```
brentschmaltz: should we modify this parameter name and comment, we don't really know or verify that the string is a JWS in Compact Serialization Format.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -469,7 +580,35 @@ public string EncryptToken(string innerJwt, EncryptingCredentials encryptingCred
...
+        /// <exception cref="SecurityTokenEncryptionFailedException">if the CryptoProviderFactory being used does not support the <see cref="EncryptingCredentials.Enc"/> (algorithm), <see cref="EncryptingCredentials.Key"/> pair.</exception>
+        /// <exception cref="SecurityTokenEncryptionFailedException">if unable to create a token encryption provider for the <see cref="EncryptingCredentials.Enc"/> (algorithm), <see cref="EncryptingCredentials.Key"/> pair.</exception>
+        /// <exception cref="SecurityTokenEncryptionFailedException">if encryption fails using the <see cref="EncryptingCredentials.Enc"/> (algorithm), <see cref="EncryptingCredentials.Key"/> pair.</exception>
+        /// <exception cref="SecurityTokenEncryptionFailedException">if not using one of the supported content encryption key (CEK) algorithms: 128, 384 or 512 AesCbcHmac (this applies in the case of key wrap only, not direct encryption).</exception>
+        public string EncryptToken(string innerJwt, EncryptingCredentials encryptingCredentials, IDictionary<string, object> additionalHeaderClaims)
```
brentschmaltz: might want to think about placing this method after the one without additionalHeaderClaims

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
```diff
@@ -76,6 +76,15 @@ public class SecurityTokenDescriptor
...
         public IDictionary<string, object> Claims { get; set; }
 
+        /// <summary>
+        /// Gets or sets the <see cref="Dictionary{TKey, TValue}"/> which contains any custom header claims that need to be added to the JWT token header.
+        /// The 'alg', 'kid', 'typ', 'x5t', 'enc', and 'zip' claims are added by default based on the <see cref="SigningCredentials"/>,
+        /// <see cref="EncryptingCredentials"/>, and/or <see cref="CompressionAlgorithm"/> provided and SHOULD NOT be included in this dictionary as this
+        /// will result in an exception being thrown. 
+        /// <remarks> These claims are only added to the outer header (in case of a JWE).</remarks>
+        /// </summary>
+        public IDictionary<string, object> AdditionalHeaderClaims { get; set; }
```
brentschmaltz: This property is only for JWT's. What will SAML handlers do with this?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: I agree that the parameter name and comment should be changed, or perhaps maybe we should verify that the string is in JWS format?

In the [RFC](https://tools.ietf.org/html/rfc7516#appendix-A.1.5) the value to be encrypted is referred to as 'plaintext', so we could call it that. 

However, if something other than a JWS is passed into this method, the result produced by this method will no longer be something our library can verify. I feel like this may cause confusion.

The name of the method also happens to be 'EncryptToken', which may also get confusing.

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
mafurman: I was thinking this would be a property that's only used when creating JWT tokens. I could make the comment clearer with regards to this if you think that'll help.

On a related note, `IDictionary<string, object> Claims` is currently not being used by the SAML handlers either.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
GeoK: This condition looks too complicated. Perhaps we should refactor it.

There might be one more issue here. Let's say that we are creating a JWE token with some additional header claims.
The expected behavior is to add these additional claims only to the outer token header.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/0f506ac2ef4e7a2a26ddec4369ae4cbe9dfeebcd/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L243

`encryptingCredentials` wouldn't be null, `additionalHeaderClaims` wouldn't be null and let's suppose that the header cache is empty => (false && true) || true => true, so this condition will execute.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/0f506ac2ef4e7a2a26ddec4369ae4cbe9dfeebcd/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L341

Additional header claims will be added to the inner token header too, which is not in line with the expected behavior. 
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/0f506ac2ef4e7a2a26ddec4369ae4cbe9dfeebcd/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L353-L354

GeoK (changes requested):

leastprivilege:
I haven't looked at the API yet - but per our recent discussion, please make sure it is easily possible to set the `typ` header value.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -227,12 +289,56 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
         }
 
-        private string CreateTokenPrivate(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, string algorithm)
+        private string CreateTokenPrivate(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, string algorithm, IDictionary<string, object> additionalHeaderClaims)
         {
-            if (!JsonWebTokenManager.KeyToHeaderCache.TryGetValue(JsonWebTokenManager.GetHeaderCacheKey(signingCredentials), out string rawHeader))
+
+            if (additionalHeaderClaims != null && (additionalHeaderClaims.ContainsKey(JwtHeaderParameterNames.Alg)
+                || additionalHeaderClaims.ContainsKey(JwtHeaderParameterNames.Kid)
+                || additionalHeaderClaims.ContainsKey(JwtHeaderParameterNames.Typ)
```
GeoK: Let's relax this condition by removing "typ" and allow users to set "typ" via additional header claims.
If there is not "typ" claim within _additionalHeaderClaims_, we should use the default value ("JWT").

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Thanks for noticing this! I fixed this bug and added in tests to make sure the additional header claims aren't added to both the inner and the outer JWT headers.

I tried to refactor the logic to make it clearer (and added in some comments), but if you have any other suggestions for improvement let me know! I rebased against the dev branch and the combination of allowing unsigned JWTs and additional header claims has made this method somewhat confusing in my opinion.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: I moved these restricted header claim names into a list for ease of readability and removed 'typ' from it.

GeoK: Perhaps we should use `StringComparer.OrdinalIgnoreCase` here.

mafurman: Good catch! Just made this fix and added some tests. :)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -303,6 +413,34 @@ private string CreateTokenPrivate(JObject payload, SigningCredentials signingCre
 
                 rawHeader = Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(header.ToString(Formatting.None)));
                 JsonWebTokenManager.KeyToHeaderCache.TryAdd(JsonWebTokenManager.GetHeaderCacheKey(signingCredentials), rawHeader);
+            } // Otherwise, if there is no outer JWT header, add additional header claims to this header.
+            else if (additionalHeaderClaims != null)
+            {
+                JObject header;
+                if (signingCredentials != null)
+                {
+                    header = new JObject
```
GeoK: nit: The process of creating a default header is repeated twice in this method. Perhaps we should add an internal utility method that creates a default header using singing credentials.

mafurman: Good suggestion! :) I added utility methods for creating default JWE and JWS headers.

GeoK (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -109,6 +118,41 @@ public virtual bool CanValidateToken
             get { return true; }
         }
 
+        private JObject CreateDefaultJWEHeader(EncryptingCredentials encryptingCredentials, string compressionAlgorithm)
+        {
+            var header = new JObject();
+
+            if (!string.IsNullOrEmpty(encryptingCredentials.Alg))
```
brentschmaltz: Alg, Enc can not be null.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b16a4d5c90bef3a20b35282ef4dc1fb1b4d8cd49/src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs#L100

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b16a4d5c90bef3a20b35282ef4dc1fb1b4d8cd49/src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs#L109

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -109,6 +118,41 @@ public virtual bool CanValidateToken
...
+
+            return header;
+        }
+
+        private JObject CreateDefaultJWSHeader(SigningCredentials signingCredentials)
+        {
+            var header = new JObject
+                {
+                    { JwtHeaderParameterNames.Alg, signingCredentials.Algorithm },
+                    { JwtHeaderParameterNames.Kid, signingCredentials.Key.KeyId },
```
brentschmaltz: KeyId can be null.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -199,7 +269,7 @@ public virtual string CreateToken(SecurityTokenDescriptor tokenDescriptor)
             if (!payload.Any())
```
brentschmaltz: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1237 was opened as an empty payload should be allowed.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -283,26 +389,72 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
+        /// <param name="compressionAlgorithm">Defines the compression algorithm that will be used to compress the JWT token payload.</param>       
+        /// <param name="additionalHeaderClaims">Defines the dictionary containing any custom header claims that need to be added to the outer JWT token header.</param>
+        /// <exception cref="ArgumentNullException">if <paramref name="payload"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="signingCredentials"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="encryptingCredentials"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="compressionAlgorithm"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="additionalHeaderClaims"/> is null.</exception>
+        /// <returns>A JWE in compact serialization format.</returns>
+        public virtual string CreateToken(string payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, string compressionAlgorithm, IDictionary<string, object> additionalHeaderClaims)
```
brentschmaltz: need to add a comment to capture the SecurityTokenException thrown if any default header claims are found.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -137,7 +183,31 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
+        /// Creates a JWS (Json Web Signature).
+        /// </summary>
+        /// <param name="payload">A string containing JSON which represents the JWT token payload.</param>
+        /// <param name="signingCredentials">Defines the security key and algorithm that will be used to sign the JWS.</param>
+        /// <param name="additionalHeaderClaims">Defines the dictionary containing any custom header claims that need to be added to the JWT token header.</param>
+        /// <exception cref="ArgumentNullException">if <paramref name="payload"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="signingCredentials"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="additionalHeaderClaims"/> is null.</exception>
+        /// <returns>A JWS in Compact Serialization Format.</returns>
+        public virtual string CreateToken(string payload, SigningCredentials signingCredentials, IDictionary<string, object> additionalHeaderClaims)
```
brentschmaltz: need to add a comment to capture the SecurityTokenException thrown if any default header claims are found.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -283,26 +389,72 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
+        {
+
+            if (additionalHeaderClaims != null && additionalHeaderClaims.Keys.Intersect(_defaultHeaderParameters, StringComparer.OrdinalIgnoreCase).Any())
+                throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX14116, nameof(additionalHeaderClaims), string.Join(", ", _defaultHeaderParameters))));
+
             var rawHeader = Base64UrlEncodedUnsignedJWSHeader;
-            if (signingCredentials != null && !JsonWebTokenManager.KeyToHeaderCache.TryGetValue(JsonWebTokenManager.GetHeaderCacheKey(signingCredentials), out rawHeader))
+            // If there's no additional header claims to be added to the header and the token will be signed, try to retrieve a header value from the cache.
+            if (additionalHeaderClaims == null && signingCredentials != null && !JsonWebTokenManager.KeyToHeaderCache.TryGetValue(JsonWebTokenManager.GetHeaderCacheKey(signingCredentials), out rawHeader))
```
brentschmaltz: (nit) could add additionalHeaderClaims.Count > 0

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Good point! Fixed.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Fixed.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Fixed. Also added in some tests.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Good catch! I fixed this in all the relevant methods.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Fixed!

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Do you mean that we should check for this before seeing if 'additionalHeaderClaims' intersects _defaultHeaderParameters? Or that we should check for this so that we still use the header cache if 'additionalHeaderClaims' is empty?

I added in a check in both places.

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
brentschmaltz: this captures the issue: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1155

brentschmaltz (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -283,26 +391,75 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
+
+            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, encryptingCredentials, compressionAlgorithm, additionalHeaderClaims);
         }
 
-        private string CreateTokenPrivate(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, string compressionAlgorithm)
+        private string CreateTokenPrivate(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, string compressionAlgorithm, IDictionary<string, object> additionalHeaderClaims)
         {
+
+            if (additionalHeaderClaims != null && additionalHeaderClaims.Count > 0 && additionalHeaderClaims.Keys.Intersect(_defaultHeaderParameters, StringComparer.OrdinalIgnoreCase).Any())
+                throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX14116, nameof(additionalHeaderClaims), string.Join(", ", _defaultHeaderParameters))));
```
GeoK: 👍

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -283,26 +391,75 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
-                {
-                    { JwtHeaderParameterNames.Alg, signingCredentials.Algorithm },
-                    { JwtHeaderParameterNames.Kid, signingCredentials.Key.KeyId },
-                    { JwtHeaderParameterNames.Typ, JwtConstants.HeaderType }
-                };
+                var header = CreateDefaultJWSHeader(signingCredentials);
+                rawHeader = Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(header.ToString(Formatting.None)));
+                JsonWebTokenManager.KeyToHeaderCache.TryAdd(JsonWebTokenManager.GetHeaderCacheKey(signingCredentials), rawHeader);
+            } // Otherwise, if there is no outer JWT header, add additional header claims to this header.
+            else if (additionalHeaderClaims != null)
```
GeoK: nit: We should probably add `&& additionalHeaderClaims.Count > 0` to this condition too.
If additionalHeaderClaims are not null (but the count is 0), we will do a few unnecessary operations when a header is found in the cache. This is a pretty narrow use-case, though.

mafurman: Good catch! :) Fixed.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: 😃

GeoK (approved):

steveoshima:
FYI, this currently isnt working with RSA-PSS support.

GeoK:
Hi @steveoshima - It's not clear what exactly is not working. Can you provide more information or/and a small repro project?

GeoK:
@steveoshima - Are you still hitting the issue when using RSA-PSS?
We are planning on releasing 5.6.0 soon and we would appreciate your input.

steveoshima:
> @steveoshima - Are you still hitting the issue when using RSA-PSS?
> We are planning on releasing 5.6.0 soon and we would appreciate your input.

I last check different branches to get RSA-PSS with custom headers working but no branches worked that i tried. I spotted the cusom header test with RSA-PSS was incomplete. I messaged about this on the issue. See my comment here - https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1210

Issue #1222, opened 2019-07-18
For NETSTANDARD2_0 target X509SecurityKey should use RSACertificateExtensions


brentschmaltz:
Obtaining public / private keys should use the extension that is available in 2.0.

brentschmaltz:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1225

Issue #1221, opened 2019-07-18
AsymmetricAdapter should use JsonWebKeyConverter.TryConvertToSecurityKey transform keys.


brentschmaltz:
TryConvertToSecurityKey provides consistent logic when transforming JsonWebKey to a Key that is usable by the runtime. In if the internal key is set TryConvertToSecurityKey will recognize this and the transform will not be performed.

brentschmaltz:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1225

PR #1225, opened 2019-07-19
Use JsonWebKeyConvertor consistently

brentschmaltz:
We want to use JsonWebKeyConveter when converting from a JsonWebKey.
This provides some consistency.
We also noticed that there was some cleanup possible in the JsonWebKeyConverter tests.
We also included #ifdef for Netstandard2.0 for X509SecurityKey when obtaining private and public keys

This address issues: 
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1221
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1222

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
@@ -82,52 +82,9 @@ namespace Microsoft.IdentityModel.Tokens
 #if NET461 || NETSTANDARD1_4 || NETSTANDARD2_0
         // HasAlgorithmName was introduced into Net46
         internal AsymmetricAdapter(SecurityKey key, string algorithm, HashAlgorithm hashAlgorithm, HashAlgorithmName hashAlgorithmName, bool requirePrivateKey)
+            : this(key, algorithm, hashAlgorithm, requirePrivateKey)
         {
-            HashAlgorithm = hashAlgorithm;
             HashAlgorithmName = hashAlgorithmName;
-
-            if (key is RsaSecurityKey rsaKey)
-            {
-                if (rsaKey.Rsa != null)
-                    Initialize(rsaKey.Rsa, algorithm);
-                else
-                {
-                    var rsa = RSA.Create();
-                    rsa.ImportParameters(rsaKey.Parameters);
-                    Initialize(rsa, algorithm);
-                    _disposeCryptoOperators = true;
-                }
-            }
-            else if (key is X509SecurityKey x509Key)
-            {
-                if (requirePrivateKey)
-                    Initialize(x509Key.Certificate.GetRSAPrivateKey(), algorithm);
-                else
-                    Initialize(x509Key.Certificate.GetRSAPublicKey(), algorithm);
-            }
-            else if (key is JsonWebKey rsaWebKey && rsaWebKey.Kty == JsonWebAlgorithmsKeyTypes.RSA)
-            {
-                var rsa = RSA.Create();
-                rsa.ImportParameters(rsaWebKey.CreateRsaParameters());
-                Initialize(rsa, algorithm);
-                _disposeCryptoOperators = true;
-            }
-            else if (key is JsonWebKey ecdsaWebKey && ecdsaWebKey.Kty == JsonWebAlgorithmsKeyTypes.EllipticCurve)
-            {
-                var ecdsaAdapter = ECDsaAdapter.Instance;
-                ECDsa = ecdsaAdapter.CreateECDsa(ecdsaWebKey, requirePrivateKey);
-                SignatureFunction = SignWithECDsa;
-                VerifyFunction = VerifyWithECDsa;
-                _disposeCryptoOperators = true;
-            }
-            else if (key is ECDsaSecurityKey ecdKey)
-            {
-                ECDsa = ecdKey.ECDsa;
-                SignatureFunction = SignWithECDsa;
-                VerifyFunction = VerifyWithECDsa;
-            }
-            else
-                throw LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant(LogMessages.IDX10684, algorithm, key)));
         }
 #endif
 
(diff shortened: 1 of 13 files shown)


src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -146,64 +103,65 @@ internal AsymmetricAdapter(SecurityKey key, string algorithm, HashAlgorithm hash
...
-                    Initialize(x509Key.Certificate.GetRSAPrivateKey(), algorithm);
-                else
-                    Initialize(x509Key.Certificate.GetRSAPublicKey(), algorithm);
-#endif
+                InitializeUsingX509SecurityKey(x509Key, algorithm, requirePrivateKey);
             }
             else if (key is JsonWebKey jsonWebKey)
             {
-                if (jsonWebKey.Kty == JsonWebAlgorithmsKeyTypes.RSA)
+                if (JsonWebKeyConverter.TryConvertToSecurityKey(jsonWebKey, out SecurityKey securityKey))
```
GeoK: If a jsonWebKey can't be converted to a security key, Signature/Verify functions won't be set.
When a user calls signatureProvider.Sign() we would throw NotSupportedException with "Unable to Sign, Internal SignFunction is not available."

I think that we should be more explicit and throw an exception here if we see that we can't covert a jsonWebKey to a security key.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -335,11 +335,17 @@ public virtual void ValidateAsymmetricSecurityKeySize(SecurityKey key, string al
             if (string.IsNullOrEmpty(algorithm))
                 throw LogHelper.LogArgumentNullException(nameof(algorithm));
 
-            int keySize;
-            if (key is JsonWebKey jsonWebKey && jsonWebKey.ConvertedSecurityKey is AsymmetricSecurityKey)
-                keySize = (jsonWebKey.ConvertedSecurityKey as AsymmetricSecurityKey).KeySize;
-            else
-                keySize = key.KeySize;
+            int keySize = key.KeySize;
+            if (key is AsymmetricSecurityKey securityKey)
```
GeoK: I'm not sure why this if-condition is needed.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -146,64 +103,65 @@ internal AsymmetricAdapter(SecurityKey key, string algorithm, HashAlgorithm hash
...
+        {
+            if (requirePrivateKey)
+                InitializeUsingRsa(x509SecurityKey.PrivateKey as RSA, algorithm);
+            else
+                InitializeUsingRsa(x509SecurityKey.PublicKey as RSA, algorithm);
+        }
+
+        private void InitializeUsingEcdsaSecurityKey(ECDsaSecurityKey ecdsaSecurityKey)
+        {
+            ECDsaSecurityKey = ecdsaSecurityKey;
```
GeoK: Do we need to set  `_disposeCryptoOperators = true` here?
And possibly here:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b553f2db7433ec87fd925147aceb0b83948920f8/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs#L286-L291

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
brentschmaltz: I had originally changed to follow that logic, but choose a different exception as it made more sense.
ill try again and be more careful with the exception and see where that goes.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
brentschmaltz: I want to convert from a JsonWebKey before checking the KeySize. Sometimes a JsonWebKey can have 0 as the key size.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
brentschmaltz: I'll check

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
GeoK: What I was thinking about is to remove this if-condition and change the line 343 from
`else if (key is JsonWebKey jsonWebKey)` to `if (key is JsonWebKey jsonWebKey)`.
Not sure what is the value of casting a Security Key into AsymmetricSecurityKey and getting its key size if we are doing that already in line 338.

brentschmaltz: @GeoK This is a public method, perhaps we should throw if the key is not Asymmetric.
Perhaps logic of
if key is Asymmetric use keysize
else try to convert to asymmetric
	throw if no conversion

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
brentschmaltz: @GeoK I don't think so because we are only creating a single 'derived' key each time because JsonWebKey.ConvertedSecurityKey is set once.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -146,64 +103,65 @@ internal AsymmetricAdapter(SecurityKey key, string algorithm, HashAlgorithm hash
...
-                if (rsaKey.Rsa != null)
-                    Initialize(rsaKey.Rsa, algorithm);
-                else
-                {
-                    var rsa = RSA.Create();
-                    rsa.ImportParameters(rsaKey.Parameters);
-                    Initialize(rsa, algorithm);
-                    _disposeCryptoOperators = true;
-                }
+                IntializeUsingRsaSecurityKey(rsaKey, algorithm);
```
mafurman: nit: "Initialize" is misspelled

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -190,12 +190,15 @@ internal static class LogMessages
         //public const string IDX10691 = "IDX10691:"
         public const string IDX10692 = "IDX10692: The RSASS-PSS signature algorithm is not available on .NET 4.5 and .NET 4.5.1 targets. The list of supported algorithms is available here: https://aka.ms/IdentityModel/supported-algorithms";
         public const string IDX10693 = "IDX10693: RSACryptoServiceProvider doesn't support the RSASSA-PSS signature algorithm. The list of supported algorithms is available here: https://aka.ms/IdentityModel/supported-algorithms";
+        public const string IDX10694 = "IDX10694: JsonWebKeyConverter threw attempting to convert JsonWebKey: '{0}'. Exception: '{1}'.";
+
```
mafurman: nit: extra new line here

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
brentschmaltz: fixed

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
brentschmaltz: fixed

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
brentschmaltz: @GeoK 	Opened https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1240 to capture this issue.

PR #1229, opened 2019-07-24
JsonWebToken now retrieves claim values in a case sensitive manner.

mafurman:
Claim names are case sensitive according to the JWT RFC:
https://tools.ietf.org/html/rfc7519#section-10.1.1

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -497,7 +497,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         {
             string issuer = Issuer ?? ClaimsIdentity.DefaultIssuer;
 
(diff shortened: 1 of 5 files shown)


GeoK (commented):
We should also retrieve _iat_, _nbf_, and _exp_ claims in a case sensitive manner.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/1bece30388abfc85218e415747ebd7ceec91aa31/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L316
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/1bece30388abfc85218e415747ebd7ceec91aa31/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs#L163-L165

mafurman:
Closing as we have decided not to add this feature in.

Issue #1086, opened 2018-12-28
Obsolete message seems wrong


GeoK:
This message seems wrong, there is no SecurityTokenDescriptor.Payload.
 https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/7742fd30d8b81a9eca1f31e0f769f090db33fb40/src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs#L87

PR #1235, opened 2019-08-09
JWT tokens can now be created from both SecurityTokenDescriptor.Subject and SecurityTokenDescriptor.Claims

### LINKED ISSUE #1193: Should JsonWebTokenHandler take SecurityTokenDescriptor.Subject into account?
For some reasons, it seems that the new `JsonWebTokenHandler` no longer honors the `Claim` claims stored in `SecurityTokenDescriptor.Subject`. Instead, you have to populate the `Claims` property (an `IDictionary<string, object>`).

Is it the intended behavior?

mafurman:
Fixes #1193.

If both SecurityTokenDescriptor.Subject and SecurityTokenDescriptor.Claims are set, both are used when creating a JWT token. However, if duplicate values are present, the ones found in SecurityTokenDescriptor.Claims will override those found in SecurityTokenDescriptor.Subject.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -150,11 +150,20 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (tokenDescriptor == null)
                 throw LogHelper.LogArgumentNullException(nameof(tokenDescriptor));
 
-            if (tokenDescriptor.Claims == null || !tokenDescriptor.Claims.Any())
-                LogHelper.LogWarning(LogMessages.IDX14114);
+            if ((tokenDescriptor.Subject == null || !tokenDescriptor.Subject.Claims.Any()) 
+                && (tokenDescriptor.Claims == null || !tokenDescriptor.Claims.Any()))
+                LogHelper.LogWarning(LogMessages.IDX14114, nameof(SecurityTokenDescriptor), nameof(SecurityTokenDescriptor.Subject), nameof(SecurityTokenDescriptor.Claims));
+            
+            JObject payload;
+            if (tokenDescriptor.Subject != null)
+                payload = JObject.FromObject(JwtTokenUtilities.CreateDictionaryFromClaims(tokenDescriptor.Subject.Claims));
+            else
+                payload = new JObject();
 
-            // JObject needs to be upcast to an IDictionary<string, JToken> so that we can access the ContainsKey() and Any() methods
-            IDictionary<string, JToken> payload = tokenDescriptor.Claims == null ? new JObject() : JObject.FromObject(tokenDescriptor.Claims);
+            // If a key is present in both tokenDescriptor.Subject.Claims and tokenDescriptor.Claims, the value present in tokenDescriptor.Claims is the
+            // one that takes precedence and will remain after the merge. Key comparison is case sensitive. 
+            if (tokenDescriptor.Claims != null && tokenDescriptor.Claims.Count > 0)
+                payload.Merge(JObject.FromObject(tokenDescriptor.Claims), new JsonMergeSettings { MergeArrayHandling = MergeArrayHandling.Replace });
 
             if (tokenDescriptor.Audience != null)
             {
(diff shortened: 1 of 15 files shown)


GeoK:
Also fixes #1086

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -150,11 +150,21 @@ public virtual string CreateToken(SecurityTokenDescriptor tokenDescriptor)
...
+                payload = JObject.FromObject(JwtTokenUtilities.CreateDictionaryFromClaims(tokenDescriptor.Subject.Claims));
+            else
+                payload = new JObject();
 
-            // JObject needs to be upcast to an IDictionary<string, JToken> so that we can access the ContainsKey() and Any() methods
-            IDictionary<string, JToken> payload = tokenDescriptor.Claims == null ? new JObject() : JObject.FromObject(tokenDescriptor.Claims);
+            // If a key is present in both tokenDescriptor.Subject.Claims and tokenDescriptor.Claims, but the key is not the same case in both (e.g. "key" and "KEY"),
+            // the key present in tokenDescriptor.Subject.Claims is the one that will remain after the merge. However, the corresponding value will be overriden with the value
+            // that was present in tokenDescriptor.Claims.
+            if (tokenDescriptor.Claims != null)
```
brentschmaltz: nit: add logic ( tokenDescriptor.Claims != null && count is > 0 )

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -54,6 +55,45 @@ public class JwtTokenUtilities
         /// </summary>
         public static Regex RegexJwe = new Regex(JwtConstants.JweCompactSerializationRegex, RegexOptions.Compiled | RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(100));
 
+        internal static Dictionary<string, object> CreateDictionaryFromClaims(IEnumerable<Claim> claims)
+        {
+            if (claims == null)
```
brentschmaltz: since this is an internal method, how about in both case (claims == null) && foreach loop never is entered ( there are no claims ), just return empty dictionary.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -151,6 +191,49 @@ public static IEnumerable<SecurityKey> GetAllDecryptionKeys(TokenValidationParam
             return decryptionKeys;
 
         }
+
+        internal static object GetClaimValueUsingValueType(Claim claim)
+        {
+            if (claim.ValueType == ClaimValueTypes.Boolean)
+            {
+                bool boolValue;
```
brentschmaltz: use the pattern in ParseTimeValue, if (double.TryParse(claim.Value, out int intValue))

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -778,7 +795,66 @@ public static TheoryData<CreateTokenTheoryData> CreateJWSUsingSecurityTokenDescr
                             ValidIssuer = "Issuer"
                         }
                     },
-
+                    new CreateTokenTheoryData // Test checks that values in SecurityTokenDescriptor.Subject.Claims
```
brentschmaltz: can we start comments on new line?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Done!

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Made this change.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Fixed.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
mafurman: Fixed.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -150,11 +150,21 @@ public virtual string CreateToken(SecurityTokenDescriptor tokenDescriptor)
...
+            
+            JObject payload;
+            if (tokenDescriptor.Subject != null)
+                payload = JObject.FromObject(JwtTokenUtilities.CreateDictionaryFromClaims(tokenDescriptor.Subject.Claims));
+            else
+                payload = new JObject();
 
-            // JObject needs to be upcast to an IDictionary<string, JToken> so that we can access the ContainsKey() and Any() methods
-            IDictionary<string, JToken> payload = tokenDescriptor.Claims == null ? new JObject() : JObject.FromObject(tokenDescriptor.Claims);
+            // If a key is present in both tokenDescriptor.Subject.Claims and tokenDescriptor.Claims, but the key is not the same case in both (e.g. "key" and "KEY"),
```
GeoK: I believe that this comment is no longer true.

mafurman: You're right! I fixed the comment so it now describes the current logic.

PR #1243, opened 2019-08-26
Added the ISecurityTokenValidator interface to JsonWebTokenHandler

### LINKED ISSUE #1227: JsonWebTokenHandler should support ISecurityTokenValidator
This would allow asp.net users to easily swap out System.IdentityModel.Tokens.Jwt.JwtSecurityToken handler and avoid all the claims 'mapping' issues.

In addition, JsonWebTokenHandler has much better perf (10 - 20%).

mafurman:
Fixes #1227.

Also got rid of any mentions to` SecurityTokenEncryptionKeyNotFoundException` from comments as it's not actually thrown anywhere.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -43,7 +43,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// A <see cref="SecurityTokenHandler"/> designed for creating and validating Json Web Tokens. 
     /// See: http://tools.ietf.org/html/rfc7519 and http://www.rfc-editor.org/info/rfc7515.
     /// </summary>
-    public class JsonWebTokenHandler : TokenHandler
+    public class JsonWebTokenHandler : TokenHandler, ISecurityTokenValidator
     {
         private List<string> _defaultHeaderParameters = new List<string>()
         {
@@ -610,7 +610,6 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// <exception cref="ArgumentNullException">if <paramref name="validationParameters"/>  is null.</exception>
         /// <exception cref="SecurityTokenException">if '<paramref name="jwtToken"/> .Enc' is null or empty.</exception>
         /// <exception cref="SecurityTokenDecompressionFailedException">if decompression failed.</exception>
-        /// <exception cref="SecurityTokenEncryptionKeyNotFoundException">if '<paramref name="jwtToken"/> .Kid' is not null AND decryption fails.</exception>
         /// <exception cref="SecurityTokenDecryptionFailedException">if the JWE was not able to be decrypted.</exception>
         public string DecryptToken(JsonWebToken jwtToken, TokenValidationParameters validationParameters)
         {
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -424,7 +424,6 @@ private ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, TokenValidati
         /// <exception cref="ArgumentNullException">if 'validationParameters' is null.</exception>
         /// <exception cref="SecurityTokenException">if 'jwtToken.Enc' is null or empty.</exception>
         /// <exception cref="SecurityTokenDecompressionFailedException">if decompression failed.</exception>
-        /// <exception cref="SecurityTokenEncryptionKeyNotFoundException">if 'jwtToken.Kid' is not null AND decryption fails.</exception>
```
GeoK: We should get rid of SecurityTokenEncryptionKeyNotFoundException mention here as well:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/93cdb72619325df2d86c3a8afb79d22348191d7a/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L1353

mafurman: Good catch. This reference has been removed.

alrz:
could we get rid of exception-based failure contract? relaying on exceptions for token validation is strange (in JwtBearerHandler).

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1113,6 +1112,54 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
+                throw LogHelper.LogArgumentNullException(nameof(token));
+
+            if (validationParameters == null)
+                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
+
+            if (token.Length > MaximumTokenSizeInBytes)
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(TokenLogMessages.IDX10209, token.Length, MaximumTokenSizeInBytes)));
+
+            var tokenValidationResult = ValidateTokenPrivate(token, validationParameters);
+            if (!tokenValidationResult.IsValid)
```
brentschmaltz: We will benefit from introducing a standard method for detecting if the TokenValidationResult is valid.
This will abstract away any new validation conditions that arise.
In the future, we could allow users to hook into this. 

```
internal void ThrowOnTokenValidationError(TokenValidationResult result, string token, CallContext context)
{
	if (result.Exception != null)
    {
    	if (result.Exception is <SomeIdentityModelExceptio>Exception)
        	throw LogHelper.LogExceptionMessage(result.Exception);
        else
            throw LogHelper.LogExceptionMessage(new <SomeIdentityModelException>, LogHelper.FormatInvariant(LogMessages.IDX?????, this, token), result.Exception), context);
    }

    if (!result.IsValid)
    	throw LogHelper.LogExceptionMessage(new <SomeIdentityModelException>(LogHelper.FormatInvariant(LogMessages.IDX???, this, token)), context);
}
```

brentschmaltz:
@mafurman @henrik-me @GeoK we actually shouldn't do this as asp.net customers will be broken if they switch to JsonWebTokenHandler as ValidateToken's out parameter is a JsonWebToken. Internally asp.net expects a JwtSecurityToken.

We will need a new interface / abstract type that new call graphs will use.

PR #1249, opened 2019-09-05
Ordered TokenValidationParameter methods in order

brentschmaltz:
put methods in order
put { get; set; } on single line


brentschmaltz:
merged in with some other changes directly into dev when fixing a build break.

Issue #1232, opened 2019-08-08
Add support for proof-of-possesion (PoP) protocol to Microsoft.IdentityModel


brentschmaltz:
We want to provide some type of proof of possession(POP). 
There are some specs that are in draft stage:
https://tools.ietf.org/html/draft-ietf-ace-cwt-proof-of-possession-06
https://tools.ietf.org/html/draft-fett-oauth-dpop-02
https://tools.ietf.org/html/draft-ietf-oauth-pop-key-distribution-07
https://tools.ietf.org/html/draft-ietf-oauth-signed-http-request-03

The spec for the 'cnf' key is ratified.
https://tools.ietf.org/html/rfc7800

We want to provide support for POP as it mitigates tokens that leak.
The specs aren't there, but we will push forward with the protocol "PoP" based off of the draft - http-request-03 above.
When the specs are ratified, we will support the ratified version.
This is similar to WSTrust2005 which was released before the ratified 1.3 spec.

We will add classes that a http listener and client would call to create and validate the authenticator.

We should add a new assembly: Microsoft.IdentityModel.Protocols.Pop

PR #1260, opened 2019-09-30
Add support for SignedHttpRequest

GeoK:
Proof-of-Possession (PoP) provides a mechanism to bind (sender-constrain) access_tokens at the application layer. PoP provides mitigation as access_token are only accepted inside a signed envelope. An access_token obtained by an attacker can only be used if the attacker has also obtained private the key associated with the access_token.

A SignedHttpRequest provides a mechanism to bind an http request to a (PoP) token by signing specific parts of the request. 

Proof-of-Possession (PoP) represents a concept that can be implemented in many different ways. This implementation of PoP is based on the SignedHttpRequest spec and it's named _SignedHttpRequest_ to make a clear distinction between other existing and future PoP implementations.

See the [SignedHttpRequestDraft](https://tools.ietf.org/html/draft-ietf-oauth-signed-http-request-03)

SignedHttpRequest will be released as another IdentityModel assembly (_Microsoft.IdentityModel.Protocols.SignedHttpRequest_).

---
Other changes:
* Mark JsonWebTokenHandler.DecryptToken as public.
* Add ConvertFromECDsaSecurityKey to JsonWebKeyConverter (netstandard2.0 target)
* Add ConvertFromX509SecurityKey as an RSA key to to JsonWebKeyConverter.
* Add ComputeJwkThumbprint to SecurityKey.
---

Resolves: #1232


src/Microsoft.IdentityModel.Protocols.PoP/SignedHttpRequest/SignedHttpRequestValidationResult.cs
```diff
@@ -0,0 +1,47 @@
...
+//
+
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Protocols.Pop.SignedHttpRequest
+{
+    /// <summary>
+    /// Contains artifacts obtained when a SignedHttpRequest is validated.
+    /// </summary>
+    public class SignedHttpRequestValidationResult : PopTokenValidationResult
```
GeoK: transferred comment:
@brentschmaltz:
> did you consider deriving from TokenValidationResult?

src/Microsoft.IdentityModel.Protocols.PoP/SignedHttpRequest/SignedHttpRequestCreationData.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <summary>
+        /// Initializes a new instance of <see cref="SignedHttpRequestCreationData"/>.
+        /// </summary>
+        /// <remarks>
+        /// <paramref name="accessToken"/> has to contain the 'cnf' claim so that PoP key can be resolved on the validation side.
+        /// https://tools.ietf.org/html/rfc7800#section-3.1
+        /// </remarks> 
+        /// <param name="accessToken">An access token that contains the 'cnf' claim.</param>
+        /// <param name="httpRequestData">A structure that represents an outgoing http request.</param>
+        /// <param name="httpRequestSigningCredentials">Defines the security key and algorithm that will be used to sign the (Signed)HttpRequest.</param>
```
GeoK: transferred comments:
@brentschmaltz: 

> suggest shorting the name to: signingCredentials

src/Microsoft.IdentityModel.Protocols.PoP/SignedHttpRequest/SignedHttpRequestCreationData.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <param name="signedHttpRequestCreationPolicy">A policy for creating the SignedHttpRequest.</param>
+        /// <param name="callContext">An opaque context used to store work when working with authentication artifacts.</param> 
+        public SignedHttpRequestCreationData(string accessToken, HttpRequestData httpRequestData, SigningCredentials httpRequestSigningCredentials, SignedHttpRequestCreationPolicy signedHttpRequestCreationPolicy, CallContext callContext)
```
GeoK: transferred comment:
@brentschmaltz:
> did you consider the name PopTokenDescriptor, this would be in line with SecurityTokenDescriptor.
it has similar semantics.

GeoK: @brentschmaltz  _SecurityTokenDescriptor_ is used only when creating a token. I needed classes to hold data necessary for both signed http request creation and validation.

src/Microsoft.IdentityModel.Protocols.PoP/SignedHttpRequest/SignedHttpRequestValidationResult.cs
GeoK: This can be an option. Initially I wanted to distance the Pop-related code from TokenValidationResult class, as that is one of the classes that might have some breaking changes between 5x and 6x.

src/Microsoft.IdentityModel.Protocols.PoP/SignedHttpRequest/SignedHttpRequestCreationData.cs
GeoK: Renamed to SignedHttpRequestDescriptor.

GeoK:
@brentschmaltz @henrik-me @mafurman 
Re-requesting the PR review.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/LogMessages.cs
```diff
@@ -0,0 +1,74 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+//  Microsoft.IdentityModel.Protocols.Pop
```
brentschmaltz: s/b Microsoft.IdentityModel.Protocols.SignedHttpRequest

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/ConfirmationClaimTypes.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        /// <summary>
+        /// https://tools.ietf.org/html/rfc7800#section-6.1.1
+        /// </summary>
+        public const string Cnf = "cnf";
+
+        /// <summary>
+        /// https://tools.ietf.org/html/rfc7800#section-6.2.2
+        /// </summary>
+        public const string Jwk = "jwk";
```
brentschmaltz: we have some of these defined in JwtHeaderParameterNames.cs
can we use those, then add the ones related to the payload to JwtRegisteredClaim.cs?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Exceptions/SignedHttpRequestCreationException.cs
```diff
@@ -0,0 +1,63 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
```
brentschmaltz: In general for desktop for exceptions to support serialization we need to have the following
```
#if DESKTOPNET45  
	[Serializable]  
#endif
```
Then add method

```
#if DESKTOPNET45
        /// <summary>
        /// Initializes a new instance of the <see cref="SignedHttpRequestCreationException"/> class.
        /// </summary>
        /// <param name="info">the <see cref="SerializationInfo"/> that holds the serialized object data.</param>
        /// <param name="context">The contextual information about the source or destination.</param>
        protected SignedHttpRequestCreationException(SerializationInfo info, StreamingContext context)
            : base(info, context)
        {
        }
#endif
```

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/LogMessages.cs
```diff
@@ -0,0 +1,74 @@
...
+        public const string IDX23004 = "IDX23004: The following query parameters will not be processed as they are repeated: '{0}'.";
+        public const string IDX23005 = "IDX23005: The following headers will not be processed as they are repeated: '{0}'.";
+        public const string IDX23006 = "IDX23006: The address specified '{0}' is not valid as per the HTTPS scheme. Please specify an HTTPS address for security reasons. For testing with an HTTP address, set the RequireHttpsForJkuResourceRetrieval property on SignedHttpRequestValidationParameters to false.";
+        public const string IDX23007 = "IDX23007: HttpRequestUri is an invalid relative URI: '{0}'.";
+        public const string IDX23008 = "IDX23008: Exception caught while creating the '{0}' claim. Inner exception: '{1}'.";
+        public const string IDX23009 = "IDX23009: Signed http request signature validation failed. Exceptions caught: '{0}'.";
+        public const string IDX23010 = "IDX23010: Lifetime validation of the signed http request failed. Current time: '{0}', signed http request is valid until: '{1}'.";
```
brentschmaltz: Is this utc time?
perhaps we should say so.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/LogMessages.cs
```diff
@@ -0,0 +1,74 @@
...
+        public const string IDX23011 = "IDX23011: '{0}' claim validation failed. Expected value: '{1}', value found: '{2}'.";
+        public const string IDX23012 = "IDX23012: '{0}' claim validation failed. Expected values: '{1}' or '{2}', value found: '{3}'.";
+        public const string IDX23013 = "IDX23013: 'at' token validation failed. Inner exception: '{0}'.";
+        public const string IDX23014 = "IDX23014: Unable to resolve a PoP key. The 'cnf' object must have one of the following claims: 'jwk', 'jwe', 'jku', 'kid'. The 'cnf' claim value: '{0}'.";
+        public const string IDX23015 = "IDX23015: A security key resolved from the 'jwk' claim is not an asymmetric key. Resolved key type: '{0}'.";
+        public const string IDX23016 = "IDX23016: Unable to convert the key found in the 'jwk' claim to a security key. JsonWebKey: '{0}'.";
```
brentschmaltz: will this be needed if we fix up the cnf claim?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/LogMessages.cs
```diff
@@ -0,0 +1,74 @@
...
+        public const string IDX23017 = "IDX23017: No decryption keys found. Unable to decrypt a key found in the 'jwe' claim without decryption keys.";
+        public const string IDX23018 = "IDX23018: Unable to decrypt a 'jwe' claim. Decryption keys used: '{0}'. Inner exception: '{1}'.";
+        public const string IDX23019 = "IDX23019: A security key resolved from the 'jwe' claim is not a symmetric key. Resolved key type: '{0}'.";
+        public const string IDX23020 = "IDX23020: Only one PoP key should be resolved using the 'jku' claim. Number of resolved keys: '{0}'.";
```
brentschmaltz: this is used when the number of keys 0 or > 1 or null returned from ResolvePopKeyFromJkuAsync, can we be specific?
0 keys,
> 1 key
null

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,45 @@
...
+
+[assembly: AssemblyInformationalVersion("0.0.1")]
+[assembly: AssemblyFileVersion("0.0.1")]
+[assembly: AssemblyMetadata("Serviceable", "True")]
+[assembly: AssemblyVersion("0.0.1")]
+[assembly: CLSCompliant(true)]
+[assembly: ComVisible(false)]
+
+[assembly: InternalsVisibleTo("Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
+[assembly: InternalsVisibleTo("Microsoft.IdentityModel.S2S, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
brentschmaltz: Why do we need these internals?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestClaimTypes.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <summary>
+        /// https://tools.ietf.org/html/draft-ietf-oauth-signed-http-request-03#section-3
+        /// </summary>
+        public const string B = "b";
+
+        /// <summary>
+        /// The "cnf" claim might be present in SignedHttpRequest payload.
+        ///  https://tools.ietf.org/html/rfc7800#section-6.1.1
+        /// </summary>
+        public const string Cnf = "cnf";
```
brentschmaltz: This is also in ConfirmationClaimTypes.cs

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestClaimTypes.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary>
+        /// Default "nonce" claim.
+        /// </summary>
+        public const string Nonce = "nonce";
```
brentschmaltz: this is also in JwtRegisteredClaimNames.cs

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestCreationParameters.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// <summary>
+        /// Default value for the <see cref="TimeAdjustment"/>.
+        /// </summary>
+        public static readonly TimeSpan DefaultTimeAdjustment = TimeSpan.Zero;
+
+        /// <summary>
+        /// Gets or sets a time adjustment to apply when creating the timestamp ("ts") claim.
+        /// </summary>
+        /// <remarks>Allows for adjusting the local time so it matches a server time.</remarks>
+        public TimeSpan TimeAdjustment { get; set; } = DefaultTimeAdjustment;
```
brentschmaltz: Should we allow this to be negative?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestDescriptor.cs
```diff
@@ -0,0 +1,142 @@
...
+        public CallContext CallContext { get; }
+
+        ///<summary> Gets or sets a "cnf" claim value as a JSON string.</summary>
+        /// <remarks>
+        /// If <see cref="SignedHttpRequestCreationParameters.CreateCnf"/> flag is set to <c>true</c>, <see cref="CnfClaimValue"/> can be used 
+        /// as a "cnf" claim value when creating a SignedHttpRequest payload.
+        /// If <see cref="SignedHttpRequestCreationParameters.CreateCnf"/> flag is set to <c>true</c>, and <see cref="CnfClaimValue"/> is null or empty,
+        /// a "cnf" claim value will be derived from a <see cref="SigningCredentials"/>.<see cref="SecurityKey"/>.
+        /// </remarks>
+        public string CnfClaimValue { get; set; }
```
brentschmaltz: Why do we allow this to be set?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs
```diff
@@ -0,0 +1,229 @@
...
+        /// <summary>
+        /// Default value for the <see cref="SignedHttpRequestLifetime"/>.
+        /// </summary>
+        public static readonly TimeSpan DefaultSignedHttpRequestLifetime = TimeSpan.FromMinutes(5);
+
+        /// <summary>
+        /// Gets or sets a custom HttpClient when obtaining a JWK set using the 'jku' claim.
+        /// </summary>
+        /// <remarks>https://tools.ietf.org/html/rfc7800#section-3.5</remarks> s
+        public HttpClient HttpClientForJkuResourceRetrieval { get; set; }
```
brentschmaltz: Could this be HttpClientFactory for Core?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -0,0 +1,1417 @@
...
+        /// <param name="signedHttpRequestDescriptor">A structure that wraps parameters needed for SignedHttpRequest creation.</param>
+        /// <remarks>
+        /// This method will be executed only if <see cref="SignedHttpRequestCreationParameters.CreateTs"/> is set to <c>true</c>.
+        /// </remarks>    
+        protected virtual void AddTsClaim(Dictionary<string, object> payload, SignedHttpRequestDescriptor signedHttpRequestDescriptor)
+        {
+            if (payload == null)
+                throw LogHelper.LogArgumentNullException(nameof(payload));
+
+            var signedHttpRequestCreationTime = DateTime.UtcNow.Add(signedHttpRequestDescriptor.SignedHttpRequestCreationParameters.TimeAdjustment);
```
brentschmaltz: Can we use EpochTime.cs?

src/Microsoft.IdentityModel.Tokens/CallContext.cs
```diff
@@ -55,6 +55,14 @@ public CallContext(Guid activityId)
             ActivityId = activityId;
         }
 
+        /// <summary>
+        /// Creates a default context that doesn't capture any logs.
+        /// </summary>
+        public static CallContext Default
```
brentschmaltz: I don't think this buys us much.
Same as:
`new CallContext();`

src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// </summary>
+        public IDictionary<string, IEnumerable<string>> Headers { get; set; }
+
+        /// <summary>
+        /// A utility method that appends <paramref name="headers"/> to the <see cref="Headers"/>.
+        /// </summary>
+        /// <param name="headers">A collection of http request headers.</param>
+        public void AppendHeaders(HttpHeaders headers)
+        {
+            if (Headers == null)
```
brentschmaltz: Why do we allow null?

src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http.Headers;
+
+namespace Microsoft.IdentityModel.Protocols
+{
+    /// <summary>
+    /// Structure that represents an incoming or an outgoing http request.
+    /// </summary>
+    public class HttpRequestData
```
brentschmaltz: Can you put add a propertybag here?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/ConfirmationClaimTypes.cs
GeoK: sure, sounds like a good idea.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Exceptions/SignedHttpRequestCreationException.cs
GeoK: I didn't put this code in as I thought that is never used. 
I'm not sure if the DESKTOPNET45 is a correct target i.e. will this code ever be compiled?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/LogMessages.cs
GeoK: Agreed.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/LogMessages.cs
GeoK: Still needed. We might get an incorrect JWK that we can't use to create a SecurityKey.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/LogMessages.cs
GeoK: If I'm understanding correctly, the idea is to report IDX23032 when GetPopKeysFromJkuAsync method returns null, and otherwise report the exact count of keys (0 or >1).
Is this correct?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Properties/AssemblyInfo.cs
GeoK: Great catch. There was an internal methods that was removed during the process. 
I'll remove these.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestClaimTypes.cs
GeoK: Would you prefer keeping the member here and getting the value from ConfirmationClaimTypes or removing them from SignedHttpRequestClaimTypes?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestClaimTypes.cs
GeoK: Would you prefer keeping the member here and getting the value from ConfirmationClaimTypes or removing them from SignedHttpRequestClaimTypes?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestCreationParameters.cs
GeoK: I think that there is no harm in allowing negative values, thought I can't think of a use case where it might be applicable.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestDescriptor.cs
GeoK: It's an extensibility point to support different type of keys.
It might be useful when EC keys are used on on targets other than netstandard2.0.

Another possible use case might be symmetric keys. A user could ask for a PoP protected access token using the symmetric key thumbprint as req_cnf claim. A symmetric key can't be included in plain text to a SHR payload, it has to be encrypted.
As we don't support this scenario out-of-the-box, users could manually create the outer (SHR's) cnf claim and set it as a jwe.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: TimeAdjustment is a TimeSpan. I'm not sure how to use it with EpochTime.
I'm using EpochTime to calculate the 'ts' claim.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/541bfca2bbe83d65db4df44f783b2c966f1fd67b/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs#L196-L197

Are you suggesting that I should replace ln. 197 with:

```csharp
payload.Add(SignedHttpRequestClaimTypes.Ts, EpochTime.GetIntDate(signedHttpRequestCreationTime));
```

src/Microsoft.IdentityModel.Tokens/CallContext.cs
GeoK: Agreed. I'll remove it.

src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs
GeoK: This check is there for the class property that might be null. 
If 'headers' parameter is null, we return:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/541bfca2bbe83d65db4df44f783b2c966f1fd67b/src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs#L69-L70

src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs
GeoK: Sure, sounds like a good idea.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs
GeoK: I'm not sure how to implement that. We don't have a Core-only target.

src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs
GeoK: Addressed with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/b19c53cf2297adcd4ad0989b361b9faf5a20efae

src/Microsoft.IdentityModel.Tokens/CallContext.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/a7c7b608fbd0da084e479c531b45120cf8f738c6

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Properties/AssemblyInfo.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/0f1a8a812cbf112dc90e319739091d88521563ce

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/LogMessages.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/fba00ba395f903d4080d68b9d81cd47b929c1e56

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/LogMessages.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/fba00ba395f903d4080d68b9d81cd47b929c1e56

src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs
GeoK: Addressed with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/b19c53cf2297adcd4ad0989b361b9faf5a20efae

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Exceptions/SignedHttpRequestCreationException.cs
brentschmaltz: OH, we should make sure that this code is in .NET DESKTOP.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestClaimTypes.cs
brentschmaltz: Getting values from JwtRegisteredClaimNames.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestDescriptor.cs
```diff
@@ -0,0 +1,142 @@
...
+
+using Microsoft.IdentityModel.Logging;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
+{
+    /// <summary>
+    /// Structure that wraps parameters needed for SignedHttpRequest creation. 
+    /// </summary>
+    public class SignedHttpRequestDescriptor
```
brentschmaltz: I think we could collapse SignedHttpRequestDescriptor and SignedHttpCreationParameters into a single class. 
This would simplify the API.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestCreationParameters.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// A delegate that will be called to create and add the nonce claim, overriding the default behavior.
+    /// </summary>
+    /// <param name="payload">A SignedHttpRequest payload.</param>
+    /// <param name="signedHttpRequestDescriptor">A structure for wrapping parameters needed for SignedHttpRequest creation.</param>
+    public delegate void CustomNonceCreator(IDictionary<string, object> payload, SignedHttpRequestDescriptor signedHttpRequestDescriptor);
+
+    /// <summary>
+    /// Defines a set of parameters that are used by a <see cref="SignedHttpRequestHandler"/> when creating a SignedHttpRequest.
+    /// </summary>
+    public class SignedHttpRequestCreationParameters
```
brentschmaltz: Consider merging with SignedHttpRequestDescriptor

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -0,0 +1,1417 @@
...
+
+            return ConvertToJson(payload);
+        }
+
+        /// <summary>
+        /// Converts a dictionary representation of HttpRequest into a JSON string.
+        /// </summary>
+        /// <param name="payload">HttpRequest payload represented as a <see cref="Dictionary{TKey, TValue}"/>.</param>
+        /// <returns>A JSON string.</returns>
+        protected virtual string ConvertToJson(Dictionary<string, object> payload)
```
brentschmaltz: consider moving to utilities class

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -0,0 +1,1417 @@
...
+        }
+
+        /// <summary>
+        /// Creates a JSON representation of a HttpRequest header.
+        /// </summary>
+        /// <param name="signedHttpRequestDescriptor">A structure that wraps parameters needed for SignedHttpRequest creation.</param>
+        /// <returns>A JSON representation of an HttpRequest header.</returns>
+        protected virtual string CreateHttpRequestHeader(SignedHttpRequestDescriptor signedHttpRequestDescriptor)
+        {
+            var header = new JObject
```
brentschmaltz: Let JsonWebTokenHandler create the header

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -0,0 +1,1417 @@
...
+        private readonly HttpClient _defaultHttpClient = new HttpClient();
+
+        #region SignedHttpRequest creation
+        /// <summary>
+        /// Creates a signed http request using the <paramref name="signedHttpRequestDescriptor"/>.
+        /// /// </summary>
+        /// <param name="signedHttpRequestDescriptor">A structure that wraps parameters needed for SignedHttpRequest creation.</param>
+        /// <param name="cancellationToken">Propagates notification that operations should be canceled.</param>
+        /// <returns>A signed http request as a JWS in Compact Serialization Format.</returns>
+        public async Task<string> CreateSignedHttpRequestAsync(SignedHttpRequestDescriptor signedHttpRequestDescriptor, CancellationToken cancellationToken)
```
brentschmaltz: Our code would be simpler if when creating a JWS this class focused on building the payload and then calling JsonWebTokenHandler.CreateToken(string, SigningCredentials, IDictionary<string, object>)

see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/110ced71d96efaaaa4db42eb5f3737b7f31ec2d0/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L221

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestCreationParameters.cs
```diff
@@ -0,0 +1,129 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
+{
+    /// <summary>
+    /// A delegate that will be called to create and add additional claim(s), if set.
+    /// </summary>
+    /// <param name="payload">A SignedHttpRequest payload.</param>
+    /// <param name="signedHttpRequestDescriptor">A structure for wrapping parameters needed for SignedHttpRequest creation.</param>
+    public delegate void AdditionalClaimCreator(IDictionary<string, object> payload, SignedHttpRequestDescriptor signedHttpRequestDescriptor);
```
brentschmaltz: Can we add a property on SignedHttpRequestDescriptor?
Do we need this delegate?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -0,0 +1,1417 @@
...
+                throw LogHelper.LogArgumentNullException(nameof(confirmationClaim));
+
+            var cnf = JObject.Parse(confirmationClaim);
+            if (cnf.TryGetValue(ConfirmationClaimTypes.Jwk, StringComparison.Ordinal, out var jwk))
+            {
+                return ResolvePopKeyFromJwk(jwk.ToString(), signedHttpRequest, validatedAccessToken, signedHttpRequestValidationContext);
+            }
+            else if (cnf.TryGetValue(ConfirmationClaimTypes.Jwe, StringComparison.Ordinal, out var jwe))
+            {
+                return await ResolveP
```
brentschmaltz: Can we put this in a utility class?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestDescriptor.cs
```diff
@@ -0,0 +1,142 @@
...
+        /// <remarks>
+        /// <paramref name="accessToken"/> has to contain the 'cnf' claim so that PoP key can be resolved on the validation side.
+        /// https://tools.ietf.org/html/rfc7800#section-3.1
+        /// Default <see cref="SignedHttpRequestCreationParameters"/> and <see cref="CallContext"/> will be created. 
+        /// </remarks>
+        /// <param name="accessToken">An access token that contains the 'cnf' claim.</param>
+        /// <param name="httpRequestData">A structure that represents an outgoing http request.</param>
+        /// <param name="signingCredentials">A security key and algorithm that will be used to sign the (Signed)HttpRequest.</param>
+        /// <param name="callContext">An opaque context used to store work when working with authentication artifacts.</param>
+        public SignedHttpRequestDescriptor(string accessToken, HttpRequestData httpRequestData, SigningCredentials signingCredentials, CallContext callContext)
```
brentschmaltz: Context represents a single call.
Since SignedHttpRequestDescriptor could be reused, perhaps it shouldn't be a property, but only passed to SignedHttpRequestHandler methods.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: I'll move it, it seems like the method fits better in the util class.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: I'll refactor the code, it makes sense to utilize JsonWebTokenHandler. The method will be synchronous for now, but we can add another async method when we support async in JsonWebTokenHandler.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: I'll refactor the code, it makes sense to utilize JsonWebTokenHandler. The method will be synchronous for now, but we can add another async method when we support async in JsonWebTokenHandler.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestCreationParameters.cs
GeoK: It can be a property. I had a delegate counterpart on the validation side so I wanted a delegate on the creation side, for consistency.
Should I remove the `AdditionalClaimValidatorAsync` delegate?
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/a7c7b608fbd0da084e479c531b45120cf8f738c6/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs#L39-L46

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: Yes, makes sense. Will do.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestDescriptor.cs
GeoK: It makes sense to decouple CallContext from the descriptor. I'll refactor the code so that the CallContext is passed as a method parameter. If a user didn't create a context, a default one would be created.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestDescriptor.cs
GeoK: The reason behind having two classes is a coding model. 
CreationParameters are usually static (defined once), while the Descriptor is created for each new SignedHttpRequest.
This models fits with our S2S scenarios, and outbound policies.

Please let me know if you would still prefer to see these as a single class.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestCreationParameters.cs
GeoK: The reason behind having two classes is a coding model. 
CreationParameters are usually static (defined once), while the Descriptor is created for each new SignedHttpRequest.
This models fits with our S2S scenarios, and outbound policies.

Please let me know if you would still prefer to see these as a single class.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestClaimTypes.cs
GeoK: Ok, I'll make the changes.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Exceptions/SignedHttpRequestCreationException.cs
GeoK: I'm not sure if this code ever worked. Even if there was a DESKTOPNET45 directive defined, I think that this code wouldn't compile.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/a1656abf4f653567a5d6e97a49aeed2559d331a5

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/a1656abf4f653567a5d6e97a49aeed2559d331a5

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/0fac2cc80a8875dea87c068ef99978ca51ad68ba

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/0fac2cc80a8875dea87c068ef99978ca51ad68ba

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/ConfirmationClaimTypes.cs
GeoK: I'd actually avoid putting these claims to JwtRegisteredClaims as they are only specified in the [rfc7800](https://tools.ietf.org/html/rfc7800#section-6.1), in the confirmation method context. 
As these are payload claims I'd also like not to use JwtHeaderParameterNames.
I'd prefer keeping this class as is, and alter SignedHttpRequestClaimTypes.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestClaimTypes.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/67d5dad46791ea015501fdc414f65ac6afb60a1e

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestClaimTypes.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/67d5dad46791ea015501fdc414f65ac6afb60a1e

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/LogMessages.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/656a75672b90545592cd87d09a8c88c9901e6811

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestCreationParameters.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/c465dc691d052b7ad13baee9219b69eb972c65d0 and https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/c5694079f20b5961bf3b4f100c12d9a144a8f79f

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestDescriptor.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/a038e36f12bd6e6be52f812f2bbbd2afa3c214ca

GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/a038e36f12bd6e6be52f812f2bbbd2afa3c214ca

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestDescriptor.cs
GeoK: After the review the team decided not to do this.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestCreationParameters.cs
GeoK: After the review the team decided not to do this.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: Todo: Move to M.IM.Tokens assembly.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Exceptions/SignedHttpRequestCreationException.cs
GeoK: todo: open up a new issue on custom exception serialization.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/LogMessages.cs
GeoK: todo: resolving pop key might throw.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs
GeoK: todo: Replace with a delegate.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: todo: the assumption is correct.

src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs
GeoK: todo: headers.Any is unnecessary. 
todo: Initialize Headers dictionary. Don't allow null setter.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -0,0 +1,1417 @@
...
+        {
+            if (payload == null)
+                throw LogHelper.LogArgumentNullException(nameof(payload));
+
+            var httpRequestUri = signedHttpRequestDescriptor.HttpRequestData.Uri;
+
+            if (httpRequestUri == null)
+                throw LogHelper.LogArgumentNullException(nameof(signedHttpRequestDescriptor.HttpRequestData.Uri));
+
+            if (!httpRequestUri.IsAbsoluteUri)
```
rymeskar: Should the utility EnsureAbsoluteUri be used here as well?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -0,0 +1,1417 @@
...
+
+            if (!(signedHttpRequest is JsonWebToken jwtSignedHttpRequest))
+                throw LogHelper.LogExceptionMessage(new SignedHttpRequestValidationException(LogHelper.FormatInvariant(LogMessages.IDX23031, signedHttpRequest.GetType(), typeof(JsonWebToken), signedHttpRequest)));
+
+            if (!jwtSignedHttpRequest.TryGetPayloadValue(SignedHttpRequestClaimTypes.Ts, out long tsClaimValue))
+                throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidTsClaimException(LogHelper.FormatInvariant(LogMessages.IDX23003, SignedHttpRequestClaimTypes.Ts)));
+
+            DateTime utcNow = DateTime.UtcNow;
+            DateTime signedHttpRequestCreationTime = EpochTime.DateTime(tsClaimValue);
+            DateTime signedHttpRequestExpirationTime = signedHttpRequestCreationTime.Add(signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.SignedHttpRequestLifetime);
```
rymeskar: Usually JWT validation libraries have the ClockSkew setting. This setting is not intended for the signed http request?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -0,0 +1,1417 @@
...
+                throw LogHelper.LogExceptionMessage(new SignedHttpRequestValidationException(LogHelper.FormatInvariant(LogMessages.IDX23031, validatedAccessToken.GetType(), typeof(JsonWebToken), validatedAccessToken)));
+
+            // use the decrypted jwt if the jwtValidatedAccessToken is encrypted.
+            if (jwtValidatedAccessToken.InnerToken != null)
+                jwtValidatedAccessToken = jwtValidatedAccessToken.InnerToken;
+            
+            if (jwtValidatedAccessToken.TryGetPayloadValue(ConfirmationClaimTypes.Cnf, out JObject cnf) && cnf != null)
+                return cnf.ToString();
+            else if (jwtValidatedAccessToken.TryGetPayloadValue(ConfirmationClaimTypes.Cnf, out string cnfString) && !string.IsNullOrEmpty(cnfString))
+                return cnfString;
```
rymeskar: Perf consideration:
GetCnfClaimValue returns string. Either directly from the cnf claim or from the JObject.ToString.

Then, this string is again parsed to JObject in ResolvePopKeyFromCnfClaimAsync.

What about returning the JObject from GetCnfClaimValue directly?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -0,0 +1,1417 @@
...
+        protected virtual string CreateHttpRequestHeader(SignedHttpRequestDescriptor signedHttpRequestDescriptor)
+        {
+            var header = new JObject
+            {
+                { JwtHeaderParameterNames.Alg, signedHttpRequestDescriptor.SigningCredentials.Algorithm },
+                { JwtHeaderParameterNames.Typ, SignedHttpRequestConstants.TokenType }
+            };
+
+            if (signedHttpRequestDescriptor.SigningCredentials.Key is X509SecurityKey x509SecurityKey)
+                header[JwtHeaderParameterNames.X5t] = x509SecurityKey.X5t;
```
rymeskar: Curious:
Under what conditions should the x5t claim in the JWT header be used?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -0,0 +1,1417 @@
...
+                throw LogHelper.LogArgumentNullException(nameof(signedHttpRequestDescriptor.HttpRequestData.Uri));
+
+            httpRequestUri = EnsureAbsoluteUri(httpRequestUri);
+            var sanitizedQueryParams = SanitizeQueryParams(httpRequestUri);
+
+            StringBuilder stringBuffer = new StringBuilder();
+            List<string> queryParamNameList = new List<string>();
+            try
+            {
+                var lastQueryParam = sanitizedQueryParams.LastOrDefault();
```
rymeskar: Perf:
wouldn't it be faster to instead have a first query parameter bool switch ?

Dirty example:

```               
                var firstQueryParam = true;
 
                foreach (var queryParam in sanitizedQueryParams)
                {
		       if (!firstQueryParam)
		       {
			  stringBuffer.Append("&");
		       }
					
                    queryParamNameList.Add(queryParam.Key);
                    var encodedValue = $"{queryParam.Key}={queryParam.Value}";

                    stringBuffer.Append(encodedValue);
					
					firstQueryParam = false;
                }
```

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: The 'u' claim represents an http host component, which requires presence of an absolute URI.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: SignedHttpRequestLifetime is set to 5 minutes (by default), but it's configurable via SignedHttpRequestValidationParameters.
For signed http request, the clock skew is moved to the client side i.e. clients should take care of computing the skew and adjusting the time.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/7d9235b0dd4aea3b6082c4c1080408ae377f4b06/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestCreationParameters.cs#L98-L102

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: That's right, great catch.
These methods were initially protected, and we didn't want to expose Newtonsoft objects.
I'll make the change.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: This code is not used anymore and JsonWebToken handler is utilized instead: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/a1656abf4f653567a5d6e97a49aeed2559d331a5

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: I agree, it's simpler too https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/8898f83d7adde7173222729b31c1338f164baf01

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
GeoK: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1260/commits/4ad150d3bb53a80d672b8afda8202e70eeaa13a8

brentschmaltz:
:shipit:

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
rymeskar: @GeoK and @brentschmaltz and may I ask why the change in the skewing logic compared to regular JWT tokens?

Considerations is that the actual skew could be both + and - and the TimeAdjustment is always in just one direction.

PR #1265, opened 2019-10-10
OpenIdConnectConfiguration.SigningKeys property is no longer used during deserialization

### LINKED ISSUE #1242: OpenIdConnectConfiguration.Create(string json) throws deserialization exception when security keys present in json
Version: 5.5.0

When trying to create a new `OpenIdConnectConfiguration` object from a json string created with the `OpenIdConnectConfiguration.Write` static method a deserialization exception is thrown with the following error message: 
```Could not create an instance of type Microsoft.IdentityModel.Tokens.SecurityKey. Type is an interface or abstract class and cannot be instantiated. Path 'SigningKeys[0].KeySize'```

I used the `OpenIdConnectConfigurationRetriever` class to initially create the OpenIdConnectConfiguration object. So doing something like the following throws the exception for me:

```
var config = OpenIdConnectConfiguration.Create(
                OpenIdConnectConfiguration.Write(
                    await OpenIdConnectConfigurationRetriever.GetAsync($"https://myidentityserveruri.com/.well-known/openid-configuration", CancellationToken.None)));

mafurman:
Fixes #1242.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
@@ -519,17 +519,6 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         }
 
         /// <summary>
-        /// Gets a bool that determines if the 'SigningKeys' property should be serialized.
-        /// This is used by Json.NET in order to conditionally serialize properties.
-        /// </summary>
-        /// <return>true if 'SigningKeys' is not empty; otherwise, false.</return>
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        public bool ShouldSerializeSigningKeys()
-        {
-            return SigningKeys.Count > 0;
-        }
-
-        /// <summary>
         /// Gets a bool that determines if the 'scopes_supported' (ScopesSupported) property should be serialized.
         /// This is used by Json.NET in order to conditionally serialize properties.
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
```diff
@@ -306,6 +306,7 @@ public OpenIdConnectConfiguration(string json)
         /// <summary>
         /// Gets the <see cref="ICollection{SecurityKey}"/> that the IdentityProvider indicates are to be used signing tokens.
         /// </summary>
+        [JsonIgnore]
```
GeoK: Marking a property with `[JsonIgnore]` will exclude a property from both serialization and deserialization. 

We are still using conditional serialization for the SigningKeys property: `[bool ShouldSerialize{SigningKeys}()]`.
I'm wondering what takes priority, JsonIgnore or conditional serialization? Also, should we even serialize SigningKeys to JSON i.e. should we remove `ShouldSerializeSigningKeys`?

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/d4d6d3598952410ff96ef8f38c9ca9ce74bca8bf/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs#L526-L529

brentschmaltz: I don't think we should serialize SigningKeys since it's not the standard.
Users can overload this method if they need to do this.

mafurman: Will remove that method.

mafurman: This method has been removed.

PR #1272, opened 2019-10-16
Validators call delegates before checking if validation should occur

### LINKED ISSUE #1100: If user sets TokenValidationParameters.TokenReplayValidator, it should be called
If a users has set TokenValidationParameters.TokenReplayValidator delegate, it should be called.
Currently, the user has to set the delegate AND set TokenValidationParameters.ValidateTokenReplay

This could be solved by setting the default to true || examining both values.

mafurman:
Fixes #1100.

Ex. `TokenValidationParameters.AudienceValidator` is called before the value of `TokenValidationParameters.ValidateAudience` is checked.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
@@ -214,8 +214,8 @@ namespace Microsoft.IdentityModel.Tokens
         /// Gets or sets a delegate that will be used to validate the audience.
         /// </summary>
         /// <remarks>
-        /// If set, this delegate will be called to validate the 'audience' instead of normal processing.
-        /// If <see cref="ValidateAudience"/> is false, this delegate will not be called.
+        /// If set, this delegate will be called to validate the 'audience', instead of normal processing. This means that no default 'audience' validation will occur.
+        /// Even if <see cref="ValidateAudience"/> is false, this delegate will still be called.
         /// </remarks>
         public AudienceValidator AudienceValidator { get; set; }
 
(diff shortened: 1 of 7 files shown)


henrik-me:
>                             return;

for my education: Would there ever be a situation where there would be multiple audiences which could be valid? (the method takes an enum of audiences), can you pls. help describe the scenario in which this would happen?

---
Refers to: src/Microsoft.IdentityModel.Tokens/Validators.cs:93 in dc9b6b5. [](commit_id = dc9b6b56aeb96a4bcce7ab80f74a3e8a90942d0d, deletion_comment = False)

mafurman:
> 
> 
> > ```
> >                         return;
> > ```
> 
> for my education: Would there ever be a situation where there would be multiple audiences which could be valid? (the method takes an enum of audiences), can you pls. help describe the scenario in which this would happen?
> 
> Refers to: src/Microsoft.IdentityModel.Tokens/Validators.cs:93 in [dc9b6b5](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/dc9b6b56aeb96a4bcce7ab80f74a3e8a90942d0d). [](commit_id = [dc9b6b5](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/dc9b6b56aeb96a4bcce7ab80f74a3e8a90942d0d), deletion_comment = False)

@henrik-me I believe this would mainly be used in the scenario where you could have multiple different represenations of the same audience.

mafurman:
This PR has been rebased against _dev_ (previously was _dev5x_).

AndersAbel:
> > ```
> >                         return;
> > ```
> 
> for my education: Would there ever be a situation where there would be multiple audiences which could be valid? (the method takes an enum of audiences), can you pls. help describe the scenario in which this would happen?
> 
> Refers to: src/Microsoft.IdentityModel.Tokens/Validators.cs:93 in [dc9b6b5](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/dc9b6b56aeb96a4bcce7ab80f74a3e8a90942d0d). [](commit_id = [dc9b6b5](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/dc9b6b56aeb96a4bcce7ab80f74a3e8a90942d0d), deletion_comment = False)

Yes, it is sometimes used for multitenancy SAS deployments. When each tenant/customer has their own domain that is just a CNAME to a common hosting instance each tenant might have their own audience value. The validation on the common host would then accept an audience value for any of the tenant hosts.

PR #1275, opened 2019-10-17
Added 'typ' header claim validation to JwtSecurityTokenHandler and JsonWebTokenHandler

### LINKED ISSUE #1220: Add validation for JWT type
According to the latest JWT BCP - different types of JWTs should use different values for the `typ` header field.

https://tools.ietf.org/html/draft-ietf-oauth-jwt-bcp-06#section-3.11

It would be useful to have a `ValidType` property on the `TokenValidationParameters` to be able to easily add this validation check.

mafurman:
Fixes #1220.

Changes include:

- Added 'typ' header claim validation for both `JwtSecurityTokenHandler` and `JsonWebTokenHandler`.
- Added a `SecurityTokenInvalidTypeException`.
- Added a `ValidTypes` property to the TokenValidationParameters.
- Added an overloaded method for creating an unsigned JWS (using `JsonWebTokenHandler`) that also allows for additional header parameters.
- Fixed comment on `SecurityTokenDescriptor.AdditionalHeaderClaims` to clarify that the 'typ' header claim can be overridden.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
@@ -86,6 +86,10 @@ namespace Microsoft.IdentityModel.Tokens
         public const string IDX10252 = "IDX10252: RequireSignedTokens property on ValidationParameters is set to false and the issuer signing key is null. Exiting without validating the issuer signing key.";
         public const string IDX10253 = "IDX10253: RequireSignedTokens property on ValidationParameters is set to true, but the issuer signing key is null.";
         public const string IDX10254 = "IDX10254: '{0}.{1}' failed. The virtual method '{2}.{3}' returned null. If this method was overridden, ensure a valid '{4}' is returned.";
+        public const string IDX10255 = "IDX10255: ValidTypes property on ValidationParameters is either null or empty. Exiting without validating the token type.";
+        public const string IDX10256 = "IDX10256: Unable to validate the token type. TokenValidationParameters.ValidTypes is set, but the 'typ' header claim is null or empty.";
+        public const string IDX10257 = "IDX10257: Token type validation failed. Type: '{0}'. Did not match: validationParameters.TokenTypes: '{1}'.";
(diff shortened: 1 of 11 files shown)


mafurman:
Since it's relevant to this PR, I created a documentation issue for showing users how to create a JWE that has an inner token with a different type: #1278

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -86,6 +86,10 @@ internal static class LogMessages
         public const string IDX10252 = "IDX10252: RequireSignedTokens property on ValidationParameters is set to false and the issuer signing key is null. Exiting without validating the issuer signing key.";
         public const string IDX10253 = "IDX10253: RequireSignedTokens property on ValidationParameters is set to true, but the issuer signing key is null.";
         public const string IDX10254 = "IDX10254: '{0}.{1}' failed. The virtual method '{2}.{3}' returned null. If this method was overridden, ensure a valid '{4}' is returned.";
+        public const string IDX10255 = "IDX10255: ValidTypes property on ValidationParameters null. Exiting without validating the token type.";
```
brentschmaltz: might want to write: ValidationParameters.ValidTypes is null or empty. Exiting ....

reason the logic checks for null and count > 0

mafurman: Good catch! Fixed this.

brentschmaltz (approved):
:shipit:

GeoK (approved):

PinpointTownes:
Woot, excellent! Is that part of the latest nightly builds? If so, any chance you could pin one version? 😄

brentschmaltz:
@PinpointTownes we released this as as 5.6.0. see https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/releases
Try it out, we are planning another release with support for: PoP {signedHttpRequests, WsTrust 1.3} at end of November.
Let us know if this works well for you or needs some tweaks.

Issue #1261, opened 2019-10-01
JWT claim in the form of DateTime get extra double-quote when converted to claims


oliviervaillancourt:
Hi,

After an upgrade from 4.x to 5.5 of *System.IdentityModel.Tokens.Jwt*, we are noticing that if we read a Jwt that has a claim as a DateTime, then the value of Claim object is surrounded with double-quote.

Repro:
run this code:
```
var backendToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJjbGFpbV9hc19kYXRldGltZSI6IjIwMTktMTEtMTVUMTQ6MzE6MjEuNjEwMTMyNloifQ.yYcHSl-rNT2nHe8Nb0aWe6Qu3E0ZOn2_OUidpxuw0wk";

            var handler = new JwtSecurityTokenHandler();

            var token = handler.ReadToken(backendToken) as JwtSecurityToken;
            var dateClaim = token.Claims.First();

            var claimValue = dateClaim.Value;
            Console.WriteLine(claimValue);
```
 
Notice that the value of `claimValue` is `"2019-11-15T14:31:21.6101326Z"` instead  of `2019-11-15T14:31:21.6101326Z` (extra double-quote).

From looking into the code a little it seems that this related to the use of JsonConvert.DeserializeObject without specifying a DateFormatHandling property.

Thanks

GeoK:
Thank you for reporting this issue @oliviervaillancourt.
The fix will be released in v5.6.0.

PR #1276, opened 2019-10-17
Fix DateTime claims processing

GeoK:
Currently, claims created from a deserialized JSON string have a `System.DateTime` type.
This PR changes that behavior, so claims that hold DateTime values have `ClaimValueTypes.DateTime` as a type.

Also, by default, Newtonsoft.JSON serializer encloses a DateTime string with quotes which results in "\"dateTimeValue\"" (see #1261).
This behavior is changed, so that resulting string doesn't contain the double quotes.

DateTime claim value (string) is now created as an UTC string represented in ISO 8061 date-time format.

Resolves: #1261


GeoK:
Closing in favor of a simple solution.

Issue #1262, opened 2019-10-04
Time related properties ignore timezone information


SamuelDebruyn:
I use the following code to check if my token has expired or not:

```
var decoded = new JwtSecurityTokenHandler().ReadJwtToken(accessToken);
return decoded.ValidTo > DateTime.UtcNow;
```

However, it's not clear if I should use `UtcNow` or `Now`.

Timezone information seems to be lost in this conversion. If I'm not mistaken a JWT token stores the expiration timestamp as seconds since unix epoch in UTC so we should be sure that the `ValidTo` property is in UTC.

[RFC 7519](https://tools.ietf.org/html/rfc7519#section-4.1.4) specifies the following about the expiration:

```
The "exp" (expiration time) claim identifies the expiration time on
   or after which the JWT MUST NOT be accepted for processing.  The
   processing of the "exp" claim requires that the current date/time
   MUST be before the expiration date/time listed in the "exp" claim.
   Implementers MAY provide for some small leeway, usually no more than
   a few minutes, to account for clock skew.  Its value MUST be a number
   containing a NumericDate value.  Use of this claim is OPTIONAL.
```

The same RFC has this to say about the `NumericDate`:
```
A JSON numeric value representing the number of seconds from
      1970-01-01T00:00:00Z UTC until the specified UTC date/time,
      ignoring leap seconds.  This is equivalent to the IEEE Std 1003.1,
      2013 Edition [POSIX.1] definition "Seconds Since the Epoch", in
      which each day is accounted for by exactly 86400 seconds, other
      than that non-integer values can be represented.  See RFC 3339
      [RFC3339] for details regarding date/times in general and UTC in
      particular.
```

I suggest one of either 2 solutions:

- Make `ValidTo` and other time related properties a `DateTimeOffset` so that TZ info is included
- Add clear documentation on the property indicating that the value is in UTC
- Out of the box solution: rely on NodaTime :)

Sidenote: the RFC also says that the expiration is an optional claim so maybe it should even be nullable.

mafurman:
@SamuelDebruyn We will add documentation that states that we use UTC everywhere for this release (5.6.0). We can't change the method to return a DateTimeOffset as that would be a breaking change.

PR #1279, opened 2019-10-17
All DateTime properties are clearly marked as UTC

mafurman:
Addresses #1262.

I went through all our public DateTime properties and made sure that the documentation comments clearly state that the properties are in UTC.

In the case where a property can be set by a user, the comment states that it "should" be UTC, whereas if it's set by our library the comment states that it's "always" in UTC. 

Some of the properties have setters that convert DateTimes to UTC, so I added a different comment for these that explains this behavior.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
+++ b/src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
@@ -52,7 +52,7 @@ namespace Microsoft.IdentityModel.Tokens
         public EncryptingCredentials EncryptingCredentials { get; set; }
 
         /// <summary>
-        /// Gets or sets the value of the 'expiration' claim.
+        /// Gets or sets the value of the 'expiration' claim. This value should be in UTC.
         /// </summary>
         public DateTime? Expires { get; set; }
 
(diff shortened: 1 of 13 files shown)


mafurman:
Documentation page for DateTime usage has been added here:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/How-are-DateTime-values-treated%3F

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
```diff
@@ -52,7 +52,7 @@ public class SecurityTokenDescriptor
         public EncryptingCredentials EncryptingCredentials { get; set; }
 
         /// <summary>
-        /// Gets or sets the value of the 'expiration' claim.
+        /// Gets or sets the value of the 'expiration' claim. This value should be in UTC.
```
henrik-me: >This value should be in UTC. [](start = 62, length = 28)

consider adding an aka.ms link to the wiki page you created

henrik-me (approved):
:shipit:

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
henrik-me: [https://aka.ms/msal-net-on-behalf-of](https://aka.ms/msal-net-on-behalf-of)

---
In reply to: [336267198](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1279#discussion_r336267198) [](ancestors = 336267198)

GeoK (approved):

brentschmaltz (approved):
:shipit:

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
mafurman: I created an issue to track this: #1283

PR #1280, opened 2019-10-17
Added a setter for the AuthenticationProtocolMessage.Script property

### LINKED ISSUE #1263: 'unsafe-eval' required for .net core 2.2 POST to IssuerAddress (OIDC)
As first raised in https://github.com/aspnet/AspNetCore/issues/14721. We are using ADFS with OpenID Connect for an ASP.NET core application, which we just upgraded from 2.0 to 2.2. 

Once upgraded to 2.2, we found that the form POST to the ADFS server required our content security policy to have ['unsafe-eval'](https://developer.chrome.com/extensions/contentSecurityPolicy#relaxing-eval). The string in question is `window.setTimeout('document.forms[0].submit()', 0);`. 
In 2.0, the script just reads `document.form.submit()`, and therefore works fine even with our CSP. 

It _seems_ to be coming from [BuildFormPost()](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/58bd5959f6210771071f676e1b68a217666a48f1/src/Microsoft.IdentityModel.Protocols/AuthenticationProtocolMessage.cs#L60) in [AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet). I'm not sure how that repository relates to the standard .net core ones, as we aren't using any additional nuget packages for authentication. It was also hard for me to tell how those versions relate to the .net core versions and what that code used to look like. 

We managed to get around this and only relax the policy when the user has no claims, aka, is trying to login, but it still seems like we are missing something. 
* What is the point of setTimeout with a 0? 
* Why does the argument of setTimeout need to be a string? 

### To Reproduce
1. Setup OIDC
2. Set a content security policy
3. Run code
4. Note how 2.2 will get blocked by browser (Chrome) but 2.0 will not. 

### Expected behavior
To not need to have a more relaxed content security policy for page that has the authentication form that is submitted via JavaScript.

mafurman:
Fixes #1263.

Also modified the setters for the other public properties on `AuthenticationProtocolMessage` so that if an `ArgumentNullException` is thrown it includes the name of the property (not just "value").

NOTE: `LogHelper.LogArgumentNullException()` uses the following message:
`"IDX10000: The parameter '{0}' cannot be a 'null' or an empty object."`
This doesn't really make sense for the properties on this class since they can be empty strings. 
Should another log message be added that's just "the parameter cannot be null"?

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/AuthenticationProtocolMessage.cs
+++ b/src/Microsoft.IdentityModel.Protocols/AuthenticationProtocolMessage.cs
@@ -40,6 +40,7 @@ namespace Microsoft.IdentityModel.Protocols
     public abstract class AuthenticationProtocolMessage
     {
         private string _postTitle = "Working...";
+        private string _script = "<script language=\"javascript\">window.setTimeout('document.forms[0].submit()', 0);</script>"; 
         private string _scriptButtonText = "Submit";
         private string _scriptDisabledText = "Script is disabled. Click Submit to continue.";
 
@@ -148,7 +149,7 @@ namespace Microsoft.IdentityModel.Protocols
             set
             {
                 if (value == null)
-                    throw LogHelper.LogArgumentNullException("value");
+                    throw LogHelper.LogArgumentNullException(nameof(IssuerAddress));
 
                 _issuerAddress = value;
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Protocols/AuthenticationProtocolMessage.cs
```diff
@@ -148,7 +149,7 @@ public string IssuerAddress
             set
             {
                 if (value == null)
-                    throw LogHelper.LogArgumentNullException("value");
+                    throw LogHelper.LogArgumentNullException(nameof(IssuerAddress));
```
brentschmaltz: nice.
We should follow this pattern.

PR #1281, opened 2019-10-17
Fix DateTime claims processing

GeoK:
Currently, claims created from a deserialized JSON string might have 'System.DateTime' (invalid claim type) or ClaimValueTypes.String type, depending on the format of a datetime json string, the way how JSON string was deserialized, and json handler being used.

This PR changes fixes this inconsistency, so claims that hold DateTime values in ISO8061 datetime format will have ClaimValueTypes.DateTime as a claim value type. DateTime strings represented in datetime formats other than ISO8061 will have ClaimValueTypes.String as a claim value type.

Also, by default, Newtonsoft.JSON serializer encloses some DateTime strings with quotes which results in ""dateTimeValue"" (see #1261). This behavior is changed, so that resulting string doesn't contain the
double quotes.

DateTime claim value (string) is now created as an UTC string represented in ISO 8061 date-time format, if a DateTime claim is added directly (via 'JwtPayload.Add()').

Wiki link: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/How-are-DateTime-values-treated%3F

Resolves: #1261

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -267,6 +267,12 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     return longValue;
             }
 
+            if (claim.ValueType == ClaimValueTypes.DateTime)
+            {
+                if (DateTime.TryParse(claim.Value, out DateTime dateTimeValue))
+                    return dateTimeValue;
+            }
+
             if (claim.ValueType == JsonClaimValueTypes.Json)
                 return JObject.Parse(claim.Value);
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -267,6 +267,12 @@ internal static object GetClaimValueUsingValueType(Claim claim)
                     return longValue;
             }
 
+            if (claim.ValueType == ClaimValueTypes.DateTime)
+            {
+                if (DateTime.TryParse(claim.Value, out DateTime dateTimeValue))
```
brentschmaltz: what happens if TryParse fails?

GeoK: claim.Value will be returned, same as for the other types where parsing might fail.

Issue #179, opened 2015-06-13
RSACryptoServiceProviderProxy crashes on Mono


PinpointTownes:
```
System.Security.Cryptography.CryptographicException: Keyset does not exist
  at System.Security.Cryptography.RSACryptoServiceProvider.Common (System.Security.Cryptography.CspParameters p) [0x00000] in <filename unknown>:0 
  at System.Security.Cryptography.RSACryptoServiceProvider..ctor (Int32 dwKeySize, System.Security.Cryptography.CspParameters parameters) [0x00000] in <filename unknown>:0 
  at System.Security.Cryptography.RSACryptoServiceProvider..ctor (System.Security.Cryptography.CspParameters parameters) [0x00000] in <filename unknown>:0 
  at System.IdentityModel.Tokens.RSACryptoServiceProviderProxy..ctor (System.Security.Cryptography.RSACryptoServiceProvider rsa) [0x00000] in <filename unknown>:0 
  at System.IdentityModel.Tokens.AsymmetricSignatureProvider..ctor (System.IdentityModel.Tokens.AsymmetricSecurityKey key, System.String algorithm, Boolean willCreateSignatures) [0x00000] in <filename unknown>:0 
  at System.IdentityModel.Tokens.SignatureProviderFactory.CreateProvider (System.IdentityModel.Tokens.SecurityKey key, System.String algorithm, Boolean willCreateSignatures) [0x00000] in <filename unknown>:0 
  at System.IdentityModel.Tokens.SignatureProviderFactory.CreateForSigning (System.IdentityModel.Tokens.SecurityKey key, System.String algorithm) [0x00000] in <filename unknown>:0 
  at System.IdentityModel.Tokens.JwtSecurityTokenHandler.CreateSignature (System.String inputString, System.IdentityModel.Tokens.SecurityKey key, System.String algorithm, System.IdentityModel.Tokens.SignatureProvider signatureProvider) [0x00000] in <filename unknown>:0 
  at System.IdentityModel.Tokens.JwtSecurityTokenHandler.CreateToken (System.String issuer, System.String audience, System.Security.Claims.ClaimsIdentity subject, Nullable`1 notBefore, Nullable`1 expires, System.IdentityModel.Tokens.SigningCredentials signingCredentials, System.IdentityModel.Tokens.SignatureProvider signatureProvider) [0x00000] in <filename unknown>:0 
[...]

PR #1287, opened 2019-10-23
Work around the RSACryptoServiceProviderProxy crash on mono

main--:
While I understand that #179 was closed as wontfix because mono is not officially supported, this is a really trivial workaround that greatly helps mono users. To put this into perspective, we are currently IL-patching `Microsoft.IdentityModel.Tokens.dll` on every startup of our application because there is no easier way to work around this problem in our case (it's behind other libraries).

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
+++ b/src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
@@ -86,7 +86,7 @@ namespace Microsoft.IdentityModel.Tokens
             // 3. we are not running on mono (which reports PROV_RSA_FULL but doesn't need a workaround)
             var isSha1Provider = rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_FULL || rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_SCHANNEL;
             var isMono = Type.GetType("Mono.Runtime") != null;
-            if (isSha1Provider && !rsa.CspKeyContainerInfo.HardwareDevice && !isMono)
+            if (isSha1Provider && !rsa.CspKeyContainerInfo.HardwareDevice)
             {
                 var csp = new CspParameters();
                 csp.ProviderType = PROV_RSA_AES;
@@ -99,7 +99,16 @@ namespace Microsoft.IdentityModel.Tokens
                 // With this flag, a CryptographicException is thrown instead.
                 csp.Flags |= CspProviderFlags.UseExistingKey;
 
-                _rsa = new RSACryptoServiceProvider(csp);
+                try
+                {
+                    _rsa = new RSACryptoServiceProvider(csp);
+                }
+                catch (CryptographicException) when (isMono)
+                {
+                    // On mono, this exception is expected behavior.
+                    // The solution is to simply not level up the provider as this workaround is not needed on mono.
+                    _rsa = rsa;
+                }
 
                 // since we created a new RsaCryptoServiceProvider we need to dispose it
                 _disposeRsa = true;
--- a/src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
+++ b/src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
@@ -83,7 +83,10 @@ namespace Microsoft.IdentityModel.Tokens
             // Level up the provider type only if:
             // 1. it is PROV_RSA_FULL or PROV_RSA_SCHANNEL which denote CSPs that only understand Sha1 algorithms
             // 2. it is not associated with a hardware key
-            if ((rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_FULL || rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_SCHANNEL) && !rsa.CspKeyContainerInfo.HardwareDevice)
+            // 3. we are not running on mono (which reports PROV_RSA_FULL but doesn't need a workaround)
+            var isSha1Provider = rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_FULL || rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_SCHANNEL;
+            var isMono = Type.GetType("Mono.Runtime") != null;
+            if (isSha1Provider && !rsa.CspKeyContainerInfo.HardwareDevice && !isMono)
             {
                 var csp = new CspParameters();
                 csp.ProviderType = PROV_RSA_AES;
--- a/src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
+++ b/src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
@@ -102,6 +102,8 @@ namespace Microsoft.IdentityModel.Tokens
                 try
                 {
                     _rsa = new RSACryptoServiceProvider(csp);
+                    // since we created a new RsaCryptoServiceProvider we need to dispose it
+                    _disposeRsa = true;
                 }
                 catch (CryptographicException) when (isMono)
                 {
@@ -109,9 +111,6 @@ namespace Microsoft.IdentityModel.Tokens
                     // The solution is to simply not level up the provider as this workaround is not needed on mono.
                     _rsa = rsa;
                 }
-
-                // since we created a new RsaCryptoServiceProvider we need to dispose it
-                _disposeRsa = true;
             }
             else
             {


GeoK:
@main-- Which version of IdentityModel are you using?
This class is not available in our netstandard 2.0 target since v5.4.0.

main--:
This is on version 5.5.0 but on the `net472` target.

main--:
Can you tell me how? From what I understand there is no way to build executables against netstandard because [netstandard is not a runtime](https://stackoverflow.com/a/45845351/1590632). Creating a class library with a `Main()` function and running that is not possible either because the resulting assembly will be missing an entry point. Using .NET Core is not an option as it doesn't run on FreeBSD. Mono is the only way for this and plenty of other platforms that are neither Windows nor Linux (nor Mac).

The point of this PR is to add a tiny trivial fix that costs basically nothing and brings a great improvement to all mono users (RSA JWT is broken completely -> RSA JWT works perfectly fine). While a merge would make things easier, the cost and risk of moving to netstandard is definitely not worth it for us. Sorry, I can't tell you whether or not that would solve this problem.

GeoK:
It looks like Mono supports libraries compiled against netstandard2.0:

> Loading and running a library compiled against the 2.0 standard should work fine

https://www.mono-project.com/docs/about-mono/releases/5.2.0/#net-standard-20-groundwork
https://www.mono-project.com/docs/about-mono/releases/6.4.0/#net-standard-21-support

You should continue using net472, but consume IdentityModel compiled against netstandard2.0.
NuGet will automatically select the closest target, in this case net461. You can override this behavior by explicitly referencing IdentityModel assembly compiled against netstandard2.0.

```xml  
<ItemGroup>
    <PackageReference Include="Microsoft.IdentityModel.Tokens" GeneratePathProperty="true">
      <Version>5.6.0</Version>
    </PackageReference>
    <Reference Include="Microsoft.IdentityModel.Tokens">
      <HintPath>$(PkgMicrosoft_IdentityModel_Tokens)\lib\netstandard2.0\Microsoft.IdentityModel.Tokens.dll</HintPath>
    </Reference>
</ItemGroup>
```

If you are not using NuGet, just manually reference IdentityModel assembly compiled against netstandard2.0.

main--:
Thank you! This also resolved the problem. As the project is using `packages.config` I simply change the `HintPath` in the csproj file.

Although this does seem like a really hacky workaround to me. What is the risk/drawback of doing it this way? Do I have to reapply this fix every time I update nuget packages?

GeoK:
I don't see any risks, expect that I don't know how well Mono plays w/ netstandard.
Just wanted to make sure that we look at alternative solutions before we review the PR as the previous issue was closed as "wontfix" (like you pointed out).

moritzuehling:
While this does fix the error, just including the fix would probably make life easier for everyone involved. Editing csproj files because a comment in an issue said so (with no documentation) is always more of a hassle than "it just works."

The PR will likely not affect code complexity, runtime performance or binary size significantly - but it makes the library work on a whole new platform. Sure, that platform is not the target, nonetheless people use it. 

While perhaps the benefits of this PR aren't huge, I don't see much cost associated with clicking merge, either.

GeoK:
I agree that the proposed solution is not ideal, but It's still a working workaround.
Would you be interested in making some changes to this PR and testing one of the preview builds?

main--:
Sure, what changes?

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
```diff
@@ -83,7 +83,10 @@ public RSACryptoServiceProviderProxy(RSACryptoServiceProvider rsa)
             // Level up the provider type only if:
             // 1. it is PROV_RSA_FULL or PROV_RSA_SCHANNEL which denote CSPs that only understand Sha1 algorithms
             // 2. it is not associated with a hardware key
-            if ((rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_FULL || rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_SCHANNEL) && !rsa.CspKeyContainerInfo.HardwareDevice)
+            // 3. we are not running on mono (which reports PROV_RSA_FULL but doesn't need a workaround)
+            var isSha1Provider = rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_FULL || rsa.CspKeyContainerInfo.ProviderType == PROV_RSA_SCHANNEL;
+            var isMono = Type.GetType("Mono.Runtime") == null;
```
GeoK: `Type.GetType("Mono.Runtime") != null` doesn't necessarily indicate that Mono is being used.
For example, one can create **Runtime** class under **Mono** namespace and this check would return true.

A large number of people is using these libraries [(69.9m downloads)](https://www.nuget.org/packages/System.IdentityModel.Tokens.Jwt/) and we should make sure that we are not introducing regressions.

My suggestion would be to check if Mono is being used with `Type.GetType("Mono.Runtime") != null`, and (only) if the answer is `true`, **try** to sign some dummy data using the `rsa` key and SHA-256 hash.
If the operation succeeds, there is no need to up-level the provider type.

GeoK:
Great! I'll put my comments in the review.

GeoK (changes requested):

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
moritzuehling: Checking if `Mono.Runtime` exists is the recommended check. For example [here in the official mono documentation](https://www.mono-project.com/docs/gui/winforms/porting-winforms-applications/#runtime-conditionals) ([secondary place](https://www.mono-project.com/docs/faq/technical/#how-can-i-detect-if-am-running-in-mono)), it's also the [top result on Google / Stack Overflow](https://stackoverflow.com/questions/721161/how-to-detect-which-net-runtime-is-being-used-ms-vs-mono), and is generally the preferred way of doing it. 
If you know of another way, we're of course happy - but that is the only way we know, and that one is used all over C# projects in the wild already. Adding an additional check is of course possible, and we're happy to add it - just clarifying that this isn't anything we've come up with ourselves.

GeoK: Sure - thank you for bringing this out. It's clearly a dominant approach in determining if Mono is being used or not, but it's not necessarily correct in all cases (see the example above).
An extra check would allow us to eliminate doubt in breaking some corner cases and would add a minimal performance hit, but only for Mono users who are blocked now anyways.

main--:
Thank you for your feedback. I fixed the typo.

While it's clearly unfortunate that `Type.GetType("Mono.Runtime") != null` is not a definitive check I would argue that it's so widespread that anyone who accidentally creates a `Mono.Runtime` class in their application should expect to see plenty of breakage from other dependencies as well. A quick search on github alone yields >35k results and includes big names like NUnit and even [nuget](https://github.com/NuGet/NuGet.Client/blob/d62db666c710bf95121fe8f5c6a6cbe01985456f/src/NuGet.Core/NuGet.Common/RuntimeEnvironmentHelper.cs#L16)!

And since the class is not used as of the netstandard2.0 target many users can't be impacted at all.

Alternatively, https://github.com/main--/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/f043db6847ca082b5b8ed1e5eb6b838472547199 catches the exception from re-creating the CSP on mono. This avoids the performance hit of performing a dummy RSA operation every time.

mafurman:
@main-- The alternate solution seems good to me as it handles the corner cases where  `Type.GetType("Mono.Runtime") != null ` but Mono isn't being used. Can you modify the PR to include it?

main--:
@mafurman I added the commit.

moritzuehling:
Hi, I understand that we submitted the final fix during the US thanksgiving weekend, so I'm poking just in case this was overlooked during all the festives.

GeoK:
@main, @moritzuehling - Apologies for the delay in response.
Please rebase this PR against the _dev5x_ branch. We will publish a nightly build so you can verify that the fix is working for you.

main--:
@GeoK Rebase done.

GeoK:
Thank you @main--.
We released a nightly/preview build ([5.7.1-preview-10122233743](https://www.myget.org/feed/azureadwebstacknightly/package/nuget/Microsoft.IdentityModel.Tokens/5.7.1-preview-10122233743)).

Can you please verify if the fix works fine in your (Mono) environment?

Vaskinn:
I've been having this problem when running on mono and validating jwt tokens, and can confirm that 5.7.1-preview-10122233743 solved it for me.

main--:
Thank you, it works!

Vaskinn:
The preview package seems to have disappeared off myget, (If I had to guess because of a retention policy) stopping our testing in its tracks. 

When will it be decided if this makes it into a release, and if it is accepted will there be new preview builds containing it between now and next release?

GeoK (approved):

GeoK:
@Vaskinn - Correct, the package is removed because of a retention policy.
This change will be available in all preview builds starting with "5.7.1-preview-10214203806".
We don't have an ETA for a new release yet.

@main--, @moritzuehling This is great, thank you for your contribution!

lcorneliussen:
It looks like this never got released since the 5.6 is the latest stable release, then followed by 6.5 (ref https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1361)

I tried switching to the latest 6x - but I'm still getting "Keyset does not exist" on Mono.

Interesting to see it is even the latest commit on dev5x:
image

PR #1299, opened 2019-12-04
Fixed the JWT segment counting bug in the JsonWebToken constructor

### LINKED ISSUE #1298: JsonWebToken constructor doesn't count JWT segments properly
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/110ced71d96efaaaa4db42eb5f3737b7f31ec2d0/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L54-L76

As soon as 'count' gets to JwtConstants.JwsSegmentCount (3), the while loop will terminate. This means that as long as the number of dots in the 'jwtEncodedString' is >= 2, the value of 'count' will always be equal to 3. This also means that we'll split the string and try to decode it even if it isn't a properly formatted JWT token.

mafurman:
Fixes #1298.


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -56,16 +56,8 @@ public JsonWebToken(string jwtEncodedString)
...
-            while ((next = jwtEncodedString.IndexOf('.', next + 1)) != -1)
-            {
-                count++;
-                if (count >= JwtConstants.JwsSegmentCount)
-                    break;
-            }
-
-            // JWS or JWE
+            // Number of segments is always one more than the number of '.'s.
+            var count = jwtEncodedString.Count(c => c == '.') + 1;
```
brentschmaltz: What happens with a large number of dots, say 100,000?
This was the reason to avoid split.
You can change the condition to count >= JwtConstants.JweSegmentCount, then check afterwards that it equals Jws or Jwe.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
```diff
@@ -516,6 +517,98 @@ public static TheoryData<ParseTimeValuesTheoryData> ParseTimeValuesTheoryData
             }
         }
 
+        // Test ensures that we only try to populate a JsonWebToken from a string if it is a properly formatted JWT. More specifically, we only want to try and decode
+        // a JWT token if it has the correct number of (JWE or JWS) token parts.
+        [Theory, MemberData(nameof(ParseTokenTheoryData))]
+        public void ParseToken(JwtTheoryData theoryData)
```
brentschmaltz: Nice.

mafurman: Thanks!

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: That makes sense. I've changed the logic here to make use of `string.Split()`, where the maximum number of substrings is equal to `JwConstants.MaxJwtSegmentCount + 1`. This way any string with greater than 5 segments will have `tokenParts.Length == 6`.

GeoK (approved):

brentschmaltz (approved):

Issue #1273, opened 2019-10-16
Feature: Support for validating audience with or with-out '/' in the end


henrik-me:
Add a bool to the validation parameters to indicate that audience validation will accept audience if there is a match for everything but an ending /

Note: Some IDP's returns the audience with a / and others don't

Workaround today is to use audiences where both the audience with the / and the one with-out the / can be added as validaudiences.

PR #1314, opened 2020-01-16
Ignore single ending slash when validating audience

brentschmaltz:
It is common for '/' to show up at the end of the audience.
From a security standpoint, this is not meaningful, we added support for ignoring such slash's.

Issue: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1273

### DIFF
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -38,14 +38,6 @@ namespace Microsoft.IdentityModel.Logging
     /// </summary>
     public class LogHelper
     {
-        private static readonly List<string> CustomExceptionTypePrefixes = new List<string>()
-        {
-            "Microsoft.IdentityModel.Protocols",
-            "Microsoft.IdentityModel.Tokens.SecurityToken",
-            "Microsoft.IdentityModel.Tokens.Saml",
-            "Microsoft.IdentityModel.Xml"
-        };
-
         /// <summary>
         /// Logs an exception using the event source logger and returns new <see cref="ArgumentNullException"/> exception.
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -76,31 +77,48 @@ public static void ValidateAudience(IEnumerable<string> audiences, SecurityToken
             if (string.IsNullOrWhiteSpace(validationParameters.ValidAudience) && (validationParameters.ValidAudiences == null))
                 throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidAudienceException(LogMessages.IDX10208) { InvalidAudience = Utility.SerializeAsSingleCommaDelimitedString(audiences) });
 
-            foreach (string audience in audiences)
+            // create enumeration of all valid audiences from validationParameters
+            IEnumerable<string> validationParametersAudiences = validationParameters.ValidAudiences == null
```
GeoK: nit: var instead of IEnumerable<string>

test/Microsoft.IdentityModel.Tokens.Tests/ValidatorsTests.cs
```diff
@@ -37,40 +37,258 @@ namespace Microsoft.IdentityModel.Tokens.Tests
...
+        }
 
-                return dataset;
+        public static TheoryData<AudienceValidationTheoryData> ValidateAudienceTheoryData
+        {
+            get
+            {
+                var audience1 = "http://audience1.com";
+                var audience2 = "http://audience2.com";
+                List<string> audiences1 = new List<string> { "", audience1 };
```
GeoK: Are the empty strings added to these list so that the code below is tested? https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/5315c1404d94c596ea92bf267411b4e05ee515af/src/Microsoft.IdentityModel.Tokens/Validators.cs#L92-L96

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -344,27 +336,15 @@ public static string FormatInvariant(string format, params object[] args)
...
-                if (IsCustomException(e))
-                    return e.ToString();
-                else
-                    return e.GetType().ToString();
-            }
-            else
-            {
-                return IdentityModelEventSource.HiddenPIIString;
-            }
+            if (arg is Exception ex && ex.GetType().FullName.StartsWith("Microsoft.IdentityModel."))
```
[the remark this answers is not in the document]
brentschmaltz: I agree, will adjust.

src/Microsoft.IdentityModel.Tokens/Validators.cs
brentschmaltz: I agree, will change.

test/Microsoft.IdentityModel.Tokens.Tests/ValidatorsTests.cs
brentschmaltz: yes

PR #1318, opened 2020-01-26
Replace split with explicit token segments count

krlm:
This small PR is changing the way how token segments are count. `Split` is replaced by explicit `for` loop to  counts segments. It should reduce number of allocations made during token validation and positively impact performance of that operation.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -96,13 +96,11 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 return false;
             }
 
-            // Set the maximum number of segments to MaxJwtSegmentCount + 1. This controls the number of splits and allows detecting the number of segments is too large.
-            // For example: "a.b.c.d.e.f.g.h" => [a], [b], [c], [d], [e], [f.g.h]. 6 segments.
-            // If just MaxJwtSegmentCount was used, then [a], [b], [c], [d], [e.f.g.h] would be returned. 5 segments.
-            string[] tokenParts = token.Split(new char[] { '.' }, JwtConstants.MaxJwtSegmentCount + 1);
-            if (tokenParts.Length == JwtConstants.JwsSegmentCount)
+            int tokenPartsCount = CountTokenSegments(token);
+
+            if (tokenPartsCount == JwtConstants.JwsSegmentCount)
                 return JwtTokenUtilities.RegexJws.IsMatch(token);
-            else if (tokenParts.Length == JwtConstants.JweSegmentCount)
+            else if (tokenPartsCount == JwtConstants.JweSegmentCount)
                 return JwtTokenUtilities.RegexJwe.IsMatch(token);
 
             LogHelper.LogInformation(LogMessages.IDX14107);
@@ -116,6 +114,23 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         public virtual bool CanValidateToken
         {
             get { return true; }
+        }
+
+        private int CountTokenSegments(string token)
+        {
+            if (String.IsNullOrEmpty(token))
+                return 0;
+
+            // Counts number of Jwt segments, capped to MaxJwtSegmentCount + 1 to allow detecting when number of segments is too large.
+            // For example: "a.b.c.d.e.f.g.h" => [a], [b], [c], [d], [e], [f.g.h]. 6 segments.
+            // If just MaxJwtSegmentCount was used, then [a], [b], [c], [d], [e.f.g.h] would be returned. 5 segments.
+            int tokenPartsCount = 1;
+            for (int i = 0; i < token.Length && tokenPartsCount < JwtConstants.MaxJwtSegmentCount + 1; i++)
+            {
+                if (token[i] == '.')
+                    tokenPartsCount++;
+            }
+            return tokenPartsCount;
         }
 
         private JObject CreateDefaultJWEHeader(EncryptingCredentials encryptingCredentials, string compressionAlgorithm)
@@ -1133,13 +1148,14 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (token.Length > MaximumTokenSizeInBytes)
                 return new TokenValidationResult { Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(TokenLogMessages.IDX10209, token.Length, MaximumTokenSizeInBytes))) };
 
-            var tokenParts = token.Split(new char[] { '.' }, JwtConstants.MaxJwtSegmentCount + 1);
-            if (tokenParts.Length != JwtConstants.JwsSegmentCount && tokenParts.Length != JwtConstants.JweSegmentCount)
+            var tokenPartsCount = CountTokenSegments(token);
+
+            if (tokenPartsCount != JwtConstants.JwsSegmentCount && tokenPartsCount != JwtConstants.JweSegmentCount)
                 return new TokenValidationResult { Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14111, token))) };
 
             try
             {
-                if (tokenParts.Length == JwtConstants.JweSegmentCount)
+                if (tokenPartsCount == JwtConstants.JweSegmentCount)
                 {
                     var jwtToken = new JsonWebToken(token);
                     var decryptedJwt = DecryptToken(jwtToken, validationParameters);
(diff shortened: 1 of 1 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -118,6 +116,20 @@ public virtual bool CanValidateToken
             get { return true; }
         }
 
+        private int CountTokenSegments(string token)
```
henrik-me: nit: callers of this method checks that this is not null or empty. Code can be moved around and those checks might disappear, perhaps there should be a nullorempty check here as well returning count == 0 for that case? Or alternatively throw an exception. Nit as this is not an issue in the current code...

henrik-me:
would ideally have liked a few unit tests to be exercised ensuring similar behavior, but overall I see no issues with this change and fewer string allocations makes a lot of sense

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -118,6 +116,20 @@ public virtual bool CanValidateToken
...
+        {
+            // Set the maximum number of segments to MaxJwtSegmentCount + 1. This controls the number of splits and allows detecting the number of segments is too large.
```
GeoK: This comment addresses the line that you are replacing with CountTokenSegments() method.
Can you update the comment to reflect your new logic?
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/18ebb69783db82c1911f38c75311aa33425f1ee8/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L102

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
krlm: I've added null check and return 0 in that case. I find it hard to move those checks here as they're serving different purposes in each case, but I'm open to advice.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
krlm: Fixed.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -118,6 +116,23 @@ public virtual bool CanValidateToken
...
+        private int CountTokenSegments(string token)
+        {
+            if (String.IsNullOrEmpty(token))
+                return 0;
+
+            // Counts number of Jwt segments, capped to MaxJwtSegmentCount + 1 to allow detecting when number of segments is too large.
+            // For example: "a.b.c.d.e.f.g.h" => [a], [b], [c], [d], [e], [f.g.h]. 6 segments.
+            // If just MaxJwtSegmentCount was used, then [a], [b], [c], [d], [e.f.g.h] would be returned. 5 segments.
+            int tokenPartsCount = 1;
+            for (int i = 0; i < token.Length && tokenPartsCount < JwtConstants.MaxJwtSegmentCount + 1; i++)
```
KrisVandermotten, line 128: You don't have to test `tokenPartsCount < JwtConstants.MaxJwtSegmentCount + 1` on every iteration.

Consider

            for (int i = 0; i < token.Length; i++)
            {
                if (token[i] == '.')
                {
                    tokenPartsCount++;
                    if (tokenPartsCount > JwtConstants.MaxJwtSegmentCount)
                        break;
                }
            }

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -96,13 +96,11 @@ public virtual bool CanReadToken(string token)
                 return false;
             }
 
-            // Set the maximum number of segments to MaxJwtSegmentCount + 1. This controls the number of splits and allows detecting the number of segments is too large.
-            // For example: "a.b.c.d.e.f.g.h" => [a], [b], [c], [d], [e], [f.g.h]. 6 segments.
-            // If just MaxJwtSegmentCount was used, then [a], [b], [c], [d], [e.f.g.h] would be returned. 5 segments.
-            string[] tokenParts = token.Split(new char[] { '.' }, JwtConstants.MaxJwtSegmentCount + 1);
-            if (tokenParts.Length == JwtConstants.JwsSegmentCount)
+            int tokenPartsCount = CountTokenSegments(token);
+
```
brentschmaltz, line 100: Our style is that normally we would not put an empty line here.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -118,6 +116,23 @@ public virtual bool CanValidateToken
             get { return true; }
         }
 
+        private int CountTokenSegments(string token)
+        {
+            if (String.IsNullOrEmpty(token))
+                return 0;
+
+            // Counts number of Jwt segments, capped to MaxJwtSegmentCount + 1 to allow detecting when number of segments is too large.
```
brentschmaltz, line 124: this comment doesn't seem to apply to this code.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1133,13 +1148,14 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
             if (token.Length > MaximumTokenSizeInBytes)
                 return new TokenValidationResult { Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(TokenLogMessages.IDX10209, token.Length, MaximumTokenSizeInBytes))) };
 
-            var tokenParts = token.Split(new char[] { '.' }, JwtConstants.MaxJwtSegmentCount + 1);
-            if (tokenParts.Length != JwtConstants.JwsSegmentCount && tokenParts.Length != JwtConstants.JweSegmentCount)
+            var tokenPartsCount = CountTokenSegments(token);
+
```
brentschmaltz, line 1152: Our style is that normally we would not put an empty line here

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -118,6 +116,23 @@ public virtual bool CanValidateToken
             get { return true; }
         }
 
+        private int CountTokenSegments(string token)
```
brentschmaltz, line 119: The name of the method is technically misleading since it breaks when number of '.''s is greater than JwtConstants.MaxJwtSegmentCount +1.

brentschmaltz:
@krlm can you please rebase this against dev, thanks.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
lechu445: consider this solution as I think it is more efficient for this use-case:
```c#
        private static int CountSegments(string token)
        {
            if (string.IsNullOrEmpty(token))
                return 0;

            int tokenPartsCount = 1;
            int index = 0;

            while ((index = input.IndexOf('.', index)) != -1)
            {
                index++;
                tokenPartsCount++;
                if (tokenPartsCount > JwtConstants.MaxJwtSegmentCount)
                    break;
            }
            return tokenPartsCount;
        }
```

pharring:
@brentschmaltz, @krlm what's the verdict on this PR? It's over a year old. Is it going to make it?

Reason I'm here is that we've been reviewing expensive string.Split calls across all 1st party Azure services and this one bubbled to the top. In fact, there are four in `System.IdentityModel.Tokens.Jwt`:

* `System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.ValidateToken`
* `System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.ReadJwtToken`
* `System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.CanReadToken`
* `System.IdentityModel.Tokens.Jwt.JsonWebToken..ctor`

The first three be eliminated entirely by "counting dots" -- like this PR does.
The last one is harder to avoid -- particularly since the substrings are actually needed.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
pharring: That's not more efficient because of the extra function calls. IndexOf isn't able to avoid a linear scan.
You can't get much better than @KrisVendermotten's code -- although, I wouldn't be afraid to use `foreach(char ch in token)` here -- the JIT compiler does a stellar job of optimizing CharEnumerator. It's simple and compact.

But really, this is nit-picking over the main improvement here which is avoiding string.Split allocations.

lechu445:
last activity on Github of the author of this PR was March 2020. I am afraid we are going to wait forever for the fixes. I propose to create a new PR with all the changes and fixes after the code review. @pharring @KrisVandermotten  what do you think?

brentschmaltz:
This pr is more comprehensive https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1645, we will try and get this one in

PR #1320, opened 2020-01-29
JsonWebTokenHandler implements ISecurityTokenValidator and the M.IM.Protocols.OpenIdConnect assembly makes use of M.IM.JsonWebTokens

### LINKED ISSUE #1160: M.IM.Protocols.OpenIdConnect needs to support JsonWebTokens
We are moving away from using System.IdentityModel.Tokens.Jwt, so we should have support for Microsoft.IdentityModel.JsonWebTokens in our OpenIdConnect assembly.

OpenIdConnectProtocolValidationContext should have another property on it (either a SecurityToken or simply json or a new interface) that will allow the OpenIdConnectProtocolValidator to use JsonWebToken rather than JwtSecurityToken.

### LINKED ISSUE #1227: JsonWebTokenHandler should support ISecurityTokenValidator

mafurman:
This PR combines the following two PRs: #1243 and #1127.

Tests still need to be added to make sure that the TryGetPayloadValue() (and other similar methods on the IJsonWebToken interface) are implemented properly by JwtSecurityToken.

It may also be a good idea to include a RawData property on either the IJsonWebToken or ISecurityToken interface, as otherwise a user of the library will need to cast to a specific token type in order to obtain the original encoded token. 

(Fixes #1160)
(Fixes #1227)

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -50,6 +50,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// <exception cref="ArgumentException">'jwtEncodedString' is not in JWS or JWE Compact serialization format.</exception>
         /// <remarks>
         /// The contents of the returned <see cref="JsonWebToken"/> have not been validated, the JSON Web Token is simply decoded. Validation can be accomplished using the validation methods in <see cref="JsonWebTokenHandler"/>
+        /// We recommend using the JsonWebToken class over the System.IdentityModel.Tokens.Jwt.JwtSecurityToken class as it is newer, faster, and has more functionality.
         /// </remarks>
         public JsonWebToken(string jwtEncodedString)
         {
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -146,7 +146,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// <summary>
         /// Gets a <see cref="IEnumerable{Claim}"/><see cref="Claim"/> for each JSON { name, value }.
         /// </summary>
-        public IEnumerable<Claim> Claims
+        public virtual IEnumerable<Claim> Claims
         {
             get
             {
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -322,7 +322,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// Gets the 'value' of the 'zip' claim { zip, 'value' }.
         /// </summary>
         /// <remarks>If the 'zip' claim is not found, an empty string is returned.</remarks>   
-        public string Zip => Header.Value<string>(JwtHeaderParameterNames.Zip) ?? String.Empty;
+        public string Zip => Header.Value<string>(JwtHeaderParameterNames.Zip) ?? string.Empty;
 
         /// <summary>
         /// Decodes the string into the header, payload and signature.
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -40,7 +40,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// <summary>
     /// A <see cref="SecurityToken"/> designed for representing a JSON Web Token (JWT). 
     /// </summary>
-    public class JsonWebToken : SecurityToken
+    public class JsonWebToken : SecurityToken, IJsonWebToken
     {
         /// <summary>
         /// Initializes a new instance of <see cref="JsonWebToken"/> from a string in JWS or JWE Compact serialized format.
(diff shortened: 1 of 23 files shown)


src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
```diff
@@ -330,6 +375,20 @@ internal set
...
+        /// Gets the 'value' of the 'kid' claim { kid, 'value' } from the <see cref="JwtHeader"/>.  
+        /// </summary>
+        /// <remarks>If the 'kid' claim is not found, an empty string is returned.</remarks>
+        public string Kid
+        {
+            get
+            { 
+                if (Header != null)
+                    return Header.Kid;
+                return String.Empty;
```
brentschmaltz: It is recommended to use lowercase 'string'. 
We normally see a empty line after an 'if' clause.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
```diff
@@ -212,7 +212,7 @@ public JwtSecurityToken(string issuer = null, string audience = null, IEnumerabl
         /// <summary>
         /// Gets the 'value' of the 'actor' claim { actort, 'value' }.
         /// </summary>
-        /// <remarks>If the 'actor' claim is not found, null is returned.</remarks> 
+        /// <remarks>If the 'actor' claim is not found, an empty string is returned.</remarks>
```
brentschmaltz: an empty string is logically different than null.
Sometimes the existence of a value may open a gate, should we consider returning null?

mafurman: My concern about returning null would be that changing it from an empty string to null is considered a breaking change (for existing properties like `Issuer` and `Actor`). We could do this for new properties that are being added but that would be inconsistent behavior....

src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
mafurman: I've changed it to a lowercase 'string' here and in all other places for consistency.

I've also added empty lines after all the 'if' statements in this class.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -40,7 +40,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// <summary>
     /// A <see cref="SecurityToken"/> designed for representing a JSON Web Token (JWT).
```
pmaytak: Should we add a remark to this class and the handler that this is newer and preferred for use than other Jwt classes in System namespace?

src/Microsoft.IdentityModel.Tokens/IJsonWebToken.cs
```diff
@@ -0,0 +1,117 @@
...
+//------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Security.Claims;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Interface for a json web token (JWT).
```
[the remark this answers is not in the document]
mafurman: Thanks! Fixed. 👍

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -358,73 +383,107 @@ protected virtual void ValidateIdToken(OpenIdConnectProtocolValidationContext va
...
 
-                if (!idToken.Payload.Iat.HasValue)
-                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, JwtRegisteredClaimNames.Iat.ToLowerInvariant(), idToken)));
+            if (string.IsNullOrEmpty(idToken.Issuer))
+                throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, JwtRegisteredClaimNames.Iss.ToLowerInvariant(), idToken)));
 
-                if (idToken.Payload.Iss == null)
-                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, JwtRegisteredClaimNames.Iss.ToLowerInvariant(), idToken)));
+            // sub is required in OpenID spec; but we don't want to block valid idTokens provided by some identity providers
+            if (RequireSub && (string.IsNullOrWhiteSpace(idToken.Subject)))
```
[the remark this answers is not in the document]
mafurman: Whoops, this has been fixed!

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
mafurman: This seems like its a source of confusion for people so I definitely think it's a good idea to add this in! It's also probably a good idea to add this on the old Jwt token and handler classes as well.

Did both in this [commit](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1320/commits/6cc69b1500a8507cde6c582e7314e248a27dc5a2).

henrik-me: Isn't the old supposed to be marked as obsolete? with details saying to use the new one?

henrik-me (approved):
:shipit:

brentschmaltz:
@mafurman @henrik-me @GeoK we actually shouldn't do this as asp.net customers will be broken if they switch to JsonWebTokenHandler as ValidateToken's out parameter is a JsonWebToken. Internally asp.net expects a JwtSecurityToken.

Issue #1300, opened 2019-12-06
Make sure that custom exceptions can be serialized properly


GeoK:
https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.iserializable?view=netframework-4.8

https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229064(v=vs.100)?redirectedfrom=MSDN

https://blog.gurock.com/articles/creating-custom-exceptions-in-dotnet/

PR #1322, opened 2020-02-05
Removed netstandard1.4 from the library targets

### LINKED ISSUE #175: Add back ...xyzException(SerializationInfo info, StreamingContext context) for 4.5.x and desktop
These constructors should be added back but #ifdef'd for FxCore.

### LINKED ISSUE #1310: Remove dead code from ECDsaAdapter
The following method has no references and appears not to be used anywhere. 

Microsoft.IdentityModel.Tokens.ECDsaAdapter.GetKeySize(string curveId)

I am wondering if it really is dead code and maybe consider removing it or explaining how it used.

### LINKED ISSUE #1321: Remove netstandard1.4 from Wilson 6.x
netstandard1.4 is only supported by netcore1.0, which is not an LTS (long term support) release.

mafurman:
Fixes #1321 - _Remove netstandard1.4 from Wilson 6.x_ 
Fixes #175 - _Add back ...xyzException(SerializationInfo info, StreamingContext context_
Fixes  #1310 - _Remove dead code from ECDsaAdapter_


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Exceptions/OpenIdConnectProtocolException.cs
```diff
@@ -26,17 +26,15 @@
...
 {
 
-    #if DESKTOPNET45
-        [Serializable]
-    #endif
-
     /// <summary>
     /// This exception is thrown when an OpenIdConnect protocol handler encounters a protocol error.
     /// </summary>
+    [Serializable]
```
GeoK: w.r.t. Serializable, we need to make sure that the exceptions do serialize properly (tests).
Also, we need to account for additional properties (e.g. SecurityTokenExpiredException has Expires) by overriding GetObjectData #1300 

We should probably add [Serializable] attribute to SignedHttpRequest exceptions (we didn't set it as Serializable didn't work for other assemblies).

src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs
```diff
@@ -207,36 +204,6 @@ private uint GetKeyByteCount(string curveId)
             return keyByteCount;
         }
 
-        /// <summary>
```
GeoK: Thanks :)

src/Microsoft.IdentityModel.ManagedKeyVaultSecurityKey/Microsoft.IdentityModel.ManagedKeyVaultSecurityKey.csproj
```diff
@@ -9,8 +9,8 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <LangVersion>latest</LangVersion>
     <PackageId>Microsoft.IdentityModel.ManagedKeyVaultSecurityKey</PackageId>
-    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net452;netstandard1.4</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netstandard1.4</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net452;netstandard2.0</TargetFrameworks>
```
GeoK: I don't think that current versions of ManagedKeyVaultSecurityKey dependencies support netstandard2.0.
We should probably upgrade, as the new versions support netstandard2.0.

GeoK:
We should remove netstandard1.4 from buildConfiguration.xml too.

mafurman:
@GeoK Good catch; I'll make sure to remove it from there. I completely forgot we also had it in the buildConfiguration.xml.

henrik-me:
not sure I understand how the serialization change is related to the target change. And why "everywhere" should now be serializable where it wouldn't be before. (removal of DESKTOPNET45 constraint)

mafurman:
@henrik-me The _DESKTOPNET45_ symbol was never defined, so there was actually no target where that code would execute (and so these classes were not serializable regardless of target).

The only target where the SerializableAttribute wasn't available for us was netstandard1.4, which is why we can safely include it now. For reference: https://docs.microsoft.com/en-us/dotnet/api/system.serializableattribute?view=netstandard-2.0

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Exceptions/OpenIdConnectProtocolException.cs
mafurman: I created a separate issue to add `[Serializable]` to `SignedHttpRequest` exceptions: #1324

I've also assigned both #1300 and #1324 to myself; I'll issue separate PRs for both.

src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs
mafurman: :grin:

src/Microsoft.IdentityModel.ManagedKeyVaultSecurityKey/Microsoft.IdentityModel.ManagedKeyVaultSecurityKey.csproj
mafurman: I've updated the `MicrosoftAzureKeyVaultVersion `to 3.0.5 in dependencies.props and `SystemNetHttpVersion` to 4.3.4 in both dependencies.props and depedenciesTest.props.

It also seems like [`Microsoft.Azure.KeyVault`](https://www.nuget.org/packages/Microsoft.Azure.KeyVault/) has been replaced by a set of different libraries, so it may be worth switching to those as well...

henrik-me: I suggest updating all references to the latest available. if not at least to the min recommended version (considering security, performance and feature updates etc.)

src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs
```diff
@@ -62,9 +62,6 @@ internal ECDsaAdapter()
                 CreateECDsaFunction = CreateECDsaUsingECParams;
             else
                 CreateECDsaFunction = CreateECDsaUsingCNGKey;
-#elif NETSTANDARD1_4
-            if (SupportsCNGKey())
-                CreateECDsaFunction = CreateECDsaUsingCNGKey;
 #elif DESKTOP
```
brentschmaltz: I opened up this issue to address how we deal with DESKTOP
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1325

src/Microsoft.IdentityModel.ManagedKeyVaultSecurityKey/Microsoft.IdentityModel.ManagedKeyVaultSecurityKey.csproj
mafurman: @henrik-me I'm merging this PR in now, but I created a separate issue to track this: #1326

PR #1328, opened 2020-02-07
check for VSINSTALLDIR and adjust path for msbuild.exe

brentschmaltz:
This PR has two small items.
1. Checking the environment variable for the installation path to Visual Studio and setting the msbuildDir variable from that.
2. Adding '< / clear>' in nuget.config. This ensures that our build will use a controlled set of packages.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -64,6 +64,11 @@ function ClearBaselineFiles($root)
 
 ################################################# Functions ############################################################
 
+if ($env:VSINSTALLDIR)
+{
+    $msbuildDir = $env:VSINSTALLDIR+"\MSBuild\Current\Bin\amd64";
+}
+
 WriteSectionHeader("build.ps1 - parameters");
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -64,6 +64,11 @@ function ClearBaselineFiles($root)
 
 ################################################# Functions ############################################################
 
+if ($env:VSINSTALLDIR)
+{
+    $msbuildDir = $env:VSINSTALLDIR+"\MSBuild\Current\Bin\amd64";
```
GeoK: Not sure why we are appending "amd64" here.
$msbuildDir is defined without it:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/5d1bee56a0612d5187eba7ddb3edf11f05f8d589/build.ps1#L4

brentschmaltz: @GeoK I took the amd64 off and performed a check if the folder exists.

PR #1333, opened 2020-02-08
Removed the net451 desktop target

### LINKED ISSUE #1332: Remove net451 desktop target from Wilson6x
The behavior of the library is the same for both the net45 and net451 targets, so having both is unnecessary.

mafurman:
Fixes #1332 - _Remove net451 desktop target from Wilson6x_


src/Microsoft.IdentityModel.Protocols/Microsoft.IdentityModel.Protocols.csproj
```diff
@@ -20,7 +20,7 @@
     <ProjectReference Include="..\Microsoft.IdentityModel.Tokens\Microsoft.IdentityModel.Tokens.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net45'  Or  '$(TargetFramework)' == 'net451' Or '$(TargetFramework)' == 'net461'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'net45' Or '$(TargetFramework)' == 'net461'">
```
Saibamen: Missing space in `'">`

PR #1336, opened 2020-02-21
Fix exp claim type when using SetDefaultTimesOnTokenCreation

dzsibi:
Due to what I think is small programming error, the value used for the exp claim is promoted to a double, and is added as a floating point value in the payload. I saw no real need for using TimeSpan for this calculation, so I simply multiplied by 60 and now the value is an integer as expected.

Please note that [RFC 7519](https://tools.ietf.org/html/rfc7519#section-4.1.4) specifies that the exp claim must be a NumericDate, which CAN be a non-integer value - so the current implementation is still compliant with the RFC. I would still vote for using an integer value consistently for the three claims (since the extra precision is not required). Feel free to reject this request if you don't agree.


GeoK:
LGTM! 
TotalSeconds [returns double](https://docs.microsoft.com/en-us/dotnet/api/system.timespan.totalseconds?view=netframework-4.8) and TimeSpan.FromMinutes() introduces additional operations that we don't require.

brentschmaltz:
@dzsibi thanks.

PR #1337, opened 2020-02-24
Disable caching of PoP (verifying) signature providers

GeoK:


src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -573,7 +573,16 @@ internal virtual async Task<SecurityKey> ValidateSignatureAsync(JsonWebToken sig
 
             try
             {
-                if (_jwtTokenHandler.ValidateSignature(Encoding.UTF8.GetBytes(signedHttpRequest.EncodedHeader + "." + signedHttpRequest.EncodedPayload), Base64UrlEncoder.DecodeBytes(signedHttpRequest.EncodedSignature), popKey, signedHttpRequest.Alg, new TokenValidationParameters()))
+                // disable caching of PoP (verifying) signature providers
+                var tokenValidationParameters = new TokenValidationParameters()
```
brentschmaltz: We can just put this inline, then the async runtime will not have to remember the local variable.

pmaytak: One comment for keeping as a separate variable is that it might be more readable. (Unless it's inlined without any new lines?)

pmaytak (approved):
One thought - are we sure we don't need caching for these requests; would adding cache limit and expiry be a better long-term solution or unnecessary?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
mafurman: I do agree that keeping it as a separate variable is slightly more readable than having it inline.

If we were to inline it, I think it would better to keep some new lines so that it's easier to see that caching of signature providers is being turned off.

PR #1340, opened 2020-02-29
Adding an OSS notice

mafurman:
This notice contains all OSS that is used by this library.

I've kept the name as the default generated by the tool, but maybe it should be changed to OSS_NOTICE to avoid any confusion with other notice files we have in the library?

This can also be a .txt file if preferable, but I went with HTML because it's a little easier to view with all the licenses it contains.


henrik-me (approved):
:shipit:

pmaytak (approved):

pmaytak:
Yea. Maybe we can rename it to something clearer. I don't remember reading that it has to be named something specific.

GeoK (approved):

PR #1343, opened 2020-02-29
Improve doc of JwtPayload and JwtSecurityToken re claims

### LINKED ISSUE #1342: "audience" behaviour of JwtSecurityToken, JwtPayload confusing
Due to these lines:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/cc6d48f33df400362d261b6272ba5e4215f64e52/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L115-L117

The behaviour can be confusing, and some implementations of refresh tokens that I have seen (and is promulgated in a few tutorials out there) will pass in issuer, audience, etc., along with previous claims (from a deserialized JwtSecurityToken) to the JwtSecurityToken constructor. The resulting new token will have everything overwritten except for the audience claim, which will be a list with one new element added.

To be more clear:
If you call the constructor once with audience = "users", and claims = null, everything is fine;
If you call the constructor a second time with audience = "users" and claims including { "aud", "users" } you will end up with { "aud", ["users, "users'] }. etc.
After many refreshes of a long-lasting chain of tokens, the "aud" list can be very, very long, eventually causing problems (e.g. HTTP header size too big)

I am not saying the implementation is incorrect, and it's clear in the RFC that aud can be a list, but I suggest the documentation should be improved to clarify this behaviour. For example, instead of "a claim will be added", you could write "a claim will be added, appended to any 'aud' claims already in 'claims', if present."

The three places this documentation is copied are:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/cc6d48f33df400362d261b6272ba5e4215f64e52/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L80
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/cc6d48f33df400362d261b6272ba5e4215f64e52/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L67
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/cc6d48f33df400362d261b6272ba5e4215f64e52/src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs#L193

mgillis:
Improves documentation of behaviour of multiple overlapping claims in JwtPayload and JwtSecurityToken. 

Fixes #1342

### DIFF
--- a/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
@@ -63,11 +63,11 @@ namespace System.IdentityModel.Tokens.Jwt
         /// <summary>
         /// Initializes a new instance of the <see cref="JwtPayload"/> class with claims added for each parameter specified. Default string comparer <see cref="StringComparer.Ordinal"/>. 
         /// </summary>
-        /// <param name="issuer">If this value is not null, a { iss, 'issuer' } claim will be added.</param>
-        /// <param name="audience">If this value is not null, a { aud, 'audience' } claim will be added</param>
+        /// <param name="issuer">If this value is not null, a { iss, 'issuer' } claim will be added, overwriting any 'iss' claim in 'claims' if present.</param>
+        /// <param name="audience">If this value is not null, a { aud, 'audience' } claim will be added, appending to any 'aud' claims in 'claims' if present.</param>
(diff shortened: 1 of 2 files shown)


brentschmaltz:
@mgillis thanks for improving these comments. Can you squash and rebase against 'dev', We would be happy to take this PR.

mgillis:
@brentschmaltz done.

(for future reference, I think you can give yourself a "merge and squash" button for PRs if you edit the repository settings? but maybe then you have to squash all PRs, I'm not sure.)

brentschmaltz:
@mgillis thanks. I see a rebase and merge button now, perhaps before you squashed the button was different.

Again, thanks for this work.

PR #1346, opened 2020-03-06
Change SecurityKey.InternalId and SignatureProvider caching logic

### LINKED ISSUE #1219: SecurityKey internal unique keyid's should be based off of key data.
For example:
X509 - thumbprint
RSA - N, E
ECD = TBD

GeoK:
Cache key in InMemoryCryptoProviderCache is now based on SecurityKey's InternalId.
A SecurityKey with empty InternalId will not be cached.

InternalId is now lazy-initialized, so that there is no extra cost when caching is disabled.

Rules for creating InternalId for different key types:
* X509SecurityKey: x5t.
* RsaSecurityKey, SymmetricSecurityKey, JsonWebSecurityKey: base64url
encoded jwk thumbprint.
* EcdsaSecurityKey: base64url encoded jwk thumbprint in runtimes where
it's possible to compute it, otherwise empty string.
* Custom derived keys: base64url encoded jwk thumbprint (if
implemented), otherwise empty string.

fixes: #1219

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
+++ b/src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
@@ -97,10 +97,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// <returns><c>True</c> if <paramref name="signatureProvider"/> should be cached, <c>false</c> otherwise.</returns>
         private bool ShouldCacheSignatureProvider(SignatureProvider signatureProvider)
         {
-            if (signatureProvider.Key.InternalId.Length == 0)
-                return false;
-            else
-                return true;
+            return signatureProvider.Key.InternalId.Length != 0;
         }
 
         /// <summary>
--- a/src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
+++ b/src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
@@ -90,14 +90,14 @@ namespace Microsoft.IdentityModel.Tokens
         }
 
         /// <summary>
-        /// For some security key types, in some runtimes, it's not possible to extract public key material and create <see cref="SecurityKey.InternalId"/>.
+        /// For some security key types, in some runtimes, it's not possible to extract public key material and create an <see cref="SecurityKey.InternalId"/>.
         /// In these cases, <see cref="SecurityKey.InternalId"/> will be an empty string, and these keys should not be cached.
         /// </summary>
         /// <param name="signatureProvider"><see cref="SignatureProvider"/> to be examined.</param>
         /// <returns><c>True</c> if <paramref name="signatureProvider"/> should be cached, <c>false</c> otherwise.</returns>
         private bool ShouldCacheSignatureProvider(SignatureProvider signatureProvider)
         {
-            if (string.IsNullOrEmpty(signatureProvider?.Key?.InternalId))
+            if (signatureProvider.Key.InternalId.Length == 0)
                 return false;
             else
                 return true;
--- a/src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
+++ b/src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
@@ -84,9 +84,23 @@ namespace Microsoft.IdentityModel.Tokens
             return string.Format(CultureInfo.InvariantCulture,
                                  "{0}-{1}-{2}-{3}",
                                  securityKey.GetType(),
-                                 string.IsNullOrEmpty(securityKey.KeyId) ? securityKey.InternalId : securityKey.KeyId,
+                                 securityKey.InternalId,
                                  algorithm,
                                  typeofProvider);
+        }
+
+        /// <summary>
+        /// For some security key types, in some runtimes, it's not possible to extract public key material and create <see cref="SecurityKey.InternalId"/>.
+        /// In these cases, <see cref="SecurityKey.InternalId"/> will be an empty string, and these keys should not be cached.
+        /// </summary>
+        /// <param name="signatureProvider"><see cref="SignatureProvider"/> to be examined.</param>
+        /// <returns><c>True</c> if <paramref name="signatureProvider"/> should be cached, <c>false</c> otherwise.</returns>
+        private bool ShouldCacheSignatureProvider(SignatureProvider signatureProvider)
+        {
+            if (string.IsNullOrEmpty(signatureProvider?.Key?.InternalId))
+                return false;
+            else
+                return true;
         }
 
         /// <summary>
@@ -94,12 +108,17 @@ namespace Microsoft.IdentityModel.Tokens
         /// </summary>
         /// <param name="signatureProvider"><see cref="SignatureProvider"/> to cache.</param>
         /// <exception cref="ArgumentNullException">if signatureProvider is null.</exception>
-        /// <returns>true if the <see cref="SignatureProvider"/> was added, false if the cache already contained the <see cref="SignatureProvider"/></returns>
+        /// <returns>
+        /// <c>true</c> if the <see cref="SignatureProvider"/> was added, <c>false</c> if the cache already contained the <see cref="SignatureProvider"/> or if <see cref="SignatureProvider"/> should not be cached.
+        /// </returns>
         /// <remarks>if the <see cref="SignatureProvider"/> is added <see cref="SignatureProvider.CryptoProviderCache"/> will be set to 'this'.</remarks>
         public override bool TryAdd(SignatureProvider signatureProvider)
         {
             if (signatureProvider == null)
                 throw LogHelper.LogArgumentNullException(nameof(signatureProvider));
+
+            if (!ShouldCacheSignatureProvider(signatureProvider))
+                return false;
 
             var cacheKey = GetCacheKey(signatureProvider);
             if (signatureProvider.WillCreateSignatures)
--- a/src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
+++ b/src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
@@ -374,10 +374,7 @@ namespace Microsoft.IdentityModel.Tokens
 
         private bool CanComputeOctThumbprint()
         {
-            if (string.IsNullOrEmpty(K))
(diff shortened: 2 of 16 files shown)


src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
```diff
@@ -37,11 +37,23 @@ namespace Microsoft.IdentityModel.Tokens
...
+            {
+                try
+                {
+                    return Base64UrlEncoder.Encode(ComputeJwkThumbprint());
+                }
+                catch
+                {
+                    return string.Empty;
+                }
+            });
```
henrik-me: call base? or refactor into a method?

src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
```diff
@@ -37,11 +37,23 @@ namespace Microsoft.IdentityModel.Tokens
...
 
         internal SecurityKey(SecurityKey key)
         {
             _cryptoProviderFactory = key._cryptoProviderFactory;
             KeyId = key.KeyId;
+            _internalId = new Lazy<string>(() =>
+            {
+                try
+                {
+                    return Base64UrlEncoder.Encode(ComputeJwkThumbprint());
```
henrik-me: >ComputeJwkThumbprint() [](start = 51, length = 22)

consider moving out? Assuming this is the place where there could be an expected exception? Do you have expected exceptions from encoding?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -84,23 +84,42 @@ private string GetCacheKeyPrivate(SecurityKey securityKey, string algorithm, str
...
 
+        /// <summary>
+        /// For some security key types, in some runtimes, it's not possible to extract public key material and create <see cref="SecurityKey.InternalId"/>.
+        /// In these cases, <see cref="SecurityKey.InternalId"/> will be an empty string, and these keys should not be cached.
+        /// </summary>
+        /// <param name="signatureProvider"><see cref="SignatureProvider"/> to be examined.</param>
+        /// <returns><c>True</c> if <paramref name="signatureProvider"/> should be cached, <c>false</c> otherwise.</returns>
+        private bool ShouldCacheSignatureProvider(SignatureProvider signatureProvider)
+        {
+            if (string.IsNullOrEmpty(signatureProvider?.Key?.InternalId))
```
brentschmaltz: nit: signatureProvider would not be null for current call graph. Key cannot be null as ctor to SignatureProvider ensures Key is non-null;

src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
brentschmaltz: If we had a method: CanCreateJwkThumbprint(), then if that returns false, return set to string.Empty, we  would not need to throw in known cases.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
GeoK: That's right. I'll remove the null-conditional operators.

src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
GeoK: I'll create a new private method as cryptoProviderFactory differs in these two constructors.

src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
GeoK: That's a good point. I'll add a public virtual method that returns false by default.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -84,23 +84,42 @@ private string GetCacheKeyPrivate(SecurityKey securityKey, string algorithm, str
...
                                  "{0}-{1}-{2}-{3}",
                                  securityKey.GetType(),
-                                 string.IsNullOrEmpty(securityKey.KeyId) ? securityKey.InternalId : securityKey.KeyId,
+                                 securityKey.InternalId,
                                  algorithm,
                                  typeofProvider);
         }
 
+        /// <summary>
+        /// For some security key types, in some runtimes, it's not possible to extract public key material and create <see cref="SecurityKey.InternalId"/>.
```
mafurman: minor grammar nit: create **a**

GeoK: Replaced with "create an", as "create an InternalId" is what will be shown on MSFT docs website.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1346/commits/0f00581f0a75b4972b983006eda9aad40e728414.

src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1346/commits/0f00581f0a75b4972b983006eda9aad40e728414.

src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1346/commits/0f00581f0a75b4972b983006eda9aad40e728414.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -84,23 +84,42 @@ private string GetCacheKeyPrivate(SecurityKey securityKey, string algorithm, str
...
 
+        /// <summary>
+        /// For some security key types, in some runtimes, it's not possible to extract public key material and create an <see cref="SecurityKey.InternalId"/>.
+        /// In these cases, <see cref="SecurityKey.InternalId"/> will be an empty string, and these keys should not be cached.
+        /// </summary>
+        /// <param name="signatureProvider"><see cref="SignatureProvider"/> to be examined.</param>
+        /// <returns><c>True</c> if <paramref name="signatureProvider"/> should be cached, <c>false</c> otherwise.</returns>
+        private bool ShouldCacheSignatureProvider(SignatureProvider signatureProvider)
+        {
+            if (signatureProvider.Key.InternalId.Length == 0)
```
brentschmaltz: could we write: return !string.IsNullOrEmpty(signatureProvider.Key.InternalId);

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -352,6 +372,14 @@ public override byte[] ComputeJwkThumbprint()
                 throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10706, nameof(Kty), string.Join(", ", JsonWebAlgorithmsKeyTypes.EllipticCurve, JsonWebAlgorithmsKeyTypes.RSA, JsonWebAlgorithmsKeyTypes.Octet), nameof(Kty))));
         }
 
+        private bool CanComputeOctThumbprint()
+        {
+            if (string.IsNullOrEmpty(K))
```
brentschmaltz: could we return: !string.IsNullOrEmpty(K);

src/Microsoft.IdentityModel.Tokens/RsaSecurityKey.cs
```diff
@@ -186,22 +186,32 @@ public override int KeySize
...
         /// </summary>
         /// <returns>A JWK thumbprint.</returns>
         /// <remarks>https://tools.ietf.org/html/rfc7638</remarks>
         public override byte[] ComputeJwkThumbprint()
         {
-            if (Parameters.Exponent == null)
-            {
-                if (Rsa == null)
-                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10705, nameof(Rsa)), nameof(Rsa)));
+            var rsaParameters = Parameters;
```
brentschmaltz: nice fix, the previous code change the state of the object.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
GeoK: IsNullOrEmpty checks both for null and that length of the string is 0.
As InternalId can't be null we can skip the extra check.

I can remove the if-branching.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
GeoK: I'll remove all unnecessary if-branches.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1346/commits/303138374134fe6c096ded0ade0b448252f8f695.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
GeoK: Addressed with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1346/commits/303138374134fe6c096ded0ade0b448252f8f695.

brentschmaltz:
:shipit: we will pull this in and then we will make two additional changes:
1. signedhttprequestHandler do not cache signature providers
2. do not add kid to signedhttprequesthandler header, simply require cnf claim.

brentschmaltz (approved):
Opened two issues as follow up.

PR #1353, opened 2020-03-17
Update token caching to use IDistributedCache

mjrousos:
This makes the following updates to WSTrustChannelSecurityTokenProvider:

* Replaces the custom `ISecurityTokenResponseCache` with `IDistributedCache`
* Makes the cache non-public
* Improves some comments and moves local WCF packages to the Wilson directory instead of WsFederation
* Reworks the unit tests to follow the pattern from `TestUtils.TestStub`


src/System.ServiceModel.Federation/WSTrustChannelSecurityTokenProvider.cs
```diff
@@ -90,7 +92,7 @@ public SecurityTokenRequirement SecurityTokenRequirement
         /// <summary>
         /// Gets or sets the cache used for storing issued security tokens.
         /// </summary>
-        public ISecurityTokenResponseCache<WsTrustRequestKey, WsTrustResponse> IssuedTokensCache
+        protected IDistributedCache IssuedTokensCache
```
Tratcher: protected? How do you expect people to actually set it? IDistributedCache isn't very useful unless I can specify where it stores content.

mjrousos: This will be public eventually. @brentschmaltz recommended we keep it internal for now until we've had an opportunity to give due thought to what the public API for managing caches will be. This is just getting us ready for that.

PR #1366, opened 2020-03-30
Simplify caching since a token provider will always use the same RST

mjrousos:
We should add back fuller cache support (including using `IDistributedCache`) in the future
if/when caches can be shared between providers. With the current API surface area, though,
there's no way for a given `WSTrustChannelSecurityTokenProvider` to use more than one RST,
so it will only ever have to cache a single response. Given that, the design that we had here
was more complicated and slower than it needed to be.


RLlane2126:
i signed

PR #1370, opened 2020-04-06
Generate request entropy

mjrousos:
Generate request entropy based on KeyEntropyMode

### DIFF
--- a/test/System.ServiceModel.Federation.Tests/WSTrustChannelSecurityTokenProviderTests.cs
+++ b/test/System.ServiceModel.Federation.Tests/WSTrustChannelSecurityTokenProviderTests.cs
@@ -279,7 +279,7 @@ namespace System.ServiceModel.Federation.Tests
                 {
                     // Default scenario
                     ExpectedProofKey = new byte[0], // Empty proof key means the proof key should be present but the value is unpredictable
-                    TestId = "Symmetric, default entropy"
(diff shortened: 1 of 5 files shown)


mconnew:
LGTM

test/System.ServiceModel.Federation.Tests/WSTrustChannelSecurityTokenProviderTests.cs
```diff
@@ -373,6 +374,7 @@ public static TheoryData<ProofTokenGenerationTheoryData> ProofTokenTheoryData
                 {
                     // Computed key, key size from non-default SecurityAlgorithmSuite
                     RequestSecurityAlgorithmSuite = SecurityAlgorithmSuite.TripleDes,
+                    RequestKeySize = 0,
```
RLlane2126, line 377: y

mjrousos: This is just fixing the test to not include a KeySize parameter in the RST. This is necessary so that we can test that the key size is picked based on the security algorithm suite. If a key size were preset in the RST, that would be used without ever looking at the security algorithm suite.

RLlane2126: merge

PR #1375, opened 2020-04-13
Consider WsTrust version when issuing tokens

### LINKED ISSUE #1371: WsTrustChannelSecuritytokenProvider.GetProoftoken should not assume wstrust1.3
KeyTypes are version specific (2005 != trust1.3), we need to know the version of the version of trust OR allow different version of trust.

This requires some thought through-out the call graph.
We need to establish the version of trust in the beginning of the read and pass a context through the call graph.

We need to also think about the rst/rstr exchange should versions match?

mjrousos:
This updates `WSTrustChannelSecurityTokenProvider` to get message security version from the issuer binding's security binding element or from the issued token parameter's default message security version. One question is where the issuer binding should be retrieved from. This change gets the issuer binding from the incoming security token requirement (which is where it came from in NetFx). Other code paths in WSTrustChannelSecurityTokenProvider (such as `CreateChannelFactory`) get the issuer binding from issued token parameters, though. We should review whether there's a reason to get the binding from these different sources and, if not, standardize on one input.

This PR also fixes a small bug that was introduced when the token provider's "is unexpired" method was changed to "is expired".

Fixes #1371

### DIFF
--- a/src/System.ServiceModel.Federation/WSTrustChannelSecurityTokenProvider.cs
+++ b/src/System.ServiceModel.Federation/WSTrustChannelSecurityTokenProvider.cs
@@ -374,7 +374,7 @@ namespace System.ServiceModel.Federation
                 // with an unspecified expiration. Similarly, a response lifetime without an expiration
                 // time represents an unspecified expiration. In any of these cases, err on the side of
                 // retrieving a new response instead of possibly using an invalid or expired one.
-                return false;
+                return true;
             }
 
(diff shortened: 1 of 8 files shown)


mjrousos:
@brentschmaltz, I've updated this PR to use the IssuerBinding from issued token parameters, as we discussed yesterday. I've also rebased to the latest brentsch/wstrust commit.

Please take a look and let me know if there's anything else we need here before merging.

PR #1377, opened 2020-04-17
Re-enable ClientCredentials tests.

mjrousos:
Qualify SecurityToken and SecurityKey usage to deal with conflicting namespaces


brentschmaltz:
@mjrousos considering that System.ServiceModel.Federation will move to the WCF repo for releasing, I am not in favor of these changes. Instead we should have a separate IdentityComparer for S.S.Federation.

mjrousos:
These changes should just be about namespace-qualifying `SecurityToken` and `SecurityKey` to differentiate between Microsoft.IdentityModel... and System... namespaces. Is that the change you're talking about @brentschmaltz. The diff for this PR is weird because so much has changed in underlying branches. I can rebase this off of the latest to get a more coherent diff.

brentschmaltz:
I understand that this is a namespace only fix, but the PR adds unnecessary confusion that will not be needed long term

mjrousos:
This PR is old enough that it's probably best to just abandon it and open newer ones with the correct test fixes for what the tree looks like now.

PR #1385, opened 2020-04-25
Introduce TokenValidationParameters.TypeValidator to allow implementing custom token type validation

### LINKED ISSUE #1378: Please consider adding a TokenValidationParameters.TypeValidator property
Thanks to the excellent `typ` support introduced in recent versions of Wilson, I've been able to remove the custom validation logic used in ASOS/OpenIddict 1.x/2.x (which was based on a private `token_usage` claim). Due to this change, it's currently hard for me to keep JWT tokens produced by older versions compatible with OpenIddict 3.0. For that, a new API in TVP - very similar to the existing hooks - would help me a lot:

```csharp
public delegate string TypeValidator(string type, SecurityToken securityToken, TokenValidationParameters validationParameters);
```

```csharp
public TypeValidator TypeValidator { get; set; }
```

@brentschmaltz do you think it's something that could be part of the next 6.x release? I'd love to send a PR if it could help.

kevinchalet:
Fixes https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1378.

Please don't merge it before running the tests locally, as there were a few tests/projects I couldn't run on my machine (due to weird strong naming issues)

/cc @brentschmaltz @GeoK @mafurman

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -32,7 +32,6 @@ using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Globalization;
-using System.Linq;
 using System.Security.Claims;
 using System.Security.Cryptography;
 using System.Text;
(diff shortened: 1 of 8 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -319,35 +319,42 @@ private static long ParseTimeValue(JToken jToken, string claimName)
...
+        /// <param name="token">The security token.</param>
         /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
         /// <exception cref="ArgumentNullException">If <paramref name="validationParameters"/> is null or whitespace.</exception>
         /// <exception cref="SecurityTokenInvalidTypeException">If <paramref name="type"/> is null or whitespace and <see cref="TokenValidationParameters.ValidTypes"/> is not null.</exception>
         /// <exception cref="SecurityTokenInvalidTypeException">If <paramref name="type"/> failed to match <see cref="TokenValidationParameters.ValidTypes"/>.</exception>
         /// <remarks>An EXACT match is required. <see cref="StringComparison.Ordinal"/> (case sensitive) is used for comparing <paramref name="type"/> against <see cref="TokenValidationParameters.ValidTypes"/>.</remarks>
-        internal static void ValidateTokenType(string type, TokenValidationParameters validationParameters)
+        internal static string ValidateTokenType(string type, SecurityToken token, TokenValidationParameters validationParameters)
```
kevinchalet: FWIW, this method doesn't have any JWT-specific logic, so it could be moved to `Validators.cs`.

brentschmaltz: I agree it should be in Validators.cs

brentschmaltz:
@kevinchalet if you make the move to Validators.cs, then we can pull in the fix and we can push a new release.

kevinchalet:
> I agree it should be in Validators.cs

Done.

> @kevinchalet if you make the move to Validators.cs, then we can pull in the fix and we can push a new release.

Great, thanks! Before pushing them to NuGet, is there a chance you could push the new packages to the MyGet feed so I can try them in some actual code? (just to be sure... 😄)

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -367,5 +367,51 @@ public static void ValidateTokenReplay(string securityToken, DateTime? expiratio
         {
             ValidateTokenReplay(expirationTime, securityToken, validationParameters);
         }
+
+        /// <summary>
+        /// Validates the type of the token.
+        /// </summary>
+        /// <param name="type">The token type or <c>null</c> if it couldn't be resolved (e.g from the 'typ' header for a JWT).</param>
```
kevinchalet, line 374: FYI, I updated the XML documentation to be less JWT-specific. I also renamed `token` to `securityToken` for consistency and added a null check.

brentschmaltz: nice

brentschmaltz:
@kevinchalet super, I pulled in the PR and will play with it.
This will be in our nightly builds, I will make sure you have access to a signed build.

kevinchalet:
@brentschmaltz I just tested it with the `6.5.2-preview-10429031352` nightly builds and it worked great, but if you don't mind, I have another favor to ask: would it be possible to populate `TokenValidationParameters.TokenType` with the result returned by `Validators.ValidateTokenType`? https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1389

brentschmaltz:
@kevinchalet that's a great idea. Let's get that into next weeks release.

Did you mean put the Type in TokenValidationResult?

PR #1388, opened 2020-04-28
Compare jwt.x5t with signingKey.x5t when using validationParameters.IssuerSigningKeys in JwtSecurityTokenHandler.ResolveIssuerSigningKey

### LINKED ISSUE #1339: JwtSecurityTokenHandler ResolveIssuerSigningKey
JwtSecurityTokenHandler.ResolveIssuerSigningKey doesn't compare jwt.x5t with signingKey.x5t if validationParameters.IssuerSigningKeys is used rather than validationParameters.IssuerSigningKey. Is there a reason for this?

RojaEnnam:
Fixes #1339 

1. Adding functionality to compare X5t of each signingKey in IssuerSigningKeys to JWT token's `X5t.`
2. Adding a new instance of TheoryData in both JwtSecurityTokenHandlerTests and JsonWebTokenHandlerTests to include X509SigningCredential while creating a token, in order for jwtheader to assign token X5t value.

All the test runs are successful.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -972,51 +972,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (jwtToken == null)
                 throw LogHelper.LogArgumentNullException(nameof(jwtToken));
 
-            if (!string.IsNullOrEmpty(jwtToken.Kid))
-            {
-                string kid = jwtToken.Kid;
-                if (validationParameters.IssuerSigningKey != null
-                    && string.Equals(validationParameters.IssuerSigningKey.KeyId, kid, validationParameters.IssuerSigningKey is X509SecurityKey ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
-                    return validationParameters.IssuerSigningKey;
-
-                if (validationParameters.IssuerSigningKeys != null)
-                {
-                    foreach (SecurityKey signingKey in validationParameters.IssuerSigningKeys)
-                    {
-                        if (signingKey != null && string.Equals(signingKey.KeyId, kid, signingKey is X509SecurityKey ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
-                        {
-                            return signingKey;
-                        }
-                    }
-                }
-            }
-
-            if (!string.IsNullOrEmpty(jwtToken.X5t))
-            {
-                string x5t = jwtToken.X5t;
-                if (validationParameters.IssuerSigningKey != null)
-                {
-                    if (string.Equals(validationParameters.IssuerSigningKey.KeyId, x5t, validationParameters.IssuerSigningKey is X509SecurityKey ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
-                        return validationParameters.IssuerSigningKey;
-
-                    var x509Key = validationParameters.IssuerSigningKey as X509SecurityKey;
-                    if (x509Key != null && string.Equals(x509Key.X5t, x5t, StringComparison.OrdinalIgnoreCase))
-                        return validationParameters.IssuerSigningKey;
-                }
-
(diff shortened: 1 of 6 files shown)


brentschmaltz (commented):
This is the right idea. 

We have two handlers that work with JsonTokens.
1. JwtSecurityTokenHandler
2. JsonWebTokenHandler

The code in the JsonWebTokenHandler is fairly correct, see ResolveIssuerSigningKey. We may want to think about moving that code to a central spot, JwtTokenUtilities comes to mind, and create an internal method with a method signature perhaps like: (string kid, string x5t, SecurityToken jwt, TokenValidationParameters validationParameters).
The SecurityToken parameter is for extensibility that is sometimes helpful.

Keep in mind that JwtSecurityTokenHandler.ResolveIssuerSigningKey is virtual so it must be called, the utility method should be called from inside this method. Make sure same errors and exceptions are thrown.

Make sure to add tests, +ve and -ve.
See: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/How-We-Test

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -1852,6 +1852,20 @@ public static TheoryData<JwtTheoryData> ValidateJwsTheoryData
             {
                 return new TheoryData<JwtTheoryData>
                 {
+                     new JwtTheoryData
+                    {
+                        TestId = nameof(Default.AsymmetricJws) + "RequireSignedTokens",
```
brentschmaltz: nit. can you put a "_" between two names in the TestId (in all these test cases).
I know that was how the code was, but it will make reading the results better.

test/Microsoft.IdentityModel.TestUtils/Default.cs
```diff
@@ -96,6 +96,10 @@ public static SigningCredentials AsymmetricSigningCredentials
         {
             get => new SigningCredentials(KeyingMaterial.DefaultX509SigningCreds_2048_RsaSha2_Sha2.Key, KeyingMaterial.DefaultX509SigningCreds_2048_RsaSha2_Sha2.Algorithm, KeyingMaterial.DefaultX509SigningCreds_2048_RsaSha2_Sha2.Digest);
         }
+        public static X509SigningCredentials X509AsymmetricSigningCredentials
```
brentschmaltz, line 100: nit: can you add a new line after the closing brace?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -972,51 +972,7 @@ internal virtual SecurityKey ResolveIssuerSigningKey(JsonWebToken jwtToken, Toke
...
-                        if (signingKey != null && string.Equals(signingKey.KeyId, x5t, signingKey is X509SecurityKey ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
-                        {
-                            return signingKey;
-                        }
-                    }
-                }
-            }
-
-            return null;
+            return JwtTokenUtilities.ResolveTokenSigningKey(jwtToken.Kid, jwtToken.X5t, jwtToken, validationParameters);
```
brentschmaltz, line 975: Nicely done :-)

RojaEnnam: Thanks :)

test/Microsoft.IdentityModel.TestUtils/Default.cs
RojaEnnam: done.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
RojaEnnam: done.

brentschmaltz (approved):

brentschmaltz:
@RojaEnnam i had to rebase on top of dev the push -f.

PR #1389, opened 2020-04-29
Set TokenValidationResult.TokenType with the token type returned by Validators.ValidateTokenType

kevinchalet:
If you approve the approach, I'll try to add a unit test.

/cc @brentschmaltz

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -1146,7 +1146,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     {
                         SecurityToken = jwtToken,
                         ClaimsIdentity = innerTokenValidationResult.ClaimsIdentity,
-                        IsValid = true
+                        IsValid = true,
+                        TokenType = innerTokenValidationResult.TokenType
                     };
                 }
                 else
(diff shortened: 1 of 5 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1146,7 +1146,8 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
                     {
                         SecurityToken = jwtToken,
                         ClaimsIdentity = innerTokenValidationResult.ClaimsIdentity,
-                        IsValid = true
+                        IsValid = true,
+                        TokenType = innerTokenValidationResult.TokenType
```
kevinchalet, line 1150: Not 100% sure about this one. Although I would expect the two to match in most cases, Wilson seems to make a difference between the `typ` of the JWE container and the `typ` of the inner JWT token, so maybe it should be another property, like `InnerTokenType`?

brentschmaltz: I think it should be as you mentioned the inner, as the JWE is just a wrapper without any info.

brentschmaltz:
@kevinchalet I agree with your approach, feel free to add the tests.
fyi, we will probably put another release out next week.

kevinchalet:
I updated the existing type validation tests to validate `TokenValidationResult.TokenType`.

I also opened https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1397 to track a possible improvement for token creation, that would make the story nicer and consistent with this PR.

PR #1394, opened 2020-05-01
Remove System.Net.Http package reference for .NET Framework (4.5/4.6)

tpersettmicrosoft:
The out-of-band [System.Net.Http nuget](https://www.nuget.org/packages/System.Net.Http/) package is known for causing issues with .NET 4.5/4.6 deployments, and the general guidance is to [avoid using it when targeting the .NET Framework](https://github.com/dotnet/runtime/issues/26131#issuecomment-388199236). 

The proposed change replaces the nuget reference with a framework reference for .NET 4.5 and .NET 4.6.


brentschmaltz:
@tpersettmicrosoft @Tratcher thanks.

PR #1395, opened 2020-05-01
Serialize custom exceptions

### LINKED ISSUE #1300: Make sure that custom exceptions can be serialized properly

keegan-caruso:
Fixes #1300


src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenInvalidSigningKeyException.cs
```diff
@@ -73,6 +73,8 @@ public SecurityTokenInvalidSigningKeyException(string message, Exception inner)
         protected SecurityTokenInvalidSigningKeyException(SerializationInfo info, StreamingContext context)
             : base(info, context)
         {
+            //TODO - investigate if SigningKey type should be serialized
```
keegan-caruso, line 76: For the serializer to work with the SigningKey, the abstract base class (SecurityKey) would also need to be marked as serializable as well as the types on SecurityKey. The list of types that would need to be marked as serializable would be:

- CryptoProviderCache
- CryptoProviderFactory
- InMemoryCryptoProviderCache
- SecurityKey

keegan-caruso:
For the serializer to work with the SigningKey, the abstract base class (SecurityKey) would also need to be marked as serializable as well as the types on SecurityKey. The list of types that would need to be marked as serializable would be:

- CryptoProviderCache
- CryptoProviderFactory
- InMemoryCryptoProviderCache
- SecurityKey

Is this the approach that we want to follow?

keegan-caruso:
In opensource/json there are a few exceptions with `HAVE_BINARY_EXCEPTION_SERIALIZATION` ifdef blocks. Do we want to roll the changes into opensource/json/*Exceptions with this change set?

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenExpiredException.cs
```diff
@@ -73,6 +73,15 @@ public SecurityTokenExpiredException(string message, Exception inner)
...
         {
+            Expires = (DateTime)info.GetValue(nameof(Expires), typeof(DateTime));
+        }
+
+        /// <inheritdoc/>
+        public override void GetObjectData(SerializationInfo info, StreamingContext context)
+        {
+            base.GetObjectData(info, context);
+
+            info.AddValue(nameof(Expires), Expires, typeof(DateTime));
```
brentschmaltz: Since this could be derived from, a users could set the same value.
If we use the fully qualified name such as: GetType().FullName + "." + nameof(Expires), the collusions would be reduced.
Also this can throw, if the type cast fails on info.GetValue.

Have a look at: https://referencesource.microsoft.com/#mscorlib/system/security/claims/ClaimsIdentity.cs,8437a543724aff81 where fully qualified names are used. Also the SerializationInfo enumerator is queried for values. See the Deserialize method.

keegan-caruso: Thanks for the link! I will refer to it.

How should we handle throwing an exception if the type cast fails on GetValue? We can check the type based on the enumerator, but silently failing doesn't seem right here either.

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenInvalidSigningKeyException.cs
henrik-me: > //TODO - investigate if SigningKey type should be serialized [](start = 11, length = 61)

The answer to this should be why we are doing this work for serialization. Are we transferring these over the wire?

henrik-me:
Those seems to not be custom exceptions? I would not expect those changes to be here. Ideally we should not make any changes in the opensource area (will make it very hard to update).

---
In reply to: [622439825](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1395#issuecomment-622439825) [](ancestors = 622439825)

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenExpiredException.cs
henrik-me: @brentschmaltz @GeoK perspectives?

henrik-me: IMO, we should not do anything, we should let the original exception be thrown as described here.
https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.serializationinfo.getvalue?view=netcore-3.1

@keegan-caruso  Can you elaborate on when/how this situation will occur?

keegan-caruso: @henrik-me should only happen in error / not during normal execution. If a user derived from the class and mistakenly created a conflict with a different type for example.

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenInvalidSigningKeyException.cs
mafurman: I don't think so. I think this would be used by users for writing exceptions to files.

mafurman:
> 
> 
> Those seems to not be custom exceptions? I would not expect those changes to be here. Ideally we should not make any changes in the opensource area (will make it very hard to update).
> 
> In reply to: [622439825](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1395#issuecomment-622439825) [](ancestors = 622439825)

I agree with this. I believe 'custom exceptions' is solely referring to exceptions that belong to our library.

Anything in 'opensource' isn't our code and we shouldn't modify it (we learned the hard way that this creates very odd issues, haha :sweat_smile:).

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenInvalidSigningKeyException.cs
brentschmaltz: I agree we should remove that TODO.
The coding of this PR was in sync with the desired result.
We serialize the internal data of 'IdentityModel' exceptions such as Expires on SecurityTokenExpiredException.

I am OK with not serializing SecurityKey.

brentschmaltz (approved):
I think it is fine not to serialize the SecurityKey on SecurityTokenInvalidSigningKeyException

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenExpiredException.cs
brentschmaltz: we should add documentation i opened: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1408

GeoK (approved):

PR #1398, opened 2020-05-04
Remove the JwtTokenUtilities.ValidateTokenType() method

kevinchalet:
Most likely due to a bad merge, https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/f5628e689e97b80fba3efe81fc33903076c97ad5 added back `JwtTokenUtilities.ValidateTokenType()` (but without a `System.Linq` using, so the solution no longer compiles).

This PR removes this method (for good, hopefully 😅)

/cc @brentschmaltz


brentschmaltz:
@kevinchalet hard to tell :-). Thanks.

PR #1399, opened 2020-05-04
Adding Switch to control if ALL keys are  tried when validating signature

### LINKED ISSUE #1128: Add configurable switch to control if ALL keys are tried when validating signature
If a user has populated TokenValidationParameters with a large number of keys, then we want to give them a way to only check the signature if a key was matched.
If there was no kid in the header, then this is considered not matched.

Adding a new property to TokenValidationParameters may work.

RojaEnnam:
Fixes #1128 

1. Currently we try all keys to validate signature by default if a Issuer key is not resolved to match token's 'kid' or 'x5t'
2. Adding _TryAllIssuerSigningKeys_ switch to control if all keys need to be tried when validating signature.

All test runs are successful.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -1234,30 +1234,34 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 throw new SecurityTokenInvalidSignatureException(TokenLogMessages.IDX10508, e);
             }
 
-            foreach (var key in keys)
+            if(keys!=null)
             {
-                try
+                foreach (var key in keys)
                 {
-                    if (ValidateSignature(encodedBytes, signatureBytes, key, jwtToken.Alg, validationParameters))
+                    try
                     {
-                        LogHelper.LogInformation(TokenLogMessages.IDX10242, token);
-                        jwtToken.SigningKey = key;
-                        return jwtToken;
-                    };
-                }
-                catch (Exception ex)
-                {
-                    exceptionStrings.AppendLine(ex.ToString());
+                        if (ValidateSignature(encodedBytes, signatureBytes, key, jwtToken.Alg, validationParameters))
+                        {
+                            LogHelper.LogInformation(TokenLogMessages.IDX10242, token);
+                            jwtToken.SigningKey = key;
+                            return jwtToken;
+                        };
+                    }
+                    catch (Exception ex)
+                    {
+                        exceptionStrings.AppendLine(ex.ToString());
+                    }
+
+                    if (key != null)
+                    {
+                        keysAttempted.AppendLine(key.ToString() + " , KeyId: " + key.KeyId);
+                        if (kidExists && !kidMatched && key.KeyId != null)
+                            kidMatched = jwtToken.Kid.Equals(key.KeyId, key is X509SecurityKey ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal);
+                    }
                 }
 
-                if (key != null)
-                {
-                    keysAttempted.AppendLine(key.ToString() + " , KeyId: " + key.KeyId);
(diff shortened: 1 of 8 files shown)


src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -901,8 +901,8 @@ protected virtual JwtSecurityToken ValidateSignature(string token, TokenValidati
                     return jwtToken;
             }
 
-            bool kidMatched = false;           
-            IEnumerable<SecurityKey> keys = null;
+            bool kidMatched = false;
+            IEnumerable<SecurityKey> keys = null;
```
brentschmaltz: please remove trailing space (it looks like it was there before :-) )

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1234,30 +1234,34 @@ private JsonWebToken ValidateSignature(string token, TokenValidationParameters v
                 throw new SecurityTokenInvalidSignatureException(TokenLogMessages.IDX10508, e);
             }
 
-            foreach (var key in keys)
+            if(keys!=null)
```
brentschmaltz: nit: spacing needed after 'if'

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1209,8 +1209,8 @@ private JsonWebToken ValidateSignature(string token, TokenValidationParameters v
                     keys = new List<SecurityKey> { key };
                 }
             }
-
-            if (keys == null)
+
```
brentschmaltz: nit: please remove whitespace

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
RojaEnnam: done

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: done

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: done.

brentschmaltz:
@RojaEnnam can you add a simple test?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1234,30 +1234,34 @@ private JsonWebToken ValidateSignature(string token, TokenValidationParameters v
                 throw new SecurityTokenInvalidSignatureException(TokenLogMessages.IDX10508, e);
             }
 
-            foreach (var key in keys)
+            if (keys!=null)
```
brentschmaltz, line 1237: nice, our current code would throw a null ref exception

brentschmaltz (approved):
nice.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: yes

GeoK:
@RojaEnnam, @brentschmaltz - Sorry, I'm a bit late here. Would we need the same behavior for Saml token handlers?

brentschmaltz:
@GeoK yes we should have, we should open another issue.
@RamonPage In addition we should investigate if we could factor out the logic into a common method.

PR #1400, opened 2020-05-04
Do not initialize JwtSecurityTokenHandler.ValidateToken in case of an exception

### LINKED ISSUE #1290: JwtSecurityTokenHandler.ValidateToken can set the out parameter in case of exception in ValidateToken
```cs
validatedToken = ValidateSignature(token, validationParameters);
return ValidateTokenPayload(validatedToken as JwtSecurityToken, validationParameters);
```

It's unexpected to have the validatedToken set in case of exception.

RojaEnnam:
Fixes #1290 

1. Adding additional variables to store token from _ValidateSignature_  temporarily and initialize out variable after _ValidateTokenPayload_ is done.

All test runs are successful.

### DIFF
--- a/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
@@ -705,8 +705,12 @@ namespace System.IdentityModel.Tokens.Jwt
                 throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(TokenLogMessages.IDX10209, token.Length, MaximumTokenSizeInBytes)));
 
             var tokenParts = token.Split(new char[] { '.' }, JwtConstants.MaxJwtSegmentCount + 1);
+
             if (tokenParts.Length != JwtConstants.JwsSegmentCount && tokenParts.Length != JwtConstants.JweSegmentCount)
                 throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX12741, token)));
+
+            ClaimsPrincipal claimsPrincipal = null;
+            SecurityToken signatureValidatedToken = null;
 
             if (tokenParts.Length == JwtConstants.JweSegmentCount)
             {
@@ -714,15 +718,17 @@ namespace System.IdentityModel.Tokens.Jwt
                 var decryptedJwt = DecryptToken(jwtToken, validationParameters);
                 var innerToken = ValidateSignature(decryptedJwt, validationParameters);
                 jwtToken.InnerToken = innerToken;
-                validatedToken = jwtToken;
-                return ValidateTokenPayload(innerToken, validationParameters);
+                signatureValidatedToken = jwtToken;
+                claimsPrincipal = ValidateTokenPayload(innerToken, validationParameters);
             }
             else
             {
-                validatedToken = ValidateSignature(token, validationParameters);
-                return ValidateTokenPayload(validatedToken as JwtSecurityToken, validationParameters);
-                    
+                signatureValidatedToken = ValidateSignature(token, validationParameters);
+                claimsPrincipal = ValidateTokenPayload(signatureValidatedToken as JwtSecurityToken, validationParameters);
             }
+
+            validatedToken = signatureValidatedToken;
+            return claimsPrincipal;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -705,6 +705,8 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
                 throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(TokenLogMessages.IDX10209, token.Length, MaximumTokenSizeInBytes)));
 
             var tokenParts = token.Split(new char[] { '.' }, JwtConstants.MaxJwtSegmentCount + 1);
+            ClaimsPrincipal claimsPrincipal = null;
```
GeoK: No need for redundant assignment. I'd move these variables closer to where they are used.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -714,14 +716,17 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
                 var decryptedJwt = DecryptToken(jwtToken, validationParameters);
                 var innerToken = ValidateSignature(decryptedJwt, validationParameters);
                 jwtToken.InnerToken = innerToken;
-                validatedToken = jwtToken;
-                return ValidateTokenPayload(innerToken, validationParameters);
+                signatureValidatedToken = jwtToken;
+                claimsPrincipal = ValidateTokenPayload(innerToken, validationParameters);
+                validatedToken = signatureValidatedToken;
```
GeoK: You could set validatedToken and return claimsPrincipal after the if/else statement.

keegan-caruso (approved):

brentschmaltz:
@RojaEnnam we have to have a couple of tests.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
```diff
@@ -40,6 +40,75 @@ namespace System.IdentityModel.Tokens.Jwt.Tests
 {
     public class JwtSecurityTokenHandlerTests
     {
+        [Theory, MemberData(nameof(VerifyValidatedTokenTheoryData))]
+        public void VerifyValidatedToken(JwtTheoryData theoryData)
```
GeoK: minor comment: Can we extend our current _ValidateToken_ tests and add asserts there?
I'm not sure if we need a new method and a theory data set just to test a new model (validated token is null in case of an exception). We are also adding some other tests here, like expecting IDX10501, which are not the goal of this test.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
```diff
@@ -40,6 +40,75 @@ namespace System.IdentityModel.Tokens.Jwt.Tests
...
+        public static TheoryData<JwtTheoryData> VerifyValidatedTokenTheoryData
+        {
+            get
+            {
+                var theoryData = new TheoryData<JwtTheoryData>();
+                var handler = new JwtSecurityTokenHandler();
+
+                ClaimsIdentity claimsIdentity = new ClaimsIdentity(ClaimSets.DefaultClaimsIdentity);
+                var validationParameters = Default.AsymmetricSignTokenValidationParameters;
+                validationParameters.ValidateIssuer = true;
```
GeoK: Why is this assignment important? Isn't this true by default?

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
RojaEnnam: Sure. That would cut out lot of code for a minor validation. 
I was wondering if a separate test case is needed for a new change and so created one.

Right, we don't need IDX10501. It is covered by other tests.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
RojaEnnam: Yes, ValidateIssuer is true by default.

I added it with the below reference in mind. Assuming that the below is set to make the test case behavior easy to understand.
Eg: ExpectedException for JwtTheory data is NoExceptionExpected by default but it is still assigned in few places.

GeoK: Got it, thanks.

GeoK (approved):

brentschmaltz:
squashed and rebased on dev

PR #1405, opened 2020-05-07
Using TryAllIssuerSigningKeys in saml/saml2 token handlers 

### LINKED ISSUE #1407: Add support for TokenValidatonParameters.TryAllIssuerSigningKeys in saml/saml2 token handlers
We introduced TryAllIssuerSigningKeys  as a property of TokenValidatonParameters to control if all keys need to be tried while validating signature.

Only JsonWebToken and JwtSecurityToken handlers support this currently.
We also need to add it in SAML/SAML2 token handlers.

RojaEnnam:
Fixes #1407 

1. Adding usage of _TryAllIssuerSigningKeys_ in saml/saml2 token handlers 
2. Moving the logic of gathering keys to validate signature into jwttoken and samltoken utilities.

All test runs are successful.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -372,5 +372,63 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
             return null;
         }
+
+        /// <summary>
+        /// Returns all <see cref="SecurityKey"/> to use when validating the signature of a token.
+        /// </summary>
+        /// <param name="token">The <see cref="string"/> representation of the token that is being validated.</param>
+        /// <param name="kid">The <see cref="string"/> kid field of the token being validated</param>
+        /// <param name="x5t">The <see cref="string"/> x5t field of the token being validated</param>
+        /// <param name="jwtToken">The <see cref="SecurityToken"/> that is being validated.</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <param name="kidMatched">A <see cref="bool"/> to represent if a a issuer signing key matched with token kid or x5t</param>
+        /// <returns>Returns all <see cref="SecurityKey"/> to use for signature validation.</returns>
+        internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(string token, string kid, string x5t, SecurityToken jwtToken, TokenValidationParameters validationParameters,out bool kidMatched)
+        {
+            IEnumerable<SecurityKey> keys = null;
+            kidMatched = false;
+
+            if (validationParameters.IssuerSigningKeyResolver != null)
+            {
+                keys = validationParameters.IssuerSigningKeyResolver(token, jwtToken, kid, validationParameters);
+            }
+            else
+            {
+                SecurityKey key = ResolveTokenSigningKey(kid, x5t, jwtToken, validationParameters);
+
+                if (key != null)
+                {
+                    kidMatched = true;
+                    keys = new List<SecurityKey> { key };
+                }
+                else
+                {
+                    kidMatched = false;
+                    if (validationParameters.TryAllIssuerSigningKeys)
+                    {
+                        keys = GetAllSigningKeys(token, validationParameters);
+                    }
+                }
+            }
+            return keys;
+        }
+
+        /// <summary>
+        /// Returns all <see cref="SecurityKey"/> provided in validationParameters.
+        /// </summary>
+        /// <param name="token">The <see cref="string"/> representation of the token that is being validated.</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <returns>Returns all <see cref="SecurityKey"/> provided in validationParameters.</returns>
+        internal static IEnumerable<SecurityKey> GetAllSigningKeys(string token, TokenValidationParameters validationParameters)
+        {
+            LogHelper.LogInformation(TokenLogMessages.IDX10243);
+            if (validationParameters.IssuerSigningKey != null)
+                yield return validationParameters.IssuerSigningKey;
+
+            if (validationParameters.IssuerSigningKeys != null)
+                foreach (SecurityKey key in validationParameters.IssuerSigningKeys)
+                    yield return key;
+        }
     }
 }
+
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -383,14 +383,13 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
         /// <param name="kidMatched">A <see cref="bool"/> to represent if a a issuer signing key matched with token kid or x5t</param>
         /// <returns>Returns all <see cref="SecurityKey"/> to use for signature validation.</returns>
-        internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(string token, string kid, string x5t, SecurityToken jwtToken, TokenValidationParameters validationParameters,out bool kidMatched)
+        internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(string token, string kid, string x5t, SecurityToken jwtToken, TokenValidationParameters validationParameters, out bool kidMatched)
         {
-            IEnumerable<SecurityKey> keys = null;
             kidMatched = false;
 
             if (validationParameters.IssuerSigningKeyResolver != null)
             {
-                keys = validationParameters.IssuerSigningKeyResolver(token, jwtToken, kid, validationParameters);
+                return validationParameters.IssuerSigningKeyResolver(token, jwtToken, kid, validationParameters);
             }
             else
             {
@@ -399,35 +398,18 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 if (key != null)
                 {
                     kidMatched = true;
-                    keys = new List<SecurityKey> { key };
+                    return new List<SecurityKey> { key };
                 }
                 else
                 {
                     kidMatched = false;
                     if (validationParameters.TryAllIssuerSigningKeys)
                     {
-                        keys = GetAllSigningKeys(token, validationParameters);
+                        return TokenUtilities.GetAllSigningKeys(validationParameters);
                     }
                 }
             }
-            return keys;
-        }
-
-        /// <summary>
-        /// Returns all <see cref="SecurityKey"/> provided in validationParameters.
-        /// </summary>
-        /// <param name="token">The <see cref="string"/> representation of the token that is being validated.</param>
-        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
-        /// <returns>Returns all <see cref="SecurityKey"/> provided in validationParameters.</returns>
-        internal static IEnumerable<SecurityKey> GetAllSigningKeys(string token, TokenValidationParameters validationParameters)
-        {
-            LogHelper.LogInformation(TokenLogMessages.IDX10243);
-            if (validationParameters.IssuerSigningKey != null)
-                yield return validationParameters.IssuerSigningKey;
-
-            if (validationParameters.IssuerSigningKeys != null)
-                foreach (SecurityKey key in validationParameters.IssuerSigningKeys)
-                    yield return key;
+            return null;
         }
     }
 }
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
@@ -547,17 +547,6 @@ namespace Microsoft.IdentityModel.Tokens.Saml
             }
         }
 
-        private IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParameters validationParameters)
-        {
-            LogHelper.LogInformation(TokenLogMessages.IDX10243);
-            if (validationParameters.IssuerSigningKey != null)
-                yield return validationParameters.IssuerSigningKey;
-
-            if (validationParameters.IssuerSigningKeys != null)
-                foreach (SecurityKey securityKey in validationParameters.IssuerSigningKeys)
-                    yield return securityKey;
-        }
-
         /// <summary>
         /// Creates claims from a <see cref="SamlAttributeStatement"/>.
         /// </summary>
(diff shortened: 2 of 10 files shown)


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -384,51 +385,35 @@ private Saml2SecurityToken ValidateSignature(Saml2SecurityToken samlToken, strin
...
             // keep track of exceptions thrown, keys that were tried
             var exceptionStrings = new StringBuilder();
             var keysAttempted = new StringBuilder();
             bool canMatchKey = samlToken.Assertion.Signature.KeyInfo != null;
-            foreach (var key in keys)
+
+            if (keys != null)
             {
-                try
+                foreach (var key in keys)
```
AndersAbel, line 397: I think this loop would be better to put in Signature.Verify() and pass the key collection instead of one key at a time. The logic to try multiple keys is needed when other signed XML blocks are verified too (complete SAML2P responses, metadata). If that logic is in the security token handler, it will have to be recreated in other places.

henrik-me:
@RojaEnnam, which issues does this PR belong to?

RojaEnnam:
@henrik-me It is related to [#1128](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1128). After the issue was closed we realized saml/saml2 token handlers also need this.

henrik-me:
> @henrik-me It is related to [#1128](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1128). After the issue was closed we realized saml/saml2 token handlers also need this.

Thanks @RojaEnnam for the explanation.

Seems like the issue should be re-opened then or a new issue created to track this work and ensure it's properly described in the release notes?

brentschmaltz:
@AndersAbel @RojaEnnam @henrik-me I agree that other places may need to have logic where possibly different keys are available.

We may benefit from some middle layer abstraction that handles this.
I do not think that Signature.Verify is the right place as an XML Signature Element has only one signature ALL the possible key clauses should reference the same key.

RojaEnnam:
> > @henrik-me It is related to [#1128](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1128). After the issue was closed we realized saml/saml2 token handlers also need this.
> 
> Thanks @RojaEnnam for the explanation.
> 
> Seems like the issue should be re-opened then or a new issue created to track this work and ensure it's properly described in the release notes?

Agreed. I created #1407 to track this.

RojaEnnam:
> @AndersAbel @RojaEnnam @henrik-me I agree that other places may need to have logic where possibly different keys are available.
> 
> We may benefit from some middle layer abstraction that handles this.
> I do not think that Signature.Verify is the right place as an XML Signature Element has only one signature ALL the possible key clauses should reference the same key.

Currently, we use kid/key Match result to throw appropriate exceptions in case of failure. If we move the key iteration to Signature.Verify, we could move kid/key Match logic to some place like GetAllSigningKeys. Although this could be done, I agree with @brentschmaltz regarding one signature in XML Signature Element.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -372,5 +372,63 @@ internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, Secur
...
+        /// Returns all <see cref="SecurityKey"/> to use when validating the signature of a token.
+        /// </summary>
+        /// <param name="token">The <see cref="string"/> representation of the token that is being validated.</param>
+        /// <param name="kid">The <see cref="string"/> kid field of the token being validated</param>
+        /// <param name="x5t">The <see cref="string"/> x5t field of the token being validated</param>
+        /// <param name="jwtToken">The <see cref="SecurityToken"/> that is being validated.</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <param name="kidMatched">A <see cref="bool"/> to represent if a a issuer signing key matched with token kid or x5t</param>
+        /// <returns>Returns all <see cref="SecurityKey"/> to use for signature validation.</returns>
+        internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(string token, string kid, string x5t, SecurityToken jwtToken, TokenValidationParameters validationParameters,out bool kidMatched)
```
brentschmaltz: nit: can you put a space between the comma 'validationParameters,out bool'

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -372,5 +372,63 @@ internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, Secur
...
+        {
+            IEnumerable<SecurityKey> keys = null;
```
brentschmaltz: we don't need this variable as we can just return from the places we set 'keys'
the statement at the end could be - return null.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -1048,53 +1029,34 @@ private SamlSecurityToken ValidateSignature(SamlSecurityToken samlToken, string
...
-                keys = GetAllSigningKeys(validationParameters);
-            }
+            keys = ResolveIssuerSigningKey(token, samlToken, validationParameters, out keyMatched);
 
             // keep track of exceptions thrown, keys that were tried
             var exceptionStrings = new StringBuilder();
             var keysAttempted = new StringBuilder();
             bool canMatchKey = samlToken.Assertion.Signature.KeyInfo != null;
-            foreach (var key in keys)
+            if (keys != null)
```
brentschmaltz, line 1038: nice catch of null-ref

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <summary>
+        /// Returns a <see cref="SecurityKey"/> to use when validating the signature of a token.
+        /// </summary>
+        /// <param name="tokenKeyInfo">The <see cref="KeyInfo"/> field of the token being validated</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <returns>Returns a <see cref="SecurityKey"/> to use for signature validation.</returns>
+        /// <remarks>If key fails to resolve, then null is returned</remarks>
+        internal static SecurityKey ResolveTokenSigningKey(KeyInfo tokenKeyInfo, TokenValidationParameters validationParameters)
+        {
+
```
brentschmaltz: nit: remove extra line

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -0,0 +1,124 @@
...
+            if (validationParameters.IssuerSigningKey != null && tokenKeyInfo.MatchesKey(validationParameters.IssuerSigningKey))
```
brentschmaltz, line 53: while we are here, could you fix: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/6a08f9d2dad68acc5a6238f7a5879b6170d6f0fc/src/Microsoft.IdentityModel.Xml/KeyInfo.cs#L147

Matches should return false if 'key' is null, then you don't need the two checks.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// </summary>
+        /// <param name="token">The <see cref="string"/> representation of the token that is being validated.</param>
+        /// <param name="samlToken">The <see cref="SecurityToken"/> that is being validated.</param>
+        /// <param name="tokenKeyInfo">The <see cref="KeyInfo"/> field of the token being validated</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <param name="keyMatched">A <see cref="bool"/> to represent if a a issuer signing key matched with token kid or x5t</param>
+        /// <returns>Returns all <see cref="SecurityKey"/> to use for signature validation.</returns>
+        internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(string token,SecurityToken samlToken, KeyInfo tokenKeyInfo, TokenValidationParameters validationParameters, out bool keyMatched)
+        {
+            IEnumerable<SecurityKey> keys = null;
```
brentschmaltz: same note about just returning and not having this variable.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <summary>
+        /// Returns all <see cref="SecurityKey"/> to use when validating the signature of a token.
+        /// </summary>
+        /// <param name="token">The <see cref="string"/> representation of the token that is being validated.</param>
+        /// <param name="samlToken">The <see cref="SecurityToken"/> that is being validated.</param>
+        /// <param name="tokenKeyInfo">The <see cref="KeyInfo"/> field of the token being validated</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <param name="keyMatched">A <see cref="bool"/> to represent if a a issuer signing key matched with token kid or x5t</param>
+        /// <returns>Returns all <see cref="SecurityKey"/> to use for signature validation.</returns>
+        internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(string token,SecurityToken samlToken, KeyInfo tokenKeyInfo, TokenValidationParameters validationParameters, out bool keyMatched)
```
brentschmaltz: nit: space after comma between: 'token,SecurityToken'

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -372,5 +372,63 @@ internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, Secur
...
+            return keys;
+        }
+
+        /// <summary>
+        /// Returns all <see cref="SecurityKey"/> provided in validationParameters.
+        /// </summary>
+        /// <param name="token">The <see cref="string"/> representation of the token that is being validated.</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <returns>Returns all <see cref="SecurityKey"/> provided in validationParameters.</returns>
+        internal static IEnumerable<SecurityKey> GetAllSigningKeys(string token, TokenValidationParameters validationParameters)
```
brentschmaltz: could be shared between saml and jwt, the method is internal and we never use the token.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1216,17 +1184,26 @@ protected virtual void ValidateTokenReplay(DateTime? expires, string securityTok
         /// <param name="token">The <see cref="string"/> representation of the token that is being validated.</param>
         /// <param name="jwtToken">The <see cref="JwtSecurityToken"/> that is being validated.</param>
         /// <param name="validationParameters">A <see cref="TokenValidationParameters"/>  required for validation.</param>
+        /// <param name="kidMatched">A <see cref="bool"/> to represent if a issuer signing key matched with token kid or x5t</param>
         /// <returns>Returns a <see cref="SecurityKey"/> to use for signature validation.</returns>
         /// <remarks>If key fails to resolve, then null is returned</remarks>
-        protected virtual SecurityKey ResolveIssuerSigningKey(string token, JwtSecurityToken jwtToken, TokenValidationParameters validationParameters)
+        protected virtual IEnumerable<SecurityKey> ResolveIssuerSigningKey(string token, JwtSecurityToken jwtToken, TokenValidationParameters validationParameters,out bool kidMatched)
```
brentschmaltz: nit: space after comma

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1216,17 +1184,26 @@ protected virtual void ValidateTokenReplay(DateTime? expires, string securityTok
...
             if (validationParameters == null)
                 throw LogHelper.LogArgumentNullException(nameof(validationParameters));
 
             if (jwtToken == null)
                 throw LogHelper.LogArgumentNullException(nameof(jwtToken));
 
-            return JwtTokenUtilities.ResolveTokenSigningKey(jwtToken.Header.Kid, jwtToken.Header.X5t, jwtToken, validationParameters);
+            string kid = null;
+            string x5t = null;
+            if (jwtToken.Header != null)
```
brentschmaltz: I don't believe jwtToken.Header can be null.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
RojaEnnam: Done.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
RojaEnnam: Done.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
RojaEnnam: Done.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
RojaEnnam: Done.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
RojaEnnam: Adding it to TokenUtilities.cs

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
RojaEnnam: Done.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
RojaEnnam: Okie..Thanks for letting me know.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
RojaEnnam: Done.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
RojaEnnam: Done.

src/Microsoft.IdentityModel.Xml/KeyInfo.cs
```diff
@@ -145,6 +145,9 @@ public override int GetHashCode()
         /// </summary>
         internal bool MatchesKey(SecurityKey key)
         {
+            if (key == null)
+                return false;
+
             if (key is X509SecurityKey x509SecurityKey)
```
RojaEnnam, line 151: not sure if this line throws null ref exception. That's why adding code to return false if key is null.

brentschmaltz:
@RojaEnnam nice.

PR #1409, opened 2020-05-11
Validate signed http requests claims if present

### LINKED ISSUE #1393: SignedHttpRequest values should be validated if they are found.
make these optional by default

keegan-caruso:
Log a warning for skipped validation of a present claim for signed http requests

Fixes #1393

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
+++ b/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
@@ -545,28 +545,61 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
             if (signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ReplayValidatorAsync != null)
                 await signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ReplayValidatorAsync(jwtSignedHttpRequest, signedHttpRequestValidationContext, cancellationToken).ConfigureAwait(false);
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -545,30 +545,42 @@ protected virtual async Task<SecurityToken> ValidateSignedHttpRequestPayloadAsyn
...
+        /// <summary>
+        /// Determine if validation of a claim should happen.
+        /// </summary>
+        /// <param name="jwtSignedHttpRequest">The request being considered to validate the claim on.</param>
+        /// <param name="validateClaim">Force validation to always occur.</param>
+        /// <param name="claimName">The name of the claim to validate.</param>
+        /// <returns>Whether the given claim should be validated</returns>
+        internal virtual bool ShouldValidate(JsonWebToken jwtSignedHttpRequest, bool validateClaim, string claimName)
+        {
+            return validateClaim || (jwtSignedHttpRequest.TryGetClaim(claimName, out var claimValue) && claimValue != null);
```
brentschmaltz: This is a change that may affect users where different clients are running different versions.
Suppose an older client is sending a value that is causing issues, the receiver currently has it turned off (validateClaim false). That client will succeed. With this change, that client will now start to fail.

keegan-caruso:
Closing as there are still questions around how we want to implement this.

brentschmaltz:
@keegan-caruso i agree it makes sense to close this for now as we aren't sure what 'done' looks like.

PR #1411, opened 2020-05-12
Add introspection support in OpenIdConnectConfiguration

### LINKED ISSUE #1209: OpenIdConnectConfiguration doesn't have TokenIntrospectionEndpoint
OpenIdConnectConfiguration doesn't have TokenIntrospectionEndpoint property.

It is very useful property in access token introspection scenario. 

```
var response = await client.IntrospectTokenAsync(new TokenIntrospectionRequest {
    Address = "https://demo.identityserver.io/connect/introspect", // <- here
    ClientId = "id",
    ClientSecret = "secret",
    Token = accessToken
});
```

kevinchalet:
Fixes https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1209.

@brentschmaltz is it something that could be included in the next release? 😃


test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectMetadata.json
```diff
@@ -15,6 +15,9 @@
     "id_token_encryption_alg_values_supported" : ["RSA1_5", "A256KW"],
     "id_token_encryption_enc_values_supported" : ["A128CBC-HS256","A256CBC-HS512"],
     "id_token_signing_alg_values_supported": [ "RS256" ],
+    "introspection_endpoint" : "https://login.windows.net/d062b2b0-9aca-4ff7-b32a-ba47231a4002/oauth2/introspect",
```
kevinchalet, line 18: Note: Azure AD doesn't support introspection so these values are fake.

brentschmaltz:
@kevinchalet we can fit this in, thanks.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdProviderMetadataNames.cs
```diff
@@ -51,6 +51,9 @@ public static class OpenIdProviderMetadataNames
         public const string IdTokenEncryptionAlgValuesSupported = "id_token_encryption_alg_values_supported";
         public const string IdTokenEncryptionEncValuesSupported = "id_token_encryption_enc_values_supported";
         public const string IdTokenSigningAlgValuesSupported = "id_token_signing_alg_values_supported";
+        public const string IntrospectionEndpoint = "introspection_endpoint";
```
brentschmaltz, line 54: @kevinchalet is there a spec we can reference for these values?

kevinchalet: Yep, these values are defined in the OAuth 2.0 Authorization Server Metadata specification (RFC8414): https://tools.ietf.org/html/rfc8414#section-2

brentschmaltz: @kevinchalet super thanks!

PR #1413, opened 2020-05-13
Allow users to specify the acceptable algorithms for crypto operations

### LINKED ISSUE #1224: Introduce API's to allow user to restrict cryptographic algorithms.
Best practice suggests users should be able to easily control acceptable algorithms.
While users can use extensibility, it should designed as an inbound policy.

See: https://tools.ietf.org/html/draft-ietf-oauth-jwt-bcp-06#section-3.1

keegan-caruso:
Add a collection and a delegate to token validation parameters to allow
a user to specify which cyrpto algorithms are considered acceptable.

fixes #1224

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -636,6 +636,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                 try
                 {
+                    Validators.ValidateAlgorithm(jwtToken.Enc, key, jwtToken, validationParameters);
                     decryptedTokenBytes = DecryptToken(jwtToken, cryptoProviderFactory, key);
                     decryptionSucceeded = true;
                     break;
@@ -1213,7 +1214,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 {
                     try
                     {
-                        if (ValidateSignature(encodedBytes, signatureBytes, key, jwtToken.Alg, validationParameters))
+                        if (ValidateSignature(encodedBytes, signatureBytes, key, jwtToken.Alg, jwtToken, validationParameters))
                         {
                             LogHelper.LogInformation(TokenLogMessages.IDX10242, token);
                             jwtToken.SigningKey = key;
@@ -1258,9 +1259,10 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// <param name="signature">Signature to compare against.</param>
         /// <param name="key"><See cref="SecurityKey"/> to use.</param>
         /// <param name="algorithm">Crypto algorithm to use.</param>
+        /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
         /// <param name="validationParameters">Priority will be given to <see cref="TokenValidationParameters.CryptoProviderFactory"/> over <see cref="SecurityKey.CryptoProviderFactory"/>.</param>
         /// <returns>'true' if signature is valid.</returns>
-        internal bool ValidateSignature(byte[] encodedBytes, byte[] signature, SecurityKey key, string algorithm, TokenValidationParameters validationParameters)
+        internal bool ValidateSignature(byte[] encodedBytes, byte[] signature, SecurityKey key, string algorithm, SecurityToken securityToken, TokenValidationParameters validationParameters)
         {
             var cryptoProviderFactory = validationParameters.CryptoProviderFactory ?? key.CryptoProviderFactory;
             if (!cryptoProviderFactory.IsSupportedAlgorithm(algorithm, key))
@@ -1268,6 +1270,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 LogHelper.LogInformation(LogMessages.IDX14000, algorithm, key);
                 return false;
             }
+
+            Validators.ValidateAlgorithm(algorithm, key, securityToken, validationParameters);
 
             var signatureProvider = cryptoProviderFactory.CreateForVerifying(key, algorithm);
             if (signatureProvider == null)
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
@@ -420,6 +420,8 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
                 {
                     try
                     {
+                        Validators.ValidateAlgorithm(samlToken.Assertion.Signature.SignedInfo.SignatureMethod, key, samlToken, validationParameters);
+
                         samlToken.Assertion.Signature.Verify(key, validationParameters.CryptoProviderFactory ?? key.CryptoProviderFactory);
                         LogHelper.LogInformation(TokenLogMessages.IDX10242, token);
                         samlToken.SigningKey = key;
--- a/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
@@ -197,6 +197,8 @@ namespace Microsoft.IdentityModel.Tokens
         public const string IDX10693 = "IDX10693: RSACryptoServiceProvider doesn't support the RSASSA-PSS signature algorithm. The list of supported algorithms is available here: https://aka.ms/IdentityModel/supported-algorithms";
         public const string IDX10694 = "IDX10694: JsonWebKeyConverter threw attempting to convert JsonWebKey: '{0}'. Exception: '{1}'.";
         public const string IDX10695 = "IDX10695: Unable to create a JsonWebKey from an ECDsa object. Required ECParameters structure is not supported by .NET Framework < 4.7.";
+        public const string IDX10696 = "IDX10696: The algorithm '{0}' is not in the user-defined accepted list of algorithms.";
+        public const string IDX10697 = "IDX10697: The user defined 'Delegate' AlgorithmValidator specified on TokenValidationParameters returned false when validating Algorithm: '{0}', SecurityKey: '{1}'.";
 
         // security keys
         public const string IDX10700 = "IDX10700: {0} is unable to use 'rsaParameters'. {1} is null.";
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
@@ -34,6 +34,16 @@ using Microsoft.IdentityModel.Logging;
 namespace Microsoft.IdentityModel.Tokens
 {
     /// <summary>
+    /// Definition for AlgorithmValidator
+    /// </summary>
+    /// <param name="algorithm">The algorithm to validate.</param>
+    /// <param name="securityKey">The <see cref="SecurityKey"/> that signed the <see cref="SecurityToken"/>.</param>
+    /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
+    /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+    /// <returns><c>true</c> if the algorithm is considered valid</returns>
+    public delegate bool AlgorithmValidator(string algorithm, SecurityKey securityKey, SecurityToken securityToken, TokenValidationParameters validationParameters);
+
+    /// <summary>
     /// Definition for AudienceValidator.
     /// </summary>
     /// <param name="audiences">The audiences found in the <see cref="SecurityToken"/>.</param>
@@ -156,6 +166,7 @@ namespace Microsoft.IdentityModel.Tokens
             if (other == null)
                 throw LogHelper.LogExceptionMessage(new ArgumentNullException(nameof(other)));
 
+            AlgorithmValidator = other.AlgorithmValidator;
             ActorValidationParameters = other.ActorValidationParameters?.Clone();
             AudienceValidator = other.AudienceValidator;
             _authenticationType = other._authenticationType;
@@ -192,6 +203,7 @@ namespace Microsoft.IdentityModel.Tokens
             ValidateIssuerSigningKey = other.ValidateIssuerSigningKey;
             ValidateLifetime = other.ValidateLifetime;
             ValidateTokenReplay = other.ValidateTokenReplay;
+            ValidAlgorithms = other.ValidAlgorithms;
             ValidAudience = other.ValidAudience;
             ValidAudiences = other.ValidAudiences;
             ValidIssuer = other.ValidIssuer;
@@ -221,6 +233,15 @@ namespace Microsoft.IdentityModel.Tokens
         /// Gets or sets <see cref="TokenValidationParameters"/>.
         /// </summary>
         public TokenValidationParameters ActorValidationParameters { get; set; }
+
+        /// <summary>
+        /// Gets or sets a delegate used to validate the cryptographic algorithm used.
+        /// </summary>
+        /// <remarks>
+        /// If set, this delegate will validate the cryptographic algorithm used and
+        /// the algorithm will not be checked against <see cref="ValidAlgorithms"/>.
+        /// </remarks>
+        public AlgorithmValidator AlgorithmValidator { get; set; }
 
         /// <summary>
         /// Gets or sets a delegate that will be used to validate the audience.
@@ -604,6 +625,14 @@ namespace Microsoft.IdentityModel.Tokens
         /// </remarks>
         [DefaultValue(false)]
         public bool ValidateTokenReplay { get; set; }
+
+        /// <summary>
+        /// Gets or sets the valid algorithms for cryptographic operations.
+        /// </summary>
+        /// <remarks>
+        /// If set to a non-empty collection, only the algorithms listed will be considered valid.
+        /// </remarks>
+        public IEnumerable<string> ValidAlgorithms { get; set; }
 
         /// <summary>
         /// Gets or sets a string that represents a valid audience that will be used to check against the token's audience.
--- a/src/Microsoft.IdentityModel.Tokens/Validators.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validators.cs
@@ -39,6 +39,40 @@ namespace Microsoft.IdentityModel.Tokens
     public static class Validators
     {
         /// <summary>
+        /// Validates if a given algorithm for a <see cref="SecurityKey"/> is valid.
+        /// </summary>
+        /// <param name="algorithm">The algorithm to be validated.</param>
+        /// <param name="securityKey">The <see cref="SecurityKey"/> that signed the <see cref="SecurityToken"/>.</param>
+        /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
+        /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+        public static void ValidateAlgorithm(string algorithm, SecurityKey securityKey, SecurityToken securityToken, TokenValidationParameters validationParameters)
+        {
+            if (validationParameters == null)
+                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
+
+            if (validationParameters.AlgorithmValidator != null)
+            {
+                if (!validationParameters.AlgorithmValidator(algorithm, securityKey, securityToken, validationParameters))
+                {
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidAlgorithmException(LogHelper.FormatInvariant(LogMessages.IDX10697, algorithm, securityKey))
+                    {
+                        InvalidAlgorithm = algorithm,
+                    });
+                }
+
+                return;
+            }
+
+            if (validationParameters.ValidAlgorithms != null && validationParameters.ValidAlgorithms.Any() && !validationParameters.ValidAlgorithms.Contains(algorithm, StringComparer.Ordinal))
+            {
+                throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidAlgorithmException(LogHelper.FormatInvariant(LogMessages.IDX10696, algorithm))
+                {
+                    InvalidAlgorithm = algorithm,
+                });
+            }
+        }
+
+        /// <summary>
         /// Determines if the audiences found in a <see cref="SecurityToken"/> are valid.
         /// </summary>
         /// <param name="audiences">The audiences found in the <see cref="SecurityToken"/>.</param>
--- a/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
+++ b/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
@@ -1853,7 +1853,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
             {
                 return new TheoryData<JwtTheoryData>
                 {
-                     new JwtTheoryData
+                    new JwtTheoryData
                     {
                         TestId = nameof(Default.AsymmetricJws) + "_" + "RequireSignedTokens",
                         Token = Default.AsymmetricJws,
@@ -1867,7 +1867,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
                             ValidateLifetime = false,
                         }
                     },
-                     new JwtTheoryData
+                    new JwtTheoryData
                     {
                         TestId = nameof(Default.SymmetricJws) + "_" + "RequireSignedTokens",
                         Token = Default.SymmetricJws,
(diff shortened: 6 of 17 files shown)


src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -33,6 +33,14 @@
 
 namespace Microsoft.IdentityModel.Tokens
 {
+    /// <summary>
+    /// Definition for AcceptedAlgorithmValidator
+    /// </summary>
+    /// <param name="algorithm">The algorithm to validate.</param>
+    /// <param name="securityKey">THe securityKey to validate.</param>
+    /// <returns><c>true</c> if the algorithm is considered valid</returns>
+    public delegate bool AcceptedAlgorithmValidator(string algorithm, SecurityKey securityKey);
```
brentschmaltz: while not specifically required, we usually pass additional parameters when validating as this gives users the opportunity to apply additional logic.

i am suggesting the parameters are:
string algorithm, SecurityKey securityKey, string purpose, SecurityToken securityToken, TokenValidationParameters validationParameters.

purpose - describes the cryptographic operation: encrypting, decrypting, creating signature, validating signature, ...
validationParameters - TokenValidationParameters is derivable and contains a PropertyBag that has helped users in corner cases.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -197,6 +197,8 @@ internal static class LogMessages
         public const string IDX10693 = "IDX10693: RSACryptoServiceProvider doesn't support the RSASSA-PSS signature algorithm. The list of supported algorithms is available here: https://aka.ms/IdentityModel/supported-algorithms";
         public const string IDX10694 = "IDX10694: JsonWebKeyConverter threw attempting to convert JsonWebKey: '{0}'. Exception: '{1}'.";
         public const string IDX10695 = "IDX10695: Unable to create a JsonWebKey from an ECDsa object. Required ECParameters structure is not supported by .NET Framework < 4.7.";
+        public const string IDX10696 = "IDX10696: The algorithm '{0}' is not in the user defined accepted list of algorithms.";
+        public const string IDX10697 = "IDX10697: The user defined 'Delegate' specified on TokenValidationParameters returned false when validating the algorithm for Algorithm: '{0}', SecurityKey: '{1}'.";
```
brentschmaltz: would be helpful to include the name of the Delegate: AcceptedAlgorithmValidator.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -634,6 +634,18 @@ public string DecryptToken(JsonWebToken jwtToken, TokenValidationParameters vali
                     continue;
                 }
 
+                if (validationParameters.AcceptedAlgorithms != null && validationParameters.AcceptedAlgorithms.Any() && !validationParameters.AcceptedAlgorithms.Contains(jwtToken.Enc, StringComparer.OrdinalIgnoreCase))
```
brentschmaltz: we will need to add this to other call graphs: Saml1, Saml2, SignedHttpRequests, perhaps this logic should live in Validators.cs.

the pattern is to apply the delegate, check if 'validation' should be applied, then check.
some of the 'checks' have specific bools that control behavior. ValidateAudience for example.
in this case, that is probably not required as by default an empty collection implies 'all'.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1296,6 +1308,18 @@ internal bool ValidateSignature(byte[] encodedBytes, byte[] signature, SecurityK
                 return false;
             }
 
+            if (validationParameters.AcceptedAlgorithms != null && validationParameters.AcceptedAlgorithms.Any() && !validationParameters.AcceptedAlgorithms.Contains(algorithm, StringComparer.OrdinalIgnoreCase))
```
brentschmaltz: https://tools.ietf.org/html/rfc7515#section-4.1.1 
'alg' is case-insensitive

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
keegan-caruso: I agree. This is also a potential minor improvement to some of the other log messages in this class

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
keegan-caruso: @brentschmaltz Good point. This brings up a few questions when combined with changing the signature on the delegate

Should the SignedHttpRequestValidationParameters object also include restrictions on crypto algorithms? It would be confusing to reuse the TVP used for Access Token validation.

For SAML, there is not a SecurityToken available at the time that the signature is verified over the SignedInfo. Just a SecurityKey, TVP, and algorithm.

Is it better to broaden the inputs of the delegate with the trade off being that the logic will need to be able to handle potential null values?

keegan-caruso: Following up, the saml token just needs to be brought into the context. and the SHR validation is out of scope for this task.

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -38,6 +38,34 @@ namespace Microsoft.IdentityModel.Tokens
...
     {
+        /// <summary>
+        /// Validates if a given algorithm for a <see cref="SecurityKey"/> is acceptable
+        /// </summary>
+        /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <param name="algorithm">The algorithm being considered if it is valid</param>
+        /// <param name="securityKey">The <see cref="SecurityKey"/> that signed the <see cref="SecurityToken"/>.</param>
+        /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
+        /// <returns><c>true</c> if the algorithm is acceptable</returns>
+        public static bool ValidateAcceptedAlgorithms(TokenValidationParameters validationParameters, string algorithm, SecurityKey securityKey, SecurityToken securityToken)
```
brentschmaltz: can we have parameters in a similar order as the other validators and matching delegate?
Parameters of interest first - algorithm, securitykey
SecurityToken, 
TokenValidationParameters last,

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -37,9 +37,11 @@ namespace Microsoft.IdentityModel.Tokens
     /// Definition for AcceptedAlgorithmValidator
     /// </summary>
     /// <param name="algorithm">The algorithm to validate.</param>
-    /// <param name="securityKey">THe securityKey to validate.</param>
+    /// <param name="securityKey">The <see cref="SecurityKey"/> that signed the <see cref="SecurityToken"/>.</param>
+    /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
+    /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
     /// <returns><c>true</c> if the algorithm is considered valid</returns>
-    public delegate bool AcceptedAlgorithmValidator(string algorithm, SecurityKey securityKey);
+    public delegate bool AcceptedAlgorithmValidator(string algorithm, SecurityKey securityKey, SecurityToken securityToken, TokenValidationParameters validationParameters);
```
brentschmaltz, line 44: should have parameters in similar order as other delegates, see other comment.

src/Microsoft.IdentityModel.Tokens/Validators.cs
keegan-caruso: good idea

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -197,6 +197,8 @@ internal static class LogMessages
         public const string IDX10693 = "IDX10693: RSACryptoServiceProvider doesn't support the RSASSA-PSS signature algorithm. The list of supported algorithms is available here: https://aka.ms/IdentityModel/supported-algorithms";
         public const string IDX10694 = "IDX10694: JsonWebKeyConverter threw attempting to convert JsonWebKey: '{0}'. Exception: '{1}'.";
         public const string IDX10695 = "IDX10695: Unable to create a JsonWebKey from an ECDsa object. Required ECParameters structure is not supported by .NET Framework < 4.7.";
+        public const string IDX10696 = "IDX10696: The algorithm '{0}' is not in the user defined accepted list of algorithms.";
```
mafurman: nit: "user-defined"

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -197,6 +197,8 @@ internal static class LogMessages
...
+        public const string IDX10697 = "IDX10697: The user defined 'Delegate' AcceptedAlgorithmValidator specified on TokenValidationParameters returned false when validating the algorithm for Algorithm: '{0}', SecurityKey: '{1}'.";
```
mafurman: nit: Do we need to capitalize 'Delegate' and 'Algorithm' here?

Also, can the sentence be simplified by removing 'for Algorithm' at the end?

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -33,6 +33,16 @@
 
 namespace Microsoft.IdentityModel.Tokens
 {
+    /// <summary>
+    /// Definition for AcceptedAlgorithmValidator
```
mafurman: Can this just be called AlgorithmValidator? That seems to better fit the naming model of the other delegates.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -217,6 +229,19 @@ public TokenValidationParameters()
             ValidateTokenReplay = false;
         }
 
+        /// <summary>
+        /// Gets or sets the accepted algorithms for cryptographic operations.
+        /// </summary>
+        /// <remarks>
+        /// If set to a non-empty collection, only the algorithms listed will be considered valid.
+        /// </remarks>
```
mafurman: Could this be renamed to ValidAlgorithms to better follow the naming conventions of the other properties on TokenValidationParameters?

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -38,6 +38,34 @@ namespace Microsoft.IdentityModel.Tokens
...
     public static class Validators
     {
+        /// <summary>
+        /// Validates if a given algorithm for a <see cref="SecurityKey"/> is acceptable
+        /// </summary>
+        /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <param name="algorithm">The algorithm being considered if it is valid</param>
+        /// <param name="securityKey">The <see cref="SecurityKey"/> that signed the <see cref="SecurityToken"/>.</param>
+        /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
+        /// <returns><c>true</c> if the algorithm is acceptable</returns>
```
mafurman: nit: I think we generally try to put '.' at the end of comments for consistency.

src/Microsoft.IdentityModel.Xml/LogMessages.cs
```diff
@@ -73,6 +73,7 @@ internal static class LogMessages
         internal const string IDX30211 = "IDX30211: The TransfromFactory does not support the canonicalizing transform: '{0}'.";
         internal const string IDX30212 = "IDX30212: Unable to verify Signature as Signature.SignedInfo is null.";
         internal const string IDX30213 = "IDX30213: The CryptoProviderFactory: '{0}', CreateForSigning returned null for key: '{1}', SignatureMethod: '{2}'.";
+        internal const string IDX30214 = "IDX30214: User supplied validation of the algorithm '{0}' failed.";
```
mafurman: Could this be reworded as "Algorithm '{0}' did not match any of the valid algorithms supplied by the user. Valid algorithms: {1}.'

mafurman (approved):

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
keegan-caruso: This is a fair question. I was trying to keep this consistent with IDX10505, IDX10506, IDX10508, and IDX10509. Will take another look

src/Microsoft.IdentityModel.Xml/LogMessages.cs
keegan-caruso: This is a good idea. It would make debugging easier.

keegan-caruso: Okay, can't quite use that. This capture the failure from Validators.ValidateAcceptedAlgorithms. Which also include the case for the validation failing from the delegate.

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -38,6 +38,34 @@ namespace Microsoft.IdentityModel.Tokens
...
+        public static bool ValidateAcceptedAlgorithms(TokenValidationParameters validationParameters, string algorithm, SecurityKey securityKey, SecurityToken securityToken)
+        {
+            if (validationParameters == null)
+                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
+
+            if (validationParameters.AcceptedAlgorithms != null && validationParameters.AcceptedAlgorithms.Any() && !validationParameters.AcceptedAlgorithms.Contains(algorithm, StringComparer.Ordinal))
+            {
+                LogHelper.LogInformation(LogMessages.IDX10696, algorithm);
+                return false;
```
GeoK: This model is inconsistent with the existing validator model. Other methods throw exceptions in case of unsuccessful validations.

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -38,6 +38,34 @@ namespace Microsoft.IdentityModel.Tokens
...
+            }
+
+            if (validationParameters.AcceptedAlgorithmValidator != null && !validationParameters.AcceptedAlgorithmValidator(algorithm, securityKey, securityToken, validationParameters))
```
GeoK: AcceptedAlgorithmValidator delegate should have priority over validation of AcceptedAlgorithms.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -197,6 +197,8 @@ internal static class LogMessages
         public const string IDX10693 = "IDX10693: RSACryptoServiceProvider doesn't support the RSASSA-PSS signature algorithm. The list of supported algorithms is available here: https://aka.ms/IdentityModel/supported-algorithms";
         public const string IDX10694 = "IDX10694: JsonWebKeyConverter threw attempting to convert JsonWebKey: '{0}'. Exception: '{1}'.";
         public const string IDX10695 = "IDX10695: Unable to create a JsonWebKey from an ECDsa object. Required ECParameters structure is not supported by .NET Framework < 4.7.";
+        public const string IDX10696 = "IDX10696: The algorithm '{0}' is not in the user-defined accepted list of algorithms.";
+        public const string IDX10697 = "IDX10697: The user defined 'Delegate' AcceptedAlgorithmValidator specified on TokenValidationParameters returned false when validating Algorithm: '{0}', SecurityKey: '{1}'.";
```
keegan-caruso: missed this in the refactor

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenInvalidAlgorithm.cs
```diff
@@ -0,0 +1,110 @@
...
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// This exception is thrown when a cryptographic algorithm is invalid.
+    /// </summary>
+    [Serializable]
+    public class SecurityTokenInvalidAlgorithm : SecurityTokenValidationException
```
mafurman: We typically put "Exception" at the end of an Exception class name.

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -38,6 +38,36 @@ namespace Microsoft.IdentityModel.Tokens
...
     {
+        /// <summary>
+        /// Validates if a given algorithm for a <see cref="SecurityKey"/> is valid.
+        /// </summary>
+        /// <param name="algorithm">The algorithm being considered if it is valid</param>
+        /// <param name="securityKey">The <see cref="SecurityKey"/> that signed the <see cref="SecurityToken"/>.</param>
+        /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
+        /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <returns><c>true</c> if the algorithm is acceptable.</returns>
+        public static void ValidateAlgorithms(string algorithm, SecurityKey securityKey, SecurityToken securityToken, TokenValidationParameters validationParameters)
```
mafurman: I think it may make more sense to call this "ValidateAlgorithm" (singular).

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -38,6 +38,36 @@ namespace Microsoft.IdentityModel.Tokens
     /// </summary>
     public static class Validators
     {
+        /// <summary>
+        /// Validates if a given algorithm for a <see cref="SecurityKey"/> is valid.
+        /// </summary>
+        /// <param name="algorithm">The algorithm being considered if it is valid</param>
```
mafurman: Maybe rephrase this comment as "the algorithm to be validated".

src/Microsoft.IdentityModel.Xml/Signature.cs
```diff
@@ -91,29 +91,34 @@ public SignedInfo SignedInfo
...
         /// </summary>
         /// <param name="key">the <see cref="SecurityKey"/> to use for cryptographic operations.</param>
+        /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
+        /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
         /// <exception cref="ArgumentNullException"> if <paramref name="key"/> is null.</exception>
         /// <exception cref="XmlValidationException"> if <see cref="SignedInfo"/> null.</exception>
         /// <exception cref="XmlValidationException"> if <see cref="SignedInfo.SignatureMethod"/> is not supported.</exception>
         /// <exception cref="XmlValidationException"> if signature does not validate.</exception>
-        public void Verify(SecurityKey key)
+        public void Verify(SecurityKey key, SecurityToken securityToken, TokenValidationParameters validationParameters)
```
mafurman: Modifying a public method like this would be considered a breaking change. I would add this method in as an overload (same for the other Verify() method as well).

src/Microsoft.IdentityModel.Tokens/Validators.cs
keegan-caruso: Good idea. It is only checking one

src/Microsoft.IdentityModel.Tokens/Validators.cs
keegan-caruso: that is more clear, good idea.

src/Microsoft.IdentityModel.Xml/Signature.cs
keegan-caruso: Very good point! How should input validation look for the potential null arguments?

Validate algorithm only if the Security token and TVP are present?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -420,7 +420,7 @@ private Saml2SecurityToken ValidateSignature(Saml2SecurityToken samlToken, strin
                 {
                     try
                     {
-                        samlToken.Assertion.Signature.Verify(key, validationParameters.CryptoProviderFactory ?? key.CryptoProviderFactory);
+                        samlToken.Assertion.Signature.Verify(key, validationParameters.CryptoProviderFactory ?? key.CryptoProviderFactory, samlToken, validationParameters);
```
brentschmaltz: We could just verify that the key and algorithm are valid before calling Signature.Verify.
The reasoning is that an XML Signature may or not be associated with a token. 
TokenValidationParameters may not the right object conceptually.

We could write in the SamlSecurityTokenHandler and Saml2SecurityTokenHandler: 
                        Validators.ValidateAlgorithm(samlToken.Assertion.Signature.SignedInfo.SignatureMethod, key, samlToken, validationParameters);

@keegan-caruso thoughts?

keegan-caruso: We can do that, and it would simplify the changes a bit. I think that would also address the issue you brought up about the conceptual mapping?

brentschmaltz: @keegan-caruso super....  ping me when the changes are in....

keegan-caruso: @brentschmaltz changes are in and build is passing

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -38,6 +38,36 @@ namespace Microsoft.IdentityModel.Tokens
...
     public static class Validators
     {
+        /// <summary>
+        /// Validates if a given algorithm for a <see cref="SecurityKey"/> is valid.
+        /// </summary>
+        /// <param name="algorithm">The algorithm to be validated.</param>
+        /// <param name="securityKey">The <see cref="SecurityKey"/> that signed the <see cref="SecurityToken"/>.</param>
+        /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
+        /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <returns><c>true</c> if the algorithm is acceptable.</returns>
```
GeoK: this is no longer true.

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -38,6 +38,36 @@ namespace Microsoft.IdentityModel.Tokens
...
+        public static void ValidateAlgorithm(string algorithm, SecurityKey securityKey, SecurityToken securityToken, TokenValidationParameters validationParameters)
+        {
+            if (validationParameters == null)
+                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
+
+            if (validationParameters.AlgorithmValidator != null && !validationParameters.AlgorithmValidator(algorithm, securityKey, securityToken, validationParameters))
```
GeoK: If the delegate succeeded, we shouldn't proceed with the default validation.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1269,6 +1271,8 @@ internal bool ValidateSignature(byte[] encodedBytes, byte[] signature, SecurityK
                 return false;
             }
 
+            Validators.ValidateAlgorithm(algorithm, key, securityToken, validationParameters);
```
GeoK, line 1274: Would it make sense to move this call to the private ValidateSignature method?
I'm not sure what we are gaining by putting the check in this method.
...or do we need a SecurityKey on the delegate?

GeoK (commented):
I think that JwtSecurityTokenHandler should be covered too.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
```diff
@@ -1286,6 +1286,153 @@ public static TheoryData<ValidateSignedHttpRequestTheoryData> ValidateSignedHttp
...
+                            ValidateU = false,
+                        },
+                        ValidationParameters =  new TokenValidationParameters()
+                        {
+                            IssuerSigningKey = KeyingMaterial.RsaSigningCreds_2048_Public.Key,
+                            ValidIssuer = Default.Issuer,
+                            ValidAudience = Default.Audience,
+                            TokenDecryptionKey = KeyingMaterial.DefaultSymmetricEncryptingCreds_Aes128_Sha2.Key,
+                            ValidateLifetime = false,
+                            AlgorithmValidator = ValidationDelegates.AlgorithmValidatorBuilder(true),
```
GeoK, line 1431: Nice!

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
keegan-caruso: We currently expect the SecurityKey. The thought being that it could be useful for making a validation decision in the delegate.

GeoK: What would be a potential use case where a SecurityKey is expected/useful? The alg/enc is already defined in a token and TVP holds a collection of valid algorithms.

keegan-caruso: @brentschmaltz can correct me if I am wrong here.

But the biggest one that comes to mind is AdditionalData on JsonWebKey typed arguments

brentschmaltz: @GeoK @keegan-caruso passing the SecurityKey we are about to use can be helpful as the 'SecurityKey' could be user specified as we have a delegate that the user could have coded into the call graph.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -1939,7 +1939,196 @@ public static TheoryData<JwtTheoryData> ValidateJwsTheoryData
...
+                        {
+                            ValidateIssuerSigningKey = true,
+                            RequireSignedTokens = false,
+                            IssuerSigningKey = Default.SymmetricSigningKey,
+                            ValidateIssuer = false,
+                            ValidateAudience = false,
+                            ValidateLifetime = false,
+                            AlgorithmValidator = (alg, key, token, validationParameters) => throw new TestException("expected error validating algorithm")
+                        },
+                        ExpectedException = ExpectedException.SecurityTokenInvalidSignatureException("IDX10511")
```
brentschmaltz, line 2034: info: while I would not ask for this change in this PR, it's best to end the expected substring including a ':'. It has to do with the was we check the exception message [contains](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/8ca7558ca2d5c7a27d108d6ecd7c9babf2b30a95/test/Microsoft.IdentityModel.TestUtils/ExpectedException.cs#L140).  I updated our wiki [How We Test](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/How-We-Test) to include this idea.

brentschmaltz:
@keegan-caruso can you squash, rebase on top of dev and commit?

brentschmaltz (approved):
please squash, rebase on top of dev and commit.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
GeoK: I'm not pushing back on the change, just trying to identity a use case.
If the idea is to be future-proof, and provide more extensibility, I'm ok with the change.

PR #1414, opened 2020-05-14
Support for SecurityTokenDescriptor.Claims in JwtSecurity/Saml/Saml2 Tokens

### LINKED ISSUE #1155: Add support for creating Saml and Saml2 tokens with TokenDescriptor.Claims

RojaEnnam:
Fixes #1155 

1. Adding method to convert dictionary of claims to list of Claim objects.
2. Merging both SecurityTokenDescriptor.Claims and SecurityTokenDescriptor.Subject.Claims and using them in creating token.

All test runs are successful.


src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -382,7 +383,7 @@ public virtual string CreateEncodedJwt(string issuer, string audience, ClaimsIde
         /// <exception cref="ArgumentException">If 'expires' &lt;= 'notBefore'.</exception>
         public virtual string CreateEncodedJwt(string issuer, string audience, ClaimsIdentity subject, DateTime? notBefore, DateTime? expires, DateTime? issuedAt, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials)
```
RojaEnnam, line 384: Do we need to add an input parameter in CreateEncodedJwt to provide an option to pass in SecurityTokenDescriptor.Claims?

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +58,77 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
+                return ClaimValueTypes.DateTime;
+            }
+            if (value.GetType().Name == typeof(List<>).Name)
+            {
+                foreach (var item in (IList)value)
+                {
+                    return GetClaimValueTypeFromValue(item);
+                }
+            }
+            return null;
```
RojaEnnam, line 123: If none of these types are matched, it returns null for ValueType of the Claim.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
             return claim.Value;
         }
 
+        /// <summary>
+        /// Gets ValueType of the claim from it's Value.
+        /// </summary>
+        /// <param name="value">Represents value of a claim.</param>
+        /// <returns>String representing claim's ValueType.</returns>
+        internal static string GetClaimValueTypeFromValue(object value)
```
RojaEnnam: Adding this method in JwtTokenUtilities to allow token specific ClaimTypes matching.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +58,77 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
         }
+
+        /// <summary>
+        /// Merges claims, If an item with same type exists in both the lists, the one in claims is picked.
+        /// </summary>
+        /// <param name="claims">Collection of claims.</param>
+        /// <param name="subjectClaims">Collection of claims.</param>
+        /// <param name="replace">Tells to either replace repeating items in subjectClaims or keep them.</param>
+        /// <returns></returns>
+        internal static IEnumerable<Claim> MergeClaims(IEnumerable<Claim> claims, IEnumerable<Claim> subjectClaims, bool replace)
```
RojaEnnam, line 69: <param name ="replace> is added for extensibility.

RojaEnnam: adding "replace" for extensibility.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -453,7 +455,7 @@ public virtual JwtSecurityToken CreateJwtSecurityToken(string issuer, string aud
         /// <exception cref="ArgumentException">If 'expires' &lt;= 'notBefore'.</exception>
         public virtual JwtSecurityToken CreateJwtSecurityToken(string issuer = null, string audience = null, ClaimsIdentity subject = null, DateTime? notBefore = null, DateTime? expires = null, DateTime? issuedAt = null, SigningCredentials signingCredentials = null)
```
RojaEnnam, line 456: Do we need to add a input parameter in CreateJwtSEcurityToken to provide an option to pass in SecurityTokenDescriptor.Claims?

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +58,77 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
+            }
+            if (value.GetType().Name == typeof(Double).Name)
+            {
+                return ClaimValueTypes.Double;
+            }
+            if (value.GetType().Name == typeof(DateTime).Name)
+            {
+                return ClaimValueTypes.DateTime;
+            }
+            if (value.GetType().Name == typeof(List<>).Name)
```
RojaEnnam: Checking for List because multiple Claims's can have same value for "Type" attribute and JwtTokenUtilities.CreateDictionaryFromClaims creates a list in this case.

test/Microsoft.IdentityModel.TestUtils/Default.cs
```diff
@@ -407,6 +407,48 @@ public static List<Claim> PayloadClaims
...
+        {
+            get => new Dictionary<string, object>()
+            {
+                { JwtRegisteredClaimNames.Aud, Audience },
+                { JwtRegisteredClaimNames.Iss, Issuer },
+                { "ClaimValueTypes.String", "ClaimValueTypes.String.Value" },
+                { "ClaimValueTypes.Boolean.true", true },
+                { "ClaimValueTypes.Boolean.false", false },
+                { "ClaimValueTypes.Double", 123.4 },
+                { "ClaimValueTypes.DateTime.IS8061", DateTime.TryParse("2019-11-15T14:31:21.6101326Z", out DateTime dateTimeValue1) ? dateTimeValue1 : new DateTime()},
```
RojaEnnam, line 442: 1. While using claims from dictionary to create a token:

```
- DateTime.TryParse("2019-11-15T14:31:21.6101326Z") result is 11/15/2019 6:31:21 AM( which will be the value of Claim after converting dictionary to Claim list)
- 11/15/2019 6:31:21 AM after serialization becomes 2019-11-15T06:31:21
- 2019-11-15T06:31:21 after decoding the token becomes 2019-11-15T14:31:21.0000000Z
```

2. While using Subject.Claims to create a token:

`- "2019-11-15T14:31:21.6101326Z" is the value of Claim and remains same after decoding the token.`

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -102,5 +104,79 @@ internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(stri
...
+
+        /// <summary>
+        /// Gets ValueType of the claim from it's Value.
+        /// </summary>
+        /// <param name="value">Represents value of a claim.</param>
+        /// <returns>String representing claim's ValueType.</returns>
+        internal static string GetClaimValueTypeFromValue(object value)
+        {
+            string claimValueType;
+            claimValueType = TokenUtilities.GetClaimValueTypeFromValue(value);
```
brentschmaltz: why not return TokenUtilities.GetClaimValueTypeFromValue, since it seems ok to return null?
not sure what value this method has over just calling TokenUtilities.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -65,6 +66,11 @@ public class JwtTokenUtilities
             JwtHeaderParameterNames.Zip
         };
 
+        /// <summary>
+        /// Creates a dictionary from set of Claim's 
+        /// </summary>
+        /// <param name="claims"></param>
+        /// <returns></returns>
```
keegan-caruso: Returns section should be filled out

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
         }
 
+        /// <summary>
+        /// Gets ValueType of the claim from it's Value.
+        /// </summary>
+        /// <param name="value">Represents value of a claim.</param>
+        /// <returns>String representing claim's ValueType.</returns>
+        internal static string GetClaimValueTypeFromValue(object value)
+        {
+            if (value.GetType().Name == typeof(String).Name && value.ToString() == string.Empty)
```
keegan-caruso: Can we use the `is` operator here instead of checking type via string comparison?

keegan-caruso: Same below with checking for JObject and JArray

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
RojaEnnam: Good point, I think I wanted to look for any SAML specific ClaimValueTypes like JSONArray etc.. for Jwt. For now we can call TokenUtilities directly. I forgot to clean up.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+            if (value.GetType().Name == typeof(JArray).Name)
+            {
+                return JsonClaimValueTypes.JsonArray;
+            }
+
+            return null;
+        }
+
+        /// <summary>
+        /// Creats claims from dictionary.
```
keegan-caruso: Creats => Creates

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
RojaEnnam: Good catch.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+        /// </summary>
+        /// <param name="claimsCollection">Represents claims stored in dictionary.</param>
+        /// <returns>claims</returns>
```
keegan-caruso: This should have a little more info

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+        internal static IEnumerable<Claim> CreateClaimsFromDictionary(IDictionary<string, object> claimsCollection)
+        {
+            List<Claim> claims = new List<Claim>();
+            object value;
+            foreach (string claimtype in claimsCollection.Keys)
+            {
+                claimsCollection.TryGetValue(claimtype, out value);
+                if (value != null)
```
keegan-caruso: It is conventional to use the bool returned by `TryGetValue` to check for existence

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.idictionary-2.trygetvalue?view=netcore-3.1

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+                {
+                    string valueType = GetClaimValueTypeFromValue(value);
+
+                    if (value.GetType().Name == typeof(List<>).Name)
+                    {
+                        foreach (var item in (IList)value)
```
keegan-caruso: using `is` with pattern matching directly above will remove the need for a cast here

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -102,5 +104,79 @@ internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(stri
...
+
+            if (claimValueType != null)
+                return claimValueType;
+
+            return null;
+        }
+
+        /// <summary>
+        /// Creats claims from dictionary.
```
keegan-caruso: Creats => Creates

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +58,77 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
                     yield return key;
         }
+
+        /// <summary>
+        /// Merges claims, If an item with same type exists in both the lists, the one in claims is picked.
+        /// </summary>
+        /// <param name="claims">Collection of claims.</param>
+        /// <param name="subjectClaims">Collection of claims.</param>
+        /// <param name="replace">Tells to either replace repeating items in subjectClaims or keep them.</param>
+        /// <returns></returns>
```
keegan-caruso: empty returns should be filled out

test/Microsoft.IdentityModel.TestUtils/Default.cs
```diff
@@ -407,6 +407,48 @@ public static List<Claim> PayloadClaims
...
+                { "ClaimValueTypes.DateTime", DateTime.TryParse("2019-11-15", out DateTime dateTimeValue2) ? dateTimeValue2 : new DateTime()},
+                { "ClaimValueTypes.JsonClaimValueTypes.Json1", JObject.Parse(@"{""jsonProperty1"":""jsonvalue1""}") },
+                { "ClaimValueTypes.JsonClaimValueTypes.Json2", JObject.Parse(@"{""jsonProperty2"":""jsonvalue2""}") },
+                { "ClaimValueTypes.JsonClaimValueTypes.JsonNull", "" },
+                { "ClaimValueTypes.JsonClaimValueTypes.JsonArray1", JArray.Parse(@"[1,2,3]") },
+                {"ClaimValueTypes.JsonClaimValueTypes.JsonArray2", JArray.Parse(@"[1,""2"",3]") },
```
keegan-caruso: nitpick: these two lines should have a space after the `{`

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
```diff
@@ -40,6 +41,78 @@ namespace System.IdentityModel.Tokens.Jwt.Tests
...
+                        Payload = Default.PayloadString,
+                        TokenDescriptor =  new SecurityTokenDescriptor
+                        {
+                            SigningCredentials = Default.AsymmetricSigningCredentials,
+                            EncryptingCredentials = null,
+                            Claims = Default.PayloadJsonDictionary,
+                            Subject = new ClaimsIdentity(Default.PayloadJsonClaims)
+                        },
+                        JwtSecurityTokenHandler = new JwtSecurityTokenHandler(),
+                        ValidationParameters =Default.AsymmetricSignTokenValidationParameters
```
keegan-caruso: nitpick: space after `=`

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
```diff
@@ -40,6 +41,78 @@ namespace System.IdentityModel.Tokens.Jwt.Tests
...
+                    },
+                    new CreateTokenTheoryData
+                    {
+                        First = true,
```
keegan-caruso: Only the first one should be First

test/Microsoft.IdentityModel.TestUtils/Default.cs
RojaEnnam: Following up, found that multiple dictionary to claim conversions is the reason for this. Will be adding fix in next commit.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: <param> section should be filled out as well.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: I think the comment here sounds a little confusing. Could it just be "Gets the value type of the claim"?

I think the comment for the param would be a little simpler if it was just:
`<param name="value">The claim value.</param>`

And same for the return value:
`<returns>The value type of the claim.</returns> `

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Another option is to use `value.GetType().Equals(typeof(String))`.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+        {
+            if (value.GetType().Name == typeof(String).Name && value.ToString() == string.Empty)
+                return JsonClaimValueTypes.JsonNull;
+
+            string claimValueType = TokenUtilities.GetClaimValueTypeFromValue(value);
+
+            if (!String.IsNullOrEmpty(claimValueType))
+                return claimValueType;
+
+            if (value.GetType().Name == typeof(JObject).Name)
```
mafurman: nit: remove the `{` here for consistency

We typically don't include `{` for single line 'if' statements.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+            {
+                return JsonClaimValueTypes.Json;
+            }
+            if (value.GetType().Name == typeof(JArray).Name)
```
mafurman: nit: We typically try to leave a space between the end of one 'if' statement and the start of another.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: grammar nit: "from **a** dictionary'

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+                return JsonClaimValueTypes.JsonArray;
+            }
+
+            return null;
+        }
+
+        /// <summary>
+        /// Creats claims from dictionary.
+        /// </summary>
+        /// <param name="claimsCollection">Represents claims stored in dictionary.</param>
```
mafurman: nit: I'd reword this as "A dictionary that represents a set of claims."

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Would it be simpler to just iterate through the 'key, value' pairs instead of calling TryGetValue? 

`foreach(var entry in claimsCollection)`

Then you can access `entry.Key` and `entry.Value`.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+                            claims.Add(new Claim(claimtype, item.ToString(), valueType));
+                        }
+                    }
+                    else
+                    {
+                        claims.Add(new Claim(claimtype, value.ToString(), valueType));
+                    }
+                }
+            }
+            return claims;
```
mafurman: nit: include a space after the last `}`

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+            {
+                claimsCollection.TryGetValue(claimtype, out value);
+                if (value != null)
+                {
+                    string valueType = GetClaimValueTypeFromValue(value);
+
+                    if (value.GetType().Name == typeof(List<>).Name)
+                    {
+                        foreach (var item in (IList)value)
+                        {
```
mafurman: nit: I think you can remove the `{` in the foreach loop

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+                    {
+                        claims.Add(new Claim(claimtype, value.ToString(), valueType));
+                    }
+                }
+            }
+            return claims;
+        }
+
+        /// <summary>
+        /// Merges SecurityTokenDescriptor.Claims and SecurityTokenDescriptor.Subject.Claims
```
mafurman: nit: We typically put '.' at the end of comments.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+        /// </summary>
+        /// <param name="claims">Represents SecurityTokenDescriptor.Claims.</param>
+        /// <param name="subjectClaims">SecurityTokenDescriptor.Subject.Claims.</param>
+        /// <returns>Merged list of claims </returns>
```
mafurman: I'd reword this comment as: 
`<param name="claims">A dictionary of claims (represents SecurityTokenDescriptor.Claims).</param>`
`<param name="subjectClaims">An IEnumerable of claims (represents SecurityTokenDescriptor.Subject.Claims).</param>`
`<returns>**A** merged list of claims.</returns> `

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -278,6 +284,87 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+        internal static IEnumerable<Claim> GetAllClaims(IDictionary<string , object> claims, IEnumerable<Claim> subjectClaims)
+        {
+            IEnumerable<Claim> allClaims = null;
+            if (claims != null)
+                allClaims = CreateClaimsFromDictionary(claims);
+
+            if (allClaims != null && allClaims.Any())
+            {
```
mafurman: nit: `{` unnecessary

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +58,77 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
                 foreach (SecurityKey key in validationParameters.IssuerSigningKeys)
                     yield return key;
         }
+
+        /// <summary>
+        /// Merges claims, If an item with same type exists in both the lists, the one in claims is picked.
+        /// </summary>
+        /// <param name="claims">Collection of claims.</param>
+        /// <param name="subjectClaims">Collection of claims.</param>
+        /// <param name="replace">Tells to either replace repeating items in subjectClaims or keep them.</param>
```
mafurman: I'd change the param name to 'shouldKeepRepeatingClaims' and modify the comment slightly to "if true, repeating items in subjectClaims will be kept. Otherwise, they will be removed." (if I'm understanding this correctly)

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +58,77 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
                 foreach (SecurityKey key in validationParameters.IssuerSigningKeys)
                     yield return key;
         }
+
+        /// <summary>
+        /// Merges claims, If an item with same type exists in both the lists, the one in claims is picked.
```
mafurman: nit: I'd put a '.' after "Merges claims."

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -359,7 +360,7 @@ public virtual string CreateEncodedJwt(SecurityTokenDescriptor tokenDescriptor)
         /// <returns>A Base64UrlEncoded string in 'Compact Serialization Format'.</returns>
         public virtual string CreateEncodedJwt(string issuer, string audience, ClaimsIdentity subject, DateTime? notBefore, DateTime? expires, DateTime? issuedAt, SigningCredentials signingCredentials)
         {
-            return CreateJwtSecurityTokenPrivate(issuer, audience, subject, notBefore, expires, issuedAt, signingCredentials, null).RawData;
+            return CreateJwtSecurityTokenPrivate(issuer, audience, subject, notBefore, expires, issuedAt, signingCredentials, null,null).RawData;
```
mafurman: nit: should be a space before the new null argument

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
RojaEnnam: I agree. Sounds simple and easy to understand.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
RojaEnnam: I find it repeating to use entry.Key and entry.Value every time I need them. I can define a variable for each of them and use , but this necessarily is doing the same.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
RojaEnnam: When using` is`, not sure why but JObject and JSONArray also resolve to match IList, which is not expected in this case, as multiple Claims should not be created from a JObject/JSONArray.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
RojaEnnam: `value.GetType().Equals(typeof(String))`  option does not work for List. As list can contain object of any type.

1. `value.GetType()` resolves to System.Collections.Generic.List`1[System.String] - For String object in list. 
2. `typeof(List<>)` resolves to System.Collections.Generic.List`1[T] .

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -29,6 +29,7 @@
 using Microsoft.IdentityModel.Logging;
 using Microsoft.IdentityModel.Tokens;
 using System;
+using System.Collections;
```
keegan-caruso: This reference is unused.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -102,5 +104,60 @@ internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(stri
             }
             return null;
         }
+
+        /// <summary>
+        /// Creates claims from  a dictionary.
```
keegan-caruso: nit: two spaces between from and a

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -102,5 +104,60 @@ internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(stri
...
+        /// </summary>
+        /// <param name="claimsCollection"> A dictionary that represents a set of claims.</param>
+        /// <returns> A collection of Claim objects created from the dictionary.</returns>
+        internal static IEnumerable<Claim> CreateClaimsFromDictionary(IDictionary<string, object> claimsCollection)
+        {
+            List<Claim> claims = null;
+            object value;
```
keegan-caruso: nit: this can be inlined with the out parameter

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -102,5 +104,60 @@ internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(stri
...
+            if (claimsCollection == null)
+                return claims;
+
+            claims = new List<Claim>();
+            foreach (string claimtype in claimsCollection.Keys)
+            {
+                if (claimsCollection.TryGetValue(claimtype, out value))
```
keegan-caruso: The other reason for iterating over the KeyValuePair instead of the Key is that you don't need the `TryGetValue`. You have direct access to the Key and the value and end up with a lower cyclomatic complexity value / more simple code for this method.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
keegan-caruso: The why would be that JObject and JSONArray implement IList. That makes sense. Thank you for looking into that.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -102,5 +104,60 @@ internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(stri
...
+                        claims.Add(new Claim(claimtype, value.ToString(), valueType));
+                    }
+                }
+            }
+
+            return claims;
+        }
+
+        /// <summary>
+        /// Merges IDictionary of claims and IEnumerable of claims.
```
keegan-caruso: When referencing Symbols in XML style comments it is standard to use a `<see cref="{symbolName}"/>`

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -26,10 +26,13 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Security.Claims;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.IdentityModel.Json.Linq;
```
keegan-caruso: unused reference

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +58,70 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
                 foreach (SecurityKey key in validationParameters.IssuerSigningKeys)
                     yield return key;
         }
+
+        /// <summary>
+        /// Merges claims. If a claim with same type exists in both the lists, the one in claims will be kept.
```
keegan-caruso: when referencing parameters it is standard to use paramref eg `<paramref name="claims"/>`

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +58,70 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
+                return subjectClaims;
+
+            if (subjectClaims == null)
+                return claims;
+
+            List<Claim> result = claims.ToList();
+
+            foreach (Claim claim in subjectClaims)
+            {
+                if (claims.Where(i => i.Type == claim.Type).FirstOrDefault() == null)
```
keegan-caruso: Consider using `!Any` with the same lambda

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +58,70 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
+        /// <summary>
+        /// Gets the value type of the claim.
+        /// </summary>
+        /// <param name="value"> The claim value.</param>
+        /// <returns> The value type of the claim.</returns>
+        internal static string GetClaimValueTypeFromValue(object value)
+        {
+            if (value != null)
+            {
+                if (value.GetType().Name == typeof(String).Name)
```
keegan-caruso: Prefer language keywords instead of BCL types

See item 11 here https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/coding-style.md

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -722,7 +741,7 @@ private DateTime GetDateTime(string key)
         /// <returns>This instance as JSON.</returns>
         /// <remarks>Use <see cref="JsonExtensions.Serializer"/> to customize JSON serialization.</remarks>
         public virtual string SerializeToJson()
-        {
+                {
```
keegan-caruso: Should be a single indent

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
mafurman: I think it might be beneficial to add overloaded methods for both CreateEncodedJwt() and CreateJwtSecurityToken() that take in a Dictionary<string, object> instead of a ClaimsIdentity.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -75,59 +74,53 @@ internal static IEnumerable<Claim> MergeClaims(IEnumerable<Claim> claims, IEnume
...
+                if (value.GetType().Name == typeof(Double).Name)
+                    return ClaimValueTypes.Double;
+
+                if (value.GetType().Name == typeof(DateTime).Name)
+                    return ClaimValueTypes.DateTime;
+
+                if (value.GetType().Name == typeof(List<>).Name)
                 {
-                    return GetClaimValueTypeFromValue(item);
+                    foreach (var item in (IList)value)
```
mafurman, line 113: Wouldn't this return the value type for the first item in the list only?

If so, should this be a for each loop?

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -79,18 +79,23 @@ public JwtPayload(string issuer, string audience, IEnumerable<Claim> claims, Dat
...
+        /// values will be overridden.</param>
+        /// <param name="notBefore">If notbefore.HasValue a { nbf, 'value' } claim is added, overwriting any 'nbf' claim in 'claims' and 'claimcollection' if present.</param>
+        /// <param name="expires">If expires.HasValue a { exp, 'value' } claim is added, overwriting any 'exp' claim in 'claims' and 'claimcollection' if present.</param>
+        /// <param name="issuedAt">If issuedAt.HasValue is 'true' a { iat, 'value' } claim is added, overwriting any 'iat' claim in 'claims' and 'claimcollection' if present.</param>
         /// <remarks>Comparison is set to <see cref="StringComparer.Ordinal"/>
-        /// <para>The 4 parameters: 'issuer', 'audience', 'notBefore', 'expires' take precednece over <see cref="Claim"/>(s) in 'claims'. The values in 'claims' will be overridden.</para></remarks>
+        /// <para>The 4 parameters: 'issuer', 'audience', 'notBefore', 'expires' take precednece over <see cref="Claim"/>(s) in 'claims' and 'claimcollection'. The values will be overridden.</para></remarks>
```
mafurman: nit: "precedence"

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -79,18 +79,23 @@ public JwtPayload(string issuer, string audience, IEnumerable<Claim> claims, Dat
...
         /// <exception cref="ArgumentException">If 'expires' &lt;= 'notbefore'.</exception>
-        public JwtPayload(string issuer, string audience, IEnumerable<Claim> claims, DateTime? notBefore, DateTime? expires, DateTime? issuedAt)
+        public JwtPayload(string issuer, string audience, IEnumerable<Claim> claims, IDictionary<string, object> claimsCollection, DateTime? notBefore, DateTime? expires, DateTime? issuedAt)
```
mafurman, line 134: Modifying the parameters for this public method would be considered a breaking change. I would make this a new overloaded method.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
RojaEnnam: Done.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
RojaEnnam: Added symbol referencing.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
RojaEnnam: New Learning :)

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
RojaEnnam: Yes, it returns value type for first item only. 

I used `foreach` here to handle the case when there are objects of different types in the list. From the below comment in JwtTokenUtilities.CreateDictionaryFromClaims method, looks like we have not decided yet on how to handle this case.
`// What to do if the 'ClaimValueType' is not the same.`

I kept `foreach `   in this method to have the comment below make more sense for someone looking at the code.
`// What to do if the value type of objects in the list are different.`

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
RojaEnnam: Thank you. Its a new learning for me. It makes sense considering cyclomatic complexity.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
RojaEnnam: Agreed.` Any` is more intuitive.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
RojaEnnam: Good to know.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
RojaEnnam: Yes. Indeed it is. Thank you for clearing that up.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
RojaEnnam: Good idea. Don't we need both ClaimsIdentity and Dictionary<string, object>?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -65,6 +65,11 @@ public class JwtTokenUtilities
             JwtHeaderParameterNames.Zip
         };
 
+        /// <summary>
+        /// Creates a dictionary from a list of Claim's.
+        /// </summary>
+        /// <param name="claims"> A list of claims.</param>
+        /// <returns> A Dictionary representing claims.</returns>
         internal static Dictionary<string, object> CreateDictionaryFromClaims(IEnumerable<Claim> claims)
```
brentschmaltz: It's preferable to return IDictionary<string, object> than Dictionary<string, object>.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -102,5 +100,62 @@ internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(stri
             }
             return null;
         }
+
```
brentschmaltz, line 103: I know SamlTokenUtilities was added in a separate PR, but it should be internal.
I added 1022496 to our backlog to add checking for new public apis to our API compat analysis.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +55,70 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
+        /// <summary>
+        /// Gets the value type of the <see cref="Claim"/> from its value <paramref name="value"/>
+        /// </summary>
+        /// <param name="value"> The <see cref="Claim"/> value.</param>
+        /// <returns> The value type of the <see cref="Claim"/>.</returns>
+        internal static string GetClaimValueTypeFromValue(object value)
+        {
+            if (value != null)
+            {
+                if (value.GetType().Name == typeof(string).Name)
```
brentschmaltz: We don't have to convert to a string, we can write: 
if (value is string)
    return ClaimsValueTypes.String

if we don't have a good suggestion, then we should return 
1. null, if value is null
2. value.GetType().ToString();

We will need some type of extensibility here for complex types.
But for now, this will handle the basic JSON types.

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -510,6 +559,20 @@ public void AddClaims(IEnumerable<Claim> claims)
             }
         }
 
+        /// <summary>
+        /// Adds claims from dictionary.
+        /// </summary>
+        /// <param name="claimsCollection"> A dictionary of claims.</param>
+        /// <remark> If a key is already present in target dictionary, its value is overridden by the value of the key in claimsCollection.</remark>
+        public void MergeDictionaryClaims(IDictionary<string, object> claimsCollection)
```
brentschmaltz, line 581: this should be internal

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
brentschmaltz: @mafurman @RojaEnnam we are trying to steer users towards JsonWebToken. I am wondering if we should have this method?

All it means is that JwtSecurityTokenHandler consolidates to 'Claim' before calling this method.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +55,70 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
                 foreach (SecurityKey key in validationParameters.IssuerSigningKeys)
                     yield return key;
         }
+
```
brentschmaltz, line 58: TokenUtilities should be internal.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -429,7 +455,34 @@ public virtual JwtSecurityToken CreateJwtSecurityToken(SecurityTokenDescriptor t
...
+        /// <para><see cref="SigningCredentials.SigningCredentials(SecurityKey, string)"/> is used to sign <see cref="JwtSecurityToken.RawData"/>.</para>
+        /// <para><see cref="EncryptingCredentials.EncryptingCredentials(SecurityKey, string, string)"/> is used to encrypt <see cref="JwtSecurityToken.RawData"/> or <see cref="JwtSecurityToken.RawPayload"/> .</para>
+        /// </remarks>
+        /// <returns>A <see cref="JwtSecurityToken"/>.</returns>
+        /// <exception cref="ArgumentException">If 'expires' &lt;= 'notBefore'.</exception>
+        public virtual JwtSecurityToken CreateJwtSecurityToken(string issuer, string audience, ClaimsIdentity subject, DateTime? notBefore, DateTime? expires, DateTime? issuedAt, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, IDictionary<string, object> claimCollection)
+        {
+            return CreateJwtSecurityTokenPrivate(issuer, audience, subject, notBefore, expires, issuedAt, signingCredentials, encryptingCredentials, claimCollection);
```
brentschmaltz, line 485: @mafurman @RojaEnnam we could make a similar argument as for JwtPayload. Do we really need this?

Every public API we ship, we have to support, sometimes less is more.

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
mafurman: @brentschmaltz That's actually a good point! I don't think this necessarily needs to be a public API; we can always add it later if it's needed for some reason.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
RojaEnnam: This method is only used for Saml/Saml2

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
RojaEnnam: @brentschmaltz @mafurman  Originally I implemented it by consolidating dictionary to 'Claim'(s) and then call JwtPayload but, 'AddClaims' method again converts the 'Claim'(s) to dictionary. When this is happening on DateTime objects I observed the original value is not retained.

```
- DateTime.TryParse("2019-11-15T14:31:21.6101326Z") result is "11/15/2019 6:31:21 AM"( which will be the value of Claim after converting dictionary to Claim list)
- After 'AddClaims' Date.TryParse("11/15/2019 6:31:21 AM") results in 2019-11-15T14:31:21.0000000Z instead of 2019-11-15T14:31:21.6101326Z 
```

brentschmaltz:
@RojaEnnam you need to rebase roennam/TokenDescriptor.Claims against 'dev' since: 
protected virtual IEnumerable<SecurityKey> ResolveIssuerSigningKey(string token, Saml2SecurityToken samlToken, TokenValidationParameters validationParameters, out bool keyMatched);
was reverted back to:
protected virtual SecurityKey ResolveIssuerSigningKey(string token, Saml2SecurityToken samlToken, TokenValidationParameters validationParameters);

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
RojaEnnam: @brentschmaltz my thought process was, If we don't have this, users can use both SecurityTokenDescriptor 'Claims' and  'Subject.Claims' only when they call CreateToken(SecurityTokenDescriptor tokenDescriptor)

brentschmaltz:
The PR has moved to roennam/SecurityTokenDescriptor.Claims

PR #1415, opened 2020-05-19
Allow validation of signed http requests claims if present

### LINKED ISSUE #1393: SignedHttpRequest values should be validated if they are found.

keegan-caruso:
The truth table is the general idea of my proposed changes. Validate is always respected, if !Validate, then check if the user opted in to check present claims. The claims to check are also configurable which has a sensible default. 

The alternative here would be to add another bool that is used for every claim to control the validate if present behavior

| Claim Valid | Claim present | Validate On | Validate if present | Validate | Fail |
| :---------: | :-----------: | :---------: | :-----------------: | :------: | :--: |
|             | x             | x           |                     |  x       | x    |
|             | x             |             |                     |          |      |
| x           | x             | x           |                     |  x       |      |
| x           | x             |             |                     |          |      |
|             |               | x           |                     | x        | x    |
|             |               |             |                     |          |      |
|             | x             | x           | x                   | x        | x    |
|             | x             |             | x                   | x        | x    |
| x           | x             | x           | x                   | x        |      |
| x           | x             |             | x                   | x        |      |
|             |               | x           | x                   | x        | x    |
|             |               |             | x                   |          |      |

---

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
+++ b/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
@@ -556,7 +556,7 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
             if (validationParameters.ReplayValidatorAsync != null)
                 await signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ReplayValidatorAsync(jwtSignedHttpRequest, signedHttpRequestValidationContext, cancellationToken).ConfigureAwait(false);
 
-            if (ShouldValidate(jwtSignedHttpRequest,validationParameters.ValidateTs, ValidateIfPresent(validationParameters, SignedHttpRequestClaimTypes.Ts), SignedHttpRequestClaimTypes.Ts))
+            if (ShouldValidate(jwtSignedHttpRequest, validationParameters.ValidateTs, ValidateIfPresent(validationParameters, SignedHttpRequestClaimTypes.Ts), SignedHttpRequestClaimTypes.Ts))
                 ValidateTsClaim(jwtSignedHttpRequest, signedHttpRequestValidationContext);
 
             if (ShouldValidate(jwtSignedHttpRequest, validationParameters.ValidateM, ValidateIfPresent(validationParameters, SignedHttpRequestClaimTypes.M), SignedHttpRequestClaimTypes.M))
--- a/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
+++ b/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
@@ -551,31 +551,53 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
             if (!(signedHttpRequest is JsonWebToken jwtSignedHttpRequest))
                 throw LogHelper.LogExceptionMessage(new SignedHttpRequestValidationException(LogHelper.FormatInvariant(LogMessages.IDX23030, signedHttpRequest.GetType(), typeof(JsonWebToken), signedHttpRequest)));
 
-            if (signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ReplayValidatorAsync != null)
+            var validationParameters = signedHttpRequestValidationContext.SignedHttpRequestValidationParameters;
+
+            if (validationParameters.ReplayValidatorAsync != null)
                 await signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ReplayValidatorAsync(jwtSignedHttpRequest, signedHttpRequestValidationContext, cancellationToken).ConfigureAwait(false);
 
-            if (signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ValidateTs)
+            if (ShouldValidate(jwtSignedHttpRequest,validationParameters.ValidateTs, ValidateIfPresent(validationParameters, SignedHttpRequestClaimTypes.Ts), SignedHttpRequestClaimTypes.Ts))
                 ValidateTsClaim(jwtSignedHttpRequest, signedHttpRequestValidationContext);
 
-            if (signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ValidateM)
+            if (ShouldValidate(jwtSignedHttpRequest, validationParameters.ValidateM, ValidateIfPresent(validationParameters, SignedHttpRequestClaimTypes.M), SignedHttpRequestClaimTypes.M))
                 ValidateMClaim(jwtSignedHttpRequest, signedHttpRequestValidationContext);
 
-            if (signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ValidateU)
+            if (ShouldValidate(jwtSignedHttpRequest, validationParameters.ValidateU, ValidateIfPresent(validationParameters, SignedHttpRequestClaimTypes.U), SignedHttpRequestClaimTypes.U))
                 ValidateUClaim(jwtSignedHttpRequest, signedHttpRequestValidationContext);
 
-            if (signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ValidateP)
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -542,33 +542,55 @@ protected virtual async Task<SecurityToken> ValidateSignedHttpRequestPayloadAsyn
...
+        internal virtual bool ShouldValidate(JsonWebToken jwtSignedHttpRequest, bool validateClaim, bool validateIfPresent, string claimName)
+        {
+            return validateClaim || (validateIfPresent && jwtSignedHttpRequest.TryGetClaim(claimName, out var claimValue) && claimValue != null);
+        }
+
+        private bool ValidateIfPresent(SignedHttpRequestValidationParameters validationParameters, string claim)
+        {
+            return validationParameters.ValidatePresentClaims &&
+                validationParameters.ClaimsValidatedWhenPresent != null &&
+                validationParameters.ClaimsValidatedWhenPresent.Contains(claim) == true;
```
RojaEnnam: Is `==true `needed ? Contains method returns true/false.

keegan-caruso: No, it is not needed here.

Originally had this as `validationParameters.ClaimsValidatedWhenPresent?.Contains(claim) == true` to coerce the nullable bool to a bool. I ended up not doing that as I thought this was a little more clear

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -623,8 +645,9 @@ internal virtual async Task<SecurityKey> ValidateSignatureAsync(JsonWebToken sig
         /// <param name="signedHttpRequest">A SignedHttpRequest.</param>
         /// <param name="signedHttpRequestValidationContext">A structure that wraps parameters needed for SignedHttpRequest validation.</param>
         /// <remarks>
-        /// This method will be executed only if <see cref="SignedHttpRequestValidationParameters.ValidateTs"/> is set to <c>true</c>.
-        /// </remarks>    
+        /// This method will be executed only if <see cref="SignedHttpRequestValidationParameters.ValidateTs"/> is set to <c>true</c> or a ts claim is present and
```
GeoK, line 657: This comment is not entirely true, as a claim (ts in this case) has to present in the ClaimsValidatedWhenPresent collection.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -623,8 +645,9 @@ internal virtual async Task<SecurityKey> ValidateSignatureAsync(JsonWebToken sig
...
+        /// <see cref="SignedHttpRequestValidationParameters.ValidatePresentClaims"/> is set to true.
```
mafurman: Might be helpful to include a `<see cref="SignedHttpRequestClaimTypes.Ts"/>` in the comment.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs
```diff
@@ -216,5 +216,28 @@ public TimeSpan SignedHttpRequestLifetime
...
+        /// </remarks>
+        public bool ValidatePresentClaims { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets the claims to validate if present.
+        /// </summary>
+        /// <remarks>
+        /// Validation will only occur if <see cref="ValidatePresentClaims"/> is set to <c>true</c>.
+        /// </remarks>
+        public IEnumerable<string> ClaimsValidatedWhenPresent { get; set; } = new List<string>
```
mafurman: I personally find the name of this property somewhat confusing. Maybe rename to "ClaimsToValidateWhenPresent"?

mafurman (approved):

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs
keegan-caruso: Thanks for the suggestion! I will also sort it appropriately.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
RojaEnnam: Got it, Thanks.

RojaEnnam (approved):

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -551,33 +551,55 @@ protected virtual async Task<SecurityToken> ValidateSignedHttpRequestPayloadAsyn
...
                 throw LogHelper.LogExceptionMessage(new SignedHttpRequestValidationException(LogHelper.FormatInvariant(LogMessages.IDX23030, signedHttpRequest.GetType(), typeof(JsonWebToken), signedHttpRequest)));
 
-            if (signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ReplayValidatorAsync != null)
+            var validationParameters = signedHttpRequestValidationContext.SignedHttpRequestValidationParameters;
+
+            if (validationParameters.ReplayValidatorAsync != null)
                 await signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ReplayValidatorAsync(jwtSignedHttpRequest, signedHttpRequestValidationContext, cancellationToken).ConfigureAwait(false);
 
-            if (signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.ValidateTs)
+            if (ShouldValidate(jwtSignedHttpRequest,validationParameters.ValidateTs, ValidateIfPresent(validationParameters, SignedHttpRequestClaimTypes.Ts), SignedHttpRequestClaimTypes.Ts))
```
brentschmaltz: nit: space after ValidateTs,

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs
```diff
@@ -109,6 +109,20 @@ public class SignedHttpRequestValidationParameters
...
 
+        /// <summary>
+        /// Gets or sets the claims to validate if present.
+        /// </summary>
+        /// <remarks>
+        /// Validation will only occur if <see cref="ValidatePresentClaims"/> is set to <c>true</c>.
+        /// </remarks>
+        public IEnumerable<string> ClaimsToValidateWhenPresent { get; set; } = new List<string>
+        {
+            SignedHttpRequestClaimTypes.Ts,
```
GeoK: Should we remove 'Ts' and 'U' from the collection? 
'M' and 'P' were mentioned as recommended claims.

keegan-caruso: Good question. I was defaulting recommended and required claims. 
We would expect the usual case is Ts and U are part of the always validated claims.

GeoK: I agree. The reason for this feature is using 'M' and 'P' as recommended claims, but we are providing additional extensibility so that any claim could be validated in this fashion.

Having only 'M' and 'P' in this collection would reduce number of steps that users have to do to enable this scenario. In this case users would flip ValidateM and ValidateP switches to `false` and ValidatePresentClaims to `true`.

keegan-caruso: This is a good point. I'll make the change

PR #1416, opened 2020-05-19
Checking if arg is null in RemovePII

### LINKED ISSUE #1406: NullReferenceException in LogHelper.RemovePII when arg is null
According to description, WsFederationMessage FromQueryString should allow "empty or null" queryString. But it fails with NullReferenceException when queryString is null 

```c#
 public static WsFederationMessage FromQueryString(string queryString) {
            LogHelper.LogVerbose(FormatInvariant(LogMessages.IDX22900, queryString));
```
due to issue with LogHelper.RemovePII:
```c#
private static string RemovePII(object arg)
        {
            if (arg is Exception ex && ex.GetType().FullName.StartsWith("Microsoft.IdentityModel.", StringComparison.Ordinal))
                return ex.ToString();

            return arg.GetType().ToString();
        }
```

RojaEnnam:
Fixes #1406

### DIFF
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -29,7 +29,6 @@ using System;
 using System.Diagnostics.Tracing;
 using System.Globalization;
 using System.Linq;
-using System.Collections.Generic;
 
 namespace Microsoft.IdentityModel.Logging
 {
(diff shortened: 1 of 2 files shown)


keegan-caruso:
Looking at the issue description, it might be worth adding a test to WsFederationMessage:FromQueryString with an input of empty string and null to prove that the code matches the description on the method. Specifically the remarks section on FromQueryString.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -339,7 +339,7 @@ private static string RemovePII(object arg)
             if (arg is Exception ex && ex.GetType().FullName.StartsWith("Microsoft.IdentityModel.", StringComparison.Ordinal))
                 return ex.ToString();
 
-            return arg.GetType().ToString();
+            return arg?.GetType().ToString();
```
brentschmaltz: Can you also add a check in FormatInvariant above.
if args is null, then we will throw a null ref.

brentschmaltz: @RojaEnnam ill make this change and push the commit.

GeoK (approved):

PR #1417, opened 2020-05-20
Support for SecurityTokenDescriptor.Claims in JwtSecurity/Saml/Saml2 Tokens

### LINKED ISSUE #1155: Add support for creating Saml and Saml2 tokens with TokenDescriptor.Claims

RojaEnnam:
Fixes #1155

1. Adding method to convert dictionary of claims to list of Claim objects.
2. Merging both SecurityTokenDescriptor.Claims and SecurityTokenDescriptor.Subject.Claims and using them in creating token.

[ New PR to resolve rebase issues from #1414 ]

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
@@ -28,8 +28,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+using System.Security.Claims;
 using Microsoft.IdentityModel.Logging;
 using TokenLogMessages = Microsoft.IdentityModel.Tokens.LogMessages;
 
@@ -38,7 +37,7 @@ namespace Microsoft.IdentityModel.Tokens
     /// <summary>
     /// A class which contains useful methods for processing tokens.
     /// </summary>
-    public class TokenUtilities
+    internal class TokenUtilities
     {
         /// <summary>
         /// Returns all <see cref="SecurityKey"/> provided in validationParameters.
@@ -54,6 +53,31 @@ namespace Microsoft.IdentityModel.Tokens
             if (validationParameters.IssuerSigningKeys != null)
                 foreach (SecurityKey key in validationParameters.IssuerSigningKeys)
                     yield return key;
+        }
+
+        /// <summary>
+        /// Merges claims. If a claim with same type exists in both <paramref name="claims"/> and <paramref name="subjectClaims"/>, the one in claims will be kept.
+        /// </summary>
+        /// <param name="claims"> Collection of <see cref="Claim"/>'s.</param>
+        /// <param name="subjectClaims"> Collection of <see cref="Claim"/>'s.</param>
+        /// <returns> A Merged list of <see cref="Claim"/>'s.</returns>
+        internal static IEnumerable<Claim> MergeClaims(IEnumerable<Claim> claims, IEnumerable<Claim> subjectClaims)
+        {
+            if (claims == null)
+                return subjectClaims;
+
+            if (subjectClaims == null)
+                return claims;
+
+            List<Claim> result = claims.ToList();
+
+            foreach (Claim claim in subjectClaims)
+            {
+                if (!claims.Any(i => i.Type == claim.Type))
+                    result.Add(claim);
+            }
+
+            return result;
         }
     }
 }
--- a/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
@@ -27,6 +27,7 @@
 
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Security.Claims;
 using Microsoft.IdentityModel.Json;
 using Microsoft.IdentityModel.Json.Linq;
@@ -83,7 +84,7 @@ namespace System.IdentityModel.Tokens.Jwt
         /// <param name="expires">If expires.HasValue a { exp, 'value' } claim is added, overwriting any 'exp' claim in 'claims' if present.</param>
         /// <param name="issuedAt">If issuedAt.HasValue is 'true' a { iat, 'value' } claim is added, overwriting any 'iat' claim in 'claims' if present.</param>
         /// <remarks>Comparison is set to <see cref="StringComparer.Ordinal"/>
-        /// <para>The 4 parameters: 'issuer', 'audience', 'notBefore', 'expires' take precednece over <see cref="Claim"/>(s) in 'claims'. The values in 'claims' will be overridden.</para></remarks>
+        /// <para>The 4 parameters: 'issuer', 'audience', 'notBefore', 'expires' take precedence over <see cref="Claim"/>(s) in 'claims'. The values will be overridden.</para></remarks>
         /// <exception cref="ArgumentException">If 'expires' &lt;= 'notbefore'.</exception>
         public JwtPayload(string issuer, string audience, IEnumerable<Claim> claims, DateTime? notBefore, DateTime? expires, DateTime? issuedAt)
             : base(StringComparer.Ordinal)
@@ -91,6 +92,45 @@ namespace System.IdentityModel.Tokens.Jwt
             if (claims != null)
                 AddClaims(claims);
 
+            AddFirstPriorityClaims(issuer, audience, notBefore, expires, issuedAt);
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="JwtPayload"/> class with claims added for each parameter specified. Default string comparer <see cref="StringComparer.Ordinal"/>. 
+        /// </summary>
+        /// <param name="issuer">If this value is not null, a { iss, 'issuer' } claim will be added, overwriting any 'iss' claim in 'claims' and 'claimCollection' if present.</param>
+        /// <param name="audience">If this value is not null, a { aud, 'audience' } claim will be added, appending to any 'aud' claims in 'claims' or 'claimCollection' if present.</param>
+        /// <param name="claims">If this value is not null then for each <see cref="Claim"/> a { 'Claim.Type', 'Claim.Value' } is added. If duplicate claims are found then a { 'Claim.Type', List&lt;object&gt; } will be created to contain the duplicate values.</param>
+        /// <param name="claimsCollection">If both <paramref name="claims"/> and <paramref name="claimsCollection"/> are not null then the values in claims will be combined with the values in claimsCollection. The values found in claimCollection take precedence over those found in claims, so any duplicate
+        /// values will be overridden.</param>
+        /// <param name="notBefore">If notbefore.HasValue a { nbf, 'value' } claim is added, overwriting any 'nbf' claim in 'claims' and 'claimcollection' if present.</param>
+        /// <param name="expires">If expires.HasValue a { exp, 'value' } claim is added, overwriting any 'exp' claim in 'claims' and 'claimcollection' if present.</param>
+        /// <param name="issuedAt">If issuedAt.HasValue is 'true' a { iat, 'value' } claim is added, overwriting any 'iat' claim in 'claims' and 'claimcollection' if present.</param>
+        /// <remarks>Comparison is set to <see cref="StringComparer.Ordinal"/>
+        /// <para>The 4 parameters: 'issuer', 'audience', 'notBefore', 'expires' take precedence over <see cref="Claim"/>(s) in 'claims' and 'claimcollection'. The values will be overridden.</para></remarks>
+        /// <exception cref="ArgumentException">If 'expires' &lt;= 'notbefore'.</exception>
+        public JwtPayload(string issuer, string audience, IEnumerable<Claim> claims, IDictionary<string, object> claimsCollection, DateTime? notBefore, DateTime? expires, DateTime? issuedAt)
+            : base(StringComparer.Ordinal)
+        {
+            if (claims != null)
+                AddClaims(claims);
+
+            if (claimsCollection != null && claimsCollection.Any())
+                AddDictionaryClaims(claimsCollection);
+
+            AddFirstPriorityClaims(issuer, audience, notBefore, expires, issuedAt);
+        }
+
+        /// <summary>
+        /// Adds Nbf, Exp, Iat, Iss and Aud claims to payload
+        /// </summary>
+        /// <param name="issuer">If this value is not null, a { iss, 'issuer' } claim will be added, overwriting any 'iss' claim in <see cref="JwtPayload"/> instance.</param>
+        /// <param name="audience">If this value is not null, a { aud, 'audience' } claim will be added, appending to any 'aud' claims in <see cref="JwtPayload"/> instance.</param>
+        /// <param name="notBefore">If notbefore.HasValue a { nbf, 'value' } claim is added, overwriting any 'nbf' claim in <see cref="JwtPayload"/> instance.</param>
+        /// <param name="expires">If expires.HasValue a { exp, 'value' } claim is added, overwriting any 'exp' claim in <see cref="JwtPayload"/> instance.</param>
+        /// <param name="issuedAt">If issuedAt.HasValue is 'true' a { iat, 'value' } claim is added, overwriting any 'iat' claim in <see cref="JwtPayload"/> instance.</param>
+        internal void AddFirstPriorityClaims(string issuer, string audience, DateTime? notBefore, DateTime? expires, DateTime? issuedAt)
+        {
             if (expires.HasValue)
             {
                 if (notBefore.HasValue)
@@ -508,6 +548,20 @@ namespace System.IdentityModel.Tokens.Jwt
                     this[jsonClaimType] = jsonClaimValue;
                 }
             }
+        }
+
+        /// <summary>
+        /// Adds claims from dictionary.
+        /// </summary>
+        /// <param name="claimsCollection"> A dictionary of claims.</param>
+        /// <remark> If a key is already present in target dictionary, its value is overridden by the value of the key in claimsCollection.</remark>
+        internal void AddDictionaryClaims(IDictionary<string, object> claimsCollection)
+        {
+            if (claimsCollection == null)
+                throw LogHelper.LogExceptionMessage(new ArgumentNullException(nameof(claimsCollection)));
+
+            foreach (string type in claimsCollection.Keys)
+                this[type] = claimsCollection[type];
         }
 
         internal static string GetClaimValueType(object obj)
--- a/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
@@ -359,7 +359,7 @@ namespace System.IdentityModel.Tokens.Jwt
         /// <returns>A Base64UrlEncoded string in 'Compact Serialization Format'.</returns>
         public virtual string CreateEncodedJwt(string issuer, string audience, ClaimsIdentity subject, DateTime? notBefore, DateTime? expires, DateTime? issuedAt, SigningCredentials signingCredentials)
         {
-            return CreateJwtSecurityTokenPrivate(issuer, audience, subject, notBefore, expires, issuedAt, signingCredentials, null).RawData;
+            return CreateJwtSecurityTokenPrivate(issuer, audience, subject, notBefore, expires, issuedAt, signingCredentials, null, null).RawData;
         }
 
         /// <summary>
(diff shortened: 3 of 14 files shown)


src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -117,6 +117,55 @@ public JwtPayload(string issuer, string audience, IEnumerable<Claim> claims, Dat
...
+        /// values will be overridden.</param>
+        /// <param name="notBefore">If notbefore.HasValue a { nbf, 'value' } claim is added, overwriting any 'nbf' claim in 'claims' and 'claimcollection' if present.</param>
+        /// <param name="expires">If expires.HasValue a { exp, 'value' } claim is added, overwriting any 'exp' claim in 'claims' and 'claimcollection' if present.</param>
+        /// <param name="issuedAt">If issuedAt.HasValue is 'true' a { iat, 'value' } claim is added, overwriting any 'iat' claim in 'claims' and 'claimcollection' if present.</param>
+        /// <remarks>Comparison is set to <see cref="StringComparer.Ordinal"/>
+        /// <para>The 4 parameters: 'issuer', 'audience', 'notBefore', 'expires' take precedence over <see cref="Claim"/>(s) in 'claims' and 'claimcollection'. The values will be overridden.</para></remarks>
+        /// <exception cref="ArgumentException">If 'expires' &lt;= 'notbefore'.</exception>
+        public JwtPayload(string issuer, string audience, IEnumerable<Claim> claims, IDictionary<string, object> claimsCollection, DateTime? notBefore, DateTime? expires, DateTime? issuedAt)
```
RojaEnnam, line 112: Quoting  @brentschmaltz Comments from previous PR
```
we are trying to steer users towards JsonWebToken. I am wondering if we should have this method?
All it means is that JwtSecurityTokenHandler consolidates to 'Claim' before calling this method.
```

RojaEnnam: Originally I implemented it by consolidating dictionary to 'Claim'(s) and then call JwtPayload but, 'AddClaims' method again converts the 'Claim'(s) to dictionary. When this is happening on DateTime objects I observed the original value is not retained.

```
- DateTime.TryParse("2019-11-15T14:31:21.6101326Z") result is "11/15/2019 6:31:21 AM"( which will be the value of Claim after converting dictionary to Claim list)
- After 'AddClaims' Date.TryParse("11/15/2019 6:31:21 AM") results in 2019-11-15T14:31:21.0000000Z instead of 2019-11-15T14:31:21.6101326Z 
```

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -429,7 +455,34 @@ public virtual JwtSecurityToken CreateJwtSecurityToken(SecurityTokenDescriptor t
...
+        /// <para>Each <see cref="Claim"/> on the <paramref name="subject"/> added will have <see cref="Claim.Type"/> translated according to the mapping found in
+        /// <see cref="OutboundClaimTypeMap"/>. Adding and removing to <see cref="OutboundClaimTypeMap"/> will affect the name component of the Json claim.</para>
+        /// <para><see cref="SigningCredentials.SigningCredentials(SecurityKey, string)"/> is used to sign <see cref="JwtSecurityToken.RawData"/>.</para>
+        /// <para><see cref="EncryptingCredentials.EncryptingCredentials(SecurityKey, string, string)"/> is used to encrypt <see cref="JwtSecurityToken.RawData"/> or <see cref="JwtSecurityToken.RawPayload"/> .</para>
+        /// </remarks>
+        /// <returns>A <see cref="JwtSecurityToken"/>.</returns>
+        /// <exception cref="ArgumentException">If 'expires' &lt;= 'notBefore'.</exception>
+        public virtual JwtSecurityToken CreateJwtSecurityToken(string issuer, string audience, ClaimsIdentity subject, DateTime? notBefore, DateTime? expires, DateTime? issuedAt, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, IDictionary<string, object> claimCollection)
```
RojaEnnam, line 482: Quoting @brentschmaltz comment from previous PR
```
we could make a similar argument as for JwtPayload. Do we really need this?
Every public API we ship, we have to support, sometimes less is more.
```

RojaEnnam: My thought process was, If we don't have this, users can use both SecurityTokenDescriptor 'Claims' and 'Subject.Claims' only when they call CreateToken(SecurityTokenDescriptor tokenDescriptor)

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +55,72 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
                     yield return key;
         }
+
+        /// <summary>
+        /// Merges claims. If a claim with same type exists in both <paramref name="claims"/> and <paramref name="subjectClaims"/>, the one in claims will be kept.
+        /// </summary>
+        /// <param name="claims"> Collection of <see cref="Claim"/>'s.</param>
+        /// <param name="subjectClaims"> Collection of <see cref="Claim"/>'s.</param>
+        /// <returns> A Merged list of <see cref="Claim"/>'s.</returns>
+        internal static IEnumerable<Claim> MergeClaims(IEnumerable<Claim> claims, IEnumerable<Claim> subjectClaims)
```
henrik-me, line 64: >internal static IEnumerable<Claim> MergeClaims(IEnumerable<Claim> claims, IEnumerable<Claim> subjectClaims) [](start = 8, length = 107)

Could this be replaced by [https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.sortedset-1.unionwith?view=netcore-3.1](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.sortedset-1.unionwith?view=netcore-3.1) or https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.iset-1.unionwith?view=netframework-4.5.2

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +55,72 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
+
+                if (value.GetType().Name == typeof(DateTime).Name)
+                    return ClaimValueTypes.DateTime;
+
+                if (value.GetType().Name == typeof(List<>).Name)
+                {
+                    foreach (var item in (IList)value)
+                    {
+                        // What to do if the value type of objects in the list are different.
+                        return GetClaimValueTypeFromValue(item);
```
henrik-me: given you are not actually looking to figure out if there are different types perhaps write the code to simply look only at the first element. not sure if it's worth the effort to go through all elements here, depends on what the use case is.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -102,5 +100,62 @@ internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(stri
...
+                object claimValue = claim.Value;
+
+                if (claimValue != null)
+                {
+                    string valueType = TokenUtilities.GetClaimValueTypeFromValue(claimValue);
+
+                    if (claimValue.GetType().Name == typeof(List<>).Name)
+                    {
+                        foreach (var item in (IList)claimValue)
+                            claims.Add(new Claim(claimType, item.ToString(), valueType));
```
henrik-me, line 132: Assume you would have to get the value type for each of the claims? 
if valueType was typeof(List<>).Name) you would not need to look at claimValue.GEtType again?

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
henrik-me: perhaps you should return typeof(List<>).Name) as your call pattern to this has to do special handling here?

---
In reply to: [428331612](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1417#discussion_r428331612) [](ancestors = 428331612)

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -102,5 +100,62 @@ internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(stri
...
+
+        /// <summary>
+        /// Merges <paramref name="claims"/> and <paramref name="subjectClaims"/>
+        /// </summary>
+        /// <param name="claims"> A dictionary of claims.</param>
+        /// <param name="subjectClaims"> A collection of <see cref="Claim"/>'s</param>
+        /// <returns> A merged list of <see cref="Claim"/>'s.</returns>
+        internal static IEnumerable<Claim> GetAllClaims(IDictionary<string, object> claims, IEnumerable<Claim> subjectClaims)
+        {
+            IEnumerable<Claim> allClaims = null;
```
keegan-caruso: It is worth consider if this should ever be allowed to return null.

Null doesn't appear to have much meaning to callers, only if there are items in the returned collection. We can get close to a Null Object Pattern with just returning an empty list.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +55,72 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
+                return subjectClaims;
+
+            if (subjectClaims == null)
+                return claims;
+
+            List<Claim> result = claims.ToList();
+
+            foreach (Claim claim in subjectClaims)
+            {
+                if (!claims.Where(i => i.Type == claim.Type).Any())
```
keegan-caruso: It is slightly more clean and efficient to directly use this lambda passed to Any.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
RojaEnnam: I think we need to get valueType for every item in list and GetClaimValueTypeFromValue returns value.GetType().ToString() if a List contains any object types that are not covered. I will make that change.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -55,5 +55,72 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
+        /// <param name="value"> The <see cref="Claim"/> value.</param>
+        /// <returns> The value type of the <see cref="Claim"/>.</returns>
+        internal static string GetClaimValueTypeFromValue(object value)
+        {
+            if (value != null)
+            {
+                if (value is string)
+                    return ClaimValueTypes.String;
+
+                if (value.GetType().Name == typeof(bool).Name)
```
keegan-caruso: the checks for bool, int, long, double, and Datetime can all use the `is` operator as well.

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -117,6 +117,55 @@ public JwtPayload(string issuer, string audience, IEnumerable<Claim> claims, Dat
...
+            : base(StringComparer.Ordinal)
+        {
+            if (claims != null)
+                AddClaims(claims);
+
+            if (claimsCollection != null && claimsCollection.Count > 0)
```
keegan-caruso: `Any` is a little more clear than checking against Count

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -117,6 +117,55 @@ public JwtPayload(string issuer, string audience, IEnumerable<Claim> claims, Dat
...
+                MergeDictionaryClaims(claimsCollection);
+
+            if (expires.HasValue)
```
keegan-caruso, line 134: This block looks duplicated from above. Consider refactoring so that this block is extracted to a method

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -510,6 +559,20 @@ public void AddClaims(IEnumerable<Claim> claims)
             }
         }
 
+        /// <summary>
+        /// Adds claims from dictionary.
+        /// </summary>
+        /// <param name="claimsCollection"> A dictionary of claims.</param>
+        /// <remark> If a key is already present in target dictionary, its value is overridden by the value of the key in claimsCollection.</remark>
+        internal void MergeDictionaryClaims(IDictionary<string, object> claimsCollection)
```
keegan-caruso: This name is misleading. It doesn't look like merging is done

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -35,6 +35,7 @@
 using Microsoft.IdentityModel.Tokens;
 using Microsoft.IdentityModel.JsonWebTokens;
 using TokenLogMessages = Microsoft.IdentityModel.Tokens.LogMessages;
+using System.Linq;
```
keegan-caruso: System imports should come first

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -1,5 +1,7 @@
 ﻿using Microsoft.IdentityModel.Logging;
 using Microsoft.IdentityModel.Xml;
+using System.Security.Claims;
```
keegan-caruso: System imports should come first

test/Microsoft.IdentityModel.TestUtils/Default.cs
```diff
@@ -407,6 +407,50 @@ public static List<Claim> PayloadClaims
             };
         }
 
+        public static List<Claim> PayloadJsonClaims
+        {
+            get => new List<Claim>
+                    {
```
keegan-caruso, line 413: This should not be indented twice (4 spaces instead of 8)

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
```diff
@@ -40,6 +41,156 @@ namespace System.IdentityModel.Tokens.Jwt.Tests
...
+                        {
+                            SigningCredentials = Default.AsymmetricSigningCredentials,
+                            EncryptingCredentials = null,
+                            Claims = Default.PayloadJsonDictionary,
+                        },
+                        JwtSecurityTokenHandler = tokenHandler,
+                        JsonWebTokenHandler = jsonTokenHandler,
+                        ValidationParameters = Default.AsymmetricSignTokenValidationParameters
+                    },
+                     new CreateTokenTheoryData
```
keegan-caruso, line 155: nit: spacing is off

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
RojaEnnam: I took a look at UnionWith and Union.

1. UnionWith is not available on IEnumerable, to use it we need to cast it to ISet
2. Union is available but we need to provide a comparer. For that 'Claim' class need to implement IEqualityComparer<T> or IEquatable<T>

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
RojaEnnam: Right. Its more like adding.

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
brentschmaltz: @RojaEnnam 
I think we should support 1 level of enumeration for a single item and throw NotSupported for lists of lists (and other collections).

RojaEnnam: @brentschmaltz 
For second level of enumeration should we throw Exception or continue by assigning value.GetType() as valueType?

RojaEnnam: After looking into it a bit more. I think we should throw NotSupported

src/Microsoft.IdentityModel.Tokens.Saml/SamlTokenUtilities.cs
```diff
@@ -102,5 +100,65 @@ internal static IEnumerable<SecurityKey> GetKeysForTokenSignatureValidation(stri
...
+                {
+                    valueType = TokenUtilities.GetClaimValueTypeFromValue(claimValue);
+                    if (valueType == null)
+                    {
+                        if (claimValue is IEnumerable claimList)
+                        {
+                            foreach (var item in claimList)
+                            {
+                                valueType = TokenUtilities.GetClaimValueTypeFromValue(item);
+                                claims.Add(new Claim(claimType, item.ToString(), valueType == null? item.GetType().ToString() : valueType));
```
brentschmaltz: @RojaEnnam 

the valueType must be a valid XSI type { "http://www.w3.org/2001/XMLSchema#string" is assumed}
see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/8ca7558ca2d5c7a27d108d6ecd7c9babf2b30a95/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs#L1639 and https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/8ca7558ca2d5c7a27d108d6ecd7c9babf2b30a95/src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlAttribute.cs#L92

If we add the following to CreateSaml2TokenUsingTokenDescriptor and CreateSamlTokenUsingTokenDescriptor tests, they will throw on creation.

{ "https://wwww.listinlist.com", new List<object>{ new List<object> { "bob", new SecurityTokenDescriptor(), 12, 1.45 }, new List<object> { "bob", new SecurityTokenDescriptor(), 12, 1.45 } } }

We should throw a NotSupportedException for enumerations inside enumerations and leave the valueType to null if we can not resolve it to a well known xsiType from GetClaimValueType. 

Since this type is only for Saml tokens, how about renaming GetClaimValueTypeFromValue to GetXsiTypeForValue and move the method to SamlTokenUtilities.

RojaEnnam: I think that makes it more clear. Pushed the changes.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -382,7 +383,31 @@ public virtual string CreateEncodedJwt(string issuer, string audience, ClaimsIde
...
+        /// <remarks>If <see cref="ClaimsIdentity.Actor"/> is not null, then a claim { actort, 'value' } will be added to the payload. <see cref="CreateActorValue"/> for details on how the value is created.
+        /// <para>See <seealso cref="JwtHeader"/> for details on how the HeaderParameters are added to the header.</para>
+        /// <para>See <seealso cref="JwtPayload"/> for details on how the values are added to the payload.</para>
+        /// <para>Each <see cref="Claim"/> in the <paramref name="subject"/> will map <see cref="Claim.Type"/> by applying <see cref="OutboundClaimTypeMap"/>. Modifying <see cref="OutboundClaimTypeMap"/> could change the outbound JWT.</para>
+        /// </remarks>
+        /// <returns>A Base64UrlEncoded string in 'Compact Serialization Format'.</returns>
+        /// <exception cref="ArgumentException">If 'expires' &lt;= 'notBefore'.</exception>
+        public virtual string CreateEncodedJwt(string issuer, string audience, ClaimsIdentity subject, DateTime? notBefore, DateTime? expires, DateTime? issuedAt, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, IDictionary<string, object> claimCollection)
```
GeoK, line 407: Do we need this public method? It a bit confusing to me. Could a user accomplish the same by calling CreateEncodedJwt(SecurityTokenDescriptor)?

RojaEnnam: Yes, accomplishes same as CreateEncodedJwt(SecurityTokenDescriptor).

Before this change [this](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e4928aff114bf17ee04bd0360b15da6d33694b06/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L383) was also acheiving the same as CreateEncodedJwt(SecurityTokenDescriptor) ? Please correct me if I am wrong.

brentschmaltz:
Squashed and rebased on dev.

Issue #1436, opened 2020-06-09
SamlSerializer fails to validate token using XmlReader created from XDocument


gislikonrad:
I have a really obscure bug. The application I'm working on has need for a non-forward-only reader for XML based tokens. When validating these tokens, I'm using the SecurityTokenHandlers from this repository and sending in an XmlReader that is created by XDocument.CreateReader(). The SamlSecurityTokenHandler is unable to use this reader to validate tokens. It fails to read the attributes in the Assertion. This is not a problem in the Saml2SecurityTokenHandler. The problem is because of how the SamlSerializer reads the attributes in a different manner from the Saml2Serializer.

How SamlSerializer reads its attributes:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/a9a4da0263cdf28d97056159247618040e8d1f87/src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSerializer.cs#L233

How Saml2Serializer reads its attributes:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/a9a4da0263cdf28d97056159247618040e8d1f87/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs#L218

gislikonrad:
I have a [repository](https://github.com/gislikonrad/saml-token-validation-error-repro) that contains a reproduction.

brentschmaltz:
@gislikonrad yes, this is a error on our part.
During development of WsTrust, i coded a different solution for attributes.
When time permits, we will integrate this solution.
For now, I think your PR will fix the issue without any side effects.
see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e019cbea21f186903ad5a17e641d42bb93a856e3/src/Microsoft.IdentityModel.Protocols.WsTrust/XmlAttributeHolder.cs#L32
and: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e019cbea21f186903ad5a17e641d42bb93a856e3/src/Microsoft.IdentityModel.Protocols.WsTrust/WsPolicy/WsPolicySerializer.cs#L85

Thanks!

brentschmaltz:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1437

PR #1437, opened 2020-06-09
fixes issue with SamlSerializer and adds tests

gislikonrad:
This PR fixes an issue #1436 by remove the null value that is passed to XmlReader.GetAttribute in SamlSerializer.

Tests are added to make sure that Saml and Saml2 Assertions can be read using a standard XmlReader and an XDocument created XmlReader.


brentschmaltz:
@gislikonrad i rebased your commit on top of dev just to keep the branch linear.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/fa6f9d4de718f4f488121d0720ce452fb9fe90de

Thanks again.

PR #1441, opened 2020-06-10
Add Analyzers to Logging

keegan-caruso:
Add Roslyn analysis to Logging

### DIFF
--- a/src/Microsoft.IdentityModel.Logging/GlobalSuppressions.cs
+++ b/src/Microsoft.IdentityModel.Logging/GlobalSuppressions.cs
@@ -1,0 +1,10 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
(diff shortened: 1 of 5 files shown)


pmaytak (commented):
I would suggest to look into using .editorconfig instead of suppressions for rules that don't apply to a wide variety of classes. EditorConfig is also hierarchy based, so you can have a solution wide file and then more specific project located ones, etc. Although if you want to suppress for a specific class/method/etc I think you have to use suppressions.

You can take a look at how it's done in [Identity Web](https://github.com/AzureAD/microsoft-identity-web).

Some docs [here](https://docs.microsoft.com/en-us/visualstudio/code-quality/configure-fxcop-analyzers?view=vs-2019) and [here](https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2019).

keegan-caruso:
@pmaytak editorconfig is a great idea. I opened an issue to track it #1442 

We might want to do that for CA1303

But yeah, there are a few things we need to suppress for backwards compatibility, but would like to cover in future work.

brentschmaltz (approved):

PR #1451, opened 2020-06-15
Adds ReadReferences method

### LINKED ISSUE #1449: DSigSerializer is difficult to override for multiple references
The ReadSignedInfo method in DSigSerializer calls the ReadReference method once and adds the Reference returned into the References collection. If another reference is available, then ReadSignedInfo throws an exception.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/brentsch/wstrust/src/Microsoft.IdentityModel.Xml/DsigSerializer.cs#L400

If the check for another reference and exception would be moved to a seperate method called ReadReferences, then that method could be easilly overridden to support multiple references.

gislikonrad:
fixes #1449 

Adds a ReadReferences method that is called by ReadSignedInfo. ReadReferences makes the same checks as were done previously in ReadSignedInfo.

### DIFF
--- a/src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
+++ b/src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.IO;
 using System.Text;
@@ -78,10 +79,7 @@ namespace Microsoft.IdentityModel.Xml
         {
             XmlUtil.CheckReaderOnEntry(reader, XmlSignatureConstants.Elements.KeyInfo, XmlSignatureConstants.Namespace);
 
-            var keyInfo = new KeyInfo
-            {
-                Prefix = reader.Prefix
-            };
+            var keyInfo = CreateKeyInfo(reader);
 
             try
             {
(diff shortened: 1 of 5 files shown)


brentschmaltz:
@gislikonrad this is a good fix as it will maintain the existing behavior, yet allow for extensibility.
I would like to add some tests to see explore behavior with multiple references.

gislikonrad:
@brentschmaltz You mind if I attempt to add some more enhancements to extendability to this class with backwards compatability in mind?

brentschmaltz:
@gislikonrad sure, ill review any changes you make.

brentschmaltz:
@gislikonrad i would like to put this in next months release as we will be working on our WsTrust work and I would like to coordinate with this PR.

gislikonrad:
@brentschmaltz Sounds good. There are a few things I might want to add in the mean time, if I don't run out of time. I'm working on an X509SecurityTokenValidator to be used with my [SOAP hosting extension](https://github.com/SOLIDSoftworks/Solid.Extensions.AspNetCore.Soap) for AspNetCore 2.1 and 3.1. Signatures are a bit different with the BinarySecurityToken than they are in Saml and Saml2.

gislikonrad:
@brentschmaltz So I've been looking into this, but I can't find a good solution. I'm trying to work in WS-Security SecurityTokenReference when reading the KeyInfo of the digital signature. I have a custom implementation that works, just trying to figure out a way to add it to the WS-Trust or Xml libraries.

gislikonrad:
@brentschmaltz I think the few items I need/want to add should go into the WsTrust project. So, for me at least, this could be merged into dev. 

What I want to add is the ability to use a SecurityToken to validate a digital signature. So extending DSigSerializer to read the WS-Security SecurityTokenReference element. Then adding an extension method to Signature named Validate where you can pass it an IEnumerable&lt;SecurityToken&gt;.

brentschmaltz:
@gislikonrad we are looking to release an updated version within a week or so and we would like to get this in. Can you rebase against the current dev branch. Sorry for the extra work as we didn't get this in last release.

gislikonrad:
@brentscmaltz I'll try to get it done this weekend.

gislikonrad:
@brentschmaltz The commits have been rebased and squashed.

brentschmaltz:
@gislikonrad thanks!

brentschmaltz:
@gislikonrad if you can rebase this against current dev, we will get this into the next release.

src/Microsoft.IdentityModel.Xml/Reference.cs
```diff
@@ -161,13 +161,10 @@ private static byte[] ProcessAndDigest(XmlTokenStream tokenStream, HashAlgorithm
...
-                using (var writer = XmlDictionaryWriter.CreateTextWriter(Stream.Null))
-                {
-                    tokenStream.WriteTo(writer);
-                    writer.Flush();
-                }
+                using (var writer = XmlWriter.Create(stream, new XmlWriterSettings { CloseOutput = false }))
+                using (var dictionaryWriter = XmlDictionaryWriter.CreateDictionaryWriter(writer))
+                    tokenStream.WriteTo(dictionaryWriter);
 
-                stream.Flush();
```
brentschmaltz, line 170: why are the two Flush() calls removed?

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -174,13 +135,176 @@ public virtual KeyInfo ReadKeyInfo(XmlReader reader)
...
         }
 
+
+        /// <summary>
+        /// Creates a new <see cref="KeyInfo"/> object.
+        /// </summary>
+        /// <param name="reader">The current <see cref="XmlReader"/>.</param>
+        protected virtual KeyInfo CreateKeyInfo(XmlReader reader)
+        {
+            XmlUtil.CheckReaderOnEntry(reader, XmlSignatureConstants.Elements.KeyInfo, XmlSignatureConstants.Namespace);
```
brentschmaltz, line 129: This is going to get called twice now.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -101,10 +102,7 @@ public virtual KeyInfo ReadKeyInfo(XmlReader reader)
         {
             XmlUtil.CheckReaderOnEntry(reader, XmlSignatureConstants.Elements.KeyInfo, XmlSignatureConstants.Namespace);
 
-            var keyInfo = new KeyInfo
-            {
-                Prefix = reader.Prefix
-            };
+            var keyInfo = CreateKeyInfo(reader);
```
brentschmaltz, line 82: Is there a lot of value in creating a method here?

brentschmaltz:
@gislikonrad i think there is value here and we can get this in.
There are some conflicts that need to be resolved.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
gislikonrad: ~Maybe not. It might be a better idea to make the `TryReadKeyInfoType` have an out parameter for keyInfo rather than a ref parameter.~

It has merrit, I have another project where I have an overridden `KeyInfo` that has more properties. I had to override this whole method to handle that, but wouldn't have needed to if my PR had been pulled in earlier.

https://github.com/SOLIDSoftworks/Solid.Identity.Protocols.WsTrust/blob/e3e677abd802962c404d7bb377e7359c3671a1f8/src/Solid.Identity.Protocols.WsTrust/Protocols/WsSecurity/Xml/WsUtilityDSigSerializer.cs#L80

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
gislikonrad: Yeah, that was an oversight when this was done originally.

src/Microsoft.IdentityModel.Xml/Reference.cs
gislikonrad: I assumed it was ok, since XmlTextWriters auto flush when disposed.

gislikonrad:
I've rebased my branch on your dev branch and answered the comments.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
gislikonrad: Actually, when removing this check, I get a warning.

> CA1062: In externally visible method 'KeyInfo DSigSerializer.CreateKeyInfo(XmlReader reader)', validate parameter 'reader' is non-null before using it. If appropriate, throw an ArgumentNullException when the argument is null or add a Code Contract precondition asserting non-null argument.

That's probably why I had it in there. It has been 3 years... 😄

brentschmaltz: Yeah, 3 years is too long.
Let's leave it for now.

src/Microsoft.IdentityModel.Xml/Reference.cs
brentschmaltz: @gislikonrad 
I tend to assume that sometimes the 'Flush' is not always consistent with different writers and like to be consistent.
For example, I have found Utf8JsonWriter and MemoryStreams have to be managed in certain order when calling Flush.

If you can add the Flush back on the dictionaryWriter, we should be good to go.

gislikonrad: I'll re-add the flush.

I am, genuinely confused as to what this method did originally. Writing the token stream to a null stream and flushing the writer, then flushing an empty memory stream and hashing nothing?

gislikonrad:
The missing flush of the dictionary writer has been re-added.

brentschmaltz:
@gislikonrad we will try and be faster next time :-)
Thanks!

brentschmaltz (approved):

PR #1454, opened 2020-06-16
Add analyzers to Tokens assembly

keegan-caruso:


brentschmaltz:
I opened: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1459 as we need to fix these.

brentschmaltz:
@keegan-caruso i think we can fix some of our ArgumentOutOfRangeException errors by passing the parameter to the exception and / or slightly modifying the error message.

For example: 
IDX10653 in AuthenticatedEncryptionProvider.ValidateKeySize could take nameof(key) instead of "key.KeySize".

IDX10675 in ECDsaAdapter.CreateECDsaUsingCNGKey would need a slight change.
"IDX10675: The byte count of '{0}' must be less than or equal to '{1}', but was {2}.";
=>
"IDX10675: Cannot create a ECDsa object from the '{0}',  the bytes from the decoded value of '{1}' must be less than the size associated with the curve: '{2}'. Size was: '{3}'.";

keegan-caruso:
@brentschmaltz  good idea. No reason to wait on those.

brentschmaltz:
@keegan-caruso i agree we should address any larger issues post getting the analyzers in.

brentschmaltz (approved):
I added issue: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1461 for recording any possible changes.

PR #1456, opened 2020-06-17
Add analyzers to Tokens.Saml

keegan-caruso:


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Action.cs
```diff
@@ -81,7 +81,7 @@ public Uri Namespace
                     throw LogArgumentNullException(nameof(value));
 
                 if (!value.IsAbsoluteUri)
-                    throw LogExceptionMessage(new ArgumentException(nameof(value), FormatInvariant(LogMessages.IDX13300, value)));
+                    throw LogExceptionMessage(new ArgumentException(FormatInvariant(LogMessages.IDX13300, value), nameof(value)));
```
keegan-caruso, line 84: This is backwards. Should be (message, paramName)

https://docs.microsoft.com/en-us/dotnet/api/system.argumentexception.-ctor?view=netcore-3.1#System_ArgumentException__ctor_System_String_System_String_

brentschmaltz (approved):

PR #1458, opened 2020-06-17
Fix usage of IDX13300 and IDX13107

### LINKED ISSUE #1455: Usage of IDX13300 and IDX13107 is incorrect
IDX13300 requires 2 arguments, It is passed into many FormatInvariant calls with only one argument.

IDX13107 in Saml2Serializer.ReadSimpleUriElement is used with 2 arguments and requires 3

Instead of raising an ArgumentException, it will raise a FormatException

### LINKED ISSUE #1457: Should AuthnContextClassRef  be required?
The zero parameter constructor for Saml2AuthenticationContext passes null as a ClassReference.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthenticationContext.cs#L56

ClassReference does not accept a null value

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthenticationContext.cs#L104

Saml2Core, 2.7.2.2 lists AuthnContextClassRef as optional. Is there a reason behavior deviates from Saml2Core? Is there an overriding errata? The zero parameter constructor should either be removed or the ClassReference property should accept null depending on the answer

keegan-caruso:
fixes #1455

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthenticationContext.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthenticationContext.cs
@@ -53,7 +53,6 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
         /// Creates an instance of Saml2AuthenticationContext.
         /// </summary>
         public Saml2AuthenticationContext()
-            : this(null, null)
         {
         }
 
@@ -62,8 +61,8 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
         /// </summary>
         /// <param name="classReference">The class reference of the authentication context.</param>
         public Saml2AuthenticationContext(Uri classReference)
-            : this(classReference, null)
         {
+            ClassReference = classReference;
         }
 
         /// <summary>
@@ -73,15 +72,8 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
         /// <param name="declarationReference">The declaration reference of the authentication context.</param>
         public Saml2AuthenticationContext(Uri classReference, Uri declarationReference)
         {
-            // Class and declaration reference are optional, so don't try and set the respective properties if
-            // the parameters are null.
-            if (classReference != null)
-                ClassReference = classReference;
-
-            if (declarationReference != null)
-                DeclarationReference = declarationReference;
-
-            AuthenticatingAuthorities = new List<Uri>();
+            ClassReference = classReference;
+            DeclarationReference = declarationReference;
         }
 
         /// <summary>
@@ -93,7 +85,7 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
         public ICollection<Uri> AuthenticatingAuthorities
         {
             get;
-        }
+        } = new List<Uri>();
 
         /// <summary>
         /// Gets or sets a URI reference identifying an authentication context class that 
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthenticationContext.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthenticationContext.cs
@@ -73,8 +73,14 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
         /// <param name="declarationReference">The declaration reference of the authentication context.</param>
         public Saml2AuthenticationContext(Uri classReference, Uri declarationReference)
         {
-            ClassReference = classReference;
-            DeclarationReference = declarationReference;
+            // Class and declaration reference are optional, so don't try and set the respective properties if
+            // the parameters are null.
+            if (classReference != null)
+                ClassReference = classReference;
+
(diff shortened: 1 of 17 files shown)


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthenticationContext.cs
```diff
@@ -122,7 +122,7 @@ public Uri DeclarationReference
             set
             {
                 if (value != null && !value.IsAbsoluteUri)
-                    throw LogExceptionMessage(new ArgumentException(FormatInvariant(LogMessages.IDX13300, nameof(value), value)));
+                    throw LogExceptionMessage(new ArgumentException(FormatInvariant(LogMessages.IDX13300, nameof(DeclarationReference), value)));
```
brentschmaltz, line 126: @keegan-caruso could you fix up the difference between ClassReference and DeclarationReference where ClassReference throws ArgumentNullException when value == null. I would vote for differentiating null.

keegan-caruso: @brentschmaltz I'm happy to normalize the two. I am wondering if this goes back to the Saml2Core spec. Section 2.7.2.2 calls out both as being optional. That seems to imply neither throw on null. This looks safe as all access to this field checks for null conditions.

brentschmaltz: @keegan-caruso in rethinking this, the ctors seem to be fine. Given the spec.
The issue seems to be that the ctors try and set a null value to the property.

It's OK that the value is null, but the reason we throw when setting the property is that seems to indicate a programming error.

So, perhaps the error is the constructor trying to set a null value.

keegan-caruso: This is a good idea. It also fixes #1457

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthenticationContext.cs
```diff
@@ -73,8 +73,14 @@ public Saml2AuthenticationContext(Uri classReference)
         /// <param name="declarationReference">The declaration reference of the authentication context.</param>
         public Saml2AuthenticationContext(Uri classReference, Uri declarationReference)
         {
-            ClassReference = classReference;
-            DeclarationReference = declarationReference;
+            // Class and declaration reference are optional, so don't try and set the respective properties if
```
brentschmaltz: I would recommend a slightly different pattern.
We have three ctors.
Check the parameters as we did before, but call a private method to initialize where we know the parameters were checked.
Otherwise we are not helping devs by helping them catch programming errors.

Here is an example here where we know in the private method we checked parameters in the public  methods.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/2cb5b3e600b0e9f978f1d468f096e17fd5c1bc5f/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L454

keegan-caruso: I took a slightly different approach, but hopefully captured the spirit of what your feedback was addressing.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
```diff
@@ -659,7 +659,10 @@ protected virtual Saml2AuthenticationContext ReadAuthenticationContext(XmlDictio
                     reader.ReadStartElement(Saml2Constants.Elements.AuthnContextDeclRef, Saml2Constants.Namespace);
 
                 // Now we have enough data to create the object
-                var authnContext = new Saml2AuthenticationContext(classRef, declRef);
+                var authnContext = new Saml2AuthenticationContext(classRef);
```
keegan-caruso, line 662: This is to keep the behavior the same. Previously, classRef was always passed in and always validated as non-null. declRef was allowed to be null.

brentschmaltz: Makes sense.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthenticationContext.cs
brentschmaltz: Yep, looks good.

brentschmaltz:
:shipit:

PR #1464, opened 2020-06-22
Add editorconfig

### LINKED ISSUE #1442: Add editorconfig
Use [this](https://github.com/dotnet/runtime/blob/master/.editorconfig) as a starting point.

It would be good to codify the style used in this repository. Will also give an good spot to do cross-project analysis configuration.

keegan-caruso:
Add editorconfig and move global analyzer setting to editorconfig

Fixes #1442


.editorconfig
```diff
@@ -0,0 +1,191 @@
...
+# name all constant fields using PascalCase
+dotnet_naming_rule.constant_fields_should_be_pascal_case.severity = suggestion
+dotnet_naming_rule.constant_fields_should_be_pascal_case.symbols  = constant_fields
+dotnet_naming_rule.constant_fields_should_be_pascal_case.style    = pascal_case_style
+dotnet_naming_symbols.constant_fields.applicable_kinds   = field
+dotnet_naming_symbols.constant_fields.required_modifiers = const
+dotnet_naming_style.pascal_case_style.capitalization = pascal_case
+
+# static fields should have s_ prefix
+dotnet_naming_rule.static_fields_should_have_prefix.severity = suggestion
```
keegan-caruso, line 62: This guidance comes from [here](https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/coding-style.md)

In general it looks like we don't follow this, though. Do we want to keep the editor suggestion or not?

brentschmaltz (approved):
thanks

.editorconfig
GeoK: How many warnings are you seeing? How far are we from the .net core guidance?

keegan-caruso: Currently, "s_" prefix violations are at 144. But most of those are coming from opensource/json and test assemblies. 
After ignoring those two sources we are down to 48 I will write up some non root editorconfig files for those directories

GeoK: We should disable analyzers and code style formatting for opensource/json as we are not maintaining that code.

I'm debating if we should fix the violations and start enforcing the rules? 
@brentschmaltz - Do you have an option about this?

brentschmaltz: @GeoK @keegan-caruso I don't think we should change any internal json code.
But i like to see what warnings they issue, so i am OK with having the BIG list of exceptions as long as we can easily distinguish between the two groups.

keegan-caruso: fixed in a569748 Don't show any style warnings for tests or opensource/json.
Bringing the suggestions down to ~400.

Some of these would be good to fix (static prefix, don't qualify access with `this`) some of them are more in the realm of polish (simplifying object/collection initialization, use pattern matching, etc)

Issue #1467, opened 2020-06-24
JwtPayload uses current culture to parse claim values


gislikonrad:
Running the tests in the solution on my machine (I live in Iceland), the [RoundTrip ](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/0567929e7c7f4a299925f16a26d43ad5f8c60dc1/test/System.IdentityModel.Tokens.Jwt.Tests/JwtPayloadTest.cs#L211) tests fail because of the value of the **ClaimValueTypes.Double** claim. This is due to the value of **123.4** being changed to **1234** when it's being parsed from string using the **is-IS** culture. I'm sure other cultures could have this error aswell.

brentschmaltz:
@gislikonrad thanks, we will fix this up.

brentschmaltz:
@gislikonrad  fixed with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/8433d584bf3890894fd9b8f3ceef156d0e16e47a

PR #1468, opened 2020-06-24
Uses CultureInfo.InvariantCulture to parse claim values

gislikonrad:
Fixes issue #1467 by using CultureInfo.InvariantCulture to parse number claim values.


brentschmaltz:
@gislikonrad there are other instances where we parse without considering culture.

gislikonrad:
The test referenced in the original issue was the only one failing for me because of parsing using an implicit culture. 

I've seen this all to many times in my career and it's usually because of how numbers are parsed. The differences between the thousands seperator and decimal seperator are the main offenders.

brentschmaltz:
@gislikonrad i rebased this against dev and committed.

PR #1470, opened 2020-06-25
target lts version for test runtime

keegan-caruso:


brentschmaltz (approved):
So weird that someone changed the agent.....
We can check this in, but should define our own agent setup in the future.

Issue #1476, opened 2020-06-29
Microsoft.IdentityModel.Tokens.Saml2.Saml2Serializer Write Signature if included in Assertion


alexjebens:
If I Serialize a Saml Assertion to store and use later using Microsoft.IdentityModel.Tokens.Saml2.Saml2Serializer when I try to write this assertion again to an XmlWriter the Signature is left out because I am not creating a new signature but want to use the existing one.

Also for correctness reading an assertion, writing it and reading it again should be idempotent. That is currently not the case as the written assertion will not contain the signature.

brentschmaltz:
@alexjebens reading and writing the assertion is tricky due to the transforms that are applied and the xmlreader that is used.
Internally we have a property 'CannonicalString' that maintains a version of the token that can be 'reused'.

Would this be of help if this property was public?

alexjebens:
@brentschmaltz you are correct that this would help. It is all I would really need.

I still feel that it is inconsistent that reserializing using the Saml2Serializer modifies the string, however that can be a separate issue.

brentschmaltz:
@alexjebens it's not always possible to determine what the original xml was. As it moves through hops, rules of xml allow it to change to equivalent forms. Kind of a big equivalence class. Extra whitespace, normalization of newlines (/r/n -> /n), reordering of attributes, etc, ...

This is one reason the signature of a JsonWebToken  (JWS) is over the base64urlencoded octets. The original json is available as created by the issuer.

This is why we save the CannonicalString as per the transformations specified in the SignedInfo.
This allows us to use the token in the traditional On-Behalf-Of flows.

When reading the xml, we remember the XML nodes in a 'XmlTokenStream' see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/a22c4f474b806c7c90f6dd9b5fd371fd55db51d4/src/Microsoft.IdentityModel.Xml/XmlTokenStream.cs#L39

We will make access to these public in a future release.

brentschmaltz:
@alexjebens the Saml2Serializer does not change the inbound string.
I am not clear on your ask, if you need something else please open a new issue.

PR #1477, opened 2020-06-29
Write Assertion.Signature if it exists and SigningCredentials were not used

alexjebens:
#1476


alexjebens:
Making the Saml2Assertion CanonicalString public is sufficient and more correct.

PR #1480, opened 2020-06-29
Allow Saml2AuthenticationContext.DeclarationReference to be null

### LINKED ISSUE #1478: Breaking change in 6.7.0: Saml2AuthenticationContext.DeclarationReference no longer allows null values
See https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/0567929e7c7f4a299925f16a26d43ad5f8c60dc1#diff-d904bfa7b70e16ce90571f9a418a9362L123-L128, from #1458. 

Even if this is intentional (which I'm not sure about, given that the previous version explicitly allowed `null`s), 6.7.0 should be retracted and republished as 7.0.0 to be semver-compliant (as recommended by your own docs: https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/versioning).

/cc @catn8r

brentschmaltz:
Customer noted this would break some users code.
We are reverting to previous behavior.


test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2AuthenticationContextTests.cs
```diff
@@ -61,8 +61,7 @@ public void Saml2AuthenticationContext_CtorNullClassAndDeclaration_ArgumentNullE
         [Fact]
         public void Saml2AuthenticationContext_CtorNullDeclaration_ArgumentNullException()
```
keegan-caruso, line 62: name on these test methods no longer match

keegan-caruso (approved):

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2AuthenticationContextTests.cs
brentschmaltz: Thanks, fixed.

PR #1487, opened 2020-07-15
Deduplicated code from LogHelper.RemovePII

tkhadimullin:
It appears to be a pretty obvious case of DRY


brentschmaltz:
@tkhadimullin i see that that code was changed so i am closing out this PR.

PR #1488, opened 2020-07-16
Improve Empty audience error message

### LINKED ISSUE #1410: Improve error message when 'aud' claim is empty
I'm currently attempting to verify a JWT received from Cognito, which [does not include the aud claim](https://stackoverflow.com/q/53148711/10100823).  As [the aud claim is optional](https://tools.ietf.org/html/rfc7519#section-4.1.3), this should be fine.  However, I'm receiving the error:

```
Error validating identity token: Microsoft.IdentityModel.Tokens.SecurityTokenInvalidAudienceException: IDX10214: Audience validation failed. Audiences: 'empty'. Did not match: validationParameters.ValidAudience: '[CENSORED]' or validationParameters.ValidAudiences: 'null'.
   at Microsoft.IdentityModel.Tokens.Validators.ValidateAudience(IEnumerable`1 audiences, SecurityToken securityToken, TokenValidationParameters validationParameters)
   at System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.ValidateTokenPayload(JwtSecurityToken jwtToken, TokenValidationParameters validationParameters)
```

It looks like [that error is being thrown here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/f5628e689e97b80fba3efe81fc33903076c97ad5/src/Microsoft.IdentityModel.Tokens/Validators.cs#L126), and 'empty' is just what`SerializeAsSingleCommaDelimitedString` gives back for [an empty list](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/f5628e689e97b80fba3efe81fc33903076c97ad5/src/Microsoft.IdentityModel.Tokens/Utility.cs#L45).  Per discussion on #1402 , I'm disabling `validationParameters.ValidateAudience` as a workaround.

As the default PII settings won't show the 'empty' `aud` by default, the error message makes it seem as though something is being set as a default value, and adds some digging to figure out what's going on.  A clearer error message when `aud` is required but not present would be appreciated.

An example could just be "missing required claim: aud"

keegan-caruso:
Fixes #1410

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validators.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validators.cs
@@ -112,6 +112,11 @@ namespace Microsoft.IdentityModel.Tokens
             if (string.IsNullOrWhiteSpace(validationParameters.ValidAudience) && (validationParameters.ValidAudiences == null))
                 throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidAudienceException(LogMessages.IDX10208) { InvalidAudience = Utility.SerializeAsSingleCommaDelimitedString(audiences) });
 
+            if (!audiences.Any())
+                throw LogHelper.LogExceptionMessage(
+                    new SecurityTokenInvalidAudienceException(LogHelper.FormatInvariant(LogMessages.IDX10206))
+                    { InvalidAudience = Utility.SerializeAsSingleCommaDelimitedString(audiences) });
+
             // create enumeration of all valid audiences from validationParameters
             var validationParametersAudiences = validationParameters.ValidAudiences == null
                 ? new [] { validationParameters.ValidAudience }
--- a/src/Microsoft.IdentityModel.Tokens/Validators.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validators.cs
@@ -30,7 +30,6 @@ using System.Collections.Generic;
 using System;
 using System.Linq;
 using System.Security.Cryptography.X509Certificates;
-using System.Globalization;
 
 namespace Microsoft.IdentityModel.Tokens
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -129,40 +146,58 @@ public static void ValidateAudience(IEnumerable<string> audiences, SecurityToken
...
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        private static bool AudiencesMatchIgnoringTrailingSlash(string tokenAudience, string validAudience)
+        {
+            int length;
```
RojaEnnam: length can be assigned -1 here itself ?

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -129,40 +146,58 @@ public static void ValidateAudience(IEnumerable<string> audiences, SecurityToken
...
-                    { InvalidAudience = Utility.SerializeAsSingleCommaDelimitedString(audiences) });
+            return false;
+        }
+
+        private static bool AudiencesMatch(TokenValidationParameters validationParameters, string tokenAudience, string validAudience)
+        {
+            if (validAudience.Length == tokenAudience.Length)
+            {
+                if (string.Equals(validAudience, tokenAudience, StringComparison.Ordinal))
+                {
```
RojaEnnam: nit: remove braces.

src/Microsoft.IdentityModel.Tokens/Validators.cs
keegan-caruso: Good call, that is easier to read. Originally just taking a naïve conversion from the nested ternary

RojaEnnam (approved):

GeoK (approved):

PR #1494, opened 2020-07-21
Dont call overridable methods in ctors

### LINKED ISSUE #1459: Do not call overridable methods in constructors
We recently added analyzers that show we are calling virtual methods inside constructors.
This could lead to unexpected results.

https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2214?view=vs-2019 

We need to:
1. Fix these issues
2. Remove the exclusion from GlobalSuppressions.cs

keegan-caruso:
Fixes 1459


src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
```diff
@@ -66,13 +66,13 @@ public SymmetricKeyWrapProvider(SecurityKey key, string algorithm)
...
 
-            if (!IsSupportedAlgorithm(key, algorithm))
+            if (!SupportedAlgorithms.IsSupportedSymmetricKeyWrap(algorithm, key))
                 throw LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant(LogMessages.IDX10661, algorithm, key)));
 
             Algorithm = algorithm;
             Key = key;
 
-            _symmetricAlgorithm = GetSymmetricAlgorithm(key, algorithm);
+            _symmetricAlgorithm = GetSymetricAlgorithmInternal(key, algorithm);
```
GeoK: I agree that this represents bad design and a code smell, but we might break some users who are leveraging this approach. Wondering if we could instead resolve an algorithm once (before use) and not invoke the method from the constructor? This seems to be harder to accomplish with AsymmetricSignatureProvider, though.

PR #1497, opened 2020-07-28
Fix errors from Sonar

### LINKED ISSUE #1503: Fix findings from Sonar static analysis
Sonar analyzers found a few issues that should be fixed

See: https://sonarcloud.io/dashboard?id=AzureAD_azure-activedirectory-identitymodel-extensions-for-dotnet

keegan-caruso:
Fixes these findings: https://sonarcloud.io/project/issues?id=AzureAD_azure-activedirectory-identitymodel-extensions-for-dotnet&resolved=false&types=BUG

Fixed, simplified, and added tests for `Equals` and `GetHashCode` to data transfer objects in M.IM.Xml assembly. Generated methods for `Equals` and `GetHashCode` then edited to respect previous behavior for string casing, and expected behavior for collection types for `Equals`. Added `unchecked` for `GetHashCode` as it was another Sonar finding

Removed null checks on value type objects

Moved null check before first access

Fixed NRE condition when calling `GetType` on a null object

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -1096,8 +1096,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
         private TokenValidationResult ValidateTokenPayload(JsonWebToken jsonWebToken, TokenValidationParameters validationParameters)
         {
-            var expires = (jsonWebToken.ValidTo == null) ? null : new DateTime?(jsonWebToken.ValidTo);
-            var notBefore = (jsonWebToken.ValidFrom == null) ? null : new DateTime?(jsonWebToken.ValidFrom);
+            var expires = jsonWebToken.ValidTo;
+            var notBefore = jsonWebToken.ValidFrom;
 
             Validators.ValidateLifetime(notBefore, expires, jsonWebToken, validationParameters);
             Validators.ValidateAudience(jsonWebToken.Audiences, jsonWebToken, validationParameters);
@@ -1138,7 +1138,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                 var validatedJsonWebToken = validatedToken as JsonWebToken;
                 if (validatedJsonWebToken == null)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10506, typeof(JsonWebToken), validatedJsonWebToken.GetType(), token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10506, typeof(JsonWebToken), validatedToken.GetType(), token)));
 
                 return validatedJsonWebToken;
             }
@@ -1220,7 +1220,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                             LogHelper.LogInformation(TokenLogMessages.IDX10242, token);
                             jwtToken.SigningKey = key;
                             return jwtToken;
-                        };
+                        }
                     }
                     catch (Exception ex)
                     {
--- a/src/Microsoft.IdentityModel.Tokens/CallContext.cs
+++ b/src/Microsoft.IdentityModel.Tokens/CallContext.cs
@@ -48,9 +48,6 @@ namespace Microsoft.IdentityModel.Tokens
         /// </summary>
         public CallContext(Guid activityId)
         {
-            if (activityId == null)
-                throw new ArgumentNullException(nameof(activityId));
-
             ActivityId = activityId;
         }
 
--- a/src/Microsoft.IdentityModel.Xml/IssuerSerial.cs
+++ b/src/Microsoft.IdentityModel.Xml/IssuerSerial.cs
@@ -65,10 +65,13 @@ namespace Microsoft.IdentityModel.Xml
         /// <inheritdoc/>
         public override int GetHashCode()
         {
-            int hashCode = -1073543679;
-            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(IssuerName);
-            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(SerialNumber);
-            return hashCode;
+            unchecked
+            {
+                int hashCode = -1073543679;
+                hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(IssuerName);
+                hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(SerialNumber);
+                return hashCode;
+            }
         }
     }
 }
--- a/src/Microsoft.IdentityModel.Xml/IssuerSerial.cs
+++ b/src/Microsoft.IdentityModel.Xml/IssuerSerial.cs
@@ -26,6 +26,7 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 
 namespace Microsoft.IdentityModel.Xml
 {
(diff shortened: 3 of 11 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1096,8 +1096,8 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
 
         private TokenValidationResult ValidateTokenPayload(JsonWebToken jsonWebToken, TokenValidationParameters validationParameters)
         {
-            var expires = (jsonWebToken.ValidTo == null) ? null : new DateTime?(jsonWebToken.ValidTo);
-            var notBefore = (jsonWebToken.ValidFrom == null) ? null : new DateTime?(jsonWebToken.ValidFrom);
```
henrik-me, line 1100: Any idea what the reason was behind doing the datetime conversion before?

src/Microsoft.IdentityModel.Tokens/CallContext.cs
```diff
@@ -48,9 +48,6 @@ public CallContext()
         /// </summary>
         public CallContext(Guid activityId)
         {
-            if (activityId == null)
-                throw new ArgumentNullException(nameof(activityId));
```
henrik-me, line 52: should we instead set it to empty or skip setting the value? Why potentially setting it to null? Also not sure why it's not ok to throw here, if the developer doesn't want to set the activity id they should use the base ctr?

henrik-me:
IMO, a few tricks are used for the hashcode, and I'm assuming that the updates for where these objects are used are properly tested with the added tests. What I might be missing for the utests is some usages of the objects in lists/collections for equals/compare to ensure things works as expected there.

src/Microsoft.IdentityModel.Tokens/CallContext.cs
keegan-caruso: Guid is a value type, so it could never be null, This came up from sonar, marked line 52 as non-reachable code

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
keegan-caruso: Similar to the Guid comment, DateTime is a value type.

keegan-caruso: As far as the conversion from `DateTime` to `DateTime?`, looks to only be to explicitly match the signature of some of the validators, which is not needed.

src/Microsoft.IdentityModel.Xml/IssuerSerial.cs
```diff
@@ -53,27 +54,24 @@ public IssuerSerial(string issuerName, string serialNumber)
...
-        /// <summary>
-        /// Serves as a hash function for IssuerSerial.
-        /// </summary>
+        /// <inheritdoc/>
         public override int GetHashCode()
         {
-            return base.GetHashCode();
+            unchecked
+            {
+                int hashCode = -1073543679;
```
GeoK, line 70: I'm guessing that this code is emitted by VS, right?

src/Microsoft.IdentityModel.Xml/IssuerSerial.cs
```diff
@@ -53,27 +54,24 @@ public IssuerSerial(string issuerName, string serialNumber)
...
         }
 
-        /// <summary>
-        /// Serves as a hash function for IssuerSerial.
-        /// </summary>
+        /// <inheritdoc/>
         public override int GetHashCode()
         {
-            return base.GetHashCode();
+            unchecked
```
GeoK, line 68: just learned about _unchecked_. It fits great in this scenario.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1138,7 +1138,7 @@ private static JsonWebToken ValidateSignature(string token, TokenValidationParam
 
                 var validatedJsonWebToken = validatedToken as JsonWebToken;
                 if (validatedJsonWebToken == null)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10506, typeof(JsonWebToken), validatedJsonWebToken.GetType(), token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10506, typeof(JsonWebToken), validatedToken.GetType(), token)));
```
GeoK, line 1141: Great catch. Was this flagged by Sonar too?

src/Microsoft.IdentityModel.Xml/X509Data.cs
```diff
@@ -120,31 +120,24 @@ public string CRL
...
-        /// <summary>
-        /// Serves as a hash function for X509Data.
-        /// </summary>
+        /// <inheritdoc/>
         public override int GetHashCode()
         {
-            return base.GetHashCode();
+            unchecked
+            {
+                // Certificates is the only immutable property
```
GeoK, line 138: Not sure why this method was originally there, but also not sure why we decided to change it. Was this flagged too? What are the benefits of the new implementation?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
keegan-caruso: Correct,

https://sonarcloud.io/project/issues?id=AzureAD_azure-activedirectory-identitymodel-extensions-for-dotnet&open=AXNs_zszUUMedrjWA3Lk&resolved=false&types=BUG

src/Microsoft.IdentityModel.Xml/X509Data.cs
keegan-caruso: Sonar has a good write up on the finding

https://rules.sonarsource.com/csharp/type/Bug/RSPEC-3249

Here is the finding:

https://sonarcloud.io/project/issues?id=AzureAD_azure-activedirectory-identitymodel-extensions-for-dotnet&open=AXNs_zk7UUMedrjWA3JI&resolved=false&types=BUG

ms docs has good remarks as well
https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=netcore-3.1#remarks

src/Microsoft.IdentityModel.Xml/IssuerSerial.cs
keegan-caruso: Correct. It didn't originally add `unchecked` though.

src/Microsoft.IdentityModel.Xml/X509Data.cs
GeoK: Thanks. I'm still a bit confused about this.
_EqualityComparer<ICollection<string>>.Default.GetHashCode(Certificates)_ returns a different hash code for each X509Data object, right? This implementations satisfies the guideline that GetHashCode function should return the same hash code over the lifetime of the object (immutability), but the following is not true: _Two objects that are equal return hash codes that are equal._

If both arguments above are true, it seems like the is no perfect solution and that we are choosing to be back-compat.

keegan-caruso: GetHashCode with classes that contain collections that are not set by reference in the ctor end up with this limitation. From the GetHashCode perspective, the object is seen as not equal. 

That is my understanding at least.

GeoK (approved):

src/Microsoft.IdentityModel.Xml/X509Data.cs
keegan-caruso: You do raise a good point. There is a slight behavioral change in collections that use the hash code

brentschmaltz: This is interesting as the recommendation is that if obj1.HashCode() ! obj2.HashCode() => obj1 != obj2. But this will not be true, right?

src/Microsoft.IdentityModel.Xml/IssuerSerial.cs
brentschmaltz: @keegan-caruso what is the purpose of the multiplication step, how is that different than a fix number?
With the fixed number, what if two objects have the same "IssuerName" and use the same 'magic' number? A follow up is needed to determine 'AreEqual'?

src/Microsoft.IdentityModel.Xml/X509Data.cs
keegan-caruso: @brentschmaltz Correct. I don't think we can get there in this case though. Functionally this is somewhat similar to the previous behavior. Every X509Data object will have a different hashcode

src/Microsoft.IdentityModel.Xml/IssuerSerial.cs
keegan-caruso: This was generated code. Honestly, was just relying on the tool to create what we hope is a good distribution. 

And yes, I believe a follow up call (Equals) will always be required. HashCode should not be used to determine equality.

PR #1500, opened 2020-07-31
Serialize entropy, if present, on an RST

mjrousos:


brentschmaltz:
We added the code directly as this needed to be rebased.

PR #1505, opened 2020-08-12
Add Claims property to TokenValidationResult

### LINKED ISSUE #1194: Add property bag to TokenValidationResult
In the new world, it seems `ClaimsIdentity` is no longer king, as `JsonWebTokenHandler` exclusively uses `SecurityTokenDescriptor.Claims` (an "untyped" dictionary) instead of `SecurityTokenDescriptor.Subject`.

Would it make sense to have a property in `TokenValidationResult` exposing an equivalent to `SecurityTokenDescriptor.Claims`?

RojaEnnam:
Fixes #1194 

- Adds `Claims` property, that is equivalent to `SecurityTokenDescriptor.Claims` on `TokenValidationResult`.
- Sets `Claims` on `TokenValidationResult` after token validation.


test/Microsoft.IdentityModel.Tokens.Tests/TokenValidationResultTests.cs
```diff
@@ -45,8 +45,8 @@ public void GetSets()
             TokenValidationResult tokenValidationResult = new TokenValidationResult();
             Type type = typeof(TokenValidationResult);
             PropertyInfo[] properties = type.GetProperties();
-            if (properties.Length != 7)
-                Assert.True(false, "Number of public fields has changed from 7 to: " + properties.Length + ", adjust tests");
+            if (properties.Length != 8)
```
keegan-caruso, line 48: `PropertyNamesAndSetGetValue` looks like it is missing a new entry

src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
```diff
@@ -35,6 +36,11 @@ namespace Microsoft.IdentityModel.Tokens
     /// </summary>
     public class TokenValidationResult
     {
+        /// <summary>
+        /// The <see cref="Dictionary{TKey, TValue}"/> created from the validated security token.
+        /// </summary>
+        public IDictionary<string, object> Claims { get; } = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
```
keegan-caruso, line 42: Correct me if I am wrong, but claim names should be case sensitive, yes?

Either way, would be good to add a line to the summary or a remarks section that indicates this uses OrdinalIgnoreCase or Ordinal

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1070,13 +1070,19 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
-                    return new TokenValidationResult
+                    var tokenValidationResult = new TokenValidationResult
                     {
                         SecurityToken = jwtToken,
                         ClaimsIdentity = innerTokenValidationResult.ClaimsIdentity,
                         IsValid = true,
                         TokenType = innerTokenValidationResult.TokenType
                     };
+
+                    var claims = JwtTokenUtilities.CreateDictionaryFromClaims(innerTokenValidationResult.ClaimsIdentity.Claims);
```
keegan-caruso, line 1081: You should be getting a suggestion from the editorconfig to use the type name instead of var here

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1109,14 +1115,20 @@ private TokenValidationResult ValidateTokenPayload(JsonWebToken jsonWebToken, To
...
             {
                 SecurityToken = jsonWebToken,
-                ClaimsIdentity = CreateClaimsIdentity(jsonWebToken, validationParameters, issuer),
+                ClaimsIdentity = claimsIdentity,
                 IsValid = true,
                 TokenType = type
             };
+
+            var claims = JwtTokenUtilities.CreateDictionaryFromClaims(claimsIdentity.Claims);
+            foreach (var claim in claims)
```
brentschmaltz, line 1128: @RojaEnnam  We should delay performing this until it is asked for as this will have a performance impact on all call graphs.

RojaEnnam: @brentschmaltz are you saying we should add the property but not set it on TokenValidationResult ?

RojaEnnam:
Closing this PR and moving it to #1514

Issue #1268, opened 2019-10-14
Throw SecurityTokenSignatureKeyNotFoundException when none of the supplied keys verify the signature.


brentschmaltz:
If there are keys made available to the runtime and none of them result in a verified signature, then throw SecurityTokenSignatureKeyNotFoundException. This will allow asp.net to trigger a refresh of metadata. 

Currently we do not throw if a kid is not found in the inbound jwt.

see: https://github.com/aspnet/AspNetCore/issues/12310

PR #1506, opened 2020-08-12
Throw SecurityTokenSignatureValidationFailedException when none of the provided keys verify the signature

### LINKED ISSUE #1431: Token decryption fails if incoming token doesn't have kid (key id) in the protected header.
My application receives encrypted web tokens from AAD. I am using the OWIN middleware for setting up the token decryption parameters. I am setting the `TokenValidationParameters.TokenDecryptionKeys` with a set of two certificates either of which might be used to encrypt the token. (I am testing the certificate rotation flow). 
However, when I am testing the code, I see the following error being reported -

```
Microsoft.Owin.Security.OAuth.OAuthBearerAuthenticationMiddleware Error: 0 : Authentication failed
Microsoft.IdentityModel.Tokens.SecurityTokenKeyWrapException: IDX10659: UnwrapKey failed, exception from cryptographic operation: 'System.Security.Cryptography.CryptographicException'
   at Microsoft.IdentityModel.Tokens.RsaKeyWrapProvider.UnwrapKey(Byte[] keyBytes)
   at System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.GetContentEncryptionKeys(JwtSecurityToken jwtToken, TokenValidationParameters validationParameters)
   at System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.DecryptToken(JwtSecurityToken jwtToken, TokenValidationParameters validationParameters)
   at Microsoft.Owin.Security.Jwt.JwtFormat.Unprotect(String protectedText)
```

If I remove the certificate which is currently not being used to encrypt the tokens then the error goes away.

I believe this has to do with the code here - 
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e4928aff114bf17ee04bd0360b15da6d33694b06/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L966

It should not be enclosing the if condition at 
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e4928aff114bf17ee04bd0360b15da6d33694b06/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L981

Looking for suggestions of what is the correct way out here.

Currently to unblock myself, I am adding a token decryption key resolver with the following code - (which is heavily influenced with the code in JsonWebTokenHandler)

```c#
private List<SecurityKey> TokenDecryptionKeyResolver(string token, SecurityToken securityToken, string kid, TokenValidationParameters validationParameters)
{
	using (X509Store store = new X509Store(StoreName.My, StoreLocation.LocalMachine))
	{
			store.Open(OpenFlags.ReadOnly | OpenFlags.OpenExistingOnly);
			X509Certificate2Collection certs = store.Certificates.Find(X509FindType.FindByThumbprint, <thumbprint_new>, true);
			var tokenEncryptionCert = certs[0];			
			certs = store.Certificates.Find(X509FindType.FindByThumbprint, <thumbprint_old>, true);
			var tokenEncryptionCertOld = certs[0];
	}
	List<X509SecurityKey> tokenDecryptionKeys = new List<X509SecurityKey> {
																	new X509SecurityKey(tokenEncryptionCert),
																	new X509SecurityKey(tokenEncryptionCertOld) };
	if (!string.IsNullOrEmpty(kid))
	{
			foreach (var key in tokenDecryptionKeys)
			{
					if (key != null && string.Equals(key.KeyId, kid, StringComparison.OrdinalIgnoreCase))
					{
							return new List<SecurityKey>() { key };
					}
			}
	}
	if (securityToken is JwtSecurityToken)
	{
			JwtSecurityToken jwtToken = securityToken as JwtSecurityToken;
			if (!string.IsNullOrEmpty(jwtToken?.Header.X5t))
			{
					foreach (var key in tokenDecryptionKeys)
					{
							if (string.Equals(key.X5t, jwtToken.Header.X5t, StringComparison.OrdinalIgnoreCase))
							{
									return new List<SecurityKey>() { key };
							}
					}
			}
	}
	return null;
}
```

RojaEnnam:
Fixes #1431, #1268 

1. Add `SecurityTokenSignatureValidationFailedException`  derived from `SecurityTokenSignatureKeyNotFoundException`.
2. Throw `SecurityTokenSignatureValidationFailedException` when a security token does not contain a key identifier and the signature validation is not successful with keys found by the runtime.
3. Modify current logic to also resolve encryption keys when security token does not contain key identifier.


src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenSignatureValidationFailedException.cs
```diff
@@ -0,0 +1,76 @@
...
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// This exception is thrown when a security token does not contain a key identifier and the signature validation is not successful with keys found by the runtime.
+    /// </summary>
+    [Serializable]
+    public class SecurityTokenSignatureValidationFailedException : SecurityTokenSignatureKeyNotFoundException
```
keegan-caruso, line 37: We need to be _careful_ with this

It should work with [OIDC](https://github.com/dotnet/aspnetcore/blob/bbf7c8780c42e3d32aeec8018367037784eb9181/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs#L748)

It won't with [WsFed](https://github.com/dotnet/aspnetcore/blob/5ff9ed68d1cf6c89d72d27a69b00ed0ecd34daed/src/Security/Authentication/WsFederation/src/WsFederationHandler.cs#L298)

GeoK:
My understanding is that a fix for #1431 should be in JwtSecurityTokenHandler.

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenSignatureValidationFailedException.cs
GeoK: Good point. JwtBearer is coded in a similar way as OIDC (`ex is SecurityTokenSignatureKeyNotFoundException`) which should work fine with the proposed model. 

However, we are not breaking the refresh scenario in WsFed with this change as Wilson is currently throwing `SecurityTokenInvalidSignatureException` when there is no 'kid', and keys couldn't validate the signature.

The right thing to do here would be to create a pull request for WsFed, so it's consistent with other handlers.
We would need similar PRs for Katana:
[oidc](https://github.com/aspnet/AspNetKatana/blob/635c92f641ad1e014eead31cc7a365004949fda5/src/Microsoft.Owin.Security.OpenIdConnect/OpenidConnectAuthenticationHandler.cs#L526)
[wsfed](https://github.com/aspnet/AspNetKatana/blob/e2b18ec84ceab7ffa29d80d89429c9988ab40144/src/Microsoft.Owin.Security.WsFederation/WsFederationAuthenticationHandler.cs#L372)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -977,30 +977,30 @@ protected virtual SecurityKey ResolveTokenDecryptionKey(string token, JsonWebTok
                             return key;
                     }
                 }
+            }
 
-                if (!string.IsNullOrEmpty(jwtToken.X5t))
```
GeoK, line 981: It would be great to have tests that verify the new behavior.

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenSignatureValidationFailedException.cs
RojaEnnam: That is a Good point, for WsFed it wouldn't break the refresh scenario but #1268 will not be addressed. 

I agree about creating pull request for other handlers.

That makes me think of another scenario where the user is also using similar method as WsFed ( `exception.GetType().Equals(typeof(SecurityTokenSignatureKeyNotFoundException` ) to check what exception it is and is performing some action based on it, that would break things ? as we are currently throwing a derived exception( `SecurityTokenSignatureValidationFailedException` ) in place of `SecurityTokenInvalidSignatureException` .

brentschmaltz:
@RojaEnnam 1431 seems to be caused because in JwtSecurityTokenHandler.GetContentEncryptionKeys we will not try all keys. We could have a try / catch similar to what we do when we try all signing keys.

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenSignatureValidationFailedException.cs
brentschmaltz: @keegan-caruso nice catch. I opened https://github.com/dotnet/aspnetcore/issues/24847 against asp.net.

@RojaEnnam  We can not make this change to a new exception. We will have stay with the existing exception.

RojaEnnam:
Closing this PR and moving it to #1515

PR #1508, opened 2020-08-12
Issue: https://github.com/AzureAD/azure-activedirectory-identitymodel…

### LINKED ISSUE #1200: EnvelopedSignatureWriter should be able to determine the Digest algorithm from signing algorithm
If SigningCredentials.Digest is null a fault is generated.
We should ask the CryptoProviderFactory to create the HashAlgorithm with similar logic as [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e69d0c77909334ddba19ac15a6d993adcc77e458/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs#L187)

brentschmaltz:
This address issue https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1200
Where we want to determine the digest value from signature algorithm if the SigningCredentials.Digest is null.


test/Microsoft.IdentityModel.Xml.Tests/EnvelopedSignatureWriterTests.cs
```diff
@@ -150,13 +237,11 @@ public void RoundTripWsMetadata(EnvelopedSignatureTheoryData theoryData)
...
                 };
+
                 var buffer = new MemoryStream();
                 var esw = new EnvelopedSignatureWriter(XmlWriter.Create(buffer, settings), theoryData.SigningCredentials, theoryData.ReferenceId);
 
-                foreach (var action in theoryData.Action.GetInvocationList())
-                {
-                    action.DynamicInvoke(esw);
-                }
+                theoryData.Action.DynamicInvoke(esw);
```
keegan-caruso, line 244: nit: DynamicInvoke is going to be much slower than using Invoke or directly calling the delegate.

keegan-caruso (approved):

RojaEnnam (approved):

test/Microsoft.IdentityModel.Xml.Tests/EnvelopedSignatureWriterTests.cs
brentschmaltz: @keegan-caruso thanks, i am on the fence about actions in general. it makes it easy to set a breakpoint on the specific test, but adds a layer.

Issue #1501, opened 2020-08-03
Investigate issue with DTD processing suppression in XML in WsFederation


brentschmaltz:
We need to ensure this is not an issue.

aasembly: SuppressMessage("Security", "CA3075:Insecure DTD processing in XML", Justification = "Not created from factory, dtd is still prohibited", Scope = "member", Target = "~M:Microsoft.IdentityModel.Protocols.WsFederation.WsFederationMessage.GetTokenUsingXmlReader~System.String")]

PR #1509, opened 2020-08-13
Code cleanup to remove warning

brentschmaltz:
The code removed triggered a warning, also set XmlResolver to null as a precautionary item.
This cleans up https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1501

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
+++ b/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
@@ -251,11 +251,8 @@ namespace Microsoft.IdentityModel.Protocols.WsFederation
 
             string token = null;
             using (var sr = new StringReader(Wresult))
-            using (var xmlReader = new XmlTextReader(sr) { DtdProcessing = DtdProcessing.Prohibit })
+            using (var xmlReader = new XmlTextReader(sr) { DtdProcessing = DtdProcessing.Prohibit, XmlResolver = null })
             {
-                if (xmlReader.Settings != null)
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
```diff
@@ -251,11 +251,8 @@ public virtual string GetTokenUsingXmlReader()
 
             string token = null;
             using (var sr = new StringReader(Wresult))
-            using (var xmlReader = new XmlTextReader(sr) { DtdProcessing = DtdProcessing.Prohibit })
+            using (var xmlReader = new XmlTextReader(sr) { DtdProcessing = DtdProcessing.Prohibit, XmlResolver = null })
             {
-                if (xmlReader.Settings != null)
```
keegan-caruso, line 256: what is the history on this?

keegan-caruso (approved):

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
brentschmaltz: @keegan-caruso  That's part of what was causing the warning.
Some of our older code for netstandard 1.4 would not allow XmlreaderSettings to be set, so we had to check for null.
However, XmlTextReader allows setting DtdProcessing.Prohibit the code could be removed.

RojaEnnam (approved):

brentschmaltz:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1509

PR #1511, opened 2020-08-17
Resolve encryption key when Kid is not present in token

### LINKED ISSUE #1431: Token decryption fails if incoming token doesn't have kid (key id) in the protected header.

RojaEnnam:
Fixes #1431 in both `JsonWebTokenHandler `and `JwtSecurityTokenHandler`.

- Modify current logic to also resolve encryption keys when security token does not contain key identifier.
- Add try catch in `GetContentEncryptionKeys` to catch exception and continue with other keys
- Add unit tests to test new behavior.


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -965,7 +979,7 @@ protected virtual SecurityKey ResolveTokenDecryptionKey(string token, JsonWebTok
 
             if (!string.IsNullOrEmpty(jwtToken.Kid))
             {
-                if (validationParameters.TokenDecryptionKey != null
+                 if (validationParameters.TokenDecryptionKey != null
```
keegan-caruso: nit: bonus whitespace?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -303,6 +303,97 @@ public static TheoryData<CreateTokenTheoryData> CreateJWETheoryData
...
+
+                IdentityComparer.AreEqual(encryptionKeysFromJsonHandlerWithKid, encryptionKeysFromJsonHandlerWithNoKid, context);
+                theoryData.ExpectedException.ProcessNoException(context);
+            }
+            catch (Exception ex)
+            {
+                theoryData.ExpectedException.ProcessException(ex.InnerException, context);
+            }
+            TestUtilities.AssertFailIfErrors(context);
+        }
```
keegan-caruso, line 334: nit: usually have a new line between a brace and a method

src/Microsoft.IdentityModel.JsonWebTokens/GlobalSuppressions.cs
```diff
@@ -20,3 +20,4 @@
...
 [assembly: SuppressMessage("Performance", "CA1815:Override equals and operator equals on value types", Justification = "Has no fields or properties", Scope = "type", Target = "~T:Microsoft.IdentityModel.JsonWebTokens.JwtHeaderParameterNames")]
 [assembly: SuppressMessage("Performance", "CA1815:Override equals and operator equals on value types", Justification = "Has no fields or properties", Scope = "type", Target = "~T:Microsoft.IdentityModel.JsonWebTokens.JwtRegisteredClaimNames")]
+[assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "Exception is wrote to string", Scope = "member", Target = "~M:Microsoft.IdentityModel.JsonWebTokens.JsonWebTokenHandler.GetContentEncryptionKeys(Microsoft.IdentityModel.JsonWebTokens.JsonWebToken,Microsoft.IdentityModel.Tokens.TokenValidationParameters)~System.Collections.Generic.IEnumerable{Microsoft.IdentityModel.Tokens.SecurityKey}")]
```
mafurman: nit: I believe the proper grammar here is "exception is written to a string"

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: Yes, removed it.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
```diff
@@ -2360,6 +2361,99 @@ public static TheoryData<ParametersCheckTheoryData> ParametersCheckTheoryData()
...
             };
         }
+
+        [Theory, MemberData(nameof(SecurityTokenDecryptionTheoryData))]
+        public void GetEncryptionKeys(CreateTokenTheoryData theoryData)
+        {
+            var context = TestUtilities.WriteHeader($"{this}.EncryptionKeysCheck", theoryData);
+            try
+            {
+                TestUtilities.AssertFailIfErrors(context);
```
GeoK: Why do we need this line here?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -303,6 +303,98 @@ public static TheoryData<CreateTokenTheoryData> CreateJWETheoryData
...
 
+        [Theory, MemberData(nameof(SecurityTokenDecryptionTheoryData))]
+        public void GetEncryptionKeys(CreateTokenTheoryData theoryData)
+        {
+            var context = TestUtilities.WriteHeader($"{this}.EncryptionKeysCheck", theoryData);
+            try
+            {
+                string jweFromJsonHandlerWithKid = theoryData.JsonWebTokenHandler.CreateToken(theoryData.Payload, theoryData.TokenDescriptor.SigningCredentials, theoryData.TokenDescriptor.EncryptingCredentials);
+                var jwtTokenFromJsonHandlerWithKid = new JsonWebToken(jweFromJsonHandlerWithKid);
+                var getContentEncryptionKeysMethod = theoryData.JsonWebTokenHandler.GetType().GetMethod("GetContentEncryptionKeys", BindingFlags.Instance | BindingFlags.NonPublic);
```
GeoK: Consider making GetContentEncryptionKeys internal to eliminate reflection from tests, and refactor the code to test if decryption keys for a given token and set of parameters matches an expected set of keys.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -934,17 +934,31 @@ private IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken,
...
+                        var kwp = key.CryptoProviderFactory.CreateKeyWrapProviderForUnwrap(key, jwtToken.Alg);
+                        var unwrappedKey = kwp.UnwrapKey(Base64UrlEncoder.DecodeBytes(jwtToken.EncryptedKey));
+                        unwrappedKeys.Add(new SymmetricSecurityKey(unwrappedKey));
+                    }
                 }
+                catch (Exception ex)
+                {
+                    exceptionStrings.AppendLine(ex.ToString());
+                }
+                keysAttempted.AppendLine(key.ToString() + " , KeyId: " + key.KeyId);
```
GeoK: Not sure if we need the keyId here, as SecurityKey.ToString() prints it too:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/19092cb44fd556084c4cef9dfef08ec38e2ee130/src/Microsoft.IdentityModel.Tokens/SecurityKey.cs#L92-L95

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -119,7 +119,7 @@ internal static class LogMessages
         public const string IDX10610 = "IDX10610: Decryption failed. Could not create decryption provider. Key: '{0}', Algorithm: '{1}'.";
         public const string IDX10611 = "IDX10611: Decryption failed. Encryption is not supported for: Algorithm: '{0}', SecurityKey: '{1}'.";
         public const string IDX10612 = "IDX10612: Decryption failed. Header.Enc is null or empty, it must be specified.";
-        // public const string IDX10613 = "IDX10613:";
+        public const string IDX10613 = "IDX10613: Key unwrap failed using decryption Keys: '{0}'.\nExceptions caught:\n '{1}'.\ntoken: '{2}'.";
```
GeoK: You should use a new log id for this message. Log Id is considered deprecated if it's commented out. We are using this approach to make sure not to re-use a log id as it might lead to some confusion.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
RojaEnnam: We don't need it. I think I forgot to delete it. I removed it now.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: You're right, we don't need KeyId. I made the change.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
RojaEnnam: That answers why they were commented out :) Thank you. I added a new message ID.

GeoK (approved):

PR #1514, opened 2020-08-18
Add Claims and PropertyBag properties to TokenValidationResult

### LINKED ISSUE #1194: Add property bag to TokenValidationResult

RojaEnnam:
Fixes #1194

- Adds `Claims `property, that is equivalent to `SecurityTokenDescriptor.Claims` on `TokenValidationResult`.
- Adds `PropertBag`


src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -65,50 +63,6 @@ public class JwtTokenUtilities
             JwtHeaderParameterNames.Zip
         };
 
-        /// <summary>
-        /// Creates a dictionary from a list of Claim's.
-        /// </summary>
-        /// <param name="claims"> A list of claims.</param>
-        /// <returns> A Dictionary representing claims.</returns>
-        internal static IDictionary<string, object> CreateDictionaryFromClaims(IEnumerable<Claim> claims)
```
GeoK, line 73: What is the reason for moving this method to M.IM.Tokens?

RojaEnnam: `TokenValidationResult  ` is in M.IM.Tokens, and to use JwtTokenUtilities it has to refer to M.IM.JsonWebTokens . Which causes circular dependency.

src/Microsoft.IdentityModel.Tokens/JsonClaimValueTypes.cs
```diff
@@ -0,0 +1,53 @@
...
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Constants for Json Web tokens.
+    /// </summary>
+    public static class JsonClaimValueTypes
```
brentschmaltz: This type is in M.IM.JsonWebTokens, we shouldn't have two public types with the same name.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -39,6 +41,79 @@ namespace Microsoft.IdentityModel.Tokens
...
+                    payload[jsonClaimType] = jsonClaimValue;
+                }
+            }
+
+            return payload;
+        }
+
+        internal static object GetClaimValueUsingValueType(Claim claim)
+        {
+            if (claim.ValueType == ClaimValueTypes.Boolean && bool.TryParse(claim.Value, out bool boolValue))
```
brentschmaltz, line 112: most likely, the ValueType will be ClaimValueTypes.String, perhaps we should check that first.

src/Microsoft.IdentityModel.Tokens/JsonClaimValueTypes.cs
RojaEnnam: JsonClaimValueTypes is used by GetClaimValueUsingValueType , due to circular dependency problem I had to create a new one as System.IdentityModel.Tokens.Jwt also has one :)

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
RojaEnnam: I think that's a good idea if its most likely string. I made the change.

src/Microsoft.IdentityModel.Tokens/JsonClaimValueTypes.cs
keegan-caruso: I recognize the problem, but I agree with Brent here. I'm not sure this is how we want to solve this.

RojaEnnam: Defining `JsonClaimValueTypes `consts locally in `TokenUtilities `to solve the problem at this moment.

test/Microsoft.IdentityModel.Tokens.Tests/TokenValidationResultTests.cs
```diff
@@ -45,8 +45,8 @@ public void GetSets()
             TokenValidationResult tokenValidationResult = new TokenValidationResult();
             Type type = typeof(TokenValidationResult);
             PropertyInfo[] properties = type.GetProperties();
-            if (properties.Length != 7)
-                Assert.True(false, "Number of public fields has changed from 7 to: " + properties.Length + ", adjust tests");
+            if (properties.Length != 9)
```
keegan-caruso, line 48: +2 properties but +1 test case. Should Claims not be included here?

RojaEnnam: I believe every time a new object is created when accessing Claims, [here ](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/92580d7f1a3027c6f400cc11be46d9e061dedea2/test/Microsoft.IdentityModel.TestUtils/TestUtilities.cs#L161) they don't match on all properties as they are different objects.

GeoK (approved):

keegan-caruso (approved):

PR #1515, opened 2020-08-19
Throw SecurityTokenSignatureValidationFailedException when none of the provided keys verify the signature

### LINKED ISSUE #1268: Throw SecurityTokenSignatureKeyNotFoundException when none of the supplied keys verify the signature.

RojaEnnam:
Fixes  #1268

- Throw `SecurityTokenSignatureKeyNotFoundException` when a security token does not contain a key identifier and the signature validation is not successful with keys found by the runtime.
- Modify current logic to also resolve encryption keys when security token does not contain key identifier.


RojaEnnam:
The duplicated code is not all from this PR. I looked into reformatting ValidateSignature method in Saml/Saml2SecurityTokenHandlers. One thing I found was Saml/Saml2Assertion does not have a base class.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -96,7 +96,7 @@ internal static class LogMessages
         // 10500 - SignatureValidation
         public const string IDX10500 = "IDX10500: Signature validation failed. No security keys were provided to validate the signature.";
         public const string IDX10501 = "IDX10501: Signature validation failed. Unable to match key: \nkid: '{0}'.\nExceptions caught:\n '{1}'. \ntoken: '{2}'.";
-        public const string IDX10503 = "IDX10503: Signature validation failed. Keys tried: '{0}'.\nExceptions caught:\n '{1}'.\ntoken: '{2}'.";
+        public const string IDX10503 = "IDX10503: Signature validation failed. Token does not have a kid. Keys tried: '{0}'.\nExceptions caught:\n '{1}'.\ntoken: '{2}'.";
```
keegan-caruso, line 99: This doesn't work for SAML/SAML2

RojaEnnam: You're right. I missed that. I made the change.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -1104,7 +1104,7 @@ private SamlSecurityToken ValidateSignature(SamlSecurityToken samlToken, string
                 throw LogExceptionMessage(new SecurityTokenSignatureKeyNotFoundException(FormatInvariant(TokenLogMessages.IDX10501, samlToken.Assertion.Signature.KeyInfo, exceptionStrings, samlToken)));
```
brentschmaltz: We could change message IDX10501 to be specific for SAML and use 'KeyInfo' as you did with IDX10512

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -105,6 +105,7 @@ internal static class LogMessages
         public const string IDX10509 = "IDX10509: Signature validation failed. The user defined 'Delegate' specified in TokenValidationParameters did not return a '{0}', but returned a '{1}' when reading token: '{2}'.";
         public const string IDX10510 = "IDX10510: Signature validation failed. The user defined 'Delegate' specified in TokenValidationParameters returned null when reading token: '{0}'.";
         public const string IDX10511 = "IDX10511: Signature validation failed. Keys tried: '{0}'. \nkid: '{1}'. \nExceptions caught:\n '{2}'.\ntoken: '{3}'.";
```
brentschmaltz, line 107: Change 'kid' to 'KeyInfo'

brentschmaltz:
@RojaEnnam looks good, only change i have to suggest is to use 'KeyInfo' instead of 'kid' for saml tokens.
The reason that there is no base class for Saml1 and Saml2 is that they are very different objects.

We could have ValidateSignature across all tokens perhaps in the future.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
RojaEnnam: IDX10511 is not used for saml. For IDX10501 I added IDX10513 to be saml specific.

brentschmaltz (approved):

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
keegan-caruso: Looks like 10503 should have kid -> KeyInfo then?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -444,10 +444,10 @@ private Saml2SecurityToken ValidateSignature(Saml2SecurityToken samlToken, strin
 
             //if there was a key match with what was found in tokenValidationParameters most likely metadata is stale. throw SecurityTokenSignatureKeyNotFoundException
             if (!keyMatched && canMatchKey && keysAttempted.Length > 0)
-                throw LogExceptionMessage(new SecurityTokenSignatureKeyNotFoundException(FormatInvariant(TokenLogMessages.IDX10501, samlToken.Assertion.Signature.KeyInfo, exceptionStrings, samlToken)));
+                throw LogExceptionMessage(new SecurityTokenSignatureKeyNotFoundException(FormatInvariant(TokenLogMessages.IDX10513, samlToken.Assertion.Signature.KeyInfo, exceptionStrings, samlToken)));
 
             if (keysAttempted.Length > 0)
-                throw LogExceptionMessage(new SecurityTokenInvalidSignatureException(FormatInvariant(TokenLogMessages.IDX10503, keysAttempted, exceptionStrings, samlToken)));
+                throw LogExceptionMessage(new SecurityTokenSignatureKeyNotFoundException(FormatInvariant(TokenLogMessages.IDX10512, keysAttempted, exceptionStrings, samlToken)));
```
keegan-caruso, line 454: Is this logically correct still?

If we look at the test cases, we are raising KeyNotFound for cases where attribute or signatures have been tampered with

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
RojaEnnam: 10503  is used in M.IM.JsonWebTokens and S.IM.T.jwt, saml specific for 10503 is 10512.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
RojaEnnam: I was wondering about that myself. Upon more testing I think its best to match saml logic to jwt. 
I went ahead made the changes in logic and tests.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -1099,14 +1099,20 @@ private SamlSecurityToken ValidateSignature(SamlSecurityToken samlToken, string
...
 
-            // if there was a keymatch with what was found in tokenValidationParameters most likely metadata is stale. throw SecurityTokenSignatureKeyNotFoundException
-            if (!keyMatched && canMatchKey && keysAttempted.Length > 0)
-                throw LogExceptionMessage(new SecurityTokenSignatureKeyNotFoundException(FormatInvariant(TokenLogMessages.IDX10501, samlToken.Assertion.Signature.KeyInfo, exceptionStrings, samlToken)));
-
-            if (keysAttempted.Length > 0)
-                throw LogExceptionMessage(new SecurityTokenInvalidSignatureException(FormatInvariant(TokenLogMessages.IDX10503, keysAttempted, exceptionStrings, samlToken)));
-
-            throw LogExceptionMessage(new SecurityTokenInvalidSignatureException(TokenLogMessages.IDX10500));
+            if (canMatchKey)
```
keegan-caruso, line 1102: This could be simplified (a lot of the `else`s are redundant)

keegan-caruso (approved):

PR #1518, opened 2020-08-19
fixes typo in WsTrustSerializer

gislikonrad:
This is a PR for brentsch/wstrust. There was a typo in the ReadRequestSecurityTokenResponse method name.


gislikonrad:
@brentschmaltz Thought you might want to fix this.

brentschmaltz:
This is a bit too old now, if we still need it we should push to WSTrust6.8.0

PR #1522, opened 2020-08-26
Add TokenType to SecurityTokenDescriptor.

### LINKED ISSUE #1397: Consider adding a TokenType property to SecurityTokenDescriptor
When creating an encrypted JWT using the `CreateToken(SecurityTokenDescriptor tokenDescriptor)` API, the headers added to `SecurityTokenDescriptor.AdditionalHeaderClaims` are only used for the JWE container and not the JWS. If you want to assign the JWS a specific `typ`, you must first create an unencrypted token with `AdditionalHeaderClaims` populated and then encrypt it manually, potentially with `AdditionalHeaderClaims` populated too if you want both the JWE and the JWS to have a `typ` header.

This procedure is a bit counter-intuitive and frankly error-prone. Things would be much simpler if `SecurityTokenDescriptor` had a `TokenType` property that would be used when creating both the JWS and the JWE.

/cc @brentschmaltz @mafurman

RojaEnnam:
Fixes #1397 in `JsonWebTokenHandler `and `JwtSecurityTokenHandler`.

- Adds `TokenType `property to `SecurityTokenDescriptor`.
- Sets 'typ' claim to 'TokenType' property value in JWS and JWE headers.
- Adds unit tests to verify the 'typ' claim value is correct.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -748,7 +748,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             var cryptoProviderFactory = encryptingCredentials.CryptoProviderFactory ?? encryptingCredentials.Key.CryptoProviderFactory;
 
             if (cryptoProviderFactory == null)
-                throw LogHelper.LogExceptionMessage(new ArgumentException(TokenLogMessages.IDX10619));
+                throw LogHelper.LogExceptionMessage(new ArgumentException(TokenLogMessages.IDX10620));
 
             byte[] wrappedKey = null;
             SecurityKey securityKey = JwtTokenUtilities.GetSecurityKey(encryptingCredentials,cryptoProviderFactory, out wrappedKey);
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -109,7 +109,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             get { return true; }
         }
 
-        private static JObject CreateDefaultJWEHeader(EncryptingCredentials encryptingCredentials, string compressionAlgorithm)
+        private static JObject CreateDefaultJWEHeader(EncryptingCredentials encryptingCredentials, string compressionAlgorithm, string tokenType)
         {
             var header = new JObject();
             header.Add(JwtHeaderParameterNames.Alg, encryptingCredentials.Alg);
@@ -121,12 +121,15 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (!string.IsNullOrEmpty(compressionAlgorithm))
                 header.Add(JwtHeaderParameterNames.Zip, compressionAlgorithm);
 
-            header.Add(JwtHeaderParameterNames.Typ, JwtConstants.HeaderType);
+            if (string.IsNullOrEmpty(tokenType))
+                header.Add(JwtHeaderParameterNames.Typ, JwtConstants.HeaderType);
+            else
+                header.Add(JwtHeaderParameterNames.Typ, tokenType);
 
             return header;
         }
 
-        private static JObject CreateDefaultJWSHeader(SigningCredentials signingCredentials)
+        private static JObject CreateDefaultJWSHeader(SigningCredentials signingCredentials, string tokenType)
         {
             var header = new JObject()
             {
@@ -136,7 +139,10 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (signingCredentials.Key.KeyId != null)
                 header.Add(JwtHeaderParameterNames.Kid, signingCredentials.Key.KeyId);
 
-            header.Add(JwtHeaderParameterNames.Typ, JwtConstants.HeaderType);
+            if (string.IsNullOrEmpty(tokenType))
+                header.Add(JwtHeaderParameterNames.Typ, JwtConstants.HeaderType);
+            else
+                header.Add(JwtHeaderParameterNames.Typ, tokenType);
 
             if (signingCredentials.Key is X509SecurityKey x509SecurityKey)
                 header[JwtHeaderParameterNames.X5t] = x509SecurityKey.X5t;
@@ -155,7 +161,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (string.IsNullOrEmpty(payload))
                 throw LogHelper.LogArgumentNullException(nameof(payload));
 
-            return CreateTokenPrivate(JObject.Parse(payload), null, null, null, null);
+            return CreateTokenPrivate(JObject.Parse(payload), null, null, null, null, null);
         }
 
         /// <summary>
@@ -174,7 +180,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (additionalHeaderClaims == null)
                 throw LogHelper.LogArgumentNullException(nameof(additionalHeaderClaims));
 
-            return CreateTokenPrivate(JObject.Parse(payload), null, null, null, additionalHeaderClaims);
+            return CreateTokenPrivate(JObject.Parse(payload), null, null, null, additionalHeaderClaims, null);
         }
 
         /// <summary>
@@ -193,7 +199,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (signingCredentials == null)
                 throw LogHelper.LogArgumentNullException(nameof(signingCredentials));
 
-            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, null, null, null);
+            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, null, null, null, null);
         }
 
         /// <summary>
@@ -220,7 +226,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (additionalHeaderClaims == null)
                 throw LogHelper.LogArgumentNullException(nameof(additionalHeaderClaims));
 
-            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, null, null, additionalHeaderClaims);
+            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, null, null, additionalHeaderClaims, null);
         }
 
         /// <summary>
@@ -288,7 +294,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 payload[JwtRegisteredClaimNames.Nbf] = EpochTime.GetIntDate(tokenDescriptor.NotBefore.Value);
             }
 
-            return CreateTokenPrivate(payload, tokenDescriptor.SigningCredentials, tokenDescriptor.EncryptingCredentials, tokenDescriptor.CompressionAlgorithm, tokenDescriptor.AdditionalHeaderClaims);
+            return CreateTokenPrivate(payload, tokenDescriptor.SigningCredentials, tokenDescriptor.EncryptingCredentials, tokenDescriptor.CompressionAlgorithm, tokenDescriptor.AdditionalHeaderClaims, tokenDescriptor.TokenType);
         }
 
         /// <summary>
@@ -305,7 +311,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (encryptingCredentials == null)
                 throw LogHelper.LogArgumentNullException(nameof(encryptingCredentials));
 
-            return CreateTokenPrivate(JObject.Parse(payload), null, encryptingCredentials, null, null);
+            return CreateTokenPrivate(JObject.Parse(payload), null, encryptingCredentials, null, null, null);
         }
 
         /// <summary>
@@ -329,7 +335,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (encryptingCredentials == null)
                 throw LogHelper.LogArgumentNullException(nameof(encryptingCredentials));
 
-            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, encryptingCredentials, null, null);
+            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, encryptingCredentials, null, null, null);
         }
 
         /// <summary>
@@ -361,7 +367,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (additionalHeaderClaims == null)
                 throw LogHelper.LogArgumentNullException(nameof(additionalHeaderClaims));
 
-            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, encryptingCredentials, null, additionalHeaderClaims);
+            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, encryptingCredentials, null, additionalHeaderClaims, null);
         }
 
         /// <summary>
@@ -382,7 +388,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (string.IsNullOrEmpty(compressionAlgorithm))
                 throw LogHelper.LogArgumentNullException(nameof(compressionAlgorithm));
 
-            return CreateTokenPrivate(JObject.Parse(payload), null, encryptingCredentials, compressionAlgorithm, null);
+            return CreateTokenPrivate(JObject.Parse(payload), null, encryptingCredentials, compressionAlgorithm, null, null);
         }
 
         /// <summary>
@@ -411,7 +417,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (string.IsNullOrEmpty(compressionAlgorithm))
                 throw LogHelper.LogArgumentNullException(nameof(compressionAlgorithm));
 
-            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, encryptingCredentials, compressionAlgorithm, null);
+            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, encryptingCredentials, compressionAlgorithm, null, null);
         }
 
         /// <summary>
@@ -448,37 +454,27 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (additionalHeaderClaims == null)
                 throw LogHelper.LogArgumentNullException(nameof(additionalHeaderClaims));
 
-            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, encryptingCredentials, compressionAlgorithm, additionalHeaderClaims);
+            return CreateTokenPrivate(JObject.Parse(payload), signingCredentials, encryptingCredentials, compressionAlgorithm, additionalHeaderClaims, null);
         }
 
-        private string CreateTokenPrivate(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, string compressionAlgorithm, IDictionary<string, object> additionalHeaderClaims)
+        private string CreateTokenPrivate(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, string compressionAlgorithm, IDictionary<string, object> additionalHeaderClaims, string tokenType)
         {
-            if (additionalHeaderClaims != null && additionalHeaderClaims.Count > 0 && additionalHeaderClaims.Keys.Intersect(JwtTokenUtilities.DefaultHeaderParameters, StringComparer.OrdinalIgnoreCase).Any())
+            if (additionalHeaderClaims?.Count > 0 && additionalHeaderClaims.Keys.Intersect(JwtTokenUtilities.DefaultHeaderParameters, StringComparer.OrdinalIgnoreCase).Any())
                 throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX14116, nameof(additionalHeaderClaims), string.Join(", ", JwtTokenUtilities.DefaultHeaderParameters))));
 
-            var rawHeader = Base64UrlEncodedUnsignedJWSHeader;
-            // If there's no additional header claims to be added to the header and the token will be signed, try to retrieve a header value from the cache.
-            if ((additionalHeaderClaims == null || additionalHeaderClaims.Count == 0) && signingCredentials != null)
-            {
-                var header = CreateDefaultJWSHeader(signingCredentials);
-                rawHeader = Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(header.ToString(Formatting.None)));
-            } // Otherwise, if there is no outer JWT header, add additional header claims to this header.
-            else if (additionalHeaderClaims != null && additionalHeaderClaims.Count != 0)
-            {
-                JObject header;
-                if (signingCredentials != null)
-                    header = CreateDefaultJWSHeader(signingCredentials);
-                else
-                    header = new JObject
+            JObject header = null;
+            if (signingCredentials != null)
+                header = CreateDefaultJWSHeader(signingCredentials, tokenType);
+            else
+                header = new JObject
                     {
                         { JwtHeaderParameterNames.Alg, SecurityAlgorithms.None},
                     };
 
-                if (encryptingCredentials == null)
-                    header.Merge(JObject.FromObject(additionalHeaderClaims));
+            if (encryptingCredentials == null && additionalHeaderClaims != null && additionalHeaderClaims.Count > 0)
+                header.Merge(JObject.FromObject(additionalHeaderClaims));
 
-                rawHeader = Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(header.ToString(Formatting.None)));
-            }
+            var rawHeader = Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(header.ToString(Formatting.None)));
 
             if (SetDefaultTimesOnTokenCreation)
             {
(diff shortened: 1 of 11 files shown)


src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
```diff
@@ -102,6 +102,43 @@ public JwtHeader(SigningCredentials signingCredentials, IDictionary<string,strin
...
         }
 
+        /// <summary>
+        /// Initializes a new instance of <see cref="JwtHeader"/>.
+        /// With the Header Parameters:
+        /// <para>{ { typ, JWT }, { alg, SigningCredentials.Algorithm } }</para>
+        /// </summary>
+        /// <param name="signingCredentials"><see cref="SigningCredentials"/> used when creating a JWS Compact JSON.</param>
+        /// <param name="outboundAlgorithmMap">provides a mapping for the 'alg' value so that values are within the JWT namespace.</param>
+        /// /// <param name="tokenType"> provides the token type</param>
```
keegan-caruso: nit: extra `///`

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
```diff
@@ -102,6 +102,43 @@ public JwtHeader(SigningCredentials signingCredentials, IDictionary<string,strin
...
+        public JwtHeader(SigningCredentials signingCredentials, IDictionary<string, string> outboundAlgorithmMap, string tokenType)
+            : base(StringComparer.Ordinal)
+        {
+            if (signingCredentials == null)
+            {
```
keegan-caruso: nit: prefer no braces for single line statements

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -448,37 +454,27 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
-            } // Otherwise, if there is no outer JWT header, add additional header claims to this header.
-            else if (additionalHeaderClaims != null && additionalHeaderClaims.Count != 0)
-            {
-                JObject header;
-                if (signingCredentials != null)
-                    header = CreateDefaultJWSHeader(signingCredentials);
-                else
-                    header = new JObject
+            JObject header = null;
+            if (signingCredentials != null)
```
mafurman: I really like this change. I think it makes the code a lot easier to understand.

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
```diff
@@ -102,6 +102,43 @@ public JwtHeader(SigningCredentials signingCredentials, IDictionary<string,strin
...
+                    Alg = outboundAlg;
+                else
+                    Alg = signingCredentials.Algorithm;
+
+                if (!string.IsNullOrEmpty(signingCredentials.Key.KeyId))
+                    Kid = signingCredentials.Key.KeyId;
+
+                if (signingCredentials is X509SigningCredentials x509SigningCredentials)
+                    this[JwtHeaderParameterNames.X5t] = Base64UrlEncoder.Encode(x509SigningCredentials.Certificate.GetCertHash());
+            }
```
keegan-caruso, line 111: everything here and above looks the same as the ctor directly above.

Can you call that one, and then only keep the typ setting logic here?

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
```diff
@@ -135,6 +172,43 @@ public JwtHeader(EncryptingCredentials encryptingCredentials, IDictionary<string
...
+
+            if (outboundAlgorithmMap != null && outboundAlgorithmMap.TryGetValue(encryptingCredentials.Enc, out outboundAlg))
+                Enc = outboundAlg;
+            else
+                Enc = encryptingCredentials.Enc;
+
+            if (!string.IsNullOrEmpty(encryptingCredentials.Key.KeyId))
+                Kid = encryptingCredentials.Key.KeyId;
+
+            if (string.IsNullOrEmpty(tokenType))
```
keegan-caruso, line 163: similar comment as above. Only thing that looks different is the logic to set Typ

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -558,12 +560,12 @@ private JwtSecurityToken CreateJwtSecurityTokenPrivate(string issuer, string aud
             LogHelper.LogInformation(LogMessages.IDX12722, rawHeader, rawPayload, rawSignature);
 
             if (encryptingCredentials != null)
-                return EncryptToken(new JwtSecurityToken(header, payload, rawHeader, rawPayload, rawSignature), encryptingCredentials);
-            else
-                return new JwtSecurityToken(header, payload, rawHeader, rawPayload, rawSignature);
+                return EncryptToken(new JwtSecurityToken(header, payload, rawHeader, rawPayload, rawSignature), encryptingCredentials, tokenType);
+
+            return new JwtSecurityToken(header, payload, rawHeader, rawPayload, rawSignature);
```
keegan-caruso, line 565: Why do we include the token type for encrypted but not for un-encrypted?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -795,118 +791,85 @@ public string EncryptToken(string innerJwt, EncryptingCredentials encryptingCred
...
-                        throw LogHelper.LogExceptionMessage(new SecurityTokenEncryptionFailedException(LogMessages.IDX14103));
+                kwProvider = cryptoProviderFactory.CreateKeyWrapProvider(encryptingCredentials.Key, encryptingCredentials.Alg);
+                var symmetricKey = securityKey as SymmetricSecurityKey;
+                wrappedKey = kwProvider.WrapKey(symmetricKey.Key);
+            }
 
-                    var header = CreateDefaultJWEHeader(encryptingCredentials, compressionAlgorithm);
-                    if (additionalHeaderClaims != null)
-                        header.Merge(JObject.FromObject(additionalHeaderClaims));
+            using (var encryptionProvider = cryptoProviderFactory.CreateAuthenticatedEncryptionProvider(securityKey, encryptingCredentials.Enc))
```
mafurman, line 756: I really like the refactoring in this method as well. :)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -448,37 +454,27 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
-                JObject header;
-                if (signingCredentials != null)
-                    header = CreateDefaultJWSHeader(signingCredentials);
-                else
-                    header = new JObject
+            JObject header = null;
+            if (signingCredentials != null)
+                header = CreateDefaultJWSHeader(signingCredentials, tokenType);
+            else
+                header = new JObject
```
mafurman: should 'typ' be added to the header here as well (if it was provided)?

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
```diff
@@ -71,6 +71,12 @@ public class SecurityTokenDescriptor
         /// </summary>
         public DateTime? NotBefore { get; set; }
 
+        /// <summary>
+        /// Gets or sets the token type
```
mafurman: nit: maybe put a '.' at the end of the comment

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
RojaEnnam: For un-encrypted the token type is included while creating `Jwtheader`.
Here it is included again so that the 'typ' claim in the outer header also have the token type value. Does that makes sense?

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
RojaEnnam: That sounds reasonable :)

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
keegan-caruso: makes sense, thanks!

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
```diff
@@ -71,6 +71,12 @@ public class SecurityTokenDescriptor
...
+        /// <remarks> This value will be replaced in the JWS header(in case of no encryption) and in outer header of JWE by 'typ' claim from <see cref="AdditionalHeaderClaims"/>, if it is provided.</remarks>
```
mafurman: Is the 'typ' value added to both headers if the token happens to be encrypted? If so, it might be worth adding that in since I think it's a little unclear.

"If provided, this will be added as the value for the 'typ' header parameter. In the case of a JWE, this will be added to both the inner (JWS) and the outer token (JWE) header. By default, the value used is 'JWT'.

"If AdditionalHeaderClaims also contains 'typ' header claim value, it will override the TokenType provided here."

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -878,7 +870,7 @@ public override string WriteToken(SecurityToken token)
                 if (jwtToken.InnerToken.SigningCredentials != null)
                     encodedSignature = JwtTokenUtilities.CreateEncodedSignature(string.Concat(jwtToken.InnerToken.EncodedHeader, ".", jwtToken.EncodedPayload), jwtToken.InnerToken.SigningCredentials);
 
-                return EncryptToken(new JwtSecurityToken(jwtToken.InnerToken.Header, jwtToken.InnerToken.Payload, jwtToken.InnerToken.EncodedHeader, encodedPayload, encodedSignature), jwtToken.EncryptingCredentials).RawData;
+                return EncryptToken(new JwtSecurityToken(jwtToken.InnerToken.Header, jwtToken.InnerToken.Payload, jwtToken.InnerToken.EncodedHeader, encodedPayload, encodedSignature), jwtToken.EncryptingCredentials, jwtToken.InnerToken.Header?.Typ).RawData;
```
mafurman: When would the inner token header be null?

(asking this to better understand this myself :))

keegan-caruso: Oh good catch, it could never be. Header is directly accessed above, if it was null, a NRE would be raised before this line.

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
```diff
@@ -111,6 +127,20 @@ public JwtHeader(SigningCredentials signingCredentials, IDictionary<string,strin
...
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="JwtHeader"/>.
+        /// With the Header Parameters:
+        /// <para>{ { typ, JWT }, { alg, SigningCredentials.Algorithm } }</para>
+        /// </summary>
+        /// <param name="encryptingCredentials"><see cref="EncryptingCredentials"/> used when creating a JWS Compact JSON.</param>
+        /// <param name="outboundAlgorithmMap">provides a mapping for the 'alg' value so that values are within the JWT namespace.</param>
+        /// <param name="tokenType"> provides the token type</param>
+        /// <exception cref="ArgumentNullException">If 'signingCredentials' is null.</exception>
```
keegan-caruso: should be if 'encryptingCredentials' is null

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -448,37 +454,27 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
                     {
                         { JwtHeaderParameterNames.Alg, SecurityAlgorithms.None},
                     };
 
-                if (encryptingCredentials == null)
-                    header.Merge(JObject.FromObject(additionalHeaderClaims));
+            if (encryptingCredentials == null && additionalHeaderClaims != null && additionalHeaderClaims.Count != 0)
```
keegan-caruso: little more clear / expressive to say this as 

`additionalHeaderClaims.Count > 0`

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
RojaEnnam: Good idea to mention default value as well. I made the change.

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
RojaEnnam: Good catch, went unnoticed in previous releases 😆

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: I agree.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
RojaEnnam: That's right. It is accessed before this statement. Not sure if it would be null ever, I think we don't need it.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: That is a good point. It makes sense to me to add as it was provided but originally it was added only in the CreateDefaultJWSHeader, not sure if there is a reason.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -261,6 +270,26 @@ public static TheoryData<CreateTokenTheoryData> CreateJWETheoryData
                     new CreateTokenTheoryData
                     {
                         First = true,
+                        Payload = Default.PayloadString,
```
GeoK: Is Payload still relevant here?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -776,6 +812,25 @@ public static TheoryData<CreateTokenTheoryData> CreateJWSTheoryData
                     new CreateTokenTheoryData
                     {
                         First = true,
+                        TestId = "ValidUsingTokenType",
+                        Payload = Default.PayloadString,
```
GeoK: Same here, Payload seems to be irrelevant.

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
```diff
@@ -78,12 +78,24 @@ public JwtHeader(EncryptingCredentials encryptingCredentials)
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="JwtHeader"/>.
+        /// With the Header Parameters:
+        /// <para>{ { typ, JWT }, { alg, SigningCredentials.Algorithm } }</para>
+        /// </summary>
+        /// <param name="signingCredentials"><see cref="SigningCredentials"/> used when creating a JWS Compact JSON.</param>
+        /// <param name="outboundAlgorithmMap">provides a mapping for the 'alg' value so that values are within the JWT namespace.</param>
+        /// <param name="tokenType"> provides the token type</param>
```
GeoK: nit: extra blank space at the beginning.
Perhaps we could clarify this comment a bit, saying which claim will be populated and what happens if null or empty string is passed.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
GeoK: I just hit the same thing. Can't see a reason why we shouldn't add a token type together with alg: none. Unsigned JWT is still a JWT.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -573,77 +575,67 @@ private JwtSecurityToken EncryptToken(JwtSecurityToken innerJwt, EncryptingCrede
...
+            using (var encryptionProvider = cryptoProviderFactory.CreateAuthenticatedEncryptionProvider(securityKey, encryptingCredentials.Enc))
+            {
+                if (encryptionProvider == null)
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenEncryptionFailedException(LogMessages.IDX12730));
+
+                try
+                {
+                    var header = new JwtHeader(encryptingCredentials, OutboundAlgorithmMap, tokenType);
+                    var encryptionResult = encryptionProvider.Encrypt(Encoding.UTF8.GetBytes(innerJwt.RawData), Encoding.ASCII.GetBytes(header.Base64UrlEncode()));
+                    return JwtConstants.DirectKeyUseAlg.Equals(encryptingCredentials.Alg, StringComparison.Ordinal)? new JwtSecurityToken(
```
GeoK: nit: space in front of "?" (same in M.IM.JWT).

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -573,77 +575,67 @@ private JwtSecurityToken EncryptToken(JwtSecurityToken innerJwt, EncryptingCrede
...
                     throw LogHelper.LogExceptionMessage(new SecurityTokenEncryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10617, SecurityAlgorithms.Aes128CbcHmacSha256, SecurityAlgorithms.Aes192CbcHmacSha384, SecurityAlgorithms.Aes256CbcHmacSha512, encryptingCredentials.Enc)));
 
-                var kwProvider = cryptoProviderFactory.CreateKeyWrapProvider(encryptingCredentials.Key, encryptingCredentials.Alg);
-                var wrappedKey = kwProvider.WrapKey(symmetricKey.Key);
-                using (var encryptionProvider = cryptoProviderFactory.CreateAuthenticatedEncryptionProvider(symmetricKey, encryptingCredentials.Enc))
-                {
-                    if (encryptionProvider == null)
-                        throw LogHelper.LogExceptionMessage(new SecurityTokenEncryptionFailedException(LogMessages.IDX12730));
+                kwProvider = cryptoProviderFactory.CreateKeyWrapProvider(encryptingCredentials.Key, encryptingCredentials.Alg);
+                var symmetricKey = securityKey as SymmetricSecurityKey;
```
GeoK: nit: no real need to initialize a new variable (same in M.IM.JWT)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -795,118 +791,85 @@ public string EncryptToken(string innerJwt, EncryptingCredentials encryptingCred
...
+                else
+                {
+                    plainText = Encoding.UTF8.GetBytes(innerJwt);
+                }
+
+                try
+                {
+                    var rawHeader = Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(header.ToString(Formatting.None)));
+                    var encryptionResult = encryptionProvider.Encrypt(plainText, Encoding.ASCII.GetBytes(rawHeader));
+                    return JwtConstants.DirectKeyUseAlg.Equals(encryptingCredentials.Alg, StringComparison.Ordinal)? string.Join(".", rawHeader, string.Empty, Base64UrlEncoder.Encode(encryptionResult.IV), Base64UrlEncoder.Encode(encryptionResult.Ciphertext), Base64UrlEncoder.Encode(encryptionResult.AuthenticationTag)) :
```
GeoK: nit: consider spliting this line to increase readability.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: Made the change to add 'typ' for unsigned JWT.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
RojaEnnam: Not relevant  here. I removed it.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
RojaEnnam: Fixed it.

src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs
RojaEnnam: Fixed it.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
RojaEnnam: Fixed it.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
RojaEnnam: Done.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: Done.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -795,118 +796,85 @@ public string EncryptToken(string innerJwt, EncryptingCredentials encryptingCred
...
 
-                var kwProvider = cryptoProviderFactory.CreateKeyWrapProvider(encryptingCredentials.Key, encryptingCredentials.Alg);
-                var wrappedKey = kwProvider.WrapKey(symmetricKey.Key);
-
-                using (var encryptionProvider = cryptoProviderFactory.CreateAuthenticatedEncryptionProvider(symmetricKey, encryptingCredentials.Enc))
-                {
-                    if (encryptionProvider == null)
-                        throw LogHelper.LogExceptionMessage(new SecurityTokenEncryptionFailedException(LogMessages.IDX14103));
+                kwProvider = cryptoProviderFactory.CreateKeyWrapProvider(encryptingCredentials.Key, encryptingCredentials.Alg);
+                wrappedKey = kwProvider.WrapKey((securityKey as SymmetricSecurityKey).Key);
```
keegan-caruso: nit: since this will always be created directly above we know it will be a `SymmetricSecurityKey`. But, in the case that this does fail, we would want an InvalidCastException, not a NRE. It would be more clear to do a direct cast rather than an `as` cast.

RojaEnnam: That makes sense. Used to `as `, this is something to keep in mind :)

kevinchalet:
Once you guys have a chance to merge this PR, I'd love to give the new API a try. Could you please make sure nightly builds are pushed to MyGet when it's merged? Thanks!

keegan-caruso:
LGTM other than the duplication

RojaEnnam:
@kevinchalet Nightly build artifact is pushed to MyGet see [here](https://www.myget.org/feed/azureadwebstacknightly/package/nuget/Microsoft.IdentityModel.JsonWebTokens/6.7.2-preview-10917235921)

kevinchalet:
Thanks! However, I'm hitting a strange issue: when using the nightly builds, I reach this assertion: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/opensource/json/Utilities/CollectionUtils.cs#L384

It's not unexpected as `Enumerable.Empty<T>()` doesn't return an array on recent versions of .NET Core, but I'm surprised I see this assert error at all. Do you guys push debug binaries to MyGet?

RojaEnnam:
@kevinchalet that is strange. [This ](https://www.myget.org/feed/azureadwebstacknightly/package/nuget/Microsoft.IdentityModel.JsonWebTokens/6.7.2-preview-10918222843) is from release, can you please try this and let us know how it goes.

kevinchalet:
@RojaEnnam as expected, it works like a charm! Do you have a rough idea of when a stable version including this change will be pushed to NuGet.org?

In a future PR, you'll probably want to sync your local copy of the JSON.NET types, as JSON.NET was updated to remove this assertion, that is no longer true on .NET Core 3.0: https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Utilities/CollectionUtils.cs#L369-L380

RojaEnnam:
@kevinchalet That's great! We are targeting to do the next release in couple of weeks.

Thanks for letting us know regarding JSON.NET types.

brentschmaltz (approved):

brentschmaltz:
@kevinchalet we are looking to release this with a couple of other changes in a few days. We still had to punt a couple of features slated for 6.7.2, as the team had some other internal priorities. Hopefully we can deliver the next release out in a smaller time interval that between 6.7.1 and 6.7.2.

kevinchalet:
@brentschmaltz no problem. It's a welcome improvement that makes things simpler and cleaner, but we can objectively live without it, so I understand it's not a high-pri thing 😃

PR #1526, opened 2020-09-09
Make CanonicalString public

### LINKED ISSUE #1476: Microsoft.IdentityModel.Tokens.Saml2.Saml2Serializer Write Signature if included in Assertion

RojaEnnam:
Makes `CanonicalString `property public on `SamlAssertion `and `Saml2Assertion`.


keegan-caruso:
This looks like a duplicate of #1481

RojaEnnam:
> This looks like a duplicate of #1481

Oh you're right I have seen this couple of weeks ago. Forgot about it. I will close this PR.

PR #1532, opened 2020-09-17
Fixes WS-Security serialization (AzureAD#1531)

gislikonrad:
Fixes WS-Security serialization within the WS-Trust package.


brentschmaltz:
Any work here will have to be against the WSTrust6.8.0

PR #1535, opened 2020-10-01
Remove locks in SignatureProviders

brentschmaltz:
This PR adds an object pool that allows for the removal of locks when preforming crypto operations involving signatures. The 'objectpool' was copied from .net core as it is not available in 4.x.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
@@ -67,18 +67,12 @@ namespace Microsoft.IdentityModel.Tokens
 
 #if NET461 || NETSTANDARD2_0
         private RSAEncryptionPadding _rsaEncryptionPadding;
-        private object _signRsaLock = new object();
 #endif
 
         private bool _disposeCryptoOperators = false;
         private bool _disposed = false;
         private SignDelegate SignatureFunction;
         private VerifyDelegate VerifyFunction;
-
-
-        private object _signEcdsaLock = new object();
-        private object _verifyRsaLock = new object();
-        private object _verifyEcdsaLock = new object();
 
 #if NET461 || NETSTANDARD2_0
         // HasAlgorithmName was introduced into Net46
@@ -344,19 +338,13 @@ namespace Microsoft.IdentityModel.Tokens
 
         private byte[] SignWithECDsa(byte[] bytes)
         {
-            lock (_signEcdsaLock)
-            {
-                return ECDsaSecurityKey.ECDsa.SignHash(HashAlgorithm.ComputeHash(bytes));
-            }
+            return ECDsaSecurityKey.ECDsa.SignHash(HashAlgorithm.ComputeHash(bytes));
         }
 
 #if NET461 || NETSTANDARD2_0
         private byte[] SignWithRsa(byte[] bytes)
         {
-            lock (_signRsaLock)
-            {
-                return RSA.SignHash(HashAlgorithm.ComputeHash(bytes), HashAlgorithmName, RSASignaturePadding);
-            }
+            return RSA.SignHash(HashAlgorithm.ComputeHash(bytes), HashAlgorithmName, RSASignaturePadding);
         }
 #endif
 
@@ -378,29 +366,20 @@ namespace Microsoft.IdentityModel.Tokens
 
         private bool VerifyWithECDsa(byte[] bytes, byte[] signature)
         {
-            lock (_verifyEcdsaLock)
-            {
-                return ECDsaSecurityKey.ECDsa.VerifyHash(HashAlgorithm.ComputeHash(bytes), signature);
-            }
+            return ECDsaSecurityKey.ECDsa.VerifyHash(HashAlgorithm.ComputeHash(bytes), signature);
         }
 
 #if NET461 || NETSTANDARD2_0
         private bool VerifyWithRsa(byte[] bytes, byte[] signature)
         {
-            lock (_verifyRsaLock)
-            {
-                return RSA.VerifyHash(HashAlgorithm.ComputeHash(bytes), signature, HashAlgorithmName, RSASignaturePadding);
-            }
+            return RSA.VerifyHash(HashAlgorithm.ComputeHash(bytes), signature, HashAlgorithmName, RSASignaturePadding);
         }
 #endif
 
 #if DESKTOP
         private bool VerifyWithRsaCryptoServiceProviderProxy(byte[] bytes, byte[] signature)
         {
-            lock (_verifyRsaLock)
-            {
-                return RsaCryptoServiceProviderProxy.VerifyData(bytes, HashAlgorithm, signature);
-            }
+            return RsaCryptoServiceProviderProxy.VerifyData(bytes, HashAlgorithm, signature);
         }
 #endif
 
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
@@ -71,8 +71,8 @@ namespace Microsoft.IdentityModel.Tokens
 
         private bool _disposeCryptoOperators = false;
         private bool _disposed = false;
-        private SignDelegate SignatureFunction;
-        private VerifyDelegate VerifyFunction;
+        private SignDelegate SignatureFunction = SignatureFunctionNotFound;
+        private VerifyDelegate VerifyFunction = VerifyFunctionNotFound;
 
 #if NET461 || NETSTANDARD2_0
         // HasAlgorithmName was introduced into Net46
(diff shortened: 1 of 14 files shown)


src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -153,10 +153,10 @@ public AsymmetricSignatureProvider(SecurityKey key, string algorithm, bool willC
             if (!_cryptoProviderFactory.IsSupportedAlgorithm(algorithm, key))
                 throw LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant(LogMessages.IDX10634, (algorithm ?? "null"), key)));
 
-            ValidateAsymmetricSecurityKeySize(key, algorithm, willCreateSignatures);
-            _asymmetricAdapter = ResolveAsymmetricAdapter(jsonWebKey?.ConvertedSecurityKey ?? key, algorithm, willCreateSignatures);
             WillCreateSignatures = willCreateSignatures;
-        }
+            ValidateAsymmetricSecurityKeySize(key, algorithm, WillCreateSignatures);
+            _asymmetricAdapterObjectPool = new DisposableObjectPool<AsymmetricAdapter>(CreateAsymmetricAdapter);
```
keegan-caruso: It would be good if we could provide a configuration option so we can use the other ctor and allow the user to specify pool size.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -203,10 +203,10 @@ protected virtual HashAlgorithmName GetHashAlgorithmName(string algorithm)
             return SupportedAlgorithms.GetHashAlgorithmName(algorithm);
         }
 
-        private AsymmetricAdapter ResolveAsymmetricAdapter(SecurityKey key, string algorithm, bool requirePrivateKey)
+        private AsymmetricAdapter CreateAsymmetricAdapter()
         {
-            var hashAlgoritmName = GetHashAlgorithmName(algorithm);
-            return new AsymmetricAdapter(key, algorithm, _cryptoProviderFactory.CreateHashAlgorithm(hashAlgoritmName), hashAlgoritmName, requirePrivateKey);
+            var hashAlgoritmName = GetHashAlgorithmName(Algorithm);
```
keegan-caruso, line 208: We could use a Lazy<bool> with ValidateAsymmetricSecurityKeySize to remove the virtual call in the ctor.

src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
```diff
@@ -238,20 +226,23 @@ public override bool Verify(byte[] input, byte[] signature)
...
-                lock (_verifyLock)
-                {
-                    return Utility.AreEqual(signature, KeyedHashAlgorithm.ComputeHash(input));
-                }
+                return Utility.AreEqual(signature, keyedHashAlgorithm.ComputeHash(input));
             }
             catch
             {
                 CryptoProviderCache?.TryRemove(this);
+                Dispose(true);
```
keegan-caruso, line 243: Suppose two threads are using the object pool

T1 acquires an element from the pool
T2 acquires an element and throws and hits this, disposing the the element that T1 acquires, T1 throws an object disposed exception on use.

Applies more so with larger object pools.

Am I missing something here? Is there anything that we can do in that case?

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
GeoK: +1

GeoK: Looking at this again, this property should probably be on the CryptoProviderFactory, otherwise it will have no effect on Wilson.

src/Microsoft.IdentityModel.Tokens/ObjectPools.cs
```diff
@@ -0,0 +1,139 @@
+//------------------------------------------------------------------------------
```
GeoK, line 1: Should we rename this class to "DisposableObjectPool"?

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -378,29 +366,20 @@ internal bool Verify(byte[] bytes, byte[] signature)
...
 #endif
 
 #if DESKTOP
         private bool VerifyWithRsaCryptoServiceProviderProxy(byte[] bytes, byte[] signature)
         {
-            lock (_verifyRsaLock)
-            {
-                return RsaCryptoServiceProviderProxy.VerifyData(bytes, HashAlgorithm, signature);
-            }
+            return RsaCryptoServiceProviderProxy.VerifyData(bytes, HashAlgorithm, signature);
```
GeoK, line 380: RsaCryptoServiceProviderProxy still has sign/verify locks. We should be able to remove them now, correct?

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -203,10 +203,10 @@ protected virtual HashAlgorithmName GetHashAlgorithmName(string algorithm)
...
+            return new AsymmetricAdapter(Key, Algorithm, _cryptoProviderFactory.CreateHashAlgorithm(hashAlgoritmName), hashAlgoritmName, WillCreateSignatures);
```
GeoK, line 209: Was the initial reason for adding locks the lack of thread-safety guarantees for HashAlgorithm and/or some other objects? 

Even though each thread uses it's own AsymmetricAdapter, they are all using the same key. Are sign/verify ops on our Key objects thread-safe?

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
brentschmaltz: @keegan-caruso yes, but isn't this https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1459 ?

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
brentschmaltz: I deliberately didn't put a limit on the pool size as it is effectively tied to the number of threads.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
brentschmaltz: @GeoK yes they use the same key and algorithm. For RSA params, we should be ok. I will have to review what happens if a RSA object is passed in. 

ComputeHash is not thread safe, Sign is, so we this is fine.

src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
brentschmaltz: @keegan-caruso T1 and T2 will have different instances of the element, that is the idea of the object pool.
As George pointed out below, we will have to think about when an RSA or ECD object is passed to the constructor of the SecurityKey, we may have to recreate a new object. However that could be cause issues with derived types.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
keegan-caruso: Correct, I will be able to more easily add that with this change

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
GeoK: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/26396d50b996d9e19cda286f614b58068aa4528f/src/Microsoft.IdentityModel.Tokens/ObjectPools.cs#L63
In a given moment, (2 * #logicalCores) objects would be used from a pool, while AsymmetricSignatureProvider needs to create a new AsymmetricAdapter for all other request that are happening simultaneously.
I guess the question here is whether we should allow users to change this i.e. to increase/decrease object pool size.

brentschmaltz:
@GeoK @keegan-caruso i upadate the PR, removing the locks in RsaCryptoServiceProxy and adding a programming possibility (not particularly elegant) opportunity to set the pool size.

I struggled with naming and settled with DefaultSignatureProviderObjectPoolCacheSize, specifying using SignatureProvider in the name, rather than a more generic DefaultObjectPoolCacheSize, in case someone wanted to have different sizes (highly unlikely, however who knows).

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -41,6 +41,8 @@ public class CryptoProviderFactory
         private static CryptoProviderFactory _default;
         private static ConcurrentDictionary<string, string> _typeToAlgorithmMap = new ConcurrentDictionary<string, string>();
         private static object _cacheLock = new object();
+        private static int _defaultSignatureProviderObjectPoolCacheSize = Environment.ProcessorCount * 4;
```
GeoK, line 44: Just curios on why we decided to change the default object pool size?
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/26396d50b996d9e19cda286f614b58068aa4528f/src/Microsoft.IdentityModel.Tokens/ObjectPools.cs#L63

keegan-caruso (approved):
LGTM, curious about @GeoK 's questions and stress / perf findings

PR #1549, opened 2020-11-10
Do not call overridable methods in constructors

### LINKED ISSUE #1459: Do not call overridable methods in constructors

keegan-caruso:
AsymmetricSignautreProvider, AuthenticatedEncryptionProvider,
RsaKeyWrapProvider, and SymmetricAlgorithm called virtuals in their ctor.
This moves that call to a lazy<t> and removes the suppression entries


src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
```diff
@@ -150,6 +158,9 @@ protected virtual SymmetricAlgorithm GetSymmetricAlgorithm(SecurityKey key, stri
             if (key == null)
                 throw LogHelper.LogArgumentNullException(nameof(key));
 
+            if (!IsSupportedAlgorithm(key, algorithm))
```
GeoK, line 161: Not sure if this is required considering that ValidateKeySize() later on makes a similar check.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -1924,7 +1924,7 @@ public static TheoryData<CreateTokenTheoryData> RoundTripJWEKeyWrapTestCases
                         Payload = Default.PayloadString,
                         SigningCredentials = Default.SymmetricSigningCredentials,
                         EncryptingCredentials = new EncryptingCredentials(Default.SymmetricEncryptionKey256, SecurityAlgorithms.Aes128KW, SecurityAlgorithms.Aes128CbcHmacSha256),
-                        ExpectedException = ExpectedException.ArgumentOutOfRangeException("IDX10662:")
+                        ExpectedException = ExpectedException.SecurityTokenKeyWrapException("IDX10662:")
```
GeoK, line 1927: Something doesn't add up here. Only these two lines are throwing exceptions with IDX10662, and both of them are ArgumentOutOfRangeExceptions.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/1a1ff1f75bf042451c136e1246690c036a37a4d6/src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs#L343

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/1a1ff1f75bf042451c136e1246690c036a37a4d6/src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs#L351

keegan-caruso: I would have change this to include the inner exception, but the exception code is using the exception as a message instead of as an inner exception, this is coming from the wrap or unwrap, since the validation is deferred 

So something like this is going through the deferred validation code then catching and setting the exception message.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/1a1ff1f75bf042451c136e1246690c036a37a4d6/src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs#L238

GeoK: Oh, I see, e.g. SecurityTokenKeyWrapException IDX10658 was thrown, and you are checking for the exception type and id of the original message (IDX10662).

GeoK (approved):

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -154,9 +155,9 @@ public AsymmetricSignatureProvider(SecurityKey key, string algorithm, bool willC
                 throw LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant(LogMessages.IDX10634, (algorithm ?? "null"), key)));
 
             WillCreateSignatures = willCreateSignatures;
-            ValidateAsymmetricSecurityKeySize(key, algorithm, WillCreateSignatures);
+            _keySizeIsValid = new Lazy<bool>(ValidKeySize);
```
brentschmaltz, line 158: @keegan-caruso This is a clever solution.
Extending the idea, is there a need for the Lazy<bool>, could we call the method directly inside CreateAsymmetricAdapter?

src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
brentschmaltz: @keegan-caruso @GeoK reviewing this points out a kink in our extensibility. IsSupportedAlgorithm is virtual, however ValidateKeySize is not and requires one of: Aes128KW, Aes128KeyWrap, Aes256KW, Aes256KeyWrap.

I opened https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1569

brentschmaltz (approved):

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
keegan-caruso: @brentschmaltz That is a good question. We don't need the lazy, I included it to keep `ValidateAsymmetricSecurityKeySize` as only being called once.  Previously in the ctor, now once in the Lazy.

It would simplify this if we move the call to CreateAsymmetricAdapter. Do you think it is important to keep ValidateAsymmetricSecurityKeySize as only being called once?

PR #1552, opened 2020-11-13
Added net472 target

### LINKED ISSUE #1507: Add Target 4.7.2
With 4.7.2, we will be able to move the adapters for RSA and ECD for just 461 and 45.

For this issue, we can open up Creating a JsonWebKey from EC key.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1317

mafurman:
Addresses #1507 

The net472 ECDsaAdaper calls CreateECDsaUsingECParams (similar to the netstandard2_0 target).

However, the AsymmetricAdapter still makes use of RsaCryptoProviderProxy on the net472 target.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
@@ -33,7 +33,7 @@ using Microsoft.IdentityModel.Logging;
 using System.Reflection;
 #endif
 
-#if NET461 || NETSTANDARD2_0
+#if NET461 || NET472 || NETSTANDARD2_0
 using System.Security.Cryptography.X509Certificates;
 #endif
 
@@ -65,7 +65,7 @@ namespace Microsoft.IdentityModel.Tokens
         private bool _useRSAOeapPadding = false;
 #endif
 
-#if NET461 || NETSTANDARD2_0
+#if NET461 || NET472 || NETSTANDARD2_0
         private RSAEncryptionPadding _rsaEncryptionPadding;
 #endif
 
@@ -74,7 +74,7 @@ namespace Microsoft.IdentityModel.Tokens
         private SignDelegate SignatureFunction = SignatureFunctionNotFound;
         private VerifyDelegate VerifyFunction = VerifyFunctionNotFound;
 
-#if NET461 || NETSTANDARD2_0
+#if NET461 || NET472 || NETSTANDARD2_0
         // HasAlgorithmName was introduced into Net46
         internal AsymmetricAdapter(SecurityKey key, string algorithm, HashAlgorithm hashAlgorithm, HashAlgorithmName hashAlgorithmName, bool requirePrivateKey)
             : this(key, algorithm, hashAlgorithm, requirePrivateKey)
@@ -169,8 +169,8 @@ namespace Microsoft.IdentityModel.Tokens
                 return RSA.DecryptValue(data);
 #endif
 
-            // NET461 could have been passed RSACryptoServiceProvider
-#if NET461
+            // NET461 or NET472 could have been passed RSACryptoServiceProvider.
+#if NET461 || NET472
             if (RsaCryptoServiceProviderProxy != null)
                 return RsaCryptoServiceProviderProxy.Decrypt(data, _useRSAOeapPadding);
             else
@@ -227,8 +227,8 @@ namespace Microsoft.IdentityModel.Tokens
                 return RSA.EncryptValue(data);
 #endif
 
-            // NET461 could have been passed RSACryptoServiceProvider
-#if NET461
+            // NET461 or NET472 could have been passed RSACryptoServiceProvider
+#if NET461 || NET472
             if (RsaCryptoServiceProviderProxy != null)
                 return RsaCryptoServiceProviderProxy.Encrypt(data, _useRSAOeapPadding);
 
@@ -243,7 +243,7 @@ namespace Microsoft.IdentityModel.Tokens
 
         private HashAlgorithm HashAlgorithm { get; set; }
 
-#if NET461 || NETSTANDARD2_0
+#if NET461 || NET472 || NETSTANDARD2_0
         private HashAlgorithmName HashAlgorithmName { get; set; }
 
         private RSASignaturePadding RSASignaturePadding { get; set; }
@@ -252,7 +252,7 @@ namespace Microsoft.IdentityModel.Tokens
         private void InitializeUsingRsa(RSA rsa, string algorithm)
         {
 
-#if NET461 || NETSTANDARD2_0
+#if NET461 || NET472 || NETSTANDARD2_0
             if (algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha256, StringComparison.Ordinal) ||
                 algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha256Signature, StringComparison.Ordinal) ||
                 algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha384, StringComparison.Ordinal) ||
@@ -310,7 +310,7 @@ namespace Microsoft.IdentityModel.Tokens
             }
 #endif
 
-#if NET461 || NETSTANDARD2_0
+#if NET461 || NET472 || NETSTANDARD2_0
             // Here we can use RSA straight up.
             _rsaEncryptionPadding = (algorithm.Equals(SecurityAlgorithms.RsaOAEP, StringComparison.Ordinal) || algorithm.Equals(SecurityAlgorithms.RsaOaepKeyWrap, StringComparison.Ordinal))
                         ? RSAEncryptionPadding.OaepSHA1
@@ -337,7 +337,7 @@ namespace Microsoft.IdentityModel.Tokens
             return ECDsaSecurityKey.ECDsa.SignHash(HashAlgorithm.ComputeHash(bytes));
         }
 
-#if NET461 || NETSTANDARD2_0
+#if NET461 || NET472 || NETSTANDARD2_0
         private byte[] SignWithRsa(byte[] bytes)
         {
             return RSA.SignHash(HashAlgorithm.ComputeHash(bytes), HashAlgorithmName, RSASignaturePadding);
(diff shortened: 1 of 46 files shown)


test/Microsoft.IdentityModel.Tokens.Tests/SecurityKeyTests.cs
```diff
@@ -198,7 +198,7 @@ public static TheoryData<SecurityKeyTheoryData> CreateInternalIdsTestCases
                         ExpectedInternalId = Base64UrlEncoder.Encode(KeyingMaterial.JsonWebKeyRsa_2048_Public.ComputeJwkThumbprint()),
                         TestId = nameof(KeyingMaterial.JsonWebKeyRsa_2048_Public)
                     },
-#if NET_CORE
+#if NET472 || NET_CORE
```
GeoK, line 201: nit: comment on ln216 is no longer true.

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
```diff
@@ -271,7 +271,7 @@ public static TheoryData<CryptoProviderCacheTheoryData> TryAddTheoryData
                         SignatureProvider = new SymmetricSignatureProvider(KeyingMaterial.DefaultSymmetricSecurityKey_384, ALG.HmacSha384, true),
                         TestId = nameof(KeyingMaterial.DefaultSymmetricSecurityKey_256)
                     },
-#if NET_CORE
+#if NET472 || NET_CORE
```
GeoK, line 274: nit: comment on ln275 is no longer true

GeoK:
Are there any changes required here?
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/30c16ab82cf4338c6d6e04a5d4094795aa83c587/src/Microsoft.IdentityModel.Xml/Microsoft.IdentityModel.Xml.csproj#L33

GeoK:
[Pss ](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rsasignaturepadding.pss?view=net-5.0)is available on net472
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/30c16ab82cf4338c6d6e04a5d4094795aa83c587/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs#L1507

GeoK:
We should test RsaCryptoServiceProviderProxy using net472, imo. RSA keys could be created of X509 certs, where keys are created with CAPI and old cryptographic providers that don't know about sha2.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/30c16ab82cf4338c6d6e04a5d4094795aa83c587/test/Microsoft.IdentityModel.Tokens.Tests/RsaCryptoServiceProviderProxyTests.cs#L28

test/Microsoft.IdentityModel.Tokens.Tests/ECDsaSecurityKeyTests.cs
```diff
@@ -74,7 +74,7 @@ public void Defaults()
         [Fact]
         public void CanComputeJwkThumbprint()
         {
-#if NET_CORE
+#if NET472 || NET_CORE
```
GeoK, line 77: nit: messages below are no longer true.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -170,7 +170,7 @@ internal byte[] Decrypt(byte[] data)
 #endif
 
             // NET461 could have been passed RSACryptoServiceProvider
-#if NET461
+#if NET461 || NET472
```
GeoK, line 173: nit: comment above could be updated. same with ln 231.

GeoK:
We should update these to reflect the changes:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/Supported-Algorithms
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/IDX10690

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
```diff
@@ -116,7 +116,7 @@ public override int KeySize
         /// <remarks>https://tools.ietf.org/html/rfc7638</remarks>
         public override bool CanComputeJwkThumbprint()
         {
-#if NETSTANDARD2_0
+#if NET472 || NETSTANDARD2_0
```
GeoK, line 119: net472 supports [ECParameters](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecparameters?view=net-5.0#moniker-applies-to), there is no need for this check. Similar applies to ln 133.

You can also keep this logic as is, but refactor SupportsECParameters to return **true** on net472, without the unnecessary check.

test/Microsoft.IdentityModel.Tokens.Tests/SecurityKeyTests.cs
mafurman: Good catch. Fixed!

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
mafurman: Fixed!

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
mafurman: Fixed!

src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
mafurman: I've refactored SupportsECParameters to return true.

mafurman:
> 
> 
> Are there any changes required here?
> 
> https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/30c16ab82cf4338c6d6e04a5d4094795aa83c587/src/Microsoft.IdentityModel.Xml/Microsoft.IdentityModel.Xml.csproj#L33

I don't think so? It doesn't seem like those two assemblies are directly referenced in this project.

> 
> 
> [Pss ](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rsasignaturepadding.pss?view=net-5.0)is available on net472
> 
> https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/30c16ab82cf4338c6d6e04a5d4094795aa83c587/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs#L1507

Fixed!




> 
> 
> We should test RsaCryptoServiceProviderProxy using net472, imo. RSA keys could be created of X509 certs, where keys are created with CAPI and old cryptographic providers that don't know about sha2.
> 
> https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/30c16ab82cf4338c6d6e04a5d4094795aa83c587/test/Microsoft.IdentityModel.Tokens.Tests/RsaCryptoServiceProviderProxyTests.cs#L28

Good point, thanks! :) I think this was left over from before we realized RsaCryptoServiceProviderProxy was necessary on the net472 target.



> 
> 
> We should update these to reflect the changes:
> https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/Supported-Algorithms
> https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/IDX10690

I created a separate issue to track this so I can add this in after the PR is merged in. :)
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1556

GeoK (approved):
LGTM. Still unsure about Microsoft.IdentityModel.Xml.csproj; should we add references (Microsoft.CSharp and System.Net.Http) to 472 or remove them for net45 and net461?

test/Microsoft.IdentityModel.Tokens.Tests/AsymmetricAdapterTests.cs
```diff
@@ -39,7 +39,7 @@ namespace Microsoft.IdentityModel.Tokens.Tests
     public class AsymmetricAdapterTests
     {
         [Theory, MemberData(nameof(AsymmetricAdapterUsageTestCases))]
-        public void AsymmetricAdapterUsageTests(AsymmetricAdatperTheoryData theoryData)
```
brentschmaltz, line 42: thanks!

src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs
```diff
@@ -52,12 +52,14 @@ internal class ECDsaAdapter
         /// </exception>
         internal ECDsaAdapter()
         {
-#if NETSTANDARD2_0
+#if NET472
+            CreateECDsaFunction = CreateECDsaUsingECParams;
+#elif NETSTANDARD2_0
```
brentschmaltz, line 57: @mafurman Could you add a comment here indicating that although NETSTANDARD2_0 specifies that ECParameters is supported, we still need to call SupportsECParameters() as 4..6.1 is listed as supporting NETSTANDARD2_0, but DOES NOT support ECParameters. see: https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecparameters?view=netstandard-2.0

This may help future devs as they may not understand the issues.

brentschmaltz (approved):
suggested adding a comment.

mafurman:
@GeoK I've gone ahead and removed the Microsoft.CSharp and System.Net.Http references from the net45 and net461 targets in Microsoft.IdentityModel.Xml.csproj. I think this is something we may want to do for the other .csproj files we have as well, so I created a separate issue to track this: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1572

@brentschmaltz That's a good suggestion. I added your comment in.

PR #1565, opened 2020-12-09
Use new script code

### LINKED ISSUE #1564: Provide CSP 'unsafe-eval' compliant default value for _script
The initial value for `AuthenticationProtocolMessage._script` 
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/824068adaeb7b158df2b36d8787f689646e715e0/src/Microsoft.IdentityModel.Protocols/AuthenticationProtocolMessage.cs#L43 
requires the CSP value 'unsafe-eval' to be executed. This is a no-go from security perspective.

Please provide a value which does not require this CSP setting:

`<script language="javascript">window.setTimeout(function() {document.forms[0].submit();}, 0);</script>`

twenzel:
Use new script code which does not use JS evaluation.
Fixes #1564


jasonnutter (approved):

mafurman:
@twenzel Could you please rebase against the dev branch and update the PR? We'll pull it in after that. :)

PR #1575, opened 2021-01-05
Added cache eviction for the InMemoryCryptoProviderCache

### LINKED ISSUE #1447: Crypto provider cache entries should be automatically evicted
Currently, items added to crypto provider cache will never expire.
We should consider evicting these items from the cache after certain time, or if they are not accessed in a given span of time.

mafurman:
Fixes #1447.

The net461, net472 and netstandard2.0 make use of Microsoft.Extensions.Caching.Memory.MemoryCache, a LRU cache which will begin to evict items once it's close to max capacity.

The net45 target makes use of a simple wrapper (MaxCapacityRefreshCache) over a ConcurrentDictionary that fully clears the cache whenever the max capacity is reached. After careful consideration, it was decided that adding a more intelligent caching solution for the net45 target wasn't worth it (but if anyone believes otherwise please let me know!).

I've added a simple test that checks to ensure that the cache is cleared when max capacity is reached, but I haven't added tests for any of the other methods as the tests for the InMemoryCryptoProviderCache already cover this functionality + adding these tests would just be testing the underlying ConcurrentDictionary.

I haven't done it yet, but I think doing some kind of performance testing for this change would have a lot of value.


src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
```diff
@@ -0,0 +1,137 @@
...
+{
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// A cache that clears itself upon reaching max capacity.
+    /// </summary>
+    /// <typeparam name="TKey">The type of cache key.</typeparam>
+    /// <typeparam name="TValue">The type of cached value.</typeparam>
+    internal class MaxCapacityRefreshCache
```
mafurman, line 36: Not sure if this is the best name for this class.

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        /// <summary>
+        /// Recycles the cache if needed.
+        /// </summary>
+        private void RecycleCacheIfNeeded()
+        {
+            if (SizeLimit == Cache.Keys.Count)
+            {
+                lock (RecycleLock)
```
mafurman: I'm not sure if we really need to lock here.

src/Microsoft.IdentityModel.Tokens/TokenCacheOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <summary>
+    /// Specifies the TokenCacheOptions which can be used to configure the internal cryptoprovider cache.
+    /// For the net461, net472, and netstandard2.0 targets we are using the Microsoft.Extensions.Caching.Memory.MemoryCache class:
+    /// https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycache?view=dotnet-plat-ext-3.1
+    /// For the net45 target we are using our own simple caching implementation which clears the entire cache
+    /// once the cache is at max capacity. We recommend upgrading to net461+ for a more comprehensive caching experience.
+    /// Any property on these TokenCacheOptions that corresponds directly to a property 
+    /// used by Microsoft.Extensions.Caching.Memory.MemoryCache has the same name.
+    /// </summary>
+    public class TokenCacheOptions
```
mafurman: I'll be adding documentation/samples on how to pass these options in once this PR is merged in.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -26,20 +26,60 @@
...
+    public class InMemoryCryptoProviderCache : CryptoProviderCache, IDisposable
+#elif NET45
+    public class InMemoryCryptoProviderCache: CryptoProviderCache
+#endif
     {
-        private ConcurrentDictionary<string, SignatureProvider> _signingSignatureProviders = new ConcurrentDictionary<string, SignatureProvider>();
-        private ConcurrentDictionary<string, SignatureProvider> _verifyingSignatureProviders = new ConcurrentDictionary<string, SignatureProvider>();
+        internal TokenCacheOptions _tokenCacheOptions;
+#if NET461 || NET472 || NETSTANDARD2_0 
+        private MemoryCache _signingSignatureProviders;
```
keegan-caruso: _signingSignatureProviders and _verifyingSignatureProviders should be readonly in both targets

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
```diff
@@ -0,0 +1,137 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    using System.Collections.Concurrent;
```
keegan-caruso: we typically put these outside the namespace

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// <typeparam name="TKey">The type of cache key.</typeparam>
+    /// <typeparam name="TValue">The type of cached value.</typeparam>
+    internal class MaxCapacityRefreshCache
+        <TKey, TValue>
+        where TValue : class
+    {
+        /// <summary>
+        /// Gets or sets the cache.
+        /// </summary>
+        private ConcurrentDictionary<TKey, TValue> Cache { get; set; }
```
keegan-caruso: none of these properties need the setter

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
```diff
@@ -0,0 +1,137 @@
...
+            RecycleCacheIfNeeded();
+            return Cache.TryRemove(cacheKey, out removedValue);
+        }
+
+        /// <summary>
+        /// Recycles the cache if needed.
+        /// </summary>
+        private void RecycleCacheIfNeeded()
+        {
+            if (SizeLimit == Cache.Keys.Count)
```
keegan-caruso: safer to use <= , multiple threads calling set could bump this over SizeLimit

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
keegan-caruso: ConcurrentDictionary.Clear should do a "table level" lock. it does in 4.8, need to verify that it does in 4.5

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
```diff
@@ -0,0 +1,137 @@
...
+            return Cache.TryGetValue(cacheKey, out value);           
+        }
+
+        /// <summary>
+        /// Sets the cache value.
+        /// </summary>
+        /// <param name="cacheKey">The cache key.</param>
+        /// <param name="value">The value.</param>
+        /// <returns><paramref name="value"/> if it was successfully set, null otherwise.</returns>
+        /// <remarks>This method is non-blocking and thread-safe.</remarks>
```
keegan-caruso: misleading to call these non-blocking when some of the methods in ConcurrentDictionary can take a lock and block

src/Microsoft.IdentityModel.Tokens/TokenCacheOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Specifies the TokenCacheOptions which can be used to configure the internal cryptoprovider cache.
+    /// For the net461, net472, and netstandard2.0 targets we are using the Microsoft.Extensions.Caching.Memory.MemoryCache class:
+    /// https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycache?view=dotnet-plat-ext-3.1
```
keegan-caruso: link is misleading since it takes you to the wrong view, might be better to just drop the view param completely?

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <remarks>This value will not be set if the cache already contains a key with the same name as <paramref name="cacheKey"/>.</remarks>
```
keegan-caruso: This is a bit surprising behavior, why not let a set update the value in the cache?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -163,9 +221,23 @@ public override bool TryGetSignatureProvider(SecurityKey securityKey, string alg
...
+                return signatureProvider != null;
+#endif
+            }
             else
+            {
+#if NET461 || NET472 || NETSTANDARD2_0 
                 return _verifyingSignatureProviders.TryGetValue(cacheKey, out signatureProvider);
+#elif NET45
+                _verifyingSignatureProviders.TryGetValue(cacheKey, out signatureProvider);
+                return signatureProvider != null;
```
keegan-caruso: can this not be the same as directly above and return the TryGetValue's bool directly?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -184,24 +256,76 @@ public override bool TryRemove(SignatureProvider signatureProvider)
...
                     provider.CryptoProviderCache = null;
                     return true;
                 }
+                else
+                {
+                    return false;
+                }
+#endif
+            }
+            catch
```
keegan-caruso: we at least need the exception for diagnostics right? What exceptions can be thrown in the try block?

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
```diff
@@ -0,0 +1,137 @@
...
+        internal bool TryRemove(TKey cacheKey, out TValue removedValue)
+        {
+            RecycleCacheIfNeeded();
+            return Cache.TryRemove(cacheKey, out removedValue);
+        }
+
+        /// <summary>
+        /// Recycles the cache if needed.
+        /// </summary>
+        private void RecycleCacheIfNeeded()
```
keegan-caruso, line 117: Do we need any read after write guarantees on this cache? I think this behavior would be surprising to most users. I think a simple LRU implementation would be better in most cases. Thoughts?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
RojaEnnam: +1, would be nice if logged when an exception is caught.

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
```diff
@@ -0,0 +1,137 @@
...
+            Cache = new ConcurrentDictionary<TKey, TValue>(comparer ?? EqualityComparer<TKey>.Default);
+        }
+
+        internal bool Contains(TKey key)
+        {
+            return Cache.ContainsKey(key);
+        }
+
+        /// <summary>
+        /// Gets the value corresponding to <paramref name="value"/> from the cache.
```
RojaEnnam: Guessing it should be name ="cacheKey"?

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <param name="value">The cached value.</param>
+        /// <remarks>This method is non-blocking and thread-safe.</remarks>
+        internal bool TryGetValue(TKey cacheKey, out TValue value)
+        {
+            RecycleCacheIfNeeded();
+            return Cache.TryGetValue(cacheKey, out value);           
+        }
+
+        /// <summary>
+        /// Sets the cache value.
```
RojaEnnam: Should this be CacheKey's value or similar?

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
RojaEnnam: +1, if we want to keep the Cache behavior consistent across different targets.
That said, for my knowledge I'd like to know if there is a compelling reason to go with this approach.

src/Microsoft.IdentityModel.Tokens/TokenCacheOptions.cs
GeoK: Would it be more honest to call this class CryptoProviderCacheOptions?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -118,22 +158,40 @@ public override bool TryAdd(SignatureProvider signatureProvider)
...
+            MaxCapacityRefreshCache<string, SignatureProvider> signatureProviderCache;
+#endif
+            // Determine if we are caching a signing or a verifying SignatureProvider.
             if (signatureProvider.WillCreateSignatures)
+                signatureProviderCache = _signingSignatureProviders;
+            else
+                signatureProviderCache = _verifyingSignatureProviders;
+
+#if NET461 || NET472 || NETSTANDARD2_0 
+            var cacheEntryOptions = new MemoryCacheEntryOptions
```
GeoK: nit: cacheEntryOptions could be created only if signatureProvider is not already cached to save some cycles.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -26,20 +26,60 @@
...
 {
     /// <summary>
     /// Defines a cache for crypto providers.
     /// Current support is limited to <see cref="SignatureProvider"/> only.
     /// </summary>
-    public class InMemoryCryptoProviderCache : CryptoProviderCache
+#if NET461 || NET472 || NETSTANDARD2_0 
+    public class InMemoryCryptoProviderCache : CryptoProviderCache, IDisposable
+#elif NET45
+    public class InMemoryCryptoProviderCache: CryptoProviderCache
```
GeoK, line 45: Are SignatureProviders stored in MaxCapacityRefreshCache going to be properly disposed in net45?

build/dependencies.props
```diff
@@ -1,5 +1,6 @@
 <Project>
   <PropertyGroup>
+    <CachingVersion>2.1.1</CachingVersion>
```
GeoK, line 3: Dependency on [Microsoft.Extensions.Caching.Memory/2.1.1](https://www.nuget.org/packages/Microsoft.Extensions.Caching.Memory/2.1.1) will require users on .net >= 4.6.1 to consume .netstandard2.0 metapackage. In some cases this could cause conflicts.
It might be worthwhile exploring if it would be better to take a dependency on [5.0.0](https://www.nuget.org/packages/Microsoft.Extensions.Caching.Memory/5.0.0) as it has a dedicated .net461 target.

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: How did you know it does a table level lock in 4.8? I can't seem to find this information in the official documentation.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
mafurman: Fixed!

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: Yikes, don't know how that happened. Thanks for catching this!

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: 👍

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: Good point! Fixed.

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: I removed this misleading comment from all the relevant methods.

src/Microsoft.IdentityModel.Tokens/TokenCacheOptions.cs
mafurman: Good catch. I've removed the unnecessary view param.

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: This is the default behavior of the ConcurrentDictionary class + is consistent with the current behavior of the InMemoryCryptoProviderCache.

I do agree that this is somewhat confusing behavior though; more than happy to change it if we agree this approach doesn't make sense.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
mafurman: Good catch! Fixed!

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
mafurman: I'm honestly not sure. I believe the only method in this try block that would throw an exception is [Remove](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycache.remove), but the documentation doesn't provide any information on this. :/

I've added a log for this scenario.

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: That's a valid point. I think my current approach only works if we operate under the assumption that the net45 target is going to be rarely used.

It's also worth taking into account that no one has actually run into any kind of overflow error with the InMemoryCryptoProviderCache, so the work that's currently being done here is more of a preventative measure. Even this simple approach is better than what we're currently doing.

A simple LRU implementation shouldn't be too difficult to implement, so I can modify the cache to do that instead if we believe the extra complexity is necessary.

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: Yes, good catch!

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: Good point! This comment was copy/pasted and definitely doesn't make much sense.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
mafurman: Good point, fixed. :)

src/Microsoft.IdentityModel.Tokens/TokenCacheOptions.cs
mafurman: Haha, that makes a lot more sense. Completely forgot to rename it.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
mafurman: I think so? The InMemoryCryptoProviderCache is only responsible for disposing the cache itself, not the SignatureProviders contained inside of it. I believe each SignatureProvider is responsible for disposing itself, so this behavior should remain consistent across all targets.

build/dependencies.props
mafurman: Oh, that's a good point. I've switched the dependency over to 5.0.0 and there doesn't seem to be any issues with it.

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
keegan-caruso: I **think** the idea is it is implied. That is the only way Clear can be done in a thread safe way. But I agree, it isn't in the docs and leaves you to hope.... or look at the reference source: https://github.com/microsoft/referencesource/blob/NetFramework48ZDP/mscorlib/system/collections/Concurrent/ConcurrentDictionary.cs#L615

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
keegan-caruso: MemoryCache's Set method will update. It is the InMemoryCryptoProvider cache that we do a set if not exists.

The layering feels off here, Set should Set. Concurrent dictionary supports that with an indexer.

Pls let me know if I'm missing something

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
brentschmaltz: @mafurman we can change the perf tests to use 'mafurman/cacheEviction' topic branch and compare to 'dev'. I will do that today.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        public int SizeLimit
+        {
+            get
+            {
+                return _sizeLimit;
+            }
+            set
+            {
+                _sizeLimit = (value > 0) ? value : throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(value)));
```
brentschmaltz: Is '1' a reasonable minimum.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -26,20 +26,60 @@
...
+        /// Creates a new instance of <see cref="InMemoryCryptoProviderCache"/> using the default <see cref="CryptoProviderCacheOptions"/>.
+        /// </summary>
+        public InMemoryCryptoProviderCache() : this(new CryptoProviderCacheOptions())
+        {
+        }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="InMemoryCryptoProviderCache"/> using the specified <paramref name="cryptoProviderCacheOptions"/>.
+        /// </summary>
+        /// <param name="cryptoProviderCacheOptions">The options which can be used to configure the crypto provider cache.</param>
```
brentschmaltz: suggest: 'The options used to configure &lt;see cref="InMemoryCryptoProviderCache" /&gt;'
don't see a reason to include "which can be used to".

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// Specifies the CryptoProviderCacheOptions which can be used to configure the internal cryptoprovider cache.
+    /// For the net461, net472, and netstandard2.0 targets we are using the Microsoft.Extensions.Caching.Memory.MemoryCache class:
+    /// https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycache
+    /// For the net45 target we are using our own simple caching implementation which clears the entire cache
+    /// once the cache is at max capacity. We recommend upgrading to net461+ for a more comprehensive caching experience.
+    /// Any property on these CryptoProviderCacheOptions that corresponds directly to a property 
+    /// used by Microsoft.Extensions.Caching.Memory.MemoryCache has the same name.
+    /// </summary>
+    public class CryptoProviderCacheOptions
+
```
brentschmaltz: extra line?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -118,22 +158,40 @@ public override bool TryAdd(SignatureProvider signatureProvider)
...
+                signatureProviderCache = _signingSignatureProviders;
+            else
+                signatureProviderCache = _verifyingSignatureProviders;
+
+#if NET461 || NET472 || NETSTANDARD2_0 
+            // The cache does NOT already have a crypto provider associated with this key.
+            if (!signatureProviderCache.TryGetValue(cacheKey, out _))
             {
-                if (_signingSignatureProviders.TryAdd(cacheKey, signatureProvider))
+                var cacheEntryOptions = new MemoryCacheEntryOptions
```
brentschmaltz: we could inline this 'new' in the call to .Set and remove the local variable.

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: Right, I agree that the MemoryCache's set method will update the cache.

I do think the [TryAdd()](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.tryadd?view=net-5.0) method from the ConcurrentDictionary prevents an already set value from being overridden though.

> `true` if the key/value pair was added to the ConcurrentDictionary<TKey,TValue> successfully; `false` if the key already exists.

I might be misunderstanding something as well.

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: Thanks @brentschmaltz! 👍

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
mafurman: That's a good point... I personally don't think so.

We should probably set the minimum to a reasonable value instead of allowing the user to do something that doesn't really make sense.

Maybe 10? I'm not sure how to determine this.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
mafurman: That's a really good suggestion; the previous comment was far more verbose than it needed to be.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
mafurman: Thanks, fixed!

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
mafurman: Good point; I've removed the local variable and inlined it.

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: Thank you for confirming that a table level lock occurs in 4.5 as well! I've gone ahead and removed the lock from this caching implementation. :)

src/Microsoft.IdentityModel.Tokens/MaxCapacityRefreshCache.cs
mafurman: Seems like I confused myself by focusing too much on the ConcurrentDictionary APIs, haha. I've gone ahead and modified this code to make use of the indexer and to override the current value if it already exists.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
mafurman: I added two tests that show SignatureProviders are disposed of properly across all targets. These are currently present as unit tests to make review easier.

They make use of GC.Collect() and GC.WaitForPendingFinalizers(), so both tests should definitely be moved to a different location later to avoid slowing down our tests.

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
```diff
@@ -26,6 +26,7 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.Net.Http.Headers;
```
keegan-caruso: unused statement

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
```diff
@@ -426,6 +427,73 @@ public static TheoryData<CryptoProviderCacheTheoryData> TryGetSignatureProviderT
...
 
+        /// <summary>
+        /// Checks that the cache and any crypto providers contained inside of it are properly garbage collected.
+        /// </summary>
+        [Fact]
+        public void CryptoProviderCacheDisposal()
+        {
+            TestUtilities.WriteHeader($"{this}.CryptoProviderCacheDisposal");
+            var context = new CompareContext();
+            WeakReference cacheReference = null;
```
keegan-caruso: This mostly works, but it would be better to test this from the perspective of how we expect it to be used.

e.g. If we make these regular references and test _disposed and DisposeCalled.

Debugging this, looks like Dispose is not called on  cryptoProviderReference

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
```diff
@@ -426,6 +427,73 @@ public static TheoryData<CryptoProviderCacheTheoryData> TryGetSignatureProviderT
...
+                var customCryptoProvider = new CustomSignatureProvider(new DerivedSecurityKey("kid", 256), ALG.HmacSha256);
+                cache.TryAdd(customCryptoProvider);
+                cryptoProviderReference = new WeakReference(customCryptoProvider, true);
+                cache.TryRemove(customCryptoProvider);
+            })();
+
+            // Both the cache and the crypto provider added to it should have gone out of scope about now, 
+            // so the garbage collector can clean them both up.
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
```
keegan-caruso: This one shouldn't be needed, neither InMemoryCryptoProviderCache or CustomSignatureProvider define a finalizer

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
mafurman: Fixed!

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
mafurman: That makes sense. I think the tests checking for garbage collection don't really have a place here.

I've gone ahead and simplified the tests to check that DisposeCalled is properly set on the InMemoryCryptoProviderCache.

It makes sense that Dispose() wouldn't be called on the cryptoProviderReference, since that's something that would need to be manually done by the user...

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
mafurman: Good point; I've removed it.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
brentschmaltz: Setting it to 1, essentially turns it off.
Let's check in with a minimum of 10. 
I will write some additional tests to test these thresholds.

brentschmaltz:
I finished with the comprehensive tests and I can confirm that there is an improvement when caching is turned on.
I have seen one instance of an increase of memory when RSACryptoServiceProvider is used for the RSA on .net core 2.1in the working set.
I will track that down.

brentschmaltz (approved):

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,260 @@
...
+
+            // just need to update value and move it to the top
+            if (_map.ContainsKey(key))
+            {
+                // make sure node hasn't been removed by a different thread
+                if (_map.TryGetValue(key, out var cacheItem))
+                    _eventQueue.Add(() => _doubleLinkedList.Remove(cacheItem));
+
+                // add a new item regardless of whether the old item was removed or not
+                _eventQueue.Add(() => _doubleLinkedList.AddFirst(new LRUCacheItem<TKey, TValue>(key, value, expirationTime)));
```
RojaEnnam: Should we batch Remove and AddFirst in set and get to keep these operations as one and not having to worry about what impact an another thread adding another event in between these two would cause?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,316 @@
...
+        /// 
+        /// </summary>
+        /// <param name="capacity"></param>
+        /// <param name="comparer"></param>
+        public EventBasedLRUCache(int capacity, IEqualityComparer<TKey> comparer = null)
+        {
+            _capacity = capacity > 0 ? capacity : throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(capacity)));
+            _map = new ConcurrentDictionary<TKey, LRUCacheItem<TKey, TValue>>(comparer ?? EqualityComparer<TKey>.Default);
+
+            var thread = new Thread(new ThreadStart(OnStart));
```
keegan-caruso: Why not use a task here?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,290 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    internal class EventBasedLRUCache<TKey, TValue> : ILRUCache<TKey,TValue>, IDisposable
+    {
+        private int _capacity;
+        private ConcurrentDictionary<TKey, LRUCacheItem<TKey, TValue>> _map;
```
keegan-caruso: _map and _doubleLinkedList  should be readonly

keegan-caruso: and capacity

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,290 @@
...
+    /// 
+    /// </summary>
+    /// <typeparam name="TKey"></typeparam>
+    /// <typeparam name="TValue"></typeparam>
+    internal class LRUCacheItem<TKey, TValue>
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        internal TKey Key { get; set; }
```
keegan-caruso: Key likely shouldn't be settable outside of a ctor

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,290 @@
...
+        internal DateTime ExpirationTime { get; set; }
+
+        internal LRUCacheItem(TKey key, TValue value)
+        {
+
+            Key = key ?? throw LogHelper.LogArgumentNullException(nameof(key));
+            Value = value ?? throw LogHelper.LogArgumentNullException(nameof(value));
+        }
+
+        internal LRUCacheItem(TKey key, TValue value, DateTime? expirationTime)
```
keegan-caruso: instead of accepting a nullable datetime and then checking if it is null, we can just accept a DateTime and assign it.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,290 @@
...
+            {
+                value = default;
+                return false;
+            }
+
+            _eventQueue.Add(() => _doubleLinkedList.Remove(cacheItem));
+            _map.TryRemove(key, out _);
+            value = cacheItem.Value;
+
+            return true;
```
keegan-caruso: should this be the return value of map.TryRemove?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
keegan-caruso: Similar for Set() above

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,290 @@
...
+        internal EventBasedLRUCache(int capacity, IEqualityComparer<TKey> comparer = null)
+        {
+            _capacity = capacity > 0 ? capacity : throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(capacity)));
+            _map = new ConcurrentDictionary<TKey, LRUCacheItem<TKey, TValue>>(comparer ?? EqualityComparer<TKey>.Default);
+
+            var thread = new Thread(new ThreadStart(OnStart));
+            thread.IsBackground = true;
+            thread.Start();
+
+            Task timerTask = RemoveExpiredValuesPeriodically(TimeSpan.FromMinutes(5));
```
keegan-caruso: we don't do anything with this. If that is intentional Should either do a discard assign `_ =` or remove the assignment completely

src/Microsoft.IdentityModel.Tokens/LockingLRUCache.cs
```diff
@@ -0,0 +1,283 @@
...
+
+        public bool Contains(TKey key)
+        {
+            if (key == null)
+                throw LogHelper.LogArgumentNullException(nameof(key));
+
+            return _map.ContainsKey(key);
+        }
+
+        public int RemoveExpiredValues()
```
keegan-caruso, line 76: Because this is public, you likely have a race condition if this called both from a user directly and near simultaneously this gets called from the periodic function

src/Microsoft.IdentityModel.Tokens/LockingLRUCache.cs
```diff
@@ -0,0 +1,283 @@
...
+        {
+            int numItemsRemoved = 0;
+            var node = _doubleLinkedList.First;
```
keegan-caruso, line 79: I believe this is a race condition. If this gets called externally and from the periodic function, you don't have a guarantee that first will be the same after a run of the loop below.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,290 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    internal class EventBasedLRUCache<TKey, TValue> : ILRUCache<TKey,TValue>, IDisposable
```
keegan-caruso, line 50: Let's document the behavior and the assumptions here

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        public bool Contains(TKey key)
+        {
+            if (key == null)
+                throw LogHelper.LogArgumentNullException(nameof(key));
+
+            return _map.ContainsKey(key);
+        }
+
+        public int RemoveExpiredValues()
```
keegan-caruso: similar, I believe this has a race condition because it is public. It can be called externally and concurrently by a schedule periodic cleanup

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -520,6 +521,17 @@ private SignatureProvider CreateSignatureProvider(SecurityKey key, string algori
             return signatureProvider;
         }
 
+        /// <summary>
+        /// For some security key types, in some runtimes, it's not possible to extract public key material and create an <see cref="SecurityKey.InternalId"/>.
+        /// In these cases, <see cref="SecurityKey.InternalId"/> will be an empty string, and these keys should not be cached.
+        /// </summary>
+        /// <param name="signatureProvider"><see cref="SignatureProvider"/> to be examined.</param>
+        /// <returns><c>True</c> if <paramref name="signatureProvider"/> should be cached, <c>false</c> otherwise.</returns>
+        internal static bool ShouldCacheSignatureProvider(SignatureProvider signatureProvider)
```
mafurman, line 532: I've moved this method out of the cache implementation after discussing an associated issue with George.

In the case where a SignatureProvider has an empty string as an internal ID, it should **never** be cached, regardless of the caching implementation.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: This makes sense to me! Fixed. :) 

I didn't group them initially since I figured the exact order didn't really matter (e.g. we're okay as long as we're close enough to LRU), but I do think it's better to be more deterministic about this if we can.

Commit: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/45c168bd46aa16f03a5e5f83c8b9b2b8cef9b070

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: As discussed offline, no reason not to. I've modified this to be a task with TaskCreationOptions.LongRunning set.

Commit: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/45c168bd46aa16f03a5e5f83c8b9b2b8cef9b070

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Good catch, thanks!

Addressed in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/45c168bd46aa16f03a5e5f83c8b9b2b8cef9b070.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Good point! Addressed in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/45c168bd46aa16f03a5e5f83c8b9b2b8cef9b070.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Makes sense! Doesn't really make sense to allow it to be nullable to begin with. Fixed in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/45c168bd46aa16f03a5e5f83c8b9b2b8cef9b070.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Yes, that makes sense. Fixed here https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/45c168bd46aa16f03a5e5f83c8b9b2b8cef9b070.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Going with discard assign as removing it completely results in the following error: 

`CS4014 - Because this call is not awaited, execution of the current method continues before the call is completed. Consider applying the 'await' operator to the result of the call.`

I realize we could add a suppression for this... which might be more truthful than going with the discard assign.

Changed here https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/45c168bd46aa16f03a5e5f83c8b9b2b8cef9b070.

src/Microsoft.IdentityModel.Tokens/LockingLRUCache.cs
mafurman: I do think that could be a race condition...

I've made the method internal for now, as it was only public so that we could have an interface to easily switch between the two caching models. I believe this removes the potential for a race condition, but if we decide we need to expose the method publicly this will need to be looked at again. https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/45c168bd46aa16f03a5e5f83c8b9b2b8cef9b070

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Great suggestion. I've added a short description with (what I believe to be) the main points. Please let me know if there's anything you think I'm missing! Change can be seen here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/45c168bd46aa16f03a5e5f83c8b9b2b8cef9b070

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: I've gone ahead and made it internal for now (same as the LockingLRUCache). https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/45c168bd46aa16f03a5e5f83c8b9b2b8cef9b070

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
mafurman: I changed the minimum size limit to 10 in this commit: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/c88f7b8cf56ea93006f1d4f6d5eeff3b204626e7.

I think the type of exception being thrown should also be changed to avoid confusing people.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,302 @@
...
+            while (true)
+            {
+                _eventQueue.Add(() => RemoveExpiredValues());
+                await Task.Delay(interval).ConfigureAwait(false);
+            }
+        }
+
+        public void SetValue(TKey key, TValue value)
+        {
+            SetValue(key, value, DateTime.MaxValue);
```
henrik-me, line 125: >DateTime.MaxValue [](start = 33, length = 17)

Should it be considered to use e.g. 24h by default?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,302 @@
...
+                    _eventQueue.Add(() =>
+                    {
+                        var lru = _doubleLinkedList.Last;
+                        _map.TryRemove(lru.Value.Key, out _);
+                        _doubleLinkedList.Remove(lru);
+                    });
+                }
+
+                // add a new node
+                var node = new LRUCacheItem<TKey, TValue>(key, value, expirationTime);
```
henrik-me: >var node = new LRUCacheItem<TKey, TValue>(key, value, expirationTime); [](start = 16, length = 70)

seems to be the same item you are also creating above, perhaps move this out of the if/else statement since you are using it regardless? 
Nit: This solution hangs on to the objects for longer and there is an additional CPU/memory cost of recreating the item in the scenario where it's not needed. Not sure how much of a perf/size impact thig really has, perf. tests also report on memory if I'm not mistaken, we should make sure to validate memory numbers as well as perf. numbers.

henrik-me:
Do you have numbers somewhere, I'm curious. Also curious about memory consumption with various number of items in the cache. 

---
In reply to: [764776400](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575#issuecomment-764776400) [](ancestors = 764776400)

test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
```diff
@@ -0,0 +1,242 @@
...
+
+            for (int i = 0; i <= 100; i++)
+            {
+                cache.SetValue(i, Guid.NewGuid().ToString());
+
+                // print the value of the list every 100 items
+                if (i % 10 == 0 && i != 0)
+                {
+                    // wait for the cache events to process
+                    Thread.Sleep(1000);
```
henrik-me: >Thread.Sleep(1000); [](start = 20, length = 19)

perhaps the cache should have an event for events processed? avoiding the needed for a thread.sleep? Or alternatively a WaitForProcessing on the LRUCache?

test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
```diff
@@ -0,0 +1,242 @@
...
+            for (int i = 0; i <= 10; i++)
+            {
+                // Only even values should expire.
+                if (i % 2 == 0)
+                    cache.SetValue(i, i.ToString(), DateTime.UtcNow + TimeSpan.FromSeconds(5));
+                else
+                    cache.SetValue(i, i.ToString());
+            }
+
+            Thread.Sleep(5000);
```
henrik-me, line 83: >Thread.Sleep(5000); [](start = 12, length = 19)

see other comment on thread.sleep

henrik-me:
Change looks good to me. Assuming we will keep only one cache implementation?
Would be interested in seeing the perf/memory numbers. Possibly update this PR with some numbers as well.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,302 @@
...
+
+            if (!_map.ContainsKey(key))
+            {
+                value = default;
+                return false;
+            }
+
+            LRUCacheItem<TKey, TValue> cacheItem;
+            // make sure node hasn't been removed by a different thread
+            if (_map.TryGetValue(key, out cacheItem))
```
keegan-caruso: cacheItem can be inline in the out statement, as is done above

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: The cache works the way it does now because this is the approach used by both the System.Runtime.Caching and Microsoft.Extensions.Caching.Memory libraries. Because of this, I was operating under the assumption that InMemoryCryptoProviderCache would be responsible for setting this value.

That being said, I can see the argument for this given our use case. I'm open to modifying it if we agree using a finite default value makes more sense.

test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
mafurman: I do think this is a good idea. I think the simplest way of doing this would be to add a bool flag (e.g. 'IsBusy') that tells the user whether the _eventQueue is empty or not, but this would require the user to figure out a way to wait until this condition is false. If we're only adding this for the purpose of testing I think this solution would be fine, but if we expect to expose it publicly having an event be fired or creating a WaitForProcessing() method seems more ideal.

I'm not too familiar with how I'd go about implementing something like this; are there any best practices that you're aware of?

mafurman:
@henrik-me I believe Brent should have numbers from the perf tests today. I can do additional tests with regards to memory usage if it turns out the perf tests don't cover this.

@keegan-caruso I can modify the code to use a TaskFactory; I was just under the impression that it was unnecessary for our use case.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Good point; I've fixed this.https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/38b55da469dcd67dfc96114d15fc2edf2a3802ba

I do believe the perf tests also report on memory, but I agree this is something we should look into if they don't. I believe Brent should have the results of the perf tests by tomorrow.

Do you mind elaborating on what you mean by "recreating the item in the scenario where it's not needed"? I'm not sure I follow.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Thanks, fixed! https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575/commits/38b55da469dcd67dfc96114d15fc2edf2a3802ba

test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
henrik-me: Thought: Instead of a timer. Move the "wait" to a method on the cache. The method can force push queue processing (internal method)

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,300 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading;
```
GeoK, line 32: nit: seems to be unnecessary.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,300 @@
...
+        internal LRUCacheItem(TKey key, TValue value)
+        {
+
+            Key = key ?? throw LogHelper.LogArgumentNullException(nameof(key));
+            Value = value ?? throw LogHelper.LogArgumentNullException(nameof(value));
+        }
+
+        internal LRUCacheItem(TKey key, TValue value, DateTime expirationTime)
+        {
+
```
GeoK: nit: extra new line

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// </summary>
+        internal TValue Value { get; set; }
+        /// <summary>
+        /// 
+        /// </summary>
+        internal DateTime ExpirationTime { get; set; }
+
+        internal LRUCacheItem(TKey key, TValue value)
+        {
+
```
GeoK: nit: extra new line

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,300 @@
...
+            _map = new ConcurrentDictionary<TKey, LRUCacheItem<TKey, TValue>>(comparer ?? EqualityComparer<TKey>.Default);
+            new Task(() => OnStart(), TaskCreationOptions.LongRunning).Start();
+            _ = RemoveExpiredValuesPeriodically(TimeSpan.FromMinutes(5));
+        }
+
+        private void OnStart()
+        {
+            while (true)
+            {
+                _eventQueue.Take().Invoke();
```
GeoK: I'm worried about disposal. This object might get disposed while the thread is still running and we might get an ObjectDisposedException that could be problematic.
Consider breaking this loop when _disposed == true and utilizing TryTake. One of the downsides here is that a thread would be constantly checking for work.

Similar approach could be considered for _RemoveExpiredValuesPeriodically_ too.

This would also allow us to gracefully exit from these threads when the cache is disposed.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,300 @@
...
+                throw LogHelper.LogArgumentNullException(nameof(key));
+
+            if (!_map.ContainsKey(key))
+            {
+                value = default;
+                return false;
+            }
+
+            // check to make sure node wasn't removed by a different thread
+            if (!_map.TryGetValue(key, out var cacheItem))
```
GeoK, line 215: Not sure why is this necessary considering the check on ln 199 and the actual removal on ln 214.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,332 @@
...
+            }
+            else
+            {
+                // if cache is full, then remove the least recently used node
+                if (_map.Count >= _capacity)
+                {
+                    _eventQueue.Add(() =>
+                    {
+                        var lru = _doubleLinkedList.Last;
+                        _map.TryRemove(lru.Value.Key, out _);
```
henrik-me, line 159: >_map.TryRemove(lru.Value.Key, out _); [](start = 24, length = 37)

Suggestion: Perhaps more than 1 should be removed? Perhaps 5% or 10% of the capacity?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,332 @@
...
+            value = cacheItem.Value;
+            _eventQueue.Add(() => _doubleLinkedList.Remove(cacheItem));
+            return _map.TryRemove(key, out _);
+        }
+
+        /// <summary>
+        /// FOR TESTING ONLY.
+        /// </summary>
+        /// <returns></returns>
+        public LinkedList<LRUCacheItem<TKey, TValue>> LinkedListValues => _doubleLinkedList;
```
henrik-me: >public [](start = 8, length = 6)

Should all these "FOR TESTING ONLY" be marked as internal ?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,332 @@
...
+        /// FOR TESTING ONLY.
+        /// </summary>
+        public long LinkedListCount => _doubleLinkedList.Count;
+
+        /// <summary>
+        /// FOR TESTING ONLY.
+        /// </summary>
+        public long MapCount => _map.Count;
+
+        public List<TKey> Dupes => _doubleLinkedList.GroupBy(x => x.Key).Where(g => g.Count() > 1).Select(y => y.Key).ToList();
```
henrik-me: >Dupes [](start = 26, length = 5)

testing?

test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
henrik-me: I see you have the method you can use instead now?

---
In reply to: [574028899](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575#discussion_r574028899) [](ancestors = 574028899)

henrik-me:
Would be great to include some numbers in the PR. 

---
In reply to: [777685435](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575#issuecomment-777685435) [](ancestors = 777685435)

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,332 @@
...
+                _eventQueue.Add(() =>
+                {
+                    _doubleLinkedList.Remove(cacheItem);
+                    _doubleLinkedList.AddFirst(cacheItem);
+                });
+            }
+            else
+            {
+                // if cache is full, then remove the least recently used node
+                if (_map.Count >= _capacity)
```
RojaEnnam, line 154: Should we change this to _doubleLinkedList.Count?

mafurman:
Closing this PR as the design has changed dramatically since it was first opened. Will create a new PR with the updated design.

PR #1579, opened 2021-01-19
Added RSACng support for encryption / decryption

brentschmaltz:
1. This PR enables 'lightup' for scenarios NET45, if RSACng is passed as the RSA type and is available from the runtime, then delegates are available for encryption. This is similar to the logic for Signing.
2. There was a bug where these statics would remember the first padding set, these have been changed to instance variables
        private static Type _rsaSignaturePaddingType
        private static volatile Func<RSA, byte[], string, byte[]> _rsaPkcs1SignMethod
        private static volatile Func<RSA, byte[], byte[], string, bool> _rsaPkcs1VerifyMethod
3. Code was refactored so it is easier to follow the logic for different targets
4. Added test versions in MultiThreadingTokenTests that use RSACng for signing and encrypting.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
@@ -39,6 +39,8 @@ using System.Security.Cryptography.X509Certificates;
 
 namespace Microsoft.IdentityModel.Tokens
 {
+    delegate byte[] EncryptDelegate(byte[] bytes);
+    delegate byte[] DecryptDelegate(byte[] bytes);
     delegate byte[] SignDelegate(byte[] bytes);
     delegate bool VerifyDelegate(byte[] bytes, byte[] signature);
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -53,36 +55,29 @@ internal class AsymmetricAdapter : IDisposable
...
         // We use reflection to detect that 4.6+ is available and access the appropriate signing or verifying methods.
         private static Type _hashAlgorithmNameType = typeof(object).Assembly.GetType("System.Security.Cryptography.HashAlgorithmName", false);
+        private static Type _rsaEncryptionPaddingType = typeof(object).Assembly.GetType("System.Security.Cryptography.RSAEncryptionPadding", false);
         private static Type _rsaSignaturePaddingType = typeof(object).Assembly.GetType("System.Security.Cryptography.RSASignaturePadding", false);
-        private static volatile Func<RSA, byte[], string, byte[]> _rsaPkcs1SignMethod;
-        private static volatile Func<RSA, byte[], byte[], string, bool> _rsaPkcs1VerifyMethod;
+
+        private volatile Func<RSA, byte[], byte[]> _rsaDecrypt45Method;
+        private volatile Func<RSA, byte[], byte[]> _rsaEncrypt45Method;
+        private volatile Func<RSA, byte[], string, byte[]> _rsaPkcs1SignMethod;
```
keegan-caruso: I see this is existing, but I'm not quite following why this needs to be volatile

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -311,86 +236,245 @@ private void InitializeUsingRsa(RSA rsa, string algorithm)
...
+                            typeof(byte[]));
+
+                PropertyInfo prop;
+                if (_useRSAOeapPadding)
+                    prop = _rsaEncryptionPaddingType.GetProperty("OaepSHA1", BindingFlags.Static | BindingFlags.Public);
+                else
+                    prop = _rsaEncryptionPaddingType.GetProperty("Pkcs1", BindingFlags.Static | BindingFlags.Public);
+
+                Delegate openDelegate = Delegate.CreateDelegate(delegateType, encryptMethod);
+                _rsaDecrypt45Method = (rsaArg, bytesArg) =>
```
keegan-caruso, line 415: This could be wrote to more than once (multiple threads calling Decrypt) Does that matter?

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
```diff
@@ -605,6 +609,23 @@ public static RsaSecurityKey RsaSecurityKey_2048
             }
         }
 
+        public static RsaSecurityKey RsaSecurityKeyCng_2048
+        {
+            get
+            {
+                if (_rsaCng == null)
```
keegan-caruso: This looks like it could be simplified with a Lazy<T>

GeoK (approved):

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
brentschmaltz: I don't think it is necessary since these are instance variables.
I will remove this, thanks.

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
brentschmaltz: I will make that change as i hardly ever use Lazy and perhaps this will make it top of mind.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
brentschmaltz: the AsymmetricSignatureProvider maintains a pool of AsymmetricAdapters and they function as a singleton.

PR #1581, opened 2021-01-20
Fix Actor Claim Processing Issue in Saml2SecurityTokenHandler

KevinTan-Engineer:
https://msazure.visualstudio.com/One/_workitems/edit/9170507

SetClaimsIdentityActorFromAttribute method is checking for "Actor" instead of "http://schemas.xmlsoap.org/ws/2009/09/identity/claims/actor". This results in SetClaimsIdentityActorFromAttribute(recursive method) returns right away without executing any of the logics


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -1027,7 +1027,7 @@ protected virtual void ValidateOneTimeUseCondition(Saml2SecurityToken securityTo
         protected virtual void SetClaimsIdentityActorFromAttribute(Saml2Attribute attribute, ClaimsIdentity identity, string issuer)
         {
             // bail here; nothing to add.
-            if (identity == null || attribute == null || attribute.Name != _actor || attribute.Values == null || attribute.Values.Count < 1)
+            if (identity == null || attribute == null || attribute.Name != ClaimTypes.Actor || attribute.Values == null || attribute.Values.Count < 1)
```
brentschmaltz: This could be breaking, perhaps we could think about accepting two different claim types.

brentschmaltz: Another alternative is have a property that allows the value of the expected actor to be set by the user.

KevinTan-Engineer: Hi Brent,

Thank you for quickly pointing out the issue. Should I be changing the return condition to something like this:

if (identity == null || attribute == null || (attribute.Name != ClaimTypes.Actor && attribute.Name != _actor) || attribute.Values == null || attribute.Values.Count < 1)      
    return;

I assume this is what you meant by accepting two different claim type as ‘actor’.

Thanks,
Kevin

brentschmaltz: That is an option OR we could think of a way allowing the caller to specify what the actor claim is.

KevinTan-Engineer: Hi Brent,

Thank you for the suggestion. Due to the time constraint I'm thinking about going with option of changing the if check to include two different actor claim. 

I've updated the pull request. Please let me know I can merge this change.

Thanks,
Kevin 

Thanks,
Kevin

brentschmaltz: @KevinTan-Engineer the other thing you can do to unblock yourself is override SetClaimsIdentityActorFromAttribute method and have it work as you wish.

corogers-msft: Hi Brent, we have done something similar to what you have suggested to be unblocked for our private preview customers. However, it is not a long-term solution since we would then be out of date with any changes in this method.

The issue I see is the change for allowing the user to specify what the actor claim is would maybe require a signature change/larger changes that could take time in rolling out, while this change that Kevin made to check for both actor claim strings is sufficient for our needs and does not break anything else as far as we can tell.

Is it possible to approve this change as is so we can release our production code or will there be pushback to have a more long term solution here? 

Thank you!

PR #1582, opened 2021-01-20
Check if Exp and Nbf values exist.

### LINKED ISSUE #1567: Invalid validation of non-present expiration JWT claims
The validators accepts nullable `notBefore` and `expires` `DateTime`s but return values from `JsonWebToken` `ValidFrom` and `ValidTo` properties return `DateTime.MinValue` if the claim is not present. So in my opinion `JsonWebTokenHandler` fails to properly convert it to nullable values.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/824068adaeb7b158df2b36d8787f689646e715e0/src/Microsoft.IdentityModel.Tokens/Validators.cs#L330

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/824068adaeb7b158df2b36d8787f689646e715e0/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L303-L313

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/824068adaeb7b158df2b36d8787f689646e715e0/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L995-L1000

As a result `SecurityTokenNoExpirationException` with `IDX10225` is never thrown.

The best way to handle this would be to change `JsonWebToken` results to nullable, but I guess it would be a big breaking change.

*Note: the above code snippets are from `6.8.0` branch but `dev` is at the same state (as of authoring this issue).*

RojaEnnam:
1. Instead of using JsonWebToken.Validto/ValidFrom in payload validation, adding logic to check for the presence of 'exp' and 'nbf' values in the token and returning the value accordingly. This means DateTime.MinValue will not be returned if these claims are absent.
2. Adding tests to verify if it throws when 'exp' claim is not present.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -994,8 +994,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
         private TokenValidationResult ValidateTokenPayload(JsonWebToken jsonWebToken, TokenValidationParameters validationParameters)
         {
-            var expires = jsonWebToken.ValidTo;
-            var notBefore = jsonWebToken.ValidFrom;
+            var expires = jsonWebToken.TryGetClaim(JwtRegisteredClaimNames.Exp, out var _) ? (DateTime?)jsonWebToken.ValidTo : null;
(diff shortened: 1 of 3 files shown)


brentschmaltz (approved):

keegan-caruso (approved):
LGTM

PR #1588, opened 2021-01-22
Set 'cty' claim in JWE header

### LINKED ISSUE #1295: CTY header is not set for JWE
I'm using the following code to issue my JWEs:

```csharp
var signCreds = new SigningCredentials(new SymmetricSecurityKey(Encoding.ASCII.GetBytes(Configuration["Jwt:SigningKey"])), SecurityAlgorithms.HmacSha256);
var encryptionCreds = new EncryptingCredentials(new SymmetricSecurityKey(Encoding.ASCII.GetBytes(Configuration["Jwt:Encryptionkey"])), SecurityAlgorithms.Aes128KW, SecurityAlgorithms.Aes128CbcHmacSha256);

var handler = new JwtSecurityTokenHandler();


var jwtSecurityToken = handler.CreateJwtSecurityToken(
  Configuration["Jwt:Issuer"],
  Configuration["Jwt:Audience"],
  new ClaimsIdentity(claims),
  DateTime.UtcNow,
  expiresIn,
  DateTime.UtcNow,
  signCreds,
  encryptionCreds);
```

But it doesn't specify "cty" header of the token - just only alg, enc and typ. If I understand correctly, the header must be set for encrypted JWT so I have an issue while parsing the token in golang because of the headers absence.

I also tried the following ways to issue JWE:

```csharp
var signCreds = new SigningCredentials(new SymmetricSecurityKey(Encoding.ASCII.GetBytes(Configuration["Jwt:SigningKey"])), SecurityAlgorithms.HmacSha256);
var encryptionCreds = new EncryptingCredentials(new SymmetricSecurityKey(Encoding.ASCII.GetBytes(Configuration["Jwt:Encryptionkey"])), SecurityAlgorithms.Aes128KW, SecurityAlgorithms.Aes128CbcHmacSha256);

var handler = new JwtSecurityTokenHandler();

var tokenDescriptor1 = new SecurityTokenDescriptor
  {
     Audience = "you",
     Issuer = "me",
     Subject = new ClaimsIdentity(claims),
     EncryptingCredentials = encryptionCreds
};

var tokenDescriptor2 = new SecurityTokenDescriptor
  {
     Audience = "you",
     Issuer = "me",
     Subject = new ClaimsIdentity(claims),
     EncryptingCredentials = encryptionCreds,
     SigningCredentials = signCreds
};

var tokenDescriptor3 = new SecurityTokenDescriptor
  {
     Audience = "you",
     Issuer = "me",
     Subject = new ClaimsIdentity(claims),
     EncryptingCredentials = encryptionCreds,
     SigningCredentials = signCreds,
     AdditionalHeaderClaims = new Dictionary<string, object> { { "cty", "JWT" } }
   };

var enc = handler.CreateEncodedJwt(tokenDescriptor1);
var encSigned = handler.CreateEncodedJwt(tokenDescriptor2);
var encSignedWithCty = handler.CreateEncodedJwt(tokenDescriptor3);
```

But have the same result:  

image

### LINKED ISSUE #1592: JsonWebTokenHandler.CreateToken(string payload, EncryptingCredentials encryptingCredentials) should not assume the 'payload' is json.
JsonWebTokenHandler.CreateToken calls CreateTokenPrivate passing a JObject(payload).
This assumes that payload is well formatted json. This does not need to be true.

CreateTokenPrivate  wants to create a JObject to apply logic for SetDefaultTimesOnTokenCreation.
Instead requiring well formatted json, figure out if payload is json, if so, then apply logic for SetDefaultTimesOnTokenCreation.

Otherwise use payload as is.

see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/c2a6ed5877030f4a7829b5ebbb7242d987c14be8/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L469

RojaEnnam:
Addresses #1592 and #1588 

Setting 'cty' claim with value 'JWT' in outer header for a JWE - based on [the spec](https://tools.ietf.org/html/rfc7519#section-5.2)
(Considering that requiring 'cty' in JWE header for validation is a breaking change, not making the changes to enforce it in the validation.)

JsonWebTokenHandler:

- Adding a new api CreateToken(Payload, EncryptingCredentials, AdditionalHeaderClaims)
- Check if the payload is json string, If AdditionalHeaderClaims has 'cty', use that value or else set 'cty' claim value to default value 'JWT'.
- If payload is not a json string, use it as is.

JwtSecurityTokenHandler:

- Always set 'cty==JWT' when creating a JWE.


src/Microsoft.IdentityModel.JsonWebTokens/JwtConstants.cs
```diff
@@ -32,6 +32,11 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// </summary>
     public static class JwtConstants
     {
+        /// <summary>
+        /// Token content type.
+        /// </summary>
+        public const string ContentType = "JWT";
```
brentschmaltz: I don't think we need this, we can use: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/53e19a887bfed8ee55a91f747494cf419c10f97f/src/Microsoft.IdentityModel.JsonWebTokens/JwtConstants.cs#L38

RojaEnnam: As it is type of the 'Content' and not type of the 'Header' I've created a new const( Following the same pattern of creating a new constant for TokenType as well 😃) But I see the argument of reusing the existing constant. I can go ahead change it if we think using HeaderType should be ok.

brentschmaltz: I think we should stick with HeaderType.

mafurman: Since the default value for both the 'typ' and 'cty' claim is always going to be the same, I think using the same constant is okay. I think modifying the comment to mention this information would be useful though.

It also looks like the 'TokenType' constant isn't actually used anywhere.

RojaEnnam: I went ahead removed ContentType constant.

GeoK:
Should we create the 'cty' claim by default in JsonWebTokens when creating a JWE?
It seems like we are doing that in Tokens.Json.

RojaEnnam:
> Should we create the 'cty' claim by default in JsonWebTokens when creating a JWE?
> It seems like we are doing that in Tokens.Json.

Sure, that's a valid point.

Currently we assume that the payload is a well formatted json but it can be anything, that is why we are letting the users set 'cty'. 
This [issue ](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1592) is created to change the CreateTokenPrivate method to expect payload to be anything and use it as is.

JwtSecurityTokenHandler does not use AdditionalHeaderClaims and as we are encouraging users to move away from it, introducing another API did not seem necessary.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -480,21 +507,37 @@ private string CreateTokenPrivate(JObject payload, SigningCredentials signingCre
                 header.Merge(JObject.FromObject(additionalHeaderClaims));
 
             var rawHeader = Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(header.ToString(Formatting.None)));
-
-            if (SetDefaultTimesOnTokenCreation)
+            try
```
RojaEnnam, line 607: More attention here is appreciated :)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -480,21 +507,37 @@ private string CreateTokenPrivate(JObject payload, SigningCredentials signingCre
...
+                        jsonPayload.Add(JwtRegisteredClaimNames.Nbf, now);
+                }
+                payload = jsonPayload.ToString(Formatting.None);
 
-                if (!payload.TryGetValue(JwtRegisteredClaimNames.Nbf, out _))
-                    payload.Add(JwtRegisteredClaimNames.Nbf, now);
+                if (encryptingCredentials != null)
+                {
+                    if (additionalHeaderClaims == null)
+                        additionalHeaderClaims = new Dictionary<string, object> { { JwtHeaderParameterNames.Cty, JwtConstants.HeaderType } };
```
brentschmaltz: if the parameter tokenType is not empty or null, should that be used?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -480,21 +507,37 @@ private string CreateTokenPrivate(JObject payload, SigningCredentials signingCre
...
 
             var rawHeader = Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(header.ToString(Formatting.None)));
-
-            if (SetDefaultTimesOnTokenCreation)
+            try
             {
-                var now = EpochTime.GetIntDate(DateTime.UtcNow);
-                if (!payload.TryGetValue(JwtRegisteredClaimNames.Exp, out _))
-                    payload.Add(JwtRegisteredClaimNames.Exp, now + TokenLifetimeInMinutes * 60);
+                var jsonPayload = JObject.Parse(payload);
```
brentschmaltz: If SetDefaultTimesOnTokenCreation is false, we don't have to parse the payload.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz, line 607: Above, CreateDefaultJWSHeader takes the 'tokenType'.
It's possible that a user wants 'typ' to be different in each header.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: If the 'cty' not sent in additionalclaims and tokentype is provided, I think we can use it for 'cty'

RojaEnnam: Discussed offline: tokenType is different from cty.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: Fixed it.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: Addressed it through AdditionalInnerHeaderClaims.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -433,7 +470,48 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
+        /// <exception cref="ArgumentNullException">if <paramref name="encryptingCredentials"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="compressionAlgorithm"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="additionalHeaderClaims"/> is null.</exception>
+        /// <exception cref="SecurityTokenException">if <see cref="JwtHeaderParameterNames.Alg"/>, <see cref="JwtHeaderParameterNames.Kid"/>
+        /// <see cref="JwtHeaderParameterNames.X5t"/>, <see cref="JwtHeaderParameterNames.Enc"/>, and/or <see cref="JwtHeaderParameterNames.Zip"/>
+        /// are present inside of <paramref name="additionalHeaderClaims"/>.</exception>
+        /// <returns>A JWE in compact serialization format.</returns>
+        public virtual string CreateToken(string payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, string compressionAlgorithm, IDictionary<string, object> additionalHeaderClaims, IDictionary<string, object> additionalInnerHeaderClaims)
```
brentschmaltz: Can you stack these parameters, it's a bit long
```cs
public virtual string CreateToken(
    string payload,
    SigningCredentials signingCredentials,
```

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -470,40 +548,59 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
+                            jsonPayload.Add(JwtRegisteredClaimNames.Nbf, now);
+                    }
+                }
             }
-
-            var rawPayload = Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(payload.ToString(Formatting.None)));
+            catch(Exception ex)
+            {
+                LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX14307, ex, payload)));
+            }
```
brentschmaltz, line 629: please add a new line after clauses

brentschmaltz (changes requested):
some small changes around formatting and stacking of parameters.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: Fixed

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: Fixed

brentschmaltz (approved):
thanks!

PR #1595, opened 2021-02-11
Adding the IDisposable interface to ECDsaSecurityKey, RsaSecurityKey and X509SecurityKey

### LINKED ISSUE #1434: Examine disposable behavior for classes that hold AsymmetricAlgorithm
At a minimum:

- RsaSecurityKey has a reference to Rsa of type RSA which extends AsymmetricAlgorithm.
- X509SecurityKey has a reference to AsymmetricAlgorithm

AsymmetricAdapter already implements IDisposable

mafurman:
Fixes #1434.

One of the points brought up in the issue mentioned above is that we should avoid caching and retrieving signature providers that have been disposed of. I've made a note to handle this in the caching PR (https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1575).

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
+++ b/src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs
@@ -34,8 +34,9 @@ namespace Microsoft.IdentityModel.Tokens
     /// <summary>
     /// Represents a ECDsa security key.
     /// </summary>
-    public class ECDsaSecurityKey : AsymmetricSecurityKey
+    public class ECDsaSecurityKey : AsymmetricSecurityKey, IDisposable
     {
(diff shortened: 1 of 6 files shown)


kevinchalet:
Are you sure it's the right solution to that problem? In most cases, components that create instances are responsible of disposing them. Yet, `RsaSecurity` and `ECDsaSecurityKey` offer constructors that accept `RSA` or `ECDsa` instances that already constructed and that they don't create. As such it would be weird for them to dispose them.

It's also worth noting that these types are often stored in-memory for quite some time (e.g in a typically ASP.NET Core JWT validation scenario, they are attached to the configuration object and basically treated as singletons).

brentschmaltz:
we decided not to do this.

PR #1604, opened 2021-03-02
Update NuGet.Config to use only the dotnet-public feed

GeoK:
* Remove SauceControl.InheritDoc and add missing comments
* Downgrade ApiCompat version
* Remove redundant NuGet.Config files

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
+++ b/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
@@ -189,13 +189,26 @@ namespace Microsoft.IdentityModel.Tokens
             return CreateKeyWrapProvider(key, algorithm, false);
         }
 
-#pragma warning disable 1573
-        /// <inheritdoc cref="CreateForSigning(SecurityKey, string)" />
+        /// <summary>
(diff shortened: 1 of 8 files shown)


src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -189,13 +189,26 @@ public virtual KeyWrapProvider CreateKeyWrapProvider(SecurityKey key, string alg
             return CreateKeyWrapProvider(key, algorithm, false);
         }
 
-#pragma warning disable 1573
-        /// <inheritdoc cref="CreateForSigning(SecurityKey, string)" />
+        /// <summary>
+        /// Creates an instance of <see cref="KeyWrapProvider"/> for a specific &lt;SecurityKey, Algorithm>.
```
keegan-caruso, line 193: Seems odd that there is both `&lt;` and ">" in the same line

GeoK: It's consistent with the rest of the library. '<' gets interpreted as an opening tag.

keegan-caruso (approved):

PR #1605, opened 2021-03-03
Added an EventBasedLRUCache for use by InMemoryCryptoProviderCache on the .NET desktop targets

### LINKED ISSUE #1447: Crypto provider cache entries should be automatically evicted

mafurman:
Fixes #1447.

Some notes on the changes that were made:
- ShouldCacheSignatureProvider has been moved to the CryptoProviderFactory and KeyVaultCryptoProvider from the InMemoryCryptoProviderCache. Moving this property a level up means that a SignatureProvider that has a key with an empty internal ID will never be cached regardless of the caching implementation being used. This is important if we ever want to allow users to set their own cache.
- A getter has been added for the CryptoProviderCache on the KeyVaultCryptoProvider so that proper usage of ShouldCacheSignatureProvider() could be tested.
- Certain properties such as EventQueueCount are currently marked as public so that they can be used by performance tests, but should be made internal before this PR is merged in.


src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultCryptoProvider.cs
```diff
@@ -48,6 +48,11 @@ public KeyVaultCryptoProvider()
             _cache = new InMemoryCryptoProviderCache();
         }
 
+        /// <summary>
+        /// Gets the <see cref="CryptoProviderCache"/>
+        /// </summary>
+        internal CryptoProviderCache CryptoProviderCache => _cache;
```
mafurman, line 54: Added a getter for testing purposes.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -520,6 +523,17 @@ private SignatureProvider CreateSignatureProvider(SecurityKey key, string algori
             return signatureProvider;
         }
 
+        /// <summary>
+        /// For some security key types, in some runtimes, it's not possible to extract public key material and create an <see cref="SecurityKey.InternalId"/>.
+        /// In these cases, <see cref="SecurityKey.InternalId"/> will be an empty string, and these keys should not be cached.
+        /// </summary>
+        /// <param name="signatureProvider"><see cref="SignatureProvider"/> to be examined.</param>
+        /// <returns><c>True</c> if <paramref name="signatureProvider"/> should be cached, <c>false</c> otherwise.</returns>
+        internal static bool ShouldCacheSignatureProvider(SignatureProvider signatureProvider)
```
mafurman, line 593: Moved one level up from the cache as it shouldn't be it's responsibility.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,317 @@
...
+
+            return numItemsRemoved;
+        }
+
+        internal void RemoveLRUs()
+        {
+            // use the _capacity for the newCacheSize calculation in the case where the cache is experiencing overflow
+            int currentCount = _map.Count <= _capacity ? _capacity : _map.Count;
+            var newCacheSize = currentCount - (int)(currentCount * _compactionPercentage);
+            while (_map.Count > newCacheSize && _doubleLinkedList.Count > 0)
```
mafurman, line 155: Since the _map can grow to be quite a bit larger than the linked list, we need to make sure we're not trying to remove more items than the list actually contains.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,317 @@
...
+        /// <summary>
+        /// FOR TESTING ONLY.
+        /// </summary>
+        /// <returns></returns>
+        internal ICollection<LRUCacheItem<TKey, TValue>> MapValues => _map.Values;
+
+        /// <summary>
+        /// FOR TESTING ONLY.
+        /// </summary>
+        public long EventQueueCount => _eventQueue.Count;
```
mafurman: Some of these properties need to remain public until the PR is ready to be merged in because of the performance tests.

test/Microsoft.IdentityModel.Tokens.Tests/CustomSecurityKeys.cs
```diff
@@ -47,13 +47,22 @@ public CustomRsaSecurityKey(int keySize, PrivateKeyStatus privateKeyStatus, RSAP
...
+
 #pragma warning disable CS0672
         public override bool HasPrivateKey => true;
 #pragma warning restore CS0672
 
         public override PrivateKeyStatus PrivateKeyStatus => _privateKeyStatus;
 
         public override int KeySize => _keySize;
+
+        internal override string InternalId  => "";
```
mafurman, line 65: This property has been added to the test key for testing that SignatureProviders corresponding to keys with empty internal IDs are not cached.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Specifies the CryptoProviderCacheOptions which can be used to configure the internal cryptoprovider cache.
+    /// For the net461, net472, and netstandard2.0 targets we are using the Microsoft.Extensions.Caching.Memory.MemoryCache class:
+    /// https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycache
+    /// For the net45 target we are using our own simple caching implementation which clears the entire cache
```
keegan-caruso: Comment is out of date

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+        private int _sizeLimit = DefaultSizeLimit;
+
+        /// <summary>
+        /// Default value for <see cref="SizeLimit"/>.
+        /// </summary>
+        public static readonly int DefaultSizeLimit = 1000;
+
+        /// <summary>
+        /// Gets or sets the size of the cache (in number of items). 
+        /// On the net45 target, the entire cache will be cleared once the cache reaches max capacity.
```
keegan-caruso: comment is out of date

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,317 @@
...
+    /// through eventual processing of the event queue. This implementation relies on the principle of 'eventual consistency':
+    /// though the map and it's corresponding linked list may be out of sync at any given point in time, they will eventually line up.
+    /// </summary>
+    /// <typeparam name="TKey">The key type to be used by the cache.</typeparam>
+    /// <typeparam name="TValue">The value type to be used by the cache</typeparam>
+    internal class EventBasedLRUCache<TKey, TValue> : IDisposable
+    {
+        private readonly int _capacity;
+        // The percentage of the cache to be removed when _maxCapacityPercentage is reached.
+        private double _compactionPercentage = .20;
```
keegan-caruso: _compactionPercentage and _maxCapacityPercentage can be readonly yes?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,317 @@
...
+            new Task(OnStart, TaskCreationOptions.LongRunning).Start();
+            _ = RemoveExpiredValuesPeriodically(TimeSpan.FromMinutes(5));
+        }
+
+        private void OnStart()
+        {
+            while (!_disposed)
+            {
+                if (_eventQueue.TryTake(out var action, 500))
+                    action.Invoke();
```
keegan-caruso: try / catch with logging around this?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
keegan-caruso: What could we do to allow the performance test to continue to be runnable?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,317 @@
...
+
+        public override bool Equals(object obj)
+        {
+            LRUCacheItem<TKey, TValue> item = obj as LRUCacheItem<TKey, TValue>;
+            return item == null ? false : Key.Equals(item.Key);
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
```
keegan-caruso: this will use objects GetHashCode, this is undesirable. Hashcode should be stable for the lifetime of an object, so probably a seed value combined with the HashCode of Key #Closed

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
```diff
@@ -591,6 +608,8 @@ public class CryptoProviderCacheTheoryData : TheoryDataBase
 
     public class InMemoryCryptoProviderCachePublic : InMemoryCryptoProviderCache
     {
+        public bool DisposeCalled { get; set; } = false;
```
keegan-caruso, line 670: this should also be in a #if block if it only gets used in a single target

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// </summary>
+        public int SizeLimit
+        {
+            get
+            {
+                return _sizeLimit;
+            }
+            set
+            {
+                _sizeLimit = (value > 10) ? value : throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(value)));
```
keegan-caruso: pls use the overload that takes the param name as well as a message. this isn't obvious

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -103,7 +103,9 @@ public CryptoProviderFactory(CryptoProviderFactory other)
         /// <summary>
         /// Gets the <see cref="CryptoProviderCache"/>
         /// </summary>
-        public CryptoProviderCache CryptoProviderCache { get; } = new InMemoryCryptoProviderCache();
+        /// Set was added for simplifying stress and perf tests.
```
keegan-caruso: remarks section? notes for review?

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
```diff
@@ -0,0 +1,53 @@
+﻿//------------------------------------------------------------------------------------------------
+// <copyright file="TokenValidator.cs" company="Microsoft">
+//     Copyright (c) Microsoft Corporation.  All rights reserved.
```
keegan-caruso: MIT license?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -26,20 +26,57 @@
...
+        {
+            if (cryptoProviderCacheOptions == null)
+                throw LogHelper.LogArgumentNullException(nameof(cryptoProviderCacheOptions));
+
+            _cryptoProviderCacheOptions = cryptoProviderCacheOptions;
+#if NETSTANDARD2_0
+            _signingSignatureProviders = new MemoryCache(Options.Create(new MemoryCacheOptions() { SizeLimit = _cryptoProviderCacheOptions.SizeLimit }));
+            _verifyingSignatureProviders = new MemoryCache(Options.Create(new MemoryCacheOptions() { SizeLimit = _cryptoProviderCacheOptions.SizeLimit }));
+#elif NET45 || NET461 || NET472
+            _signingSignatureProviders = new EventBasedLRUCache<string, SignatureProvider>(cryptoProviderCacheOptions.SizeLimit, StringComparer.Ordinal);
```
keegan-caruso: should we add a control that would allow reverting to current behavior? we could put it as a deprecated property from the start.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: @brentschmaltz Could the performance tests be modified to rely on internal rather than public properties?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
RojaEnnam: Why does map grow larger than the linked list?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
mafurman: Hmmm... my opinion on this is that it isn't necessary. Even if we put it as a deprecated property from the beginning, removing it would still be considered a breaking change.

If this caching work turns out to be an issue for people using Wilson directly, we can just tell them to revert back to version 6.8.0. If they're using SAL, CryptoProvider caching is turned off as is, so they can simply set that flag back to false in order to revert back to the original behavior.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
mafurman: Good catch. Fixed here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/commits/af3243229ca3f4885398299f0d733066d3b9be46.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
mafurman: Fixed here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/commits/af3243229ca3f4885398299f0d733066d3b9be46.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Yep! Fixed: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/commits/af3243229ca3f4885398299f0d733066d3b9be46.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Good idea. I added a try, catch {} block around the TryTake() and Action.Invoke() operations: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/commits/af3243229ca3f4885398299f0d733066d3b9be46. #Closed

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Thanks for pointing this out. I've addressed the issue here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/commits/af3243229ca3f4885398299f0d733066d3b9be46. #Closed

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
mafurman: InMemoryCryptoProviderCache now implements IDisposable on all targets, so actually it is the #if block around the Dispose() method that should be removed. Fix made here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/commits/af3243229ca3f4885398299f0d733066d3b9be46.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
mafurman: Agreed. Added here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/commits/af3243229ca3f4885398299f0d733066d3b9be46.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
mafurman: I believe Brent added this in. I'm interpreting this as notes for review; I don't think it's the kind of content that belongs in a remarks section....

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
mafurman: Whoops, fixed here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/commits/af3243229ca3f4885398299f0d733066d3b9be46.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,322 @@
...
+                try
+                {
+                    if (_eventQueue.TryTake(out var action, 500))
+                        action.Invoke();
+                }
+                catch (Exception ex)
+                {
+                    LogHelper.LogWarning(LogHelper.FormatInvariant(LogMessages.IDX10900, ex));
+                }
+
```
RojaEnnam: nit: extra line.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
keegan-caruso: set is good for testing too, I say we keep it and remove the comment, thoughts?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -184,24 +259,157 @@ public override bool TryRemove(SignatureProvider signatureProvider)
...
+                _disposed = true;
+                if (disposing)
+                {
+                    _signingSignatureProviders.Dispose();
+                    _verifyingSignatureProviders.Dispose();
+                }
+            }
+        }
+
+#region FOR TESTING TO BE REMOVED OR INTERNAL BEFORE RELEASE
```
keegan-caruso: how do we want to resolve these before merging in?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -103,7 +103,9 @@ public CryptoProviderFactory(CryptoProviderFactory other)
...
+        /// We may want to remove it.
```
henrik-me: >/// We may want to remove it. [](start = 8, length = 29)

want to make it internal then?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,329 @@
...
+        }
+
+        async Task RemoveExpiredValuesPeriodically(TimeSpan interval)
+        {
+            while (!_disposed)
+            {
+                _eventQueue.Add(() => RemoveExpiredValues());
+                await Task.Delay(interval).ConfigureAwait(false);
+            }
+        }
```
henrik-me, line 179: I'm wondering if this is causing a dispose of resources challenge if there are several LRUCaches created as the thread is waiting for 5 minutes. It can be considered to add the capability to the OnStart process (and do a simple comparison there on time instead of this). However if we haven't seen any issues in tests so far there is not likely a need to change this now. #Closed

henrik-me:
I believe unit test time has gone up significantly with this change or am I misreading previous test results?  #Closed

test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
```diff
@@ -0,0 +1,323 @@
...
+                for (int i = 0; i <= 10; i++)
+                {
+                    // Only even values should expire.
+                    if (i % 2 == 0)
+                        cache.SetValue(i, i.ToString(), DateTime.UtcNow + TimeSpan.FromSeconds(5));
+                    else
+                        cache.SetValue(i, i.ToString());
+                }
+
+                Thread.Sleep(5000);
```
henrik-me, line 110: >Thread.Sleep(5000); [](start = 16, length = 19)

Any reason this is still needed? Would WaitForProcessing not do the trick?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: @henrik-me i think internal makes good sense could be helpful for testing

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: The map is always modified at the time a get/set/remove method is called, while the list is always processed on a background thread. Though the list will eventually have the exact same values as the map, it's state at any given point in time is highly dependent on how fast that background task is processing. #Closed

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
mafurman: I think it makes sense to make these properties internal, since we'd likely want to be able to run cache performance tests even after the PR goes in.

@brentschmaltz Can the performance tests be modified to use internal rather than public properties?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Task.Delay() is non-blocking, so I don't think this should be causing an issue? The method itself is never added to the eventQueue, it's just responsible for adding a RemoveExpiredValues event every 5 minutes. I believe if we dispose while still in the delay interval, the task will return, see _disposed is true, and terminate.

I can see how adding this to the OnStart() method might be cleaner though.... #Closed

test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
mafurman: This Thread.Sleep() is specifically added so that values in the cache have time to expire.

mafurman:
> 
> 
> I believe unit test time has gone up significantly with this change or am I misreading previous test results?

@henrik-me We ran into an issue with the unit tests due to the task.Wait() call in the Dispose() method. Unit test time has been significantly reduced after modifying the delay interval in the TryTake() method of the OnStart() method from 500 ms to 50 ms. #Closed

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
mafurman: That makes sense to me. Removed here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/commits/9c7858b65c91a66f869a97950ee007dc282bd6fb

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Fixed: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/commits/9c7858b65c91a66f869a97950ee007dc282bd6fb

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
mafurman: Made the change here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/commits/9c7858b65c91a66f869a97950ee007dc282bd6fb

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -103,7 +103,7 @@ public CryptoProviderFactory(CryptoProviderFactory other)
         /// <summary>
         /// Gets the <see cref="CryptoProviderCache"/>
         /// </summary>
-        public CryptoProviderCache CryptoProviderCache { get; } = new InMemoryCryptoProviderCache();
+        internal CryptoProviderCache CryptoProviderCache { get; set; } = new InMemoryCryptoProviderCache();
```
GeoK: Looks like a breaking change. Let me know if ApiCompat is not catching this as we recently downgraded ApiCompat version.

mafurman: ApiCompat does catch this! I've fixed this so that only the set is internal.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,328 @@
...
+            return _map.ContainsKey(key);
+        }
+
+        /// <summary>
+        /// FOR TESTING PURPOSES ONLY.
+        /// </summary>
+        internal void WaitForProcessing()
+        {
+            while (_eventQueue.Count != 0)
+                continue;
```
henrik-me: Seems like this is what sonarcube complains about. Perhaps Task.Delay(1) instead? #Closed

test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
henrik-me: Hmm, perhaps a shorter expiry time then? can be 1ms or even 0ms, right? either it has an expiration or not? Not a fan of waiting 5s for any reason in a utest

---
In reply to: [590681840](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605#discussion_r590681840) [](ancestors = 590681840)

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
henrik-me: alternative 
while(true) 
  if (eventQueue.Count == 0) return;

---
In reply to: [591016250](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605#discussion_r591016250) [](ancestors = 591016250)

henrik-me (approved):
:shipit:

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,329 @@
...
+        private readonly double _maxCapacityPercentage = .95;
+        private bool _disposed = false;
+
+        internal EventBasedLRUCache(int capacity, TaskCreationOptions options = TaskCreationOptions.LongRunning, IEqualityComparer<TKey> comparer = null)
+        {
+            _capacity = capacity > 0 ? capacity : throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(capacity)));
+            _map = new ConcurrentDictionary<TKey, LRUCacheItem<TKey, TValue>>(comparer ?? EqualityComparer<TKey>.Default);
+            _eventQueueTask = new Task(OnStart, options);
+            _eventQueueTask.Start();
+            _ = RemoveExpiredValuesPeriodically(TimeSpan.FromMinutes(5));
```
brentschmaltz: For the scenario of SignatureProviders, it may not be necessary to schedule removal of expired items. 
We could introduce a variable to control removal and the interval in CryptoProviderCacheOptions. #Closed

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -163,9 +230,17 @@ public override bool TryGetSignatureProvider(SecurityKey securityKey, string alg
 
             var cacheKey = GetCacheKeyPrivate(securityKey, algorithm, typeofProvider);
             if (willCreateSignatures)
+#if NETSTANDARD2_0
                 return _signingSignatureProviders.TryGetValue(cacheKey, out signatureProvider);
```
GeoK, line 247: nit: can these be pulled out from ifdefs?

test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
keegan-caruso: +1 

pass expiry as UtcNow then waiting any amount should be fine right?

GeoK (approved):

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
RojaEnnam: Makes sense. Thanks! #Closed

RojaEnnam (approved):

test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
mafurman: If the _removeExpiredValues flag is set to true, the set API will not add expired values. This can be an issue when when just using Utc.Now, as the item could expire before it's added.

henrik-me: understood, but is it a real problem for the ut?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -84,6 +84,15 @@ static CryptoProviderFactory()
         /// </summary>
         public CryptoProviderFactory()
         {
+            CryptoProviderCache = new InMemoryCryptoProviderCache();
+        }
+
+        /// <summary>
```
keegan-caruso, line 91: should be updated

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -84,6 +84,15 @@ static CryptoProviderFactory()
...
+        /// Default constructor for <see cref="CryptoProviderFactory"/>.
+        /// </summary>
+        public CryptoProviderFactory(CryptoProviderCache cache)
```
keegan-caruso, line 97: copy ctor below should create new instance of cache

keegan-caruso, line 97: Should this be public? I think yes, it allows ctor injection as an extensibility point. other thoughts?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -104,6 +107,7 @@ public CryptoProviderFactory(CryptoProviderFactory other)
             if (other == null)
                 throw LogHelper.LogArgumentNullException(nameof(other));
 
+            CryptoProviderCache = new InMemoryCryptoProviderCache();
```
RojaEnnam: should we create if null?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,344 @@
...
+        private readonly int _tryTakeTimeout;
+        // if true, expired values will not be added to the cache and clean-up of expired values will occur on a 5 minute interval
+        private readonly bool _removeExpiredValues;
+
+        internal EventBasedLRUCache(
+            int capacity,
+            TaskCreationOptions options = TaskCreationOptions.LongRunning,
+            IEqualityComparer<TKey> comparer = null,
+            int tryTakeTimeout = 500,
+            bool removeExpiredValues = false,
```
henrik-me: >false [](start = 39, length = 5)

shouldn't this default to true? #Closed

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
keegan-caruso: This is consistent with the behavior that was prior. Wanted to keep the behavior consistent

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
henrik-me: ack. I think we should have a task to move this code to onstart

---
In reply to: [590681156](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605#discussion_r590681156) [](ancestors = 590681156)

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,344 @@
...
+                _map.TryRemove(lru.Value.Key, out _);
+                _doubleLinkedList.Remove(lru);
+            }
+        }
+
+        async Task RemoveExpiredValuesPeriodically(TimeSpan interval)
+        {
+            while (!_disposed)
+            {
+                _eventQueue.Add(() => RemoveExpiredValues());
```
henrik-me: >_eventQueue [](start = 16, length = 11)

there is a theoretical probability that this could be disposed... perhaps we should have try catch around it? #Closed

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,344 @@
...
+        internal void WaitForProcessing()
+        {
+            while (true)
+            {
+                if (_eventQueue.Count == 0)
+                    return;
+            }
+        }
+
+        internal int RemoveExpiredValues()
```
henrik-me, line 121: >RemoveExpiredValues [](start = 21, length = 19)

perhaps we should check for disposed and also have a try catch around this entire operation, during dispose this could fail? #Closed

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,344 @@
...
+
+            return _map.ContainsKey(key);
+        }
+
+        /// <summary>
+        /// FOR TESTING PURPOSES ONLY.
+        /// </summary>
+        internal void WaitForProcessing()
+        {
+            while (true)
```
henrik-me: >true [](start = 19, length = 4)

&& !_disposed ? #Closed

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,344 @@
...
+                    numItemsRemoved++;
+                }
+
+                node = nextNode;
+            }
+
+            return numItemsRemoved;
+        }
+
+        internal void RemoveLRUs()
```
henrik-me, line 150: >RemoveLRUs [](start = 22, length = 10)

this might need to check for isdisposed? (not sure about try/catch) #Closed

keegan-caruso: This will run in the context of in OnStart() which will be wrapped in a try catch #Closed

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
keegan-caruso: for the case of signature providers, no. Depends if we want to optimize for signature provider case or not

henrik-me: yeah, I thought the signature provider would pass in false her :) 

---
In reply to: [591992559](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605#discussion_r591992559) [](ancestors = 591992559)

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
keegan-caruso: #1609 #Closed

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
keegan-caruso: fixed! #Closed

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,358 @@
...
+        {
+            // use the _capacity for the newCacheSize calculation in the case where the cache is experiencing overflow
+            int currentCount = _map.Count <= _capacity ? _capacity : _map.Count;
+            var newCacheSize = currentCount - (int)(currentCount * _compactionPercentage);
+            while (_map.Count > newCacheSize && _doubleLinkedList.Count > 0)
+            {
+                var lru = _doubleLinkedList.Last;
```
brentschmaltz, line 157: Is it possible at this time that _doubleLinkedList.Last is null?
I guess it hinges if RemoveExpiredValues runs parallel to this task.
A race condition could have the last item removed between the check for _doubleLinkedList.Count > 0 and var lru = _doubleLinkedList.Last.
 #Closed

mafurman: RemoveLRUs and RemoveExpiredValues are both processed by the _eventQueue, so they should only ever run sequentially one after the other. #Closed

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,358 @@
...
+        {
+            try
+            {
+                while (!_disposed)
+                {
+                    _eventQueue.Add(() => RemoveExpiredValues());
+                    await Task.Delay(interval).ConfigureAwait(false);
+                }
+            }
+            catch (ObjectDisposedException ex)
```
brentschmaltz, line 175: why do we need to catch ObjectDisposedException ? #Closed

mafurman: I believe this is handling the case where the queue get disposed of while inside the while loop. #Closed

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,358 @@
...
+            try
+            {
+                var node = _doubleLinkedList.First;
+                while (node != null)
+                {
+                    var nextNode = node.Next;
+                    if (node.Value.ExpirationTime < DateTime.UtcNow)
+                    {
+                        _doubleLinkedList.Remove(node);
+                        _map.TryRemove(node.Value.Key, out _);
```
brentschmaltz: When a item is removed, when is it disposed?
In the case of SignatureProviders, CryptoProviderFactory.Release should be called.

keegan-caruso: Either by the GC or by the owner, we didn't create and we don't know if another reference exists elsewhere.

keegan-caruso: For SignatureProviders, TryRemove will zero out the cache reference, which CryptoProviderFactory.ReleaseSignatureProviderwill check if the ref is null null, when null, will dispose

henrik-me: hmm, it doesn't seem right that the cache should control this life time. Seems like a responsibility of the CryptoProviderFactory or ideally the GC imo. Why would anyone use this object still if we have removed it?

---
In reply to: [592695435](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605#discussion_r592695435) [](ancestors = 592695435)

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
henrik-me: +1: we discussed this yesterday as we were exploring possible reasons for the object disposed exception we were hunting.

---
In reply to: [592623019](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605#discussion_r592623019) [](ancestors = 592623019)

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -817,6 +829,144 @@ public static TheoryData<CryptoProviderFactoryTheoryData> ReleaseRsaKeyWrapProvi
...
+            Assert.Equal(2, signing._ReferenceCount);
+            Assert.Equal(2, signing2._ReferenceCount);
+
+            cryptoProviderFactory.ReleaseSignatureProvider(signing2);
+
+            Assert.Equal(1, signing._ReferenceCount);
+
+            cryptoProviderFactory.ReleaseSignatureProvider(signing);
+
+            Assert.Equal(0, signing._ReferenceCount);
```
henrik-me: >signing [](start = 28, length = 7)

Should this also be validated for being disposed?

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -817,6 +829,144 @@ public static TheoryData<CryptoProviderFactoryTheoryData> ReleaseRsaKeyWrapProvi
...
+            var signing = cryptoProviderFactory.CreateForSigning(Default.AsymmetricSigningKey, Default.AsymmetricSigningAlgorithm);
+            Assert.Equal(0, signing._ReferenceCount);
+
+            var signing2 = cryptoProviderFactory.CreateForSigning(Default.AsymmetricSigningKey, Default.AsymmetricSigningAlgorithm);
+            Assert.Equal(0, signing._ReferenceCount);
+            Assert.Equal(0, signing2._ReferenceCount);
+
+            cryptoProviderFactory.ReleaseSignatureProvider(signing2);
+
+            Assert.Equal(0, signing._ReferenceCount);
```
henrik-me: >signing [](start = 28, length = 7)

signing2 ? or perhaps assert both?

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -817,6 +829,144 @@ public static TheoryData<CryptoProviderFactoryTheoryData> ReleaseRsaKeyWrapProvi
...
+
+            cryptoProviderFactory.ReleaseSignatureProvider(signing);
+
+            Assert.Equal(0, signing._ReferenceCount);
```
henrik-me: >signing [](start = 28, length = 7)

should also be disposed? the same with signing2?

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -817,6 +829,144 @@ public static TheoryData<CryptoProviderFactoryTheoryData> ReleaseRsaKeyWrapProvi
...
+                    cryptoProviderFactory.ReleaseSignatureProvider(hmacSha384);
+
+                    cryptoProviderFactory.ReleaseSignatureProvider(rsaSha256Verifying);
+                    cryptoProviderFactory.ReleaseSignatureProvider(hmacSha256Verifying);
+                    cryptoProviderFactory.ReleaseSignatureProvider(hmacSha512Verifying);
+                    cryptoProviderFactory.ReleaseSignatureProvider(hmacSha384Verifying);
+                });
+            }
+
+            Task.WaitAll(tasks);
```
henrik-me, line 986: > Task.WaitAll(tasks); [](start = 11, length = 21)

should it be asserted that right after the wait the ref count should be 0 ?

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -817,6 +829,144 @@ public static TheoryData<CryptoProviderFactoryTheoryData> ReleaseRsaKeyWrapProvi
...
+            cryptoProviderFactory.ReleaseSignatureProvider(signing2);
+            var disposeCalled = (bool)signing2.GetType().GetField("_disposed", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(signing2);
+            Assert.False(disposeCalled);
+            cryptoProviderFactory.ReleaseSignatureProvider(signing);
+            disposeCalled = (bool)signing.GetType().GetField("_disposed", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(signing);
+            Assert.True(disposeCalled);
+        }
+
+        [Fact]
+        public void ReferenceCountingTest_MultiThreaded()
```
henrik-me, line 953: >public void ReferenceCountingTest_MultiThreaded() [](start = 8, length = 49)

Do you have a timing for this test? Would we be able to prove the same with fewer tasks?

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -841,6 +844,129 @@ public void ShouldCacheSignatureProvider()
...
+        {
+            var context = new CompareContext($"{this}.ReferenceCountingTest_NoCaching");
+            var cryptoProviderFactory = new CryptoProviderFactory(CryptoProviderCacheTests.CreateCacheForTesting());
+            cryptoProviderFactory.CacheSignatureProviders = false;
+
+            var signing = cryptoProviderFactory.CreateForSigning(Default.AsymmetricSigningKey, Default.AsymmetricSigningAlgorithm);
+            Assert.Equal(0, signing._ReferenceCount);
+
+            var signing2 = cryptoProviderFactory.CreateForSigning(Default.AsymmetricSigningKey, Default.AsymmetricSigningAlgorithm);
+            Assert.Equal(0, signing._ReferenceCount);
```
mafurman: Maybe use context.AddDiff() here rather than Assert?

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
keegan-caruso: locally about 1 second, on the build machine about 6-7 seconds. Could we use less? Sure.

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
keegan-caruso: using context is more consistent, fair point

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
keegan-caruso: Makes for an easier to understand test perhaps.

src/Microsoft.IdentityModel.Tokens/SignatureProvider.cs
```diff
@@ -38,7 +39,7 @@ public abstract class SignatureProvider : IDisposable
         /// <summary>
         /// Maintained by <see cref="CryptoProviderFactory"/>
         /// </summary>
-        internal int _ReferenceCount;
+        private int _referenceCount;
```
keegan-caruso, line 43: Good change. Ownership model is better #Closed

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -86,8 +88,15 @@ public InMemoryCryptoProviderCache(CryptoProviderCacheOptions cryptoProviderCach
...
-            _signingSignatureProviders = new EventBasedLRUCache<string, SignatureProvider>(cryptoProviderCacheOptions.SizeLimit, removeExpiredValues: false, comparer: StringComparer.Ordinal);
-            _verifyingSignatureProviders = new EventBasedLRUCache<string, SignatureProvider>(cryptoProviderCacheOptions.SizeLimit, removeExpiredValues: false, comparer: StringComparer.Ordinal);
+            _signingSignatureProviders = new EventBasedLRUCache<string, SignatureProvider>(cryptoProviderCacheOptions.SizeLimit, removeExpiredValues: false, comparer: StringComparer.Ordinal) { OnItemRemoved = OnSignatureProviderRemovedFromCache };
+            _verifyingSignatureProviders = new EventBasedLRUCache<string, SignatureProvider>(cryptoProviderCacheOptions.SizeLimit, removeExpiredValues: false, comparer: StringComparer.Ordinal) { OnItemRemoved = OnSignatureProviderRemovedFromCache };
+        }
+
+        private void OnSignatureProviderRemovedFromCache(SignatureProvider signatureProvider)
+        {
+            signatureProvider.CryptoProviderCache = null;
+            if (signatureProvider.RefCount == 0)
```
keegan-caruso, line 98: Any chance this can be under 0? would it be safer for this to be less than or equal?

src/Microsoft.IdentityModel.Tokens/SignatureProvider.cs
```diff
@@ -38,7 +39,7 @@ public abstract class SignatureProvider : IDisposable
         /// <summary>
         /// Maintained by <see cref="CryptoProviderFactory"/>
```
keegan-caruso: Comment is no longer accurate  #Closed

keegan-caruso (approved):

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -569,9 +569,7 @@ private SignatureProvider CreateSignatureProvider(SecurityKey key, string algori
                         signatureProvider = new SymmetricSignatureProvider(key, algorithm, willCreateSignatures);
 
                     if (ShouldCacheSignatureProvider(signatureProvider))
-                        if (CryptoProviderCache.TryAdd(signatureProvider))
-                            Interlocked.Increment(ref signatureProvider._ReferenceCount);
-
+                        CryptoProviderCache.TryAdd(signatureProvider);
```
mafurman, line 572: Why do we not need to increment the ref count here?

keegan-caruso: this is the creation case, See Signature Provider ctor, ref count is set to 1 on creation

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
brentschmaltz: See: SignatureProvider.Release()
There was a condition it never went below 0.

src/Microsoft.IdentityModel.Tokens/SignatureProvider.cs
brentschmaltz: addessed - https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/58e807326fca6a7252db6e8f952a4bf221321810 #Closed

PR #1615, opened 2021-03-19
Add SecurityTokenSignatureKeyNotFoundWithValidationErrorsException

### LINKED ISSUE #1614: Modify when SecurityTokenSignatureKeyNotFoundException is thrown
The library should have a way to distinguish when a token being validated that doesn't have a matching key is either expired or the issuer doesn't match

Current behavior:

Given a kid exists, 
    kid is matched but didn't validate: throw SecurityTokenInvalidSignatureException
    is not matched: throw SecurityTokenSignatureKeyNotFoundException
kid doesn't exists :
    throw SecurityTokenSignatureKeyNotFoundException

Proposed behavior:

Given a kid exists,
    kid is matched but didn't validate: throw SecurityTokenInvalidSignatureException
    token is expired or issuer is invalid: throw new exception recording validity of issuer and expiry
    token is not expired and issuer is valid: throw key not found exception
kid doesn't exists:
    throw SecurityTokenSignatureKeyNotFoundException

issuer and lifetime are validation are done through Validators.ValidateLifetime and Validators.ValidateIssuer
to provide extensibility and respect configuration

Similar behavior needs to be implemented for SAML1, SAML2, JwtSecurityTokenHandler, and JsonWebTokenHandler

This gives the layer above more signals as to when it can choose to refresh metadata.

Currently only SecurityTokenSignatureKeyNotFoundException is used for such a purpose

keegan-caruso:
SecurityTokenSignatureKeyNotFoundWithValidationErrorsException represents
SecurityTokenSignatureKeyNotFoundException but with validation errors
such that a different signal is sent to the layer above


test/Microsoft.IdentityModel.TestUtils/Default.cs
```diff
@@ -279,7 +279,7 @@ public static DateTime Expires
 
         public static string ExpiresString
         {
-            get => "2021-03-17T18:33:37.080Z";
+            get => "2025-03-17T18:33:37.080Z";
```
GeoK, line 282: Wow, what timing! This date was created 3y ago.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2687,6 +2687,96 @@ public static TheoryData<JWEDecompressionTheoryData> JWEDecompressionTheoryData(
...
+                    TokenDescriptor = new SecurityTokenDescriptor
+                    {
+                        Subject = new ClaimsIdentity(Default.PayloadClaims),
+                        SigningCredentials = Default.AsymmetricSigningCredentials,
+                    },
+                    ValidationParameters = new TokenValidationParameters
+                    {
+                        IssuerSigningKey = Default.SymmetricSigningKey,
+                    },
+                    ExpectedException = ExpectedException.SecurityTokenSignatureKeyNotFoundWithValidationErrorsException("IDX10516:")
```
GeoK: Issuer mismatch? Consider adding a testId for clarity.

test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlSecurityTokenHandlerTests.cs
```diff
@@ -1285,6 +1285,96 @@ public static TheoryData<CreateTokenTheoryData> CreateSamlTokenUsingTokenDescrip
...
+                    TokenDescriptor = new SecurityTokenDescriptor
+                    {
+                        Subject = new ClaimsIdentity(Default.SamlClaims),
+                        SigningCredentials = Default.AsymmetricSigningCredentials,
+                        Issuer = Default.Issuer,
+                    },
+                    ValidationParameters = new TokenValidationParameters
+                    {
+                        IssuerSigningKey = Default.SymmetricSigningKey,
+                        ValidateIssuer = false,
```
GeoK: Any reason not to set 'ValidIssuer'? (here and on other places)

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -1104,7 +1104,13 @@ private SamlSecurityToken ValidateSignature(SamlSecurityToken samlToken, string
                 if (keyMatched)
                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10514, keysAttempted, samlToken.Assertion.Signature.KeyInfo, exceptionStrings, samlToken)));
 
-                throw LogHelper.LogExceptionMessage(new SecurityTokenSignatureKeyNotFoundException(LogHelper.FormatInvariant(TokenLogMessages.IDX10513, samlToken.Assertion.Signature.KeyInfo, exceptionStrings, samlToken)));
+                InternalValidators.ValidateLifetimeAndIssuerAfterSignatureNotValidatedSaml(
+                    samlToken,
+                    samlToken.Assertion.Conditions.NotBefore,
```
GeoK: Check for null Conditions? (for both saml handlers)

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1017,10 +1017,19 @@ protected virtual JwtSecurityToken ValidateSignature(string token, TokenValidati
...
             {
-                if (kidMatched) 
+                if (kidMatched)
                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10511, keysAttempted, jwtToken.Header.Kid, exceptionStrings, jwtToken)));
 
-                throw LogHelper.LogExceptionMessage(new SecurityTokenSignatureKeyNotFoundException(LogHelper.FormatInvariant(TokenLogMessages.IDX10501, jwtToken.Header.Kid, exceptionStrings, jwtToken)));
+                DateTime? expires = (jwtToken.Payload.Exp == null) ? null : new DateTime?(jwtToken.ValidTo);
+                DateTime? notBefore = (jwtToken.Payload.Nbf == null) ? null : new DateTime?(jwtToken.ValidFrom);
+
+                InternalValidators.ValidateLifetimeAndIssuerAfterSignatureNotValidatedJwt(
```
GeoK, line 1032: ValidateSignature/ValidateToken are missing the description about the new exception type. That could be a good place to make a point about differences between SecurityTokenSignatureKeyNotFoundException and SecurityTokenSignatureKeyNotFoundWithValidationErrorsException.

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenSignatureKeyNotFoundWithValidationErrorsException.cs
```diff
@@ -0,0 +1,136 @@
...
+using System.Runtime.Serialization;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// This exception is thrown when a security token contained a key identifier but the key was not found by the runtime.
+    /// In addition, when validation errors exist over the security token.
+    /// </summary>
+    [Serializable]
+    public class SecurityTokenSignatureKeyNotFoundWithValidationErrorsException
```
GeoK: How can we better communicate the intent behind SecurityTokenSignatureKeyNotFoundException? Should we update the summary to hint towards metadata refresh?

GeoK:
Conceptually looks good.

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenSignatureKeyNotFoundWithValidationErrorsException.cs
brentschmaltz: How about: SecurityTokenUnableToValidate and then add a public external property that describes the error.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -1104,7 +1104,14 @@ private SamlSecurityToken ValidateSignature(SamlSecurityToken samlToken, string
                 if (keyMatched)
                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10514, keysAttempted, samlToken.Assertion.Signature.KeyInfo, exceptionStrings, samlToken)));
 
-                throw LogHelper.LogExceptionMessage(new SecurityTokenSignatureKeyNotFoundException(LogHelper.FormatInvariant(TokenLogMessages.IDX10513, samlToken.Assertion.Signature.KeyInfo, exceptionStrings, samlToken)));
+                if (samlToken.Assertion.Conditions != null && samlToken.Assertion.Signature.KeyInfo != null)
```
GeoK: samlToken.Assertion.Signature.KeyInfo should be already verified at this point.

Could we do something like:
samlToken.Assertion.Conditions?.NotBefore
samlToken.Assertion.Conditions?.NotOnOrAfter

and call into InternalValidators?

keegan-caruso: Will remove the check on Signature.KeyInfo.

Saml1 doesn't use nullable datetimes, so can only check the conditions object for null.

For saml2, I think we pass along null NotBefore and NotOnOrAfter and let that contribute to the decision on lifetime validity. Does that sound right?

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenUnableToValidateException.cs
```diff
@@ -0,0 +1,121 @@
...
+//------------------------------------------------------------------------------
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// This exception is thrown when a security token contained a key identifier but the key was not found by the runtime.
+    /// In addition, when validation errors exist over the security token. This exception is not intended to be used as a signal
```
mafurman: I'd suggest rephrasing this as "key was not found by the runtime AND when validation errors...." to make it clearer both conditions need to be met (assuming I'm understanding this correctly).

src/Microsoft.IdentityModel.Tokens/InternalValidators.cs
```diff
@@ -0,0 +1,133 @@
...
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Validators meant to be kept internal
+    /// </summary>
+    internal static class InternalValidators
+    {
+        /// <summary>
+        /// Called after signature validity has already been determined
+        /// </summary>
```
mafurman, line 42: Would it be more correct to say "called after signature validation has already failed"? We always throw here, right?

src/Microsoft.IdentityModel.Tokens/Exceptions/ValidationFailure.cs
```diff
@@ -0,0 +1,48 @@
...
+        None = 0b_0000_0000_0000_0000, // 0
+        /// <summary>
+        /// Indicates that the lifetime was invalid
+        /// </summary>
+        InvalidLifetime = 0b_0000_0000_0000_0001, // 1
+        /// <summary>
+        /// Indicates that the issuer was invalid
+        /// </summary>
+        InvalidIssuer = 0b_0000_0000_0000_0010, // 2
+    }
```
mafurman, line 47: I might be misunderstanding something.... do we need an enum that represents when both are invalid?

keegan-caruso: The idea of using a bitflag: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum#enumeration-types-as-bit-flags

That way we can represent a combination. And then the end user will do .HasFlag() to determine if a particular value is set.

If you think that is confusing, we can use distinct enums values

brentschmaltz: If we have a bitmap, then would we need a helper, cause we can't use ==.

brentschmaltz: We decided a bitmap is fine.

src/Microsoft.IdentityModel.Tokens/InternalValidators.cs
brentschmaltz: I agree with this, we call these when we couldn't validate the signature.

GeoK (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2687,6 +2703,107 @@ public static TheoryData<JWEDecompressionTheoryData> JWEDecompressionTheoryData(
...
+
+        [Theory, MemberData(nameof(SecurityKeyNotFoundExceptionTestTheoryData))]
+        public void SecurityKeyNotFoundExceptionTest(CreateTokenTheoryData theoryData)
+        {
+            var context = TestUtilities.WriteHeader($"{this}.SecurityKeyNotFoundExceptionTest", theoryData);
+
+            try
+            {
+                var handler = new JsonWebTokenHandler();
+
```
brentschmaltz: nit: do we need all these empty lines?

src/Microsoft.IdentityModel.Tokens/InternalValidators.cs
```diff
@@ -0,0 +1,144 @@
...
+                validIssuer = true;
+            }
+            catch (Exception)
+            {
+                // validIssuer will remain false
+            }
+
+            if (validLifetime && validIssuer)
+                throw LogHelper.LogExceptionMessage(new SecurityTokenSignatureKeyNotFoundException(LogHelper.FormatInvariant(TokenLogMessages.IDX10501, kid, exceptionStrings, securityToken)));
+            else
```
brentschmaltz, line 76: code could read slightly simplier.
if (validLifetime && validIssuer) ...
else if (validLifetime)
    throw InvalidIssuer ...
else if (validIssuer)
    throw InvalidLifetime ...

Similar in saml

keegan-caruso: Because of the bitfield enum I don't think we can reduce it that way. it would become roughly

if (validLifetime && validIssuer) => Key not found
if (validLifetime && !validIssuer) => UnableToValidate(invalid issuer)
if (!validLifetime && validIssuer) => UnableToValidate(invalid lifetime)
else => UnableToValdiate(invalid issuer & invalid lifetime)

I think

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
keegan-caruso: removed

src/Microsoft.IdentityModel.Tokens/InternalValidators.cs
keegan-caruso: I worked through this and it looked more complicated to me

PR #1617, opened 2021-03-22
Removing references to Microsoft.Extensions.Caching.Memory from the netstandard2.0 target

mafurman:
The results of the performance tests have shown that our EventBasedLRUCache only differs 1% or 2% from the Microsoft.Extensions.Caching.Memory.MemoryCache. Because of this, we've decided that we can just use the EventBasedLRUCache across all targets.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
+++ b/src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
@@ -43,6 +43,8 @@ namespace Microsoft.IdentityModel.Tokens
     /// the map is a <see cref="ConcurrentDictionary{TKey, TValue}"/> which may be modified directly inside an API call or
     /// through eventual processing of the event queue. This implementation relies on the principle of 'eventual consistency':
     /// though the map and it's corresponding linked list may be out of sync at any given point in time, they will eventually line up.
+    /// See here for more details:
+    /// https://aka.ms/identitymodel/caching
     /// </summary>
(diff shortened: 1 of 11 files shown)


src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -43,6 +43,8 @@ namespace Microsoft.IdentityModel.Tokens
     /// the map is a <see cref="ConcurrentDictionary{TKey, TValue}"/> which may be modified directly inside an API call or
     /// through eventual processing of the event queue. This implementation relies on the principle of 'eventual consistency':
     /// though the map and it's corresponding linked list may be out of sync at any given point in time, they will eventually line up.
+    /// See here for more details:
+    /// https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/Caching-in-Microsoft.IdentityModel
```
keegan-caruso: This is probably stable, but consider an aka.ms link

henrik-me (approved):
:shipit:

GeoK (approved):

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Good suggestion; added in link: https://aka.ms/identitymodel/caching

keegan-caruso (approved):

PR #1620, opened 2021-03-23
Improve speed of urlencoder by reducing string copies

brentschmaltz:
Our current encoding uses string operations that result in numerous string copies.
This change uses byte manipulation where possible to reduce to copying.


src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -66,41 +68,105 @@ public static string Encode(string arg)
...
-            s = s.Split(base64PadCharacter)[0]; // Remove any trailing padding
-            s = s.Replace(base64Character62, base64UrlCharacter62);  // 62nd char of encoding
-            s = s.Replace(base64Character63, _base64UrlCharacter63);  // 63rd char of encoding
+            return EncodeString(Convert.ToBase64String(inArray, 0, inArray.Length));
+        }
+
+        internal static string EncodeString(string str)
+        {
+#if NET45
+            str = str.Split(base64PadCharacter)[0]; // Remove any trailing padding
```
keegan-caruso, line 101: do we need to check that the length is greater than 0 in-addition to not null?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -66,41 +68,105 @@ public static string Encode(string arg)
...
+            str = str.Replace(base64Character62, base64UrlCharacter62);  // 62nd char of encoding
+            str = str.Replace(base64Character63, base64UrlCharacter63);  // 63rd char of encoding
```
keegan-caruso, line 103: would it make any difference if these used a string builder?

keegan-caruso (approved):
few small comments, otherwise LGTM

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
brentschmaltz: It may, but this is just the NET45 path, which has low priority.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
brentschmaltz: i think it's ok as an empty array or string, will return empty string.

PR #1631, opened 2021-04-18
Memory optimization in JwtSecurityTokenHandler.CanReadToken

lechu445:
string.Split allocates a result array and each segment. Additionally string.Split needs to scan input string twice: first to calculate size of result array, second to do actual split. So this change should also result with better CPU usage. In method CountSegments I use string.IndexOf because internally it can be hardware accelerated. So for longer strings with a few occurrences of character this solution has better performance characteristics than iterating by each character.


lechu445:
it's duplicate of https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1318

PR #1632, opened 2021-04-19
Add metadata telemetry support

RojaEnnam:
Send client SKU and Version data in metadata request.


src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.Concurrent;
+using System.Reflection;
+using System.Net.Http;
+
+namespace Microsoft.IdentityModel.Logging
```
RojaEnnam, line 34: Not sure if this the best place for this class to live.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
...
+        internal static readonly ConcurrentDictionary<string, string> telemetryData = new ConcurrentDictionary<string, string>();
+        internal const string skuTelemetry = "x-client-SKU";
+        internal const string versionTelemetry = "x-client-Ver";
+        internal const string osTelemetry = "x-client-OS";
+        internal static List<string> defaultTelemetryValues = new List<string> { skuTelemetry, versionTelemetry };
+
+        /// <summary>
+        /// Get the string that represents the client SKU.
+        /// </summary>
+        public static string ClientSku { get;} =
```
mafurman: nit: can replace with '{ get;} =' with "=>"

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
...
+        static IdentityModelTelemetryUtil()
+        {
+            telemetryData[skuTelemetry] = ClientSku;
+            telemetryData[versionTelemetry] = ClientVer;
+        }
+
+        /// <summary>
+        /// Adds a key and its value to the collection of telemetry data.
+        /// </summary>
+        /// <param name="key"> Represents the name of the telemetry.</param>
```
mafurman: nit: can change to just "The name of the telemetry."

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
...
+                throw LogHelper.LogArgumentNullException(nameof(key));
+
+            if (value == null)
+            {
+                if (telemetryData.ContainsKey(key))
+                    telemetryData.TryRemove(key, out _);
+
+            }
+            else
+                telemetryData[key] = value;
```
mafurman: nit: add "{}" for else statement for consistency

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
...
+        public static void AddTelemetryData(string key, string value)
+        {
+            if (string.IsNullOrEmpty(key))
+                throw LogHelper.LogArgumentNullException(nameof(key));
+
+            CheckIfDefaultTelemetry(key);
+            if (value == null)
+                RemoveTelemetryData(key);
+            else
+                telemetryData[key] = value;
```
mafurman: Is there a reason why we need to call Remove() and can't just override the value by using the setter (as in the 'else' case)?

src/Microsoft.IdentityModel.Logging/LogMessages.cs
```diff
@@ -40,6 +40,7 @@ internal static class LogMessages
         internal const string MIML10000 = "MIML10000: eventData.Payload is null or empty. Not logging any messages.";
         internal const string MIML10001 = "MIML10001: Cannot create the fileStream or StreamWriter to write logs. See inner exception.";
         internal const string MIML10002 = "MIML10002: Unknown log level: {0}.";
+        internal const string MIML10003 = "MIML10003: Can not add/remove/update default parameters: {0}. They are added by default.";
```
mafurman: Maybe rephrase this to say something along the lines of "they are added by default and cannot be modified in anyway"?

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -43,6 +43,22 @@ public class HttpDocumentRetriever : IDocumentRetriever
         private HttpClient _httpClient;
         private static readonly HttpClient _defaultHttpClient = new HttpClient();
 
+        /// <summary>
+        /// Gets or sets whether additional headers are added to a <see cref="HttpRequestMessage"/> headers. Set to true by default.
```
mafurman: Maybe say "additional default headers"?

test/Microsoft.IdentityModel.Logging.Tests/IdentityModelTelemetryUtilTests.cs
```diff
@@ -0,0 +1,193 @@
...
+        public void SetTelemetry(TelemetryTheoryData theoryData)
+        {
+            var testContext = new CompareContext();
+
+            try
+            {
+                IdentityModelTelemetryUtil.SetTelemetryData(theoryData.HttpRequestMessage);
+                // check if the resulting headers are as expected
+                if (!IdentityComparer.AreEqual(theoryData.ExpectedHeaders, theoryData.HttpRequestMessage?.Headers))
+                    throw new ArgumentException("resulting headers are not matching the expected headers.");
```
mafurman: nit: should be "do not match the"

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
+﻿using System;
```
keegan-caruso: copyright header

test/Microsoft.IdentityModel.Logging.Tests/IdentityModelTelemetryUtilTests.cs
```diff
@@ -0,0 +1,193 @@
+﻿using System;
```
keegan-caruso: copyright header

test/Microsoft.IdentityModel.Logging.Tests/IdentityModelTelemetryUtilTests.cs
```diff
@@ -0,0 +1,193 @@
...
+
+    /// <summary>
+    /// Helper extension of <see cref="HttpRequestHeaders"/> class.
+    /// </summary>
+    public static class HttpRequestHeadersExtension
+    {
+        /// <summary>
+        /// Clone <paramref name="headers"/> to a <see cref="IDictionary{String, strign}"/>.
+        /// </summary>
+        /// <param name="headers">Headers to clone to a <see cref="IDictionary{String, strign}"/></param>
```
keegan-caruso: typo, inconsistent casing, {string, string}

test/Microsoft.IdentityModel.Logging.Tests/IdentityModelTelemetryUtilTests.cs
```diff
@@ -0,0 +1,193 @@
...
+        public static IDictionary<string, string> CloneToDictionary(this HttpRequestHeaders headers)
+        {
+            if (headers == null)
+                return null;
+
+            var result = new Dictionary<string, string>();
+
+            foreach (var header in headers)
+            {
+                result.Add(header.Key, header.Value.First());
```
keegan-caruso: should we add all values in header.Value?

src/Microsoft.IdentityModel.Logging/LogMessages.cs
keegan-caruso: nit "Cannot"

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
...
+{
+    /// <summary>
+    /// Provides a way to add and remove telemetry data.
+    /// </summary>
+    public static class IdentityModelTelemetryUtil
+    {
+        internal static readonly ConcurrentDictionary<string, string> telemetryData = new ConcurrentDictionary<string, string>();
+        internal const string skuTelemetry = "x-client-SKU";
+        internal const string versionTelemetry = "x-client-Ver";
+        internal const string osTelemetry = "x-client-OS";
```
keegan-caruso: unused

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
...
+        internal static List<string> defaultTelemetryValues = new List<string> { skuTelemetry, versionTelemetry };
+
+        /// <summary>
+        /// Get the string that represents the client SKU.
+        /// </summary>
+        public static string ClientSku { get;} =
+#if NET45
+            "ID_NET45";
```
keegan-caruso, line 55: Can we remove the duplication between this and SkuTelemetryValue in OpenIdConnectMessage.cs?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// </summary>
+        /// <param name="key"> Represents the name of the telemetry.</param>
+        /// <param name="value"> Represents the value of the telemetry.</param>
+        public static void AddTelemetryData(string key, string value)
+        {
+            if (string.IsNullOrEmpty(key))
+                throw LogHelper.LogArgumentNullException(nameof(key));
+
+            CheckIfDefaultTelemetry(key);
+            if (value == null)
```
keegan-caruso: Why treat Add(key, null) as equivalent to Remove(key)?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
...
+            else
+                telemetryData[key] = value;
+        }
+
+        internal static void CheckIfDefaultTelemetry(string parameter)
+        {
+            if (defaultTelemetryValues.Contains(parameter))
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.MIML10003, defaultTelemetryValues)));
+
+                return;
```
keegan-caruso: shouldn't be indented

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Net.Http;
+
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// Provides a way to add and remove telemetry data.
+    /// </summary>
+    public static class IdentityModelTelemetryUtil
+    {
+        internal static readonly ConcurrentDictionary<string, string> telemetryData = new ConcurrentDictionary<string, string>();
```
keegan-caruso: Should we support multiple values for one header key?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
...
+        internal const string skuTelemetry = "x-client-SKU";
+        internal const string versionTelemetry = "x-client-Ver";
+        internal const string osTelemetry = "x-client-OS";
+        internal static List<string> defaultTelemetryValues = new List<string> { skuTelemetry, versionTelemetry };
```
keegan-caruso: this can be readonly

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,113 @@
...
+#endif
+
+        /// <summary>
+        /// Get the string that represents the client version.
+        /// </summary>
+        public static string ClientVer { get; } = typeof(IdentityModelTelemetryUtil).GetTypeInfo().Assembly.GetName().Version.ToString();
+
+        static IdentityModelTelemetryUtil()
+        {
+            telemetryData[skuTelemetry] = ClientSku;
```
keegan-caruso: can this go on the initializer of telemetryData and we can remove the static ctor?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
keegan-caruso: should be removed actually

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
RojaEnnam: Fixed

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
RojaEnnam: Updated

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
RojaEnnam: Good question. I think it might be useful, would like to hear what other's think (added it as an open question in the design doc as well)

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
RojaEnnam: ConcurrentDictionary does not have 'Add' method, so we wouldn't be able to use a collection initializer.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
RojaEnnam: That's from my assumption that we do not want to allow keys with null values. I am open to update if we think its ok to have them.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
RojaEnnam: That's from my assumption that we do not want to allow keys with null values. I am open to update if we think its ok to have them.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
keegan-caruso: I agree we don't want to accept null values. My question is more of why treat it as a remove vs throw a null argument exception on value as well as key?

RojaEnnam: I think that makes sense,  we should throw. One can use RemoveTelemetryData to remove.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
henrik-me: Aside: header can be shortened to: 

see e.g.: [https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/IAccount.cs](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/IAccount.cs)

A change for another PR / Work Item

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
```diff
@@ -198,7 +198,7 @@ private void EnsureTelemetryValues(OpenIdConnectMessage clonedMessage)
             if (this.EnableTelemetryParameters)
             {
                 clonedMessage.SetParameter(OpenIdConnectParameterNames.SkuTelemetry, SkuTelemetryValue);
-                clonedMessage.SetParameter(OpenIdConnectParameterNames.VersionTelemetry, typeof(OpenIdConnectMessage).GetTypeInfo().Assembly.GetName().Version.ToString());
+                clonedMessage.SetParameter(OpenIdConnectParameterNames.VersionTelemetry, IdentityModelTelemetryUtil.ClientVer);
```
keegan-caruso: sku should be fine to use between both, it is held constant. version, I'm not sure about, generally we expect them to be the same, but with binding redirects for example, is that always true?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// Removes a key and its value from the collection of telemetry data.
+        /// </summary>
+        /// <param name="key"> The name of the telemetry.</param>
+        public static void RemoveTelemetryData(string key)
+        {
+            if (string.IsNullOrEmpty(key))
+                throw LogHelper.LogArgumentNullException(nameof(key));
+
+            CheckIfDefaultTelemetry(key);
+            if (telemetryData.ContainsKey(key))
```
keegan-caruso: is this needed? Isn't TryRemove safe for if the key doesn't exist?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
GeoK: Perhaps it's better to move it under Tokens. It would eliminate the need for System.Http.Net dependency and some internalsVisibleTo in this project.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// Get the string that represents the client version.
+        /// </summary>
+        public static string ClientVer => typeof(IdentityModelTelemetryUtil).GetTypeInfo().Assembly.GetName().Version.ToString();
+
+        /// <summary>
+        /// Adds a key and its value to the collection of telemetry data.
+        /// </summary>
+        /// <param name="key"> The name of the telemetry.</param>
+        /// <param name="value"> The value of the telemetry.</param>
+        public static void AddTelemetryData(string key, string value)
```
GeoK: Perhaps we should return a bool here instead of throwing.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
brentschmaltz: +1

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
brentschmaltz: If we can add that without a future breaking change, we should add in a later release.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
brentschmaltz: Yes i agree, we don't want to throw if for some reason we can't add telemetry.
Please check the call graph to ensure we have try / catch everywhere, we should log any issues.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
RojaEnnam: M.IM.Tokens also does not have System.Http.Net dependency and M.IM.Protocols(which has HttpDocumentRetreiver) is not set as a friend assembly to it. Should it still logically live under M.IM Tokens?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
RojaEnnam: Makes sense. Fixed it. Add/Remove methods are not called in Wilson.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
RojaEnnam: Its safe. Removed the ContainsKey check.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs
RojaEnnam: Reverted the change.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
brentschmaltz: @RojaEnnam i think logging is the right assembly.
To solve the dependency on System.Net, SetTelemetry would need to be replaced with a method that returns the metadata as a IDictionary or some other structure. Then, each user would need to added the values to the HttpRequestMessage. That seems like a burden for devs.
@GeoK  I don't see anyone way to take System.Net.Http out of the picture.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,139 @@
...
+            foreach (var parameter in telemetryData)
+            {
+                // remove this header if it already exists.
+                // we don't want to add an additional value in case when a telemetry header already exists, but to overwrite it.
+                request.Headers.Remove(parameter.Key);
+                request.Headers.Add(parameter.Key, parameter.Value);
+            }
+        }
+
+        internal static void UpdateDefaultTelemetryData(string key, string value)
```
GeoK: Is this one still needed? Can you operate on telemetryData (also internal) in tests?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            telemetryData[key] = value;
+            return true;
+        }
+
+        /// <summary>
+        /// Removes a key and its value from the collection of telemetry data.
+        /// </summary>
+        /// <param name="key"> The name of the telemetry.</param>
+        public static void RemoveTelemetryData(string key)
```
GeoK: could also return a bool. Throwing seems aggressive for telemetry data, imo.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
RojaEnnam: @GeoK Are you saying we can directly add to the dictionary from SAL, instead of using this method?

GeoK: Ah, thought that this was a leftover used only for tests. Thanks for clarifying.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
RojaEnnam: Fixed it.

GeoK (approved):

PR #1633, opened 2021-04-19
Set telemetry data on HttpClient

RojaEnnam:
Set telemetry data on HttpClient

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
@@ -44,6 +44,16 @@ namespace Microsoft.IdentityModel.Protocols
         private static readonly HttpClient _defaultHttpClient = new HttpClient();
 
         /// <summary>
+        /// Gets or sets whether additional headers are added to a <see cref="HttpRequestMessage"/> headers.
+        /// </summary>
+        public static bool AllowAdditionalHeaderData { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets whether additional headers are added to a <see cref="HttpRequestMessage"/> headers for this <see cref="HttpDocumentRetriever"/> instance.
+        /// </summary>
+        public bool SendAdditionalHeaderData { get; set; } = true;
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="HttpDocumentRetriever"/> class.
         /// </summary>
         public HttpDocumentRetriever()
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -43,6 +43,16 @@ public class HttpDocumentRetriever : IDocumentRetriever
         private HttpClient _httpClient;
         private static readonly HttpClient _defaultHttpClient = new HttpClient();
 
+        /// <summary>
+        /// Gets or sets whether additional headers are added to a <see cref="HttpRequestMessage"/> headers.
+        /// </summary>
+        public static bool AllowAdditionalHeaderData { get; set; } = true;
```
Tratcher, line 49: ASP.NET users always ask for an easy opt-out of IM telemetry.
https://github.com/dotnet/aspnetcore/commit/32dd435c6e02267afddbfc715c7b92439f21d8cd
It looks like we'll need to update this line:
https://github.com/dotnet/aspnetcore/blob/fd75fcdbcd668a1dbf0a3da8f2ee8685850dab5d/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectPostConfigureOptions.cs#L99-L100

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -87,6 +97,9 @@ public async Task<string> GetDocumentAsync(string address, CancellationToken can
             {
                 LogHelper.LogVerbose(LogMessages.IDX20805, address);
                 var httpClient = _httpClient ?? _defaultHttpClient;
+                if (AllowAdditionalHeaderData && SendAdditionalHeaderData)
+                    IdentityModelTelemetryUtils.SetTelemetryData(httpClient);
```
Tratcher, line 101: Do not manipulate DefaultRequestHeaders, you don't own it. It's also incorrect to modify it per request.

Instead, change the below code to create a new HttpRequestMessage and add the headers there.

Tratcher (changes requested):

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
RojaEnnam: @Tratcher
 Makes sense. This was only a draft PR to demonstrate an idea. I will go ahead and abandon this one, Original PR is [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1632)

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
RojaEnnam: That looks right.

PR #1637, opened 2021-04-29
Fix double instantiation of Uri in IsHttps()

kosta-arnorsky:
Uri instance is created two times from the address.


lechu445:
@kosta-arnorsky please rebase the change against dev branch. This is what @brentschmaltz asked for in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1659#issuecomment-889456490

kosta-arnorsky:
Created https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1676

PR #1646, opened 2021-05-11
Fixing incorrect parameters when writing Saml2ProxyConditions

### LINKED ISSUE #1644: Incorrect parameters are passed into WriteElementString(...) in Saml2Serializer.WriteProxyRestriction(...)
We pass in the following parameters: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/d6f2b66d788195b50f2b1f700beb497851194c73/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs#L2057

However, the definition of the method used is as follows:
`public void WriteElementString(string localName, string ns, string value);`

`Saml2Constants.Elements.Audience` should be passed in as the first value, and the namespace `(Saml2Constants.Namespace`) should be passed in as the second value.

mafurman:
Fixes #1644.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
@@ -2054,7 +2054,7 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
 
             // <Audience> - 0-OO
             foreach (Uri uri in proxyRestriction.Audiences)
-                writer.WriteElementString(Prefix, Saml2Constants.Elements.Audience, uri.OriginalString);
+                writer.WriteElementString(Saml2Constants.Elements.Audience, Saml2Constants.Namespace, uri.OriginalString);
 
             writer.WriteEndElement();
         }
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs
```diff
@@ -2054,7 +2054,7 @@ protected virtual void WriteProxyRestriction(XmlWriter writer, Saml2ProxyRestric
 
             // <Audience> - 0-OO
             foreach (Uri uri in proxyRestriction.Audiences)
-                writer.WriteElementString(Prefix, Saml2Constants.Elements.Audience, uri.OriginalString);
+                writer.WriteElementString(Saml2Constants.Elements.Audience, Saml2Constants.Namespace, uri.OriginalString);
```
GeoK, line 2057: Other methods in this class are using this [overload ](https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmlwriter.writeelementstring?view=net-5.0#System_Xml_XmlWriter_WriteElementString_System_String_System_String_System_String_System_String_)of WriteElementString. Curios why you didn't choose that overload.

```suggestion
                writer.WriteElementString(Prefix, Saml2Constants.Elements.Audience, Saml2Constants.Namespace, uri.OriginalString);
```

brentschmaltz: @GeoK we could specify the ns, but it is not necessary as we are assuming the ns of the outer element as we use the same prefix. If one wanted to use a different ns, then it makes sense to specify it.

brentschmaltz (approved):

PR #1652, opened 2021-06-08
Change LRU logic to optional, off by default

brentschmaltz:
When obtaining a SignatureProvider from the cache, maintaining LRU order by removing and adding to front has shown a 10% performance reduction when a small number of SignatureProviders are in play.

For SignatureProviders, it is not important to maintain the LRU logic, just make sure the cache doesn't blow up.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
+++ b/src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
@@ -65,24 +65,34 @@ namespace Microsoft.IdentityModel.Tokens
         private readonly int _tryTakeTimeout;
         // if true, expired values will not be added to the cache and clean-up of expired values will occur on a 5 minute interval
         private readonly bool _removeExpiredValues;
+        // if true, then items will be maintained in a LRU fashion, moving to front of list when accessed in the cache.
+        private readonly bool _maintainLRU;
+
+        private TaskCreationOptions _options;
 
         internal EventBasedLRUCache(
             int capacity,
             TaskCreationOptions options = TaskCreationOptions.LongRunning,
             IEqualityComparer<TKey> comparer = null,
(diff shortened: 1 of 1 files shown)


src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -222,10 +238,14 @@ public bool SetValue(TKey key, TValue value, DateTime expirationTime)
                 var newCacheItem = new LRUCacheItem<TKey, TValue>(key, value, expirationTime);
                 _eventQueue.Add(() =>
                 {
+                    // if maintaining a LRU move item to front
                     // Add a remove operation in case two threads are trying to add the same value. Only the second remove will succeed in this case.
-                    _doubleLinkedList.Remove(newCacheItem);
+                    if (_maintainLRU)
```
keegan-caruso, line 243: I believe that this adds a race condition and we could end up with multiple items in the linked list representing one item in the dictionary. 

I think we need to always remove in this case

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -222,10 +238,14 @@ public bool SetValue(TKey key, TValue value, DateTime expirationTime)
...
+                        _doubleLinkedList.Remove(newCacheItem);
```
mafurman, line 244: @brentschmaltz As Keegan mentioned, we'd need to always remove in this case. This whole 'else' block will only be executed when the value **doesn't** already exist in the cache, so it isn't actually being moved to the front here.

The remove is here in case two threads adding the same value happen to enter this block of code at the same time. We need a remove in order to get rid of the duplicate in this sort of situation.

brentschmaltz:
Danny is picking up this work.

PR #1653, opened 2021-06-08
Updated CipherMode and added rules

SameerK-MSFT:
Updated the CipherMode
Added rules for security

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,18 @@
+<Project>
+  <PropertyGroup>
+
+    <LangVersion>7.2</LangVersion>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <IsPackable>false</IsPackable>
+    
+  </PropertyGroup>
+  <ItemGroup>
+    <!-- PrivateAssets="All" is on these packages to ensure they're development dependencies
+         and aren't included as transitionary dependencies to package consumers.
+         For clarity, without PrivateAssets marked here, anyone consuming Microsoft.Identity.Client
+         would also be forced to install these dependencies.  PrivateAssets avoids this problem. -->
+    <PackageReference Include="AsyncUsageAnalyzers" Version="1.0.0-alpha003" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.3" PrivateAssets="All" />
+  </ItemGroup>
+</Project> 
--- a/src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
@@ -180,7 +180,7 @@ namespace Microsoft.IdentityModel.Tokens
             {
                 // Create the AES provider
                 SymmetricAlgorithm symmetricAlgorithm = Aes.Create();
-                symmetricAlgorithm.Mode = CipherMode.ECB;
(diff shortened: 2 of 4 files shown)


Directory.Build.props
```diff
@@ -0,0 +1,18 @@
...
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <IsPackable>false</IsPackable>
+    
+  </PropertyGroup>
+  <ItemGroup>
+    <!-- PrivateAssets="All" is on these packages to ensure they're development dependencies
+         and aren't included as transitionary dependencies to package consumers.
+         For clarity, without PrivateAssets marked here, anyone consuming Microsoft.Identity.Client
+         would also be forced to install these dependencies.  PrivateAssets avoids this problem. -->
+    <PackageReference Include="AsyncUsageAnalyzers" Version="1.0.0-alpha003" PrivateAssets="All" />
```
kevinchalet, line 15: FYI, these analyzers were deprecated 3 years ago and replaced by the VS threading analyzers: https://github.com/Microsoft/vs-threading

Directory.Build.props
```diff
@@ -0,0 +1,18 @@
...
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.3" PrivateAssets="All" />
```
kevinchalet, line 16: If you use the 5.0 .NET SDK, this package is not necessary (nor recommended). By default, analysis is not enabled for .NET 5.0, but you can enable it for older TFMs by using `<EnableNETAnalyzers>true</EnableNETAnalyzers>`.

Here's an example: https://github.com/openiddict/openiddict-core/pull/1281/files

Directory.Build.props
GeoK: In addition, AsyncUsageAnalyzers is not present in the feed we are consuming (it breaks the build) while Microsoft.VisualStudio.Threading is.

src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
```diff
@@ -180,7 +180,7 @@ protected virtual SymmetricAlgorithm GetSymmetricAlgorithm(SecurityKey key, stri
             {
                 // Create the AES provider
                 SymmetricAlgorithm symmetricAlgorithm = Aes.Create();
-                symmetricAlgorithm.Mode = CipherMode.ECB;
+                symmetricAlgorithm.Mode = CipherMode.CBC;
```
brentschmaltz, line 183: I don't think these are backward compatible, so this would break any tokens that are in flight.
So we can't do this.
We will have to add switch to control what mode we are using.
While ECB is preferred over CBC, we will have to add a meaningful IV. SymmetricAlgorithm.CreateIV() should work.

kevinchalet: IMHO, you should consider adopting this breaking change for two reasons:
  - While using it with random content - such as wrapped keys - is not as problematic as with non-random plaintext, ECB is an unsafe AES mode: no one should ever use it.
  - It's not standard: the JWA/JWE specs only list CBC as an allowed AES mode: https://datatracker.ietf.org/doc/html/rfc7518#section-5.1 / https://datatracker.ietf.org/doc/html/rfc7516#appendix-A.3

brentschmaltz: @kevinchalet i agree with you, i am thinking hard about how not to break tokens in flight. This is a real issue.
I agree ECB is less secure than CBC, but for small number of bytes it is not a significant as larger data.

Excellent point about the spec compliance, thanks yet again.

brentschmaltz:
We are going to figure out have to use CBC, but not with these changes.

Issue #1650, opened 2021-06-03
CryptoProviderFactory.Default has a reference to a disposable object


imokaythanks:
When `CryptoProviderFactory` creates an `InMemoryCryptoProviderCache`
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b45a5b4cf4e918ca2b2c12ff1951f1be7b592bac/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs#L110
Two `EventBasedLRUCache` are created
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b45a5b4cf4e918ca2b2c12ff1951f1be7b592bac/src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs#L92-L93
Each of these fire off a `OnStart` task.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b45a5b4cf4e918ca2b2c12ff1951f1be7b592bac/src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs#L81

We have a requirement that unit tests cannot leave tasks hanging around, because if they throw an unobserved exception, it can cause an unrelated test to fail. So a test that touches `CryptoProviderFactory.Default` can fail this requirement if it finishes before the `OnStart` tasks.

A solution is to call
```
var cache = (InMemoryCryptoProviderCache)CryptoProviderFactory.Default.CryptoProviderCache;
cache.Dispose();
```
before exiting the test.

But this requires some internal knowledge of the `CryptoProviderCache` type.  Since `InMemoryCryptoProviderCache` implements `IDisposable`, and `CryptoProviderFactory` creates it, I think it should dispose it or implement `IDisposable` itself.

jmprieur:
@brentschmaltz @GeoK @mafurman : how do we triage this one? Which label should it have?

mafurman:
The CryptoProviderFactory currently doesn't implement IDisposable since the type of the cache property on it is the CryptoProviderCache abstract class, which doesn't implement IDisposable.

We should probably either add IDisposable to the the abstract class or cast the object to an instance of InMemoryCryptoProviderCache. Either way, the CryptoProviderFactory should implement IDisposable itself.

@jmprieur I've labeled this as 'customer reported' and 'bug' for now.

[...]

PR #1667, opened 2021-07-15
fix the leaking tasks issue that caused some users' tests to fail and…

dannybtsai:
After some discussion it was decided to stop the task when the cache is empty and start it when a new provider is added to the cache (non-empty -> empty & empty -> non-empty) as there is no reason to keep the task running if the cache is empty.

Tests that require no tasks hanging around at the end need to remove all the providers that they added to the cache.
The maintenance task that removes expired providers periodically has been replaced with a timer which is simpler than an async task and has less overhead. The maintenance task was previously implemented using Task.Delay() which also internally uses a timer. Plus a timer is much easier to pause/resume than a task.


src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -57,14 +57,33 @@ internal class EventBasedLRUCache<TKey, TValue> : IDisposable
...
+        private const int EventQueueTaskStopped = 0; // task not started yet
+        private const int EventQueueTaskRunning = 1; // task is running
+        private const int EventQueueTaskStopRequested = 2; // a request has been received to stop the task
+
+        private int _eventQueueTaskState = EventQueueTaskStopped;
+
+        // timer that removes expired items periodically
+        private Timer _timer = null;
+
+        //////// for testing purpose only to verify the task count
```
mafurman: Will this be kept when the PR is merged in? Wondering why it's commented out this way.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -76,17 +95,31 @@ internal class EventBasedLRUCache<TKey, TValue> : IDisposable
...
         {
-            while (!_disposed)
+            Interlocked.Increment(ref _taskCount);
+
+            if (_removeExpiredValues)
+            {
+                ResumeTimer();
+            }
+
+            // keep running until it is to be disposed, or when asked to stop (EventQueueTaskStopping)
```
mafurman: Should this comment be referencing "EventQueueTaskStopRequested"?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -76,17 +95,31 @@ internal class EventBasedLRUCache<TKey, TValue> : IDisposable
...
+            while (!_disposed && Interlocked.CompareExchange(ref _eventQueueTaskState, EventQueueTaskStopped, EventQueueTaskStopRequested) != EventQueueTaskStopRequested)
```
mafurman, line 126: Can you explain why this is being compared to EventQueueTaskStopRequested (rather than EventQueueTaskStopped)?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -279,6 +324,99 @@ public bool TryRemove(TKey key, out TValue value)
...
+            }
+        }
+
+        /// <summary>
+        /// The method handling the event when an item is added to the LinkedList.
+        /// The _eventQueueTask needs to be started if the _doubleLinkedList is empty.
+        /// </summary>
+        private void StartEventQueueTasksIfNotRunning()
+        {
+            // If the event queue task is set to be stopped but has not exited yet, abort the stopping process.
```
mafurman: abort the starting process?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -279,6 +324,99 @@ public bool TryRemove(TKey key, out TValue value)
...
+                eventQueueTask.Start();
+                eventQueueTask.GetAwaiter().OnCompleted(() => DisposeTask(eventQueueTask)); // dispose the task when it is complete
+
+                _eventQueueTask = eventQueueTask;
+            }
+        }
+
+        /// <summary>
+        /// Dispose of the specified task (if it is in a disposable state).
+        /// The Dipose() method should close the event object (WaitHandle) in the task but it may not affect the active task count.
```
mafurman: nit: should be "Di**s**pose"

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -313,7 +451,7 @@ internal ItemRemoved OnItemRemoved
         /// FOR TESTING ONLY.
         /// </summary>
         internal long EventQueueCount => _eventQueue.Count;
-#endregion
+        #endregion
```
mafurman: nit: extra whitespace?

src/Microsoft.IdentityModel.Tokens/GlobalSuppressions.cs
```diff
@@ -56,3 +56,4 @@
...
 [assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "Used as validation", Scope = "member", Target = "~M:Microsoft.IdentityModel.Tokens.InternalValidators.ValidateLifetimeAndIssuerAfterSignatureNotValidatedSaml(Microsoft.IdentityModel.Tokens.SecurityToken,System.Nullable{System.DateTime},System.Nullable{System.DateTime},System.String,Microsoft.IdentityModel.Tokens.TokenValidationParameters,System.Text.StringBuilder)")]
 [assembly: SuppressMessage("Globalization", "CA1305:Specify IFormatProvider", Justification = "Not using Globalization", Scope = "member", Target = "~M:Microsoft.IdentityModel.Tokens.Interop.Kernel32.GetMessage(System.Int32,System.IntPtr)~System.String")]
+[assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "<Pending>", Scope = "member", Target = "~M:Microsoft.IdentityModel.Tokens.EventBasedLRUCache`2.EventQueueTaskAction")]
```
mafurman: Missing a justification for this suppression.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -374,5 +519,21 @@ public override bool Equals(object obj)
 
         public override int GetHashCode() => 990326508 + EqualityComparer<TKey>.Default.GetHashCode(Key);
     }
+
+    internal class LRUCacheTaskItem
```
mafurman: Is this class used anywhere?

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -1024,6 +1024,180 @@ public void ReferenceCountingTest_MultiThreaded()
...
+            // create verifying providers
+            var verifyingProviders = CreateVerifyingProviders(factory);
+
+            // make sure providers can be retrieved from the cache
+            if (cache.TryGetSignatureProvider(Default.AsymmetricSigningKey, Default.AsymmetricSigningAlgorithm, typeof(AsymmetricSignatureProvider).ToString(), true, out var tmpProvider))
+            {
+                Assert.True(tmpProvider != null);
+            }
+
+            Thread.Sleep(3000); // give it some time to allow the tasks to start
```
mafurman: Is there anyway to be more deterministic about this?

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -1024,6 +1024,180 @@ public void ReferenceCountingTest_MultiThreaded()
...
+            WaitTillTasksComplete(cache, waitTimeoutInSeconds); // wait up to 60 seconds for the task(s) to complete
+
+            Assert.True(cache.TaskCount == 0, $"ProviderCacheTest_EnsureNoLeakingTasks: unexpected task count: {cache.TaskCount}, expected: 0");
+
+            //=============================================================================================
+            // repeat the steps and verify tasks will be restarted again and stopped when cache is empty...
+            //=============================================================================================
+
+            signingProviders = CreateSigningProviders(factory); // create signing providers
+
```
mafurman, line 1070: Do you need to check somewhere here that the tasks were started up again?

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -1024,6 +1024,180 @@ public void ReferenceCountingTest_MultiThreaded()
...
+            // remove all signing providers
+            foreach (var provider in signingProviders)
+                cache.TryRemove(provider);
+
+            foreach (var provider in verifyingProviders)
+                cache.TryRemove(provider);
+
+            WaitTillTasksComplete(cache, waitTimeoutInSeconds); // wait up to 60 seconds for the task(s) to complete
+
+            Assert.True(cache.TaskCount == 0, $"ProviderCacheTest_EnsureNoLeakingTasks: unexpected task count: {cache.TaskCount}, expected: 0");
```
GeoK, line 1063: Wondering whether this will be sufficient to fully address #1650.
Would it be required for users to manually cleanup the cache in each test?

dannybtsai: This will not address the issue in #1650. We need to document the new behaviors and users need to clean up the cache in order for the task to complete. I talked to Maria earlier and she suggested the github wiki as the place to add the doc. I will find it and create a task for this after my training.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
dannybtsai: My bad, it was initially named EventQueueTaskStopping but changed to EventQueueTaskStopRequested.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
dannybtsai: More comments have been added to explain the logic as the following:

            // Keep running until it is to be disposed, or when asked to stop (_eventQueueTaskState = EventQueueTaskStopRequested that happens in the OnLinkedListItemRemoved method).
            // If _eventQueueTaskState == EventQueueTaskStopRequested, the Interlocked.CompareExchange() will set the _eventQueueTaskState to EventQueueTaskStopped and the
            // while loop will exit (as the loop only continues if the original value of _eventQueueTaskState != EventQueueTaskStopRequested).
            // Interlocked.CompareExchange() is called to check/set the _eventQueueTaskState value to avoid potential race condition that one thread
            // is trying to start the task while another trying to stop it.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
dannybtsai: The comment was not clear and I have added more document the method as the following:

        /// The method handling the event when an item is added to the LinkedList.
        /// The _eventQueueTask needs to be started if the _doubleLinkedList is empty.
        /// 
        /// This method is called after an item is added to the queue, so it needs to start the event queue task if it is not running (_eventQueueTaskState == EventQueueTaskRunning),
        /// and prevent the task from being stopped if it has been set to stop but not exited yet (the task while loop is still running).

mafurman (approved):

PR #1672, opened 2021-07-27
Enable deterministic builds for CI builds

### LINKED ISSUE #1665: Enable deterministic builds for CI builds
Open the nuget package with NugetPackageExplorer
Under Health it can be observed that deterministic (dll/exe) is flagged for an error.

The solution is that the ContinuousIntegrationBuild MSBuild property has to be set to true when built on CI. In practice that means something like this in your directory props/targets file.

```
  <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
  </PropertyGroup>
```
[Reference](https://www.mitchelsellers.com/blog/article/net-5-deterministic-builds-source-linking)

See related issue: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2754

ciaozhang:
image

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -34,6 +34,10 @@
     <LangVersion>7.3</LangVersion>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
+    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
+  </PropertyGroup>
+  
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


TimHannMSFT:
This matches my understanding of the issue too. LGTM but I'll wait for someone more familiar with the codebase to provide sign off.

mafurman (approved):

RojaEnnam (approved):

PR #1674, opened 2021-07-28
Remove wait on first force refresh

RojaEnnam:
If RequestRefresh call is made for the first time, the next call to GetConfigurationAsync() will obtain new configuration, all the subsequent calls to RequestRefresh will have to wait for RefreshInterval(5mins) to pass.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -67,6 +67,7 @@ namespace Microsoft.IdentityModel.Protocols
         private DateTimeOffset _syncAfter = DateTimeOffset.MinValue;
         private DateTimeOffset _lastRefresh = DateTimeOffset.MinValue;
         private DateTimeOffset _lastForceRefresh = DateTimeOffset.MinValue;
+        private Random _jitterer = new Random();
 
         private readonly SemaphoreSlim _refreshLock;
         private readonly string _metadataAddress;
@@ -229,7 +230,8 @@ namespace Microsoft.IdentityModel.Protocols
         public void RequestRefresh()
         {
             DateTimeOffset now = DateTimeOffset.UtcNow;
-            if (now >= DateTimeUtil.Add(_lastForceRefresh.UtcDateTime, RefreshInterval))
+            var jitter = new TimeSpan((long)(RefreshInterval.Ticks * _jitterer.Next(0, 100) * 0.01));
+            if (now >= DateTimeUtil.Add(_lastForceRefresh.UtcDateTime, RefreshInterval.Add(jitter)))
             {
                 _syncAfter = now;
                 _lastForceRefresh = now;
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -66,6 +66,7 @@ namespace Microsoft.IdentityModel.Protocols
         private TimeSpan _refreshInterval = DefaultRefreshInterval;
         private DateTimeOffset _syncAfter = DateTimeOffset.MinValue;
         private DateTimeOffset _lastRefresh = DateTimeOffset.MinValue;
+        private DateTimeOffset _lastForceRefresh = DateTimeOffset.MinValue;
 
         private readonly SemaphoreSlim _refreshLock;
         private readonly string _metadataAddress;
@@ -222,15 +223,16 @@ namespace Microsoft.IdentityModel.Protocols
 
         /// <summary>
         /// Requests that then next call to <see cref="GetConfigurationAsync()"/> obtain new configuration.
-        /// <para>If the last refresh was greater than <see cref="RefreshInterval"/> then the next call to <see cref="GetConfigurationAsync()"/> will retrieve new configuration.</para>
+        /// <para>If the last force refresh was greater than <see cref="RefreshInterval"/> then the next call to <see cref="GetConfigurationAsync()"/> will retrieve new configuration.</para>
         /// <para>If <see cref="RefreshInterval"/> == <see cref="TimeSpan.MaxValue"/> then this method does nothing.</para>
         /// </summary>
         public void RequestRefresh()
         {
             DateTimeOffset now = DateTimeOffset.UtcNow;
-            if (now >= DateTimeUtil.Add(_lastRefresh.UtcDateTime, RefreshInterval))
+            if (now >= DateTimeUtil.Add(_lastForceRefresh.UtcDateTime, RefreshInterval))
             {
                 _syncAfter = now;
+                _lastForceRefresh = now;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -222,15 +223,16 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
         /// Requests that then next call to <see cref="GetConfigurationAsync()"/> obtain new configuration.
-        /// <para>If the last refresh was greater than <see cref="RefreshInterval"/> then the next call to <see cref="GetConfigurationAsync()"/> will retrieve new configuration.</para>
+        /// <para>If the last force refresh was greater than <see cref="RefreshInterval"/> then the next call to <see cref="GetConfigurationAsync()"/> will retrieve new configuration.</para>
         /// <para>If <see cref="RefreshInterval"/> == <see cref="TimeSpan.MaxValue"/> then this method does nothing.</para>
         /// </summary>
         public void RequestRefresh()
         {
             DateTimeOffset now = DateTimeOffset.UtcNow;
-            if (now >= DateTimeUtil.Add(_lastRefresh.UtcDateTime, RefreshInterval))
+            if (now >= DateTimeUtil.Add(_lastForceRefresh.UtcDateTime, RefreshInterval))
```
Tratcher: Why does it matter if the last refresh was forced or automatic?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -222,15 +224,17 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
         /// Requests that then next call to <see cref="GetConfigurationAsync()"/> obtain new configuration.
-        /// <para>If the last refresh was greater than <see cref="RefreshInterval"/> then the next call to <see cref="GetConfigurationAsync()"/> will retrieve new configuration.</para>
+        /// <para>If the last force refresh was greater than <see cref="RefreshInterval"/> then the next call to <see cref="GetConfigurationAsync()"/> will retrieve new configuration.</para>
         /// <para>If <see cref="RefreshInterval"/> == <see cref="TimeSpan.MaxValue"/> then this method does nothing.</para>
         /// </summary>
         public void RequestRefresh()
         {
             DateTimeOffset now = DateTimeOffset.UtcNow;
-            if (now >= DateTimeUtil.Add(_lastRefresh.UtcDateTime, RefreshInterval))
+            var jitter = new TimeSpan((long)(RefreshInterval.Ticks * _jitterer.Next(0, 100) * 0.01));
```
RojaEnnam: Should we define constants for min and max of the random values and should the max be less than 100? 

With 100, the RefreshInterval may be doubled.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -176,20 +176,37 @@ public void GetConfiguration()
...
+            // Next force refresh shouldn't pickup new config, as RefreshInterval hasn't passed
+            configManager.RequestRefresh();
+            TestUtilities.SetField(configManager, "_metadataAddress", "OpenIdConnectMetadata.json");
+            var configuration3 = configManager.GetConfigurationAsync().Result;
+            IdentityComparer.AreEqual(configuration2, configuration3, context);
+            if (!object.ReferenceEquals(configuration2, configuration3))
+                context.Diffs.Add("!object.ReferenceEquals(configuration2, configuration3) (5)");
+            // Next force refresh should pickup config since, RefreshInterval is set to 1s
+            configManager.RefreshInterval = TimeSpan.FromSeconds(1);
+            Thread.Sleep(2000);
```
RojaEnnam, line 198: refresh interval is set to 1ms here, adding sleep for 2ms because with the jitter max it can be doubled.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -66,6 +66,8 @@ public class ConfigurationManager<T> : IConfigurationManager<T> where T : class
         private TimeSpan _refreshInterval = DefaultRefreshInterval;
         private DateTimeOffset _syncAfter = DateTimeOffset.MinValue;
         private DateTimeOffset _lastRefresh = DateTimeOffset.MinValue;
```
Tratcher, line 69: _lastRefresh is no longer used.

RojaEnnam: updated the logic to use _lastRefresh

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
RojaEnnam: updated the logic to use _lastRefresh

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
brentschmaltz: The jitter should be applied to _automaticRefreshInterval  when setting with DefaultAutomaticRefreshInterval not necessarily the RequestRefesh interval need not be jittered.
The issue didn't specify, but the idea is in a server farm will not all refresh at the same time.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -66,6 +66,8 @@ public class ConfigurationManager<T> : IConfigurationManager<T> where T : class
...
+        private bool _isFirstForceRefresh = true;
```
brentschmaltz: I would suggest the name: _isFirstRefreshRequest

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
RojaEnnam: Got it, added change in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1674/commits/eeb6eb8daff04b3eb8aa5d2d1be0cc79ad483e1e

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -62,10 +62,12 @@ public class ConfigurationManager<T> : IConfigurationManager<T> where T : class
         /// </summary>
         public static readonly TimeSpan MinimumRefreshInterval = new TimeSpan(0, 0, 0, 1);
 
-        private TimeSpan _automaticRefreshInterval = DefaultAutomaticRefreshInterval;
+        private static TimeSpan _jitter = new TimeSpan(0,1,0,0);
```
RojaEnnam, line 65: Picked 1hr as the value from the goals/guidance for metadata doc in wiki.

brentschmaltz (approved):

PR #1676, opened 2021-08-01
Fix double instantiation of Uri in IsHttps()

kosta-arnorsky:
Uri instance is created two times from the address.


brentschmaltz (approved):

brentschmaltz:
@kosta-arnorsky thanks

PR #1678, opened 2021-08-05
Send additional data to metadata end point

RojaEnnam:
Send additional data to metadata end point

### DIFF
--- a/src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
+++ b/src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
@@ -134,12 +134,7 @@ namespace Microsoft.IdentityModel.Logging
             if (additionalHeaders != null)
             {
                 foreach (var parameter in additionalHeaders)
-                {
-                    // remove this header if it already exists.
-                    // we don't want to add an additional value in case when a telemetry header already exists, but to overwrite it.
-                    request.Headers.Remove(parameter.Key);
                     request.Headers.Add(parameter.Key, parameter.Value);
-                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -130,6 +130,17 @@ internal static void SetTelemetryData(HttpRequestMessage request)
...
                 request.Headers.Add(parameter.Key, parameter.Value);
             }
+
+            if (additionalHeaders != null)
+            {
+                foreach (var parameter in additionalHeaders)
+                {
+                    // remove this header if it already exists.
+                    // we don't want to add an additional value in case when a telemetry header already exists, but to overwrite it.
+                    request.Headers.Remove(parameter.Key);
```
brentschmaltz: Why would an existing telemetry header exist?

Can we shorten the comment to explain why we need to remove existing?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -130,6 +130,12 @@ internal static void SetTelemetryData(HttpRequestMessage request)
                 request.Headers.Remove(parameter.Key);
                 request.Headers.Add(parameter.Key, parameter.Value);
             }
+
+            if (additionalHeaders != null)
+            {
+                foreach (var parameter in additionalHeaders)
```
mafurman: nit: would it make more sense to call this variable 'header' rather than 'parameter'?

brentschmaltz: @RojaEnnam can you make this change and then we can approve.

RojaEnnam: Done.

brentschmaltz (approved):

brockallen:
What's the requirement or issue that motivated this "feature"? It's not linked here, so it's curious what this is for.

PR #1679, opened 2021-08-06
Adding BaseConfigurationManager and BaseConfiguration classes in M.IM.Tokens

mafurman:
These classes will be necessary in order to add a ValidateToken(...) API on the JsonWebTokenHandler that takes in a ConfigurationManager object in addition to the TokenValidationParameters. This API will be added in a future PR.
 
Some notes on the changes made:

- JsonWebTokenHandler is the class that will be responsible for setting UseLKG and UseCurrentConfiguration when going through its validation flow.
- I am not sure whether moving AutomaticRefreshInterval and other public static properties to the StandardConfigurationManager base abstract class is considered a breaking change or not. Would love to hear perspectives on this.
- Also not sure about the implications of having the RequestRefresh() API on both the StandardConfigurationManager abstract class and the IConfigurationManager.
- A property for timing how long we've been using the LKG may also need to be needed to the StandardConfigurationManager class (so that we know when it's stale and we should set UseLKG = false)
- Configuration had to be renamed to StandardConfiguration as it conflicted with System.Configuration in the .NET Framework (lots of extremely misleading Visual Studio errors on this one, haha).

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
@@ -39,7 +39,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
     /// Contains OpenIdConnect configuration that can be populated from a json string.
     /// </summary>
     [JsonObject]
-    public class OpenIdConnectConfiguration
+    public class OpenIdConnectConfiguration : BaseConfiguration
     {
         /// <summary>
         /// Deserializes the json string into an <see cref="OpenIdConnectConfiguration"/> object.
@@ -224,7 +224,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         /// Gets or sets the 'issuer'.
         /// </summary>
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore, NullValueHandling = NullValueHandling.Ignore, PropertyName = OpenIdProviderMetadataNames.Issuer, Required = Required.Default)]
-        public string Issuer { get; set; }
+        public override string Issuer { get; set; }
 
         /// <summary>
         /// Gets or sets the 'jwks_uri'
@@ -325,7 +325,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         /// Gets the <see cref="ICollection{SecurityKey}"/> that the IdentityProvider indicates are to be used signing tokens.
         /// </summary>
         [JsonIgnore]
-        public ICollection<SecurityKey> SigningKeys { get; } = new Collection<SecurityKey>();
+        public override ICollection<SecurityKey> SigningKeys { get; } = new Collection<SecurityKey>();
 
         /// <summary>
         /// Gets the collection of 'subject_types_supported'.
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -40,31 +40,8 @@ namespace Microsoft.IdentityModel.Protocols
     /// </summary>
     /// <typeparam name="T">The type of <see cref="IDocumentRetriever"/>.</typeparam>
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable")]
-    public class ConfigurationManager<T> : IConfigurationManager<T> where T : class
+    public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class
     {
-        /// <summary>
-        /// 12 hours is the default time interval that afterwards, <see cref="GetConfigurationAsync()"/> will obtain new configuration.
-        /// </summary>
-        public static readonly TimeSpan DefaultAutomaticRefreshInterval = new TimeSpan(0, 12, 0, 0);
-
-        /// <summary>
-        /// 5 minutes is the default time interval that must pass for <see cref="RequestRefresh"/> to obtain a new configuration.
-        /// </summary>
-        public static readonly TimeSpan DefaultRefreshInterval = new TimeSpan(0, 0, 5, 0);
-
-        /// <summary>
-        /// 5 minutes is the minimum value for automatic refresh. <see cref="AutomaticRefreshInterval"/> can not be set less than this value.
-        /// </summary>
-        public static readonly TimeSpan MinimumAutomaticRefreshInterval = new TimeSpan(0, 0, 5, 0);
-
-        /// <summary>
-        /// 1 second is the minimum time interval that must pass for <see cref="RequestRefresh"/> to obtain new configuration.
-        /// </summary>
-        public static readonly TimeSpan MinimumRefreshInterval = new TimeSpan(0, 0, 0, 1);
-
-        private static TimeSpan _jitter = new TimeSpan(0,1,0,0);
-        private TimeSpan _automaticRefreshInterval = DefaultAutomaticRefreshInterval.Add(TimeSpan.FromMinutes(new Random().Next((int)_jitter.TotalMinutes)));
-        private TimeSpan _refreshInterval = DefaultRefreshInterval;
         private DateTimeOffset _syncAfter = DateTimeOffset.MinValue;
         private DateTimeOffset _lastRefresh = DateTimeOffset.MinValue;
         private bool _isFirstRefreshRequest = true;
@@ -73,7 +50,6 @@ namespace Microsoft.IdentityModel.Protocols
         private readonly string _metadataAddress;
         private readonly IDocumentRetriever _docRetriever;
         private readonly IConfigurationRetriever<T> _configRetriever;
-        private T _currentConfiguration;
 
         /// <summary>
         /// Static initializer for a new object. Static initializers run before the first instance of the type is created.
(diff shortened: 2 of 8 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -222,12 +172,64 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
             }
         }
 
+        /// <summary>
+        /// Obtains an updated version of Configuration.
+        /// </summary>
+        /// <param name="cancel">CancellationToken</param>
+        /// <returns>Configuration of type StandardConfiguration.</returns>
+        /// <remarks>If the time since the last call is less than <see cref="StandardConfigurationManager.AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
+        public override async Task<StandardConfiguration> GetGeneralConfigurationAsync(CancellationToken cancel)
```
mafurman: Unfortunately the name of this method cannot be the same as GetConfigurationAsync(...), as having a different return type does not count as an overload.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -131,57 +110,79 @@ public ConfigurationManager(string metadataAddress, IConfigurationRetriever<T> c
...
+                        if (CurrentConfiguration == null) // Throw an exception if there's no configuration to return.
+                            throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20803, (_metadataAddress ?? "null")), ex));
+                        else
+                            LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20806, (_metadataAddress ?? "null")), ex));
+                    }
+                }
+
+                // Stale metadata is better than no metadata
+                if (CurrentConfiguration != null)
+                    return CurrentConfiguration as T;
```
mafurman: Probably safe to cast this as type T here, but not 100% sure.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -131,57 +110,79 @@ public ConfigurationManager(string metadataAddress, IConfigurationRetriever<T> c
...
-        /// <returns>Configuration of type T.</returns>
-        /// <remarks>If the time since the last call is less than <see cref="AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
-        public async Task<T> GetConfigurationAsync(CancellationToken cancel)
+        /// <returns>Configuration of type StandardConfiguration.</returns>
+        /// <remarks>If the time since the last call is less than <see cref="StandardConfigurationManager.AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
+        public override async Task<StandardConfiguration> GetGeneralConfigurationAsync(CancellationToken cancel)
         {
             DateTimeOffset now = DateTimeOffset.UtcNow;
-            if (_currentConfiguration != null && _syncAfter > now)
+            if (UseCurrentConfiguration && CurrentConfiguration != null && _syncAfter > now)
```
mafurman: We can check for whether or not the CurrentConfiguration is valid inside of this method, but we cannot set this value until a successful validation has occurred (JsonWebTokenHandler will be responsible for this).

src/Microsoft.IdentityModel.Protocols/LogMessages.cs
```diff
@@ -40,8 +40,8 @@ internal static class LogMessages
         internal const string IDX20000 = "IDX20000: The parameter '{0}' cannot be a 'null' or an empty object.";
 
         // properties, configuration 
-        internal const string IDX20106 = "IDX20106: When setting RefreshInterval, the value must be greater than MinimumRefreshInterval: '{0}'. value: '{1}'.";
-        internal const string IDX20107 = "IDX20107: When setting AutomaticRefreshInterval, the value must be greater than MinimumAutomaticRefreshInterval: '{0}'. value: '{1}'.";
+        // internal const string IDX20106 = "";
```
mafurman, line 43: Messages removed as these exceptions are now being thrown from the Microsoft.IdentityModel.Tokens namespace.

src/Microsoft.IdentityModel.Tokens/StandardConfiguration.cs
```diff
@@ -0,0 +1,51 @@
...
+//------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    ///  Represents a generic metadata configuration which is applicable for both XML and JSON based configurations.
+    /// </summary>
+    public abstract class StandardConfiguration
```
mafurman: Open to feedback on what the names of these new classes should be, as well as whether they should be interfaces or abstract classes.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -40,28 +40,8 @@ namespace Microsoft.IdentityModel.Protocols
     /// </summary>
     /// <typeparam name="T">The type of <see cref="IDocumentRetriever"/>.</typeparam>
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable")]
-    public class ConfigurationManager<T> : IConfigurationManager<T> where T : class
+    public class ConfigurationManager<T> : StandardConfigurationManager, IConfigurationManager<T> where T : StandardConfiguration
```
brentschmaltz: I think this will be breaking the 'where T : StandardConfiguration.

mafurman: Makes sense; I've changed this back to what it was originally.

src/Microsoft.IdentityModel.Tokens/StandardConfiguration.cs
TimHannMSFT: Not sure on what the future holds for this class. Right now seems like an interface makes more sense to my eye. Standard, to me implies some form of implementation (but even just any kind of class implies that which doesn't seem to be the case yet). If it's meant to eventually provide a standard/default implementation this seems okay naming to me, otherwise if this is just defining the contract then that's a better spot for interfaces.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
dannybtsai: According to the comments, the StandardConfiguration "Represents a generic metadata configuration which is applicable for both XML and JSON based configurations.".
1. Is there anything wrong if we name it "MetadataConfiguration" or "BaseConfiguration" or 'ConfigurationBase"? "Standard" seems vague to me.
2. I think abstract class is fine as I feel that more implementation can be added. For example, adding a property that returns the type (JSON, XML,...), or a method returning a SecurityKey from the SigningKeys collection.

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfiguration.cs
```diff
@@ -55,7 +55,7 @@ public string Issuer
         /// <summary>
         /// Gets the <see cref="IList{SecurityKey}"/> that the IdentityProvider indicates are to be used signing keys.
         /// </summary>
-        public ICollection<SecurityKey> SigningKeys
+        public override ICollection<SecurityKey> SigningKeys
         {
             get;
         } = new List<SecurityKey>();
```
dannybtsai: Instead of simply exposing the collection and let SigningKeys be added from outside, consider implementing Add(), Remove() to hide the implementation details and have more control on the logic. And it seems these can be done in the base class.
For example, in OpenIdConnectConfigurationRetriever

openIdConnectConfiguration.SigningKeys.Add(key);

will be changed to

openIdConnectConfiguration.AddSigningKey(key);

In a multi-threaded env, we can control adding/removing keys safely and decide if duplicates should be allowed...,etc.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
mafurman: I'm removing the UseCurrentConfiguration bool as there is no way to know with certainty if the validation failure is the result of an invalid token or a faulty configuration.

src/Microsoft.IdentityModel.Tokens/StandardConfiguration.cs
mafurman: Thanks for your feedback Tim! I don't believe this class will ever have any kind of default implementation (though there might be some cases I'm not thinking of); the reason why I chose to go with an abstract class is because it allows for changes to be made that aren't breaking our users. Any future changes to an interface will be considered a breaking change unfortunately. :( 

This is something to think about; I don't believe we've decided as a team which design choice is better for us going forward.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
mafurman: Thanks for your feedback Danny! :)

I think I'm going to go with BaseConfiguration as per our discussion with the rest of the team. I agree that it's much less vague!

Will need to think more about the abstract class vs. interface dilemma... I guess this really depends on the likelihood of these two classes having any kind of default implementations.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
dannybtsai: Since T is class (reference type) the "CurrentConfiguration as T" is safe. On the other hand, casting can throw exception if types do not match.

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfiguration.cs
```diff
@@ -46,7 +46,7 @@ public WsFederationConfiguration()
         /// <summary>
         /// Gets or sets the token issuer.
         /// </summary>
-        public string Issuer
+        public override string Issuer
```
dannybtsai: Any need to override Issuer? If not, this can be removed.

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfiguration.cs
```diff
@@ -55,7 +55,7 @@ public string Issuer
         /// <summary>
         /// Gets the <see cref="IList{SecurityKey}"/> that the IdentityProvider indicates are to be used signing keys.
         /// </summary>
-        public ICollection<SecurityKey> SigningKeys
+        public override ICollection<SecurityKey> SigningKeys
```
dannybtsai: Any need to override SigningKeys? If not, this can be removed.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
```diff
@@ -325,7 +325,7 @@ public OpenIdConnectConfiguration(string json)
         /// Gets the <see cref="ICollection{SecurityKey}"/> that the IdentityProvider indicates are to be used signing tokens.
         /// </summary>
         [JsonIgnore]
-        public ICollection<SecurityKey> SigningKeys { get; } = new Collection<SecurityKey>();
+        public override ICollection<SecurityKey> SigningKeys { get; } = new Collection<SecurityKey>();
```
dannybtsai, line 328: The type here is Collection which is in the System.Collections.ObjectModel namespace, while the property in the base class is List that is in the  System.Collections.Generic namespace. Both implement the ICollection interface but is this intended?

brentschmaltz (approved):

dannybtsai (changes requested):

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfiguration.cs
mafurman: Fixed.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
mafurman: Thanks for the catch! I've changed it to use Collection on the base class; this was a mistake on my part.

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfiguration.cs
mafurman: Removed, thanks!

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfiguration.cs
mafurman: I've noted this down and will address it in a future PR if we decide to go with this design choice. :)

PR #1687, opened 2021-08-16
Improve comments to clarify API usage and avoid unintentional validation weakening.

TimHannMSFT:
* Updating some headers for clarity
* Marking some setters on TokenValidationParameters as obsolete

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestUtilities.cs
+++ b/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestUtilities.cs
@@ -117,9 +117,9 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
             {
                 TokenDecryptionKeys = decryptionKeys,
                 RequireSignedTokens = false,
-                ValidateIssuer = false,
-                ValidateAudience = false,
-                ValidateLifetime = false,
+                IssuerValidator = (issuer, token, tvp) => { return issuer; },
+                AudienceValidator = (audiences, token, tvp) => { return true; },
+                LifetimeValidator = (notBefore, expires, token, tvp) => { return true; },
                 ValidateIssuerSigningKey = false,
             };
 
--- a/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
@@ -47,6 +47,11 @@ namespace Microsoft.IdentityModel.Tokens
         public const string IDX10104 = "IDX10104: TokenLifetimeInMinutes must be greater than zero. value: '{0}'";
         public const string IDX10105 = "IDX10105: ClaimValue that is a collection of collections is not supported. Such ClaimValue is found for ClaimType : '{0}'";
 
+        public const string IDX10106 = "IDX10106: Using this property to alter the default value and disable expiry requirements is not recommended. https://aka.ms/wilson/tokenvalidation";
+        public const string IDX10107 = "IDX10107: Disabling issuer validation is not recommended, use IssuerValidator to set a custom validation delegate. https://aka.ms/wilson/tokenvalidation";
+        public const string IDX10108 = "IDX10108: Disabling lifetime validation is not recommended, use LifetimeValidator to set a custom validation delegate. https://aka.ms/wilson/tokenvalidation";
+        public const string IDX10109 = "IDX10109: Disabling audience validation is not recommended, use AudienceValidator to set a custom validation delegate. https://aka.ms/wilson/tokenvalidation";
+        
         // token validation
         public const string IDX10204 = "IDX10204: Unable to validate issuer. validationParameters.ValidIssuer is null or whitespace AND validationParameters.ValidIssuers is null.";
         public const string IDX10205 = "IDX10205: Issuer validation failed. Issuer: '{0}'. Did not match: validationParameters.ValidIssuer: '{1}' or validationParameters.ValidIssuers: '{2}'.";
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
@@ -141,6 +141,11 @@ namespace Microsoft.IdentityModel.Tokens
         private string _nameClaimType = ClaimsIdentity.DefaultNameClaimType;
         private string _roleClaimType = ClaimsIdentity.DefaultRoleClaimType;
 
+        private bool _requireExpirationTime;
+        private bool _validateAudience;
+        private bool _validateIssuer;
+        private bool _validateLifetime;
+
         /// <summary>
         /// This is the fallback authenticationtype that a <see cref="ISecurityTokenValidator"/> will use if nothing is set.
         /// </summary>
@@ -153,7 +158,7 @@ namespace Microsoft.IdentityModel.Tokens
         public static readonly TimeSpan DefaultClockSkew = TimeSpan.FromSeconds(300); // 5 min.
 
         /// <summary>
-        /// Default for the maximm token size.
+        /// Default for the maximum token size.
         /// </summary>
         /// <remarks>250 KB (kilobytes).</remarks>
         public const Int32 DefaultMaximumTokenSizeInBytes = 1024 * 250;
@@ -169,7 +174,7 @@ namespace Microsoft.IdentityModel.Tokens
             AlgorithmValidator = other.AlgorithmValidator;
             ActorValidationParameters = other.ActorValidationParameters?.Clone();
             AudienceValidator = other.AudienceValidator;
-            _authenticationType = other._authenticationType;
+            _authenticationType = other.AuthenticationType;
             ClockSkew = other.ClockSkew;
             CryptoProviderFactory = other.CryptoProviderFactory;
             IgnoreTrailingSlashWhenValidatingAudience = other.IgnoreTrailingSlashWhenValidatingAudience;
@@ -183,7 +188,7 @@ namespace Microsoft.IdentityModel.Tokens
             NameClaimTypeRetriever = other.NameClaimTypeRetriever;
             PropertyBag = other.PropertyBag;
             RequireAudience = other.RequireAudience;
-            RequireExpirationTime = other.RequireExpirationTime;
+            _requireExpirationTime = other.RequireExpirationTime;
             RequireSignedTokens = other.RequireSignedTokens;
             RoleClaimType = other.RoleClaimType;
             RoleClaimTypeRetriever = other.RoleClaimTypeRetriever;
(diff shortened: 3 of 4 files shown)


src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -182,9 +189,9 @@ protected TokenValidationParameters(TokenValidationParameters other)
             NameClaimType = other.NameClaimType;
             NameClaimTypeRetriever = other.NameClaimTypeRetriever;
             PropertyBag = other.PropertyBag;
-            RequireAudience = other.RequireAudience;
-            RequireExpirationTime = other.RequireExpirationTime;
-            RequireSignedTokens = other.RequireSignedTokens;
+            _requireAudience = other._requireAudience;
+            _requireExpirationTime = other._requireExpirationTime;
```
TimHannMSFT: here and elsewhere, no need to change getter usage, will keep old format for consistency.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -46,6 +46,12 @@ internal static class LogMessages
         public const string IDX10103 = "IDX10103: RoleClaimType cannot be null or whitespace.";
         public const string IDX10104 = "IDX10104: TokenLifetimeInMinutes must be greater than zero. value: '{0}'";
         public const string IDX10105 = "IDX10105: ClaimValue that is a collection of collections is not supported. Such ClaimValue is found for ClaimType : '{0}'";
+        public const string IDX10106 = "IDX10106: Using this property to alter the default value and disable expiry requirements is not recommended. https://aka.ms/WilsonTokenValidation";
```
TimHannMSFT: This seemed like a mostly appropriate place for this. If the team would prefer somewhere else or a new const file I'm happy to go that direction too.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestUtilities.cs
```diff
@@ -116,10 +116,12 @@ internal static async Task<JsonWebKey> DecryptSymmetricPopKeyAsync(JsonWebTokenH
             var tokenDecryptionParameters = new TokenValidationParameters()
             {
                 TokenDecryptionKeys = decryptionKeys,
+#pragma warning disable CS0618 // TODO: If this is a valid use of RequireSignedTokens, perhaps this property should be left as not obsolete to not disrupt our consumers.
                 RequireSignedTokens = false,
```
TimHannMSFT, line 119: Hoping for feedback/explanation here for why we don't require tokens to be signed in this case? Does it have something to do with the SHR being signed as the cryptographic protection being relied on?

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -447,19 +438,43 @@ public string NameClaimType
         /// Gets or sets a value indicating whether SAML tokens must have at least one AudienceRestriction.
         /// </summary>
         [DefaultValue(true)]
-        public bool RequireAudience { get; set; }
+        public bool RequireAudience
+        {
+            get
+            {
+                return _requireAudience;
```
RojaEnnam: I might be missing something, why this getter is different for the rest that also have respective private variables? what problem are we solving with private variables?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestUtilities.cs
GeoK: PoP key can be represented as an encrypted JsonWebKey ("jwe" claim of a confirmation key). In this case it's only important to decrypt the JsonWebKey as we don't really care about validating claims and signature (it's not an access token).

https://datatracker.ietf.org/doc/html/rfc7800#section-3.3

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestUtilities.cs
```diff
@@ -116,10 +116,12 @@ internal static async Task<JsonWebKey> DecryptSymmetricPopKeyAsync(JsonWebTokenH
             var tokenDecryptionParameters = new TokenValidationParameters()
             {
                 TokenDecryptionKeys = decryptionKeys,
+#pragma warning disable CS0618 // TODO: If this is a valid use of RequireSignedTokens, perhaps this property should be left as not obsolete to not disrupt our consumers.
```
GeoK: Some legacy tokens are/can be unsigned.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
GeoK: nit: we have aka.ms/wilson/ path reserved. Perhaps it's more consistent to use aka.ms/wilson/tokenvalidation in this case.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -614,7 +641,13 @@ public string RoleClaimType
         /// property is true or false.
         /// </remarks>
         [DefaultValue(true)]
-        public bool ValidateLifetime { get; set; }
+        public bool ValidateLifetime
+        {
+            get;
+
+            [Obsolete(LogMessages.IDX10111, false)]
```
GeoK: We should call out this as a potential _breaking_ change  (if warnings are threated as errors), in the release notes.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestUtilities.cs
TimHannMSFT: Makes sense, I'll update to no longer mark this property as obsolete then.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
TimHannMSFT: It was extra confusing because I didn't finish doing it for the rest of the properties so the code was incorrect 😳. The private variables are needed here as the constructors for TVP were using the (now obsolete) setters which were causing warnings (which are treated as errors) at compile time.

RojaEnnam: Makes sense, Thanks.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
TimHannMSFT: Draft notes are now published here https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/releases/tag/untagged-cc6fe1c38dd4b5bc70e2

TimHannMSFT:
Opted to just mark the following obsolete

            "RequireExpirationTime",  // Can be overridden by the LifetimeValidator
            "ValidateAudience", // Can be overridden by the AudienceValidator
            "ValidateIssuer",  // Can be overridden by the IssuerValidator
            "ValidateLifetime",  // Can be overridden by the LifetimeValidator

Everything else seemed to have valid usages or no good override validator (RequireAudience was always evaluated and couldn't be skipped by using the associated validator).

TimHannMSFT:
For ease of review, here's the proposed wording for the release notes (https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/releases/tag/untagged-cc6fe1c38dd4b5bc70e2):

Breaking changes from 6.6.0:
If the consumer has turned on warnings as errors they may face breaking changes from some TokenValidationParameters being marked as obsolete/deprecated. There's no intention at the moment to remove these properties but changing them from the defaults represents significant degradation in security and is strongly recommended against. (the setters for RequireExpirationTime, ValidateAudience, ValidateIssuer, and ValidateLifetime have all been marked as obsolete to reflect they should no longer be used to alter the default values--as altering the default values disables important validation). For best practices on token validation, please refer to the wiki.

kevinchalet:
> the setters for RequireExpirationTime, ValidateAudience, ValidateIssuer, and ValidateLifetime have all been marked as obsolete to reflect they should no longer be used to alter the default values--as altering the default values disables important validation

I respectfully disagree with the changes introduced in this PR: there are actual use cases for disabling validation (e.g you typically don't want to reject identity tokens used as `id_token_hint`s just because they are expired).

I'm even tempted to say that the recommended alternative - assigning a delegate that does nothing - is way more dangerous than the existing approach, which is crystal clear and can be flagged by analyzers if needed.

Do you seriously think that `IssuerValidator = (issuer, token, tvp) => { return issuer; }` is better than `ValidateIssuer = false`?

leastprivilege:
I agree. 

There are also valid use cases for not validating an audience. OAuth does not even have the concept of an audience. they call it scope instead. 

`iss`, `aud`, `exp` etc. are all optional claims per JWT spec.

TimHannMSFT:
Thanks for the feedback @kevinchalet  & @leastprivilege , still discussing options inside the team. We are also investigating analyzers. Re: delegate vs bool flag, thinking around this is that by setting the delegate the consumer is much more actively--and obviously--denoting they're taking matters into their own hands; happy to admit reasonable people can disagree on this point though.

At base we're trying to best balance pain of this change with pain of security gaps. Both of your examples are good points of data in that discussion. Much appreciated.

kevinchalet:
> Thanks for the feedback @kevinchalet & @leastprivilege , still discussing options inside the team.

I'm sure you're already well aware of that, but I'd like to emphasize that IdentityModel is used in a myriad of third-party libraries, including @leastprivilege's IdentityServer and the OpenIddict stack I maintain. In these libraries, IdentityModel often has a critical role to play and it's something I'd love Microsoft to keep in mind: discussing options internally is fine, discussing options with everyone concerned - community included - would be way better. I know IdentityModel is maintained by the Azure AD team and that first-party usages dictate its development, but please don't forget the external projects that depend on it.

leastprivilege:
I personally would like to avoid using this library at all. Given the importance and central role in the .NET stack - code quality, design decisions and versioning policies have been problematic in the past (to say the least).

Unfortunately the tight coupling with ASP.NET Core makes this not feasible right now.

JWT (and its satellites) are specs - they are immutable. Once there is an implementation for a spec, it is done. That's what the core JWT lib should be. If you then want to layer the opinionated (or AAD) approach of the day on top - do it separately, but leave the core alone.

JsonWebToken/Handler is a step in the right direction. `TokenValidationParameters` needs a rebirth as well for that plumbing.

..and instead of massaging some properties - maybe you should fix your JSON serialization problems first.

brentschmaltz:
@leastprivilege 

> and instead of massaging some properties - maybe you should fix your JSON serialization problems first.

As you noticed we have invested in JsonWebToken/Handler and will continue to do so.
We have improved the performance of our base64urlencoding and will be moving to the Concept of ClaimsSets, with JsonClaimSet as being the first.
For NET 461 and above, we are switching to System.Text.Json, decoding the JWT into bytes, then accessing claims directly from the JsonElement.
The ClaimsIdentity will be Lazy and only users that need will cause the json to be translated to claims.
Most of the work in taking place in this topic branch brentsch/json.

kevinchalet:
> We are also investigating analyzers.

For the record, it's happening here:
  - https://github.com/dotnet/roslyn-analyzers/pull/5420
  - https://github.com/dotnet/docs/pull/25911

@TimHannMSFT it's always a good idea to add links to related tickets/PRs so that folks can track the progress of a feature, specially when the approach changes quite radically.

TimHannMSFT:
Appreciate the community feedback here. After getting more feedback and discussing internally we have opted to abandon this change. As I mentioned (and @kevinchalet linked above), we're looking to solve the same problem via Roslyn analyzers which will allow a much more granular opt out which allows us to better meet our goals while providing minimal disruption for consumers of this library.

PR #1688, opened 2021-08-19
Introducing interface for authentication scheme validation result

rymeskar:
Introducing interface for authentication scheme validation result. This interface will be used for flowing low-level results of overall authentication (scheme) validation. 

The interface design allows for good compatibility with current models used.


brentschmaltz:
I am not seeing how or where this is used.
The idea may make sense, but there needs to be more of a use case.

rymeskar:
The intended use case is that this model might be used on the path for 1P/3P convergence/extensibility of the validation libraries in Microsoft.

PR #1695, opened 2021-08-31
Adding the BaseConfigurationManager to the TokenValidationParameters

mafurman:
This will need to go in once the BaseConfiguration and BaseConfigurationManager classes PR is merged in (#1679). 

I've added some of the reasoning behind the design decisions in the comments below.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -31,6 +31,7 @@ using System.Linq;
 using System.Security.Claims;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading;
 using Microsoft.IdentityModel.Json;
 using Microsoft.IdentityModel.Json.Linq;
 using Microsoft.IdentityModel.Logging;
@@ -959,15 +960,43 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (tokenParts.Length != JwtConstants.JwsSegmentCount && tokenParts.Length != JwtConstants.JweSegmentCount)
                 return new TokenValidationResult { Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14111, token))), IsValid = false };
 
+            var validationParametersCopy = validationParameters;
+            if (validationParameters.ConfigurationManager != null)
+            {
+                try
+                {
+                    var configuration = validationParametersCopy.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
+                    validationParametersCopy = validationParameters.Clone();
+                    validationParametersCopy.Configuration = configuration;
+                }
+                catch (Exception ex)
+                {
+                    // If either a signing key or a valid issuer was not provided through the TVP, we want to stop processing on configuration retrieval failure.
+                    // Otherwise, we should keep going with validation.
+                    if ((validationParametersCopy.ValidateIssuer && string.IsNullOrWhiteSpace(validationParametersCopy.ValidIssuer) && validationParametersCopy.ValidIssuers.IsNullOrEmpty() && validationParametersCopy.IssuerValidator == null
+                        && validationParametersCopy.TokenDecryptionKey == null && validationParametersCopy.TokenDecryptionKeys.IsNullOrEmpty() && validationParametersCopy.TokenDecryptionKeyResolver == null)
+                        || (validationParametersCopy.RequireSignedTokens && validationParametersCopy.IssuerSigningKey == null && validationParametersCopy.IssuerSigningKeys.IsNullOrEmpty() && validationParametersCopy.IssuerSigningKeyResolver == null))
+                    {
+                        return new TokenValidationResult
+                        {
+                            Exception = ex,
+                            IsValid = false
+                        };
+                    }
+
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(TokenLogMessages.IDX10261, validationParametersCopy.ConfigurationManager.MetadataAddress));
+                }
+            }
+
             try
             {
                 if (tokenParts.Length == JwtConstants.JweSegmentCount)
                 {
                     var jwtToken = new JsonWebToken(token);
-                    var decryptedJwt = DecryptToken(jwtToken, validationParameters);
-                    var innerToken = ValidateSignature(decryptedJwt, validationParameters);
+                    var decryptedJwt = DecryptToken(jwtToken, validationParametersCopy);
+                    var innerToken = ValidateSignature(decryptedJwt, validationParametersCopy);
                     jwtToken.InnerToken = innerToken;
-                    var innerTokenValidationResult = ValidateTokenPayload(innerToken, validationParameters);
+                    var innerTokenValidationResult = ValidateTokenPayload(innerToken, validationParametersCopy);
                     return new TokenValidationResult
                     {
                         SecurityToken = jwtToken,
@@ -978,8 +1007,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 }
                 else
                 {
-                    var jsonWebToken = ValidateSignature(token, validationParameters);
-                    return ValidateTokenPayload(jsonWebToken, validationParameters);
+                    var jsonWebToken = ValidateSignature(token, validationParametersCopy);
+                    return ValidateTokenPayload(jsonWebToken, validationParametersCopy);
                 }
             }
             catch (Exception ex)
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -25,9 +25,6 @@
 //
 //------------------------------------------------------------------------------
 
-using Microsoft.IdentityModel.Json.Linq;
-using Microsoft.IdentityModel.Logging;
-using Microsoft.IdentityModel.Tokens;
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
@@ -35,6 +32,10 @@ using System.Globalization;
 using System.Security.Cryptography;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading;
+using Microsoft.IdentityModel.Json.Linq;
+using Microsoft.IdentityModel.Logging;
+using Microsoft.IdentityModel.Tokens;
 using TokenLogMessages = Microsoft.IdentityModel.Tokens.LogMessages;
 
 namespace Microsoft.IdentityModel.JsonWebTokens
@@ -380,9 +381,43 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// <param name="validationParameters">A <see cref="TokenValidationParameters"/>  required for validation.</param>
         /// <returns>Returns a <see cref="SecurityKey"/> to use for signature validation.</returns>
         /// <remarks>If key fails to resolve, then null is returned</remarks>
+        private static SecurityKey ResolveTokenSigningKeyWithConfig(string kid, string x5t, TokenValidationParameters validationParameters)
+        {
+            if (validationParameters.Configuration?.SigningKeys == null)
+                return null;
+
+            if (!string.IsNullOrEmpty(kid))
+            {
+                foreach (SecurityKey signingKey in validationParameters.Configuration.SigningKeys)
+                {
+                    if (signingKey != null && string.Equals(signingKey.KeyId, kid, signingKey is X509SecurityKey ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
+                        return signingKey;
+                }
+            }
+
+            if (!string.IsNullOrEmpty(x5t))
+            {
+                foreach (SecurityKey signingKey in validationParameters.Configuration.SigningKeys)
+                {
+                    if (signingKey != null && string.Equals(signingKey.KeyId, x5t, StringComparison.Ordinal))
+                        return signingKey;
+                }
+
+            }
+
+            return null;
+        }
+
+        /// <summary>
+        /// Returns a <see cref="SecurityKey"/> to use when validating the signature of a token.
+        /// </summary>
+        /// <param name="kid">The <see cref="string"/> kid field of the token being validated</param>
+        /// <param name="x5t">The <see cref="string"/> x5t field of the token being validated</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/>  required for validation.</param>
+        /// <returns>Returns a <see cref="SecurityKey"/> to use for signature validation.</returns>
+        /// <remarks>If key fails to resolve, then null is returned</remarks>
         internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, TokenValidationParameters validationParameters)
         {
-
             if (!string.IsNullOrEmpty(kid))
             {              
                 if (validationParameters.IssuerSigningKey != null
@@ -425,7 +460,10 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 }
             }
 
-            return null;
+            if (validationParameters.Configuration != null)
+                return ResolveTokenSigningKeyWithConfig(kid, x5t, validationParameters);
+            else
+                return null;
         }
     }
 }
--- a/src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
@@ -72,9 +72,14 @@ namespace Microsoft.IdentityModel.Tokens
         public static readonly TimeSpan DefaultRefreshInterval = new TimeSpan(0, 0, 5, 0);
 
         /// <summary>
-        /// The last known good configuration (a configuration retrieved in the past that we were able to successfully validate a token against).
+        /// The last known good configuration or LKG (a configuration retrieved in the past that we were able to successfully validate a token against).
         /// </summary>
-        public BaseConfiguration LKGConfiguration { get; set; }
+        public BaseConfiguration LastKnownGoodConfiguration { get; set; }
+
+        /// <summary>
+        /// The metadata address to retrieve the configuration from.
+        /// </summary>
+        public string MetadataAddress { get; set; }
 
         /// <summary>
         /// 5 minutes is the minimum value for automatic refresh. <see cref="AutomaticRefreshInterval"/> can not be set less than this value.
@@ -102,9 +107,9 @@ namespace Microsoft.IdentityModel.Tokens
         }
 
         /// <summary>
-        /// Indicates whether the LKG can be used, false by default.
+        /// Indicates whether the last known good configuraton (LKG) can be used, false by default.
         /// </summary>
-        public bool UseLKG { get; set; } = false;
+        public bool UseLastKnownGoodConfiguration { get; set; } = false;
 
         /// <summary>
         /// Obtains an updated version of <see cref="BaseConfiguration"/> if the appropriate refresh interval has passed.
--- a/src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
@@ -150,6 +150,10 @@ namespace Microsoft.IdentityModel.Tokens
             if (validationParameters.IssuerSigningKeys != null)
                 foreach (SecurityKey key in validationParameters.IssuerSigningKeys)
                     yield return key;
+
+            if (validationParameters.Configuration?.SigningKeys != null)
+                foreach (SecurityKey key in validationParameters.Configuration.SigningKeys)
+                    yield return key;
         }
 
         /// <summary>
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
@@ -355,6 +355,24 @@ namespace Microsoft.IdentityModel.Tokens
         }
 
         /// <summary>
+        /// This property represents the metadata (specifically the issuer and signing keys) retrieved by the <see cref="ConfigurationManager"/>.
+        /// It will be used along with the other properties set on the TokenValidationParameters in order to validate the token.
+        /// </summary>
+        /// <remarks>
+        /// This property can be set to the <see cref="BaseConfigurationManager.CurrentConfiguration"/> or the <see cref="BaseConfigurationManager.LastKnownGoodConfiguration"/>
+        /// depending on where in the Wilson call stack it is being used.
+        /// We make a copy of the TokenValidationParameters passed in to the JsonWebTokenHandler.ValidateToken(...) method, so the value of this
+        /// property will never change on the original object.
+        /// </remarks>
+        public BaseConfiguration Configuration { get; set; }
+
+        /// <summary>
+        /// If set, this property will be used to obtain the issuer and signing keys associated with the metadata endpoint of <see cref="BaseConfiguration.Issuer"/>.
+        /// The obtained issuer and signing keys will then be used along with those present on the TokenValidationParameters for validation of the incoming token.
+        /// </summary>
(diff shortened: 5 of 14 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1075,7 +1075,12 @@ private static JsonWebToken ValidateSignature(string token, TokenValidationParam
             }
             else
             {
-                var key = JwtTokenUtilities.ResolveTokenSigningKey(jwtToken.Kid, jwtToken.X5t, validationParameters);
+                SecurityKey key;
+                if (validationParameters.ConfigurationManager == null)
+                    key = JwtTokenUtilities.ResolveTokenSigningKey(jwtToken.Kid, jwtToken.X5t, validationParameters);
+                else
+                    key = JwtTokenUtilities.ResolveTokenSigningKeyWithConfig(jwtToken.Kid, jwtToken.X5t, validationParameters);
```
mafurman: I've separated this out into two different methods in the hopes that this will make it easier for us to add a ValidateAsync(...) API in the future.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -380,9 +381,53 @@ private static long ParseTimeValue(JToken jToken, string claimName)
         /// <param name="validationParameters">A <see cref="TokenValidationParameters"/>  required for validation.</param>
         /// <returns>Returns a <see cref="SecurityKey"/> to use for signature validation.</returns>
         /// <remarks>If key fails to resolve, then null is returned</remarks>
-        internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, TokenValidationParameters validationParameters)
+        internal static SecurityKey ResolveTokenSigningKeyWithConfig(string kid, string x5t, TokenValidationParameters validationParameters)
         {
+            StandardConfiguration configuration = null;
+            if (validationParameters.ConfigurationManager != null)
+                configuration = validationParameters.ConfigurationManager.GetStandardConfigurationAsync(CancellationToken.None).Result;
```
mafurman: This needs to be synchronous for now as we do not want to add a ValidateAsync(...) API that doesn't have asynchronous verifying operations.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -136,6 +137,26 @@ internal static object GetClaimValueUsingValueType(Claim claim)
...
+        /// </summary>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <returns>Returns all <see cref="SecurityKey"/> provided in validationParameters.</returns>
+        internal static IEnumerable<SecurityKey> GetAllSigningKeysWithConfig(TokenValidationParameters validationParameters)
+        {
+            LogHelper.LogInformation(TokenLogMessages.IDX10259);
+
+            StandardConfiguration configuration = null;
+            if (validationParameters.ConfigurationManager != null)
+                configuration = validationParameters.ConfigurationManager.GetStandardConfigurationAsync(CancellationToken.None).Result;
```
mafurman: Not sure if this Configuration should be retrieved in a single place and then passed on to other methods that need it.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: could JwtTokenUtilities contain the logic to include ConfigurationManager and have a single method?

mafurman: Made this change!

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -959,6 +960,30 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
             if (tokenParts.Length != JwtConstants.JwsSegmentCount && tokenParts.Length != JwtConstants.JweSegmentCount)
                 return new TokenValidationResult { Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14111, token))), IsValid = false };
 
+            if (validationParameters.ConfigurationManager != null)
+            {
+                try
+                {
+                    validationParameters.Configuration = validationParameters.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).Result;
```
kevinchalet: Doing sync-over-async is meh (in this case you may also have to do an IO operation to retrieve the configuration from the remote server if it's not already cached), but mutating the `TokenValidationParameters` instance in a `ValidateToken` method - whose name doesn't indicate it has side effects - seems concerning to me, specially since this method is very likely to be called concurrently.

This problem was solved years ago by injecting the signing keys in the TVP at a higher level (in the ASP.NET Core JWT handler, the TVP are cloned before injecting the signing keys resolved from the config). Why isn't this layering no longer suitable?

brentschmaltz: @kevinchalet we are investing in a model where we will be able to fall back to Last-Known-Good (LKG) for metadata when the existing metadata results in a token validation failure.
This can happen if an identity provider releases metadata misspelling the 'issuer'.
Updating metadata happens automatically on a user configurable interval, default is 12 hours, so while the issue is being resolved, the site will be done for 1-2 hours while the problem is solved.
This model change will allow for LKG to be used for a period of time.

Since we sit underneath multiple runtimes, we do not want to rely on layers above to interpret our errors to determine if a refresh is required as happens today with asp.net.

Using the LKG will be an opt in feature and the runtime will work as before if users choose not to use it.

To do that, we need a clean separation of what was added by the user AND what comes from metadata.

We should add a ConfigureAwait(false) here.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -425,7 +466,10 @@ internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, Token
                 }
             }
 
-            return null;
+            if (validationParameters.Configuration != null)
```
RojaEnnam, line 463: Not sure if it matters to do this check first, if one of the keys in validationParameters.IssuerSigningKeys works does it impact the decision of saving/using LKG?

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// 5 minutes is the default time interval that must pass for <see cref="RequestRefresh"/> to obtain a new configuration.
+        /// </summary>
+        public static readonly TimeSpan DefaultRefreshInterval = new TimeSpan(0, 0, 5, 0);
+
+        /// <summary>
+        /// The last known good configuration (a configuration retrieved in the past that we were able to successfully validate a token against).
+        /// </summary>
+        public BaseConfiguration LKGConfiguration { get; set; }
+
+
```
RojaEnnam, line 78: nit: extra new line?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: @brentschmaltz How would we add ConfigureAwait(false) here?

I know we can do GetAwaiter().GetResult() and that would allow us to bypass getting an AggregateException, but I'm not sure if that's the intention here.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: I agree that the check doesn't really matter, we could execute ResolveTokenSigningKeyWithConfig regardless of whether the configuration is present or not. But if it's null we avoid an additional method call.

The LKG will be saved even if the TVP was responsible for successful validation.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
mafurman: I think this was fixed in the first LKG PR that was recently merged in. :)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -959,6 +960,30 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
+            {
+                try
+                {
+                    validationParameters.Configuration = validationParameters.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).GetAwaiter().GetResult();
+                }
+                catch (AggregateException ex)
+                {
+                    // If either a signing key or a valid issuer was not provided through the TVP, we want to stop processing on configuration retrieval failure.
+                    // Otherwise, we should keep going with validation.
+                    if ((string.IsNullOrWhiteSpace(validationParameters.ValidIssuer) && validationParameters.ValidIssuers == null)
```
brentschmaltz: It is possible that delegates are available for ValidatingIssuer or Signature. ValidateIssuer may be off or require signature may be false.
We also need to consider the token may be in JWE format.
Best not to assume too much here.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -76,6 +76,11 @@ public TimeSpan AutomaticRefreshInterval
         /// </summary>
         public BaseConfiguration LKGConfiguration { get; set; }
 
+        /// <summary>
```
brentschmaltz, line 79: Should we use the full name, such as: LastKnownGood instead of LKG?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
brentschmaltz: @mafurman i agree with the null check as for any current asp.net users Configuration will always be null.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -380,9 +381,49 @@ private static long ParseTimeValue(JToken jToken, string claimName)
         /// <param name="validationParameters">A <see cref="TokenValidationParameters"/>  required for validation.</param>
         /// <returns>Returns a <see cref="SecurityKey"/> to use for signature validation.</returns>
         /// <remarks>If key fails to resolve, then null is returned</remarks>
-        internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, TokenValidationParameters validationParameters)
+        private static SecurityKey ResolveTokenSigningKeyWithConfig(string kid, string x5t, TokenValidationParameters validationParameters)
         {
+            if (!string.IsNullOrEmpty(kid))
+            {             
+                if (validationParameters.Configuration?.SigningKeys != null)
```
brentschmaltz: We don't need the null check here as this is called only when Configuration is non null.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
dannybtsai: Should we check for empty validationParameters.ValidIssuers too?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -959,6 +960,30 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
             if (tokenParts.Length != JwtConstants.JwsSegmentCount && tokenParts.Length != JwtConstants.JweSegmentCount)
                 return new TokenValidationResult { Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14111, token))), IsValid = false };
 
+            if (validationParameters.ConfigurationManager != null)
+            {
+                try
+                {
+                    validationParameters.Configuration = validationParameters.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).GetAwaiter().GetResult();
```
dannybtsai: This method performs validation but it also updates the validationParameters.Configuration, is that the desired behavior? Typically validation methods do not change data.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -380,9 +381,49 @@ private static long ParseTimeValue(JToken jToken, string claimName)
         /// <param name="validationParameters">A <see cref="TokenValidationParameters"/>  required for validation.</param>
         /// <returns>Returns a <see cref="SecurityKey"/> to use for signature validation.</returns>
         /// <remarks>If key fails to resolve, then null is returned</remarks>
-        internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, TokenValidationParameters validationParameters)
+        private static SecurityKey ResolveTokenSigningKeyWithConfig(string kid, string x5t, TokenValidationParameters validationParameters)
         {
```
dannybtsai, line 385: Check SigningKeys at the beginning and return null, no need to continue:

```
if (validationParameters.Configuration?.SigningKeys == null)
{
    return null;
}
```

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -380,9 +381,49 @@ private static long ParseTimeValue(JToken jToken, string claimName)
         /// <param name="validationParameters">A <see cref="TokenValidationParameters"/>  required for validation.</param>
         /// <returns>Returns a <see cref="SecurityKey"/> to use for signature validation.</returns>
         /// <remarks>If key fails to resolve, then null is returned</remarks>
-        internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, TokenValidationParameters validationParameters)
+        private static SecurityKey ResolveTokenSigningKeyWithConfig(string kid, string x5t, TokenValidationParameters validationParameters)
```
dannybtsai, line 384: Check SigningKeys at the beginning and return null, no need to continue:

```
if (validationParameters.Configuration?.SigningKeys == null)
{
    return null;
}
```

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
dannybtsai: @brentschmaltz , I think it is good to perform null check in case new code is added in the future that does not perform null check first.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -353,6 +353,18 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
             return new ClaimsIdentity(authenticationType: AuthenticationType ?? DefaultAuthenticationType, nameType: nameClaimType ?? ClaimsIdentity.DefaultNameClaimType, roleType: roleClaimType ?? ClaimsIdentity.DefaultRoleClaimType);
         }
 
+        /// <summary>
+        /// This property represents the metadata (specifically the issuer and signing keys) retrieved by the <see cref="ConfigurationManager"/>.
+        /// It will be used along with the other properties set on the TokenValidationParameters in order to validate the token.
+        /// </summary>
+        public BaseConfiguration Configuration { get; set; }
```
dannybtsai, line 367: Should the Configuration always come from the ConfigurationManager (maybe the BaseConfigurationManager.CurrentConfiguration)? Right now they are two separate properties. The comment says this is the metadata retrieved by ConfigurationManager.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Good suggestion; just added this in. :)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Ahhh, that's a valid concern. I've changed the logic so that we clone the TVP before we set the configuration property on it. Does that address this particular concern?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: I'm somewhat torn on what to do here. In a perfect world this method would only be called when the Configuration is non-null, but I'm not sure if we can expect the same in practice.

Do we want to throw a null reference exception here if it's null since this method should not be called without a Configuration present? I think having the null check could potentially be hiding an issue.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
mafurman: I changed it to LastKnownGood... hoping this way the meaning of this feature will be far more clear for developers if they don't read the documentation.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: I included checks here for the delegates, the ValidateIssuer and RequireSignedTokens flags, decryption keys, and the possibility of empty collections for the signing keys, decryption keys, and valid issuers.

The logic here has gotten quite complex, trying to think of whether this complexity is warranted. 

Should we continue with validation if a decryption key is present but there is no signing key?

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -223,7 +224,7 @@ public static string ValidateIssuer(string issuer, SecurityToken securityToken,
                     { InvalidIssuer = issuer });
 
             // Throw if all possible places to validate against are null or empty
-            if (string.IsNullOrWhiteSpace(validationParameters.ValidIssuer) && (validationParameters.ValidIssuers == null))
+            if (string.IsNullOrWhiteSpace(validationParameters.ValidIssuer) && (validationParameters.ValidIssuers == null) && (validationParameters.Configuration?.Issuer == null))
```
dannybtsai: Should this be changed to the following?
`if (string.IsNullOrWhiteSpace(validationParameters.ValidIssuer) && (validationParameters.ValidIssuers == null || !validationParameters.ValidIssuers.Any()) && (string.IsNullOrWhiteSpace(validationParameters.Configuration?.Issuer)))`

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
dannybtsai: The checks, line 391 & 405, can be removed.

src/Microsoft.IdentityModel.Tokens/Validators.cs
mafurman: Good catch! I added the empty check for ValidIssuers here as well.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
mafurman: I think setting this to always be equal to BaseConfigurationManager.CurrentConfiguration makes sense....that being said, this would mean there isn't any point in cloning the TVP to avoid potential modifications (as the value of this property will change once a configuration is retrieved).

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Whoops, good catch. Fixed.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
dannybtsai: I think throwing a NRE makes sense so it is consistent with the logic in many other methods.

dannybtsai: @mafurman, are you going to throw an exception?

src/Microsoft.IdentityModel.Tokens/Validators.cs
dannybtsai: We should check the Issuer with string.IsNullOrWhiteSpace(validationParameters.Configuration?.**Issuer**) so it is consistent with the string.IsNullOrWhiteSpace(validationParameters.**ValidIssuer**) check.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
dannybtsai: Add the ConfigureAwait(false) like this:

` var configuration = validationParametersCopy.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();`

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: I suggest we just continue and let it fail later.
It is hard to calculate all the possibilities

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
brentschmaltz: I'm ok with checking for null.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: @brentschmaltz Do you think including this check is still valuable for the scenarios it currently captures?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Thanks Danny! Didn't realize you could include ConfigureAwait(false) when using a asynchronous method synchronously. 

Fixed! https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1695/commits/f5b3504e5f237e293dce12f52e01e133ace89fe3

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
mafurman: I added a comment which explains that this property could be set to either the LastKnownGoodConfiguration or the Current Configuration. https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1695/commits/f5b3504e5f237e293dce12f52e01e133ace89fe3

src/Microsoft.IdentityModel.Tokens/Validators.cs
mafurman: Got it, fixed! https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1695/commits/f5b3504e5f237e293dce12f52e01e133ace89fe3

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: I added a CollectionUtilities class as Danny suggested since I think it simplifies the logic here quite a bit:  https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1695/files#diff-d6203951e787f84d178b2d2ae5f8375daa4e79d7a333b2ca9ee91a48ca66423f

```cs
public static bool IsNullOrEmpty<T>(this IEnumerable<T> enumerable)
        {
            return enumerable == null || !enumerable.Any();
        }
```

dannybtsai (approved):

brentschmaltz (approved):

PR #1698, opened 2021-09-02
Base64UrlEncoder performance improvements

### LINKED ISSUE #1655: Base64UrlEncoder.Encode fails on empty string
`Base64UrlEncoder.Encode("")` fails with the following exception on 6.10.0+:

```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
Stack Trace:
   at Microsoft.IdentityModel.Tokens.Base64UrlEncoder.UnsafeEncode(String str)
   at Microsoft.IdentityModel.Tokens.Base64UrlEncoder.EncodeString(String str)
   at Microsoft.IdentityModel.Tokens.Base64UrlEncoder.Encode(Byte[] inArray)
   at Microsoft.IdentityModel.Tokens.Base64UrlEncoder.Encode(String arg)
```

Possible cause: missing empty string check in new `UnsafeEncode` version:

```
       private unsafe static string UnsafeEncode(string str)
        {
            bool needReplace = false;
            int reductionSize = 0;
            if (str[str.Length - 1] == base64PadCharacter) // fails on empty string
                reductionSize = 1;
```

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/8e5bb2c5ec86aa3af44636ff28c48e855a904498#diff-d8157f3c41ee79206c7c316a85950971914d143866bc8ca23dcfc8d3658338b3R116

### LINKED ISSUE #1677: [Investigate] Improve allocations and speed of Base64UrlEncoder
## **Note**: We have had [improvements to the URLEncoding](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/d6f2b66d788195b50f2b1f700beb497851194c73/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs#L37) since this suggestion was written and thus actual improvements may be less, this needs to be investigated.

Suggestion by: [Feng Yuan](https://www.linkedin.com/in/dryuan/)

CreateJWSInternal is responsible for 5.42% allocation in the specific scenario investigated. Here is BaseUrlEncoder implementation (incomplete):
_There is a string.Split call inside._

Here is the code:
```C#
 internal static string CreateJWSInternal(string payload, string header, SigningCredentials signingCredentials)
  {
      var actualHeader = header != null ? JObject.Parse(header) : new JObject
      {
          { JwtClaimTypes.Alg, signingCredentials.Algorithm },
          { JwtClaimTypes.Kid, signingCredentials.Key.KeyId },
          { JwtClaimTypes.Typ, JwtClaimValues.JWT }
      };

      if (payload == null)
          throw S2SLogger.LogArgumentNullException(nameof(payload));

      if (header == null && signingCredentials.Key is X509SecurityKey x509SecurityKey)
          actualHeader[JwtClaimTypes.X5t] = x509SecurityKey.X5t;

      var cryptoFactory = signingCredentials.CryptoProviderFactory ?? signingCredentials.Key.CryptoProviderFactory;
      var signatureProvider = cryptoFactory.CreateForSigning(signingCredentials.Key, signingCredentials.Algorithm);
      if (signatureProvider == null)
          throw S2SLogger.LogArgumentNullException(S2SLogger.FormatInvariant(LogMessages.S2S32211, cryptoFactory.GetType(), signingCredentials.Key, signingCredentials.Algorithm));

      try
      {
          var message = Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(actualHeader.ToString(Formatting.None))) + "." + Base64UrlEncoder.Encode(payload);
          return message + "." + Base64UrlEncoder.Encode(signatureProvider.Sign(Encoding.UTF8.GetBytes(message)));
      }
      finally
      {
          cryptoFactory.ReleaseSignatureProvider(signatureProvider);
      }
  }
```
Base64UrlEncoder.Encode is called three times here, two with byte array form, one with string form. The later is converted to former with UTf8.GetBytes call. String.Concat are called twice.

Here is Encode implementation:
```C#
public static string Encode(byte[] inArray)
{
    return Convert.ToBase64String(inArray, 0, inArray.Length).Split(new char[]
    {
        Base64UrlEncoder.base64PadCharacter
    })[0].
    Replace(Base64UrlEncoder.base64Character62, Base64UrlEncoder.base64UrlCharacter62).
    Replace(Base64UrlEncoder.base64Character63, Base64UrlEncoder._base64UrlCharacter63);
}
```
It calls Convert.ToBase64String first, then string.Split is called to remove possible padding (this will allocate an int32 array and a string array even when there is no padding). Afterwards, string.Replace(char, char) is callled twice to replace char 62 and 63.

Convert.ToBase64String is not flexible enough (for example, there is no no-padding option), we need another base 64 encoder implementation. 

Here is one I just wrote:
```C#
/// <summary>
/// Base64 encoder for Url
/// </summary>
public class UrlBase64Encoder
{
    /// <summary>
    /// Char 62
    /// </summary>
    private const char base64UrlCharacter62 = '-';

    /// <summary>
    /// Char 63
    /// </summary>
    private const char base64UrlCharacter63 = '_';

    /// <summary>
    // Encoding table
    /// </summary>
    internal static readonly char[] base64Table = 
    {
        'A','B','C','D','E','F','G','H','I','J','K','L','M','N','O','P','Q','R','S','T','U','V','W','X','Y','Z',
        'a','b','c','d','e','f','g','h','i','j','k','l','m','n','o','p','q','r','s','t','u','v','w','x','y','z',
        '0','1','2','3','4','5','6','7','8','9',
        base64UrlCharacter62,
        base64UrlCharacter63
    };

    /// <summary>
    /// Reused encoder.
    /// </summary>
    private static UrlBase64Encoder reusedEncoder;

    /// <summary>
    /// Char buffer
    /// </summary>
    private char[] charBuffer = new char[128];

    /// <summary>
    /// Char count in the buffer
    /// </summary>
    private int charLength;

    /// <summary>
    /// Byte buffer
    /// </summary>
    private byte[] byteBuffer;

    /// <summary>
    /// Gets char count.
    /// </summary>
    public int Length => this.charLength;

    /// <summary>
    /// Acquire encoder.
    /// </summary>
    /// <returns>Empty encoder.</returns>
    public static UrlBase64Encoder AcquireEncoder()
    {
        var encoder = Interlocked.Exchange(ref UrlBase64Encoder.reusedEncoder, null);

        if (encoder == null)
        {
            encoder = new UrlBase64Encoder();
        }

        return encoder;
    }

    /// <summary>
    /// Release encoder for reuse.
    /// </summary>
    public void Release()
    {
        this.Clear();

        UrlBase64Encoder.reusedEncoder = this;
    }

    /// <summary>
    /// Clear for reuse
    /// </summary>
    public void Clear()
    {
        this.charLength = 0;
    }

    /// <summary>
    /// Append a single char without encoding.
    /// </summary>
    /// <param name="ch">Char to append</param>
    public void Append(char ch)
    {
        this.EnsureSpace(1);

        this.charBuffer[this.charLength++] = ch;
    }

    /// <summary>
    /// Encode a string
    /// </summary>
    /// <param name="str">string to encode</param>
    public void Encode(string str)
    {
        if (str == null)
        {
            throw new ArgumentNullException(nameof(str));
        }

        int length = str.Length;

        if (length != 0)
        {
            bool ascii = true;

            foreach (char ch in str)
            {
                if (ch >= 128)
                {
                    ascii = false;
                    break;
                }
            }

            if (ascii)
            {
                this.EncodeAscii(str, 0, length);
            }
            else
            {
                // Worst case encoded length.
                int byteLength = length * 6;

                if ((this.byteBuffer == null) || (this.byteBuffer.Length < byteLength))
                {
                    this.byteBuffer = new byte[byteLength];
                }

                byteLength = Encoding.UTF8.GetBytes(str, 0, length, this.byteBuffer, 0);

                this.Encode(this.byteBuffer, 0, byteLength);
            }
        }
    }

    /// <summary>
    /// Encoding ASCII string.
    /// </summary>
    /// <param name="data">String data</param>
    /// <param name="offset">start position</param>
    /// <param name="length">data length</param>
    public void EncodeAscii(string data, int offset, int length)
    {
        this.EnsureSpace((length + 2) / 3 * 4);

        int i;

        char[] table = UrlBase64Encoder.base64Table;
        char[] output = this.charBuffer;

        int j = this.charLength;

        int lengthmod3 = length % 3;
        int limit = offset + (length - lengthmod3);

        for (i = offset; i < limit; i += 3)
        {
            char d0 = data[i];
            char d1 = data[i + 1];
            char d2 = data[i + 2];

            output[j + 0] = table[(d0 & 0xfc) >> 2];
            output[j + 1] = table[((d0 & 0x03) << 4) | ((d1 & 0xf0) >> 4)];
            output[j + 2] = table[((d1 & 0x0f) << 2) | ((d2 & 0xc0) >> 6)];
            output[j + 3] = table[d2 & 0x3f];
            j += 4;
        }

        //Where we left off before
        i = limit;

        switch (lengthmod3)
        {
            case 2:
                {
                    char d0 = data[i];
                    char d1 = data[i + 1];

                    output[j + 0] = table[(d0 & 0xfc) >> 2];
                    output[j + 1] = table[((d0 & 0x03) << 4) | ((d1 & 0xf0) >> 4)];
                    output[j + 2] = table[(d1 & 0x0f) << 2];
                    j += 3;
                }
                break;

            case 1:
                {
                    char d0 = data[i];

                    output[j + 0] = table[d0 >> 2];
                    output[j + 1] = table[(d0 & 0x03) << 4];
                    j += 2;
                }
                break;
        }

        this.charLength = j;
    }

    /// <summary>
    /// Encoding from byte array.
    /// </summary>
    /// <param name="data">Data array</param>
    /// <param name="offset">start position</param>
    /// <param name="length">data length</param>
    public void Encode(byte[] data, int offset, int length)
    {
        this.EnsureSpace((length + 2) / 3 * 4);

        int i;

        char[] table = UrlBase64Encoder.base64Table;
        char[] output = this.charBuffer;

        int j = this.charLength;

        int lengthmod3 = length % 3;
        int limit = offset + (length - lengthmod3);

        for (i = offset; i < limit; i += 3)
        {
            byte d0 = data[i];
            byte d1 = data[i + 1];
            byte d2 = data[i + 2];

            output[j + 0] = table[(d0 & 0xfc) >> 2];
            output[j + 1] = table[((d0 & 0x03) << 4) | ((d1 & 0xf0) >> 4)];
            output[j + 2] = table[((d1 & 0x0f) << 2) | ((d2 & 0xc0) >> 6)];
            output[j + 3] = table[d2 & 0x3f];
            j += 4;
        }

        //Where we left off before
        i = limit;

        switch (lengthmod3)
        {
            case 2:
                {
                    byte d0 = data[i];
                    byte d1 = data[i + 1];

                    output[j + 0] = table[(d0 & 0xfc) >> 2];
                    output[j + 1] = table[((d0 & 0x03) << 4) | ((d1 & 0xf0) >> 4)];
                    output[j + 2] = table[(d1 & 0x0f) << 2];
                    j += 3;
                }
                break;

            case 1:
                {
                    byte d0 = data[i];

                    output[j + 0] = table[d0 >> 2];
                    output[j + 1] = table[(d0 & 0x03) << 4];
                    j += 2;
                }
                break;
        }

        this.charLength = j;
    }

    /// <summary>
    /// Convert content to byte array.
    /// </summary>
    /// <param name="encoding">Encoding to use.</param>
    /// <returns>encoded data.</returns>
    public byte[] GetBytes(Encoding encoding)
    {
        if (encoding == null)
        {
            throw new ArgumentNullException(nameof(encoding));
        }

        int length = encoding.GetByteCount(this.charBuffer, 0, this.charLength);

        byte[] result = new byte[length];

        encoding.GetBytes(this.charBuffer, 0, this.charLength, result, 0);

        return result;
    }

    /// <summary>
    /// Convert content to string.
    /// </summary>
    /// <returns>String content.</returns>
    {
        return new string(this.charBuffer, 0, this.charLength);
    }

    /// <summary>
    /// Ensure char buffer has enough space.
    /// </summary>
    /// <param name="toAdd">Char to be added.</param>
    [MethodImpl(MethodImplOptions.AggressiveInlining)]
    private void EnsureSpace(int toAdd)
    {
        int total = this.charLength + toAdd;

        if (this.charBuffer.Length < total)
        {
            Array.Resize(ref this.charBuffer, Math.Max(total, this.charBuffer.Length * 2));
        }
    }
}
```

Now we can change the caller to:
```C#
static string CreateJWSInternal2(string payload, string header)
{
    var encoder = UrlBase64Encoder.AcquireEncoder();
    encoder.Encode(header);
    encoder.Append('.');
    encoder.Encode(payload);
    var encoded = encoder.GetBytes(Encoding.UTF8);

    // signing encoded.
    encoder.Append('.');
    encoder.Encode(encoded, 0, encoded.Length);
    string result = encoder.ToString();
    encoder.Release();

    return result;
}
```

We just need a single string allocation, plus byte arrays needed for signing. There is no padding, no string.Split, no string.Replace, no string.Concat. The code is even skipping converting ASCII string to byte array. To make it even faster, we can use unsafe code.

Here is a simple test (with no signing):
```C#
static string CreateJWSInternal1(string payload, string header)
{
    var message = Base64UrlEncoder.Encode(header) + "." + Base64UrlEncoder.Encode(payload);

    return message + "." + Base64UrlEncoder.Encode(Encoding.UTF8.GetBytes(message));
}

static string CreateJWSInternal2(string payload, string header)
{
    var encoder = UrlBase64Encoder.AcquireEncoder();

    encoder.Encode(header);
    encoder.Append('.');
    encoder.Encode(payload);

    var encoded = encoder.GetBytes(Encoding.UTF8);

    // signing encoded.

    encoder.Append('.');
    encoder.Encode(encoded, 0, encoded.Length);

    string result = encoder.ToString();

    encoder.Release();

    return result;
}

public static void Test()
{
    for (int i = 0; i < 3; i++)
    {
        var s1 = CreateJWSInternal1("payload.Microsoft", "header.Microsoft");

        var s2 = CreateJWSInternal2("payload.Microsoft", "header.Microsoft");

        Console.WriteLine(s1);
        Console.WriteLine(s2);
        Console.WriteLine(s1.Equals(s2));
    }
}
```

**Allocation before:**
Total size: 1841; 20 heap allocations, 1,841 bytes in total. Notice string.Split is called three times (12 allocations in total). String.Replace is not causing allocation (char 62/63 not found).

**Allocation after:**
Total size: 314; 2 heap allocations, 314 bytes in total (83% down).

cesarpayan:
Changed code paths to encode with bytes always. Temporaly breaking paths to EncodeString TBD if we can remove it.

Encode(byte, int, int) is optimized, as well as current UnsafeEncode (part of the EncodeString code path)

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
@@ -86,30 +86,17 @@ namespace Microsoft.IdentityModel.Tokens
         {
             _ = inArray ?? throw LogHelper.LogArgumentNullException("inArray");
 
-            if (length < 0)
-            {
-                throw LogHelper.LogArgumentException<ArgumentException>(nameof(length), $"{nameof(length)} must be greater or equal than zero");
-            }
-
-            if (offset < 0)
-            {
-                throw LogHelper.LogArgumentException<ArgumentException>(nameof(offset), $"{nameof(offset)} must be greater or equal than zero");
-            }
-
             if (length == 0)
-            {
                 return string.Empty;
-            }
 
-            if (inArray.Length < offset)
-            {
-                throw LogHelper.LogArgumentException<ArgumentException>(nameof(offset), $"Invalid value for {nameof(offset)} parameter");
-            }
+            if (length < 0)
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10106, nameof(length), length)));
+
+            if (offset < 0 || inArray.Length < offset)
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10106, nameof(offset), offset)));
 
             if (inArray.Length < offset + length)
-            {
-                throw LogHelper.LogArgumentException<ArgumentException>(nameof(length), $"Invalid value for {nameof(length)} parameter");
-            }
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10106, nameof(length), length)));
 
             int i;
             char[] table = s_base64Table;
--- a/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
@@ -46,6 +46,18 @@ namespace Microsoft.IdentityModel.Tokens
         private static char base64UrlCharacter63 = '_';
 
         /// <summary>
+        /// Encoding table
+        /// </summary>
+        internal static readonly char[] s_base64Table =
+        {
+            'A','B','C','D','E','F','G','H','I','J','K','L','M','N','O','P','Q','R','S','T','U','V','W','X','Y','Z',
+            'a','b','c','d','e','f','g','h','i','j','k','l','m','n','o','p','q','r','s','t','u','v','w','x','y','z',
+            '0','1','2','3','4','5','6','7','8','9',
+            base64UrlCharacter62,
+            base64UrlCharacter63
+        };
+
+        /// <summary>
         /// The following functions perform base64url encoding which differs from regular base64 encoding as follows
         /// * padding is skipped so the pad character '=' doesn't have to be percent encoded
         /// * the 62nd and 63rd regular base64 encoding characters ('+' and '/') are replace with ('-' and '_')
@@ -76,7 +88,59 @@ namespace Microsoft.IdentityModel.Tokens
             if (inArray == null)
                 throw LogHelper.LogArgumentNullException("inArray");
 
-            return EncodeString(Convert.ToBase64String(inArray, offset, length));
+            int i;
+
+            char[] table = s_base64Table;
+
+
+            int j = 0;
+
+            int lengthmod3 = length % 3;
+            int limit = offset + (length - lengthmod3);
+            char[] output = new char[(length + 2) / 3 * 4];
+
+            for (i = offset; i < limit; i += 3)
+            {
+                byte d0 = inArray[i];
+                byte d1 = inArray[i + 1];
+                byte d2 = inArray[i + 2];
+
+                output[j + 0] = table[(d0 & 0xfc) >> 2];
+                output[j + 1] = table[((d0 & 0x03) << 4) | ((d1 & 0xf0) >> 4)];
+                output[j + 2] = table[((d1 & 0x0f) << 2) | ((d2 & 0xc0) >> 6)];
+                output[j + 3] = table[d2 & 0x3f];
+                j += 4;
+            }
+
+            //Where we left off before
+            i = limit;
+
+            switch (lengthmod3)
+            {
+                case 2:
+                    {
+                        byte d0 = inArray[i];
+                        byte d1 = inArray[i + 1];
+
+                        output[j + 0] = table[(d0 & 0xfc) >> 2];
+                        output[j + 1] = table[((d0 & 0x03) << 4) | ((d1 & 0xf0) >> 4)];
+                        output[j + 2] = table[(d1 & 0x0f) << 2];
+                        j += 3;
+                    }
+                    break;
+
+                case 1:
+                    {
+                        byte d0 = inArray[i];
+
+                        output[j + 0] = table[d0 >> 2];
+                        output[j + 1] = table[(d0 & 0x03) << 4];
+                        j += 2;
+                    }
+                    break;
+            }
+
+            return new(output, 0, j);
         }
 
         /// <summary>
@@ -92,7 +156,7 @@ namespace Microsoft.IdentityModel.Tokens
             if (inArray == null)
                 throw LogHelper.LogArgumentNullException("inArray");
 
-            return EncodeString(Convert.ToBase64String(inArray, 0, inArray.Length));
+            return Encode(inArray, 0, inArray.Length);
         }
 
         internal static string EncodeString(string str)
@@ -111,60 +175,46 @@ namespace Microsoft.IdentityModel.Tokens
 #if !NET45
         private unsafe static string UnsafeEncode(string str)
         {
-            bool needReplace = false;
-            int reductionSize = 0;
-            if (str[str.Length - 1] == base64PadCharacter)
-                reductionSize = 1;
+            int length = str.Length;
 
-            if (str[str.Length - 2] == base64PadCharacter)
-                reductionSize = 2;
-
-            int encodedLength = str.Length - reductionSize;
-            for (int i = 0; i < str.Length; i++)
+            if (str[length - 1] == base64PadCharacter)
             {
-                if (str[i] == base64Character62 || str[i] == base64Character63)
+                int reductionSize = 1;
+
+                if (str[length - 2] == base64PadCharacter)
                 {
-                    needReplace = true;
-                    break;
+                    reductionSize = 2;
                 }
+
+                str = str.Substring(0, length - reductionSize);
             }
 
-            if (needReplace)
+            fixed (char* pSrc = str)
             {
-                string encodedString = new string(char.MinValue, encodedLength);
-                fixed (char* dest = encodedString)
-                {
-                    for (int i = 0; i < encodedLength; i++)
-                    {
-                        if (str[i] == base64Character62)
-                            dest[i] = base64UrlCharacter62;
-                        else if (str[i] == base64Character63)
-                            dest[i] = base64UrlCharacter63;
-                        else
-                            dest[i] = str[i];
-                    }
-                }
+                char* end = pSrc + str.Length;
+                char* p = pSrc;
 
-                return encodedString;
-            }
-            else
-            {
-                if (encodedLength == str.Length)
+                while (p < end)
                 {
-                    return str;
-                }
-                else
-                {
-                    string encodedString = new string(char.MinValue, encodedLength);
-                    fixed (char* src = str)
-                    fixed (char* dest = encodedString)
+                    char ch = *p;
+
+                    if (ch < '0')
                     {
-                        Buffer.MemoryCopy(src, dest, encodedLength * 2, encodedLength * 2);
+                        if (ch == base64Character62)
+                        {
+                            *p = base64UrlCharacter62;
+                        }
+                        else if (ch == base64Character63)
+                        {
+                            *p = base64UrlCharacter63;
+                        }
                     }
 
-                    return encodedString;
+                    p++;
                 }
             }
+
+            return str;
         }
 #endif
 
--- a/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
@@ -90,13 +90,13 @@ namespace Microsoft.IdentityModel.Tokens
                 return string.Empty;
 
             if (length < 0)
-                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10106, nameof(length), length)));
+                throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(LogHelper.FormatInvariant(LogMessages.IDX10106, nameof(length), length)));
 
             if (offset < 0 || inArray.Length < offset)
-                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10106, nameof(offset), offset)));
+                throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(LogHelper.FormatInvariant(LogMessages.IDX10106, nameof(offset), offset)));
 
             if (inArray.Length < offset + length)
-                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10106, nameof(length), length)));
+                throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(LogHelper.FormatInvariant(LogMessages.IDX10106, nameof(length), length)));
 
             int i;
             char[] table = s_base64Table;
@@ -166,64 +166,12 @@ namespace Microsoft.IdentityModel.Tokens
             return Encode(inArray, 0, inArray.Length);
         }
 
-        //TBD: Before removing EncodeString and its sub-methods, we would need to verify that InternalsVisibleTo("Microsoft.AzureAD.SmartSessionEvaluator...
-        // is not referencing this method. As of the moment of this comment, there's no reference to this method within Wilson.
         internal static string EncodeString(string str)
         {
-#if NET45
-            str = str.Split(base64PadCharacter)[0]; // Remove any trailing padding
-            str = str.Replace(base64Character62, base64UrlCharacter62);  // 62nd char of encoding
-            str = str.Replace(base64Character63, base64UrlCharacter63);  // 63rd char of encoding
+            _ = str ?? throw LogHelper.LogArgumentNullException("str");
 
-            return str;
-#else
-            return UnsafeEncode(str);
-#endif
+            return Encode(Encoding.UTF8.GetBytes(str));
         }
-
-#if !NET45
-        private unsafe static string UnsafeEncode(string str)
-        {
-            int length = str.Length;
-
-            if (length > 0 && str[length - 1] == base64PadCharacter)
-            {
-                int reductionSize = 1;
-                if (length > 1 && str[length - 2] == base64PadCharacter)
-                {
-                    reductionSize = 2;
-                }
-
-                str = str.Substring(0, length - reductionSize);
-            }
-
-            fixed (char* pSrc = str)
-            {
-                char* end = pSrc + str.Length;
-                char* p = pSrc;
-
-                while (p < end)
-                {
-                    char ch = *p;
-                    if (ch < '0')
-                    {
-                        if (ch == base64Character62)
-                        {
-                            *p = base64UrlCharacter62;
-                        }
-                        else if (ch == base64Character63)
-                        {
-                            *p = base64UrlCharacter63;
-                        }
-                    }
-
-                    p++;
-                }
-            }
-
-            return str;
-        }
-#endif
 
         /// <summary>
         ///  Converts the specified string, which encodes binary data as base-64-url digits, to an equivalent 8-bit unsigned integer array.</summary>
--- a/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
@@ -67,7 +67,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// <returns>Base64Url encoding of the UTF8 bytes.</returns>
         public static string Encode(string arg)
         {
-            _ = arg ?? throw LogHelper.LogArgumentNullException("arg");
+            _ = arg ?? throw LogHelper.LogArgumentNullException(nameof(arg));
 
             return Encode(Encoding.UTF8.GetBytes(arg));
         }
@@ -84,7 +84,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// <exception cref="ArgumentOutOfRangeException">offset or length is negative OR offset plus length is greater than the length of inArray.</exception>
         public static string Encode(byte[] inArray, int offset, int length)
         {
-            _ = inArray ?? throw LogHelper.LogArgumentNullException("inArray");
+            _ = inArray ?? throw LogHelper.LogArgumentNullException(nameof(inArray));
 
             if (length == 0)
                 return string.Empty;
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -45,6 +45,18 @@ public static class Base64UrlEncoder
         private static char base64UrlCharacter62 = '-';
```
brentschmaltz: please change these statics to const.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -92,7 +156,7 @@ public static string Encode(byte[] inArray)
             if (inArray == null)
                 throw LogHelper.LogArgumentNullException("inArray");
```
dannybtsai: Use nameof(inArray) to avoid a mismatch between variable name and the name in the exception if the variable is renamed in the future.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -76,7 +88,59 @@ public static string Encode(byte[] inArray, int offset, int length)
...
 
-            return EncodeString(Convert.ToBase64String(inArray, offset, length));
+            int i;
```
dannybtsai: nit: move i, j close to where they are used and remove the extra empty lines.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -76,7 +88,59 @@ public static string Encode(byte[] inArray, int offset, int length)
             if (inArray == null)
```
dannybtsai: Validate values (negative and 0) of offset and length.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -92,7 +156,7 @@ public static string Encode(byte[] inArray)
...
                 throw LogHelper.LogArgumentNullException("inArray");
 
-            return EncodeString(Convert.ToBase64String(inArray, 0, inArray.Length));
+            return Encode(inArray, 0, inArray.Length);
         }
 
         internal static string EncodeString(string str)
```
cesarpayan, line 169: Should we preserve this method and its sub-methods or is it safe to remove them?  It is currently unreferenced by Wilson.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -76,7 +88,59 @@ public static string Encode(byte[] inArray, int offset, int length)
             if (inArray == null)
                 throw LogHelper.LogArgumentNullException("inArray");
 
-            return EncodeString(Convert.ToBase64String(inArray, offset, length));
+            int i;
+
+            char[] table = s_base64Table;
+
```
TimHannMSFT, line 94: nit: remove double blank line

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -76,7 +88,59 @@ public static string Encode(byte[] inArray, int offset, int length)
             if (inArray == null)
                 throw LogHelper.LogArgumentNullException("inArray");
```
TimHannMSFT: nit: I'm still not 100% on the team's style guide on this but Brent was mentioning these sorts of ArgNull checks are commonly performed with discards (_) and the null-coalescing operator (??), so:

_ = inArray ?? throw LogHelper.LogArgumentNullException("inArray");

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -76,7 +88,59 @@ public static string Encode(byte[] inArray, int offset, int length)
...
+                output[j + 1] = table[((d0 & 0x03) << 4) | ((d1 & 0xf0) >> 4)];
+                output[j + 2] = table[((d1 & 0x0f) << 2) | ((d2 & 0xc0) >> 6)];
+                output[j + 3] = table[d2 & 0x3f];
+                j += 4;
+            }
+
+            //Where we left off before
+            i = limit;
+
+            switch (lengthmod3)
```
TimHannMSFT, line 124: nit: log an error on the default case, or Debug.Fail

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -111,60 +175,46 @@ internal static string EncodeString(string str)
 #if !NET45
         private unsafe static string UnsafeEncode(string str)
```
dannybtsai: Is "Encode" the right name? This function simply replace the 62 and 63 chars, so maybe rename it to UnsafeReplaceBase64UrlCharacters?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -76,7 +88,59 @@ public static string Encode(byte[] inArray, int offset, int length)
...
+            int limit = offset + (length - lengthmod3);
+            char[] output = new char[(length + 2) / 3 * 4];
+
+            for (i = offset; i < limit; i += 3)
+            {
+                byte d0 = inArray[i];
+                byte d1 = inArray[i + 1];
+                byte d2 = inArray[i + 2];
+
+                output[j + 0] = table[(d0 & 0xfc) >> 2];
```
dannybtsai: No need for the & op, can be optimized to:
table[d0 >> 2];

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -76,7 +88,59 @@ public static string Encode(byte[] inArray, int offset, int length)
...
+                output[j + 1] = table[((d0 & 0x03) << 4) | ((d1 & 0xf0) >> 4)];
```
dannybtsai: & is not needed, can be optimized to:
table[((d0 & 0x03) << 4) | (d1 >> 4)];

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -76,7 +88,59 @@ public static string Encode(byte[] inArray, int offset, int length)
...
+                output[j + 2] = table[((d1 & 0x0f) << 2) | ((d2 & 0xc0) >> 6)];
```
dannybtsai: Optimize to **table[((d1 & 0x0f) << 2) | d2 >> 6)];**

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -76,7 +88,59 @@ public static string Encode(byte[] inArray, int offset, int length)
...
+            i = limit;
+
+            switch (lengthmod3)
+            {
+                case 2:
+                    {
+                        byte d0 = inArray[i];
+                        byte d1 = inArray[i + 1];
+
+                        output[j + 0] = table[(d0 & 0xfc) >> 2];
```
dannybtsai: See comments above on optimization.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
TimHannMSFT: Seems if you were to check the projects listed as InternalsVisibleTo (\src\Microsoft.IdentityModel.Tokens\Properties\AssemblyInfo.cs) and none of them used it, removing would be a good thing imo.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -76,7 +88,59 @@ public static string Encode(byte[] inArray, int offset, int length)
...
 
-            return EncodeString(Convert.ToBase64String(inArray, offset, length));
+            int i;
+
+            char[] table = s_base64Table;
+
+
+            int j = 0;
+
+            int lengthmod3 = length % 3;
```
dannybtsai, line 101: Maybe add some comments to explain the logic of converting from 3 bytes to 4 bytes for anyone who is not familiar with the algorithm?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -111,60 +175,46 @@ internal static string EncodeString(string str)
 #if !NET45
```
dannybtsai: What is the reason for checking NET45?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
cesarpayan: The variable has 3 potential values (given that its value = length%3). In the case the value being 0, it wouldn't denote an error, but a case where no more operations are needed. Perhaps this would be more descriptive with if conditions instead of a switch?

TimHannMSFT: Oops, my bad, just skimmed wrong. switch statement is fine, might help readability a tiny bit to just comment the default case but I think this is fine how you have it.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -73,10 +84,84 @@ public static string Encode(string arg)
         /// <exception cref="ArgumentOutOfRangeException">offset or length is negative OR offset plus length is greater than the length of inArray.</exception>
         public static string Encode(byte[] inArray, int offset, int length)
         {
-            if (inArray == null)
-                throw LogHelper.LogArgumentNullException("inArray");
+            _ = inArray ?? throw LogHelper.LogArgumentNullException("inArray");
+
+            if (length < 0)
```
brentschmaltz: nit: for a single if clause, we sometimes skip the {.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -73,10 +84,84 @@ public static string Encode(string arg)
...
+            {
+                throw LogHelper.LogArgumentException<ArgumentException>(nameof(length), $"{nameof(length)} must be greater or equal than zero");
```
brentschmaltz: why does a method named 'LogArgumentException' need to be generic?
why the &lt;ArgumentException&gt;

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
brentschmaltz: Even though it seems to be a hassle, we always add an error id to Token.LogMessages so that each error has an identifier that we can link to. In this case its not that important, but its a good habit.
We could even go so far as having different message IDs for different errors.
I have rarely seen a user complain about having very distinct error messages.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -73,10 +84,84 @@ public static string Encode(string arg)
...
+            }
+
+            if (length == 0)
+            {
+                return string.Empty;
+            }
+
+            if (inArray.Length < offset)
+            {
+                throw LogHelper.LogArgumentException<ArgumentException>(nameof(offset), $"Invalid value for {nameof(offset)} parameter");
```
brentschmaltz: Its good practice to echo the value that was invalid.
So this message could read: "Invalid value for (nameof(offset)) parameter, was: '{offset}'."

Note we tend to put single quotes around such values as it helps identifying the value.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
cesarpayan: I wouldn't mind taking some time to add the LogMessages, is there an existing range I can make use of? or would I need to add a new one? In the case of the later, is there anything that needs to be documented?

-edit: Adding a generic ArgumentException message for now, if required I can add more specific message per case.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
cesarpayan: Only reason I can think of, is for the usage of custom exceptions that could inherit from ArgumentException. But there's no case of it and only found the method being called in one occasion in a different place. I will move away from it to use LogHelper.LogExceptionMessage(new ArgumentException(..... which upon second inspection is the more widely way of logging argument exceptions. 

Perhaps it's worth bringing up the LogArgumentException method up for discussion.

pmaytak:
Are there any performance tests for comparison before and after with improvements?

brentschmaltz (approved):

cesarpayan (commented):
> Are there any performance tests for comparison before and after with improvements?
@pmaytak Yes there are. 
This is an example of using the both encoders (updated and current) using BenchmarkDotNet to measure their times of executions. In this case we are encoding a few strings and joining them together. 
image

dannybtsai (approved):

henrik-me:
@bgavrilMS

bgavrilMS:
Thanks for doing this work @cesarpayan. It looks like Encoding operations are now ~2x faster and take 50% less memory (in MSAL at least). Decode remains unchanged. Pretty good win. 

MSAL PR + perf tests including memory: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2879/

Issue #1683, opened 2021-08-10
Leaking tasks that cause UTs to fail in IdentityModels 6.12.0.0


isaiah-mayerchak:
### -- Package + version number --

System.IdentityModel.Tokens.Jwt 6.12.0.0 (updated from 6.6.0.0)

### -- Behavior --

We were made aware before upgrading that we needed to upgrade to a version >= 6.12.0.0 due to an issue that caused hanging tasks in unit tests that was fixed by this PR: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1667/commits/a2430d1554c621024e59f78c13cba3c00594ae91

After upgrading to 6.12.0.0, a subset of our unit tests began to fail due to hanging tasks, and we traced the problem to a function of ours that creates a `System.IdentityModel.Tokens.Jwt.JwtSecurityToken`, creates a new `System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler` and then invokes `WriteToken()` on that handler using the token.

Skipping this code in our unit tests results in no hanging tasks, which given the presence of the issue mentioned in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1667/commits/a2430d1554c621024e59f78c13cba3c00594ae91 makes it seem very likely that it's the same issue or a closely related one.

My understanding given the description in the linked PR is that post-fix this is likely the result of a cache containing providers of some kind that when non-empty can result in this issue. What doesn't quite make sense to me is that when peeking into the code for `JwtSecurityTokenHandler` the only cache I can see is the `ITokenReplayCache` that comes inside `TokenValidationParameters` in `ValidateToken`, but I'm skeptical about how relevant that actually is to the problem.

I'm wondering if I can get some help from some folks more familiar with this codebase to understand what's potentially going on here.

brentschmaltz:
@isaiah-mayerchak we recognized this issue and are working on it.
@dannybtsai owns the solution.

dannybtsai:
A new version, 6.12.1, has been published and it should fix the "leaking task" issue. The main changes are to make the task exits as soon as it finishes processing events (plus a 2-min time to reduce potential overhead of restarting tasks too frequently). There are more improvements we plan to make but hopefully this will address the UT failures so you are not blocked.

PR #1702, opened 2021-09-19
Addressing the slow build times and associated test failures

mafurman:
Made some modifications to the logic of the EventBasedLRUCache in order to reduce test times:
- got rid of the _disposed bool (it was never set to true, so it was essentially the same as while(true))
- added StartEventQueueTaskIfNotRunning() to the Get and Remove APIs
- added a polling interval to avoid continuous execution of the task when the _eventQueue is empty
- the EventQueueTaskIdleTimeoutInSeconds has been changed for certain tests

We will need to run performance tests on these changes to ensure that they don't introduce any other issues.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
+++ b/src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
@@ -53,26 +53,34 @@ namespace Microsoft.IdentityModel.Tokens
         internal delegate void ItemRemoved(TValue Value);
 
         private readonly int _capacity;
+
         // The percentage of the cache to be removed when _maxCapacityPercentage is reached.
         private readonly double _compactionPercentage = .20;
         private LinkedList<LRUCacheItem<TKey, TValue>> _doubleLinkedList = new LinkedList<LRUCacheItem<TKey, TValue>>();
         private ConcurrentQueue<Action> _eventQueue = new();
+        private ConcurrentDictionary<TKey, LRUCacheItem<TKey, TValue>> _map;
+
+        // When the current cache size gets to this percentage of _capacity, _compactionPercentage% of the cache will be removed.
+        private readonly double _maxCapacityPercentage = .95;
+
+        // if true, expired values will not be added to the cache and clean-up of expired values will occur on a 5 minute interval
+        private readonly bool _removeExpiredValues;
+        private readonly int _removeExpiredValuesIntervalInSeconds;
+        private readonly TaskCreationOptions _options;
+
+        // timer that removes expired items periodically
+        private Timer _timer = null;
+
+        // for testing purpose only to verify the task count
+        private int _taskCount = 0;
 
         #region event queue
+  
+        private int _eventQueuePollingInterval = 50;
 
         // The idle timeout, the _eventQueueTask will end after being idle for the specified time interval (execution continues even if the queue is empty to reduce the task startup overhead), default to 120 seconds.
         // TODO: consider implementing a better algorithm that tracks and predicts the usage patterns and adjusts this value dynamically.
         private long _eventQueueTaskIdleTimeoutInSeconds = 120;
-        internal long EventQueueTaskIdleTimeoutInSeconds
-        {
-            get => _eventQueueTaskIdleTimeoutInSeconds;
-            set
-            {
-                if (value <= 0)
-                    throw new ArgumentOutOfRangeException(nameof(value), "EventQueueTaskExecutionTimeInSeconds must be positive.");
-                _eventQueueTaskIdleTimeoutInSeconds = value;
-            }
-        }
 
         // the event queue task
         private Task _eventQueueTask;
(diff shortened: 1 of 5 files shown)


src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -309,6 +325,8 @@ public bool TryRemove(TKey key, out TValue value)
                 return true;
             }
 
+            // start the event queue task if it is not running
+            StartEventQueueTaskIfNotRunning();
```
mafurman: Adding this to the Get() and Remove() APIs was necessary to avoid a build up of Actions  in the _eventQueue that were never processed (was causing memory issues).

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -353,12 +370,16 @@ private void EventQueueTaskAction()
                         // This should prevent a rare (but theoretically possible) scenario caused by context switching.
                         if (Interlocked.CompareExchange(ref _eventQueueTaskState, EventQueueTaskStopped, EventQueueTaskRunning) == EventQueueTaskRunning)
                             break;
+
+                    } else // if empty, let the thread sleep for a specified number of milliseconds before attempting to retrieve another value from the queue
+                    {
+                        Thread.Sleep(_eventQueuePollingInterval);
```
mafurman: Necessary to avoid continuous execution of the loop above when the queue is empty.

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -1130,11 +1130,11 @@ public void ProviderCache_EnsureNoException_MultipleThreads()
         /// Test to ensure no hanging task at the end when calling the JwtSecurityTokenHandler.WriteToken() method.
         /// The JwtHeader is created with SymmetricEncryptingCredentials.
         /// </summary>
-        [Fact (Skip = "test failing")]
+        [Fact]
         public void ProviderCache_EnsureNoLeakingTasks_SecurityTokenHandler_SymmetricEncryptingCredentials()
         {
-            InMemoryCryptoProviderCache cache = CryptoProviderFactory.Default.CryptoProviderCache as InMemoryCryptoProviderCache;
-            SetTaskIdleTimeoutInSeconds(cache, 10); // set the event queue task idle timeout
+            InMemoryCryptoProviderCache cache = new InMemoryCryptoProviderCache();
```
mafurman: This change was necessary in order to get these tests to start passing again, though I'm not sure about the implications of it.

@dannybtsai Do you know why the test was originally written the way it was, and why it might have started failing after the changes in this PR were made?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -331,15 +349,14 @@ private void RemoveItemFromLinkedList(LRUCacheItem<TKey, TValue> newCacheItem)
         private void EventQueueTaskAction()
         {
             Interlocked.Increment(ref _taskCount);
-
-            // Keep running until it is disposed, or expired (DateTime.UtcNow > _eventQueueTaskEndTime).
-            while (!_disposed)
+            try
```
dannybtsai: The try/catch should be moved inside of the while loop so if any of the actions throws an exception it will not exit the loop.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
dannybtsai: Good catch on starting the task anywhere new events are added.
Since we need to ensure the task will run whenever an event is added to the queue I suggest that we group the action and the StartEventQueueTaskIfNotRunning() together in a method to avoid possible mistakes in the future:

```
{
    _eventQueue.Enqueue(action);

    // start the event queue task if it is not running
    StartEventQueueTaskIfNotRunning();
}
```

And replace calls

```
_eventQueue.Enqueue(() =>
{
    _doubleLinkedList.Remove(cacheItem);
    _doubleLinkedList.AddFirst(cacheItem);
});

// start the event queue task if it is not running
StartEventQueueTaskIfNotRunning();
```

with
```
AddActionToEventQueue(() =>
{
    _doubleLinkedList.Remove(cacheItem);
    _doubleLinkedList.AddFirst(cacheItem);
});
```

The RemoveExpiredValuesPeriodically() needs to start the task as well:
```
protected void RemoveExpiredValuesPeriodically(object state)
{
    AddActionToEventQueue(() => RemoveExpiredValues());
}
```

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
dannybtsai: This test was originally written to cover the scenario that the Default CryptoProviderFactory will be referenced (and the event queue task will be started) when clients call the JwtSecurityTokenHandler.WriteToken() method.
(https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1683)

We need to use the CryptoProviderFactory.Default.CryptoProviderCache instead of creating a new InMemoryCryptoProviderCache as the Default.SymmetricEncryptingCredentials is referencing the CryptoProviderFactory.Default.

Both tests passed locally without your changes, did they fail on your machine?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -353,12 +370,16 @@ private void EventQueueTaskAction()
                         // This should prevent a rare (but theoretically possible) scenario caused by context switching.
                         if (Interlocked.CompareExchange(ref _eventQueueTaskState, EventQueueTaskStopped, EventQueueTaskRunning) == EventQueueTaskRunning)
                             break;
+
+                    } else // if empty, let the thread sleep for a specified number of milliseconds before attempting to retrieve another value from the queue
```
brentschmaltz: nit: put else on separate line.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Fixed!

Thanks for this suggestion Danny! I think this will really help future proof the code and prevent developer mistakes. :)

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Fixed!

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
mafurman: Done. :)

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
mafurman: That makes a lot of sense.... unfortunately these tests are still failing locally for me. 

Just kicked off a topic branch build with this change in place to make sure they pass on the build server.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -202,12 +273,16 @@ private void RemoveLRUs()
         }
 
         /// <summary>
-        /// The timer callback that adds a request to remove expired items from the event queue.
+        /// This is the method that determines the end time for the event queue task.
```
brentschmaltz, line 276: Above on line 271, could we write:
```
_doubleLinkedList.RemoveLast();
```

dannybtsai: And should the cleanUpIntervalInSeconds be renamed to removeExpiredValuesIntervalInSeconds as we discussed (line 144)?

mafurman: I believe the design decision to use Remove() rather than RemoveLast() was done to avoid multithreading issues where the list is somehow modified by another thread (not sure whether this is possible actually).

@dannybtsai That change should already be made. Can you not see it?

EDIT: Nevermind, just noticed that I only changed the private variable and not the method parameter. Fixed this now!

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
mafurman: Tests have been re-written to use a new instance of a CryptoProviderFactory; seems other tests were messing with CryptoProviderFactory.Default (adding signature providers, etc.) and that was causing issues.

PR #1704, opened 2021-09-21
Integrate ServerNonce validation

ciaozhang:


src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs
```diff
@@ -80,6 +80,16 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
...
 
+    /// <summary>
+    /// A delegate that will take control over SignedHttpRequest nonce validation, if set.
+    /// </summary>
+    /// <param name="key">the key use to validate server nonce.</param>
+    /// <param name="signedHttpRequest">A SignedHttpRequest.</param>
+    /// <param name="signedHttpRequestValidationContext">A structure that wraps parameters needed for SignedHttpRequest validation.</param>
+    /// <param name="cancellationToken">Propagates notification that operations should be canceled.</param>
+    /// <returns>Expected to throw an appropriate exception if SignedHttpRequest replay is detected.</returns>
+    public delegate bool NonceValidatorAsync(SecurityKey key, SecurityToken signedHttpRequest, SignedHttpRequestValidationContext signedHttpRequestValidationContext, CancellationToken cancellationToken);
```
ciaozhang, line 91: IIRC, Brent said it share the same parameter with Popkeyresolver, not sure the first parameter should be security key or security token.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -648,6 +651,31 @@ internal virtual async Task<SecurityKey> ValidateSignatureAsync(JsonWebToken sig
...
 
+        /// <summary>
+        /// Validates the nonce claim of the signed http request.
+        /// </summary>
+        /// <param name="signedHttpRequest">A SignedHttpRequest.</param>
+        /// <param name="popKey">A Pop key used to validate the signed http request nonce signature.</param>
+        /// <param name="signedHttpRequestValidationContext">A structure that wraps parameters needed for SignedHttpRequest validation.</param>
+        /// <param name="cancellationToken">Propagates notification that operations should be canceled.</param>
+        /// <returns>A PoP <see cref="SecurityKey"/> that validates signature of the <paramref name="signedHttpRequest"/>.</returns>
+        internal virtual bool ValidateNonceAsync(JsonWebToken signedHttpRequest, SecurityKey popKey, SignedHttpRequestValidationContext signedHttpRequestValidationContext, CancellationToken cancellationToken)
```
ciaozhang: Should it be async? And need clarify the return value and the exception.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Exceptions/SignedHttpRequestInvalidNonceClaimException.cs
```diff
@@ -0,0 +1,75 @@
...
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
+{
+    /// <summary>
+    /// This exception is thrown when a SignedHttpRequest handler encounters an error during 'Nonce' claim resolution.   
+    /// </summary>
+    [Serializable]
+    public class SignedHttpRequestInvalidNonceClaimException : SignedHttpRequestValidationException
```
TimHannMSFT, line 38: Can we add the error code to this as a string?

TimHannMSFT: alternatively, we could add a property bag

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -483,6 +483,9 @@ public async Task<SignedHttpRequestValidationResult> ValidateSignedHttpRequestAs
                 // validate signed http request signature
                 signedHttpRequest.SigningKey = await ValidateSignatureAsync(signedHttpRequest, popKey, signedHttpRequestValidationContext, cancellationToken).ConfigureAwait(false);
 
+                // validate nonce claim
+                _ = ValidateNonceAsync(signedHttpRequest, popKey, signedHttpRequestValidationContext, cancellationToken);
```
TimHannMSFT: if this returns false throw a generic SignedHttpRequestInvalidNonceClaimException, See what we do for the Validator delegates in TokenValidationParameters

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -648,6 +651,31 @@ internal virtual async Task<SecurityKey> ValidateSignatureAsync(JsonWebToken sig
...
         }
 
+        /// <summary>
+        /// Validates the nonce claim of the signed http request.
+        /// </summary>
+        /// <param name="signedHttpRequest">A SignedHttpRequest.</param>
+        /// <param name="popKey">A Pop key used to validate the signed http request nonce signature.</param>
+        /// <param name="signedHttpRequestValidationContext">A structure that wraps parameters needed for SignedHttpRequest validation.</param>
+        /// <param name="cancellationToken">Propagates notification that operations should be canceled.</param>
+        /// <returns>A PoP <see cref="SecurityKey"/> that validates signature of the <paramref name="signedHttpRequest"/>.</returns>
```
TimHannMSFT: update

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
TimHannMSFT: We probably don't want to ship an actual implementation of this. Just define what the delegate looks like, this signature looks good though.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
TimHannMSFT: This doesn't call a local method it should call the delegate on the validation parameters.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
TimHannMSFT: Whoops, misunderstood, this is okay so long as it just calls the delegate on the validationparams

TimHannMSFT: Ok, this can be return void and just throw if the delegate returns false.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -648,6 +651,31 @@ internal virtual async Task<SecurityKey> ValidateSignatureAsync(JsonWebToken sig
...
+        internal virtual bool ValidateNonceAsync(JsonWebToken signedHttpRequest, SecurityKey popKey, SignedHttpRequestValidationContext signedHttpRequestValidationContext, CancellationToken cancellationToken)
+        {
+            if (signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.NonceValidatorAsync != null)
+                return signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.NonceValidatorAsync(popKey, signedHttpRequest, signedHttpRequestValidationContext, cancellationToken);
```
TimHannMSFT: if this returns true do nothing, if it returns false, then we want to throw a SignedHttpRequestInvalidNonceClaimException

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs
TimHannMSFT: Good for @brentschmaltz to clarify, I think in some respects we could just have both available.

TimHannMSFT: This should also take in the nonce claim string. EDIT: okay if signedHttpRequest has this already probably not needed.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Exceptions/SignedHttpRequestInvalidNonceClaimException.cs
ciaozhang: Added property bag

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Exceptions/SignedHttpRequestInvalidNonceClaimException.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <param name="context">The contextual information about the source or destination.</param>
+        protected SignedHttpRequestInvalidNonceClaimException(SerializationInfo info, StreamingContext context)
+            : base(info, context)
+        {
+        }
+
+        /// <summary>
+        /// Gets or sets an <see cref="IDictionary{String, Object}"/> that enables custom extensibility scenarios.
+        /// </summary>
+        public IDictionary<string, object> PropertyBag { get; set; }
```
TimHannMSFT, line 79: What stops us from making this init-only or readonly and instantiate it with an empty dictionary? Then we could avoid the rule suppression.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -648,6 +651,27 @@ internal virtual async Task<SecurityKey> ValidateSignatureAsync(JsonWebToken sig
...
+        /// <param name="popKey">A Pop key used to validate the signed http request nonce signature.</param>
+        /// <param name="signedHttpRequestValidationContext">A structure that wraps parameters needed for SignedHttpRequest validation.</param>
+        /// <param name="cancellationToken">Propagates notification that operations should be canceled.</param>
+        internal virtual void ValidateNonceAsync(JsonWebToken signedHttpRequest, SecurityKey popKey, SignedHttpRequestValidationContext signedHttpRequestValidationContext, CancellationToken cancellationToken)
+        {
+            try
+            {
+                if (signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.NonceValidatorAsync != null)
+                    if (!signedHttpRequestValidationContext.SignedHttpRequestValidationParameters.NonceValidatorAsync(popKey, signedHttpRequest, signedHttpRequestValidationContext, cancellationToken))
+                        throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidNonceClaimException("Signed http request nonce validation failed."));
```
TimHannMSFT: nit: consider SignedHttpRequest

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs
```diff
@@ -80,6 +80,16 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
     /// <returns>Expected to throw an appropriate exception if SignedHttpRequest replay is detected.</returns>
     public delegate Task ReplayValidatorAsync(SecurityToken signedHttpRequest, SignedHttpRequestValidationContext signedHttpRequestValidationContext, CancellationToken cancellationToken);
 
+    /// <summary>
+    /// A delegate that will take control over SignedHttpRequest nonce validation, if set.
+    /// </summary>
+    /// <param name="key">the key use to validate server nonce.</param>
+    /// <param name="signedHttpRequest">A SignedHttpRequest.</param>
```
TimHannMSFT, line 87: nit: "A SignedHttpRequest which contains the 'nonce' claim to validate."

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
```diff
@@ -1683,8 +1683,17 @@ public static TheoryData<ValidateSignedHttpRequestTheoryData> ValidateSignedHttp
                             ValidatedSignedHttpRequest = signedHttpRequest,
                         },
                         TestId = "ValidTest",
+                    },
+                    new ValidateSignedHttpRequestTheoryData
```
TimHannMSFT, line 1687: Can we do one test that sets up a nonce string you expect and that validates being able to call signedHttpRequestToken.GetClaim("nonce") and is able to see the value and validate it?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/GlobalSuppressions.cs
```diff
@@ -11,3 +11,4 @@
 [assembly: SuppressMessage("Performance", "CA1825: Avoid zero-length array allocations", Justification = "net45 target doesn't support Array.Empty")]
 [assembly: SuppressMessage("Globalization", "CA1308:Normalize strings to uppercase", Justification = "Headers need to be lowercase to calcuate appropriate hash", Scope = "type", Target = "~T:Microsoft.IdentityModel.Protocols.SignedHttpRequest.SignedHttpRequestHandler")]
 [assembly: SuppressMessage("Design", "CA1001:Types that own disposable fields should be disposable", Justification = "Breaking change", Scope = "type", Target = "~T:Microsoft.IdentityModel.Protocols.SignedHttpRequest.SignedHttpRequestHandler")]
+[assembly: SuppressMessage("Usage", "CA2227:Collection properties should be read only", Justification = "<Pending>", Scope = "member", Target = "~P:Microsoft.IdentityModel.Protocols.SignedHttpRequest.SignedHttpRequestInvalidNonceClaimException.PropertyBag")]
```
TimHannMSFT, line 14: Change the justification to be "Want to be consistent with other PropertyBag" properties in the library

TimHannMSFT: @GeoK @mafurman @RojaEnnam - Any objections with this? I think in the perfect world we'd be able to follow this across the library but since all the other property bags are not readonly, it would be pretty confusing to have this one be a good citizen.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
```diff
@@ -1683,8 +1683,45 @@ public static TheoryData<ValidateSignedHttpRequestTheoryData> ValidateSignedHttp
...
+                        SignedHttpRequestValidationParameters = new SignedHttpRequestValidationParameters()
+                        {
+                            ValidateB = false,
+                            ValidateH = false,
+                            ValidateM = false,
+                            ValidateP = false,
+                            ValidateQ = false,
+                            ValidateTs = false,
+                            ValidateU = false,
+                            NonceValidatorAsync =  (popKey, signedHttpRequestToken, signedHttpRequestValidationContext, cancellationToken) => true
```
TimHannMSFT: This delegate just always returns true, it's nice to have this high level positive case but I think what would be even more helpful and most helpful to Substrate as a sample is showing how you grab the nonce claim off of the signedHttpRequestToken you can set it up to be anything in the test but the body of the delegate should be something like return signedHttpRequestToken.GetClaim... == expectedValueSetupInTest

TimHannMSFT (approved):
Would still like to see the unit test with a delegate validating the nonce value but everything else LGTM. Thanks for the change Xiao

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
ciaozhang: Modify the test case

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/GlobalSuppressions.cs
ciaozhang: As discussed, agree with being consistent with other PropertyBag.

TimHannMSFT (approved):

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs
brentschmaltz: SecurityKey was correct.

PR #1705, opened 2021-09-22
Updating comments to use inclusive terms

### LINKED ISSUE #1703: Inclusive language in code
To promote the usage of inclusive language, use the extension (or NuGet Package) [Inclusiveness Analyzer ](https://github.com/merill/InclusivenessAnalyzer) for the source code to avoid the use of terms that can be offensive and/or exclusive

cesarpayan:
Updating comments to use inclusive terms

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
@@ -247,7 +247,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// Gets or sets a delegate that will be used to validate the audience.
         /// </summary>
         /// <remarks>
-        /// If set, this delegate will be called to validate the 'audience', instead of typical processing.
+        /// If set, this delegate will be called to validate the 'audience', instead of default processing.
         /// This means that no default 'audience' validation will occur.
         /// Even if <see cref="ValidateAudience"/> is false, this delegate will still be called.
         /// </remarks>
@@ -370,7 +370,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// Gets or sets a delegate for validating the <see cref="SecurityKey"/> that signed the token.
         /// </summary>
         /// <remarks>
-        /// If set, this delegate will be called to validate the <see cref="SecurityKey"/> that signed the token, instead of typical processing.
+        /// If set, this delegate will be called to validate the <see cref="SecurityKey"/> that signed the token, instead of default processing.
         /// This means that no default <see cref="SecurityKey"/> validation will occur.
         /// Even if <see cref="ValidateIssuerSigningKey"/> is false, this delegate will still be called.
         /// </remarks>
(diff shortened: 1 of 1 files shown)


src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -559,7 +564,7 @@ public string RoleClaimType
         /// Implementations that need to resolve it from a different location can use the 'token' parameter.
         /// </summary>
         /// <remarks>
-        /// If set, this delegate will be called to validate the 'type' of the token, instead of normal processing.
+        /// If set, this delegate will be called to validate the 'type' of the token, instead of typical processing.
```
TimHannMSFT: nit: "default" for consistency?

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -405,7 +408,8 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
         /// Gets or sets a delegate that will be used to validate the lifetime of the token
         /// </summary>
         /// <remarks>
-        /// If set, this delegate will be called to validate the lifetime of the token, instead of normal processing. This means that no default lifetime validation will occur.
+        /// If set, this delegate will be called to validate the lifetime of the token, instead of typical processing.
+        /// This means that no default lifetime validation will occur.
```
TimHannMSFT, line 412: For my own edification, is the update here to avoid long lines you have to scroll to see? If so hooray, I'm very in favor of this. Is that an accessibility improvement or is there another inclusive angle I'm missing?

TimHannMSFT (approved):
LGTM, just one nit about typical vs default for consistency but certainly not blocking and there might be a good reason I'm not seeing. Thanks for the change.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
cesarpayan: I am changing a word (which is part of the inclusiveness analyzer, removing the word 'normal' for typical/default in this case. Other inclusive terms that could be used are excepted or healthy, depending on the context), and figured since I was editing these lines might as well make them more readable since the line was quite long.

There's no accessibility angle that Inclusive analyzer can give, other than terms that might be exclusive towards people with disabilities.

TimHannMSFT: Ah okay, the diff highlighter just didn't highlight that word change in the above and my eyeballing failed me. makes sense.

ciaozhang (approved):

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
cesarpayan: Fair, switching to default.

PR #1709, opened 2021-09-29
Remove incorrect log message when an issuer is an empty string

### LINKED ISSUE #1673: IssuerValidator logs incorrect message when an issuer is an empty string
When we iterate through ValidIssuers on the TokenValidationParameters and discover an issuer that is an empty string we throw the following incorrect exception:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/48d02044accd936ec1749ca8f5d349a5a9c5e6df/src/Microsoft.IdentityModel.Tokens/Validators.cs#L242
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/48d02044accd936ec1749ca8f5d349a5a9c5e6df/src/Microsoft.IdentityModel.Tokens/LogMessages.cs#L71

ciaozhang:
Close #1673

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validators.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validators.cs
@@ -238,10 +238,7 @@ namespace Microsoft.IdentityModel.Tokens
                 foreach (string str in validationParameters.ValidIssuers)
                 {
                     if (string.IsNullOrEmpty(str))
-                    {
-                        LogHelper.LogInformation(LogMessages.IDX10235);
                         continue;
-                    }
                         
                     if (string.Equals(str, issuer, StringComparison.Ordinal))
                     {


src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -238,10 +238,7 @@ public static string ValidateIssuer(string issuer, SecurityToken securityToken,
                 foreach (string str in validationParameters.ValidIssuers)
                 {
                     if (string.IsNullOrEmpty(str))
-                    {
-                        LogHelper.LogInformation(LogMessages.IDX10235);
```
TimHannMSFT, line 242: Should we still have something logged in this case (this string just seems like the wrong string).

This does seem like something that indicates that something odd is happening, if there are empty strings in the ValidIssuers collection something has gone awry correct?

ciaozhang: IMHO, we just need the valid issuer is logged here.

jennyf19: Looks like that error message gets returned on line 216. If this is actionable for the developer (ex. they need to set a property, etc..), if so, I would definitely return something that says what the action should be. If not, it's probably fine to not log anything, or just log that the issuer was null or empty for debugging purposes.

brentschmaltz: @ciaozhang i agree with Tim having a null or empty string is unusual.
The log message should say something like: "one of the issuers in TokenValidationParameters.ValidIssuers was null or the empty string."

PR #1713, opened 2021-10-05
Cesar/config fix

### LINKED ISSUE #1712: Use ConfigurationManager<T> without T being of type BaseConfiguration
Currently if a customer were to use ConfigurationManager with a custom class, ConfigurationManager.GetBaseConfigurationAsync ends up throwing an exception because code is assuming the class T is of type BaseConfiguration (or derived). 

This limits consumers' design implementation, and there's no reason for doing so.

cesarpayan:
Current design doesn't allow customers to use ConfigurationManager<T> when T is a custom class. ConfigurationManager.GetBaseConfigurationAsync ends up throwing an exception because it is expecting that T is of type BaseConfiguration (or derived from)

This PR separates that dependency from BaseConfiguration to the templated class.
Introducing unit tests to verify that ConfigurationManager can be used with a custom class as T and a custom IConfigurationRetriever implementation that are not related to classes in our design.
There was a case where we still needed a BaseConfiguration GetBaseConfigurationAsync implementation which prevented the idea of completely removing this method from BaseConfigurationManager.


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -180,8 +193,8 @@ internal override async Task<BaseConfiguration> GetBaseConfigurationAsync(Cancel
 
         /// <summary>
         /// Requests that then next call to <see cref="GetConfigurationAsync()"/> obtain new configuration.
-        /// <para>If it is a first force refresh or the last refresh was greater than <see cref="BaseConfigurationManager.RefreshInterval"/> then the next call to <see cref="GetConfigurationAsync()"/> will retrieve new configuration.</para>
-        /// <para>If <see cref="BaseConfigurationManager.RefreshInterval"/> == <see cref="TimeSpan.MaxValue"/> then this method does nothing.</para>
+        /// <para>If it is a first force refresh or the last refresh was greater than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then the next call to <see cref="GetConfigurationAsync()"/> will retrieve new configuration.</para>
+        /// <para>If <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> == <see cref="TimeSpan.MaxValue"/> then this method does nothing.</para>
```
GeoK: Why this change? RequestRefresh methods seems to still depend on RefreshInterval.

test/Microsoft.IdentityModel.Protocols.Tests/ExtensibilityTests.cs
```diff
@@ -0,0 +1,160 @@
...
+    {
+        /// <summary>
+        /// Issuer associated with the OIDC endpoint.
+        /// </summary>
+        public string Issuer { get; set; }
+    }
+
+    public class ExtensibilityTests
+    {
+
```
GeoK: nit: extra new line

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -123,21 +124,35 @@ public async Task<T> GetConfigurationAsync()
...
+            if (obj is BaseConfiguration)
+                return obj as BaseConfiguration;
+            return null;
+        }
+
+        /// <summary>
+        /// Obtains an updated version of Configuration.
+        /// </summary>
+        /// <param name="cancel">CancellationToken</param>
+        /// <returns>Configuration of type StandardConfiguration.</returns>
```
TimHannMSFT: Seems like the returns documentation for this and the above are swapped (or almost swapped, presumably it should mention BaseConfiguration instead?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -149,28 +164,26 @@ internal override async Task<BaseConfiguration> GetBaseConfigurationAsync(Cancel
                     {
                         // Don't use the individual CT here, this is a shared operation that shouldn't be affected by an individual's cancellation.
                         // The transport should have it's own timeouts, etc..
-                        CurrentConfiguration = await _configRetriever.GetConfigurationAsync(MetadataAddress, _docRetriever, CancellationToken.None).ConfigureAwait(false) as BaseConfiguration;
-                        Contract.Assert(CurrentConfiguration != null);
+                        _currentConfiguration = await _configRetriever.GetConfigurationAsync(MetadataAddress, _docRetriever, CancellationToken.None).ConfigureAwait(false);
+                        Contract.Assert(_currentConfiguration != null);
```
TimHannMSFT: In my brief review, it looks like this GetConfigurationAsync won't return null. Now that the as BaseConfig cast is removed, is this still necessary? Seems like it at least has different semantics

TimHannMSFT (approved):
Generally LGTM, I didn't review the tests super closely or validate that they'd help catch the issue we're preventing in the future. Is the plan to do another perf/stress run with this build or just to do the build with MSAL prior to release?

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -359,8 +359,6 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
         /// It will be used along with the other properties set on the TokenValidationParameters in order to validate the token.
         /// </summary>
         /// <remarks>
-        /// This property can be set to the <see cref="BaseConfigurationManager.CurrentConfiguration"/> or the <see cref="BaseConfigurationManager.LastKnownGoodConfiguration"/>
```
dannybtsai, line 362: Why removing the comments?

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -57,12 +57,7 @@ public TimeSpan AutomaticRefreshInterval
         }
 
         /// <summary>
-        /// The most recently retrieved configuration.
-        /// </summary>
-        internal BaseConfiguration CurrentConfiguration { get; set; }
-
-        /// <summary>
-        /// 12 hours is the default time interval that afterwards, <see cref="GetBaseConfigurationAsync(CancellationToken)"/> will obtain new configuration.
+        /// 12 hours is the default time interval that afterwards will obtain new configuration.
```
dannybtsai, line 60: With the changes, why do we still need the BaseConfigurationManager class? is there a reason the LastKnownGoodConfiguration can not be in the ConfigurationManager<BaseConfiguration>?

mafurman (approved):

PR #1718, opened 2021-10-07
Has is valid or exception been read

### LINKED ISSUE #1717: Improve Logging when the SDK is used unsafely
When claims are getting evaluated we should be able to know if IsValid or Exception were read. If they were not, that could indicate an unsafe usage of the API and we should log.

cesarpayan:
When claims are getting evaluated we should be able to know if IsValid or Exception were read. If they were not, that could indicate an unsafe usage of the API and we should log.

Introducing private variables for properties IsValid and Exception so they can have a body in their getters
Introducing a private variable to track whether or not IsValid or Exception have been read
Bodies to IsValid and Exception to set new private variable
Getter body to Claims so we can check if IsValid or Exception have been read, and log if they have not.
New warning message to describe the situation
Unit Tests to test the described behavior above
Moved SampleLogger to Utils so it can be easier to access across different unit tests when test depends on log reads

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
@@ -50,9 +50,7 @@ namespace Microsoft.IdentityModel.Tokens
             get
             {
                 if (!_hasIsValidBeenRead)
-                {
                     LogHelper.LogWarning(LogMessages.IDX10109);
-                }
 
                 return _claims.Value;
             }
@@ -61,7 +59,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// <summary>
         /// The <see cref="ClaimsIdentity"/> created from the validated security token.
         /// </summary>
-        public ClaimsIdentity ClaimsIdentity { set; get; }
+        public ClaimsIdentity ClaimsIdentity { get; set; }
 
         /// <summary>
         /// Gets or sets the <see cref="Exception"/> that occurred during validation.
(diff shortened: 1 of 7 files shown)


src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
```diff
@@ -67,7 +86,19 @@ public class TokenValidationResult
...
+        public bool IsValid
+        {
+            get
+            {
+                _hasIsValidOrExceptionBeenRead = true;
+                return _isValid;
+            }
+            set
+            {
+                _hasIsValidOrExceptionBeenRead = false;
```
TimHannMSFT: Couple Questions for my knowledge:
1. Why does setting reset in the IsValid case but not the Exception case?
2. Are there scenarios (in our code base at least) where people are actually using the setters here? I'd suspect the setters were only exposed here by accident and remain to avoid breaking changes. Maybe there's a weird historical reason for the setters but I think if folks are changing the value of IsValid or Exception we're in trouble too (but that's outside the scope here).

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -96,6 +96,107 @@ public void ValidateTokenValidationResult()
...
+
+            //check if warning message was logged
+            var warningId = "IDX10109";
+            Assert.Contains(warningId, listener.TraceBuffer);
+        }
+
+        [Fact]
+        public void ValidateTokenValidationResultDoesNotThrowWarningWithIsValidRead()
+        {
+            //create a listener and enable it for logs
```
TimHannMSFT, line 123: Consider refactoring this block out into a method since it's duplicated

TimHannMSFT (approved):

PR #1720, opened 2021-10-08
Add const for the standard scope "phone"

vanillajonathan:
The "phone" scope is a standard scope part of the OpenID specification.
See: https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims

> phone
    OPTIONAL. This scope value requests access to the `phone_number` and `phone_number_verified` Claims.


jennyf19 (approved):
LGTM...thoughts @brentschmaltz ?

mafurman (approved):

jennyf19:
thanks for the contribution @vanillajonathan

PR #1723, opened 2021-10-13
Adding support for the Last Known Good Configuration feature

mafurman:
Notes on the changes made in this PR:
- ValidateToken method has been restructured as follows: 
    (1) if a configuration is present we will try to validate using it
    (2) if validation fails we will try the LKG (if present)
    (3) regardless of whether the LKG is present, we will attempt to RequestRefresh() and revalidate ONLY if we were successfully able to get a new configuration.
 - New properties have been added to the BaseConfigurationManager, including:
    - UseLastKnownGoodConfiguration - a flag indicating whether this feature is on or off, true by default (open to hearing feedback on this though)
    - IsLastKnownGoodValid - indicates whether or not the lifetime of the LKG has passed
    - LastKnownGoodLifetime - indicates how long after first use we should continue to use a LKG
   - LastKnownGoodConfiguration - the most recent configuration that was successfully able to validate a token in the past
- Test cases are intended to capture all the different LKG + RequestRefresh() fallback scenarios; would appreciate a second look at these to ensure all possible code paths are covered.
- TODO comments are intentionally left in the code. I can address them in this PR or in a new one, but I wanted to make sure I got feedback on this feature before refactoring the code further. Tasks have been created to keep track of these future changes.


src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -62,27 +65,59 @@ public TimeSpan AutomaticRefreshInterval
...
+        {
+            get
+            {
+                // only set this value the first time the last known good configuration is used for validation
+                if (_lastKnownGoodConfigFirstUse.Equals(DateTime.MinValue)) _lastKnownGoodConfigFirstUse = DateTime.UtcNow;
+                return _lastKnownGoodConfiguration;
+            }
+            set
+            {
+                _lastKnownGoodConfiguration = value != null ? value : throw LogHelper.LogArgumentNullException(nameof(value));
```
dannybtsai: nit: use the coalesce ?? if you like:
_lastKnownGoodConfiguration = value ?? throw LogHelper.LogArgumentNullException(nameof(value));

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -62,27 +65,59 @@ public TimeSpan AutomaticRefreshInterval
...
+
+        /// <summary>
+        /// The length of time that a last known good configuration is valid for.
+        /// </summary>
+        public TimeSpan LastKnownGoodLifetime
+        {
+            get { return lastKnownGoodLifetime; }
+            set
+            {
+                if (value < TimeSpan.Zero)
```
dannybtsai, line 119: Is TimeSpan.Zero valid?
`if (value <= TimeSpan.Zero)`

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -40,6 +40,9 @@ public abstract class BaseConfigurationManager
     {
         private TimeSpan _automaticRefreshInterval = DefaultAutomaticRefreshInterval;
         private TimeSpan _refreshInterval = DefaultRefreshInterval;
+        private TimeSpan lastKnownGoodLifetime = DefaultLastKnownGoodConfigurationLifetime;
```
dannybtsai: nit: rename to `_lastKnownGoodLifetime `to be consistent.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -62,27 +65,59 @@ public TimeSpan AutomaticRefreshInterval
...
+
+        /// <summary>
+        /// The last known good configuration or LKG (a configuration retrieved in the past that we were able to successfully validate a token against).
+        /// </summary>
+        public BaseConfiguration LastKnownGoodConfiguration
+        {
+            get
+            {
+                // only set this value the first time the last known good configuration is used for validation
+                if (_lastKnownGoodConfigFirstUse.Equals(DateTime.MinValue)) _lastKnownGoodConfigFirstUse = DateTime.UtcNow;
```
GeoK: nit: consider reformatting this line

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -988,29 +1034,87 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
+                        validationParametersCopy.ConfigurationManager.LastKnownGoodConfiguration = validationParametersCopy.Configuration;
+
+                    return tokenValidationResult;
                 }
-                else
+                else if (tokenValidationResult.Exception.GetType().Equals(typeof(SecurityTokenInvalidSignatureException))
+                   || tokenValidationResult.Exception.GetType().Equals(typeof(SecurityTokenInvalidSigningKeyException))
+                   || tokenValidationResult.Exception.GetType().Equals(typeof(SecurityTokenInvalidIssuerException))
+                   || tokenValidationResult.Exception.GetType().Equals(typeof(SecurityTokenUnableToValidateException))
+                   || tokenValidationResult.Exception.GetType().Equals(typeof(SecurityTokenSignatureKeyNotFoundException)))
```
GeoK, line 1037: Do we react on 'SecurityTokenInvalidIssuerException' because issuer in metadata might be corrupted?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -960,17 +960,63 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
                     validationParametersCopy.Configuration = configuration;
                 }
                 catch (Exception ex)
                 {
+                    // TODO: Get rid of this check and continue with validation regardless of the properties set on the TVP.
+                    // It's too hard to capture everything in this check, we should just fail later on in the validation process.
+
                     // If either a signing key or a valid issuer was not provided through the TVP, we want to stop processing on configuration retrieval failure.
                     // Otherwise, we should keep going with validation.
                     if ((validationParametersCopy.ValidateIssuer && string.IsNullOrWhiteSpace(validationParametersCopy.ValidIssuer) && validationParametersCopy.ValidIssuers.IsNullOrEmpty() && validationParametersCopy.IssuerValidator == null
```
GeoK: I realize that this is existing code but not sure that I understand this condition, primarily around decryption - we might be validating a JWS on this path, ValidateIssuer might be true, and TokenDecryptionKey might be set, but no ValidIssuer(s) or IssuerValidator. In that case the condition will result as 'false' even though we actually should stop the execution and return back the result.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -988,29 +1034,87 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
-                    return new TokenValidationResult
-                    {
-                        SecurityToken = jwtToken,
-                        ClaimsIdentity = innerTokenValidationResult.ClaimsIdentity,
-                        IsValid = true,
-                        TokenType = innerTokenValidationResult.TokenType
-                    };
+                    // Set current configuration as LKG if it exists.
+                    if (validationParametersCopy.Configuration != null && validationParametersCopy.Configuration != validationParametersCopy.ConfigurationManager.LastKnownGoodConfiguration)
+                        validationParametersCopy.ConfigurationManager.LastKnownGoodConfiguration = validationParametersCopy.Configuration;
```
GeoK, line 1027: What are the guarantees that 'validationParametersCopy.Configuration' is not corrupted (but we can get and parse it successfully)? Token validation could succeed in a case where only existing parameters on TVP are used (issuer/signing keys), and not the ones from TVP.Configuration.
Is the expectation that if ConfigurationManager is set on TVP, developers shouldn't be populating ValidIssuer(s)/SigningKeys?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -988,29 +1034,87 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
+                            tokenValidationResult = decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParametersCopy) : ValidateJWS(token, validationParametersCopy); ;
+
+                            if (tokenValidationResult.IsValid) return tokenValidationResult;
+                        }
+                    }
+
+                    // If we were still unable to validate, attempt to refresh the configuration and validate using it.
+                    validationParametersCopy.ConfigurationManager.RequestRefresh();
+                    var lastConfig = validationParametersCopy.Configuration;
+                    validationParametersCopy.Configuration = validationParametersCopy.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).Result;
```
GeoK: "Sync over async" is always a code smell / bad pattern and it can lead to thread-pool starvation and deadlocks. Ideally we would have a complete async stack from top-down. 
When it's time to refresh, we might have multiple threads waiting on this lock. As we use "sync over async", number of threads used is actually double. 
Would like to see some stress tests here.

GeoK: As a side not, using _GetAwaiter().GetResult()_ might be preferable as _.Result_ wraps the exception and throws AggregateException.

GeoK: As the final note here, should we surround this call with try/catch?

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
mafurman: Whoops, thanks for picking up on this!

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
mafurman: I was thinking about this earlier, and I decided to leave this as ` if (value < TimeSpan.Zero)` on the off chance that someone wanted to only use the LKG a single time. 

If it's set to 0, it means that the LKG can only be used once.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
mafurman: Good catch; just fixed this. :)

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
mafurman: Reformatted this to make it clearer by not trying to keep it all on the same line.
Also modified _lastKnownGoodConfigFirstUse to be a nullable Datetime, since DateTime.MinValue didn't feel like the right default for it.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Yes, that was the idea behind falling back to the LKG/requesting refresh for this particular exception.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Yeah, I think the logic in this check isn't quite right. I was going to leave this for a future PR, but I've gone ahead and removed the check since I do think it's too difficult to capture all the relevant scenarios.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: I believe the question here is how do we know the required validation artifacts came from the Configuration and not properties set on the TVP? 

I had this discussion with Brent back when we agreed on the design for this, and we decided that trying to separate out which keys came from the Configuration vs. the TVP would be more complicated than it's worth. So currently the LKG is set if validation succeeds, regardless of exactly how this came about (more of a heuristic). 

Do you think this inexact approach would lead to problems for developers?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: I agree with your first statement. I originally wanted to make this an async method, but I figured an async ValidateToken(..) method would confuse developers as it would imply that we have async verifying operations. The idea would be to make this call async once we got around to creating a truly async method for validation.

I have modified the code to use GetAwaiter().GetResult(): thank you for catching this!

This call is actually surrounded by a try catch one layer above (in the method that calls into this one).

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -97,7 +98,8 @@ internal static class LogMessages
         public const string IDX10257 = "IDX10257: Token type validation failed. Type: '{0}'. Did not match: validationParameters.TokenTypes: '{1}'.";
         public const string IDX10258 = "IDX10258: Token type validated. Type: '{0}'.";
         public const string IDX10260 = "IDX10260: Issuer validation failed. Issuer: '{0}'. Did not match: validationParameters.ValidIssuer: '{1}' or validationParameters.ValidIssuers: '{2}' or validationParameters.ConfigurationManager.CurrentConfiguration.Issuer: '{3}'.";
-        public const string IDX10261 = "IDX10261: Unable to retrieve configuration. \nProceeding with token validation as the relevant properties have been set manually on the TokenValidationParameters.";
+        public const string IDX10261 = "IDX10261: Unable to retrieve configuration from authority: '{0}'. \nProceeding with token validation as the relevant properties have been set manually on the TokenValidationParameters. Exception caught: \n {1}.";
+        public const string IDX10262 = "IDX10262: Unable to re-validate with ConfigurationManager.LastKnownGoodConfiguration as it is expired.";
```
jennyf19: is there anything the developer can do here to resolve this issue? If so, maybe an aka.ms link with additional steps would be useful.

jennyf19 (approved):
:shipit:

jennyf19:
LGTM on my limited understanding of the feature based on what you've presented to the team. I would leave most of the comments, as I think they are pretty helpful.

---
In reply to: [793181229](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1723#issuecomment-793181229) [](http://example.com/codeflow?ancestors=793181229,793181229)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: @GeoK the expectation is that users of our runtime will set Configuration and only in special cases set additional values.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -960,56 +960,148 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
             if (tokenParts.Length != JwtConstants.JwsSegmentCount && tokenParts.Length != JwtConstants.JweSegmentCount)
                 return new TokenValidationResult { Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14111, token))), IsValid = false };
 
+
```
brentschmaltz: nit: extra line

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -960,56 +960,148 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
+            }
+            catch (Exception ex)
+            {
+                return new TokenValidationResult
+                {
+                    Exception = ex,
+                    IsValid = false
+                };
+            }
+
```
brentschmaltz: extra line

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -960,56 +960,148 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
+        ///  it.
+        /// </summary>
+        /// <param name="token">The JWS string, only to be supplied if the token is a JWS.</param>
+        /// <param name="outerToken">The outer token of the JWE, only to be supplied if the token is a JWE.</param>
+        /// <param name="decryptedJwt">The inner decrypted token of the JWE, only to be supplied if the token is a JWE.</param>
+        /// <param name="validationParameters">The <see cref="TokenValidationParameters"/> to be used for validation.</param>
+        /// <returns></returns>
+        private TokenValidationResult ValidateToken(string token, JsonWebToken outerToken, string decryptedJwt, TokenValidationParameters validationParameters)
+        {
+
```
brentschmaltz: extra line

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -960,56 +960,148 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
+                {
+                    Exception = ex,
+                    IsValid = false
+                };
+            }
+
+        }
+
+        /// <summary>
+        ///  Private method for token validation, responsible for (1) obtaining a configuration from the <see cref="TokenValidationParameters.ConfigurationManager"/>,
```
brentschmaltz: If we split this comment out on separate lines it will be easier to read.
... responsible for:
1) ...
2) ...

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -960,56 +960,148 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
-                        && validationParametersCopy.TokenDecryptionKey == null && validationParametersCopy.TokenDecryptionKeys.IsNullOrEmpty() && validationParametersCopy.TokenDecryptionKeyResolver == null)
-                        || (validationParametersCopy.RequireSignedTokens && validationParametersCopy.IssuerSigningKey == null && validationParametersCopy.IssuerSigningKeys.IsNullOrEmpty() && validationParametersCopy.IssuerSigningKeyResolver == null))
+                    // Keep going with the validation as the TokenValidationParameters may have the issuer and signing key set
+                    // directly on them.
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(TokenLogMessages.IDX10261, validationParametersCopy.ConfigurationManager.MetadataAddress, ex.ToString()));
+                }
+            }
+
+            TokenValidationResult tokenValidationResult = decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParametersCopy) : ValidateJWS(token, validationParametersCopy);
+
```
brentschmaltz: extra line

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -960,56 +960,148 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
+                            LogHelper.LogInformation(TokenLogMessages.IDX10262);
+                        else
                         {
-                            Exception = ex,
-                            IsValid = false
-                        };
+                            validationParametersCopy.Configuration = validationParametersCopy.ConfigurationManager.LastKnownGoodConfiguration;
+                            tokenValidationResult = decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParametersCopy) : ValidateJWS(token, validationParametersCopy); ;
+
+                            if (tokenValidationResult.IsValid) return tokenValidationResult;
```
brentschmaltz: it would be easier to read if 'return tokenValidationResult' on separate line.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -960,56 +960,148 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
+        private TokenValidationResult ValidateToken(string token, JsonWebToken outerToken, string decryptedJwt, TokenValidationParameters validationParameters)
+        {
+
             var validationParametersCopy = validationParameters;
-            if (validationParameters.ConfigurationManager != null)
+            if (validationParametersCopy.ConfigurationManager != null)
             {
                 try
                 {
+                    // TODO: Use overloads to get rid of the Clone() operation.
```
brentschmaltz, line 1007: Once this PR is comitted, we will rework this method after removing the 'Clone'.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -50,6 +50,7 @@ internal static class LogMessages
         public const string IDX10107 = "IDX10107: When setting RefreshInterval, the value must be greater than MinimumRefreshInterval: '{0}'. value: '{1}'.";
         public const string IDX10108 = "IDX10108: When setting AutomaticRefreshInterval, the value must be greater than MinimumAutomaticRefreshInterval: '{0}'. value: '{1}'.";
         public const string IDX10109 = "IDX10109: Warning: Claims is being accessed without first reading the properties TokenValidationResult.IsValid or TokenValidationResult.Exception. This could be a potential security issue.";
+        public const string IDX10110 = "IDX10110: When setting LastKnownGoodLifetime, the value must be greater than zero. value: '{0}'.";
```
cesarpayan: nit:
`the value must be greater than zero.`

But code allows for a zero value
`if (value < TimeSpan.Zero)`

should it not be greater or equal than zero?

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -302,6 +301,41 @@ public void GetConfiguration()
...
+
+            // set and retrieve config again to reset first access time
+            configurationManager.LastKnownGoodConfiguration = validConfig;
+            lkg = configurationManager.LastKnownGoodConfiguration;
+            var lkgConfigFirstUse2 = lkgConfigFirstUseField.GetValue(configurationManager as BaseConfigurationManager);
+
+            if (lkgConfigFirstUse1 == null)
+                context.AddDiff("Last known good first use time was not properly set for the first configuration.");
+
+            if (lkgConfigFirstUse1 == null)
```
cesarpayan: Should it not be 
`if (lkgConfigFirstUse2 == null)`
?

note: the number at the end of the variable

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -302,6 +301,41 @@ public void GetConfiguration()
...
+                context.AddDiff("Last known good first use time was not properly set for the second configuration.");
+
+            //LKG config first use was not reset when a new configuration was set
+            if (lkgConfigFirstUse1 == lkgConfigFirstUse2)
```
cesarpayan: question: Thinking from a position of how the test can fail. I don't think this can be ever be an unsuccessful condition (as in actually go to the context.AddDiff line). If we were to comment out lines
`            configurationManager.LastKnownGoodConfiguration = validConfig;
            lkg = configurationManager.LastKnownGoodConfiguration;`

I would think that we're getting the same configuration twice, yet because the references are different this condition would still see them as different.

Is this intended? if not, it might be better to actually change a value in the configuration and compare those values to determine whether a different configuration was obtained or not.

nit: Additionally, imo checking if the configs aren't null would make the test more explicit/descriptive when it fails, as in if one of them is null, getting a failure saying that the config wasn't reset won't add much value.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
mafurman: I think that's a really good idea! I think there's probably a few troubleshooting steps that could be done for both issues; I've created a work item to track this as I think it makes more sense to add this once the feature is fully complete. :)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Thanks; fixed. :) https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1723/commits/bd148b61e7f6bd2a1967470437dcb020e6fe14df

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
mafurman: You're right; good catch! https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1723/commits/bd148b61e7f6bd2a1967470437dcb020e6fe14df

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
mafurman: Ah, my bad. Copy paste error. 😳 Fixed! https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1723/commits/bd148b61e7f6bd2a1967470437dcb020e6fe14df

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
mafurman: Ah, good catch! I forgot that a DateTime in this language is a value type and not a reference type.

You're right that commenting out those lines does not result in failure, but changing this to make use of Equals() (combined with sleeping for 1 ms) addresses that issue. Seems like even with the use of Equals() the operations happen too fast and so the first use time ends up being the same for both. Not a fan of using Thread.Sleep(), so I'll try to think of other ways to test this particular feature. https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1723/commits/bd148b61e7f6bd2a1967470437dcb020e6fe14df

The way the code is currently written, we do not check if the value that we're setting for the LKG is the same as the value that is currently set (as it costs more to figure this out than to reset this). The aim of this test isn't to check if a different configuration was obtained, but to make sure that the firstUse property is properly reset when a new LKG is set, as it is used to determine how long an LKG can safely be used for (whether it is valid). Does that make sense?

brentschmaltz (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Yep, that will be one of the first PRs going in after this one. :)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Fixed, thanks. :) https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1723/commits/84cda78d64292c88f8bd51d42254d8bd9a272c24

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Fixed! https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1723/commits/84cda78d64292c88f8bd51d42254d8bd9a272c24

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Good suggestion! Addressed here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1723/commits/84cda78d64292c88f8bd51d42254d8bd9a272c24

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Fixed; more consistent with our code style anyway. https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1723/commits/84cda78d64292c88f8bd51d42254d8bd9a272c24

PR #1725, opened 2021-10-14
Sruthi/improve pii logging

sruke:
Context
User can turn off logging Personally Identifiable Information (PII) by setting a boolean (ShowPII = false). This is the recommended setting in production and is set by default.

Problem Statement
Currently, all log message arguments are treated as PII and scrubbed as part of message formatting making debugging challenging

Proposed Changes
Mark Non-PII arguments as such by wrapping the argument with LogHelper.MarkAsNonPII(argName) method. This includes system metadata like variable/method/class names, data types, timestamps, claim names and algorithms used. Do this for all log and exception messages

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -84,7 +84,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
             if (token.Length > MaximumTokenSizeInBytes)
             {
-                LogHelper.LogInformation(TokenLogMessages.IDX10209, token.Length, MaximumTokenSizeInBytes);
+                LogHelper.LogInformation(TokenLogMessages.IDX10209, LogHelper.MarkAsNonPII(token.Length), LogHelper.MarkAsNonPII(MaximumTokenSizeInBytes));
                 return false;
             }
 
@@ -254,7 +254,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
             if ((tokenDescriptor.Subject == null || !tokenDescriptor.Subject.Claims.Any())
                 && (tokenDescriptor.Claims == null || !tokenDescriptor.Claims.Any()))
-                LogHelper.LogWarning(LogMessages.IDX14114, nameof(SecurityTokenDescriptor), nameof(SecurityTokenDescriptor.Subject), nameof(SecurityTokenDescriptor.Claims));
+                LogHelper.LogWarning(LogMessages.IDX14114, LogHelper.MarkAsNonPII(nameof(SecurityTokenDescriptor)), LogHelper.MarkAsNonPII(nameof(SecurityTokenDescriptor.Subject)), LogHelper.MarkAsNonPII(nameof(SecurityTokenDescriptor.Claims)));
 
             JObject payload;
             if (tokenDescriptor.Subject != null)
@@ -270,7 +270,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (tokenDescriptor.Audience != null)
             {
                 if (payload.ContainsKey(JwtRegisteredClaimNames.Aud))
-                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, nameof(tokenDescriptor.Audience)));
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, LogHelper.MarkAsNonPII(nameof(tokenDescriptor.Audience))));
 
                 payload[JwtRegisteredClaimNames.Aud] = tokenDescriptor.Audience;
             }
@@ -278,7 +278,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (tokenDescriptor.Expires.HasValue)
             {
                 if (payload.ContainsKey(JwtRegisteredClaimNames.Exp))
-                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, nameof(tokenDescriptor.Expires)));
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, LogHelper.MarkAsNonPII(nameof(tokenDescriptor.Expires))));
 
                 payload[JwtRegisteredClaimNames.Exp] = EpochTime.GetIntDate(tokenDescriptor.Expires.Value);
             }
@@ -286,7 +286,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (tokenDescriptor.Issuer != null)
             {
                 if (payload.ContainsKey(JwtRegisteredClaimNames.Iss))
-                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, nameof(tokenDescriptor.Issuer)));
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, LogHelper.MarkAsNonPII(nameof(tokenDescriptor.Issuer))));
 
                 payload[JwtRegisteredClaimNames.Iss] = tokenDescriptor.Issuer;
             }
@@ -294,7 +294,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (tokenDescriptor.IssuedAt.HasValue)
             {
                 if (payload.ContainsKey(JwtRegisteredClaimNames.Iat))
-                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, nameof(tokenDescriptor.IssuedAt)));
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, LogHelper.MarkAsNonPII(nameof(tokenDescriptor.IssuedAt))));
 
                 payload[JwtRegisteredClaimNames.Iat] = EpochTime.GetIntDate(tokenDescriptor.IssuedAt.Value);
             }
@@ -302,7 +302,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (tokenDescriptor.NotBefore.HasValue)
             {
                 if (payload.ContainsKey(JwtRegisteredClaimNames.Nbf))
-                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, nameof(tokenDescriptor.NotBefore)));
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, LogHelper.MarkAsNonPII(nameof(tokenDescriptor.NotBefore))));
 
                 payload[JwtRegisteredClaimNames.Nbf] = EpochTime.GetIntDate(tokenDescriptor.NotBefore.Value);
             }
@@ -473,7 +473,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         private string CreateTokenPrivate(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, string compressionAlgorithm, IDictionary<string, object> additionalHeaderClaims, string tokenType)
         {
             if (additionalHeaderClaims?.Count > 0 && additionalHeaderClaims.Keys.Intersect(JwtTokenUtilities.DefaultHeaderParameters, StringComparer.OrdinalIgnoreCase).Any())
-                throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX14116, nameof(additionalHeaderClaims), string.Join(", ", JwtTokenUtilities.DefaultHeaderParameters))));
+                throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX14116, LogHelper.MarkAsNonPII(nameof(additionalHeaderClaims)), LogHelper.MarkAsNonPII(string.Join(", ", JwtTokenUtilities.DefaultHeaderParameters)))));
 
             var header = CreateDefaultJWSHeader(signingCredentials, tokenType);
 
@@ -523,11 +523,11 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 throw LogHelper.LogArgumentNullException(nameof(compressionAlgorithm));
 
             if (!CompressionProviderFactory.Default.IsSupportedAlgorithm(compressionAlgorithm))
-                throw LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10682, compressionAlgorithm)));
+                throw LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10682, LogHelper.MarkAsNonPII(compressionAlgorithm))));
 
             var compressionProvider = CompressionProviderFactory.Default.CreateCompressionProvider(compressionAlgorithm);
 
-            return compressionProvider.Compress(Encoding.UTF8.GetBytes(token)) ?? throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10680, compressionAlgorithm)));
+            return compressionProvider.Compress(Encoding.UTF8.GetBytes(token)) ?? throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10680, LogHelper.MarkAsNonPII(compressionAlgorithm))));
         }
 
         /// <summary>
@@ -563,7 +563,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 if (claimType == ClaimTypes.Actor)
                 {
                     if (identity.Actor != null)
-                        throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX14112, JwtRegisteredClaimNames.Actort, jwtClaim.Value)));
+                        throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX14112, LogHelper.MarkAsNonPII(JwtRegisteredClaimNames.Actort), jwtClaim.Value)));
 
                     if (CanReadToken(jwtClaim.Value))
                     {
@@ -773,7 +773,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     }
                     catch (Exception ex)
                     {
-                        throw LogHelper.LogExceptionMessage(new SecurityTokenCompressionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10680, compressionAlgorithm), ex));
+                        throw LogHelper.LogExceptionMessage(new SecurityTokenCompressionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10680, LogHelper.MarkAsNonPII(compressionAlgorithm)), ex));
                     }
                 }
                 else
@@ -791,7 +791,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 }
                 catch (Exception ex)
                 {
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenEncryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10616, encryptingCredentials.Enc, encryptingCredentials.Key), ex));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenEncryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10616, LogHelper.MarkAsNonPII(encryptingCredentials.Enc), encryptingCredentials.Key), ex));
                 }
             }
         }
@@ -922,7 +922,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 throw LogHelper.LogArgumentNullException(nameof(token));
 
             if (token.Length > MaximumTokenSizeInBytes)
-                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(TokenLogMessages.IDX10209, token.Length, MaximumTokenSizeInBytes)));
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(TokenLogMessages.IDX10209, LogHelper.MarkAsNonPII(token.Length), LogHelper.MarkAsNonPII(MaximumTokenSizeInBytes))));
 
             return new JsonWebToken(token);
         }
@@ -954,7 +954,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 return new TokenValidationResult { Exception = LogHelper.LogArgumentNullException(nameof(validationParameters)), IsValid = false };
 
             if (token.Length > MaximumTokenSizeInBytes)
-                return new TokenValidationResult { Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(TokenLogMessages.IDX10209, token.Length, MaximumTokenSizeInBytes))), IsValid = false };
+                return new TokenValidationResult { Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(TokenLogMessages.IDX10209, LogHelper.MarkAsNonPII(token.Length), LogHelper.MarkAsNonPII(MaximumTokenSizeInBytes)))), IsValid = false };
 
             var tokenParts = token.Split(new char[] { '.' }, JwtConstants.MaxJwtSegmentCount + 1);
             if (tokenParts.Length != JwtConstants.JwsSegmentCount && tokenParts.Length != JwtConstants.JweSegmentCount)
@@ -1065,7 +1065,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                 var validatedJsonWebToken = validatedToken as JsonWebToken;
                 if (validatedJsonWebToken == null)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10506, typeof(JsonWebToken), validatedToken.GetType(), token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10506, LogHelper.MarkAsNonPII(typeof(JsonWebToken)), LogHelper.MarkAsNonPII(validatedToken.GetType()), token)));
 
                 return validatedJsonWebToken;
             }
@@ -1080,7 +1080,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                 jwtToken = securityToken as JsonWebToken;
                 if (jwtToken == null)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10509, typeof(JsonWebToken), securityToken.GetType(), token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10509, LogHelper.MarkAsNonPII(typeof(JsonWebToken)), LogHelper.MarkAsNonPII(securityToken.GetType()), token)));
             }
             else
             {
@@ -1202,7 +1202,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             var cryptoProviderFactory = validationParameters.CryptoProviderFactory ?? key.CryptoProviderFactory;
             if (!cryptoProviderFactory.IsSupportedAlgorithm(algorithm, key))
             {
-                LogHelper.LogInformation(LogMessages.IDX14000, algorithm, key);
+                LogHelper.LogInformation(LogMessages.IDX14000, LogHelper.MarkAsNonPII(algorithm), key);
                 return false;
             }
 
@@ -1210,7 +1210,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
             var signatureProvider = cryptoProviderFactory.CreateForVerifying(key, algorithm);
             if (signatureProvider == null)
-                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10636, key == null ? "Null" : key.ToString(), algorithm ?? "Null")));
+                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10636, key == null ? "Null" : key.ToString(), LogHelper.MarkAsNonPII(algorithm))));
 
             try
             {
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -82,7 +82,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             var cryptoProviderFactory = signingCredentials.CryptoProviderFactory ?? signingCredentials.Key.CryptoProviderFactory;
             var signatureProvider = cryptoProviderFactory.CreateForSigning(signingCredentials.Key, signingCredentials.Algorithm);
             if (signatureProvider == null)
-                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10637, signingCredentials.Key == null ? "Null" : signingCredentials.Key.ToString(), signingCredentials.Algorithm ?? "Null")));
+                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10637, signingCredentials.Key == null ? "Null" : signingCredentials.Key.ToString(), LogHelper.MarkAsNonPII(signingCredentials.Algorithm))));
 
             try
             {
@@ -114,11 +114,11 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             var cryptoProviderFactory = signingCredentials.CryptoProviderFactory ?? signingCredentials.Key.CryptoProviderFactory;
             var signatureProvider = cryptoProviderFactory.CreateForSigning(signingCredentials.Key, signingCredentials.Algorithm, cacheProvider);
             if (signatureProvider == null)
-                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10637, signingCredentials.Key == null ? "Null" : signingCredentials.Key.ToString(), signingCredentials.Algorithm ?? "Null")));
+                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10637, signingCredentials.Key == null ? "Null" : signingCredentials.Key.ToString(), LogHelper.MarkAsNonPII(signingCredentials.Algorithm))));
 
             try
             {
-                LogHelper.LogVerbose(LogHelper.FormatInvariant(LogMessages.IDX14201, cacheProvider));
+                LogHelper.LogVerbose(LogHelper.FormatInvariant(LogMessages.IDX14201, LogHelper.MarkAsNonPII(cacheProvider)));
                 return Base64UrlEncoder.Encode(signatureProvider.Sign(Encoding.UTF8.GetBytes(input)));
             }
             finally
@@ -146,13 +146,13 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 throw LogHelper.LogArgumentNullException(nameof(algorithm));
 
             if (!CompressionProviderFactory.Default.IsSupportedAlgorithm(algorithm))
-                throw LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10682, algorithm)));
+                throw LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10682, LogHelper.MarkAsNonPII(algorithm))));
 
             var compressionProvider = CompressionProviderFactory.Default.CreateCompressionProvider(algorithm);
 
             var decompressedBytes = compressionProvider.Decompress(tokenBytes);
 
-            return decompressedBytes != null ? Encoding.UTF8.GetString(decompressedBytes) : throw LogHelper.LogExceptionMessage(new SecurityTokenDecompressionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10679, algorithm)));
+            return decompressedBytes != null ? Encoding.UTF8.GetString(decompressedBytes) : throw LogHelper.LogExceptionMessage(new SecurityTokenDecompressionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10679, LogHelper.MarkAsNonPII(algorithm))));
         }
 
         /// <summary>
@@ -192,7 +192,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 if (!cryptoProviderFactory.IsSupportedAlgorithm(decryptionParameters.Enc, key))
                 {
                     algorithmNotSupportedByCryptoProvider = true;
-                    LogHelper.LogWarning(TokenLogMessages.IDX10611, decryptionParameters.Enc, key);
+                    LogHelper.LogWarning(TokenLogMessages.IDX10611, LogHelper.MarkAsNonPII(decryptionParameters.Enc), key);
                     continue;
                 }
 
@@ -233,7 +233,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10603, keysAttempted, exceptionStrings, decryptionParameters.EncodedToken)));
 
             if (!decryptionSucceeded && algorithmNotSupportedByCryptoProvider)
-                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10619, decryptionParameters.Alg, decryptionParameters.Enc)));
+                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10619, LogHelper.MarkAsNonPII(decryptionParameters.Alg), LogHelper.MarkAsNonPII(decryptionParameters.Enc))));
 
             if (!decryptionSucceeded)
                 throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10609, decryptionParameters.EncodedToken)));
@@ -244,7 +244,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             using (AuthenticatedEncryptionProvider decryptionProvider = cryptoProviderFactory.CreateAuthenticatedEncryptionProvider(key, decryptionParameters.Enc))
             {
                 if (decryptionProvider == null)
-                    throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10610, key, decryptionParameters.Enc)));
+                    throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10610, key, LogHelper.MarkAsNonPII(decryptionParameters.Enc))));
 
                 return decryptionProvider.Decrypt(
                     Base64UrlEncoder.DecodeBytes(decryptionParameters.Ciphertext),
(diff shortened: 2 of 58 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -305,7 +305,7 @@ internal static SecurityKey GetSecurityKey(EncryptingCredentials encryptingCrede
                 else if (SecurityAlgorithms.Aes256CbcHmacSha512.Equals(encryptingCredentials.Enc, StringComparison.Ordinal))
                     securityKey = new SymmetricSecurityKey(JwtTokenUtilities.GenerateKeyBytes(512));
                 else
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenEncryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10617, SecurityAlgorithms.Aes128CbcHmacSha256, SecurityAlgorithms.Aes192CbcHmacSha384, SecurityAlgorithms.Aes256CbcHmacSha512, encryptingCredentials.Enc)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenEncryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10617, LogHelper.MarkAsNonPII(SecurityAlgorithms.Aes128CbcHmacSha256), LogHelper.MarkAsNonPII(SecurityAlgorithms.Aes192CbcHmacSha384), LogHelper.MarkAsNonPII(SecurityAlgorithms.Aes256CbcHmacSha512), encryptingCredentials.Enc)));
```
GeoK: Encryption algorithm can be also marked as non-PII.

sruke: Brent also confirmed this morning that it is safe. I was unsure earlier.

src/Microsoft.IdentityModel.Logging/NonPII.cs
```diff
@@ -0,0 +1,59 @@
...
+            Argument = argument;
+        }
+
+        /// <summary>
+        /// Returns a string that represents the <see cref="Argument"/>.
+        /// </summary>
+        /// <returns><c>null</c> if the <see cref="Argument"/> is <c>null</c>, otherwise calls <see cref="System.ValueType.ToString()"/> method of the <see cref="Argument"/>.</returns>
+        public override string ToString()
+        {
+            return Argument?.ToString() ?? "null";
```
TimHannMSFT: nit: looks like elsewhere in the code we're doing a capital "Null" so consistent style here might be to capitalize.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -82,7 +82,7 @@ public static string CreateEncodedSignature(string input, SigningCredentials sig
             var cryptoProviderFactory = signingCredentials.CryptoProviderFactory ?? signingCredentials.Key.CryptoProviderFactory;
             var signatureProvider = cryptoProviderFactory.CreateForSigning(signingCredentials.Key, signingCredentials.Algorithm);
             if (signatureProvider == null)
-                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10637, signingCredentials.Key == null ? "Null" : signingCredentials.Key.ToString(), signingCredentials.Algorithm ?? "Null")));
+                throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(TokenLogMessages.IDX10637, signingCredentials.Key == null ? "Null" : signingCredentials.Key.ToString(), LogHelper.MarkAsNonPII(signingCredentials.Algorithm ?? "Null"))));
```
TimHannMSFT: nit: looks like the NonPII to string already handles the null value to string problem so you can probably get away with just letting it get handled at that layer now (here and elsewhere)

TimHannMSFT (approved):
I reviewed the first 27 files and they look good to me. I'll try to get back to the rest of them but wanted to share progress for now.

ASIDE: Are there any tricks for turning on word wrap etc that makes these very long lines easier to see or viewing these PRs via codeflow or something? Scrolling horizontally is not a very smooth experience in gitHub reviews.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
sruke: Marked Security Algorithm as NonPII

src/Microsoft.IdentityModel.Logging/NonPII.cs
sruke: > I reviewed the first 27 files and they look good to me. I'll try to get back to the rest of them but wanted to share progress for now.
> 
> ASIDE: Are there any tricks for turning on word wrap etc that makes these very long lines easier to see or viewing these PRs via codeflow or something? Scrolling horizontally is not a very smooth experience in gitHub reviews.

@TimHannMSFT 
I couldn't find an option to do word wrap in GitHub. Another option to consider is to wrap logs during development so that they don't reviewers don't have to horizontally scroll. The log lines were already long and marking arguments as NonPII made them even longer. Is there a styling preference to keep the log in a single line?
ClodeFlow looks interesting. I would like to learn more from you on how to make use of it.

sruke: Updated

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
sruke: That makes sense. Updated all instances.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -656,24 +656,24 @@ protected virtual void ValidateNonce(OpenIdConnectProtocolValidationContext vali
                 }
                 catch (Exception ex)
                 {
-                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidNonceException(LogHelper.FormatInvariant(LogMessages.IDX21326, timestamp, nonceFoundInJwt), ex));
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidNonceException(LogHelper.FormatInvariant(LogMessages.IDX21326, LogHelper.MarkAsNonPII(timestamp), nonceFoundInJwt), ex));
```
sruke, line 659: Can NonceFoundInJWT be marked as NonPII? My understanding is that Nonce is a randomly generated string sent by the client as part of the authorization request. On receiving a response client will use nonce to validate the response was indeed for the initial request made by them.

TimHannMSFT: I would say it's not safe to log as nonPII. The nonce can be used as a sort of security key (it is in the ServerNonce scenario) if someone got access to the logs then, they'd be able to potentially craft calls that would otherwise be invalid (were they not to know a valid nonce).

sruke: That's helpful to know.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -473,7 +473,7 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
         private string CreateTokenPrivate(JObject payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, string compressionAlgorithm, IDictionary<string, object> additionalHeaderClaims, string tokenType)
         {
             if (additionalHeaderClaims?.Count > 0 && additionalHeaderClaims.Keys.Intersect(JwtTokenUtilities.DefaultHeaderParameters, StringComparer.OrdinalIgnoreCase).Any())
-                throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX14116, nameof(additionalHeaderClaims), string.Join(", ", JwtTokenUtilities.DefaultHeaderParameters))));
+                throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX14116, LogHelper.MarkAsNonPII(nameof(additionalHeaderClaims)), LogHelper.MarkAsNonPII(string.Join(", ", JwtTokenUtilities.DefaultHeaderParameters)))));
```
jennyf19, line 476: >additionalHeaderClaims [](http://example.com/codeflow?start=157&length=22)

these would not contain Pii? #Resolved

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -370,7 +371,7 @@ private static long ParseTimeValue(JToken jToken, string claimName)
                     return (long)resultDouble;
             }
 
-            throw LogHelper.LogExceptionMessage(new FormatException(LogHelper.FormatInvariant(LogMessages.IDX14300, claimName, jToken.ToString(), typeof(long))));
+            throw LogHelper.LogExceptionMessage(new FormatException(LogHelper.FormatInvariant(LogMessages.IDX14300, LogHelper.MarkAsNonPII(claimName), jToken.ToString(), LogHelper.MarkAsNonPII(typeof(long)))));
```
jennyf19, line 374: >claimName [](http://example.com/codeflow?start=139&length=9)

what is claimName in this context?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
TimHannMSFT: This is the name of the variable not the value, the other part is just the claim names.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -380,20 +380,20 @@ protected virtual void ValidateIdToken(OpenIdConnectProtocolValidationContext va
...
 
                 if (idToken.Payload.Iss == null)
-                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, JwtRegisteredClaimNames.Iss.ToLowerInvariant(), idToken)));
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, LogHelper.MarkAsNonPII(JwtRegisteredClaimNames.Iss.ToLowerInvariant()), idToken)));
 
                 // sub is required in OpenID spec; but we don't want to block valid idTokens provided by some identity providers
                 if (RequireSub && (string.IsNullOrWhiteSpace(idToken.Payload.Sub)))
-                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, JwtRegisteredClaimNames.Sub.ToLowerInvariant(), idToken)));
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, LogHelper.MarkAsNonPII(JwtRegisteredClaimNames.Sub.ToLowerInvariant()), idToken)));
```
jennyf19, line 396: >JwtRegisteredClaimNames.Sub [](http://example.com/codeflow?start=162&length=27)

I think this is considered OII, which would fall under PII

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -831,7 +831,7 @@ internal virtual void ValidateQClaim(JsonWebToken signedHttpRequest, SignedHttpR
                 {
                     if (!sanitizedQueryParams.TryGetValue(queryParamName, out var queryParamsValue))
                     {
-                        throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidQClaimException(LogHelper.FormatInvariant(LogMessages.IDX23028, queryParamName, string.Join(", ", sanitizedQueryParams.Select(x => x.Key)))));
+                        throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidQClaimException(LogHelper.FormatInvariant(LogMessages.IDX23028, LogHelper.MarkAsNonPII(queryParamName), LogHelper.MarkAsNonPII(string.Join(", ", sanitizedQueryParams.Select(x => x.Key))))));
```
jennyf19, line 834: >sanitizedQueryParams [](http://example.com/codeflow?start=233&length=20)

query params can have anything really. In MSAL.NET the developer can attached any key/value pair in the qp's. not sure if it's similar for this case.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1270,7 +1270,7 @@ private Uri EnsureAbsoluteUri(Uri uri)
 
             if (repeatedQueryParams.Any())
             {
-                LogHelper.LogWarning(LogHelper.FormatInvariant(LogMessages.IDX23004, string.Join(", ", repeatedQueryParams)));
+                LogHelper.LogWarning(LogHelper.FormatInvariant(LogMessages.IDX23004, LogHelper.MarkAsNonPII(string.Join(", ", repeatedQueryParams))));
```
jennyf19, line 1273: >repeatedQueryParams [](http://example.com/codeflow?start=126&length=19)

not sure if these could have PII

test/Microsoft.IdentityModel.Logging.Tests/PIITests.cs
```diff
@@ -178,16 +178,71 @@ public void LogXmlExceptions()
...
+
+            string algorithm = null;
+
+            // Testset 1 with PII logging off
+            IdentityModelEventSource.ShowPII = false;
+
+            LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant("Algorithm not supported exception 1: {0}", algorithm)));
+            LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant("Algorithm not supported exception 2: {0}", LogHelper.MarkAsNonPII(algorithm))));
+
+            Assert.Contains("Algorithm not supported exception 1: [PII of type 'Null' is hidden. For more details, see https://aka.ms/IdentityModel/PII.]", listener.TraceBuffer);
```
jennyf19, line 208: >[PII of type 'Null' is hidden. For more details, see https://aka.ms/IdentityModel/PII.] [](http://example.com/codeflow?start=66&length=87)

i will need to maybe try this out, and have only had time to look at a few classes, but will this message be printed with every PII message? Right now, I think it gets printed once at the top of the logs.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
jennyf19: thanks. i meant the claim names.

test/Microsoft.IdentityModel.Logging.Tests/PIITests.cs
sruke: Before logging a message all arguments are scrubbed for PII and this message gets displayed in place of every PII argument. Earlier the type of argument wasn't shown. Now you can view the type.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
sruke: Checked with George earlier with the same concern and he said we can mark as NonPII. Result is a comma separate list of repeated query parameter names. Let me know if we should leave it as PII instead

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
sruke: We're logging only the keys here. Should we leave them as PII instead?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
sruke: In this context claimName can be JwtRegisteredClaimNames.Iat or JwtRegisteredClaimNames.Nbf or JwtRegisteredClaimNames.Exp

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
sruke: DefaultHeaderParameters are string constants - JwtHeaderParameterNames.Alg, JwtHeaderParameterNames.Kid, JwtHeaderParameterNames.X5t, JwtHeaderParameterNames.Enc, JwtHeaderParameterNames.Zip. Should we leave them classified as PII instead?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
sruke: That's Organization Identifiable Information?

sruke: Updated. JwtRegisteredClaimNames.Sub is no longer marked as NonPII

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -270,39 +270,39 @@ public virtual string CreateToken(SecurityTokenDescriptor tokenDescriptor)
             if (tokenDescriptor.Audience != null)
             {
                 if (payload.ContainsKey(JwtRegisteredClaimNames.Aud))
-                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, nameof(tokenDescriptor.Audience)));
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, LogHelper.MarkAsNonPII(nameof(tokenDescriptor.Audience))));
```
brentschmaltz, line 273: There are a couple of things here that could be considered, some were here already.
1. We don't know if we are going to log this message yet, it depends on if the EventSource is setup and the level. Yet we are creating and formatting a string.
2. We are creating a new object on the stack as LogHelper.MarkAsNonPII creates a new object.

Could we delay the action of creating a string and the new object until we know we will be logging?
It is not as important when we are throwing, as that interrupts the stack and we assume this is infrequent.
This work will occur in the positive case.

KalleOlaviNiemitalo:
> ASIDE: Are there any tricks for turning on word wrap etc

The GitHub mobile app has a line wrapping option.

Screen shot of Code Options settings in GitHub mobile app on iOS Screen shot of a pull request in GitHub mobile app on iOS, demonstrating line wrapping

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
sruke: Essentially we want to make a decision at runtime after the EventLevel has been decided to choose which arguments are NonPII and display them as clear text. 

We could come up with rules on which arguments can be considered as NonPII (ex. if argument contains nameof keyword) and mark them as such. These rules can be executed after EventLevel has been decided. Downside to this approach is some arguments that were safe to be displayed might be missed and rules need to be updated regularly to keep up with new logs getting added.

What are your thoughts @brentschmaltz on this approach?

jmprieur: @sruke, @brentschmaltz 
I was wondering, if we couldn't step back and encode the non-PII in the log message. For instance, we could have a little sign rarely used in messages (`~`) added to the log messages (before or after the {0} {1}). This would tag that this data is not PII

This way this would be data driven (contained in the log message) and auto descriptive, and we would not need to do any changes in the line of code calling LogHelper.LogXX(). This would be by the LogHelper.LogXX() methods themselves

Of course LogHelper.LogXX() would apply or not the PII to each argument depending on the little sign, and strip away the little sign

For instance:

```CSharp
internal const string IDX14113 = "IDX14113: A duplicate value for 'SecurityTokenDescriptor.{0}~' exists in 'SecurityTokenDescriptor.Claims'. \nThe value of 'SecurityTokenDescriptor.{0}~' is used.";
```

would contain the `~` sign because there is no PII. No extra work, and no extra allocation here
It would be called as expected:

```CSharp
LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX14113, nameof(tokenDescriptor.Audience)));
```

Another advantage is we could automatically publish a doc of the log messages (request from office of the CTO) which would also explain what is PII or not ...

(sorry for jumping late on this, I realize it's a different design)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
TimHannMSFT: Can't see how the claim names would be pii by any definition. That said, my previous comments from the SAL version of this PR stand. Better to be overly conservative and get the initial big PR asap to save yourself the ongoing merge headache.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
TimHannMSFT: Disagree here too, "sub" is not PII or OII the claim contents would be. Moreover OII is the superset, PII is more specific, often logging OII is fine, would have to confirm on what policy in Wilson is (@brentschmaltz , @GeoK to comment)

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -344,18 +344,32 @@ public static string FormatInvariant(string format, params object[] args)
 
         private static string RemovePII(object arg)
         {
-            if (arg == null)
-                return string.Empty;
-
             if (arg is Exception ex && ex.GetType().FullName.StartsWith("Microsoft.IdentityModel.", StringComparison.Ordinal))
```
TimHannMSFT: can you reference IsCustomException here to decrease code duplication?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -707,17 +707,17 @@ protected virtual void ValidateState(OpenIdConnectProtocolValidationContext vali
...
             {
                 // 'state' was sent, but message does not contain 'state'
-                throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidStateException(LogHelper.FormatInvariant(LogMessages.IDX21330, RequireState.ToString())));
+                throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidStateException(LogHelper.FormatInvariant(LogMessages.IDX21330, LogHelper.MarkAsNonPII(RequireState))));
             }
 
             if (!string.Equals(validationContext.State, validationContext.ProtocolMessage.State, StringComparison.Ordinal))
             {
-                throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidStateException(LogHelper.FormatInvariant(LogMessages.IDX21331, validationContext.State, validationContext.ProtocolMessage.State)));
+                throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidStateException(LogHelper.FormatInvariant(LogMessages.IDX21331, LogHelper.MarkAsNonPII(validationContext.State), LogHelper.MarkAsNonPII(validationContext.ProtocolMessage.State))));
```
TimHannMSFT: What sort of data is in "State". Off the top of my head I don't think there's PII in OIDC data so might not matter but good to double check if you haven't.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -1129,7 +1134,7 @@ protected virtual void ProcessStatements(ICollection<Saml2Statement> statements,
                 else if (statement is Saml2AuthorizationDecisionStatement authzStatement)
                     ProcessAuthorizationDecisionStatement(authzStatement, identity, issuer);
                 else
-                    LogWarning(LogMessages.IDX13516, this.GetType());
+                    LogWarning(LogMessages.IDX13516, LogHelper.MarkAsNonPII(this.GetType()));
```
TimHannMSFT: shouldn't this use _className?

src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs
```diff
@@ -203,7 +203,7 @@ private static uint GetKeyByteCount(string curveId)
                     keyByteCount = 66;
                     break;
                 default:
-                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10645, curveId)));
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10645, LogHelper.MarkAsNonPII(curveId))));
```
TimHannMSFT, line 206: @cesarpayan I know you're looking into ECD stuff, is the curve ID secret or is that like the public key. I know very little about this stuff.

src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
```diff
@@ -224,7 +224,7 @@ public override byte[] UnwrapKey(byte[] keyBytes)
                 throw LogHelper.LogArgumentNullException(nameof(keyBytes));
 
             if (keyBytes.Length % 8 != 0)
-                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10664, keyBytes.Length << 3), nameof(keyBytes)));
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10664, LogHelper.MarkAsNonPII(keyBytes.Length << 3)), nameof(keyBytes)));
```
TimHannMSFT, line 227: nameof(keyBytes) can be NonPII too right?

src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
```diff
@@ -340,20 +340,20 @@ private void ValidateKeySize(byte[] key, string algorithm)
...
+                    throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(key), LogHelper.FormatInvariant(LogMessages.IDX10662, LogHelper.MarkAsNonPII(algorithm), LogHelper.MarkAsNonPII(128), Key.KeyId, LogHelper.MarkAsNonPII(key.Length << 3))));
 
                 return;
             }
 
             if (SecurityAlgorithms.Aes256KW.Equals(algorithm, StringComparison.Ordinal) || (SecurityAlgorithms.Aes256KeyWrap.Equals(algorithm, StringComparison.Ordinal)))
             {
                 if (key.Length != 32)
-                    throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(key), LogHelper.FormatInvariant(LogMessages.IDX10662, algorithm, 256, Key.KeyId, key.Length << 3)));
+                    throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(key), LogHelper.FormatInvariant(LogMessages.IDX10662, LogHelper.MarkAsNonPII(algorithm), LogHelper.MarkAsNonPII(256), LogHelper.MarkAsNonPII(Key.KeyId), LogHelper.MarkAsNonPII(key.Length << 3))));
```
TimHannMSFT: Here you're marking KeyId as NonPII, other places it's not marked. I don't understand the scenario well enough to say either way but it should be consistent I'd think.

src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
```diff
@@ -371,7 +371,7 @@ public override byte[] WrapKey(byte[] keyBytes)
                 throw LogHelper.LogArgumentNullException(nameof(keyBytes));
 
             if (keyBytes.Length % 8 != 0)
-                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10664, keyBytes.Length << 3), nameof(keyBytes)));
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10664, LogHelper.MarkAsNonPII(keyBytes.Length << 3)), nameof(keyBytes)));
```
TimHannMSFT, line 374: LogHelper.MarkAsNonPII(nameof(keyBytes)) ?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -79,7 +79,7 @@ internal InMemoryCryptoProviderCache(CryptoProviderCacheOptions cryptoProviderCa
                 throw LogHelper.LogArgumentNullException(nameof(cryptoProviderCacheOptions));
 
             if (tryTakeTimeout <= 0)
-                throw LogHelper.LogArgumentException<ArgumentException>(nameof(tryTakeTimeout), $"{nameof(tryTakeTimeout)} must be greater than zero");
+                throw LogHelper.LogArgumentException<ArgumentException>(nameof(tryTakeTimeout), $"{LogHelper.MarkAsNonPII(nameof(tryTakeTimeout))} must be greater than zero");
```
TimHannMSFT: the second param here is a string right? Does the MarkAsNonPII do anything here?

src/Microsoft.IdentityModel.Xml/EnvelopedSignatureWriter.cs
```diff
@@ -179,7 +179,7 @@ private Signature CreateSignature()
             var hashAlgorithm = cryptoProviderFactory.CreateHashAlgorithm(digestAlgorithm);
 
             if (hashAlgorithm == null)
-                throw LogExceptionMessage(new XmlValidationException(FormatInvariant(LogMessages.IDX30213, cryptoProviderFactory.ToString(), _signingCredentials.Digest)));
+                throw LogExceptionMessage(new XmlValidationException(FormatInvariant(LogMessages.IDX30213, MarkAsNonPII(cryptoProviderFactory.ToString()), _signingCredentials.Digest)));
```
TimHannMSFT: Doesn't looke like CryptoProviderFactory exposes a ToString override, does this just show the type? Is it better to just make that explicit here?

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -755,12 +755,12 @@ private DateTime GetDateTime(string key)
             {
                 if (ex is FormatException || ex is ArgumentException || ex is InvalidCastException)
                 {
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX12700, key, (dateValue ?? "<null>")), ex));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX12700, key, LogHelper.MarkAsNonPII((dateValue ?? "<null>"))), ex));
```
TimHannMSFT: nit: not your change but seems odd the null string format is different here than elsewhere

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -755,12 +755,12 @@ private DateTime GetDateTime(string key)
...
                 }
 
                 if (ex is OverflowException)
                 {
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX12701, key, (dateValue ?? "<null>")), ex));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(LogMessages.IDX12701, key, LogHelper.MarkAsNonPII((dateValue ?? "<null>"))), ex));
```
TimHannMSFT: and here...

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -95,7 +96,7 @@ public class JwtSecurityTokenHandler : SecurityTokenHandler
         /// </summary>
         static JwtSecurityTokenHandler()
         {
-            LogHelper.LogVerbose("Assembly version info: " + typeof(JwtSecurityTokenHandler).AssemblyQualifiedName);
+            LogHelper.LogVerbose("Assembly version info: " + LogHelper.MarkAsNonPII(typeof(JwtSecurityTokenHandler).AssemblyQualifiedName));
```
TimHannMSFT, line 99: why not _className?

TimHannMSFT:
Finished looking at all the files at least once. didn't make it back to review all the files that changed since the first review. Left a few comments. Thanks for all the good work here @sruke !

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
sruke: That's a really interesting idea @jmprieur and @brentschmaltz had some good comments too. However, this would require us to revisit the complete design on when to categorize arguments into PII and NonPII. 

Current design considers all arguments as PII and scrubs them before the EventLevel for logging is decided. We made a design decision to intercept this and mark system metadata as NonPII that required us to wrap individual arguments as NonPII and create new objects. A similar feature was introduced in SAL recently. To be consistent shall we proceed with what we have?

src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs
KalleOlaviNiemitalo: The curve ID is part of the public key. See [RFC 7518 §6.2.1 Parameters for Elliptic Curve Public Keys](https://www.rfc-editor.org/rfc/rfc7518#section-6.2.1) and [IANA registry for JSON Web Key Elliptic Curve](https://www.iana.org/assignments/jose/jose.xhtml#web-key-elliptic-curve).

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
KalleOlaviNiemitalo: "state" seems to be up to the application. I'd hide it to be safe. The OpenID Connect server is required to send the same "state" back; if the "state" does not match, then either the server is buggy or it's some kind of attack. I don't think the application developer needs to see the string value to continue from there.

src/Microsoft.IdentityModel.Xml/EnvelopedSignatureWriter.cs
KalleOlaviNiemitalo: CryptoProviderFactory is not sealed, and application-provided SigningCredentials.CryptoProviderFactory could override ToString() and possibly return PII from there. The IDX30213 message contains `The CryptoProviderFactory: '{0}'`.

<https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fc59ad574ee0e0b9026c405fb566f343bc314bd7/src/Microsoft.IdentityModel.Xml/LogMessages.cs#L75>

Perhaps it's best to use just `cryptoProviderFactory` as the format argument, with neither ToString nor MarkAsNonPII. If ShowPII == false, then the name of the type will be formatted anyway.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
KalleOlaviNiemitalo: _className does not include the assembly version number.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
KalleOlaviNiemitalo: The IDX13516 message is

<https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fc59ad574ee0e0b9026c405fb566f343bc314bd7/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/LogMessages.cs#L52>

so this should actually be `statement.GetType()` rather than `this.GetType()`.

src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs
KalleOlaviNiemitalo: If the call stack is as follows, then the public key belongs to the identity provider and not to the individual user.

- [Microsoft.IdentityModel.Protocols.OpenIdConnect.OpenIdConnectConfigurationRetriever.GetAsync](<https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fc59ad574ee0e0b9026c405fb566f343bc314bd7/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationRetriever.cs#L100>)
- [Microsoft.IdentityModel.Tokens.JsonWebKeySet.GetSigningKeys](<https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fc59ad574ee0e0b9026c405fb566f343bc314bd7/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs#L165>)
- [Microsoft.IdentityModel.Tokens.JsonWebKeyConverter.TryConvertToECDsaSecurityKey](<https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fc59ad574ee0e0b9026c405fb566f343bc314bd7/src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs#L353>)
- [Microsoft.IdentityModel.Tokens.ECDsaSecurityKey(JsonWebKey webKey, bool usePrivateKey)](<https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fc59ad574ee0e0b9026c405fb566f343bc314bd7/src/Microsoft.IdentityModel.Tokens/ECDsaSecurityKey.cs#L44>)
- [Microsoft.IdentityModel.Tokens.ECDsaAdapter.CreateECDsa](<https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fc59ad574ee0e0b9026c405fb566f343bc314bd7/src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs#L75>)
- [Microsoft.IdentityModel.Tokens.ECDsaAdapter.CreateECDsaUsingCNGKey](<https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fc59ad574ee0e0b9026c405fb566f343bc314bd7/src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs#L102>)
- [Microsoft.IdentityModel.Tokens.ECDsaAdapter.GetKeyByteCount](<https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fc59ad574ee0e0b9026c405fb566f343bc314bd7/src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs#L206>)

~The code in this repo doesn't seem to ever call `internal ECDsaSecurityKey(JsonWebKey webKey, bool usePrivateKey)` with `usePrivateKey: true`, not even in tests. Applications cannot do that either, because the constructor is `internal`.~ The constructor and such a call were added in commit ce2fd9dd8d12d5f3383cbbc5d9aae5d8623b32e6, which is on the `dev` branch and in the `6.5.0` tag but not on the `main` branch.

src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
sruke: nameof(keyBytes) falls outside the FormatInvariant method and cannot be marked as NonPII

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
sruke: You're right. Its a string and marking it as NonPII would just create another object.

TimHannMSFT:
[https://github.com/StylishThemes/GitHub-code-wrap](https://github.com/StylishThemes/GitHub-code-wrap) << Just do the word wrap on the website 
[https://github.com/henrik-me/Github-Codeflow-Extension-Edge](https://github.com/henrik-me/Github-Codeflow-Extension-Edge) << Allow GitHub PRs to be reviewed using CodeFlow!

---
In reply to: [961374858](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1725#issuecomment-961374858) [](http://example.com/codeflow?ancestors=961374858,961374858)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
jmprieur: I realize this is a completely different design, but this might be worth discussing? (again, sorry for being late on this).

brentschmaltz: @sruke @jmprieur that little '~' would address my concerns nicely, of course we need to be sure the ~ is from us and not an attacker. It is nice because we don't need the PII wrapper.

This idea made me think of an approach would be to use params where the first value defines the number of PII values.
For example:
params: 1,foo, bar would tell us the first string (foo) is PII.
this would force us to rearrange the strings in many of our messages.

I am not too concerned if we have to revisit some previous work as, finding the right solution that is natural and performant sometimes takes a couple of tries.

My main concern is that we create strings and objects that will never be logged in the success path.

henrik-me: +1 : create only when needed and use templating where possible so the strings are max created once. I suppose that will be addressed in the next pass of the string creation. 

@brentschmaltz wondering if any customers are currently depending on any of the strings and may see a re-ordering as a "breaking" change. hopefully they only depend on the id and not the actual string, so we can optimize how we generate the strings.

brentschmaltz: @henrik-me I should clarify the 'reorder' idea.
What i mean is if a string is $"foo {0}, {1}, {2}" with parameters x, z, y and 0 and 2 are non pii, then we would have to modify the string to be
$"foo {0}, {2}, {1}" with parameters x, y, z

the user would see the same string.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Updated

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
sruke: Hiding state as PII

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
sruke: You're right @KalleOlaviNiemitalo. Updated to statement.GetType()

src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs
sruke: Thank you for the explanation @KalleOlaviNiemitalo

src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
sruke: Thank you for pointing that out. KeyId is PII. Updated.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
sruke: Updated

src/Microsoft.IdentityModel.Xml/EnvelopedSignatureWriter.cs
sruke: Updated to display the class name of cryptoProviderFactory

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
sruke: Updated to "Null" for consistency

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
sruke: Updated to "Null" for consistency

Pikhulya:
Can you please double-check this change will address the issue outlined in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1762 ?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
sruke: I will merge this PR into Wilson as it has some worthwhile work that is helpful to the community and come up with a one pager summarizing your proposals and how their potentials implementations would look like and we can have another discussion with the main goal to limit unnecessary string generations. Once we have a path forward I'll revisit this feature.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Updated

src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs
sruke: nameof(keyBytes) falls outside Format.Invariant and cannot be marked as NonPII

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
brentschmaltz: I would agree that 'sub' is OII as it represents the id the identity provider users to track the user.

TimHannMSFT: @brentschmaltz are you saying the literal string "sub" is OII (I doubt this) or the value of the sub claim is OII (which I think we all agree on)

JwtRegisteredClaimNames.Sub.ToLowerInvariant() == "sub"

I can't imagine why this isn't safe to show in logs in plaintext

PR #1736, opened 2021-10-22
initial commit to move issuer validator to separate assembly from idweb

### LINKED ISSUE #1731: SecurityTokenInvalidIssuerException with OpenID Connect and the Azure AD "common" endpoint
When you use OpenID Connect against Azure AD's "common" endpoint, configuring `https://login.microsoftonline.com/common/v2.0` as the `OpenIdConnectOptions.Authority` value, the [metadata document](https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration) will provide an issuer value of `https://login.microsoftonline.com/{tenantid}/v2.0`. Note the "{tenantid}" placeholder.

Within an actual ID token, the issuer value no longer contains that placeholder, but the actual tenant ID determined by the user logging in (e.g., `https://login.microsoftonline.com/9188040d-6c67-4c5b-b112-36a304b66dad/v2.0`). This causes the issuer validation in `Microsoft.IdentityModel.Tokens` to fail.

I guess Azure AD decided to return an issuer with a non-standard (?) placeholder here, but shouldn't the code in 
`azure-activedirectory-identitymodel-extensions-for-dotnet` be able to deal with such Azure AD specifics? If so, I'd consider this a bug.

Also, the exception message is `IDX10205: Issuer validation failed. Issuer: 'System.String'. Did not match: validationParameters.ValidIssuer: 'System.String' or validationParameters.ValidIssuers: 'System.String'.` Which is not at all helpful, but I'll open a separate issue for this.

jennyf19:
initial commit to move issuer validator to separate assembly from MS.Id.Web

I'm unable to get the VS test runner to run, but the tests pass in the command line. will keep trying to get that sorted out.

@GeoK have addressed your [PR comments](https://github.com/AzureAD/microsoft-identity-web/pull/1493#discussion_r733158659) from the initial PR in Id.Web.

**note**: only code changes were from suggestions [1](https://github.com/AzureAD/microsoft-identity-web/pull/1493#discussion_r733158659) and [2](https://github.com/AzureAD/microsoft-identity-web/pull/1493#discussion_r733174445) from @GeoK between Id.Web and this commit.

Also resolves #1731 

**Update** @brentschmaltz 

In `OpenIdConnectOptions` configuration, we would now do the following with the latest changes (10.24):

```csharp
AadIssuerValidator aadIssuerValidator = AadIssuerValidator.GetAadIssuerValidator(authority, openIdConnectOptions.Backchannel).Validate;
 
 ```
 
 **Update 10.25**
 Tested M.IM.Validators with ID Web -> Web app calls web API calls graph & B2C calls web API

### CODE REVIEW

File: src/Microsoft.IdentityModel.Tokens/Properties/AssemblyInfo.cs [hunk 1]
```diff
@@ -47,3 +47,4 @@
...
 [assembly: InternalsVisibleTo("Microsoft.IdentityModel.Protocols.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
+[assembly: InternalsVisibleTo("Microsoft.IdentityModel.Validators, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
jennyf19, line 50: this is for the M.IM.Json opensource classes.

jmprieur (approved):
LGTM
Thanks @jennyf19.

For the error message strings that were renamed form IDW to IDX401xx, 
shall we want to do a transformation at the IdWeb level ? or just document that they had changed?

jennyf19:
I'm not sure what we will do, but can discuss once these updates are out and we can try things out more easily.

---
In reply to: [786380485](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1736#issuecomment-786380485) [](http://example.com/codeflow?ancestors=786380485,786380485)

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,297 @@
...
+        /// <param name="validationParameters">Token validation parameters.</param>
+        /// <remarks>The issuer is considered as valid if it has the same HTTP scheme and authority as the
+        /// authority from the configuration file, has a tenant ID, and optionally v2.0 (this web API
+        /// accepts both V1 and V2 tokens).</remarks>
+        /// <returns>The <c>issuer</c> if it's valid, or otherwise <c>SecurityTokenInvalidIssuerException</c> is thrown.</returns>
+        /// <exception cref="ArgumentNullException"> if <paramref name="securityToken"/> is null.</exception>
+        /// <exception cref="ArgumentNullException"> if <paramref name="validationParameters"/> is null.</exception>
+        /// <exception cref="SecurityTokenInvalidIssuerException">if the issuer is invalid or if there is a network issue. </exception>
+        public string Validate(
+            string actualIssuer,
```
brentschmaltz: suggest name: issuer rather than actualIssuer. #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,297 @@
...
+            SecurityToken securityToken,
+            TokenValidationParameters validationParameters)
+        {
+            if (string.IsNullOrEmpty(actualIssuer))
+            {
+                throw new ArgumentNullException(nameof(actualIssuer));
```
brentschmaltz: normally we log all exceptions thrown  using LogHelper.LogArgumentNullException, we should follow this same pattern here.
see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/d8ad857088d961ccccf69058c7fd7eb4f640607c/src/Microsoft.IdentityModel.Tokens/Validators.cs#L223 #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,297 @@
...
+using System.IdentityModel.Tokens.Jwt;
+using System.Net.Http;
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Validators
+{
+    /// <summary>
+    /// Generic class that validates token issuer from the Microsoft identity platform (AAD).
```
brentschmaltz: Wilson validates SAML and SAML2 tokens, we need to comment that this class only works with either a JsonWebToken or a JwtSecurityToken. #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidatorFactory.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// <summary>
+    /// Factory class for creating the AadIssuerValidator per authority.
+    /// </summary>
+    public class AadIssuerValidatorFactory
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AadIssuerValidatorFactory"/> class.
+        /// </summary>
+        /// <param name="httpClient">Optional HttpClient to use to retrieve the endpoint metadata (can be null).</param>
+        public AadIssuerValidatorFactory(
```
brentschmaltz: normally we don't use default parameters and have multiple ctors.
The reason is adding a default parameter can cause a runtime failure, so we just choose to not use them for public apis. #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: fixed

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: yes, thanks! i meant to look into this yesterday. it was on my list. updated and fixed.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidatorFactory.cs
jennyf19: make sense. we had an issue w/this w/MSAL.NET a few weeks ago. 

fixed.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: updated the comment. good point.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,298 @@
...
+                    if (IsValidIssuer(AadIssuerV1, tenantId, issuer))
+                    {
+                        return issuer;
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidIssuerException(
+                string.Format(
```
GeoK: LogHelper.FormatInvariant is used on other places to make sure log parameters are obfuscated. There is a parallel work stream that would allow non-PII arguments not to get obfuscated, but it will work in conjunction with LogHelper.FormatInvariant. #Resolved

GeoK:
We should update strongname bypass scripts as M.IM.Validators is a new assembly ([#1](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/build/strongNameBypass.reg), [#2](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/build/strongNameBypass2.reg))

---
In reply to: [949995333](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1736#issuecomment-949995333) [](http://example.com/codeflow?ancestors=949995333)

---
In reply to: [949995333](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1736#issuecomment-949995333) [](http://example.com/codeflow?ancestors=949995333)

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: should i use `LogHelper.FormatInvariant` instead?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,298 @@
...
+            {
+                throw new ArgumentNullException(nameof(securityToken));
+            }
+
+            if (validationParameters == null)
+            {
+                throw new ArgumentNullException(nameof(validationParameters));
+            }
+
+            string tenantId = GetTenantIdFromToken(securityToken);
```
GeoK, line 100: Can we pull _GetTenantId_ into _IsValidIssuer_? Thinking that for single-tenant apps there will be no need to extract tenantId and it would result in a waste of cycles for each request.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,298 @@
...
+        {
+            if (string.IsNullOrEmpty(issuer))
+            {
+                throw new ArgumentNullException(nameof(issuer));
+            }
+
+            if (securityToken == null)
+            {
+                throw new ArgumentNullException(nameof(securityToken));
+            }
```
GeoK:
```suggestion
_ = securityToken ?? throw new ArgumentNullException(nameof(securityToken));
``` #Resolved

File: src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs (hunk 7 above)
Comment (jennyf19) [reply src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs:100]: potentially IsValidIssuer could be called several times. would you recommend having the tenantId as a ref parameter of IsValidIssuer? computed the first time

File: src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs (hunk 8 above)
Comment (jennyf19) [reply src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs]: nice! thx. fixed.

**Comment by brentschmaltz** [comment]
@jmprieur there is more work to do before this should be pulled in.

File: src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs [hunk 9]
```diff
@@ -0,0 +1,290 @@
...
+        /// <exception cref="ArgumentNullException"> if <paramref name="securityToken"/> is null.</exception>
+        /// <exception cref="ArgumentNullException"> if <paramref name="validationParameters"/> is null.</exception>
+        /// <exception cref="SecurityTokenInvalidIssuerException">if the issuer is invalid or if there is a network issue. </exception>
+        public string Validate(
+            string issuer,
+            SecurityToken securityToken,
+            TokenValidationParameters validationParameters)
+        {
+            _ = issuer ?? throw new ArgumentNullException(nameof(issuer));
+
```
brentschmaltz: we usually don't have blank lines.
we try to have minimal spacing #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
brentschmaltz: yes please use LogHelper.FormatInvariant

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,290 @@
...
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Validators
+{
+    /// <summary>
+    /// Generic class that validates either JsonWebTokens or JwtSecurityTokens issued from the Microsoft identity platform (AAD).
+    /// </summary>
+    public class AadIssuerValidator
+    {
+        internal AadIssuerValidator(
```
brentschmaltz, line 46: How will this constructor be used? #Resolved

jmprieur: Something like the following:

```CSharp
AadIssuerValidatorFactory factory = new AadIssuerValidatorFactory(/* http client */);
options.TokenValidationParameters.IssuerValidator = factory.GetAadIssuerValidator(authority).Validate;
```

@jennyf19 : we should add a `<example><code>` blob in the XML comment to show how this method is used.

Or with SAL policies, [this sample code](https://identitydivision.visualstudio.com/DevEx/_git/MISE?path=/src/AuthenticationTicketProvider/S2SAuthenticationManagerFactory.cs&version=GBjmprieur/migrationGuides&line=56&lineEnd=58&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidatorFactory.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Net.Http;
+
+namespace Microsoft.IdentityModel.Validators
+{
+    /// <summary>
+    /// Factory class for creating the AadIssuerValidator per authority.
+    /// </summary>
+    public class AadIssuerValidatorFactory
```
brentschmaltz: I am not completely understanding the need for this factory.
It seems that the usage pattern is a singleton that is attached to TokenValidationParameters on startup.
Help me understand the need to cache the delegate.
We could just as easily have a static method on the class. #Pending

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: update the xml

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: sounds good. have removed several.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: fixed. thanks. very cool

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/InstanceDiscovery/IssuerConfigurationRetriever.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Protocols;
+
+namespace Microsoft.IdentityModel.Validators
+{
+    /// <summary>
+    /// An implementation of IConfigurationRetriever geared towards Azure AD issuers metadata.
+    /// </summary>
+    internal class IssuerConfigurationRetriever : IConfigurationRetriever<IssuerMetadata>
```
brentschmaltz: Why do we need a custom IConfigurationRetriever?
Are we getting a standard discovery document? #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
brentschmaltz: I am questioning the need for the factory, a static method on the class would do the same thing.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidatorFactory.cs
jennyf19: Previously we had a static, but customers wanted to inject an HttpClient to retrieve the OpenId Connect doc, so we moved to a factory and we cache the delegates because there can be several authories in an application and several token versions.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/InstanceDiscovery/IssuerConfigurationRetriever.cs
jennyf19: I'm not sure. What do you recommend?

jmprieur: we do retrieve a standard discovery document, but only are interested in the issuer property, so why deserialize everything?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,290 @@
...
+                {
+                    return issuer;
+                }
+            }
+
+            try
+            {
+                if (securityToken.Issuer.EndsWith("v2.0", StringComparison.OrdinalIgnoreCase))
+                {
+                    if (AadIssuerV2 == null)
```
brentschmaltz, line 124: In the v2 case can't TokenValidationParameters.ConfigurationManager and / or Configuration be used to get the issuer from: BaseConfiguration? #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,290 @@
...
+                return AadAuthority.Replace($"{AadIssuerValidatorConstants.Organizations}/v2.0", AadIssuerValidatorConstants.Common);
+            }
+
+            return AadAuthority.Replace("/v2.0", string.Empty);
+        }
+
+        private ConfigurationManager<IssuerMetadata> CreateConfigManager(
+            string aadAuthority)
+        {
+            if (HttpClient != null)
```
brentschmaltz, line 225: We should be using the HttpClient from BaseConfigurationManager, we will need to expose that.
The asp.net team asked us to make sure all calls use that HttpClient. #Resolved

jennyf19: that's cool. we can have a static method then.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,254 @@
...
+                if (tenantId != null)
+                    return (string)tenantId;
+
+                // Since B2C doesn't have "tid" as default, get it from issuer
+                return GetTenantIdFromIss(jwtSecurityToken.Issuer);
+            }
+
+            if (securityToken is JsonWebToken jsonWebToken)
+            {
+                jsonWebToken.TryGetPayloadValue(AadIssuerValidatorConstants.Tid, out string tid);
```
brentschmaltz: this could be written (same below):
` 
if (jsonWebToken.TryGetPayloadValue(AadIssuerValidatorConstants.Tid, out string tid)
    return tid
`  #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,254 @@
...
+        /// <remarks>Only <see cref="JwtSecurityToken"/> and <see cref="JsonWebToken"/> are acceptable types.</remarks>
+        private static string GetTenantIdFromToken(SecurityToken securityToken)
+        {
+            if (securityToken is JwtSecurityToken jwtSecurityToken)
+            {
+                if (jwtSecurityToken.Payload.TryGetValue(AadIssuerValidatorConstants.Tid, out object tid))
+                    return (string)tid;
+
+                jwtSecurityToken.Payload.TryGetValue(AadIssuerValidatorConstants.TenantId, out object tenantId);
+                if (tenantId != null)
```
brentschmaltz: why check null here but not above? #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidatorFactory.cs
brentschmaltz: There could be two overloads and one could have a HttpClient as a parameter.
When there is no HttpClient passed, we need to use the HttpClient from BaseConfigurationManager.
Reusing the same HttpClient (in the null case we create and reuse a HttpClient) could lead to stale DNS issues.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/InstanceDiscovery/IssuerConfigurationRetriever.cs
brentschmaltz: Everything will get deserialized anyway, the entire json is read into memory.
That will take place once and is a very small time compared to the http get to the discovery endpoint.
These custom classes will need to be supported for a long time, it they are not needed, i recommend using the standard.
In the future is we need anything additional from AAD, we will need to change the object design.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
brentschmaltz: This was a new feature we just added a couple of weeks ago.
@mafurman knows most about it.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,254 @@
...
+        private static bool IsValidIssuer(string validIssuerTemplate, string tenantId, string actualIssuer)
+        {
+            if (string.IsNullOrEmpty(validIssuerTemplate))
+                return false;
+
+            if (validIssuerTemplate.Contains("{tenantid}"))
+            {
+                try
+                {
+                    Uri issuerFromTemplateUri = new Uri(validIssuerTemplate.Replace("{tenantid}", tenantId));
```
brentschmaltz: Why do we need to create the Uri's?
Below we just compare the strings. #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,254 @@
...
+            }
+
+            try
+            {
+                if (securityToken.Issuer.EndsWith("v2.0", StringComparison.OrdinalIgnoreCase))
+                {
+                    if (AadIssuerV2 == null)
+                    {
+                        IssuerMetadata issuerMetadata =
+                            CreateConfigManager(AadAuthority).GetConfigurationAsync().ConfigureAwait(false).GetAwaiter().GetResult();
```
brentschmaltz: I think there is a race condition here.
Consider two threads with different tokens, one v1, one v2.
AadAuthority is used in both cases. CreateV1Authority could modify AadAuthority.
It may be simpler to have two separate variables and initialize them on creation since the change in the V1 case is not dependent on the token arriving.
In any case, two variables would help. #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,254 @@
...
+            if (string.IsNullOrEmpty(iss))
+                return string.Empty;
+
+            var uri = new Uri(iss);
+
+            if (uri.Segments.Length == 3)
+                return uri.Segments[1].TrimEnd('/');
+
+            if (uri.Segments.Length == 5 && uri.Segments[1].TrimEnd('/') == AadIssuerValidatorConstants.Tfp)
+                throw new SecurityTokenInvalidIssuerException(LogMessages.IDX40104);
```
brentschmaltz: make sure to use LogHelper.LogExceptionMessage when throwing. #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: good point. fixed.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: yes. fixed. thanks for the pointing that out.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidatorFactory.cs
jennyf19: i think is is resovled now by using the validationParams to get the issuer, in the v2 case.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: will do, but need to check on the actual string format, example if there are trailing /'s anywhere or not.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: yes missed this one. i've checked the remaining files as well and updated if needed.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: fixed.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/InstanceDiscovery/IssuerConfigurationRetriever.cs
jennyf19: got it. fixed.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: not sure, but have used what you suggested here as well.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: fixed

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,323 @@
...
+        /// </summary>
+        /// <param name="issuer">Issuer to validate (will be tenanted).</param>
+        /// <param name="securityToken">Received security token.</param>
+        /// <param name="validationParameters">Token validation parameters.</param>
+        /// <example><code>
+        /// AadIssuerValidator aadIssuerValidator = AadIssuerValidator.GetAadIssuerValidator(authority, httpClient);
+        /// TokenValidationParameters.IssuerValidator = aadIssuerValidator.Validate;
+        /// </code></example>
+        /// <remarks>The issuer is considered as valid if it has the same HTTP scheme and authority as the
+        /// authority from the configuration file, has a tenant ID, and optionally v2.0 (this web API
```
mafurman: nit: should this say "(**if** this web API...)" #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,323 @@
...
+
+            if (validIssuerTemplate.Contains("{tenantid}"))
+            {
+                try
+                {
+                    string issuerFromTemplate = validIssuerTemplate.Replace("{tenantid}", tenantId);
+
+                    return issuerFromTemplate == actualIssuer;
+                }
+#pragma warning disable CA1031 // Do not catch general exception types
```
mafurman: We generally put these in the GlobalSuppressions.cs file that's present in each assembly. #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,323 @@
...
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Logging;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Validators
+{
+    /// <summary>
+    /// Generic class that validates either JsonWebTokens or JwtSecurityTokens issued from the Microsoft identity platform (AAD).
```
mafurman: Should this comment say "that **validates the issuer** for either"? #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
mafurman: The logic here for making use of TokenValidationParameters.ConfigurationManager + Configuration looks good. :) 

This should fit well into the most recent Last Known Good (LKG) PR as one of the exceptions it looks for is SecurityTokenInvalidIssuerException in order to determine if LKG fallback or configuration refresh is necessary.

test/Microsoft.IdentityModel.Validators.Tests/MicrosoftIdentityIssuerValidatorTest.cs
```diff
@@ -0,0 +1,432 @@
...
+        {
+            var validator = new AadIssuerValidator(_httpClient, ValidatorConstants.AadIssuer);
+            var tidClaim = new Claim(tidClaimType, ValidatorConstants.TenantIdAsGuid);
+
+            var issClaim = new Claim(ValidatorConstants.ClaimNameIss, ValidatorConstants.AadIssuer);
+            var jwtSecurityToken = new JwtSecurityToken(issuer: ValidatorConstants.AadIssuer, claims: new[] { issClaim, tidClaim });
+
+            var actualIssuers = validator.Validate(ValidatorConstants.AadIssuer, jwtSecurityToken, new TokenValidationParameters() { ValidIssuers = new[] { ValidatorConstants.AadIssuer } });
+
+            Assert.Equal(ValidatorConstants.AadIssuer, actualIssuers);
```
mafurman: nit: We generally like to use our CompareContext and IdentityComparer classes during testing as they allow us to get more information out of test failures (e.g. if the first AssertEqual here fails we won't know anything about the next one). #Resolved

mafurman (approved):
Do we need to add any tests that plug this new IssuerValidator into the JwtSecurityTokenHandler/JsonWebTokenHandler? I know we have tests that use custom validators so it might not be necessary.

test/Microsoft.IdentityModel.Validators.Tests/MicrosoftIdentityIssuerValidatorTest.cs
jennyf19: cool. I didn't see that. i've used IdentityComparer. thanks

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: probably, yes. thanks. fixed.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: makes sense. was following another class that did this, but have added a GS file. thanks again

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: yes! fixed

jmprieur (approved):
LGTM
Thanks @jennyf19

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -0,0 +1,321 @@
...
+            }
+            else
+            {
+                AadAuthorityV1 = aadAuthority.TrimEnd('/');
+                AadAuthorityV2 = AadAuthorityV1 + "/v2.0";
+            }
+        }
+
+        private HttpClient HttpClient { get; }
+        internal string AadIssuerV1 { get; set; }
```
brentschmaltz, line 66: nit: normally we leave a line between these. Change or not, your choice ;-)

brentschmaltz (approved):

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
jennyf19: will change. thanks.

jennyf19: fixed.

PR #1739, opened 2021-10-22
Xiao/singleton for call context

### LINKED ISSUE #1738: Implement singleton for CallContext
Added singleton CallContext.Default to the need to new one up.

ciaozhang:
Added singleton CallContext.Default
Close #1738


src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -239,10 +239,7 @@ public static string ValidateIssuer(string issuer, SecurityToken securityToken,
                 foreach (string str in validationParameters.ValidIssuers)
                 {
                     if (string.IsNullOrEmpty(str))
-                    {
```
jennyf19, line 242: nit: should this be in this PR? I think there is a discussion for this work item in another PR.

ciaozhang: Not sure why this PR pick the commit from other branch, will remove these.

PR #1746, opened 2021-10-28
fix linux build

jennyf19:


jmprieur (approved):
LGTM.
Thanks @jennyf19

Alternatively to having new expressions that previous versions of the compiler were understanding (I must admit that I'm a bit like the older versions of the compilers, btw :-)), we could force a version of the compiler in the .csproj (` <LangVersion>8.0</LangVersion>` ?)
What do we prefer?

PR #1748, opened 2021-10-28
Add Unit Test Samples To Make Negative Testing Easier for Consumers

TimHannMSFT:
* Provide simple shim implementation in both the existing model and older model for token validation
* Generates most common problem token types and gives examples for validation.


TimHannMSFT:
Not to be merged until after the ongoing wilson release

test/Microsoft.IdentityModel.TestUtils/Default.cs
```diff
@@ -276,7 +276,7 @@ public static DateTime Expires
             get => DateTime.Parse(ExpiresString);
         }
 
-
+        // Why do we have this timebomb in our code? shouldn't this just be UtcNow + a year or something so no one ever has to update this?
```
TimHannMSFT: Plan to remove this comment but didn't want to forget to raise it to the team. Seems like I should update this to be more dynamic but perhaps there's a historical reason here?

jennyf19:
Is there a related issue or work item?

---
In reply to: [954209739](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1748#issuecomment-954209739) [](http://example.com/codeflow?ancestors=954209739,954209739)

jennyf19:
> using System;

nit: add license info to all new files

---
Refers to: test/Microsoft.IdentityModel.JsonWebTokens.Tests/TestTokenCreator.cs:1 in ff1e280. [](commit_id = ff1e2801359206a5edbb1efeed5a939248996908, deletion_comment = False)

test/Microsoft.IdentityModel.JsonWebTokens.Tests/TestTokenCreator.cs
```diff
@@ -0,0 +1,218 @@
...
+
+namespace Microsoft.IdentityModel.JsonWebTokens.Tests
+{
+    public class TestTokenCreator
+    {
+        public string Issuer { get; set; }
+        public string Audience { get; set; }
+        public SigningCredentials SigningCredentials { get; set; }
+
+        public string CreateDefaultValidToken()
```
jennyf19: >CreateDefaultValidToken [](http://example.com/codeflow?start=22&length=23)

nit: add "Test" before Token on all of these, just so it's very clear they are test tokens.

test/Microsoft.IdentityModel.SampleTests/SampleTokenValidationClass.cs
```diff
@@ -0,0 +1,67 @@
...
+            var result = ValidateTokenShimWithDeprecatedModel(token, this.TokenValidationParameters);            
+        }
+
+        public void ValidateTokenShim(string token, TokenValidationParameters tokenValidationParameters)
+        {
+            var result = this.JwtHandler.ValidateToken(token, tokenValidationParameters);
+
+            if (!result.IsValid)
+            {
+                throw new SampleTestTokenValidationException("Validaiton Issue Encountered", result.Exception);
```
jennyf19:
```suggestion
                throw new SampleTestTokenValidationException("Validation Issue Encountered", result.Exception);
```

test/Microsoft.IdentityModel.SampleTests/SampleTokenValidationClass.cs
```diff
@@ -0,0 +1,67 @@
...
+        public ClaimsPrincipal ValidateTokenShimWithDeprecatedModel(string token, TokenValidationParameters overrideTokenValidationParameters)
+        {
+            try
+            {
+                SecurityToken validatedToken;
+                return this.JwtSecurityTokenHandler.ValidateToken(token, overrideTokenValidationParameters, out validatedToken);
+            }
+            catch (Exception e)
+            {
+                throw new SampleTestTokenValidationException("Validaiton Issue Encountered", e);
```
jennyf19:
```suggestion
                throw new SampleTestTokenValidationException("Validation Issue Encountered", e);
```

test/Microsoft.IdentityModel.SampleTests/SampleTokenValidationClass.cs
```diff
@@ -0,0 +1,67 @@
...
+        public void ValidateTokenShimWithDeprecatedModel(string token)
+        {
+            var result = ValidateTokenShimWithDeprecatedModel(token, this.TokenValidationParameters);            
+        }
+
+        public void ValidateTokenShim(string token, TokenValidationParameters tokenValidationParameters)
+        {
+            var result = this.JwtHandler.ValidateToken(token, tokenValidationParameters);
+
+            if (!result.IsValid)
```
jennyf19, line 106: can result be null?

TimHannMSFT:
> Is there a related issue or work item?
> 
> In reply to: [954209739](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1748#issuecomment-954209739)

There's an item in ADO tracking Product Backlog Item 1533696: Improve Unit Testability of Wilson to allow for easier negative checks of validation scenarios

test/Microsoft.IdentityModel.SampleTests/SampleTestTokenValidationException.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.IdentityModel.SampleTests
```
jennyf19, line 31: >SampleTests [](http://example.com/codeflow?start=34&length=11)

are the sampleTests for developers to see as an example and use in their own projects? Will you have a wiki page with more info?

jennyf19 (approved):
:shipit:

test/Microsoft.IdentityModel.SampleTests/SampleTestTokenValidationException.cs
TimHannMSFT: A wiki entry is a good idea. I can follow up with that. AppModel team has expressed interest in following something like this model so I'll work with them to produce something consumer teams will find most helpful.

test/Microsoft.IdentityModel.SampleTests/SampleTokenValidationClass.cs
TimHannMSFT: My understanding is no, validate will not return null. I just did a quick review again and I didn't see anywhere where it could. Outside a bug I believe this will always return a TokenValidationResult.

jennyf19 (changes requested):
:clock1:

test/Microsoft.IdentityModel.SampleTests/SampleTestTokenValidationException.cs
jennyf19: will customers be using this code? if so, should it be moved to the product code area, and not in tests, so that it's signed and released with the other package? something like M.IM.TestExtensions? or should it be moved to our AzureAD samples in a new repo?

jmprieur: In both cases, we'd want XML comments on the public api?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/TestTokenCreator.cs
TimHannMSFT: Made it clear in the method doc and it's in the class name. Would rather not make the method name overly verbose. Let me know if you feel strongly here.

test/Microsoft.IdentityModel.SampleTests/SampleTestTokenValidationException.cs
TimHannMSFT: Moved to new MS.IM.TestExtentions project, updated xml documentation across the board.

src/Microsoft.IdentityModel.TestExtentions/Microsoft.IdentityModel.TestExtentions.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\commonTest.props" />
+
+  <PropertyGroup>
+    <AssemblyName>Microsoft.IdentityModel.TestExtentions</AssemblyName>
```
jennyf19: >TestExtentions [](http://example.com/codeflow?start=42&length=14)

spelling: Extensions

src/Microsoft.IdentityModel.TestExtentions/TestTokenCreator.cs
```diff
@@ -0,0 +1,399 @@
...
+            P = Base64UrlEncoder.DecodeBytes("z3zVMYG1_TWwnNs99hyMJI9xs1zrwbV_STv9cqpaBxgPIHqpDyoG68a_VZF2RfuYLKT9ksEfLmgDeBPuM_vwBXSUApRqQo46WOpYUaDTcQJUJcspYQENQDkvtMGTuZQd6wrfGksDkUgnuZZ2CAKjN6vzGGlUkLJ9eEihbVTsCcYpp1sSK_J5HZCu0MZcnEU4oMbWKRHHApkQC8TJkutRw1XjY6s_ABj7M2W2D9CIO-CnUaXxmh21JNaJ2zP0nK1cfWgzvIikVoey7FKbGbOoIBnVf9AZUAVfclBLWI3wfUUVod_1lwLA07-7kvQ6SCEiHpsep3RwONcYJJH3vD3WlQ"),
+            Q = Base64UrlEncoder.DecodeBytes("6l2EkIkJDFV_djyTFdpYbjrmg1lIRa1_oFF1925M0aDAEeJncRYkBOyupE9Ni6vHSTlRGN7I3Vs6V72Jnt_wpKqxz0qdhUseO89_JaqQE65euYq1QyeAjnHg64q4PcLYmBDub9Hn6y0xkTzkMCOUe5ZNGeosb5LW4qQVGyXnyeg9PmwFrBjQwr_Na6-1VrS2HCqzxvQWkOuJdjt3XB4ODXUnze9G-8nx8d2Cb9hC2qT4QxTPCq25i0n2DvIcRs_pD6RyNPg7EYOSH2Alc1vVzj7aYW9Plx69rIrtIFxH0ZwHsHua-AjDg7IZLK_Ghc3RV-9nTPEA2_6q_A_oOvfN9Q"),
+        };
+
+    /// <summary>
```
jennyf19: nit: indentation seems off

src/Microsoft.IdentityModel.TestExtentions/TestTokenCreator.cs
```diff
@@ -0,0 +1,399 @@
...
+            new Claim(JwtRegisteredClaimNames.Exp, EpochTime.GetIntDate(_defaultExpirationTime).ToString(), ClaimValueTypes.String, _defaultTestIssuer, _defaultTestIssuer),
+        };
+
+        /// <summary>
+        /// A set of <see cref="RSAParameters"/> which should be different than what a caller provides as their SigningCredentials.
+        /// </summary>
+        private static RSAParameters _rsaParameters_4096 = new RSAParameters
+        {
+            D = Base64UrlEncoder.DecodeBytes("B15EJtB2qDEtE2z4k6dR-o_fQGoFtQR10FfMWt3XSE44jPYLMKyr3cc0NQkw7DsMpU8Olqo-enIX1BD9vVdWXAY-gR4yxJtRyiVcdAJiNHH6GN8s_GA21Z-6AZR_FTGgmJStEUe-KupSwzYdjo6f5S8dHyaJpLuRhhIx8mLExdhA7543Puwt-MMKj46rcEEKXhe2jNZl1vDgj8-71ZjTvE6HW__invbNX3ZefrZJCD-rDI9SyCkktq9msKfodFtyKfQ0-EYG6rxk7hpAwtilr418N-0I7oYh4OJHC44sRQswL7aCuxHB6qJbkPm7D5QIrrdd4ts5hYNoHd3YJIhr_H09VFnql5CfIyk2NbxHwbZAkGuA00ceBJ0RuIWvrxf-66ZbcNUz2U7l2GEF5qUsy7T4Zi6JtH7I3b4EaO620X-SZ5GXBWmxWBthVi2St3z7vAGl6Nv6F-pk7IjVZ4jj3EJGgCpHZEZObAD7y5IpslBiA4F24qitYfI4RPuncz78V2aKu8Zi6O19PFqQs1Ky3fEV5-fQjPqtu9944omHiwx0Jsjy0SU5MdM6indY75sTKM0xmOFCrF4poGFGJMHW-PJmFy2pCC9O39iz9ki32vJlvYo085a5vrVzKni_fY3HOJKC0NMMlwvAasedT2ZeN1hVVjSqlQwjGQTmDjcx07k"),
```
jennyf19: are these okay to have public?

test/Microsoft.IdentityModel.SampleTests/SampleTokenValidationClass.cs
```diff
@@ -0,0 +1,132 @@
...
+namespace Microsoft.IdentityModel.SampleTests
+{
+    /// <summary>
+    /// A class which trivially wraps Microsoft.IdentityModel token validation.
+    /// </summary>
+    /// <remarks>
+    /// This class exists only as a trivial example of how to leverage Microsoft.IdentityModel token validation as to provide an
+    /// example for how one might want to construct unit tests to confirm validation is handled correctly.
+    /// </remarks>
+    class SampleTokenValidationClass
```
jennyf19, line 45: maybe this can move to a code section of the XML comments in TestTokenCreator

src/Microsoft.IdentityModel.TestExtentions/TestTokenCreator.cs
```diff
@@ -0,0 +1,399 @@
...
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.TestExtentions
+{
+    /// <summary>
+    /// A class responsible for creating test tokens for use in unit testing implementations depending on
+    /// Microsoft.IdentityModel token validation.
+    /// </summary>
+    public class TestTokenCreator
```
jmprieur: Wouldn't we want this class should also propose a method to return the token validation parameters corresponding to the issuer. audience.

src/Microsoft.IdentityModel.TestExtentions/TestTokenCreator.cs
```diff
@@ -0,0 +1,399 @@
...
+using Microsoft.IdentityModel.Json.Linq;
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.TestExtentions
+{
+    /// <summary>
+    /// A class responsible for creating test tokens for use in unit testing implementations depending on
+    /// Microsoft.IdentityModel token validation.
+    /// </summary>
```
jmprieur: I think there would be a lot of value in having the code of the tests sample be added in the `<example><code>` blocks in the XML comment of the class (for `TestWithGeneratedToken`) or of the methods.
See suggestions below

src/Microsoft.IdentityModel.TestExtentions/TestTokenCreator.cs
```diff
@@ -0,0 +1,399 @@
...
+        public string CreateDefaultValidToken()
+        {
+            var tokenDescriptor = CreateTokenDescriptorWithInstanceOverrides();
+            return CreateToken(tokenDescriptor);
+        }
+
+        /// <summary>
+        /// Creates a test token with a signature that doesn't match the payload.
+        /// </summary>
+        /// <returns>A test JWS token with a signature which doesn't match the payload.</returns>
```
jmprieur:
```suggestion
        /// <returns>A test JWS token with a signature which doesn't match the payload.</returns>
        /// <example>
        ///  The following code shows how the class under test handles a token with a malformed signature.
        /// <code>
        /// [Fact]
        /// public void TokenWithBadSignature()
        /// {
        ///     TestWithGeneratedToken(
        ///         testTokenCreator.CreateTokenWithInvalidSignature,
        ///         typeof(ArgumentException),
        ///         "IDX14111");
        ///  }
        /// </code></example>
```

This enables developers to see how to use the API from the reference documentation.
cc: @mmacy

jmprieur (commented):
LGTM
I left a few suggestions, @TimHannMSFT

src/Microsoft.IdentityModel.TestExtentions/TestTokenCreator.cs
TimHannMSFT: Is there an advantage to making it public? I'd rather keep the surface as minimal as possible to make any changes in the future easier.

src/Microsoft.IdentityModel.TestExtentions/TestTokenCreator.cs
TimHannMSFT: IMO, no, that's not a good idea. That would encourage tautological unit tests. We want them to use the TVPs they'd actually use in livesite to understand if they're turning off e.g. issuer validation. If we always give them the TVPs the tests may well pass in unit tests but their real TVPs won't actually do the validation we're hoping they will.

src/Microsoft.IdentityModel.TestExtentions/TestTokenCreator.cs
TimHannMSFT: Hmm this example assumes folks are using Xunit and abstracts away the actual test code (i.e. this is just a wrapper that doesn't give a great idea of the actual logic they'd want to implement.)

I think the overall suggestion is a good one. Let me think about how best to share code examples that would be most useful.

test/Microsoft.IdentityModel.SampleTests/SampleTokenValidationClass.cs
TimHannMSFT: Tried to improve xml comments for both classes.

src/Microsoft.IdentityModel.TestExtentions/Microsoft.IdentityModel.TestExtentions.csproj
TimHannMSFT: Good catch, thanks!

src/Microsoft.IdentityModel.TestExtensions/Microsoft.IdentityModel.TestExtensions.csproj
```diff
@@ -0,0 +1,29 @@
...
+
+  <Import Project="..\..\build\commonTest.props" />
+
+  <PropertyGroup>
+    <AssemblyName>Microsoft.IdentityModel.TestExtensions</AssemblyName>
+    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)..\..\build\35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
+    <DelaySign>true</DelaySign>
+    <Description>Microsoft.IdentityModel.TestExtensions</Description>
+    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
+    <PackageId>Microsoft.IdentityModel.TestExtensions</PackageId>
```
jennyf19, line 11: Extensions needs to be updated in the strong name bypass files as well.

src/Microsoft.IdentityModel.TestExtensions/TestTokenCreator.cs
```diff
@@ -0,0 +1,455 @@
...
+    /// to validate a trivial token validation class that depends on Microsoft.IdentityModel's token validation methods.
+    /// </remarks>
+    /// <example>
+    /// The following provides an example for how this class could be leveraged using a common testing framework Xunit. The core concepts
+    /// will be applicable to unit testing using any framework however.
+    ///
+    /// The example imagines a class, ClassWithMicrosoftIdentityModelDependency, which exposes ValidateToken, a method calling the
+    /// Microsoft.IdentityModel library and GetTokenValidationParameters which retrieves the <see cref="TokenValidationParameters"/>
+    /// the code under test actually uses. Note that it's important to use the real <see cref="TokenValidationParameters"/> since that
+    /// will allow the unit tests to actually confirm if there's a hole validation (e.g. certain important validation is disabled,
```
jennyf19:
```suggestion
    /// will allow the unit tests to actually confirm if there's a hole in the validation (e.g. certain important validation is disabled,
```

src/Microsoft.IdentityModel.TestExtensions/TestTokenCreator.cs
```diff
@@ -0,0 +1,455 @@
...
+            new Claim(JwtRegisteredClaimNames.Exp, EpochTime.GetIntDate(_defaultExpirationTime).ToString(), ClaimValueTypes.String, _defaultTestIssuer, _defaultTestIssuer),
+        };
+
+        /// <summary>
+        /// A set of <see cref="RSAParameters"/> which should be different than what a caller provides as their SigningCredentials.
+        /// </summary>
+        private static RSAParameters _rsaParameters_4096 = new RSAParameters
+        {
+            D = Base64UrlEncoder.DecodeBytes("B15EJtB2qDEtE2z4k6dR-o_fQGoFtQR10FfMWt3XSE44jPYLMKyr3cc0NQkw7DsMpU8Olqo-enIX1BD9vVdWXAY-gR4yxJtRyiVcdAJiNHH6GN8s_GA21Z-6AZR_FTGgmJStEUe-KupSwzYdjo6f5S8dHyaJpLuRhhIx8mLExdhA7543Puwt-MMKj46rcEEKXhe2jNZl1vDgj8-71ZjTvE6HW__invbNX3ZefrZJCD-rDI9SyCkktq9msKfodFtyKfQ0-EYG6rxk7hpAwtilr418N-0I7oYh4OJHC44sRQswL7aCuxHB6qJbkPm7D5QIrrdd4ts5hYNoHd3YJIhr_H09VFnql5CfIyk2NbxHwbZAkGuA00ceBJ0RuIWvrxf-66ZbcNUz2U7l2GEF5qUsy7T4Zi6JtH7I3b4EaO620X-SZ5GXBWmxWBthVi2St3z7vAGl6Nv6F-pk7IjVZ4jj3EJGgCpHZEZObAD7y5IpslBiA4F24qitYfI4RPuncz78V2aKu8Zi6O19PFqQs1Ky3fEV5-fQjPqtu9944omHiwx0Jsjy0SU5MdM6indY75sTKM0xmOFCrF4poGFGJMHW-PJmFy2pCC9O39iz9ki32vJlvYo085a5vrVzKni_fY3HOJKC0NMMlwvAasedT2ZeN1hVVjSqlQwjGQTmDjcx07k"),
```
jennyf19, line 140: are these okay to have here?

TimHannMSFT: Are you able to flush out your concern? These are fake and are already public in GitHub. Is there a place you feel they would be better placed?

src/Microsoft.IdentityModel.TestExtensions/Microsoft.IdentityModel.TestExtensions.csproj
TimHannMSFT: Doh, thanks for catching this.

src/Microsoft.IdentityModel.TestExtensions/TestTokenCreator.cs
jennyf19: that's fine then. wasn't sure.

src/Microsoft.IdentityModel.TestExtensions/Microsoft.IdentityModel.TestExtensions.csproj
jennyf19: i think you need to add the new samples testing assembly as well, to both files. I had to add Test.validators to both as well... @GeoK  would know. i just noticed.

jennyf19 (approved):
:shipit: 
i think this looks good. just one remaining issue on the strongNameByPass to add the Samples test project in both files as well.

Issue #1752, opened 2021-10-28
[Bug] AadIssuerValidator doesn't work well with multiple auth schemes in ASP.NET Core


galhag:
**Which version of Microsoft Identity Web are you using?**
Microsoft.Identity.Web 1.18

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [x] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
We wanted to our production apps to start use Microsoft.Identity.Web,
Out flow contains S2S and the web api, should accept tokens from two tenants.
We currently use the AddJwtBearer, against two tenants, using two schemas.
When we try to achieve the same using the Microsoft.Identity.Web we are not able to get both tenants to work.

**Repro**
We have two tenants in AAD, tenantA, tenantB
and our appsettings contains two sections AzureAdA, AzureAdB with the relevant details.
We use S2S tokens, with roles for authorization, and constructing the relevant policy two work against both tenants.
When we set the app to work with only section A it works and accepts the tenantA tokens,
When we set it with section B only, it works as well and accepts the tenantB tokens,.
When we use both, it fails for tenantB.

asp dotnet core 3.1 web api with aad authentication
install the Microsoft.Identity.Web  package 
appsettings should contain two AzureAd sections (AzureAdA & AzureAdB) for two aad tenants
Run the app, and try to access the api with tenantA token,
then with tenantB token.
tenantB fails.

In the below code, if you change tenants to only have one entry, it works against the single tenant.

Add the below ConfigureServices.

```csharp
public void ConfigureServices(IServiceCollection services)
        {

            var tenants= new List<(string section, string schema)>
            {
                ("AzureAdA", "schemaA"),
                ("AzureAdB", "schemaB")
            };
            
            var auth = services.AddAuthentication();
            foreach(var (section, schema) in tenants)
            {
                auth.AddMicrosoftIdentityWebApi(Configuration, section, schema);
                services.AddControllers();
                services.Configure<JwtBearerOptions>(schema, options =>
                {
                    options.Events.OnAuthenticationFailed += context =>
                    {
[...]

PR #1753, opened 2021-11-03
handle multi-schemes in AadIssuerValidator

jennyf19:
#1752 

We need to use the full authority, not just the authority host for multi-auth scheme scenarios.


jmprieur (approved):
LGTM
Thanks @jennyf19. 

I had not though of this scenario involving multiple schemes.

ciaozhang (approved):

PR #1757, opened 2021-11-04
Introduced new property to store log message and EventLevel

sruke:


src/Microsoft.IdentityModel.Tokens/CallContext.cs
```diff
@@ -65,8 +66,16 @@ public CallContext(Guid activityId)
         /// <summary>
         /// The collection of logs associated with a request. Use <see cref="CaptureLogs"/> to control capture.
         /// </summary>
+        /// <remarks>
+        /// Will be deprecated in the next major release. Start moving towards using <see cref="LogData"/> instead
```
brentschmaltz, line 70: Do we have to remove this api to add the new one?

sruke: This property doesn't capture the EventLevel so the logger service cannot categorize the incoming logs.
As LogData will have everything Logs property does and the EventLevel, George and I thought we could consider moving away from this property in the future if it makes sense. Hence the deprecated comment. SAL will continue to support both but there will some duplicate functionality until both properties are around.

brentschmaltz: I am not understanding why we need the EventLevel.
Logs get captured according to the level the user set, why does the reviewer of the log data need to know level of each individual log entry?

GeoK: It's so that developers can forward these logs to their logging implementations with the corresponding Event/Log levels, e.g. ILogger. Current Logs structure doesn't have enough fidelity as it contains only strings.

GeoK (approved):

src/Microsoft.IdentityModel.Tokens/CallContext.cs
brentschmaltz: OK, thanks.

PR #1765, opened 2021-11-06
Reduce string allocations

lechu445:
String interpolation allocates a new string on the heap.
https://github.com/JosefPihrt/Roslynator/blob/master/docs/analyzers/RCS1197.md

### DIFF
--- a/test/Microsoft.IdentityModel.TestUtils/XmlGenerator.cs
+++ b/test/Microsoft.IdentityModel.TestUtils/XmlGenerator.cs
@@ -48,7 +48,9 @@ namespace Microsoft.IdentityModel.TestUtils
 
         public static string Generate(XmlEement dataElement)
         {
-            var stringBuilder = new StringBuilder($"<{dataElement.Name}>");
+            int initialCapacity = (dataElement.Name.Length + 3) * 2;
+            var stringBuilder = new StringBuilder(initialCapacity);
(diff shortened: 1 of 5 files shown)


test/Microsoft.IdentityModel.TestUtils/XmlGenerator.cs
```diff
@@ -48,7 +48,8 @@ public XmlEement(string name, object value)
 
         public static string Generate(XmlEement dataElement)
         {
-            var stringBuilder = new StringBuilder($"<{dataElement.Name}>");
+            var stringBuilder = new StringBuilder((dataElement.Name.Length + 3) * 2);
```
brentschmaltz: @lechu445 why do you want to set the size?

lechu445: @brentschmaltz  because it's better for performance than no capacity. When no capacity is defined, the default 16 is used. In this PR I calculated the minimum size of the produced output string.
 
More information why setting capacity is important for performance:  
https://github.com/u2uconsult/codeanalyzers/blob/master/docs/U2U1108.md  
https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder?view=net-5.0#memory-allocation

brentschmaltz (approved):

brentschmaltz:
@lechu445 thanks again :-)

PR #1766, opened 2021-11-08
Fixed build failure

sruke:
IdentityModelEventSource.ShowPII property change is taking effect globally causing other unit tests to fail. 

Modified my unit tests to remove the property and associated logic. Main logic to test logging null arguments and exceptions remains intact

### DIFF
--- a/test/Microsoft.IdentityModel.Logging.Tests/PIITests.cs
+++ b/test/Microsoft.IdentityModel.Logging.Tests/PIITests.cs
@@ -199,23 +199,11 @@ namespace Microsoft.IdentityModel.Logging.Tests
 
             string algorithm = null;
 
-            // Testset 1 with PII logging off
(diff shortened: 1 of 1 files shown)


TimHannMSFT:
Are there any plans to add different testing or expose some hooks to re-capture the coverage you were hoping to get here?

TimHannMSFT (approved):
:shipit:

sruke:
> Are there any plans to add different testing or expose some hooks to re-capture the coverage you were hoping to get here?

Main goal for this PR was to fix the build failure so that it wouldn't gate upcoming Wilson releases. SAL is able to utilize 'using' keyword to localize a property change to a single unit test but this didn't work for Wilson. I'll check-in with the team on alternatives. I would like to recapture the coverage if possible.

PR #1768, opened 2021-11-09
Reduce potential large strings allocations

lechu445:


brentschmaltz:
@lechu445 going to skip this one as we are moving completely away from strings for targets 461 and above.
see this topic branch so get an idea of what we are up to: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/tree/brentsch/SystemTextJson

PR #1770, opened 2021-11-09
Improve formatting in token related classes

lechu445:
Only cleanup. No functional changes.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -41,13 +41,13 @@ using TokenLogMessages = Microsoft.IdentityModel.Tokens.LogMessages;
 namespace Microsoft.IdentityModel.JsonWebTokens
 {
     /// <summary>
-    /// A <see cref="SecurityTokenHandler"/> designed for creating and validating Json Web Tokens. 
+    /// A <see cref="SecurityTokenHandler"/> designed for creating and validating Json Web Tokens.
     /// See: http://tools.ietf.org/html/rfc7519 and http://www.rfc-editor.org/info/rfc7515.
     /// </summary>
     public class JsonWebTokenHandler : TokenHandler
     {
         /// <summary>
-        /// Gets the Base64Url encoded string representation of the following JWT header: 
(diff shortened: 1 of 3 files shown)


jennyf19 (approved):
LGTM.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -41,13 +41,13 @@
...
     /// <summary>
-    /// A <see cref="SecurityTokenHandler"/> designed for creating and validating Json Web Tokens. 
+    /// A <see cref="SecurityTokenHandler"/> designed for creating and validating Json Web Tokens.
     /// See: http://tools.ietf.org/html/rfc7519 and http://www.rfc-editor.org/info/rfc7515.
     /// </summary>
     public class JsonWebTokenHandler : TokenHandler
     {
         /// <summary>
-        /// Gets the Base64Url encoded string representation of the following JWT header: 
+        /// Gets the Base64Url encoded string representation of the following JWT header:
```
brentschmaltz, line 50: Should be a period and the end instead of at ':'.

brentschmaltz:
@lechu445 there are some conflicts here, you will need to rebase against dev.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
lechu445: I think this is not a typo but just a long text wrapped after ':' for better readability. See how Visual Studio renders it: 
image
Should I perform the change that you suggested anyway?

jennyf19: i would leave it as you have it @lechu445  w/the `:`

brentschmaltz (approved):
@lechu445 thanks again :-)

brentschmaltz (changes requested):
@lechu445 thanks again :-)

Needs a rebase.

lechu445:
@brentschmaltz I have rebased it onto the newest `brentsch/SystemTextJson`

brentschmaltz:
this is a topic branch that i frequently rebase and squash.
not a good place to merge into.

PR #1771, opened 2021-11-10
added the MaximumSizeCache and simplified the EventBasedLRUCache

### LINKED ISSUE #1742: Add a non-LRU Provider Cache
The event based LRU cache implementation is designed to keep the most frequently used providers in the cache and evict the least used ones. This is more complicated and requires starting background tasks to process events. For some users who have a small number of providers this might not be needed as the cache is never full.
This task is for implementing a simpler and more lightweight non-LRU cache.

dannybtsai:
The event based LRU cache implementation is designed to keep the most frequently used providers in the cache and evict the least used ones. For some users who have a small number of providers this might not be needed as the cache is never full.
This task is for implementing a simpler and more light weight non-LRU cache. The EventBasedLruCache is also simplified including:

- Remove the expired items in the event queue task, the timer is no longer needed (otherwise it needs to be disposed of).
- The event queue task exits as soon as all events have been processed. The task is started using thread pool thread so the cost of restarting tasks should be minimum.
- The provider caches (signing and verifying signature providers) are created based on the new CryptoProviderCacheOptions.CacheType.
- Some cleanup.


src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
```diff
@@ -0,0 +1,264 @@
...
+            var newCacheSize = CalculateNewCacheSize();
+            while (_map.Count > newCacheSize)
+            {
+                var item = _map.FirstOrDefault();
+                if (_map.TryRemove(item.Key, out var cacheItem))
+                    OnItemRemoved?.Invoke(cacheItem.Value);
+            }
+
+            // reset the _compactionState to CompactionStateNotRunning
+            _compactionState = CompactionStateNotRunning;
```
henrik-me: >_compactionState = CompactionStateNotRunning; [](http://example.com/codeflow?start=12&length=45)

seems like this should be set to running when starting the compacting operation

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
```diff
@@ -0,0 +1,264 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    internal class HashCache<TKey, TValue> : IProviderCache<TKey, TValue>
```
henrik-me: >HashCache [](http://example.com/codeflow?start=19&length=9)

since this one removes from the front, should this be called FIFOCache or similar? Or we can stick with SimpleEvictionCache

src/Microsoft.IdentityModel.Tokens/ProvidersCache/ProviderCacheItem.cs
```diff
@@ -0,0 +1,62 @@
...
+//
+//------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    internal class ProviderCacheItem<TKey, TValue>
```
henrik-me: >Provider [](http://example.com/codeflow?start=19&length=8)

Why provider?

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
```diff
@@ -37,6 +37,22 @@ namespace Microsoft.IdentityModel.Tokens
...
+            /// <summary>
+            /// EventBasedLRUCache
+            /// </summary>
+            LRU,
+
+            /// <summary>
+            /// Hash-based cache
+            /// </summary>
+            Hash
+        };
```
henrik-me: separate file and not inside the class

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
```diff
@@ -0,0 +1,264 @@
...
+        internal ItemRemoved OnItemRemoved { get; set; }
+
+        // _capacity is used for the newCacheSize calculation in the case where the cache is experiencing overflow
+        protected readonly int _capacity;
+
+        // the percentage of the cache to be removed when _maxCapacityPercentage is reached
+        protected readonly double _compactionPercentage = .20;
+
+        // When the current cache size gets to this percentage of _capacity, _compactionPercentage% of the cache will be removed.
+        protected readonly double _maxCapacityPercentage = .95;
```
henrik-me: should these be configurable?

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
```diff
@@ -0,0 +1,264 @@
...
+        /// <summary>
+        /// Remove items from the map by the desired compaction percentage.
+        /// Since this is a simple hash-based cache that does not track the last used time of items, simply remove the desired number of items.
+        /// This should be a private method.
+        /// </summary>
+        private void CompactCache(object state)
+        {
+            // use the _capacity for the newCacheSize calculation in the case where the cache is experiencing overflow
+            var newCacheSize = CalculateNewCacheSize();
+            while (_map.Count > newCacheSize)
```
henrik-me: >_map.Count [](http://example.com/codeflow?start=19&length=10)

if a process is filling this up very fast we should have a signal for that situation. E.g. if we never get out of this state the service should know. #operability

Start time for compacting and end time for compacting. total time compacting should likely also be recorded. Average time compacting. Number of times compacting e.g. per hour/day would also be good numbers. 

Tim suggestions here?

henrik-me:
Can you pls. share perf. and stress numbers for both caches?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -59,31 +59,39 @@ public InMemoryCryptoProviderCache() : this(new CryptoProviderCacheOptions())
...
+            {
+                _signingSignatureProviders = new EventBasedLRUCache<string, SignatureProvider>(_cryptoProviderCacheOptions.SizeLimit, StringComparer.Ordinal) { OnItemRemoved = (SignatureProvider signatureProvider) => signatureProvider.CryptoProviderCache = null };
+                _verifyingSignatureProviders = new EventBasedLRUCache<string, SignatureProvider>(_cryptoProviderCacheOptions.SizeLimit, StringComparer.Ordinal) { OnItemRemoved = (SignatureProvider signatureProvider) => signatureProvider.CryptoProviderCache = null };
+            }
+            else
+            {
+                _signingSignatureProviders = new HashCache<string, SignatureProvider>(_cryptoProviderCacheOptions.SizeLimit, StringComparer.Ordinal) { OnItemRemoved = (SignatureProvider signatureProvider) => signatureProvider.CryptoProviderCache = null };
+                _verifyingSignatureProviders = new HashCache<string, SignatureProvider>(_cryptoProviderCacheOptions.SizeLimit, StringComparer.Ordinal) { OnItemRemoved = (SignatureProvider signatureProvider) => signatureProvider.CryptoProviderCache = null };
+            }
         }
```
henrik-me, line 74: suggestion, extract to factory class?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -59,31 +59,39 @@ public InMemoryCryptoProviderCache() : this(new CryptoProviderCacheOptions())
...
-            _signingSignatureProviders = new EventBasedLRUCache<string, SignatureProvider>(cryptoProviderCacheOptions.SizeLimit, options, StringComparer.Ordinal, false) { OnItemRemoved = (SignatureProvider signatureProvider) => signatureProvider.CryptoProviderCache = null };
-            _verifyingSignatureProviders = new EventBasedLRUCache<string, SignatureProvider>(cryptoProviderCacheOptions.SizeLimit, options, StringComparer.Ordinal, false) { OnItemRemoved = (SignatureProvider signatureProvider) => signatureProvider.CryptoProviderCache = null };
+            _cryptoProviderCacheOptions = cryptoProviderCacheOptions ?? throw LogHelper.LogArgumentNullException(nameof(cryptoProviderCacheOptions));
+            CreateProviderCache();
+        }
+
+        /// <summary>
+        /// Create the _signingSignatureProviders and _verifyingSignatureProviders caches based on the specified cache type from _cryptoProviderCacheOptions.
+        /// </summary>
+        protected void CreateProviderCache()
```
henrik-me: >protected void CreateProviderCache() [](http://example.com/codeflow?start=8&length=36)

why protected?

henrik-me:
@bgavrilMS : simple cache / lru cache : let's explore how we best leverage in MSAL

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
```diff
@@ -44,7 +44,7 @@ public class CryptoProviderCacheTests
     {
         public static Func<CryptoProviderCache> CreateCacheForTesting = new Func<CryptoProviderCache>(() =>
         {
-            return new InMemoryCryptoProviderCache(new CryptoProviderCacheOptions(), TaskCreationOptions.None, 50);
+            return new InMemoryCryptoProviderCache(new CryptoProviderCacheOptions(), 50);
```
henrik-me: >50 [](http://example.com/codeflow?start=85&length=2)

Wondering if it's still needed to set this value? This was needed in the first lru implementation but many things have changed since. 

@mafurman thoughts?

henrik-me:
@bgavrilMS is added to the review. #Closed

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
dannybtsai: This is now created by the CryptoProviderCacheFactory factory.

src/Microsoft.IdentityModel.Tokens/ProvidersCache/ProviderCacheItem.cs
dannybtsai: Because the cached items are signing and verifying signature providers.

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
dannybtsai: I welcome suggestions on the name, Hash clearly says it is hash-based but SimpleEvictionCache sounds good too.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
```diff
@@ -37,6 +37,22 @@ namespace Microsoft.IdentityModel.Tokens
     /// </summary>
     public class CryptoProviderCacheOptions
     {
+        /// <summary>
+        /// Cache types.
+        /// </summary>
+        public enum ProviderCacheType
+        {
+            /// <summary>
+            /// EventBasedLRUCache
```
mafurman: I think a more descriptive comment for what these types mean could be helpful.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
```diff
@@ -60,5 +76,10 @@ public int SizeLimit
                 _sizeLimit = (value > 10) ? value : throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(SizeLimit), LogHelper.FormatInvariant(LogMessages.IDX10901, value)));
             }
         }
+
+        /// <summary>
+        /// The desired cache type for caching providers. Default to LRU.
```
mafurman: nit: should be "Defaults"

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -243,67 +207,74 @@ internal int RemoveExpiredValues()
...
-        /// Note: tasks are based on thread pool so the overhead should not be huge but we should still try to minimize it.
+        /// This method is called after an item is added to the event queue, and the event queue task needs to be started if not running (_eventQueueTaskState != EventQueueTaskRunning).
+        /// Setting _eventQueueTaskState to EventQueueTaskRunning should prevent multiple task being started.
         /// </summary>
-        /// <returns>the time when the event queue task should end</returns>
-        private DateTime SetTaskEndTime()
+        private void StartEventQueueTaskIfNotRunning()
         {
-            return DateTime.UtcNow.AddSeconds(EventQueueTaskIdleTimeoutInSeconds);
+            // Setting _eventQueueTaskState to EventQueueTaskDoNotStop here will force the event queue task in EventQueueTaskAction to continue even it has past the _eventQueueTaskEndTime.
```
mafurman: _eventQueueTaskEndTime doesn't exist anymore, but it's referenced in few different comments.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -243,67 +207,74 @@ internal int RemoveExpiredValues()
...
+            // It is mainly to prevent a rare (but theoretically possible) scenario caused by context switching
+            // For example:
+            //   1. the task execution in EventQueueTaskAction() checks _eventQueueTaskStopTime and it has already passed (ready to exit)
```
mafurman: I guess this comment needs to be modified to say something along the lines of "the task execution in EventQueueTaskAction() runs out of items to process and is ready to exit"?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -59,31 +59,39 @@ public InMemoryCryptoProviderCache() : this(new CryptoProviderCacheOptions())
...
         }
 
         /// <summary>
         /// Creates a new instance of <see cref="InMemoryCryptoProviderCache"/> using the specified <paramref name="cryptoProviderCacheOptions"/>.
         /// </summary>
         /// <param name="cryptoProviderCacheOptions">The options used to configure the <see cref="InMemoryCryptoProviderCache"/>.</param>
-        /// <param name="options">Options used to create the event queue thread.</param>
         /// <param name="tryTakeTimeout">The time used in ms for the timeout interval of the event queue. Defaults to 500 ms.</param>
-        internal InMemoryCryptoProviderCache(CryptoProviderCacheOptions cryptoProviderCacheOptions, TaskCreationOptions options, int tryTakeTimeout = 500)
+        internal InMemoryCryptoProviderCache(CryptoProviderCacheOptions cryptoProviderCacheOptions, int tryTakeTimeout = 500)
```
mafurman: tryTakeTimeout isn't used anywhere

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -234,7 +242,7 @@ public override bool TryRemove(SignatureProvider signatureProvider)
 
         /// <summary>
         /// Calls <see cref="Dispose(bool)"/> and <see cref="GC.SuppressFinalize"/>
-        /// Note: the EventBasedLRUCache is no longer being disposed of, but since this is a public class and can be used as base class of
+        /// Note: the cache is no longer being disposed of, but since this is a public class and can be used as base class of
```
mafurman: If we no longer have IDisposable on the LRU/Hash cache, do we still need to have IDisposable on the InMemoryCryptoProviderCache?

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
mafurman: I like FIFOCache since it's clear on exactly what happens when the cache needs to evict items.

src/Microsoft.IdentityModel.Tokens/ProvidersCache/ProviderCacheItem.cs
mafurman: Could this class eventually be reused for caching other objects? Wondering if leaving the name as the more generic 'CacheItem' makes sense...

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderCacheTests.cs
mafurman: I commented on this earlier; the tryTakeTimeout is no longer relevant. :)

test/Microsoft.IdentityModel.Tokens.Tests/HashCashTests.cs
```diff
@@ -0,0 +1,246 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Reflection;
+using Microsoft.IdentityModel.TestUtils;
+using Xunit;
+using System.Diagnostics;
+
```
mafurman: The title of this file is Hash**Cash**Tests 🤑 and not Hash**Cache**Tests.

mafurman (changes requested):

src/Microsoft.IdentityModel.Tokens/ProvidersCache/ProviderCacheItem.cs
dannybtsai: I think this class is generic that CacheItem sounds good.

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
dannybtsai: This is now renamed to MaximumSizeCache as suggested.

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
dannybtsai: I have made these values configurable via the CryptoProviderCacheOptions instead of passing them individually.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
dannybtsai: The provider cache creation has been moved to a factory. Both signingSignatureProviders and verifyingSignatureProviders caches are now injected in the new internal constructor. However, the public constructors can't be changed (to internal) and both provider caches are still being created instead of being injected.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
dannybtsai: Yeah, you are right, IDisposable is not needed any more. However, if I removed it then the ApiCompat complained that it was a breaking change as some users could have tried to dispose of the object via the Dispose().

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
bgavrilMS: Yes, doesn't seem to be set to running.

bgavrilMS:
> The event based LRU cache implementation is designed to keep the most frequently used providers in the cache and evict the least used ones. For some users who have a small number of providers this might not be needed as the cache is never full. This task is for implementing a simpler and more light weight non-LRU cache. 

As an app developer, why would I chose this simple HashCache over the LRU one? Your comments seem to suggest that I would use it when I don't have many items in the cache. Are you implying that the LRU cache has worse perf characteristics than the Hash one for small cache size? If that is the case, it should be made clear in the comments, maybe add some numbers.

dannybtsai:
> > The event based LRU cache implementation is designed to keep the most frequently used providers in the cache and evict the least used ones. For some users who have a small number of providers this might not be needed as the cache is never full. This task is for implementing a simpler and more light weight non-LRU cache.
> 
> As an app developer, why would I chose this simple HashCache over the LRU one? Your comments seem to suggest that I would use it when I don't have many items in the cache. Are you implying that the LRU cache has worse perf characteristics than the Hash one for small cache size? If that is the case, it should be made clear in the comments, maybe add some numbers.

The new cache should be more lightweight as it skips the event queue processing part and does not have the linkedlist. Right now, we don't have numbers for comparison yet.

src/Microsoft.IdentityModel.Tokens/Cache/CacheItem.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    internal class CacheItem<TKey, TValue>
+    {
+        internal TKey Key { get; }
+        internal TValue Value { get; set; }
```
brentschmaltz, line 37: Do we need setters?

src/Microsoft.IdentityModel.Tokens/Cache/CacheItem.cs
```diff
@@ -0,0 +1,61 @@
...
+        internal DateTime ExpirationTime { get; set; }
```
brentschmaltz, line 38: What is the logic if this is not set?
Could the constructor that takes: 'key, value' call the other ctor with a minimum value?

src/Microsoft.IdentityModel.Tokens/CryptoProviderCache.cs
```diff
@@ -37,7 +37,7 @@ public abstract class CryptoProviderCache
         /// </summary>
         /// <param name="signatureProvider">the <see cref="SignatureProvider"/> to create the key for.</param>
         /// <returns>the cache key to use for finding a <see cref="SignatureProvider"/>.</returns>
-        protected abstract string GetCacheKey(SignatureProvider signatureProvider);
+        protected internal abstract string GetCacheKey(SignatureProvider signatureProvider);
```
brentschmaltz, line 40: This is breaking.

src/Microsoft.IdentityModel.Tokens/Cache/RandomEvictCache.cs
```diff
@@ -0,0 +1,272 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    internal class RandomEvictCache<TKey, TValue> : IProviderCache<TKey, TValue>
```
brentschmaltz: Suggest the name: MaximumSizeCache or UpperBoundCache

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -243,67 +203,75 @@ internal int RemoveExpiredValues()
...
+            {
+                Task.Run(EventQueueTaskAction);
+            }
         }
 
-        public void SetValue(TKey key, TValue value)
+        #region public methods/interface implementation
+
+        /// <inheritdoc/>
+        public override void SetValue(TKey key, TValue value)
```
brentschmaltz: I don't think this method is needed.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
dannybtsai: @mafurman , I updated the comments, can you check if they make sense?

src/Microsoft.IdentityModel.Tokens/Cache/IProviderCache.cs
```diff
@@ -0,0 +1,80 @@
...
+using System;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// The provider interface.
+    /// </summary>
+    /// <typeparam name="TKey">The key of a provider, used for cache operations.</typeparam>
+    /// <typeparam name="TValue">The value of a provider.</typeparam>
+    public interface IProviderCache<TKey, TValue>
```
brentschmaltz: I am not understanding the need for this interface or make it public.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
dannybtsai: removed

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -102,17 +103,18 @@ public CryptoProviderFactory(CryptoProviderCache cache)
         /// Constructor that creates a deep copy of given <see cref="CryptoProviderFactory"/> object.
         /// </summary>
         /// <param name="other"><see cref="CryptoProviderFactory"/> to copy from.</param>
-        public CryptoProviderFactory(CryptoProviderFactory other)
+        public CryptoProviderFactory(CryptoProviderFactory other) : this()
         {
             if (other == null)
                 throw LogHelper.LogArgumentNullException(nameof(other));
 
-            CryptoProviderCache = new InMemoryCryptoProviderCache() { CryptoProviderFactory = this };
```
brentschmaltz, line 110: Where will the cache come from?

dannybtsai: public CryptoProviderFactory(CryptoProviderFactory other) : this()

Calling the other (default) constructor will call the CryptoProviderCacheFactory to create a default (LRU) cache.
I am trying to remove dependencies and inject them instead.

src/Microsoft.IdentityModel.Tokens/Cache/CacheItem.cs
dannybtsai: The EventBasedLRUCache.SetValue() sets the Value/ExpirationTime and I certainly think that is the right place for the business logic. Another way to look at this is that a caller decides when an item should expire, not the item itself.

src/Microsoft.IdentityModel.Tokens/Cache/IProviderCache.cs
dannybtsai: This interface is meant to abstract away the details of cache implementations. Otherwise, we would force all caches to inherit from a single base class. It is now changed to internal, thanks for pointing that out.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
dannybtsai: Yeah, that is a goof catch, thanks.

src/Microsoft.IdentityModel.Tokens/Cache/RandomEvictCache.cs
dannybtsai: This is renamed to MaximumSizeCache as suggested.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCache.cs
dannybtsai: @brentschmaltz, how can this be breaking? According MS Doc: "A protected internal member is accessible from the current assembly or from types that are derived from the containing class.". I made the change so we don't need to create the CryptoProviderCacheTests.InMemoryCryptoProviderCachePublic class in order to test the protected methods. If this is not a desired practice, I can change it back.

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
dannybtsai: This while loop is similar to the while loop in the EventBasedLRUCache event queue task, except that it should be faster as the operations are simpler. I have unit tests that covers the cache overflow scenarios with multiple threads and things work well. We can certainly enhance this with signals but that should be in a separate PR.

src/Microsoft.IdentityModel.Tokens/Cache/CacheItem.cs
mafurman: I could be misunderstanding part of this comment, but I think Brent was asking if we could have `internal CacheItem(TKey key, TValue value)` call into `internal CacheItem(TKey key, TValue value, DateTime expirationTime)` to avoid code duplication.

src/Microsoft.IdentityModel.Tokens/Cache/CacheItem.cs
```diff
@@ -31,20 +31,19 @@
...
         {
             Key = key ?? throw LogHelper.LogArgumentNullException(nameof(key));
             Value = value ?? throw LogHelper.LogArgumentNullException(nameof(value));
         }
 
-        internal ProviderCacheItem(TKey key, TValue value, DateTime expirationTime)
+        internal CacheItem(TKey key, TValue value, DateTime expirationTime)
         {
             Key = key ?? throw LogHelper.LogArgumentNullException(nameof(key));
             Value = value ?? throw LogHelper.LogArgumentNullException(nameof(value));
```
mafurman: Should we be throwing if ExpirationTime is null for this particular constructor?

src/Microsoft.IdentityModel.Tokens/Cache/CryptoProviderCacheFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+//
+//------------------------------------------------------------------------------
+
+using System;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// The provider cache factory that creates crypto provider caches.
```
mafurman: Maybe add references to the types of caches that are created in this comment as well.

src/Microsoft.IdentityModel.Tokens/Cache/CryptoProviderCacheFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    internal class CryptoProviderCacheFactory
+    {
+        internal static CryptoProviderCache Create() => Create(new CryptoProviderCacheOptions());
+
+        /// <summary>
+        /// Create a new instance of InMemoryCryptoProviderCache and the _signingSignatureProviders and _verifyingSignatureProviders caches based on the cache type in _cryptoProviderCacheOptions.
```
mafurman: nit: Would be helpful to use `<cref> `for InMemoryCryptoProviderCache and CryptoProviderCacheOptions

src/Microsoft.IdentityModel.Tokens/Cache/CryptoProviderCacheFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            // Create the signature provider caches based on the ProviderCacheType.
+            signingProvidersCache = CreateSignatureProviderCache(cryptoProviderCacheOptions);
+            verifyingProvidersCache = CreateSignatureProviderCache(cryptoProviderCacheOptions);
+
+            return new InMemoryCryptoProviderCache(cryptoProviderCacheOptions, signingProvidersCache, verifyingProvidersCache);
+        }
+
+        /// <summary>
+        /// Create a new instance of SignatureProvider based on the cache type in _cryptoProviderCacheOptions.
```
mafurman: nit: should be "SignatureProvider **cache**"

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
```diff
@@ -34,7 +34,7 @@
 
 namespace Microsoft.IdentityModel.Tokens
 {
-    internal class HashCache<TKey, TValue> : IProviderCache<TKey, TValue>
+    internal class RandomEvictCache<TKey, TValue> : IProviderCache<TKey, TValue>
```
mafurman: Filename is MaximumSizeCache but class name is still RandomEvictCache

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
```diff
@@ -44,26 +44,34 @@ internal class HashCache<TKey, TValue> : IProviderCache<TKey, TValue>
...
         private const int CompactionStateNotRunning = 0; // no compaction running
         private const int CompactionStateInProgress = 1; // compaction in progress
         private int _compactionState = CompactionStateNotRunning;
 
         #region constructors
 
-        public HashCache(
-            int capacity,
+        /// <summary>
+        /// Constructor.
```
mafurman, line 61: nit: change to "`Constructor for <see cref="MaximumSizeCache"/>`"

src/Microsoft.IdentityModel.Tokens/CryptoProviderCache.cs
mafurman: This increases the visibility of the method, so I don't think this is breaking.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
```diff
@@ -37,48 +37,79 @@ namespace Microsoft.IdentityModel.Tokens
...
+        /// Gets or sets the period to wait to remove expired items, in seconds.
+        /// </summary>
+        public int RemoveExpiredValuesIntervalInSeconds
+        {
+            get => _removeExpiredValuesIntervalInSeconds;
+            set => _removeExpiredValuesIntervalInSeconds = (value > 0) ? value : throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(RemoveExpiredValuesIntervalInSeconds), LogHelper.FormatInvariant(LogMessages.IDX10905, value)));
+        }
+
+        /// <summary>
+        /// Gets or sets the whether or not to remove expired items.
```
mafurman: nit: should be "Gets or sets the value that determines whether or not to remove expired values."

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -81,18 +83,17 @@ static CryptoProviderFactory()
 
         /// <summary>
         /// Default constructor for <see cref="CryptoProviderFactory"/>.
+        /// By default the EventBasedLRUCache will be used as the internal caches in the provider cache.
```
mafurman: nit: should be "By default the EventBasedLRUCache will be used  as the caching implementation for the InMemoryCryptoProviderCache."

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -226,25 +222,26 @@ private void RemoveLRUs()
...
+            //   4. if _eventQueueTaskState is successfully set to EventQueueTaskDoNotStop, the Interlocked.CompareExchange() in the EventQueueTaskAction() will fail
+            //      and the task will continue the while loop and the new event will keep the task running
+            //   5. if _eventQueueTaskState is NOT set to EventQueueTaskDoNotStop because of context switch back to the EventQueueTaskAction() and the _eventQueueTaskState is
             //      set to EventQueueTaskStopped (task exits), then the second Interlocked.CompareExchange() below should set the _eventQueueTaskState to EventQueueTaskRunning
             //      and start a task again (though this scenario is unlikely to happen)
             //
-            // Though this kind of context switching is probably unlikely to happen, this hopefully will take care of it.
+            // Without the EventQueueTaskDoNotStop state, this execution of this thread could pass the second CompareExchange() call below after step 2 above. The event queue task would still exit
```
mafurman: I'm finding the wording on this comment a little confusing; is this trying to say that if we didn't have EventQueueTaskDoNotStop we'd run into a scenario where the task might not be started up properly when a new event is added to the queue?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -51,47 +49,28 @@ public InMemoryCryptoProviderCache() : this(new CryptoProviderCacheOptions())
         {
         }
 
-        internal CryptoProviderFactory CryptoProviderFactory { get; set; }
-
         /// <summary>
         /// Creates a new instance of <see cref="InMemoryCryptoProviderCache"/> using the specified <paramref name="cryptoProviderCacheOptions"/>.
         /// </summary>
-        /// <param name="cryptoProviderCacheOptions">The options used to configure the <see cref="InMemoryCryptoProviderCache"/>.</param>
+        /// <param name="cryptoProviderCacheOptions"></param>
```
mafurman: Why was the comment for this parameter removed?

mafurman (changes requested):

src/Microsoft.IdentityModel.Tokens/Cache/CacheItem.cs
dannybtsai: DateTime is value type and cannot be null.

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
dannybtsai: When I added "Constructor for <see cref="MaximumSizeCache"/>", VS flagged that as an error with the message:
"XML comment has cref attribute 'MaximumSizeCache' that could not be resolved"

It seems the 'MaximumSizeCache' cannot be resolved anywhere above the constructor(s).

src/Microsoft.IdentityModel.Tokens/Cache/CacheItem.cs
dannybtsai: @mafurman,that is a good observation, I have removed the duplicated code.
The business logic of setting the ExpirationTime should not be in the CacheItem itself.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
dannybtsai: @mafurman, yes you are right, it could happen if we don't have EventQueueTaskDoNotStop, though the probability is tiny, only if the event queue task ALREADY checked the queue and it was empty and the new event was added AFTER that check.
Do you have suggestion on the comments to make it easier to understand?

mafurman: Got it. Here's how I would reword it:
"Without the EventQueueTaskDoNotStop state check below, steps (3) and (4) above will not execute. Step (5) will also not be hit, as the event queue is still running (even though the EventQueueTaskAction() method has already checked that the queue is empty and is about to stop the task). This method (StartEventQueueTaskIfNotRunning()) will return, the context will switch over to EventQueueTaskAction(), and the task will terminate. This means no new task would be started to process the newly added event.
This scenario is unlikely to happen, as it can only occur if the event queue task ALREADY checked the queue and it was empty and the new event was added AFTER that check."

mafurman: Feel free to use some/all of this if it makes more sense to you to;  this is just my opinion. :)

mafurman (approved):

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
dannybtsai: Thank you for the comments, they definitely help explain the scenario more. I made some minor changes.

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
dannybtsai: The compaction is started i the SetValue() method when the cache exceeds the limit (_compactionState will be set to CompactionStateInProgress).
This is at the end of the compacting operation so the state should be set to CompactionStateNotRunning.

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
dannybtsai: I have created a task to add the measure the new cache in the PerfAndStree, https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1672505/

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
```diff
@@ -0,0 +1,272 @@
...
+            var found = TryRemoveInternal(key, out CacheItem<TKey, TValue> val);
+            if (found)
+                value = val.Value;
+
+            return found;
+        }
+
+        #endregion
+
+
```
brentschmaltz: extra line

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
```diff
@@ -0,0 +1,272 @@
...
+        /// The internal version of TryGetValue that gets the LRUCacheItem from the hash.
+        /// </summary>
+        /// <param name="key">The key of the value to get.</param>
+        /// <param name="value">The object that has the specified key, or the default value of the type if the not found.</param>
+        /// <returns>true if the key was found, otherwise, false.</returns>
+        protected virtual bool TryGetValueInternal(TKey key, out CacheItem<TKey, TValue> value)
+        {
+            value = default;
+
+            var found = _map.TryGetValue(key, out CacheItem<TKey, TValue> val);
```
brentschmaltz: I am not seeing the need for this method.
Could derived classes could just call base.TryGetValue(...)?

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
```diff
@@ -0,0 +1,272 @@
...
+        /// Attempts to remove and return the value that has the specified key from the cache.
+        /// </summary>
+        /// <param name="key">The key of the element to remove and return.</param>
+        /// <param name="value">The object removed from the cache, or the default value of the TValue type if key does not exist.</param>
+        /// <returns>true if the object was removed successfully; otherwise, false.</returns>
+        protected virtual bool TryRemoveInternal(TKey key, out CacheItem<TKey, TValue> value)
+        {
+            value = default;
+
+            var found = _map.TryRemove(key, out CacheItem<TKey, TValue> val);
```
brentschmaltz: I am not understanding the need for this method.

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
```diff
@@ -0,0 +1,272 @@
...
+        protected static bool ValidateValues(TKey key, TValue value, DateTime expirationTime)
+        {
+            if (key == null)
+                throw LogHelper.LogArgumentNullException(nameof(key));
+
+            if (value == null)
+                throw LogHelper.LogArgumentNullException(nameof(value));
+
+            // if item already expired, do not add it to the cache
+            return expirationTime >= DateTime.UtcNow;
```
brentschmaltz: This cache is named MaximumSizeCache, why the check for expirationTime?
EventBasedLRUCache may have RemovedExpiredItems off this seems contradictory.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -243,67 +203,73 @@ internal int RemoveExpiredValues()
...
-            while (_map.Count > newCacheSize && _doubleLinkedList.Count > 0)
+            var newCacheSize = CalculateNewCacheSize();
+            while (base.Count > newCacheSize && _doubleLinkedList.Count > 0)
             {
                 var lru = _doubleLinkedList.Last;
-                if (_map.TryRemove(lru.Value.Key, out var cacheItem))
+                if (base.TryRemoveInternal(lru.Value.Key, out var cacheItem))
                     OnItemRemoved?.Invoke(cacheItem.Value);
 
                 _doubleLinkedList.Remove(lru);
```
brentschmaltz: Could you write:
```cs
_doubleLinkedList.RemoveLast();
```

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -226,7 +186,7 @@ internal int RemoveExpiredValues()
                     if (node.Value.ExpirationTime < DateTime.UtcNow)
                     {
                         _doubleLinkedList.Remove(node);
-                        if (_map.TryRemove(node.Value.Key, out var cacheItem))
+                        if (base.TryRemoveInternal(node.Value.Key, out var cacheItem))
```
brentschmaltz: what is the reason for using 'base.'?

src/Microsoft.IdentityModel.Tokens/Cache/IProviderCache.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// <returns>true if the key was found, otherwise, false.</returns>
+        public bool TryGetValue(TKey key, out TValue value);
+
+        /// <summary>
+        /// Removes a particular key from the cache.
+        /// </summary>
+        /// <param name="key">The key of the item to remove.</param>
+        /// <param name="value">The object removed from the cache, or the default value of the TValue type if key does not exist.</param>
+        /// <returns>true if the object was removed successfully; false otherwise.</returns>
+        public bool TryRemove(TKey key, out TValue value);
```
brentschmaltz, line 60: Do we need the out parameter?

src/Microsoft.IdentityModel.Tokens/Cache/IProviderCache.cs
```diff
@@ -0,0 +1,80 @@
...
+using System;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// The provider interface.
+    /// </summary>
+    /// <typeparam name="TKey">The key of a provider, used for cache operations.</typeparam>
+    /// <typeparam name="TValue">The value of a provider.</typeparam>
+    internal interface IProviderCache<TKey, TValue>
```
brentschmaltz, line 37: Since EventBaseLRUCache derives from MaximumSizeCache why is this interface needed?

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
```diff
@@ -0,0 +1,272 @@
...
+            return true;
+        }
+
+        /// <inheritdoc/>
+        public virtual bool TryGetValue(TKey key, out TValue value)
+        {
+            if (key == null)
+                throw LogHelper.LogArgumentNullException(nameof(key));
+
+            value = default;
```
brentschmaltz, line 126: could you write this as:
```cs
return _map.TryGetValue(key, out CacheItem<TKey, TValue> value);

File: src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs [hunk 43]
```diff
@@ -0,0 +1,272 @@
...
+            return found;
+        }
+
+        /// <inheritdoc/>
+        public virtual bool TryRemove(TKey key, out TValue value)
+        {
+            if (key == null)
+                throw LogHelper.LogArgumentNullException(nameof(key));
+
+            value = default;
```
brentschmaltz, line 140: as above could this be:
```cs
return _map.TryRemoveValue(key, out CacheItem<TKey, TValue> value);
```

src/Microsoft.IdentityModel.Tokens/Cache/IProviderCache.cs
dannybtsai: This interface is meant to abstract away the details of cache implementations. Otherwise, we would force all caches to inherit from a single base class. It has been changed to internal since you pointed out previously. And I still think internal interface has value vs base class as it allows more loosely coupled design.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
dannybtsai: Initially it was used for me to track all the places calling the new base class methods, I forgot to remove them after I finished.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
```diff
@@ -37,28 +37,80 @@ namespace Microsoft.IdentityModel.Tokens
...
+
+        /// <summary>
+        /// Gets or sets the value that determines whether or not to remove expired values.
+        /// </summary>
+        public bool RemoveExpiredValues { get; set; } = false;
+
+        /// <summary>
+        /// The desired cache type for caching providers. Defaults to LRU.
+        /// </summary>
+        public ProviderCacheType CacheType { get; set; } = ProviderCacheType.LRU;
```
brentschmaltz, line 114: The CryptoProviderCache most likely will never use LRU, do we need this public?

src/Microsoft.IdentityModel.Tokens/Cache/IProviderCache.cs
dannybtsai: Yes, out is needed in interface as it is part of the signature.

src/Microsoft.IdentityModel.Tokens/CryptoProviderCacheOptions.cs
dannybtsai: My idea is to allow users to select the cache types based on the needs of applications. For example, some users have concerns about LRU using tasks while they have only a few signature providers in the cache. Making this public provides an option for them to just use the new `MaximumSizeCache` or any new cache types we may have in the future.

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
dannybtsai: The base.TryGetValue(...) returns the CacheItem.Value but the derived class needs CacheItem.ExpirationTime too.

I renamed them to TryGetCacheItem() and TryRemoveCacheItem().

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
dannybtsai: The _map.TryGetValue(key, out CacheItem<TKey, TValue> value) is now wrapped in the TryGetCacheItem() which is shared by the derived class too, so it is more consistent.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -37,12 +36,11 @@ namespace Microsoft.IdentityModel.Tokens
     /// Current support is limited to <see cref="SignatureProvider"/> only.
     /// </summary>
     public class InMemoryCryptoProviderCache: CryptoProviderCache, IDisposable
-
     {
         internal CryptoProviderCacheOptions _cryptoProviderCacheOptions;
```
brentschmaltz, line 40: I don't see where _cryptoProviderCacheOptions are used, can this be removed?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -51,39 +49,28 @@ public InMemoryCryptoProviderCache() : this(new CryptoProviderCacheOptions())
...
         /// <summary>
         /// Creates a new instance of <see cref="InMemoryCryptoProviderCache"/> using the specified <paramref name="cryptoProviderCacheOptions"/>.
         /// </summary>
         /// <param name="cryptoProviderCacheOptions">The options used to configure the <see cref="InMemoryCryptoProviderCache"/>.</param>
-        /// <param name="options">Options used to create the event queue thread.</param>
-        /// <param name="tryTakeTimeout">The time used in ms for the timeout interval of the event queue. Defaults to 500 ms.</param>
-        internal InMemoryCryptoProviderCache(CryptoProviderCacheOptions cryptoProviderCacheOptions, TaskCreationOptions options, int tryTakeTimeout = 500)
+        /// <param name="signingSignatureProviders">The cache for signing signature providers.</param>
+        /// <param name="verifyingSignatureProviders">The cache for verifying signature providers.</param>
+        internal InMemoryCryptoProviderCache(CryptoProviderCacheOptions cryptoProviderCacheOptions, IProviderCache<string, SignatureProvider> signingSignatureProviders, IProviderCache<string, SignatureProvider> verifyingSignatureProviders)
```
brentschmaltz, line 69: I think there are two models here:
One where InMemoryCryptoProviderCache creates the two caches, and one where they are passed in.
Do we need both models?
Do we need this ctor, we certainly don't need the cryptoProviderCacheOptions.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -92,7 +79,7 @@ internal InMemoryCryptoProviderCache(CryptoProviderCacheOptions cryptoProviderCa
         /// <param name="signatureProvider">the <see cref="SignatureProvider"/> to create the key for.</param>
         /// <exception cref="ArgumentNullException">if signatureProvider is null.</exception>
         /// <returns>the cache key to use for finding a <see cref="SignatureProvider"/>.</returns>
-        protected override string GetCacheKey(SignatureProvider signatureProvider)
+        protected internal override string GetCacheKey(SignatureProvider signatureProvider)
```
brentschmaltz, line 82: changing the visibility to 'internal' is a breaking change.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -105,12 +92,12 @@ protected override string GetCacheKey(SignatureProvider signatureProvider)
...
         /// <param name="securityKey">the key that is used to by the crypto provider.</param>
         /// <param name="algorithm">the algorithm that is used by the crypto provider.</param>
-        /// <param name="typeofProvider">the typeof the crypto provider obtained by calling object.GetType().</param>
+        /// <param name="typeofProvider">the type of the crypto provider obtained by calling object.GetType().</param>
         /// <exception cref="ArgumentNullException">if securityKey is null.</exception>
         /// <exception cref="ArgumentNullException">if algorithm is null or empty string.</exception>
         /// <exception cref="ArgumentNullException">if typeofProvider is null or empty string.</exception>
         /// <returns>the cache key to use for finding a crypto provider.</returns>
-        protected override string GetCacheKey(SecurityKey securityKey, string algorithm, string typeofProvider)
+        protected internal override string GetCacheKey(SecurityKey securityKey, string algorithm, string typeofProvider)
```
brentschmaltz, line 100: changing the visibility to 'internal' is a breaking change.

dannybtsai: From MS Docs: "A `protected internal` member of a base class is accessible from any type within its containing assembly. It is also accessible in a derived class located in another assembly only if the access occurs through a variable of the derived class type."

So `protected internal` still allows the same level of access from outside of the assembly as `protected` does. And it also allows our tests to call the methods.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
dannybtsai: My intent is to remove the one that creates the two caches (they are dependencies) and replace it with the one the caches are passed in (injection). But since InMemoryCryptoProviderCache is public I am not allowed to remove the existing ctors.
I added the one with cryptoProviderCacheOptions so users can specify MaximumSizeCache instead of LRU. For those who are sensitive to tasks and only have a few signature providers in the cache the MaximumSizeCache should be a good one to use.

Instead of using the `CryptoProviderFactory.Default`, the users can use the new 
`public CryptoProviderFactory(CryptoProviderCacheOptions cryptoProviderCacheOptions)`, so the MaximumSizeCache will be used internally if cryptoProviderCacheOptions.CacheType is set to ProviderCacheType.MaximumSize

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
dannybtsai: It has been changed to `TryRemoveCachItem()` that wraps the `_map.TryRemove(key, out CacheItem<TKey, TValue> val)` call and shared by both classes.

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
dannybtsai: You are right, only the LRU should track the time if specified. This has been changed.

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
```diff
@@ -0,0 +1,253 @@
...
+            var newCacheSize = CalculateNewCacheSize();
+            while (_map.Count > newCacheSize)
+            {
+                var item = _map.FirstOrDefault();
+                if (_map.TryRemove(item.Key, out var cacheItem))
+                    OnItemRemoved?.Invoke(cacheItem.Value);
+            }
+
+            // reset the _compactionState to CompactionStateNotRunning
+            _compactionState = CompactionStateNotRunning;
```
arkrasik: `Volatile.Write(ref _compactionState, CompactionStateNotRunning)` ?

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
```diff
@@ -0,0 +1,253 @@
...
+        /// <summary>
+        /// FOR TESTING PURPOSES ONLY.
+        /// </summary>
+        internal virtual int TaskCount => 0;
+
+        internal virtual void WaitForProcessing()
+        {
+            while (true)
+            {
+                if (_compactionState == CompactionStateNotRunning)
```
arkrasik, line 250: `if (Volatile.Read(ref _compactionState) == ...)` ?

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
```diff
@@ -0,0 +1,253 @@
...
+        /// This should be a private method.
+        /// </summary>
+        private void CompactCache(object state)
+        {
+            // use the _capacity for the newCacheSize calculation in the case where the cache is experiencing overflow
+            var newCacheSize = CalculateNewCacheSize();
+            while (_map.Count > newCacheSize)
+            {
+                var item = _map.FirstOrDefault();
+                if (_map.TryRemove(item.Key, out var cacheItem))
```
arkrasik: check for null in if?

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
```diff
@@ -0,0 +1,253 @@
...
+        /// <summary>
+        /// Gets or sets the value associated with the specified key.
+        /// </summary>
+        /// <param name="key">The key of the value to get or set.</param>
+        /// For get, if the specified key is not found, a <exception cref="KeyNotFoundException"> will be thrown.</exception>
+        /// For set, if the key is null throw an <exception cref="ArgumentNullException"> will be thrown.</exception>
+        public CacheItem<TKey, TValue> this[TKey key]
+        {
+            get => _map[key];
+            set => _map[key] = value;
```
arkrasik, line 88: nit: why do we want to expose setter that isn't going through `SetValue`?

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
dannybtsai: This is a good catch, thanks.
The cacheItem is never null but the Key can be, so it now only TryRemove() if not `default CachItem`.

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
dannybtsai: @arkrasik, I am not familiar with Volatile.Write() but believe the Interlocked.Exchange() should be concurrently safe for multiple threads running on multiple CPUs.

src/Microsoft.IdentityModel.Tokens/Cache/MaximumSizeCache.cs
dannybtsai: @arkrasik, since this is the only operation performed in the while loop, there should not be any problem related to instructions being reordered by the compiler or processor so I don't think using Volatile.Read() is necessary based on my understanding.

src/Microsoft.IdentityModel.Tokens/ProvidersCache/HashCache.cs
brentschmaltz: @henrik-me excellent idea, are you blocked on adding these telemetry signals?

brentschmaltz:
We decided to make simpler changes just related to controlling if LRU functionality is on or off as for some scenarios such as caching SignatureProviders, the extra cpu cycles to move the SignatureProvider to the front of the is not required.

PR #1645, opened 2021-05-08
Replace split with explicit token segments count (version 2)

lechu445:
This PR is based on https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1318 as it seems author is not active on GitHub anymore.
I've added all the changes from the PR, fixed its code review issues and added some more optimizations.

PR #1772, opened 2021-11-10
Replace split with explicit token segments count

lechu445:
This is cherry-pick of https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1645 to branch `brentsch/SystemTextJson`.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -280,8 +280,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
         internal static SecurityKey GetSecurityKey(EncryptingCredentials encryptingCredentials, CryptoProviderFactory cryptoProviderFactory, out byte[] wrappedKey)
         {
-            SecurityKey securityKey = null;
-            KeyWrapProvider kwProvider = null;
+            SecurityKey securityKey;
+            KeyWrapProvider kwProvider;
             wrappedKey = null;
 
             // if direct algorithm, look for support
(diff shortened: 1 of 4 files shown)


lechu445:
@brentschmaltz I have created this PR to ease you pulling https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1645 into the branch you are working on.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -466,6 +466,26 @@ internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, Token
             else
                 return null;
         }
+
+        /// <summary>
+        /// Returns number of Jwt parts in token.
+        /// </summary>
+        /// <param name="token">Input token to count patrs.</param>
+        /// <returns>Value between 1 and <see cref="JwtConstants.MaxJwtSegmentCount"/> + 1.</returns>
+        internal static int CountTokenJwtParts(string token)
```
brentschmaltz: @lechu445 we need to be thinking about what to do if the count is greater than 5.
We can't just return as the jwt is not well formed.

brentschmaltz:
@lechu445 thanks for rebasing against the topic branch.
I will include these changes into that PR.
I will add you as a reviewer if you don't mind.

lechu445:
@brentschmaltz  Sure, thank you.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
lechu445: I've just added method DetectTokenType that returns Unknown enum value in case of not well formed token. The consumers of this method can handle the Unknown type on their way.

lechu445:
@brentschmaltz I have rebased this PR changes on the newest `brentsch/SystemTextJson`

risogolo:
when this will be merged? will this cover ? [https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1805](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1805)

lechu445:
@risogolo 

> when this will be merged?

I think this is a question to @brentschmaltz.

> will this cover ? https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1805

No, this PR does not include any functional changes to make it easier to merge. This is just refactoring that focuses on performance improvement. The issue that you mentioned is a separate issue and the best approach is to create a separate PR with the fix.

brentschmaltz:
We are moving away from splitting in general and will have the work using System.Text.Json completed by end of April.
This is a major rewrite, using SPAN, keeping octets in memory.

In the new model, as we are counting the dots, we are recording the spans we need to perform cryptographic operations.
So unfortunately, this PR doesn't quite fit.

PR #1775, opened 2021-11-16
Simplify strings comparison with Ordinal option

lechu445:
No functional changes. Actually the == operator is a little bit faster because does not need to check comparison type.

see https://source.dot.net/#System.Private.CoreLib/String.Comparison.cs,739 and https://source.dot.net/#System.Private.CoreLib/String.Comparison.cs,643.


lechu445:
Convention to use string.Equals for ordinal comparison does not bring benefits. Usage of `==` brings:
- less verbosity, so improves code readability
- operator == for string comparison is widely used in .NET world (example https://source.dot.net/#System.Private.Xml/System/Xml/Core/XmlReader.cs,761)
- C# compiler can execute `==` on two constant strings, for example `"x"==""` is compiled as false
- consistency with convention how other types are compared, e.g. integers are compared with == instead of Equals method (`list.Count == 1` instead of `list.Count.Equals(1)`)
- about 30%-40% faster operation
```
|              Method |     Mean |     Error |    StdDev | Ratio | 
|-------------------- |---------:|----------:|----------:|------:|
|              Equals | 3.957 ns | 0.0182 ns | 0.0161 ns |  1.00 | 
| DoubleEqualOperator | 2.612 ns | 0.0154 ns | 0.0120 ns |  0.66 |
```

The purpose of this PR is to change the convention in the whole library as, to me, brings more benefits.

jennyf19 (approved):

jennyf19:
@brentschmaltz 

@lechu445 Please tag either @brentschmalt or myself when you have PRs so we can ack them. Thanks.

brentschmaltz:
@lechu445 wrote some additional tests using benchmark and i didn't see a as big of a difference.
see: [benchmark code](https://github.com/brentschmaltz/CodeSnips/blob/dev/Benchmarks/StringEquals.cs)

brentschmaltz (changes requested):
please review the benchmark code provided and comment.

At the bottom of this [document](https://docs.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings) there is an explicit note to not use "==" or "!="

lechu445:
Interesting.
The current version of JIT is able to inline the Equals without StringComparison code and produce asm with direct call of System.SpanHelpers.SequenceEqual. For Equal with provided StringComparison JIT is not smart enough and calls string.Equal that calls System.SpanHelpers.SequenceEqual underneath. So, usage of `string.Equals(a, b)` or `==` reduce one method call.
https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcDMACR2DC2A3utmdgA4BOAlgG5jACm2NAdsNgPpsCuUDADa8mEANyly1eoxYBjAPZsIndpx68AnlEEKFFCNgC82JAAYLZiWnKVaDZrnPcA5rxoATY9gDi7jwB0AHJMAO5+ngAUAJQBACoKAMrAtGwuMda20g4sVExgHkqCmk5mrv4I3hGBIeH+MfFJKezp0daSZFi4ACzYyakuAKIAjrxgghAA8lQe7OMxHcSLtgBmClTYkWqs3lY7ADzcfNq6+uKsANQX0cu25DQrm3DmASNjE5Fcbp4ANOWeCD+/Ra+AUUAoYFoECUAWmszY42iNxsd1Rtg0AnGIggV0y5AAvuhFl04L1gWk3uMIAsUWQSLS7msNlsODsTHsaNhDhoTnoDGJLtdbqiHk8XpSPl9/H8pQCkcK0WQMUJsbjFoS0MScKTsBKIHqabZ6YqmZttpz2QLOdzjjo+ecaFdkYrbKLPt8vEYTLKPAhnS7UcqsaI1QyNfigA==  
In my spare time I will run the benchmarks on some runtimes.  
The note in documentation is a strong argument. I think it's a matter of choice between simplicity and preciseness.
Even .NET library breaks it https://source.dot.net/#System.Private.Xml/System/Xml/Core/XmlReader.cs,781 😄

brentschmaltz:
@lechu445 my tests have shown a big improvement by just dropping the parameter: StringComparison.Ordinal.
Would it make sense to modify this PR to just drop the parameter?

Rebase against current dev and then drop in?

You are totally right about differences in the code base, but microsoft is a large company :-)
Let's try and do the right thing here.

And thanks again for helping out, we have been focused on some internal issues, but are coming out of it....

lechu445:
@brentschmaltz yes, it totally makes sense to just drop StringComparison.Ordinal.

Working on it. The rebase will be ready in max 1 hour.

lechu445:
@brentschmaltz Done. I also found more places where StringComparison.Ordinal could be removed from string.Equals

brentschmaltz (approved):

brentschmaltz:
@lechu445 Thanks!

PR #1776, opened 2021-11-16
Xiao/better exception for get OIDC configuration async

### LINKED ISSUE #1714: [FeatureRequest] Better Transient/ClientMisconfiguration error handling for ConfigurationManager.GetConfigurationAsync?
During token validation, our service uses the below `ConfigurationManager.GetConfigurationAsync` API in order to retrieve the `OpenIdConnectConfiguration` for a tenant-specific authority (e.g.: `https://login.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/`):
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/4f24eaecffca2a0ebacc63976682df0adab54805/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs#L124

For this particular API, our service would like to distinguish between two types of errors:
- **Transient Errors:** If the underlying `HttpDocumentRetriever` fails to call `https://login.windows.net/{tenantId}/.well-known/openid-configuration` due to some transient I/O or networking issue, then we want to surface this to the customer as a `503 ServiceUnavailable` error. This signals to the customer that the issue is transient, so that they can configure a retry policy to try again later.	
- **ClientMisconfiguration Errors:** Sometimes, customers will accidentally use a tenant ID belonging to the wrong environment (e.g.: the Microsoft PPE tenant `f686d426-8d16-42db-81b7-ab578e110ccd` doesn't exist in the Public Azure environment `https://login.windows.net`). When the underlying `HttpDocumentRetriever` calls `https://login.windows.net/f686d426-8d16-42db-81b7-ab578e110ccd/.well-known/openid-configuration`, it gets back a `400 BadRequest` error (`AADSTS90002: Tenant 'f686d426-8d16-42db-81b7-ab578e110ccd' not found.`). We want to surface this to the customer as a `401 Unauthorized` error with some appropriate error code (e.g.: `InvalidTenant`).

Currently, our service is unable to distinguish between the above two classes of errors. In both scenarios, the underlying ConfigurationManager/HttpDocumentRetriever throws an `InvalidOperationException` with an `IDX208xx`-level error code. **Could the error handling for this API be improved so that we can distinguish between these two types of errors?** It would be great if the thrown exception could surface additional data, such as the following:
- **HTTP status code** (e.g.: `400 BadRequest` or `500 InternalServerError`).
- **Error message** (assuming that PII can be sanitized without making the error message useless).
- **Correlation ID** (for the scenario where our DevOps engineers need to escalate a legitimate outage and provide eSTS/AAD a correlation ID for some incident).

**Internal tracking**
- [Product Backlog Item 1601972](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1601972 ): Better error handling for MetadataManager.GetOidcConfigurationAsync
  - [Product Backlog Item 1509137](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1509137): Create new exception type that contains errors from ESTS and add as InnerException to AuthenticationException

ciaozhang:
Modify wilson to include http status code and response string in exception.Data dictionary so that SAL can parse the response from exception.Data and populate properties on the new exception.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
@@ -44,6 +44,11 @@ namespace Microsoft.IdentityModel.Protocols
         private HttpClient _httpClient;
         private static readonly HttpClient _defaultHttpClient = new HttpClient();
 
+#pragma warning disable 1591
+        public const string StatusCode = "status_code";
+        public const string ResponseContent = "response_content";
+#pragma warning disable 1591
+
         /// <summary>
         /// Gets or sets whether additional default headers are added to a <see cref="HttpRequestMessage"/> headers. Set to true by default.
         /// </summary>
@@ -120,7 +125,14 @@ namespace Microsoft.IdentityModel.Protocols
                 if (response.IsSuccessStatusCode)
                     return responseContent;
 
-                 unsuccessfulHttpResponseException = new IOException(LogHelper.FormatInvariant(LogMessages.IDX20807, address, response, responseContent));
+                 unsuccessfulHttpResponseException = new IOException(LogHelper.FormatInvariant(LogMessages.IDX20807, address, response, responseContent))
+                 {
+                     Data =
+                     {
+                         { StatusCode, response.StatusCode },
+                         { ResponseContent, responseContent }
+                     }
+                 };
             }
             catch (Exception ex)
             {
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
@@ -44,10 +44,15 @@ namespace Microsoft.IdentityModel.Protocols
         private HttpClient _httpClient;
         private static readonly HttpClient _defaultHttpClient = new HttpClient();
 
-#pragma warning disable 1591
+        /// <summary>
+        /// The key is used to add status code into ex.Data.
+        /// </summary>
         public const string StatusCode = "status_code";
+
+        /// <summary>
+        /// The key is used to add response content into ex.Data.
+        /// </summary>
         public const string ResponseContent = "response_content";
-#pragma warning disable 1591
 
         /// <summary>
         /// Gets or sets whether additional default headers are added to a <see cref="HttpRequestMessage"/> headers. Set to true by default.
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
@@ -125,14 +125,9 @@ namespace Microsoft.IdentityModel.Protocols
                 if (response.IsSuccessStatusCode)
                     return responseContent;
 
-                 unsuccessfulHttpResponseException = new IOException(LogHelper.FormatInvariant(LogMessages.IDX20807, address, response, responseContent))
-                 {
-                     Data =
-                     {
-                         { StatusCode, response.StatusCode },
-                         { ResponseContent, responseContent }
-                     }
-                 };
+                unsuccessfulHttpResponseException = new IOException(LogHelper.FormatInvariant(LogMessages.IDX20807, address, response, responseContent));
+                unsuccessfulHttpResponseException.Data.Add(StatusCode, response.StatusCode);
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -44,6 +44,11 @@ public class HttpDocumentRetriever : IDocumentRetriever
         private HttpClient _httpClient;
         private static readonly HttpClient _defaultHttpClient = new HttpClient();
 
+#pragma warning disable 1591
```
mafurman: What warning is this disabling?

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -120,7 +125,9 @@ public async Task<string> GetDocumentAsync(string address, CancellationToken can
                 if (response.IsSuccessStatusCode)
                     return responseContent;
 
-                 unsuccessfulHttpResponseException = new IOException(LogHelper.FormatInvariant(LogMessages.IDX20807, address, response, responseContent));
+                unsuccessfulHttpResponseException = new IOException(LogHelper.FormatInvariant(LogMessages.IDX20807, address, response, responseContent));
```
mafurman, line 133: Should we create a different exception name for this? If we did, we could have StatusCode and ResponseContent as properties on it.

Not sure if we'd consider changing the exception type a breaking change though.

test/Microsoft.IdentityModel.Protocols.Tests/HttpDocumentRetrieverTests.cs
```diff
@@ -91,6 +92,12 @@ public void GetMetadataTest(DocumentRetrieverTheoryData theoryData)
             {
                 aex.Handle((x) =>
                 {
+                    if (x.Data.Count > 0)
+                    {
+                        Assert.True(x.Data.Contains(HttpDocumentRetriever.StatusCode));
```
mafurman: We generally try to use CompareContext.AddDiff() rather than relying on Asserts, as Assert failures will cause the test to terminate and will not notify us of failures further on in the test.

mafurman (changes requested):

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
ciaozhang: Missing XML comment for publicly visible type or member 'Type_or_Member'

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
ciaozhang: I  create a new exception(GetOidcConfigurationException) in SAL and it has StatusCode and ResponseContent as properties there

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
TimHannMSFT: Why not populate the XML comment?

test/Microsoft.IdentityModel.Protocols.Tests/HttpDocumentRetrieverTests.cs
ciaozhang: use context.adddiff in latest commit

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
ciaozhang: added XML

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
mafurman: Would it make more sense for Wilson to throw that exception directly? Is there anything in it that is first party only?

mafurman (approved):

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
ciaozhang: George has mentioned in teams channel, it might not the best solution to parse the response from AAD in Wilson since Wilson is IDP-agnostic.

Issue #468, opened 2016-07-05
Add Async pattern for TokenValidation


MicahZoltu:
There are a number of places where `JWTSecurityTokenHandler` calls into user provided methods that may be doing some IO (e.g., database or remote calls) but they do so in a synchronous way.

One example of such a place is with JWT signature validation.  The user can provide their own `SignatureValidator` or `IssuerSigningKeyResolver`.  The act of validating a signature or resolving signing keys may require an interaction with an external server.

For a real-world example, Google rotates its signing keys regularly (rumor is daily) and you can get the latest public key in JWK format [here](https://www.googleapis.com/oauth2/v3/certs).  Unfortunately, it is not possible to know when they are going to roll their keys so an application that is attempting to validate Google issued JWTs will need to hit that endpoint to retrieve the latest  signing key.  The application could be optimized to cache the signing keys but that is still an async operation and it is entirely up to Google as to how often they rotate their keys (they could start rotating them for every request if they wanted).

Another real-world example once again with Google's authentication, Google supports sending the token to [here](https://www.googleapis.com/oauth2/v3/tokeninfo?id_token=XYZ123) for validation rather rather than doing the validation locally.  If one wants to use this mechanism then every token validation would be a remote call.

Unfortunately, I recognize that the current API is very async un-friendly and the interfaces it implements are also not async-friendly but I wanted to get this filed anyway in hopes that at some point things can be improved to support modern asynchronous use cases.

brentschmaltz:
@Zoltu I agree with you that off-box validation is a real scenario. We should have an async pattern up and down the call graph to realize this pattern.
We stopped at metadata retrieval.

brentschmaltz:
We will need to add an new return type: ValidationResult so the 'out' param is not needed.

brentschmaltz:
Include this issue with this fix.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/566

brentschmaltz:
Modified title to reflect that this is general in nature across all token types. Perhaps in SecurityTokenHandler

shagenhokie:
Do we have an ETA on this?

brentschmaltz:
@shagenhokie not yet

fragilethoughts:
[...]

PR #1779, opened 2021-11-19
Refactored the validation flow that uses the Configuration to avoid cloning the TokenValidationParameters

mafurman:
Cloning the TokenValidationParameters is expensive, but it was necessary because we did NOT want to modify the TokenValidationParameters that were passed in, but we needed to set the Configuration on them. The Configuration property has been removed from the TVP in favor of passing it in directly to the methods that require it. In addition to modifying some internal and private methods, it was also necessary to add a new type of IssuerValidator, IssuerSigningKeyResolver, and IssuerSigningKeyValidator to the TokenValidationParameters that are able to take in a BaseConfiguration object.

Additionally, since we anticipate that users will be passing in the Configuration more often than not, the validation logic has been modified so that it is checked prior to any properties that may be present on the TVP.


src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -379,17 +379,18 @@ private static long ParseTimeValue(JToken jToken, string claimName)
         /// </summary>
         /// <param name="kid">The <see cref="string"/> kid field of the token being validated</param>
         /// <param name="x5t">The <see cref="string"/> x5t field of the token being validated</param>
-        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/>  required for validation.</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <param name="configuration">The <see cref="BaseConfiguration"/> that will be used along with the <see cref="TokenValidationParameters"/> to resolve the signing key</param>
         /// <returns>Returns a <see cref="SecurityKey"/> to use for signature validation.</returns>
         /// <remarks>If key fails to resolve, then null is returned</remarks>
-        private static SecurityKey ResolveTokenSigningKeyWithConfig(string kid, string x5t, TokenValidationParameters validationParameters)
+        internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, TokenValidationParameters validationParameters, BaseConfiguration configuration)
```
mafurman, line 386: This method doesn't necessarily need an overload, though I think keeping the code for resolving the key from the configuration separate from the code for resolving from the TVP is a bit cleaner. Willing to hear perspectives on this though!

src/Microsoft.IdentityModel.Tokens/GlobalSuppressions.cs
```diff
@@ -49,12 +49,15 @@
...
 [assembly: SuppressMessage("Design", "CA1001:Types That own disposable fields should be disposable", Justification = "Exceptions can occurr if disposed of", Scope = "type", Target = "~T:Microsoft.IdentityModel.Tokens.EventBasedLRUCache`2")]
+[assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "Used as validation", Scope = "member", Target = "~M:Microsoft.IdentityModel.Tokens.InternalValidators.ValidateLifetimeAndIssuerAfterSignatureNotValidatedJwt(Microsoft.IdentityModel.Tokens.SecurityToken,System.Nullable{System.DateTime},System.Nullable{System.DateTime},System.String,Microsoft.IdentityModel.Tokens.TokenValidationParameters,Microsoft.IdentityModel.Tokens.BaseConfiguration,System.Text.StringBuilder)")]
+#if NET472 || NETSTANDARD2_0
+[assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "Used to determine appropriate code path to take.", Scope = "member", Target = "~M:Microsoft.IdentityModel.Tokens.JsonWebKeyConverter.ConvertFromECDsaSecurityKey(Microsoft.IdentityModel.Tokens.ECDsaSecurityKey)~Microsoft.IdentityModel.Tokens.JsonWebKey")]
+#endif
+#if !NET472
```
mafurman, line 61: Added #if #defs here as looking at this file without them makes it seem like certain suppressions are not used.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -150,10 +150,25 @@ internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParame
...
-                foreach (SecurityKey key in validationParameters.Configuration.SigningKeys)
-                    yield return key;
+
+        /// <summary>
+        /// Returns all <see cref="SecurityKey"/> provided in <paramref name="configuration"/> and <paramref name="configuration"/>.
+        /// </summary>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <param name="configuration">The <see cref="BaseConfiguration"/> that contains signing keys used for validation.</param>
+        /// <returns>Returns all <see cref="SecurityKey"/> provided in provided in <paramref name="configuration"/> and <paramref name="configuration"/></returns>
+        internal static IEnumerable<SecurityKey> GetAllSigningKeys(TokenValidationParameters validationParameters, BaseConfiguration configuration)
```
mafurman, line 162: Similar to ResolveTokenSigningKey(), I'm not sure if this method really needs an overload. Might be a bit cleaner not to have one since we could use `yield return ` consistently.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -125,7 +126,8 @@ public class AadIssuerValidator
                     {
                         if (IsV2Authority)
                         {
-                            AadIssuerV2 = validationParameters.Configuration.Issuer;
+                            BaseConfiguration openIdConnectConfig = validationParameters.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
```
mafurman: @jennyf19 Had to mess around with your code when adding these changes; please let me know if something doesn't look right!

jennyf19: actually @RojaEnnam has [changes there from a bug found when testing her config PR](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1780), which has the correct behavior.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -999,32 +999,29 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
+            BaseConfiguration currentConfiguration = null;
+            if (validationParameters.ConfigurationManager != null)
             {
                 try
                 {
-                    // TODO: Use overloads to get rid of the Clone() operation.
-                    var configuration = validationParametersCopy.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
-                    validationParametersCopy = validationParameters.Clone();
-                    validationParametersCopy.Configuration = configuration;
+                    currentConfiguration = validationParameters.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
```
brentschmaltz, line 1007: Referenced this PR in this issue: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/468 it is about async.
We will now need to think async.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1036,43 +1033,43 @@ private TokenValidationResult ValidateToken(string token, JsonWebToken outerToke
                    && (tokenValidationResult.Exception as SecurityTokenUnableToValidateException).ValidationFailure != ValidationFailure.InvalidLifetime)
                    || tokenValidationResult.Exception.GetType().Equals(typeof(SecurityTokenSignatureKeyNotFoundException)))
                 {
```
dannybtsai, line 1037: I commented in the other PR:

```
var resultType = tokenValidationResult.Exception.GetType();
if (resultType.Equals(typeof(SecurityTokenInvalidSignatureException))
   || resultType.Equals(typeof(SecurityTokenInvalidSigningKeyException))
   || resultType.Equals(typeof(SecurityTokenInvalidIssuerException))
   || (resultType.Equals(typeof(SecurityTokenUnableToValidateException))
   // we should not try to revalidate with the LKG or request a refresh if the token has an invalid lifetime
   && (tokenValidationResult.Exception as SecurityTokenUnableToValidateException).ValidationFailure != ValidationFailure.InvalidLifetime)
   || resultType.Equals(typeof(SecurityTokenSignatureKeyNotFoundException)))
```
And if the type applies to derived types then using the `is `operator might be better:

`if (resultType  is SecurityTokenInvalidSignatureException)`

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1036,43 +1033,43 @@ private TokenValidationResult ValidateToken(string token, JsonWebToken outerToke
...
-                    if (validationParametersCopy.ConfigurationManager.UseLastKnownGoodConfiguration
-                        && validationParametersCopy.ConfigurationManager.LastKnownGoodConfiguration != null
-                        && !(validationParametersCopy.ConfigurationManager.LastKnownGoodConfiguration == validationParametersCopy.Configuration))
+                    if (validationParameters.ConfigurationManager.UseLastKnownGoodConfiguration
+                        && validationParameters.ConfigurationManager.LastKnownGoodConfiguration != null
+                        && !(validationParameters.ConfigurationManager.LastKnownGoodConfiguration == currentConfiguration))
```
dannybtsai: >&& !(validationParameters.ConfigurationManager.LastKnownGoodConfiguration == currentConfiguration) [](http://example.com/codeflow?start=24&length=98)

Replace with validationParameters.ConfigurationManager.LastKnownGoodConfiguration != currentConfiguration

dannybtsai:
>                 else if (tokenValidationResult.Exception.GetType().Equals(typeof(SecurityTokenInvalidSignatureException))

Should this if block be moved to after line 1017 right after the tokenValidationResult is set?

---
Refers to: src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs:1028 in 6d31a6e. [](commit_id = 6d31a6e09d1f67fc623698cec1f221c61aa15ca2, deletion_comment = False)

dannybtsai:
>         {

This is probably not realistic but in theory an infinite recursion can happen:
1. Starting from the ValidateToken (line 948)
2. ValidateToken (line 972)
3. ValidateJWE (line 1017)
4. ValidateTokenPayload (1090)
5. ValidateToken (1121) 
6. Now it is back to the ValidateToken (line 948) where it starts.

Note: step 4 & 5 is calling a public from private. My feeling is that there might be common functionality that can be extracted and shared.
---
Refers to: src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs:949 in 6d31a6e. [](commit_id = 6d31a6e09d1f67fc623698cec1f221c61aa15ca2, deletion_comment = False)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1036,43 +1033,43 @@ private TokenValidationResult ValidateToken(string token, JsonWebToken outerToke
...
-                    validationParametersCopy.Configuration = validationParametersCopy.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).GetAwaiter().GetResult();
+                    validationParameters.ConfigurationManager.RequestRefresh();
+                    var lastConfig = currentConfiguration;
+                    currentConfiguration = validationParameters.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).GetAwaiter().GetResult();
 
                     // Only try to re-validate using the newly obtained config if it doesn't reference equal the previously used configuration.
-                    if (lastConfig != validationParametersCopy.Configuration)
-                        return decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParametersCopy) : ValidateJWS(token, validationParametersCopy); ;
+                    if (lastConfig != currentConfiguration)
+                        return decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParameters, currentConfiguration) : ValidateJWS(token, validationParameters, currentConfiguration); ;
```
dannybtsai: > ; [](http://example.com/codeflow?start=201&length=2)

remove

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -999,32 +999,29 @@ public virtual TokenValidationResult ValidateToken(string token, TokenValidation
...
                     // Keep going with the validation as the TokenValidationParameters may have the issuer and signing key set
                     // directly on them.
-                    LogHelper.LogInformation(LogHelper.FormatInvariant(TokenLogMessages.IDX10261, validationParametersCopy.ConfigurationManager.MetadataAddress, ex.ToString()));
+                    LogHelper.LogInformation(LogHelper.FormatInvariant(TokenLogMessages.IDX10261, validationParameters.ConfigurationManager.MetadataAddress, ex.ToString()));
                 }
             }
 
-            TokenValidationResult tokenValidationResult = decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParametersCopy) : ValidateJWS(token, validationParametersCopy);
-            if (validationParametersCopy.ConfigurationManager != null)
+            TokenValidationResult tokenValidationResult = decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParameters, currentConfiguration) : ValidateJWS(token, validationParameters, currentConfiguration);
```
dannybtsai, line 1017: >decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParameters, currentConfiguration) : ValidateJWS(token, validationParameters, currentConfiguration) [](http://example.com/codeflow?start=58&length=169)

Since this appears three times in this method, maybe consider capturing the logic in a method to ensure consistency?

        private TokenValidationResult ValidateJWSOrJWE(string token, JsonWebToken outerToken, string decryptedJwt, TokenValidationParameters validationParameters, BaseConfiguration configuration)
        {
            return decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParameters, configuration) : ValidateJWS(token, validationParameters, configuration);
        }

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
mafurman: Modified again in this PR: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1786/files

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: should we use reference equals.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: @dannybtsai @mafurman we can hold off on any additional cleanup here, if we want, as the changes associated with JsonClaimSet and using SPAN will make some changes in this code.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: Will this ever result in a null ref?
> (tokenValidationResult.Exception as SecurityTokenUnableToValidateException).ValidationFailure !

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1138,14 +1135,27 @@ private TokenValidationResult ValidateTokenPayload(JsonWebToken jsonWebToken, To
...
         {
             if (string.IsNullOrWhiteSpace(token))
                 throw LogHelper.LogArgumentNullException(nameof(token));
 
             if (validationParameters == null)
                 throw LogHelper.LogArgumentNullException(nameof(validationParameters));
 
+            if (validationParameters.SignatureValidatorWithConfiguration != null)
+            {
+                var validatedToken = validationParameters.SignatureValidatorWithConfiguration(token, validationParameters, configuration);
```
brentschmaltz: In the other PR, we suggested a name change to: SignatureValidatorUsingConfiguration instead of SignatureValidatorWithConfiguration. With -> Using.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1187,13 +1197,17 @@ private static JsonWebToken ValidateSignature(string token, TokenValidationParam
 
             bool kidMatched = false;
             IEnumerable<SecurityKey> keys = null;
-            if (validationParameters.IssuerSigningKeyResolver != null)
+            if (validationParameters.IssuerSigningKeyResolverWithConfiguration != null)
```
brentschmaltz: Same as above: With -> Using

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
brentschmaltz: I agree with the overload.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -60,26 +60,70 @@ namespace Microsoft.IdentityModel.Tokens
...
+    /// </summary>
+    /// <param name="token">The <see cref="string"/> representation of the token that is being validated.</param>
+    /// <param name="securityToken">The <see cref="SecurityToken"/> that is being validated. It may be null.</param>
+    /// <param name="kid">A key identifier. It may be null.</param>
+    /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+    /// <param name="configuration"><see cref="BaseConfiguration"/> required for validation.</param>
+    /// <returns>A <see cref="SecurityKey"/> to use when validating a signature.</returns>
+    /// <remarks> If both <see cref="IssuerSigningKeyResolverWithConfiguration"/> and <see cref="IssuerSigningKeyResolver"/> are set, IssuerSigningKeyResolverWithConfiguration takes
+    /// priority.</remarks>
+    public delegate IEnumerable<SecurityKey> IssuerSigningKeyResolverWithConfiguration(string token, SecurityToken securityToken, string kid, TokenValidationParameters validationParameters, BaseConfiguration configuration);
```
brentschmaltz: Same as above: With -> Using

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -60,26 +60,70 @@ namespace Microsoft.IdentityModel.Tokens
...
+    /// <summary>
+    /// Definition for IssuerSigningKeyValidatorWithConfiguration.
+    /// </summary>
+    /// <param name="securityKey">The <see cref="SecurityKey"/> that signed the <see cref="SecurityToken"/>.</param>
+    /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
+    /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+    /// <param name="configuration"><see cref="BaseConfiguration"/> required for validation.</param>
+    /// <remarks> If both <see cref="IssuerSigningKeyResolverWithConfiguration"/> and <see cref="IssuerSigningKeyResolver"/> are set, IssuerSigningKeyResolverWithConfiguration takes
+    /// priority.</remarks>
+    public delegate bool IssuerSigningKeyValidatorWithConfiguration(SecurityKey securityKey, SecurityToken securityToken, TokenValidationParameters validationParameters, BaseConfiguration configuration);
```
brentschmaltz: Same as above: With -> Using

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -60,26 +60,70 @@ namespace Microsoft.IdentityModel.Tokens
...
+    /// <param name="issuer">The issuer to validate.</param>
+    /// <param name="securityToken">The <see cref="SecurityToken"/> that is being validated.</param>
+    /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+    /// <param name="configuration"><see cref="BaseConfiguration"/> required for validation.</param>
+    /// <returns>The issuer to use when creating the "Claim"(s) in a "ClaimsIdentity".</returns>
+    /// <remarks>The delegate should return a non null string that represents the 'issuer'. If null a default value will be used.
+    /// If both <see cref="IssuerValidatorWithConfiguration"/> and <see cref="IssuerValidator"/> are set, IssuerValidatorWithConfiguration takes
+    /// priority.
+    /// </remarks>
+    public delegate string IssuerValidatorWithConfiguration(string issuer, SecurityToken securityToken, TokenValidationParameters validationParameters, BaseConfiguration configuration);
```
brentschmaltz: Same as above: With -> Using

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -105,6 +149,14 @@ namespace Microsoft.IdentityModel.Tokens
     /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
     public delegate SecurityToken SignatureValidator(string token, TokenValidationParameters validationParameters);
 
+    /// <summary>
+    /// Definition for SignatureValidator.
+    /// </summary>
+    /// <param name="token">A securityToken with a signature.</param>
+    /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+    /// <param name="configuration">The <see cref="BaseConfiguration"/> that is required for validation.</param>
+    public delegate SecurityToken SignatureValidatorWithConfiguration(string token, TokenValidationParameters validationParameters, BaseConfiguration configuration);
```
brentschmaltz: Same as above: With -> Using

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -394,9 +435,25 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
...
+        /// <remarks>
+        /// If set, this delegate will be called to validate the <see cref="SecurityKey"/> that signed the token, instead of default processing.
+        /// This means that no default <see cref="SecurityKey"/> validation will occur.
+        /// Even if <see cref="ValidateIssuerSigningKey"/> is false, this delegate will still be called.
+        /// This delegate should be used if properties from the configuration retrieved from the authority are necessary to validate the
+        /// issuer signing key.
+        /// If both <see cref="IssuerSigningKeyValidatorWithConfiguration"/> and <see cref="IssuerSigningKeyValidator"/> are set, IssuerSigningKeyValidatorWithConfiguration takes
+        /// priority.
+        /// </remarks>
+        public IssuerSigningKeyValidatorWithConfiguration IssuerSigningKeyValidatorWithConfiguration { get; set; }
```
brentschmaltz: Same as above: With -> Using

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -407,9 +464,24 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
...
+        /// <see cref="TokenValidationParameters"/> and <see cref="BaseConfiguration"/>.
+        /// </summary>
+        /// <remarks>
+        /// This <see cref="SecurityKey"/> will be used to check the signature. This can be helpful when the <see cref="SecurityToken"/> does not contain a key identifier.
+        /// This delegate should be used if properties from the configuration retrieved from the authority are necessary to resolve the
+        /// issuer signing key.
+        /// If both <see cref="IssuerSigningKeyResolverWithConfiguration"/> and <see cref="IssuerSigningKeyResolver"/> are set, IssuerSigningKeyResolverWithConfiguration takes
+        /// priority.
+        /// </remarks>
+        public IssuerSigningKeyResolverWithConfiguration IssuerSigningKeyResolverWithConfiguration { get; set; }
```
brentschmaltz: Same as above: With -> Using

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -422,9 +494,25 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
...
+        /// </summary>
+        /// <remarks>
+        /// If set, this delegate will be called to validate the 'issuer' of the token, instead of default processing.
+        /// This means that no default 'issuer' validation will occur.
+        /// Even if <see cref="ValidateIssuer"/> is false, this delegate will still be called.
+        /// This delegate should be used if properties from the configuration retrieved from the authority are necessary to validate the issuer.
+        /// If both <see cref="IssuerValidatorWithConfiguration"/> and <see cref="IssuerValidator"/> are set, IssuerValidatorWithConfiguration takes
+        /// priority.
+        /// </remarks>
+        public IssuerValidatorWithConfiguration IssuerValidatorWithConfiguration { get; set; }
```
brentschmaltz: Same as above: With -> Using

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -537,6 +625,15 @@ public string RoleClaimType
...
         public SignatureValidator SignatureValidator { get; set; }
 
+        /// <summary>
+        /// Gets or sets a delegate that will be used to validate the signature of the token using the <see cref="TokenValidationParameters"/> and
+        /// the <see cref="BaseConfiguration"/>.
+        /// </summary>
+        /// <remarks>
+        /// If set, this delegate will be called to signature of the token, instead of default processing.
+        /// </remarks>
+        public SignatureValidatorWithConfiguration SignatureValidatorWithConfiguration { get; set; }
```
brentschmaltz: Same as above: With -> Using

brentschmaltz:
@mafurman just looking for some name changes.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
mafurman: Ah, I see. Thanks for letting me know.
I'll incorporate some of her PR into this one; we'll need to coordinate how these PRs go in since they modify the same area.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: If this will be addressed by JsonClaimSet, I'll hold off on making changes to this.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: All instances of WithConfiguration have been replaced with UsingConfiguration

mafurman:
> Should this if block be moved to after line 1017 right after the tokenValidationResult is set?

@dannybtsai  What is your reasoning for this? We probably want to execute the success case first since (hopefully) it's more likely to happen.

> This is probably not realistic but in theory an infinite recursion can happen:

@dannybtsai I added a comment explaining in more detail what's happening here. Let me know if it makes sense!

> Note: step 4 & 5 is calling a public from private. My feeling is that there might be common functionality that can be extracted and shared.

I'm not sure I understand what you mean. I do think we need to call the whole ValidateToken method on the actor token as it is essentially separate from the outer token in terms of properties that need validating (even though it is located inside of it).

> @mafurman just looking for some name changes.

Name changes have been made. :)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: whoops, thanks for catching that!

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: @dannybtsai Done!

As far as as reference equals goes, the reasoning behind using it was to minimize computation time. Checking equality through the use of an Equals method would be more time consuming, but I can see wanting to do that in order to avoid cases where we obtain a new configuration but it's not any different than the previous. That might actually result in us attempting to validate more than necessary.

How likely is this to happen with the value we have set by default for the automatic refresh interval?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Really good suggestion Danny; helps clean up the code a lot.
It shouldn't null ref; I check the exception type before casting it.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
mafurman: Ah, thanks for letting me know! I implemented some of her changes here.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
mafurman: Thanks for your input! :)

brentschmaltz (approved):

dannybtsai (approved):

PR #1783, opened 2021-11-22
Simplify the EventBasedLRUCache and Allows Skipping LRU

### LINKED ISSUE #1782: Simplify the LRU Cache
This task is to simplify the EventBaseLRUCache a bit and improve the performance by adding a new flag that skips the LRU maintenance of the _double LinkedList as not everyone needs it (especially if the cache is never full and removeExpiredValues is false).

dannybtsai:
This task is to simplify the EventBaseLRUCache a bit and improve the performance by adding a new flag that skips the LRU maintenance of the _double LinkedList as not everyone needs it (especially if the cache is never full and removeExpiredValues is false).


src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -136,24 +102,43 @@ private int EventQueuePollingInterval
...
             IEqualityComparer<TKey> comparer = null,
             bool removeExpiredValues = false,
-            int removeExpiredValuesIntervalInSeconds = 300)
+            int removeExpiredValuesIntervalInSeconds = 300,
+            bool maintainLRU = false)
         {
             _capacity = capacity > 0 ? capacity : throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(capacity)));
             _options = options;
             _map = new ConcurrentDictionary<TKey, LRUCacheItem<TKey, TValue>>(comparer ?? EqualityComparer<TKey>.Default);
             _removeExpiredValuesIntervalInSeconds = 1000 * removeExpiredValuesIntervalInSeconds;
```
brentschmaltz: why do we multiple by 1000 here?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -272,19 +251,6 @@ private void RemoveLRUs()
             }
```
brentschmaltz, line 352: Could we write: _doubleLinkedList.RemoveLast()?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -136,24 +102,43 @@ private int EventQueuePollingInterval
...
+            _maintainLRU = maintainLRU;
+            _dueForExpiredValuesRemoval = DateTime.UtcNow.AddSeconds(_removeExpiredValuesIntervalInSeconds);
         }
 
+        /// <summary>
+        /// Occurs when the application is ready to exit.
+        /// </summary>
+        /// <param name="sender">The sender of the event.</param>
+        /// <param name="e">The event argument.</param>
+        private void DomainProcessExit(object sender, EventArgs e) => StopEventQueueTask();
```
brentschmaltz: How are these methods used?

brentschmaltz (approved):

pmaytak:
@dannybtsai 
What is the difference operationally between this EventBasedLRUCache with maintainLRU = false and MaximumSizeCache in #1771? The way I udnerstand it - when compacting MaximumSizeCache removes entries randomly from the dictionary (there's no linked list), while this one removes expired entries, and also when compacting it removes entries based on the linked list? (Although the LRU order of the list is not maintained?)


@pmaytak , what you described is correct and the idea in MaximumSizeCache is to provide a simple/lightweight, non-LRU cache that does not have Linkedlist. So, maintaining a Linkedlist is the only difference.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -307,11 +333,14 @@ public bool SetValue(TKey key, TValue value, DateTime expirationTime)
             {
                 cacheItem.Value = value;
                 cacheItem.ExpirationTime = expirationTime;
-                AddActionToEventQueue(() =>
+                if (_maintainLRU)
```
brentschmaltz, line 404: @dannybtsai I am not understanding how the map is compacted.
The method RemoveLRUs() checks the _doubleLinkedList.Count>0 on line 291, but i don't see where an item is added to _doubleLinkedList ?

brentschmaltz (commented):
Trying to understand how the cache gets cleared out

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: Danny and i reviewed this and it seems fine, i just needed some hand holding....

However, we decided, that is we are using the MRU part, just skip adding to the _doubleLinkList and modify the RemoveLRU to only consider the _doubleLinkList if MRU logic is in.

We also suggested to rename RemoveLRU to Compact.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -347,21 +415,26 @@ public bool SetValue(TKey key, TValue value, DateTime expirationTime)
                 // if cache is at _maxCapacityPercentage, trim it by _compactionPercentage
                 if ((double)_map.Count / _capacity >= _maxCapacityPercentage)
                 {
-                    _eventQueue.Enqueue(() =>
+                    if (_maintainLRU)
+                        _eventQueue.Enqueue(CompactLRU);
```
brentschmaltz, line 419: nice to create separate methods for LRU and Non-LRU

brentschmaltz (approved):

PR #1784, opened 2021-11-23
Adding immediate retry on network failure to the token validation flow using ConfigurationManager

mafurman:
When retrieving the configuration, retry logic has been added that tries to obtain the configuration again immediately on network failure (HttpStatusCode.RequestTimeout and HttpStatusCode.ServiceUnavailable). Open to hearing opinions on precisely which status codes we'd like to consider "network failure." 

The keys used by the HttpDocumentRetriever to store information related to the HttpResponse have been moved to an HttpResponseConstants class inside of Microsoft.IdentityModel.Tokens, so that JsonWebToken handler can access them. These constants, though public, have not been in a release yet, so I believe it should be safe to move them over to a different location.


src/Microsoft.IdentityModel.Tokens/HttpResponseConstants.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Constants used by Microsoft.IdentityModel.Protocols.HttpDocumentRetriever to store information regarding
+    /// an unsuccessful HTTP response in the IOException thrown by GetDocumentAsync().
+    /// </summary>
+    class HttpResponseConstants
+    {
+        /// <summary>
+        /// The key is used to add status code into <see cref="Exception.Data"/>.
```
mafurman: @ciaozhang Would like to hear your perspectives on whether this change makes sense to you.

ciaozhang: LGTM

ciaozhang (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1008,9 +1010,26 @@ private TokenValidationResult ValidateToken(string token, JsonWebToken outerToke
...
+                        {
+                            LogHelper.LogInformation(LogHelper.FormatInvariant(TokenLogMessages.IDX10265, ioException.Data[HttpResponseConstants.StatusCode], ioException.Data[HttpResponseConstants.ResponseContent], validationParameters.ConfigurationManager.MetadataAddress));
+                            currentConfiguration = validationParameters.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
+                        }
+                        catch (Exception retryException)
+                        {
+                            LogHelper.LogInformation(LogHelper.FormatInvariant(TokenLogMessages.IDX10261, validationParameters.ConfigurationManager.MetadataAddress, retryException.ToString()));
+                        }
+                    }
                 }
```
dannybtsai, line 1016: I think it is better to keep the try/catch cleaner and avoid adding too much business logic into exception handling.
Consider the following alternative using for loop and the catch/when:

```
int maxAttempt = 2;
for (int i = 1; i <= maxAttempt; i++)
{
	try
	{
		currentConfiguration = validationParameters.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
	}
	catch (InvalidOperationException e) when
			(e.Message.Contains("IDX20803 :") &&
			 e.InnerException != null && e.InnerException is IOException ioException &&
			 ioException.Data.Contains(HttpResponseConstants.StatusCode) &&
			 (ioException.Data[HttpResponseConstants.StatusCode].Equals(HttpStatusCode.RequestTimeout) || ioException.Data[HttpResponseConstants.StatusCode].Equals(HttpStatusCode.ServiceUnavailable)))
	{
		// The exception is not re-thrown here as the TokenValidationParameters may have the issuer and signing key set
		// directly on them.
		LogHelper.LogInformation(LogHelper.FormatInvariant(TokenLogMessages.IDX10261, validationParameters.ConfigurationManager.MetadataAddress, e.ToString()));

		if (i < maxAttempt)
			LogHelper.LogInformation(LogHelper.FormatInvariant(TokenLogMessages.IDX10265, ioException.Data[HttpResponseConstants.StatusCode], ioException.Data[HttpResponseConstants.ResponseContent], validationParameters.ConfigurationManager.MetadataAddress));
	}
	catch (Exception ex)
	{
		// The exception is not re-thrown here as the TokenValidationParameters may have the issuer and signing key set
		// directly on them.
		LogHelper.LogInformation(LogHelper.FormatInvariant(TokenLogMessages.IDX10261, validationParameters.ConfigurationManager.MetadataAddress, ex.ToString()));
		break;
	}
}
```

dannybtsai:
>                 {

Though small, I think it is better to avoid making the same call multiple times:

```
var resultType = tokenValidationResult.Exception.GetType();
if (resultType.Equals(typeof(SecurityTokenInvalidSignatureException))
   || resultType.Equals(typeof(SecurityTokenInvalidSigningKeyException))
   || resultType.Equals(typeof(SecurityTokenInvalidIssuerException))
   || (resultType.Equals(typeof(SecurityTokenUnableToValidateException))
   // we should not try to revalidate with the LKG or request a refresh if the token has an invalid lifetime
   && (tokenValidationResult.Exception as SecurityTokenUnableToValidateException).ValidationFailure != ValidationFailure.InvalidLifetime)
   || resultType.Equals(typeof(SecurityTokenSignatureKeyNotFoundException)))
```

And if the type includes its derived type then using the `is `operator might be better:

` if (resultType  is SecurityTokenInvalidSignatureException)`
...


---
Refers to: src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs:1054 in 538278e. [](commit_id = 538278e2d98c67e715393f2981cb1b71e51db660, deletion_comment = False)

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -102,6 +102,7 @@ internal static class LogMessages
         public const string IDX10262 = "IDX10262: One of the issuers in TokenValidationParameters.ValidIssuers was null or an empty string. See https://aka.ms/wilson/tokenvalidation for details.";
         public const string IDX10263 = "IDX10263: Unable to re-validate with ConfigurationManager.LastKnownGoodConfiguration as it is expired.";
         public const string IDX10264 = "IDX10264: Reading issuer signing keys from validation parameters and configuration.";
+        public const string IDX10265 = "IDX10265: Network error occurred. Status code: '{0}'. \nResponse content: '{1}'. \nAttempting to retrieve configuration again from authority: '{2}'.";
```
dannybtsai: nit: If the response content can be long, consider moving it to the end.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1008,9 +1010,26 @@ private TokenValidationResult ValidateToken(string token, JsonWebToken outerToke
...
                 catch (Exception ex)
                 {
-                    // Keep going with the validation as the TokenValidationParameters may have the issuer and signing key set
+                    // The exception is not re-thrown here as the TokenValidationParameters may have the issuer and signing key set
                     // directly on them.
                     LogHelper.LogInformation(LogHelper.FormatInvariant(TokenLogMessages.IDX10261, validationParameters.ConfigurationManager.MetadataAddress, ex.ToString()));
+
+                    // Retry configuration retrieval on network failure (ServiceUnavailable or RequestTimeout).
+                    if (ex.GetType().Equals(typeof(InvalidOperationException)) && ex.Message.Contains("IDX20803 :")
```
dannybtsai: >ex.Message.Contains("IDX20803 :") [](http://example.com/codeflow?start=82&length=33)

This is not needed:
1. The `GetBaseConfigurationAsync()` only throws `IOException `with "IDX20803 :".
2. If `GetBaseConfigurationAsync()` changes the message this may break.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Thanks for this suggestion Danny! It definitely simplifies the code quite a bit. For ease of review (and my own sanity when rebasing 😅) I've combined this PR with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1785, which improves logging during this whole configuration retrieval process.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1784/commits/47e0c5bf3236c4fc08d1f0f56dd250d5e6051ddb

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: That's a good point; removing this makes the code more resilient to future changes in log messages. Fixed. :) https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1784/commits/47e0c5bf3236c4fc08d1f0f56dd250d5e6051ddb

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
mafurman: That's a good point; I imagine it's likely longer than the exception in string form. 

I think status code is something that should be present earlier on in the message though (for debugging purposes); do you think that separating 'status code' from 'response content' in the log message makes sense?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Good point! Addressed :) https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1784/commits/47e0c5bf3236c4fc08d1f0f56dd250d5e6051ddb

mafurman:
> And if the type includes its derived type then using the is operator might be better:

@dannybtsai Since you mentioned this in a previous PR review, this change has already been made. :)

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
dannybtsai: I meant moving the {authority} in front of the {content}, so it is easier to see in case the {content} is long:
{Status Code}, {Authority}, {Content}

dannybtsai (approved):

src/Microsoft.IdentityModel.Tokens/HttpResponseConstants.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Constants used by Microsoft.IdentityModel.Protocols.HttpDocumentRetriever to store information regarding
+    /// an unsuccessful HTTP response in the IOException thrown by GetDocumentAsync().
+    /// </summary>
+    class HttpResponseConstants
```
brentschmaltz: use explicit visibility (private , internal)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1052,20 +1043,75 @@ private TokenValidationResult ValidateToken(string token, JsonWebToken outerToke
...
+                        if (lastConfig != currentConfiguration)
+                            return decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParameters, currentConfiguration) : ValidateJWS(token, validationParameters, currentConfiguration); ;
+                    }
                 }
             }
 
             return tokenValidationResult;
         }
 
+        private static BaseConfiguration GetConfigurationAndRetryOnNetworkError(TokenValidationParameters validationParameters)
```
brentschmaltz: Was there any thought about putting this logic in HttpDocumentRetriever?
Then all handlers would benefit.
The logic for examining the exception is simplified.
The only retry will be, for sure, http related, different document retrievers may have their own logic for retry.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
mafurman: Makes sense. :) As of this commit https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1784/commits/daa8b451296c5da82fa6bcf59d176572ea67b6e0, the logging related to this has been moved to the HttpDocumentRetriever. The value of {content} is logged at the end.

src/Microsoft.IdentityModel.Tokens/HttpResponseConstants.cs
mafurman: Not an issue anymore since the properties in this class have been moved back to the HttpDocumentRetriever. https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1784/commits/daa8b451296c5da82fa6bcf59d176572ea67b6e0

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
mafurman: Code has been moved to HttpDocumentRetriever. https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1784/commits/daa8b451296c5da82fa6bcf59d176572ea67b6e0

test/Microsoft.IdentityModel.Protocols.Tests/HttpDocumentRetrieverTests.cs
```diff
@@ -179,6 +181,36 @@ public static TheoryData<DocumentRetrieverTheoryData> GetMetadataTheoryData
                     TestId = "Client Miss Configuration"
                 });
 
+                theoryData.Add(new DocumentRetrieverTheoryData
+                {
+                    Address = "https://login.windows.net/f686d426-8d16-42db-81b7-ab578e110ccd/.well-known/openid-configuration",
+                    DocumentRetriever = new HttpDocumentRetriever(HttpResponseMessageUtils.SetupHttpClientThatReturns("ValidJson.json", HttpStatusCode.RequestTimeout)),
+                    TestId = "RequestTimeout_RefreshSucceeds"
```
ciaozhang, line 186: The test ID is same as  the next one.

src/Microsoft.IdentityModel.Protocols/HttpResponseMessageUtils.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Net;
+using System.Net.Http;
+using System.Text;
+
+namespace Microsoft.IdentityModel.Protocols
+{
+    /// <summary>
+    /// Utility methods for constructing <see cref="HttpResponseMessage"/> objects for testing purposes.
+    /// </summary>
+    public static class HttpResponseMessageUtils
```
brentschmaltz: I think this class can be internal or moved under test

brentschmaltz (changes requested):
There is a public class that is not needed.

mafurman:
@brentschmaltz Both HttpResponseMessageUtils and MockHttpMessageHandler have been moved over to Microsoft.IdentityModel.TestUtils. https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1784/commits/b7f4fa41af7dc0848266ade64d9284610c614ea6

src/Microsoft.IdentityModel.Protocols/HttpResponseMessageUtils.cs
mafurman: Moved over to Microsoft.IdentityModel.TestUtils: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1784/commits/b7f4fa41af7dc0848266ade64d9284610c614ea6

test/Microsoft.IdentityModel.Protocols.Tests/HttpDocumentRetrieverTests.cs
mafurman: Good catch. I've fixed the test name to reflect what's actually being tested. https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1784/commits/b7f4fa41af7dc0848266ade64d9284610c614ea6

brentschmaltz (approved):

PR #1797, opened 2021-12-29
Adding LKG feature into JwtSecurityTokenHandler

mafurman:
Adding the LKG feature to JwtSecurityTokenHandler (already present in JsonWebTokenHandler).

This mainly includes the following logic (the PRs below are related to JsonWebTokenHandler):
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1695 (validation using BaseConfiguration)
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1723 (LKG)
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1784 (refresh on network error)

Test data sets for this feature are intentionally shared between JwtSecurityTokenHandler and JsonWebTokenHandler to ensure feature consistency (and reduce code duplication).


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1047,7 +1047,7 @@ private TokenValidationResult ValidateToken(string token, JsonWebToken outerToke
                         else
                         {
                             currentConfiguration = validationParameters.ConfigurationManager.LastKnownGoodConfiguration;
-                            tokenValidationResult = decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParameters, currentConfiguration) : ValidateJWS(token, validationParameters, currentConfiguration); ;
+                            tokenValidationResult = decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParameters, currentConfiguration) : ValidateJWS(token, validationParameters, currentConfiguration);
```
TimHannMSFT: >  [](http://example.com/codeflow?start=222&length=1)

supernit: trailing space

TimHannMSFT: and as long as you're touching this line splitting the ternary statement to three lines makes it more readable imo but that might be just a style thing.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -742,35 +743,163 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
...
         }
 
+        /// <summary>
+        ///  Private method for token validation, responsible for:
+        ///  (1) Obtaining a configuration from the <see cref="TokenValidationParameters.ConfigurationManager"/>.
+        ///  (2) Revalidating using the Last Known Good Configuration (if present), and obtaining a refreshed configuration (if necessary) and revalidating using it.
+        /// </summary>
+        /// <param name="token">The JWS string, or the decrypted token if the token is a JWE.</param>
+        /// <param name="outerToken">If the token being validated is a JWE, this is the <see cref="JwtSecurityToken"/> that represents the outer token.
+        ///  If the token is a JWS, the value of this parameter is null.
```
TimHannMSFT: >null [](http://example.com/codeflow?start=67&length=4)

supernit: <see langword="null" /> #Resolved

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -742,35 +743,163 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
...
+        private ClaimsPrincipal ValidateToken(string token, JwtSecurityToken outerToken, TokenValidationParameters validationParameters, out SecurityToken signatureValidatedToken)
+        {
+            BaseConfiguration currentConfiguration = null;
+            if (validationParameters.ConfigurationManager != null)
+            {
+                try
+                {
+                    currentConfiguration = validationParameters.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
+                }
+                catch (Exception ex)
```
TimHannMSFT, line 781: >Exception ex [](http://example.com/codeflow?start=23&length=12)

Is there a set of library specific exceptoins we can catch/swallow here? The comment makes sense for a known exception type but this will catch everything. Logging all exceptions makes sense to me but swallowing all exceptions seems dangerous.

test/Microsoft.IdentityModel.TestUtils/TestTheoryData.cs
```diff
@@ -1,7 +1,13 @@
-﻿using Microsoft.IdentityModel.Tokens;
+﻿using System;
+using System.Collections.Generic;
+using System.IdentityModel.Tokens.Jwt;
+using System.IO;
+using System.Net;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
```
TimHannMSFT: looks like only usings changes to this file, is that correct? If these were required how was it building before, if they're not, why are they added?

test/System.IdentityModel.Tokens.Jwt.Tests/JwtTestDatasets.cs
```diff
@@ -0,0 +1,755 @@
...
+
+                var requestNotFoundException = new IOException();
+                requestNotFoundException.Data.Add(HttpDocumentRetriever.StatusCode, HttpStatusCode.NotFound);
+                requestNotFoundException.Data.Add(HttpDocumentRetriever.ResponseContent, "requestNotFoundException");
+
+                return new TheoryData<JwtTheoryData>
+                {
+                    new JwtTheoryData
+                    {
+                        TestId = nameof(Default.AsymmetricJws) + "_" + "TVPInvalid" + "_" + "ConfigValid",
```
TimHannMSFT: >"_" + "TVPInvalid" + "_" + "ConfigValid", [](http://example.com/codeflow?start=65&length=41)

nit: here and below, what's the purpose of splitting up the strings like this? Why not

TestId = nameof(Default.AsymmetricJws) + "_TVPInvalid_ConfigValid",

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -742,35 +743,163 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
...
+                {
+                    // The exception is not re-thrown as the TokenValidationParameters may have the issuer and signing key set
+                    // directly on them, allowing the library to continue with token validation.
```
TimHannMSFT, line 784: Maybe I'm misunderstanding this comment but can't we check the values of TVP here to see if this is true? If so continue, else throw?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -742,35 +743,163 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
...
+                    return claimsPrincipal;
+                }
+                // using 'GetType()' instead of 'is' as SecurityTokenUnableToValidException (and others) extend SecurityTokenInvalidSignatureException
+                // we want to make sure that the clause for SecurityTokenUnableToValidateException is hit so that the ValidationFailure is checked
+                else if (exceptionThrown.GetType().Equals(typeof(SecurityTokenInvalidSignatureException))
+                   || exceptionThrown is SecurityTokenInvalidSigningKeyException
+                   || exceptionThrown is SecurityTokenInvalidIssuerException
+                   || (exceptionThrown is SecurityTokenUnableToValidateException
+                   // we should not try to revalidate with the LKG or request a refresh if the token has an invalid lifetime
+                   && (exceptionThrown as SecurityTokenUnableToValidateException).ValidationFailure != ValidationFailure.InvalidLifetime)
```
TimHannMSFT: consider:

(exceptionThrown as SecurityTokenUnableToValidateException)?.ValidationFailure != ValidationFailure.InvalidLifetime

should have the same behavior but less verbose

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -742,35 +743,163 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
...
+            {
+                try
+                {
+                    currentConfiguration = validationParameters.ConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
+                }
+                catch (Exception ex)
+                {
+                    // The exception is not re-thrown as the TokenValidationParameters may have the issuer and signing key set
+                    // directly on them, allowing the library to continue with token validation.
+                    LogHelper.LogWarning(LogHelper.FormatInvariant(TokenLogMessages.IDX10261, validationParameters.ConfigurationManager.MetadataAddress, ex.ToString()));
```
TimHannMSFT: >validationParameters.ConfigurationManager.MetadataAddress [](http://example.com/codeflow?start=94&length=57)

This should be able to be marked non-pii right? #Resolved

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -742,35 +743,163 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
...
+            }
+
+            Exception exceptionThrown;
+            ClaimsPrincipal claimsPrincipal = outerToken != null ? ValidateJWE(token, outerToken, validationParameters, currentConfiguration, out signatureValidatedToken, out exceptionThrown) :
+                ValidateJWS(token, validationParameters, currentConfiguration, out signatureValidatedToken, out exceptionThrown);
+            if (validationParameters.ConfigurationManager != null)
+            {
+                if (claimsPrincipal != null)
+                {
+                    // Set current configuration as LKG if it exists and is not the same as the LKG.
```
TimHannMSFT: >not the same as the LKG. [](http://example.com/codeflow?start=76&length=24)

Doesn't == check the reference is identical?

Wouldn't this comment more accurately be "has not already been set to the current config" or something similar.

To me this implies that it's checking the config isn't logically the same, which I don't beleive this is actually doing.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -742,35 +743,163 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
...
+                var innerToken = ValidateSignature(decryptedJwt, validationParameters, currentConfiguration);
+                outerToken.InnerToken = innerToken;
+                signatureValidatedToken = outerToken;
+                return ValidateTokenPayload(innerToken, validationParameters, currentConfiguration);
+            }
+            catch (Exception ex)
+            {
+                exceptionThrown = ex;
+                return null;
+            }
```
TimHannMSFT, line 861: I'm not crazy about this sort of pattern in general (catching generic exceptions and then assuming the caller will always throw in the right way) but if we go down this route at least consider using ExceptionDispatchInfo instead. I believe the way the code is currently written, it will strip the stack trace info which I don't think is the intent and hurts debugging efforts.

[https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo.capture?redirectedfrom=MSDN&view=net-6.0#System_Runtime_ExceptionServices_ExceptionDispatchInfo_Capture_System_Exception](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo.capture?redirectedfrom=MSDN&view=net-6.0#System_Runtime_ExceptionServices_ExceptionDispatchInfo_Capture_System_Exception)_
 #Resolved

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: @TimHannMSFT future work that is planned right after this PR related to using System.Text.Json will clean this up.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -742,35 +743,163 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
             if (tokenParts.Length != JwtConstants.JwsSegmentCount && tokenParts.Length != JwtConstants.JweSegmentCount)
                 throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX12741, token)));
 
-            ClaimsPrincipal claimsPrincipal = null;
-            SecurityToken signatureValidatedToken = null;
-
+            ClaimsPrincipal claimsPrincipal;
+            SecurityToken signatureValidatedToken;
```
brentschmaltz: if we were to pass 'validatedToken' to ValidateToken(...), we could remove this local variable and the assignment below. 
we could then remove the local variable ClaimsPrincipal and then write:

```
if (...)
{
   ...
   return ValidateToken(decryptedJwt, jwtToken, validationParameters, out validatedToken);
}
else
{
   return ValidateToken(token, null, validationParameters, out validatedToken);
}
```

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: @TimHannMSFT any exceptions that are about to bring down the appdomain will fail again shortly, no?
What would you suggest? At one point in time there was an effort to fail fast on certain exceptions, are you suggesting we have such a list?
This model is to ensure for any reason if Configuration was not able to be obtained, we can move forward as the artifacts needed for validation may be in TVP.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: @TimHannMSFT it is not possible to check due to extensibility, for example a custom key retriever.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: I agree, metadata address is a public value.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: I prefer ReferenceEquals(...) as == can be overridden.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: I agree. We should not rethrow the exception.
We can throw a new exception, added the current exception as the inner.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
TimHannMSFT: +1 that makes it more explicit, good point.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
TimHannMSFT: It depends and fatal system exceptions can have goofy things occur if you keep carrying on as if nothing happened. I don't think anyone can confidently say how the system will behave if it's out of memory...perhpas garbage collection occurs at just the right time and now there's more but the state is corrupted, I don't think the code can or should try to reason about this, fail fast.

My suggestion is catch what you know and throw what you don't. Absolutely log everything but don't try and pretend your code is smart enough to handle all contigencies.

To my mind there are three cases:
1. System exceptions - the system is now in a state you can't really control or predict, you should just bubble this up after logging.
2. Some underlying library has thrown an exception you didn't account for in your code and don't know how to handle - your assumptions are wrong and it's no longer safe to continue processing because you'd definitionally in  state you don't understand.
3. An exceptional case you know about - here you should catch the excepiton by the known type and handle it in the way you know how.

There are exceptions to the rule of course but I think we should at least have a long comment explaining why we'd make an exception in a library. (If you're running a service that "can't" fail at some critical point maybe it makes sense to try and do everything you can to hang on but IMO we don't hae the context to make those sorts of calls at our layer...fail fast)

TimHannMSFT: Also meant to acknowledge that at least some of the sytem exceptions aren't actually caught by this pattern for just this reason (IIRC this was added later in .NET) So for some of (1) we're covered but I'm not sure about all cases and (2) still is a concern.

test/Microsoft.IdentityModel.TestUtils/TestTheoryData.cs
ciaozhang: Removed unused package and added copy right

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
ciaozhang: Marked as PII

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: @TimHannMSFT even logging can be a problem. Last i checked with the .net team is that most exceptions that would cause major issues to ignore we won't see, but we should investigate how the runtime handles system exceptions.
In this specific case, all we care about is we couldn't get metadata and are going to use what we have.

We have this same pattern throughout the runtime and i agree we should understand this better, in a holistic manner as the clr has changed since i last looked.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -742,35 +743,163 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
...
+                {
+                    // Set current configuration as LKG if it exists and is not the same as the LKG.
+                    if (currentConfiguration != null && currentConfiguration != validationParameters.ConfigurationManager.LastKnownGoodConfiguration)
+                        validationParameters.ConfigurationManager.LastKnownGoodConfiguration = currentConfiguration;
+
+                    return claimsPrincipal;
+                }
+                // using 'GetType()' instead of 'is' as SecurityTokenUnableToValidException (and others) extend SecurityTokenInvalidSignatureException
+                // we want to make sure that the clause for SecurityTokenUnableToValidateException is hit so that the ValidationFailure is checked
+                else if (exceptionThrown.GetType().Equals(typeof(SecurityTokenInvalidSignatureException))
```
brentschmaltz: If we had a helper method such as RecoverableException and put all the code to decide if we can try and use LKG in there this would be easier to read.

If this method was reachable for other handlers, we would have consistency between them and less maintenace.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
ciaozhang: @brentschmaltz  It will break some tests if we write like this. For some tests if ValidteToken() throws an exception, this implementation will still out a token, and it shouldn't be happened.
Add this branch for above implementation: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/tree/jwtsec_lkg/validateToken

test/System.IdentityModel.Tokens.Jwt.Tests/JwtTestDatasets.cs
ciaozhang: Changed

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -742,35 +745,160 @@ public override ClaimsPrincipal ValidateToken(string token, TokenValidationParam
...
+                }
+            }
+
+            if (claimsPrincipal != null)
+                return claimsPrincipal;
+
+            exceptionThrown.Throw();
+
+            // This should be unreachable code, adding to make the complier happy.
+            return null;
```
ciaozhang, line 837: Not found other solution for Getting "not all code paths return a value" when using ExceptionDispatchInfo.Throw().

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
brentschmaltz: We realized we can set 'validatedToken' to null in the JWE method that successfully decrypts the JWE but fails the inner JWS validation.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
ciaozhang: Changed to exceptionDispatchInfo.Throw()

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
TimHannMSFT: hmm the github page seems to strip the html I was saying you could do see langword = "null" here instead. very much a nit.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
TimHannMSFT: should be marked *not* for PII is my point

LogHelper.MarkAsNonPII(validationParameters.ConfigurationManager.MetadataAddress)

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
TimHannMSFT: I like this a lot better than what was there before but it really begs the question for me, why are we even passing anything around for the exception, if we're just throwing in the method just above in the call stack why not just do a regular throw here instead of capturing it, letting control return to the caller and then immediately throwing?

TimHannMSFT: Ah I see, it was too long since I looked at this last. It's to allow for the fallback in logic. This makes sense, this part LGTM.

TimHannMSFT (approved):
:shipit:

brentschmaltz (approved):

PR #1798, opened 2022-01-10
rename JwtHandler -> JsonWebTokenHandler

brentschmaltz:
Renamed JwtHandler -> JsonWebTokenHandler to make it easier to understand which handler is being used.
Adjust tests to throw invalid signature, the tests were throwing an ArgumentException as the JWT was not well formed.


TimHannMSFT (approved):
:shipit:

Issue #1795, opened 2021-12-09
AadIssuerValidator checking `Tid` claim twice for `JsonWebtoken`


afscrome:
The AadIssuerValidator seems to check the `Tid` claim twice when validating `JsonWebToken` - the second of these seems redundant.

Based on the `JwtSecurityToken` validation, should the second check instead be checking the `TenantId` claim.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/a782ff50db70fc53119ea6d7a0067222bebca409/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs#L284-L294

PR #1801, opened 2022-01-17
fix check for tenantId

jennyf19:
#1795

### DIFF
--- a/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
+++ b/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
@@ -264,7 +264,7 @@ namespace Microsoft.IdentityModel.Validators
                 if (jsonWebToken.TryGetPayloadValue(AadIssuerValidatorConstants.Tid, out string tid))
                     return tid;
 
-                if (jsonWebToken.TryGetPayloadValue(AadIssuerValidatorConstants.Tid, out string tenantId))
+                if (jsonWebToken.TryGetPayloadValue(AadIssuerValidatorConstants.TenantId, out string tenantId))
                     return tenantId;
 
                 // Since B2C doesn't have "tid" as default, get it from issuer
--- a/test/Microsoft.IdentityModel.Validators.Tests/MicrosoftIdentityIssuerValidatorTest.cs
+++ b/test/Microsoft.IdentityModel.Validators.Tests/MicrosoftIdentityIssuerValidatorTest.cs
@@ -26,6 +26,7 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.IdentityModel.Tokens.Jwt;
 using System.Net.Http;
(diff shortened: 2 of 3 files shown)


jmprieur (approved):
LGTM
Thanks @jennyf19

test/Microsoft.IdentityModel.Validators.Tests/MicrosoftIdentityIssuerValidatorTest.cs
```diff
@@ -305,6 +306,31 @@ public void Validate_IssuerNotInTokenValidationParameters_ReturnsIssuer(string t
...
+            var context = new CompareContext();
+            var validator = new AadIssuerValidator(_httpClient, issuer);
+            var tidClaim = new Claim(tidClaimType, ValidatorConstants.TenantIdAsGuid);
+
+            var issClaim = new Claim(ValidatorConstants.ClaimNameIss, issuer);
+            List<Claim> claims = new List<Claim>();
+            claims.Add(tidClaim);
+            claims.Add(issClaim);
+
+            var jsonWebToken = new JsonWebToken(Default.Jwt(Default.SecurityTokenDescriptor(Default.SymmetricSigningCredentials, claims)));
```
GeoK, line 325: Can we use one of the JsonWebTokenHandler.CreateToken (perhaps the one that takes SecurityTokenDescriptor) methods instead and avoid adding SecurityTokenDescriptor(credentials, claims)?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -264,7 +264,7 @@ private static string GetTenantIdFromToken(SecurityToken securityToken)
                 if (jsonWebToken.TryGetPayloadValue(AadIssuerValidatorConstants.Tid, out string tid))
                     return tid;
 
-                if (jsonWebToken.TryGetPayloadValue(AadIssuerValidatorConstants.Tid, out string tenantId))
+                if (jsonWebToken.TryGetPayloadValue(AadIssuerValidatorConstants.TenantId, out string tenantId))
```
dannybtsai, line 267: >TenantId [](http://example.com/codeflow?start=80&length=8)

Was the old code a mistake? It was getting Tid twice, as tid and tenantId.

test/Microsoft.IdentityModel.Validators.Tests/MicrosoftIdentityIssuerValidatorTest.cs
jennyf19: I tried to do that, but I don't want an encrypted token and also needed some custom values: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1801/files#diff-20c5742d81614f58ad5fd9e5ccc34931d030cbaf5b182bfbc8607707aa3a275bR811 and the claims were not getting added. thoughts?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
brentschmaltz: @dannybtsai there are two possibilities for the 'tid' claim, therefore there are two constants.

brentschmaltz (approved):

test/Microsoft.IdentityModel.Validators.Tests/MicrosoftIdentityIssuerValidatorTest.cs
brentschmaltz: @jennyf19 using SecurityTokenDescriptor does not imply a JWE. Only if the EncryptingCredentials are set will a JWE be created.
You can use this property on SecurityTokenDescriptor to add additional payload claims: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/669515a4ab1e3201768e0d74cd5097db0ba4152b/src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs#L88

brentschmaltz: Add this property to add additional header claims: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/669515a4ab1e3201768e0d74cd5097db0ba4152b/src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs#L97

GeoK (approved):

PR #1806, opened 2022-01-28
Xiao/configuration validatior

ciaozhang:
- Introduce an ConfigurationValidator class that can be used to apply validation rules to a retrieved OpenIdConnectConfiguration.
 - Add an ConfigurationValidationExecutor class that can use one or more ConfigurationValidationValidators to validate the retrieved OpenIdConnectConfiguration.
 - Create a new exception ConfigurationValidationException, and wrap/throw this exception in the ConfigurationValidationResult if the metadata violates the rules.


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -136,18 +136,25 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
                 {
                     try
-                    {
+                    { 
                         // Don't use the individual CT here, this is a shared operation that shouldn't be affected by an individual's cancellation.
                         // The transport should have it's own timeouts, etc..
                         _currentConfiguration = await _configRetriever.GetConfigurationAsync(MetadataAddress, _docRetriever, CancellationToken.None).ConfigureAwait(false);
                         _lastRefresh = now;
                         _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval);
+                        if ( ValidationExecutor != null && ValidationExecutor.ShouldValidateConfiguration )
```
ciaozhang: Introduce a boolean value ShouldValidateConfiguration here

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -136,18 +136,25 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+                            ValidationExecutor.ValidateConfiguration(_currentConfiguration);
                     }
                     catch (Exception ex)
                     {
                         _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
                         if (_currentConfiguration == null) // Throw an exception if there's no configuration to return.
                             throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20803, (MetadataAddress ?? "null")), ex));
+                        else if (ex is ConfigurationValidationException)
+                        {
+                            _currentConfiguration = null;
```
ciaozhang, line 155: Set _currentConfiguratioin to null if it is invalid.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
TimHannMSFT: >  [](http://example.com/codeflow?start=28&length=1)

nit: existing style is not to pad the inside statements with spaces.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationValidationExecutor.cs
```diff
@@ -0,0 +1,104 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
```
TimHannMSFT, line 26: nit: I thought I saw in some PR that the full header isn't required, perhaps only something like :

//
// Copyright (c) Microsoft Corporation.
// All rights reserved.
//
// This code is licensed under the MIT License.
//

If we don't have to, we may as well save some bits :)

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationValidationExecutor.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.ComponentModel;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Protocols
+{
+    /// <summary>
+    /// Validates configuration by using validator.
+    /// </summary>
+    /// <typeparam name="T">The type of the configuration.</typeparam>
+    public class ConfigurationValidationExecutor<T> where T : class
```
TimHannMSFT, line 39: >ConfigurationValidationExecutor [](http://example.com/codeflow?start=17&length=31)

IMO the "Executor" verb isn't really adding any clarity here. This class is in charge of validating configuration so something like "ConfigurationValidtor" 

To me "Executor" and "Manager" classes are invitations to more closely consider the naming or functionality.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationValidationResult.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>
+    /// Represents the result of validation a <see cref="IConfigurationValidator{T}"/>.
+    /// </summary>
+    public class ConfigurationValidationResult
+    {
+        /// <summary>
+        /// Default constructor.
+        /// </summary>
+        public ConfigurationValidationResult() { }
```
TimHannMSFT: Is there a reason to explicitly create the default constructor?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationValidationResult.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Microsoft.IdentityModel.Protocols
+{
+    /// <summary>
+    /// Represents the result of validation a <see cref="IConfigurationValidator{T}"/>.
+    /// </summary>
+    public class ConfigurationValidationResult
+    {
+        /// <summary>
+        /// Default constructor.
```
TimHannMSFT: >Default constructor. [](http://example.com/codeflow?start=12&length=20)

nit: style is to use the following pattern for constructor documentation:

Initializes a new instance of the <see cref="ConfigurationValidationResult"/> class.

src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationValidator.cs
```diff
@@ -0,0 +1,46 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+using System.Threading;
+using System.Threading.Tasks;
```
TimHannMSFT: remove unused usings

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationValidator.cs
```diff
@@ -0,0 +1,107 @@
...
+                }
+                else
+                {
+                    return new ConfigurationValidationResult
+                    {
+                        Exception = new ConfigurationValidationException(LogHelper.FormatInvariant(LogMessages.IDX21819, LogHelper.MarkAsNonPII(MinimumNumberOfKeys), LogHelper.MarkAsNonPII(openIdConnectConfiguration.SigningKeys.Count))),
+                        Succeeded = false
+                    };
+                }
+            }
```
TimHannMSFT, line 76: I think this can be a bit more terse and still pretty clear:

return new ConfigurationValidationResult
                    {
                        Exception = new ConfigurationValidationException(
                            openIdConnectConfiguration.JsonWebKeySet.ConvertKeyInfos.Any() ? 
                            LogHelper.FormatInvariant(LogMessages.IDX21818, LogHelper.MarkAsNonPII(MinimumNumberOfKeys), string.Join("\n", openIdConnectConfiguration.JsonWebKeySet.ConvertKeyInfos.Select(x => x.Key.Kid.ToString() + ": " + string.Join("; ", x.Value)))) :
                            LogHelper.FormatInvariant(LogMessages.IDX21819, LogHelper.MarkAsNonPII(MinimumNumberOfKeys), LogHelper.MarkAsNonPII(openIdConnectConfiguration.SigningKeys.Count))),
                        Succeeded = false
                    };

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationValidator.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        /// <summary>
+        /// The minimum number of keys.
+        /// </summary>
+        public int MinimumNumberOfKeys
+        {
+            get { return _minimumNumberOfKeys; }
+            set
+            {
+                if (value < _minimumNumberOfKeys)
```
TimHannMSFT: >(value < _minimumNumberOfKeys) [](http://example.com/codeflow?start=19&length=30)

this should just be:

if (value < DefaultMinimumNumberOfKeys)

right? Or do you actually mean that once someone sets it to 2 they can't set it back to 1?

In any case that should match the error message.

Maybe add a unit test for this too then.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationValidator.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// Defines a class for validating the OpenIdConnectConfiguration by using default policy.
+    /// </summary>
+    public class OpenIdConnectConfigurationValidator : IConfigurationValidator<OpenIdConnectConfiguration>
+    {
+        private int _minimumNumberOfKeys = DefaultMinimumNumberOfKeys;
+
+        /// <summary>
+        /// 1 is the default minimum number of keys.
+        /// </summary>
+        public static readonly int DefaultMinimumNumberOfKeys = 1;
```
TimHannMSFT: >static readonly [](http://example.com/codeflow?start=15&length=15)

const?

TimHannMSFT: >public [](http://example.com/codeflow?start=8&length=6)

we don't seem to use this outside the class, mas as well make it internal or private to make sure we don't have to support someone taking a dependency.

src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs
```diff
@@ -336,6 +368,37 @@ internal static bool TryCreateToRsaSecurityKey(JsonWebKey webKey, out SecurityKe
...
+            catch (Exception ex)
+            {
+                string convertKeyInfo = LogHelper.FormatInvariant(LogMessages.IDX10813, LogHelper.MarkAsNonPII(typeof(RsaSecurityKey)), webKey, ex);
+                if (convertKeyInfos.ContainsKey(webKey))
+                    convertKeyInfos[webKey].Add(convertKeyInfo);
+                else
+                    convertKeyInfos.Add(webKey, new List<string> { convertKeyInfo });
+
+                LogHelper.LogExceptionMessage(new InvalidOperationException(convertKeyInfo, ex));
+            }
```
TimHannMSFT, line 397: For at least this case we definitely should not use this pattern to achieve the TryCreate pattern.

We control the RsaSecurityKey type. There's no reason for us to use exception handling to perform this check, performance isn't good plus it's a place for bugs to hide.

I'd recommend refactoring RsaSecurityKey to initially call an internal validate method which can return an exception. In the Init method this can just be thrown, and in this method anything returned implies we should return false.

I suspect we can avoid the bad "catch all exceptions pattern" in the other methods too.

IMO "Used as validation" is not sufficient justification to suppress CA1031

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -92,6 +93,11 @@ public JsonWebKeySet(string json)
         [JsonExtensionData]
         public virtual IDictionary<string, object> AdditionalData { get; } = new Dictionary<string, object>();
 
+        /// <summary>
+        /// Gets a <see cref="IDictionary{JsonWebKey, ConvertKeyInfo}"/> that contains convert key information.
```
TimHannMSFT: >ConvertKeyInfo [](http://example.com/codeflow?start=54&length=14)

update?

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -119,13 +125,17 @@ public JsonWebKeySet(string json)
...
             var signingKeys = new List<SecurityKey>();
+            var errorList = new List<Exception>();
             foreach (var webKey in Keys)
             {
                 // skip if "use" (Public Key Use) parameter is not empty or "sig".
                 // https://datatracker.ietf.org/doc/html/rfc7517#section-4.2
                 if (!string.IsNullOrEmpty(webKey.Use) && !webKey.Use.Equals(JsonWebKeyUseNames.Sig, StringComparison.Ordinal))
                 {
-                    LogHelper.LogInformation(LogHelper.FormatInvariant(LogMessages.IDX10808, webKey, webKey.Use));
+                    string convertKeyInfo = LogHelper.FormatInvariant(LogMessages.IDX10808, webKey, webKey.Use);
```
TimHannMSFT: >Use [](http://example.com/codeflow?start=107&length=3)

key use can be non-pii right?

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -180,5 +193,37 @@ public IList<SecurityKey> GetSigningKeys()
 
             return signingKeys;
         }
+
+        private static bool IsValidX509SecurityKey(JsonWebKey webKey, IDictionary<JsonWebKey, List<string>> convertKeyInfos)
+        {
+            if (webKey.X5c == null || webKey.X5c.Count == 0)
+            {
+                convertKeyInfos.Add(webKey, new List<string> { LogHelper.FormatInvariant(LogMessages.IDX10814, LogHelper.MarkAsNonPII(typeof(X509SecurityKey)), webKey, JsonWebKeyParameterNames.X5c) });
```
TimHannMSFT: >JsonWebKeyParameterNames.X5c [](http://example.com/codeflow?start=168&length=28)

not pii

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -138,23 +148,25 @@ public IList<SecurityKey> GetSigningKeys()
 
                     // in this case, even though RSA was specified, we can't resolve.
                     if ((webKey.X5c == null || webKey.X5c.Count == 0) && (string.IsNullOrEmpty(webKey.E) && string.IsNullOrEmpty(webKey.N)))
-                    {
+                    {                     
+                        var missingComponent = new List<string> { JsonWebKeyParameterNames.X5c, JsonWebKeyParameterNames.E, JsonWebKeyParameterNames.N };
+                        string convertKeyInfo = LogHelper.FormatInvariant(LogMessages.IDX10814, LogHelper.MarkAsNonPII(typeof(RsaSecurityKey)), webKey, string.Join(", ", missingComponent));
```
TimHannMSFT: >string.Join(", ", missingComponent) [](http://example.com/codeflow?start=152&length=35)

not pii

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -180,5 +193,37 @@ public IList<SecurityKey> GetSigningKeys()
...
+            var missingComponent = new List<string>();
+            if (string.IsNullOrWhiteSpace(webKey.E))
+                missingComponent.Add(JsonWebKeyParameterNames.E);
+
+            if (string.IsNullOrWhiteSpace(webKey.N))
+                missingComponent.Add(JsonWebKeyParameterNames.N);
+
+            if (missingComponent.Count > 0)
+            {
+                string convertKeyInfo = LogHelper.FormatInvariant(LogMessages.IDX10814, LogHelper.MarkAsNonPII(typeof(RsaSecurityKey)), webKey, string.Join(", ", missingComponent));
```
TimHannMSFT: >string.Join(", ", missingComponent)); [](http://example.com/codeflow?start=144&length=37)

not pii

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationValidator.cs
```diff
@@ -0,0 +1,99 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+using System;
+using System.Linq;
+using Microsoft.IdentityModel.Logging;
+using Microsoft.IdentityModel.Tokens;
```
brentschmaltz, line 31: using is not needed

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationValidator.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <summary>
+        /// The minimum number of keys.
+        /// </summary>
+        public int MinimumNumberOfKeys
+        {
+            get { return _minimumNumberOfKeys; }
+            set
+            {
+                if (value < DefaultMinimumNumberOfKeys)
+                    throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(value), LogHelper.FormatInvariant(LogMessages.IDX21816, LogHelper.MarkAsNonPII(DefaultMinimumNumberOfKeys), LogHelper.MarkAsNonPII(value))));
```
brentschmaltz, line 93: Can you please stack this line, put parameters on separate lines.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -136,18 +136,25 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                 if (_syncAfter <= now)
                 {
                     try
-                    {
+                    {
```
brentschmaltz, line 139: We can remove this extra space.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -136,18 +136,25 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
                 {
                     try
-                    {
+                    { 
                         // Don't use the individual CT here, this is a shared operation that shouldn't be affected by an individual's cancellation.
                         // The transport should have it's own timeouts, etc..
                         _currentConfiguration = await _configRetriever.GetConfigurationAsync(MetadataAddress, _docRetriever, CancellationToken.None).ConfigureAwait(false);
                         _lastRefresh = now;
                         _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval);
+                        if (ValidationExecutor != null && ValidationExecutor.ShouldValidateConfiguration)
```
brentschmaltz, line 145: Why would one add a ValidationExecutor and then check if it should be used?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationValidationExecutor.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        /// <summary>
+        /// Instantiates a new <see cref="ConfigurationValidationExecutor{T}"/> that validate retrived configuration data.
+        /// </summary>
+        /// <param name="defaultConfigurationValidator">The default configuration validator.</param>
+        public ConfigurationValidationExecutor(IConfigurationValidator<T> defaultConfigurationValidator)
```
brentschmaltz, line 45: What was the reasoning behind creating this class and just having IConfigurationValidator<T> as a property on ConfigurationManager<T>?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationValidationExecutor.cs
```diff
@@ -0,0 +1,104 @@
...
+        public ConfigurationValidationExecutor(IConfigurationValidator<T> defaultConfigurationValidator, IConfigurationValidator<T> additionalConfigurationValidator)
+            : this(defaultConfigurationValidator)
+        {
+            AdditionalConfigurationValidator = additionalConfigurationValidator ?? throw LogHelper.LogArgumentNullException(nameof(additionalConfigurationValidator));
+        }
+
+        /// <summary>
+        /// The additional Configuration Validator.
+        /// </summary>
+        public IConfigurationValidator<T> AdditionalConfigurationValidator { get; }
```
brentschmaltz, line 64: If this is for chaining, then another pattern is to have the IConfigurationValidator expose an 'Inner' property of type IConfigurationValidator<T>.

Why do we need chaining?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationValidator.cs
```diff
@@ -0,0 +1,99 @@
...
+            if (openIdConnectConfiguration.JsonWebKeySet == null || openIdConnectConfiguration.JsonWebKeySet.Keys.Count == 0)
+            {
+                return new ConfigurationValidationResult
+                {
+                    Exception = new ConfigurationValidationException(LogMessages.IDX21817),
+                    Succeeded = false
+                };
+            }
+
+            if (openIdConnectConfiguration.SigningKeys.Count < MinimumNumberOfKeys)
```
brentschmaltz, line 66: I think this can be simplified.
If openIdConnectConfiguration.JsonWebKeySet.Count != openIdConnectConfiguration.JsonWebKeySet.Keys.Count, then some of the keys could not be converted from a JsonWebKey to a SecurityKey.
To determine which of those that were unable to be converted, just check the property on the JsonWebKey.ConvertedSecurityKey, which will be set when the the JsonWebKey has been converted.
This happens here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/adc8d59d4aaae502f77fd8f758e4f020045f69eb/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationRetriever.cs#L100

src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs
```diff
@@ -311,6 +314,35 @@ internal static bool TryConvertToX509SecurityKey(JsonWebKey webKey, out Security
             return false;
         }
 
+        internal static bool TryConvertToX509SecurityKey(JsonWebKey webKey, IDictionary<JsonWebKey, List<string>> convertKeyInfos, out SecurityKey key)
```
brentschmaltz, line 317: If as mentioned above, we can determine which keys were unable to be converted in a different manner, this new api may not be needed.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationValidator.cs
ciaozhang: Updated

PR #1807, opened 2022-02-02
Make M.IM.Tokens visible to S2S.Tokens

ciaozhang:
Make M.IM,Tokens visible toS2S.Tokens since the LRUCache is internal class and we want to consume it in SAL.


TimHannMSFT (approved):

brentschmaltz (commented):
looks good

PR #1810, opened 2022-02-09
added the ValidateTokenAsyc() and ReadToken() methods to all token handlers

dannybtsai:
This is to add the new ValidateTokenAsync() method to the TokenHandler class. The method provides a consistent entry point for clients to use across all token handlers to perform validation.


src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1648,5 +1649,10 @@ public override void WriteToken(XmlWriter writer, SecurityToken token)
         {
             throw new NotImplementedException();
         }
+
+        /// <inheritdoc/>
+        /// No implementation since this class has been deprecated.
+        ///
```
ciaozhang: redundant line

brentschmaltz: We can't deprecate this class and will need an implementation here.

dannybtsai: The implementation is now added.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -948,6 +949,14 @@ public virtual SecurityToken ReadToken(string token)
         /// <param name="validationParameters">A <see cref="TokenValidationParameters"/>  required for validation.</param>
         /// <returns>A <see cref="TokenValidationResult"/></returns>
         public virtual TokenValidationResult ValidateToken(string token, TokenValidationParameters validationParameters)
+        {
+            var validationResult = ValidateTokenAsync(token, validationParameters);
+            validationResult.Wait();
```
TimHannMSFT: >validationResult.Wait(); [](http://example.com/codeflow?start=12&length=24)

For my benefit can you explain why you went with Wait() here?

I don't have a lot of experience around best practices with mixing async and synchronous methods but some searching around makes this look like it might not be ideal.

[https://stackoverflow.com/questions/9343594/how-to-call-asynchronous-method-from-synchronous-method-in-c](https://stackoverflow.com/questions/9343594/how-to-call-asynchronous-method-from-synchronous-method-in-c)

For one thing, as I understand, this will wrap all exceptions into an AggregateException which I think would be a behavior change here, is that desired?

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
```diff
@@ -73,5 +74,26 @@ public int TokenLifetimeInMinutes
...
+        public virtual Task<TokenValidationResult> ValidateTokenAsync(string token, TokenValidationParameters validationParameters) => throw new NotImplementedException();
+
+        /// <summary>
+        /// Converts a string into an instance of <see cref="SecurityToken"/>.
+        /// </summary>
+        /// <param name="token">The string to be deserialized.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="token"/> is null or empty.</exception>
+        /// <exception cref="ArgumentException">'token.Length' is greater than <see cref="TokenHandler.MaximumTokenSizeInBytes"/>.</exception>
+        /// <returns>A <see cref="SecurityToken"/>.</returns>
+        public virtual SecurityToken ReadToken(string token) => throw new NotImplementedException();
```
TimHannMSFT, line 97: >ReadToken [](http://example.com/codeflow?start=37&length=9)

How do we think about adding methods that will throw be default in the "Breaking changes" paradigm? Obviously we have to give it some default behavior if we're adding something to the contract here and throwing an exception makes the most sense but do we face an issue if someone had 

A : TokenHandler (for the old version)

and now when they pass A in, it throws an exception in cases where it perhaps wouldn't have before? Or if someone inherits from JsonWebTokenHandler wouldn't this virtual method be the default now instead of the now override method in that class for ReadToken? Isn't that breaking? I'm probably missing something but would appreciate a run down :) #Resolved

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -195,6 +196,29 @@ public virtual SecurityToken CreateToken(SecurityTokenDescriptor tokenDescriptor
...
+                });
+            }
+            catch (Exception ex)
+            {
+                return Task.FromResult(new TokenValidationResult
+                {
+                    IsValid = false,
+                    Exception = ex
+                });
+            }
```
TimHannMSFT, line 219: This is outside the scope of this PR but I'd like to see us as a team get away from this pattern.

I may be misunderstanding something but I believe we moved to this return-value-houses-the-exception-no-throw model because there was a request to increase performance. We're using exceptions here as glorified goto statements which is not great IMO.

The only exceptions, as I understand, that we'd actually want to expose here are of type SecurityTokenException so it seems like we could at least only catch that. (Very unclear to me what we expect our callers to do if they get an exception of an unrelated-to-Wilson type). 

The above is the first problem, the second problem is that we've just moved the perf cost of exception handling down the stack into our library instead of in the partner libraries which I'm not certain is much of an improvement. We could instead decide to pass the relevant exceptions back through the stack, this would almost certainly be more performant as well as avoiding the awkward poor-man's goto nature of the current pattern.

TimHannMSFT (dismissed):
:clock1:

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
dannybtsai: 1. If someone has A : TokenHandler, there should not be any way for their existing code to call this method since this is new. Do you have any specific scenario in mind where this will throw an exception when it would not before?

2. If someone inherits from JsonWebTokenHandler and they don't provide their own implementation of ReadToken(), then the implementation from JsonWebTokenHandler  will be used so no exception will be thrown.

TimHannMSFT: I think you're right here, I'd gotten mixed up on how the inheritance would work.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
dannybtsai: Yeah this makes sense. We should only catch the validation exceptions, like SecurityTokenValidationException, and set them in the returned result.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1648,5 +1649,29 @@ public override void WriteToken(XmlWriter writer, SecurityToken token)
...
+                var claimsPrincipal = ValidateToken(token, validationParameters, out var validatedToken);
+                return Task.FromResult(new TokenValidationResult
+                {
+                    SecurityToken = validatedToken,
+                    ClaimsIdentity = claimsPrincipal?.Identity as ClaimsIdentity,
+                    IsValid = true,
+                });
+            }
+            // only catch and return the SecurityTokenException via the TokenValidationResult, all other types of exceptions should be handled by the caller, ex., ArgumentException, ArgumentNullException...
+            catch (SecurityTokenException ste)
```
TimHannMSFT: >catch [](http://example.com/codeflow?start=12&length=5)

Thanks Danny, this LGTM.

One concern I have (now that I actually got what I asked for) is the consistency of the devEx here. These are new methods so I don't think we'd break anyone directly but anyone taking a dependency on these might mistakenly think they operate on the same old security token model where the validate methods throw an exception in the failure condition.

They might even write some bad tests (like passing null token) and confirm that this is how it works without passing a token with , e.g. a bad aud or iss. If they do that we might be in a similar case what we faced last year where folks don't check the return value and mistakenly don't actually validate tokens.

While I agree with the direction generally for the reasons in my other comment, I'd like to get Brent's (or someone with more team history) eyes on this before approving and then pushing out a Wilson release that locks us in here.

TimHannMSFT (dismissed):
:shipit:

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
brentschmaltz: @TimHannMSFT if we assume that any application using our SDK is catching exceptions (as they should), we should have a discussion about changing our model.

TimHannMSFT: Thanks Danny I had another look at this after our discussion. This approach still seems good to me. I think in the future, we can even take a look at refactoring how JsonWebTokenHandler::ValidateSignature bubbles up exceptions. We throw from that level only to catch here which I think we can eventually approve but let's save that for later and not delay more here, nothing prevents us from addressing that later.

I would recommend adding a remarks explaining the exception behavior of the new method though, explaining it won't ever throw SecurityTokenExceptions and the caller is responsible for checking the response but that it can throw input validation exceptions etc.

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
```diff
@@ -73,5 +74,26 @@ public int TokenLifetimeInMinutes
...
         }
+
+        #region methods
+
+        /// <summary>
+        /// Validates a token.
+        /// </summary>
+        /// <param name="token">The token to be validated.</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <returns>A <see cref="TokenValidationResult"/></returns>
```
TimHannMSFT, line 87: >/// [](http://example.com/codeflow?start=8&length=3)

to clarify, adding here for remarks or exceptions thrown would be helpful for callers I think. Good practice in general for public methods IMO but especially helpful in this case.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
dannybtsai: Added comments on the exceptions that can be thrown and the properties a caller should check to verify the validation result.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
dannybtsai: All token handlers ValidateTokenAsync() now catch the general exception to be consistent as discussed.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -195,6 +196,40 @@ public virtual SecurityToken CreateToken(SecurityTokenDescriptor tokenDescriptor
...
+        /// Validates a token.
+        /// ArgumentNullException (parameters are null or empty) and ArgumentException (token length exceeds the maximum size) will be thrown on invalid parameters.
+        /// On a validation failure, no exception will be thrown; instead, the exception will be set in the returned TokenValidationResult.Exception property.
+        /// Callers should always catch and handle invalid argument exceptions, and check the TokenValidationResult.IsValid property to verify the validity of the result.
+        /// </summary>
+        /// <param name="token">The token to be validated.</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="token"/> is null or empty.</exception>
+        /// <exception cref="ArgumentNullException"><paramref name="validationParameters"/> is null or empty.</exception>
+        /// <exception cref="ArgumentException">'token.Length' is greater than <see cref="TokenHandler.MaximumTokenSizeInBytes"/>.</exception>
```
TimHannMSFT: this needs to be updated.

TimHannMSFT: can probably just go back to the inheritdoc header instead.

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
```diff
@@ -73,5 +74,29 @@ public int TokenLifetimeInMinutes
...
+
+        #region methods
+
+        /// <summary>
+        /// Validates a token.
+        /// </summary>
+        /// <param name="token">The token to be validated.</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="token"/> is null or empty.</exception>
+        /// <exception cref="ArgumentNullException"><paramref name="validationParameters"/> is null or empty.</exception>
```
TimHannMSFT: remove here too, right?

src/Microsoft.IdentityModel.Tokens.Saml/GlobalSuppressions.cs
```diff
@@ -28,4 +28,5 @@
...
 [assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "Breaking change", Scope = "member", Target = "~P:Microsoft.IdentityModel.Tokens.Saml.SamlAssertion.CanonicalString")]
 [assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "Breaking change", Scope = "member", Target = "~P:Microsoft.IdentityModel.Tokens.Saml2.Saml2Assertion.CanonicalString")]
-
+[assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "Exception is returned as string", Scope = "member", Target = "~M:Microsoft.IdentityModel.Tokens.Saml.SamlSecurityTokenHandler.ValidateTokenAsync(System.String,Microsoft.IdentityModel.Tokens.TokenValidationParameters)~System.Threading.Tasks.Task{Microsoft.IdentityModel.Tokens.TokenValidationResult}")]
```
TimHannMSFT: >Exception is returned as string [](http://example.com/codeflow?start=101&length=31)

consider something along the lines of: "Catching all exceptions to match old model, in new models, we should try to avoid this behavior"

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -1174,6 +1175,40 @@ public override ClaimsPrincipal ValidateToken(XmlReader reader, TokenValidationP
...
+        /// Validates a token.
+        /// ArgumentNullException (parameters are null or empty) and ArgumentException (token length exceeds the maximum size) will be thrown on invalid parameters.
+        /// On a validation failure, no exception will be thrown; instead, the exception will be set in the returned TokenValidationResult.Exception property.
+        /// Callers should always catch and handle invalid argument exceptions, and check the TokenValidationResult.IsValid property to verify the validity of the result.
+        /// </summary>
+        /// <param name="token">The token to be validated.</param>
+        /// <param name="validationParameters">A <see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="token"/> is null or empty.</exception>
+        /// <exception cref="ArgumentNullException"><paramref name="validationParameters"/> is null or empty.</exception>
+        /// <exception cref="ArgumentException">'token.Length' is greater than <see cref="TokenHandler.MaximumTokenSizeInBytes"/>.</exception>
```
TimHannMSFT: update

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
dannybtsai: The comments have been updated.

TimHannMSFT (approved):
:shipit:

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -195,6 +196,40 @@ public virtual SecurityToken CreateToken(SecurityTokenDescriptor tokenDescriptor
...
+        /// <returns>A <see cref="TokenValidationResult"/></returns>
+        public override Task<TokenValidationResult> ValidateTokenAsync(string token, TokenValidationParameters validationParameters)
+        {
+            try
+            {
+                var claimsPrincipal = ValidateToken(token, validationParameters, out var validatedToken);
+                return Task.FromResult(new TokenValidationResult
+                {
+                    SecurityToken = validatedToken,
+                    ClaimsIdentity = claimsPrincipal?.Identity as ClaimsIdentity,
```
brentschmaltz: We need to think about how to deal with ValidateToken returns a ClaimsPrincipal that has multiple identities.

brentschmaltz (approved):

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
dannybtsai: It is now always returning the first identity as discussed.

PR #1813, opened 2022-02-23
Create thin Abstractions package and add Telemetry interface

TimHannMSFT:
* Creates a new lightweight package meant only for interfaces, abstract classes and null implementations
* Created ITelemetryClient interface
* Created NullTelemetryClient implmentation which defaults to no-ops
* To initially be used in SAL and then with MSAL and MISE.


RojaEnnam:
LGTM

---
In reply to: [1055775882](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1813#issuecomment-1055775882) [](http://example.com/codeflow?ancestors=1055775882)

---
In reply to: [1055775882](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1813#issuecomment-1055775882) [](http://example.com/codeflow?ancestors=1055775882)

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
```diff
@@ -0,0 +1,87 @@
...
+{
+    /// <summary>
+    /// Interface for Telemetry tracking.
+    /// </summary>
+    internal interface ITelemetryClient
+    {
+        /// <summary>
+        /// Gets or sets the application or client ID that telemetry is being sent for.
+        /// </summary>
+        Guid ClientId { get; set; }
```
dannybtsai: Is it possible that we might want to add more properties like ClientName/ClientDisplayName or EventStatus,...etc?
interface will be hard to change (vs abstract class that can have more methods and not breaking the contract) if this becomes public. But for this I think internal interface makes sense,

sruke: Agree with Danny that an interface will be hard to change once it becomes public. You could also consider having a struct/another class (TelemetryContext) to store all the properties of ITelemetryClient. That class would be easy to extend and the interface would remain unchanged.

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <summary>
+        /// Checks to see if telemetry is enabled all up.
+        /// </summary>
+        /// <returns>
+        /// Returns <see langword="true"/> if telemetry should be sent; <see langword="false"/> otherwise.
+        /// </returns>
+        /// <remarks>
+        /// This check should be used to gate any resource intensive operations to generate telemetry as well.
+        /// </remarks>
+        bool IsEnabled();
```
sruke, line 37: Would it be possible to have only one method (IsEnabled(eventName)) to check if telemetry is enabled? #Resolved

TimHannMSFT: It's certainly possible. My thinking here is that it's trivial to implement the additional method and it might have some good use cases if there is general telemetry overhead that could be avoided with the general method.

dannybtsai (approved):

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
brentschmaltz: I also agree, why an interface?
Although this is internal, but will SAL / MISE use this?

brentschmaltz: Is this always a GUID or would a string work here? #Resolved

Wilson.sln
```diff
@@ -97,6 +97,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.IdentityModel.Val
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.IdentityModel.SampleTests", "test\Microsoft.IdentityModel.SampleTests\Microsoft.IdentityModel.SampleTests.csproj", "{578FDF8F-6568-448A-AB93-D94269593932}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.IdentityModel.Abstractions", "src\Microsoft.IdentityModel.Abstractions\Microsoft.IdentityModel.Abstractions.csproj", "{8057C69A-3D1E-46A3-86E4-E6B26249DD25}"
```
brentschmaltz, line 100: Are there other 'abstractions' that are in the works?
Wondering why we need a new assembly. #Resolved

bgavrilMS, line 100: I noticed that you added the `strongNameBypass` reg settings. I don't think these are needed anymore. We should just strong name all our assemblies with a key that we commit to the repo. 

You can do this by adding a `Directory.Build.props` file properties in the root directory: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/Directory.Build.props#L4
and checking in this key: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/build/MSAL.snk #Resolved

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// Data used to manage access to administrative roles or sensitive functions.
+        /// </summary>
+        AccessControlData = 0,
+
+        /// <summary>
+        ///This is the data, information and code that C1 admins, C1 users, and C2 users (if present) provide to, transfers in, stores in or
+        /// process in the Microsoft online service or product.
+        /// </summary>
+        /// <remarks>The term C1 refers to Microsoft’s customer.The term C2 refers to the customer’s customer.</remarks>
+        CustomerContent = 10,
```
bgavrilMS: I assume these were copied from MISE / Office ? 

Does a piece of info fall into 1 and only 1 category from this list? Or can the user somehow specify that they want to 2 categories from this list? Would a flags enum be more appropriate? #Resolved

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// </summary>
+    /// <remarks>
+    /// See http://aka.ms/AssetClassificationStandard for more details about the asset classification standard.
+    /// </remarks>
+    public enum DataClassification
+    {
+        /// <summary>
+        /// Data used to manage access to administrative roles or sensitive functions.
+        /// </summary>
+        AccessControlData = 0,
```
bgavrilMS: From https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum:

```
✔️ DO provide a value of zero on simple enums.

Consider calling the value something like "None." If such a value is not appropriate for this particular enum, the most common default value for the enum should be assigned the underlying value of zero.
``` #Resolved

File: src/Microsoft.IdentityModel.Abstractions/TelemetryClient.cs [hunk 6]
```diff
@@ -0,0 +1,82 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.IdentityModel.Abstractions
+{
+    /// <summary>
+    /// Interface for Telemetry tracking.
+    /// </summary>
+    public abstract class TelemetryClient
```
bgavrilMS: Looks like everything except the ClientId is abstract. Should this be an interface instead? #Resolved

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
bgavrilMS: Also, from the design guidelines for Flags enum: 

```
Designing Flag Enums

✔️ CONSIDER providing special enum values for commonly used combinations of flags.

```

So maybe we could have an `AnyPii` value for this enum, as most of our customers are not used to this kind of classification.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,197 @@
...
+                    }
+                }
+                return currentBag;
+            }
+        }
+
+        /// <summary>
+        /// Gets the <see cref="DataClassification"/> mapping to keys in <see cref="Properties"/>.
+        /// </summary>
+        public virtual Dictionary<string, DataClassification> PropertyDataClassification
```
bgavrilMS: When would I override all these virtual methods? Would it make sense to keep them as normal method (non-virtual) and add the virtual property later (as a non-breaking change) ? #Resolved

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,197 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.IdentityModel.Abstractions
+{
+    /// <summary>
+    /// Details of the telemetry event.
+    /// </summary>
+    public abstract class EventDetails
```
bgavrilMS: It would help to have a console app or a unit test that shows an implementation of system. #Resolved

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// Updates the state of the property data classification.
+        /// </summary>
+        /// <param name="key">Key of the property.</param>
+        /// <param name="dataClassification">DataClassification for the property.</param>
+        internal void AddDataClassificationIfNecessary(string key, DataClassification dataClassification)
+        {
+            if (dataClassification != DataClassification.SystemMetadata)
+                DataClassificationMapping[key] = dataClassification;
+            else
+                DataClassificationMapping.Remove(key);
```
bgavrilMS: I believe this is not thread safe. You can either use ConcurrentDictionary operations or a lock (since this is pretty much a corner case?) #Resolved

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,197 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.IdentityModel.Abstractions
+{
+    /// <summary>
+    /// Details of the telemetry event.
+    /// </summary>
+    public abstract class EventDetails
+    {
+        internal Dictionary<string, DataClassification> DataClassificationMapping = new Dictionary<string, DataClassification>();
```
bgavrilMS: @pmaytak - for an opinion about thread safety. #Resolved

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
brentschmaltz: @TimHannMSFT I see what @bgavrilMS  is saying, can data be classified in 2 categories?

How will users extend to add their own categories?

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
brentschmaltz: @TimHannMSFT what is this used for?

src/Microsoft.IdentityModel.Abstractions/TelemetryClient.cs
brentschmaltz: @bgavrilMS as @dannybtsai pointed out, an interface is hard to change.

TimHannMSFT: Generally agree (see first iteration :)) will close with team this week

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: will add comments to this effect, clients can override for thread safety, don't think intra event safety will be a huge ask, can always improve later imo.

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
TimHannMSFT: I envisioned it as listing the highest exposed classification. e.g. if both PII and OII then the item is PII

bgavrilMS: Ack, so this is a not a flags enum, i.e. app developers can't pick and choose several options, just one. Kind of like logging verbosities.

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
TimHannMSFT: I see 0 as highest PII as the right model here, makes it privacy strict by default. If someone doesn't think about it and default is NoPII then we may have a privacy breach by carelessness, this defaults to more obfuscated metrics by default which is far better IMO.

Wilson.sln
GeoK: IdentityModel also has 'SignAssembly' as 'true', and uses a public key (AssemblyOriginatorKeyFile). The difference seem to be in DelaySign (true in IdentityModel, false in MSAL).

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: might do so to be thread-safe, or otherwise extend the event details.

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    /// <summary>
+    /// Interface for Telemetry tracking.
+    /// </summary>
+    public interface ITelemetryClient
+    {
+        /// <summary>
+        /// Gets or sets the application or client ID that telemetry is being sent for.
+        /// </summary>
+        Guid ClientId { get; set; }
```
TimHannMSFT: perhaps this should be string #Resolved

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        /// <summary>
+        /// Name of the telemetry event, should be unique between events.
+        /// </summary>
+        public abstract string Name { get; set; }
+
+        /// <summary>
+        /// Properties which describe the event.
+        /// </summary>
+        public virtual Dictionary<string, object> Properties
```
TimHannMSFT: IDictionary #Resolved

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <param name="boolProperties">Key value pair of bools to long with the event.</param>
+        /// <param name="DateTimeProperties">Key value pair of DateTimes to long with the event.</param>
+        /// <param name="doubleProperties">Key value pair of doubles to long with the event.</param>
+        /// <param name="guidProperties">Key value pair of Guids to long with the event.</param>
+        /// <param name="dataClassificationMapping">
+        /// Mapping between keys across all property dictionaries to data classification.
+        /// </param>
+        void TrackEvent(
+            string eventName,
+            Dictionary<string, string> stringProperties = null,
```
TimHannMSFT: all instances should be idicionary #Resolved

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,201 @@
...
+        {
+            get
+            {
+                var currentBag = new Dictionary<string, object>();
```
brentschmaltz: @TimHannMSFT how often will this get called?
These loops will take some time.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,201 @@
...
+                    }
+                }
+                return currentBag;
+            }
+        }
+
+        /// <summary>
+        /// Gets the <see cref="DataClassification"/> mapping to keys in <see cref="Properties"/>.
+        /// </summary>
+        public virtual Dictionary<string, DataClassification> PropertyDataClassification
```
brentschmaltz: Should always return IDictionary #Resolved

brentschmaltz (changes requested):

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// <summary>
+        /// Sets a property on the event details.
+        /// </summary>
+        /// <param name="key">Property key.</param>
+        /// <param name="value">Property value.</param>
+        /// <param name="dataClassification">Optional parameter which indicates the classification of the data.</param>
+        public virtual void SetProperty(
+            string key,
+            bool value,
+            DataClassification dataClassification = DataClassification.SystemMetadata)
```
brentschmaltz: Why have the default DataClassification.SystemMetadata?
Should we not make users consciously set the classification?

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
```diff
@@ -0,0 +1,82 @@
...
+        Guid ClientId { get; set; }
+
+        /// <summary>
+        /// Perform any necessary bootstrapping for the telemetry client.
+        /// </summary>
+        /// <remarks>
+        /// The expectation is that this should only be called once for the lifetime of the object however the
+        /// implementation should be idempotent.
+        /// </remarks>
+        void Initialize();
```
brentschmaltz, line 26: Always a good idea to use modifiers such as 'internal, public, ...' #Resolved

Wilson.sln
TimHannMSFT: Sruthi's work is an immediate example. Claims mapping may be another. I general we want an easy answer for the question "We want MSAL.net and Wilson to both use this abstraction, where should it live?"

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: we can, seems like worse DevEx to me. Telemetry data should basically always be metadata so we're forcing folks to always include boiler-plate stuff if we force them to set it explicitly.

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
TimHannMSFT: for <C#8.0, you can't set access modifiers on interface methods. Is that what the comment was talking about or something else?

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
TimHannMSFT: was only internal in the first iteration, now I'll be making it public since we actually want public constructors to pass this type.

IMO interfaces are cleaner and allow folks to better reuse their existing code. I agree they have different challenges with extensibility. IMO I partly mitigated this with EventDetails and it can be further mitigated by interface inheritance. I don't feel super strongly though. I'd rather get it in as an abstract class than spin wheels here. I've went interface->abstract class->interface already. Seems like the team is split. We don't seem to have a team wide guideline so I went with what I ultimately think is the cleaner design. Open to having my mind changed or just changing back to an abstract class to keep moving 😂

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: This also follows the Aria SDK programming model.

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
TimHannMSFT: I don't believe it's a design goal to allow for extensibility here. It doesn't seem very limiting to say a dimension will only be classified by one of these. Can you give an example where one should have 2 classifications? Seems like that's generally an argument for 2 dimensions...

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: Not sure I follow the Q. It's used to capture the details of a telemetry event. It's an extensibility point to hopefully capture any embellishments that are needed for events going forward. E.g. if we decide for some reason that public certs should be able to be logged in the telemetry pipe, EventDetails could be extended and a default behavior could be added in that case. Does that answer your Q?

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: @pmaytak feel free to reopen if you'd like to comment. Marking as resolved for now.

src/Microsoft.IdentityModel.Abstractions/TelemetryClient.cs
TimHannMSFT: Changed back to interface

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: Depends on the implementation of ITelemetryClient really. For the version I created, it's just called once. The other option would be to store everything in a dictionary expose consumers to it directly. Main downside there is if we keep it object someone could put junk in that would break the ItelemetryClient so that seems worse dev ex. If someone was super concerned with perf (maybe this will even be us) they could take that usability tradeoff and override the implementation.

In practice most events will only have a half dozen or less dimensions so this should be relatively cheap. I certainly think you're right to be concerned when you see this though. Open to other ideas.

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <remarks>
+        /// This check should be used to gate any resource intensive operations to generate telemetry as well.
+        /// </remarks>
+        bool IsEnabled(string eventName);
+
+        /// <summary>
+        /// Tracks an instance of a named event.
+        /// </summary>
+        /// <param name="eventDetails">Details of the event to track.</param>
+        void TrackEvent(
```
TimHannMSFT, line 56: Perhaps this should expose a TrackEventAsync As well or paerhaps there should be an ITelemetryClientAsync instead

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
dannybtsai: If the expectation is for users to extend the TelemetryClient then maybe abstract class is a good choice. Interface should be enough if this should cover common functionality. Starting from C# 8, interface seems to be looking more like abstract class "Beginning with C# 8.0, an interface may define a default implementation for members. It may also define static members in order to provide a single implementation for common functionality."

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,201 @@
...
+                            break;
+                        case DictionaryTypeEnum.Bool:
+                            currentBag.Add(kvp.Key, BoolPropertyValues[kvp.Key]);
+                            break;
+                        case DictionaryTypeEnum.DateTime:
+                            currentBag.Add(kvp.Key, DateTimePropertyValues[kvp.Key]);
+                            break;
+                        case DictionaryTypeEnum.Guid:
+                            currentBag.Add(kvp.Key, GuidPropertyValues[kvp.Key]);
+                            break;
```
dannybtsai: Dictionary.Add() can throw exceptions when key is null or exists already.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// </summary>
+        /// <param name="key">Property key.</param>
+        /// <param name="value">Property value.</param>
+        /// <param name="dataClassification">Optional parameter which indicates the classification of the data.</param>
+        public virtual void SetProperty(
+            string key,
+            string value,
+            DataClassification dataClassification = DataClassification.SystemMetadata)
+        {
+            PropertyValues[key] = value;
```
dannybtsai: >PropertyValues[key] = value; [](http://example.com/codeflow?start=12&length=28)

ArgumentNullException will be thrown if key is null.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: This is looping over PropertyKeys so at this point we already know for sure keys are unique and non-null.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,225 @@
...
+            PropertiesKeys[key] = DictionaryTypeEnum.Guid;
+            AddDataClassificationIfNecessary(key, dataClassification);
+        }
+
+        /// <summary>
+        /// Updates the state of the property data classification.
+        /// </summary>
+        /// <param name="key">Key of the property.</param>
+        /// <param name="dataClassification">DataClassification for the property.</param>
+        internal void AddDataClassificationIfNecessary(string key, DataClassification dataClassification)
```
dannybtsai: >AddDataClassificationIfNecessary [](http://example.com/codeflow?start=22&length=32)

I am not sure if the name reflects the logic.

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
```diff
@@ -0,0 +1,82 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.IdentityModel.Abstractions
+{
+    /// <summary>
+    /// Interface for Telemetry tracking.
+    /// </summary>
+    public interface ITelemetryClient
```
dannybtsai, line 12: Are there tests?

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
dannybtsai: @TimHannMSFT, I think some unit tests should help understand the details.

brentschmaltz: @TimHannMSFT that was exactly the answer I was looking for.
If i had phrased the question" "Do you have an example of how to use this class", that would have expressed my intent better.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,225 @@
...
+namespace Microsoft.IdentityModel.Abstractions
+{
+    /// <summary>
+    /// Details of the telemetry event.
+    /// </summary>
+    /// <remarks>
+    /// This implementation is not meant to be thread-safe. This implementation would either need to be overridden or
+    /// usage should not be concurrently operated on.
+    /// </remarks>
+    public abstract class EventDetails
```
brentschmaltz: Could we name this TelemetryEventDetails?

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
```diff
@@ -0,0 +1,82 @@
+﻿// Copyright(c) Microsoft Corporation.All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.IdentityModel.Abstractions
+{
+    /// <summary>
+    /// Data classification categories defined by the asset classification standard.
+    /// </summary>
+    /// <remarks>
+    /// See http://aka.ms/AssetClassificationStandard for more details about the asset classification standard.
```
brentschmaltz: http://aka.ms/AssetClassificationStandard should be 'https', the link seems to have moved should it now be: https://microsoft.sharepoint.com/sites/privacy/SitePages/Taxonomy-and-Use-Framework.aspx

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// </remarks>
+    public enum DataClassification
+    {
+        /// <summary>
+        /// Data used to manage access to administrative roles or sensitive functions.
+        /// </summary>
+        AccessControlData = 0,
+
+        /// <summary>
+        ///This is the data, information and code that C1 admins, C1 users, and C2 users (if present) provide to, transfers in, stores in or
```
brentschmaltz: space missing "///This" -> "/// This"

should " ... , transfers in ..." be "..., transfers to ..."

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// </summary>
+        /// <remarks>The term C1 refers to Microsoft’s customer.The term C2 refers to the customer’s customer.</remarks>
+        CustomerContent = 10,
+
+        /// <summary>
+        /// End User Identifiable Information.
+        /// Data that directly identifies or could be used to identify the authenticated C1 or C2 user of a Microsoft service. EUII does not contain C1 admin data.
+        /// EUII does not extend to other personal information found in Customer Content (e.g. contact lists are Customer Content), contact lists are Customer Content). 
+        /// </summary>
+        /// <remarks>The term C1 refers to Microsoft’s customer.The term C2 refers to the customer’s customer.</remarks>
```
brentschmaltz: not sure how this remark applies.

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// Data that directly identifies or could be used to identify the authenticated C1 or C2 user of a Microsoft service. EUII does not contain C1 admin data.
+        /// EUII does not extend to other personal information found in Customer Content (e.g. contact lists are Customer Content), contact lists are Customer Content). 
+        /// </summary>
+        /// <remarks>The term C1 refers to Microsoft’s customer.The term C2 refers to the customer’s customer.</remarks> 
+        EUII = 20,
+
+        /// <summary>
+        /// All data provided to Microsoft and generated by Microsoft as part of Support activities.
+        /// </summary>
+        /// <remarks>The term C1 refers to Microsoft’s customer.The term C2 refers to the customer’s customer.</remarks>
```
brentschmaltz: not sure how this remark applies.

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <summary>
+        /// All data provided to Microsoft and generated by Microsoft as part of Support activities.
+        /// </summary>
+        /// <remarks>The term C1 refers to Microsoft’s customer.The term C2 refers to the customer’s customer.</remarks>  
+        SupportData = 30,
+
+        /// <summary>
+        /// Contact and billing/purchase/payment/license information for the C1 enterprise,  including the C1 admin and any C1 subdelegated admins.
+        /// </summary>
+        /// <remarks>The term C1 refers to Microsoft’s customer.The term C2 refers to the customer’s customer.</remarks>
```
brentschmaltz: do we need to specify 'C2' here?

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <summary>
+        /// Contact and billing/purchase/payment/license information for the C1 enterprise,  including the C1 admin and any C1 subdelegated admins.
+        /// </summary>
+        /// <remarks>The term C1 refers to Microsoft’s customer.The term C2 refers to the customer’s customer.</remarks>  
+        AccountData = 40,
+
+        /// <summary>
+        /// Publicly available personal information that Microsoft obtains from external sources. Public Personal Data is
+        /// not <see cref="CustomerContent"/>, <see cref="EUII"/>, or <see cref="EUPI"/> since the data was not input.
+        /// </summary>
```
brentschmaltz: Can we include an example?

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
```diff
@@ -0,0 +1,82 @@
...
+        PublicPersonalData = 50,
+
+        /// <summary>
+        /// End User Pseudonymous Identifiers.
+        /// An identifier created by Microsoft tied to the C1 or C2 user of a Microsoft service. When EUPI is
+        /// combined with other information (such as a mapping table), it identifies the end user. 
+        /// </summary>
+        /// <remarks>The term C1 refers to Microsoft’s customer.The term C2 refers to the customer’s customer.</remarks>
```
brentschmaltz: Example would be helpful.

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    /// <summary>
+    /// Data classification categories defined by the asset classification standard.
+    /// </summary>
+    /// <remarks>
+    /// See http://aka.ms/AssetClassificationStandard for more details about the asset classification standard.
+    /// This is meant to identify the most consequential classification level (e.g. if an entry contains public data
+    /// and EUII, it should be classified as EUII).
+    /// </remarks>
+    public enum DataClassification
```
brentschmaltz: Should we name this MicrosoftDataClassification?

It would be helpful if there were examples in the comments.

brentschmaltz: It would be helpful if the values were in lexicographic order.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,225 @@
...
+    {
+        internal Dictionary<string, DataClassification> DataClassificationMapping = new Dictionary<string, DataClassification>();
+        internal Dictionary<string, string> PropertyValues = new Dictionary<string, string>();
```
brentschmaltz: Perhaps add a prefix and name" StringPropertyValues", just to be consistent.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
brentschmaltz: Do you have a link to Aria SDK?
I am concerned that data will not be classified correctly and may inadvertently be misclassified.

TimHannMSFT: They seemed to have moved their repo so my old link is dead, I cloned it a while back though so I can show it to you in our meeting later. I'm also following up to get an updated link.

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
TimHannMSFT: What sort of interface tests would you like to see? There will certainly be tests for the implementation but there's no code here.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: Unit tests for this class are probably worthwhile. I'll add those.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
```diff
@@ -0,0 +1,225 @@
...
+            if (dataClassification != DataClassification.SystemMetadata)
+                DataClassificationMapping[key] = dataClassification;
+            else
+                DataClassificationMapping.Remove(key);
+        }
+
+        /// <summary>
+        /// Internal enum to help differentiate acceptable dimension types.
+        /// </summary>
+        internal enum DictionaryTypeEnum
```
dannybtsai: >DictionaryTypeEnum [](http://example.com/codeflow?start=22&length=18)

This limits the types of values we support. Is it possible that some users might have events associated with user defined types?
It seems this can be simplified to just one Dictionary using Tuple:

internal Dictionary<string, Tuple<object, DataClassification>> Properties = new Dictionary<string, Tuple<object, DataClassification>>();

And only one SetProperty() is needed, plus the support of user-defined types.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: Changed implementation to expose a readonly dictionary instead and avoided the looping.

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
TimHannMSFT: Can expose a different interface for that later if it becomes required.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: Added null checks

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
TimHannMSFT: Added some tests to ensure the contract isn't broken to add friction to breaking the public contract.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: Updated

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
TimHannMSFT: Updated to: https://aka.ms/EntOnlineSvcDataTaxonomy
Unsure if it's kosher to include a link one must authenticate to read? This will be akin to a dead link for those outside MSFT.

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
TimHannMSFT: updated

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
TimHannMSFT: added examples for each from original doc

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
TimHannMSFT: I think classification order makes most sense here. examples added.

src/Microsoft.IdentityModel.Abstractions/EventDetails.cs
TimHannMSFT: Changed direction here. Plus added a unit test to show an example extention.

src/Microsoft.IdentityModel.Abstractions/DataClassification.cs
TimHannMSFT: decided to go a different direction

src/Microsoft.IdentityModel.Abstractions/TelemetryEventDetails.cs
```diff
@@ -0,0 +1,128 @@
...
+    /// <remarks>
+    /// This implementation is not meant to be thread-safe. This implementation would either need to be overridden or
+    /// usage should not be concurrently operated on.
+    /// </remarks>
+    public abstract class TelemetryEventDetails
+    {
+        /// <summary>
+        /// The underlying properties making up the <see cref="TelemetryEventDetails"/>.
+        /// </summary>
+        protected internal Dictionary<string, object> PropertyValues { get; } = new Dictionary<string, object>();
```
brentschmaltz: This should return 'IDictionary' rather than 'Dictionary'.

src/Microsoft.IdentityModel.Abstractions/TelemetryEventDetails.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        /// <summary>
+        /// Name of the telemetry event, should be unique between events.
+        /// </summary>
+        public virtual string Name { get; set; }
+
+        /// <summary>
+        /// Properties which describe the event.
+        /// </summary>
+        public virtual IDictionary<string, object> Properties
```
brentschmaltz: If this returned IReadOnlyDictionary<string, object>, you could then just return PropertyValues without making a copy.
The runtime will mark the dictionary as readonly.

If you are wanting to return a snapshot, you will need to create a new Dictionary<string, object>(Properties) first
then return an IReadonlyDictionary.

src/Microsoft.IdentityModel.Abstractions/TelemetryEventDetails.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.IdentityModel.Abstractions
+{
+    /// <summary>
+    /// Details of the telemetry event.
+    /// </summary>
+    /// <remarks>
+    /// This implementation is not meant to be thread-safe. This implementation would either need to be overridden or
```
brentschmaltz, line 13: What would happen with multiple threads?

TimHannMSFT: This is just to acknowledge that it's based on an internal Dictionary which isn't thread safe
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2?redirectedfrom=MSDN&view=net-6.0#thread-safety

One could reimplement with concurrentDictionary if that was important but I don't think there would be many applications which needed concurrency for a single event (the concurrent events should usually be separated).

src/Microsoft.IdentityModel.Abstractions/TelemetryEventDetails.cs
TimHannMSFT: Good catch

src/Microsoft.IdentityModel.Abstractions/TelemetryEventDetails.cs
TimHannMSFT: Good catch

brentschmaltz (approved):

src/Microsoft.IdentityModel.Abstractions/TelemetryEventDetails.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            SetPropertyCore(key, value);
+        }
+
+        private void SetPropertyCore(
+            string key,
+            object value)
+        {
+            if (key == null)
+                throw new ArgumentNullException(nameof(key));
```
brentschmaltz, line 122: Will we be piping this through our Logger?
Should we use our LogHelper?

src/Microsoft.IdentityModel.Abstractions/ObservabilityConstants.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// Common class containing observability constants to be used as well known metric keys.
+    /// </summary>
+    public static class ObservabilityConstants
+    {
+        /// <summary>
+        /// String used for the name of the property indicating if the call was successful.
+        /// </summary>
+        public const string Succeeded = "Succeeded";
```
brentschmaltz, line 14: Should we have a constant for 'Failed'?

src/Microsoft.IdentityModel.Abstractions/TelemetryEventDetails.cs
TimHannMSFT: GoodQ: It's not related to the logger. Thought it may be injected in a similar way down into Wilson at least for now when we want to integrate telemetry for e.g. crypto operations.

src/Microsoft.IdentityModel.Abstractions/ObservabilityConstants.cs
TimHannMSFT: This is for a bool property, so true indicates success and false indicates failed. I've no objections to a failed getting added here however if there's a relevant scenario in the future though I suspect the bool should cover it (many teams have this model for metrics)

PR #1814, opened 2022-02-23
Introduced logger context

sruke:
Copied over logging properties from CallContext class to newly created LoggerContext class. 

Background -
Properties primarily used for logging are currently part of CallContext class belonging to Microsoft.IdentityModel.Tokens namespace. 
As part of the effort to adopt ILogger I'm also making some improvements to the existing logic. 
This PR's goal is to create a LoggerContext class where all logging properties will reside and CallContext will inherit LoggerContext. This way logging properties will reside in Microsoft.IdentityModel.Logging namespace.


src/Microsoft.IdentityModel.Logging/LoggerContext.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <param name="activityId"></param>
+        public LoggerContext(Guid activityId)
+        {
+            ActivityId = activityId;
+        }
+
+        /// <summary>
+        /// Gets or sets a string that helps with setting breakpoints when debugging.
+        /// </summary>
+        public string Id { get; set; } = string.Empty;
```
sruke: public string Id will replace an 'internal' property with the same name in SAL's S2SContext class. In the next SAL PR I'll make sure to remove the internal property to avoid compiler getting confused.

brentschmaltz (approved):

RojaEnnam (approved):

src/Microsoft.IdentityModel.Logging/LoggerContext.cs
GeoK: Why are we exposing this property if it's used only internally?

src/Microsoft.IdentityModel.Logging/LoggerContext.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        /// <summary>
+        /// The collection of logs associated with a request. Use <see cref="CaptureLogs"/> to control capture.
+        /// </summary>
+        public ICollection<string> Logs { get; private set; } = new Collection<string>();
+
+        /// <summary>
+        /// Gets or sets an <see cref="IDictionary{String, Object}"/> that enables custom extensibility scenarios.
+        /// </summary>
+        public IDictionary<string, object> PropertyBag { get; set; }
```
GeoK, line 55: This property technically belong more to Call than Logging context.

src/Microsoft.IdentityModel.Logging/LoggerContext.cs
```diff
@@ -0,0 +1,81 @@
...
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// An opaque context used to store work when working with authentication artifacts.
+    /// </summary>
+    public class LoggerContext
```
TimHannMSFT, line 13: >LoggerContext [](http://example.com/codeflow?start=17&length=13)

So is it the case that CallContext : LoggerContext instead of CallContext : ILoggerContext mostly from an extensibility standpoint? e.g. If LoggerContext ever changes it's nice to have a class so we don't have to make a whole new interface without it being breaking?

It seems like logically this is an interface, it's saying this is a contract we'd expect such objects to behave under, as opposed to saying CallContexts are a type of LoggingContext.

I have no strong objections but if that's the case we might want to add some comments explaining the rationale.

src/Microsoft.IdentityModel.Logging/LoggerContext.cs
```diff
@@ -0,0 +1,81 @@
...
+//------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// An opaque context used to store work when working with authentication artifacts.
```
TimHannMSFT: >An opaque context used to store work when working with authentication artifacts [](http://example.com/codeflow?start=8&length=79)

Update here. I don't think the "opaque context" is even true of the CallContext but may as well have this be more accurate on the new class.

src/Microsoft.IdentityModel.Logging/LoggerContext.cs
brentschmaltz: @GeoK It is internal in SAL, but we would like it to be public to aid in debugging.

jennyf19:
Will this be merged and does it have a corresponding issue? @sruke

src/Microsoft.IdentityModel.Logging/LoggerContext.cs
sruke: Extensibility is one reason. Another reason is this class will be used mostly internally. Right now, I don't see a reason why a customer would want to implement a LoggerContext interface and customize it. Brent wants to see a context object widely available in Wilson to aid with debugging.

src/Microsoft.IdentityModel.Logging/LoggerContext.cs
sruke: Brent's rationale for having PropertyBag under LoggerContext rather than CallContext is that Microsoft.IdentityModel.Logging namespace is more widely used compared to Microsoft.IdentityModel.Tokens which makes creating a LoggerContext object and passing it through the call stack easier.

sruke:
Rebase is not straightforward as its a very old PR. Will open a new PR.

PR #1820, opened 2022-03-05
log cert thumbprint

jennyf19:
This is related to https://github.com/AzureAD/microsoft-identity-web/issues/1646
Urgent ask from MyApps

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -49,7 +49,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// Regex that is used to figure out if a token is in JWS format.
         /// </summary>
         public static Regex RegexJws = new Regex(JwtConstants.JsonCompactSerializationRegex, RegexOptions.Compiled | RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(100));
-        
+
         /// <summary>
         /// Regex that is used to figure out if a token is in JWE format.
         /// </summary>
@@ -200,6 +200,9 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 {
                     Validators.ValidateAlgorithm(decryptionParameters.Enc, key, jwtToken, validationParameters);
                     decryptedTokenBytes = DecryptToken(cryptoProviderFactory, key, decryptionParameters);
+                    X509SecurityKey x509Key = key as X509SecurityKey;
+                    if (x509Key != null)
+                        LogHelper.LogInformation(TokenLogMessages.IDX10903, x509Key.Certificate?.Thumbprint);
                     decryptionSucceeded = true;
                     break;
                 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -143,6 +143,7 @@ internal static class LogMessages
         public const string IDX10618 = "IDX10618: Key unwrap failed using decryption Keys: '{0}'.\nExceptions caught:\n '{1}'.\ntoken: '{2}'.";
         public const string IDX10619 = "IDX10619: Decryption failed. Algorithm: '{0}'. Either the Encryption Algorithm: '{1}' or none of the Security Keys are supported by the CryptoProviderFactory.";
         public const string IDX10620 = "IDX10620: Unable to obtain a CryptoProviderFactory, both EncryptingCredentials.CryptoProviderFactory and EncryptingCredentials.Key.CrypoProviderFactory are null.";
+        public const string IDX10903 = "IDX10903: Decryption succeeded. With thumbprint: '{0}'.";
```
jmprieur:
```suggestion
        public const string IDX10903 = "IDX10903: Token decryption succeeded. With thumbprint: '{0}'.";
```

jmprieur (approved):
LGTM
Thanks @jennyf19

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -200,6 +200,9 @@ internal static string DecompressToken(byte[] tokenBytes, string algorithm)
                 {
                     Validators.ValidateAlgorithm(decryptionParameters.Enc, key, jwtToken, validationParameters);
                     decryptedTokenBytes = DecryptToken(cryptoProviderFactory, key, decryptionParameters);
+                    X509SecurityKey x509Key = key as X509SecurityKey;
+                    if (x509Key != null)
+                        LogHelper.LogInformation(TokenLogMessages.IDX10903, x509Key.Certificate?.Thumbprint);
```
henrik-me, line 205: Wondering if there is anything from the decryptionParameters we can log which will give similar information? or if we should log that regardless, just in case this is not an x509 cert/key. or perhaps from the cryptoprovider?

henrik-me: also considering if we should put all of it behind an Log.IsInformation check (right now it's just a typecast and a comparison, but it will be waste of cpu cycles in most cases as it's not recommended to run with informational logging (I believe).

henrik-me (approved):
:shipit:

henrik-me:
Thinking we should also log which certs are loaded and which are in the configuration

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
brentschmaltz: x509Key.Certificate cannot be null, no need for the null check.

brentschmaltz, line 205: Please add a new line after a clause such as 'if'

jennyf19:
Do you see any perf issues with this @brentschmaltz ?

brentschmaltz (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
jennyf19: fixed.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
jennyf19: fixed

Issue #1580, opened 2021-01-20
Returning internal JObjects in public APIs breaks updating to .net 5.0


leastprivilege:
...e.g. in the OpenIdConnect configuration when a custom element is a JSON object.

Now the consuming code get returns a JObject - but it's internal and cannot do anything with that. This is a blocker for updating to .NET 5...

```cs
var config = await _manager.GetConfigurationAsync(cancel);

// cast will fail since JObject returned is the internal version...                        
var mtls = config.AdditionalData[OidcConstants.Discovery.MtlsEndpointAliases] as JObject;

config.TokenEndpoint = mtls[OidcConstants.Discovery.TokenEndpoint].ToString();
```

brentschmaltz:
We are going to fix this in: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1516

Issue #1475, opened 2020-06-27
OpenIdConnectConfiguration not serializing properly


RogerioWagner:
The OpenIdConnectConfiguration is not serializing the field **jwks_uri** field, the bellow example show the incorrect field naming (**jwksUri**).

The package to reproduce the behavior is Microsoft.IdentityModel.Protocols.OpenIdConnect version 5.5.0.

The problem caused is that the authenticationservice could not ask the authority the public key set to validate a token.

```json
{
  "additionalData": {},
  "acrValuesSupported": [],
  "authorizationEndpoint": "https://localhost:4000/connect/authorization",
  "checkSessionIframe": null,
  "claimsSupported": [], "claimsLocalesSupported": [],
  "claimsParameterSupported": false,
  "claimTypesSupported": [],
  "displayValuesSupported": [],
  "endSessionEndpoint": null,
  "frontchannelLogoutSessionSupported": null,
  "frontchannelLogoutSupported": null,
  "grantTypesSupported": [],
  "httpLogoutSupported": false,
  "idTokenEncryptionAlgValuesSupported": [],
  "idTokenEncryptionEncValuesSupported": [],
  "idTokenSigningAlgValuesSupported": [],
  "issuer": "https://localhost:4000",
  "jwksUri": "https://localhost:4000/.well-known/openid-configuration/jwks",
  "jsonWebKeySet": {
    "additionalData": {},
    "keys": [
      {
        "additionalData": {},
        "alg": null,
        "crv": null,
        "d": null,
        "dp": null,
        "dq": null,
        "e": "AQAB",
        "k": null,
        "keyId": "319D93355237260601A940A962991084B835681A",
        "keyOps": [],
        "kid": "319D93355237260601A940A962991084B835681A",
        "kty": "RSA",
[...]

Issue #1465, opened 2020-06-23
JwtSecurityToken.EncodedPayload is having problems with my Deserialized JSON to plain object


ghstahl:
This is a "It used to work" issue.  I have attached the code that shows the issue.
I want to be able to Deserialize a JSON into an object that I can add correctly to  a JwtPayload and get properly encoded by the JwtSecurityToken.

I can add a object to the payload.
```
var custom = new Custom{}
payload.Add("custom", custom);
```
I can add a dynamic to the payload.
```
dynamic ccDynamic = new
            {
                a="b",
                b=2,
                id_token_signing_alg_values_supported = new List<string>{
                        "RS256"
                },
                d = new { 
                  name  = "bob"
                }
            };
payload.Add("ccDynamic", ccDynamic);
```
I can add an expando to the payload.
```
var converter = new ExpandoObjectConverter();
dynamic expando = JsonConvert.DeserializeObject<ExpandoObject>(json, converter);
payload.Add("expando", expando);
```
And these get encoded just fine in the final JWT.

I used to be able to deserialize a json to a dynamic or object and add that.
```
   var json = JsonConvert.SerializeObject(custom);
    var jObj = JsonConvert.DeserializeObject(json);
    dynamic jDynamic = JsonConvert.DeserializeObject<dynamic>(json);
    payload.Add("object", jObj);
    payload.Add("jDynamic", jDynamic);
```
However those Deserialized object now turn into the following.
[...]

Issue #550, opened 2016-12-15
Stop transforming inbound/outbound claim types by default


leastprivilege:
I gave this feedback a couple of times - the JwtHandler is not the place for (automatic) claims transformation. This is either application logic or explicit opt-in behavior.

I want the exact claim types in the principal that got issued from my STS - and not MS proprietary ones.

brentschmaltz:
@leastprivilege I agree with you in principal. 
If we turn it off by default, we need to consider how ClaimsIdentity.Name and ClaimsPrincipal.IsInRole are affected. These apis use ClaimsIdentity.NameClaimType and ClaimsIdentity.RoleClaimType with are set to the mapped types.

A possible mitigation is:
clearing static JwtSecurityTokenHandler.DefaultInboundClaimTypeMap in global scope.

leastprivilege:
Why is this your concern? Anybody can set the RoleClaimType and NameClaimType when validating a token to match whatever they are expecting.

Both are legacy and shouldn't be used anymore anyways.

wrt statics - it's 2016.

brentschmaltz:
My concern is back-compat. ClaimsIdentity.Name, which is tied to NameClaimType, could be null. This has caused null-refs in code that expects a non-null value.

It almost 2017 :-).

leastprivilege:
Again - why is this your concern?

There might be JWTs without a "name" claim (or something that maps to it) at all. Actually they are pretty common.

The same issue would exist.

brockallen:
> My concern is back-compat

With that? This is a new(er) technology, and no where in the specs do you see WS-* claim types recommended. Stop treating the developers like they are incompetent.

If you must make these decisions for developers, then put a layer elsewhere in the app layer to give you SOAP-flavored claim types, but don't mandate this on the rest of people by default. By forcing this, you're shoehorning people into a legacy mode of claim types. Then you have more backwards compat problems down the road.

At some point you (and the developers consuming your stuff) need to evolve. This awful mindset of wanting to babysit your developers prohibits that.

brentschmaltz:
The back-compat scenario where an AuthorizationDecisionPoint (internal or external) functions differently when S.IM.Tokens.Jwt generates different claims by default, is the concern. A comprehensive set of tests may catch such issues. It is doubtful that all issues would be shown through testing.

brockallen:
[...]

PR #1823, opened 2022-03-14
Introducing a custom logging interface

### LINKED ISSUE #1846: [Feature Request] Introduce IIdentityLogger for a unified logging story across all Identity Libraries
Problem
Our libraries have different logging mechanisms. Wilson and SAL use System.Diagnostics.Tracing.EventSource. MISE and [Microsoft.Identity.Web](http://microsoft.identity.web/)  use Microsoft.Extensions.Logging.ILogger.

Solution
Adopt a uniform logging mechanism across [Microsoft.Identity.Web](http://microsoft.identity.web/) , MISE, SAL, Wilson, and MSAL so that by setting log level at a given level it applies to all these libraries.

Design Proposal is [here](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/4779?path=/Middleware/Wilson%20ILogger%20Adoption.md&_a=files)

sruke:
Introduces a logging mechanism in addition to EventSource logging.

Adds a logging interface. 

To use it, user of this library needs to implement the interface, create an instance and set it as follows:
LogHelper.Logger = myLoggerInstance;


src/Microsoft.IdentityModel.Logging/IIdentityLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+//------------------------------------------------------------------------------
+
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// Interface for Logging.
+    /// </summary>
+    public interface IIdentityLogger
```
bgavrilMS: There is no provision for turning PII logging on / off or for PII in general.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -371,5 +417,47 @@ public static object MarkAsNonPII(object arg)
...
+        /// <param name="level"><see cref="EventLevel"/></param>
+        /// <param name="innerException"><see cref="Exception"/></param>
+        /// <param name="message">The log message.</param>
+        /// <param name="args">An object array that contains zero or more objects to format.</param>
+        private static LogEntry WriteEntry(EventLevel level, Exception innerException, string message, params object[] args)
+        {
+            if (innerException != null)
+            {
+                // if PII is turned off and 'innerException' is a System exception only display the exception type
+                if (!IdentityModelEventSource.ShowPII && !LogHelper.IsCustomException(innerException))
```
bgavrilMS, line 437: This is tightly coupled to `IdentityModelEventSource`

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -37,6 +38,36 @@ namespace Microsoft.IdentityModel.Logging
     /// </summary>
     public class LogHelper
     {
+        /// <summary>
+        /// Gets or sets a logger to which logs will be written to.
+        /// </summary>
+        public static IIdentityLogger Logger { get; set; } = NullIdentityModelLogger.Instance;
```
brockallen, line 45: This being a static will cause problems down the road. Much like the JSON serializer and the claim type mapping setting.

src/Microsoft.IdentityModel.Logging/IIdentityLogger.cs
trwalke: Yes, we need some way of determining if Pii is enabled so the decision can be made if an expensive message containing Pii should be created or not.

src/Microsoft.IdentityModel.Logging.Abstractions/Microsoft.IdentityModel.Logging.Abstractions.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <AssemblyName>Microsoft.IdentityModel.Logging.Abstractions</AssemblyName>
```
TimHannMSFT: >Logging [](http://example.com/codeflow?start=42&length=7)

Can this be Microsoft.IdentityModel.Observability.Abstractions?

That way it will be a more natural home for other telemetry stuff as well

src/Microsoft.IdentityModel.Logging.Abstractions/LogEntry.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Message to be logged.
+        /// </summary>
+        public string Message { get; set; }
+
+        /// <summary>
+        /// A unique identifier for a request that can help with diagnostics across components.
+        /// </summary>
+        public string CorrelationId { get; set; }
```
TimHannMSFT: >CorrelationId [](http://example.com/codeflow?start=22&length=13)

Consider renaming to "ActivityId' or including a remark that notes this is referred to as "ActivityId" in other parts of the code (e.g. CallContext)

src/Microsoft.IdentityModel.Logging.Abstractions/IIdentityLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+using System.Diagnostics.Tracing;
```
TimHannMSFT: >System.Diagnostics.Tracing [](http://example.com/codeflow?start=6&length=26)

Is this dependency worth it just for the enum? Doesn't MISE implement it's own clone and then it's very easy for a consumer to map (we can include an example in the main lib) Would be a shame if this blocked someone's adoption.

TimHannMSFT:
> //------------------------------------------------------------------------------

Can we get away with the truncated version for new files?

---
Refers to: src/Microsoft.IdentityModel.Logging.Abstractions/IIdentityLogger.cs:1 in e31c6b7. [](commit_id = e31c6b7933caba465b2ca4df7c012d8d2640ecfe, deletion_comment = False)

src/Microsoft.IdentityModel.Logging/LogEntry.cs
```diff
@@ -0,0 +1,52 @@
...
+//------------------------------------------------------------------------------
+
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// Defines the structure of a log entry.
+    /// </summary>
+    public class LogEntry
```
TimHannMSFT: >LogEntry [](http://example.com/codeflow?start=17&length=8)

Seems like this might be the right place to surface the PII level. would this level also eventually reference a PII scrubber or is the idea that would be at another level?

src/Microsoft.IdentityModel.Logging.Abstractions/Microsoft.IdentityModel.Logging.Abstractions.csproj
TimHannMSFT: Talked offline. I believe there was another suggestion of MS.ID.Abstractions which could serve multiple purposes.

src/Microsoft.IdentityModel.Logging.Abstractions/IIdentityLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// Interface for Logging.
+    /// </summary>
+    public interface IIdentityLogger
+    {
+        /// <summary>
+        /// Checks to see if logging is enabled at given <paramref name="eventLevel"/>.
+        /// </summary>
+        /// <param name="eventLevel">Log level of an Event.</param>
+        bool IsEnabled(EventLevel eventLevel);
```
bgavrilMS: App developer typically provide a minimal log level. I propose we change to

```csharp
EventLevel LogVerbosity { get; }
```

src/Microsoft.IdentityModel.Logging.Abstractions/LogEntry.cs
bgavrilMS: Why does this need to be here? Can't SAL / MSAL / MISE etc. just prefix the log message with the correlation id?

src/Microsoft.IdentityModel.Logging.Abstractions/IIdentityLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+//------------------------------------------------------------------------------
+
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.IdentityModel.Logging.Abstractions
+{
+    /// <summary>
+    /// Interface for Logging.
+    /// </summary>
+    public interface IIdentityLogger
```
bgavrilMS: The implementer must define their level of PII they want to log. This should be done in this top-level interface.

bgavrilMS (changes requested):
We need to expose the concept of PII in our logs. The app developer needs to select the level of PII they are confortable with when they define the `IIdentityLogger` interface. We could use the simple "is PII" approach both SAL and MSAL have today, or we use the more complex approach MISE has. 

Here's a prototype showing some of my comments - with 1 PII level (best to review from bottom up or just put it in a console app and run it) https://gist.github.com/bgavrilMS/16d44dca6370588db0b95fc370e73557

Would be good to iterate on the design to understand how we can introduce MISE's PII levels.

src/Microsoft.IdentityModel.Logging/IIdentityLogger.cs
bgavrilMS: This is a good point about expensive messages, and a scenario that should be articulated. One way of doing it: https://gist.github.com/bgavrilMS/16d44dca6370588db0b95fc370e73557#file-iidentitylogger-cs-L32

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Could you expand on this or provide any relevant links on the issue?

sruke:
> We need to expose the concept of PII in our logs. The app developer needs to select the level of PII they are confortable with when they define the `IIdentityLogger` interface. We could use the simple "is PII" approach both SAL and MSAL have today, or we use the more complex approach MISE has.
> 
> Here's a prototype showing some of my comments - with 1 PII level (best to review from bottom up or just put it in a console app and run it) https://gist.github.com/bgavrilMS/16d44dca6370588db0b95fc370e73557
> 
> Would be good to iterate on the design to understand how we can introduce MISE's PII levels.

Can we have a separate LogScrubbing interface with a MinimumDataClassificationCategory that user can set to track the level of PII that needs to be scrubbed? All log arguments whose DataClassificationCategory is below the minimum level will be scrubbed.

src/Microsoft.IdentityModel.Logging.Abstractions/LogEntry.cs
sruke: The intent behind adding a separate property was that the correlation id can be parsed and could have its own column when logs are stored is an app like Application Insights.

src/Microsoft.IdentityModel.Logging.Abstractions/IIdentityLogger.cs
sruke: Would it be possible to keep the current method signature? ASP.Net Core's ILogger.IsEnabled() method was taken as reference when creating this as we write to their ILogger. 
Link - https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.logger-1.microsoft-extensions-logging-ilogger-isenabled?view=dotnet-plat-ext-6.0#microsoft-extensions-logging-logger-1-microsoft-extensions-logging-ilogger-isenabled(microsoft-extensions-logging-loglevel)

src/Microsoft.IdentityModel.Logging/LogEntry.cs
sruke: LogEntry.Message excepts a scrubbed message. 
IdentityModelEventSource.ShowPII will control if PII is turned on/off. LogHelper class needs to scrub the message before creating a log entry.
In the next version, implementation of ILogScrubber will contains a minimumDataClassificationLevel property that will decide how much scrubbing needs to be happen in the LogHelper class.

sruke:
> > //------------------------------------------------------------------------------
> 
> Can we get away with the truncated version for new files?
> 
> // Copyright (c) Microsoft Corporation. All rights reserved. // Licensed under the MIT License.
> 
> Refers to: src/Microsoft.IdentityModel.Logging.Abstractions/IIdentityLogger.cs:1 in [e31c6b7](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/e31c6b7933caba465b2ca4df7c012d8d2640ecfe). [](commit_id = [e31c6b7](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/e31c6b7933caba465b2ca4df7c012d8d2640ecfe), deletion_comment = False)

Being opensource maybe it needs the long version? Will wait for others to address this.

src/Microsoft.IdentityModel.Logging.Abstractions/IIdentityLogger.cs
sruke: If we were to remove the dependency on System.Diagnostics.Tracing.Event and introduce a custom enum (say LogSeverityLevel enum) then log level needs to be mapped from EventLevel -> LogSeverityLevel. Not sure if there could be a performance impact. Benchmarking may be worth it here.

jmprieur: If the new enum uses the same numeric values, then the mapping would be as simple as a double cast:
(LogSererityLevel)(int)eventLevel

src/Microsoft.IdentityModel.Logging.Abstractions/LogEntry.cs
sruke: I'll add a remark. Let me know if renaming the property to ActivityId would be more appropriate for consistency. This decision was made to remain consistent up the stack.

src/Microsoft.IdentityModel.Logging.Abstractions/IIdentityLogger.cs
sruke: This property can be added to a class that is implementing the IIdentityLogger interface and IsEnabled(currentEventLevel) could check currentEventLevel >= LogVerbosity.

ASP.Net Core's ILogger.IsEnabled(..) was used a reference when creating this method to keep the IIdentityLogger interface close to what ASP.Net Core has without explicitly taking a dependency on it.

Link - https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.logger-1.microsoft-extensions-logging-ilogger-isenabled?view=dotnet-plat-ext-6.0#microsoft-extensions-logging-logger-1-microsoft-extensions-logging-ilogger-isenabled(microsoft-extensions-logging-loglevel)

src/Microsoft.IdentityModel.Logging/LogHelper.cs
brockallen: It's hard to find specific issues for 8+ years of history. I think there are relevant, but unless you lived thru those issues it's hard to explain.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/202

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/550#issuecomment-267518199

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1475

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1465

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1580

src/Microsoft.IdentityModel.Logging.Abstractions/LogEntry.cs
bgavrilMS: Got it. This is indeed a feature some other folks are requesting as well, i.e. structured logging....

src/Microsoft.IdentityModel.Logging/LogHelper.cs
AndersAbel: As the author of the Sustainsys.Saml2 library I definitely see a case for intercepting logs from underlying MIM operations from the Saml2 library and correlate them with the log output from the Saml2 library. But I of course only want to affect MIM operations invoked from the Saml2 library, not any MIM operations invoked from other part of the calling application.

src/Microsoft.IdentityModel.Logging/IIdentityLogger.cs
sruke: True. Within the IIdentityLogger currently there is no provision for PII because Wilson exposes a static IdentityEventSource.ShowPII boolean that developer can set to control if PII will be logged or not.

sruke: [The consoleapp implementation ](https://gist.github.com/bgavrilMS/16d44dca6370588db0b95fc370e73557#file-iidentitylogger-cs-L32)
=> This looks very promising for MSAL's requirements. I'll check with the team if we want to introduce a PII property that is associated with IIdentityLogger interface in Wilson.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: True but being a static its available to the LogHelper class and FormatInvariant uses it to evaluate if arguments need to be scrubbed.
entry.Message = args != null ? FormatInvariant(message, args) : message;

src/Microsoft.IdentityModel.Abstractions/IIdentityLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+//------------------------------------------------------------------------------
+
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.IdentityModel.Abstractions
+{
+    /// <summary>
+    /// Interface for Logging.
+    /// </summary>
+    public interface IIdentityLogger
```
jennyf19, line 35: have you looked into leveraging the high performance logging provided by .NET core? We use it in Id Web.
[https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage?msclkid=7a0c0097b12c11ecbb637e8c4c4a3a86&view=aspnetcore-6.0](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage?msclkid=7a0c0097b12c11ecbb637e8c4c4a3a86&view=aspnetcore-6.0)

src/Microsoft.IdentityModel.Logging/IIdentityLogger.cs
brentschmaltz: @sruke @trwalke above we seem to have identified two issues:
1. add logic to delay creation of messages until after determining the message will be logged (log level is in effect).
2. controlling PII. We currently have the switch on IdentityEventSource.ShowPII, if IIdentityLogger is independent of IdentityEventSource, then we will need logic.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
brentschmaltz: @brockallen @sruke @AndersAbel the issues are semi-related to statics but seem mostly related to the general dislike of the decision to map claims and internal newton soft.

Our goal here is to provide a natural way for plugging in asp.net ILogger so that all logs are piped to the same listener.
We also want to expose this in such a way that layers, other than asp.net, can plug in so that other SDK's (we have some in the pipeline) will have a unified logging model based on ILogger.

Any ideas are welcome.

I will look this PR over.

brentschmaltz: @AndersAbel can you help us understand clearly by an example?

brockallen: > Any ideas are welcome.

Perhaps have a per-instance logger that's optional, and then fall back to the static/shared instance.

src/Microsoft.IdentityModel.Logging.Abstractions/Microsoft.IdentityModel.Logging.Abstractions.csproj
brentschmaltz: @TimHannMSFT did you mean M.IM.Abstractions?

src/Microsoft.IdentityModel.Logging/IIdentityLogger.cs
sruke: About (2)
Moving forward Wilson will introduce ILogScrubber whose implementation will contain a minimumDataClassificationLevel property. This property will decide the level of scrubbing and will decouple the concept of PII and Logging. Until then, IIdentityLogger will use IdentityEventSource.ShowPII to dictate if messages with PII can be logged. Spoke to @trwalke offline who will explore maintaining a PII property that is separate from IIdentityLogger. 

About (1)
[Here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/586bb1a47777860b0a517f9399e2b216136ffb8c/src/Microsoft.IdentityModel.Logging/LogHelper.cs#:~:text=if%20(Logger.IsEnabled(EventLevel.Verbose))) is where logger checks if the given log level is enabled before proceeding to scrub the log arguments and formatting the message.

bgavrilMS: @sruke - the example you give for (1) is does not show how the logger would handle an expensive to compute message, This is more like: 

```csharp
SalOrMsalLog(Action<string> messageProducer)
{
   if (Logger.IsEnabled())
   {
          message = messageProducer.Invoke();
          logger.Log(message);
   }
}

// usage: 
Log(() => "this is an expensive message");
```

This is how ASP.NET Core's logger does it as well, while maintaiing the ILogger / IIdentityLogger interface simple

src/Microsoft.IdentityModel.Abstractions/IIdentityLogger.cs
brentschmaltz: @jennyf19 I think this is a excellent comment. 
@sruke can you determine if we can add the benefits of high performance logging without any additional changes in a future release. Keep in mind that there will be target version (desktop, core) issues.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
brentschmaltz: @brockallen LogHelper is essentially a static class i am not sure how a per-instance would work.
Are you thinking that it would be helpful to have two instances of JwtSecurityTokenHandler use two different instances of LogHelper or a perhaps a different ILogger?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -371,5 +418,47 @@ public static object MarkAsNonPII(object arg)
...
+                // if PII is turned off and 'innerException' is a System exception only display the exception type
+                if (!IdentityModelEventSource.ShowPII && !LogHelper.IsCustomException(innerException))
+                    message = string.Format(CultureInfo.InvariantCulture, "Message: {0}, InnerException: {1}", message, innerException.GetType());
+                else // otherwise it's safe to display the entire exception message
+                    message = string.Format(CultureInfo.InvariantCulture, "Message: {0}, InnerException: {1}", message, innerException.Message);
+            }
+
+            LogEntry entry = new LogEntry();
+
+            // Logs basic information (library version, DateTime, weather PII is ON/OFF) once before any log messages are written.
```
brentschmaltz: grammer: weather -> whether

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -371,5 +418,47 @@ public static object MarkAsNonPII(object arg)
...
+                Logger.Log(entry);
+
+                entry.Message = IdentityModelEventSource.ShowPII ? _piiOnLogMessage : _piiOffLogMessage;
+                Logger.Log(entry);
+
+                HeaderWritten = true;
+            }
+
+            entry.EventLevel = level;
+            entry.Message = args != null ? FormatInvariant(message, args) : message;
```
brentschmaltz, line 462: Will the message always be logged?
We are formatting the string.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -266,6 +301,9 @@ public static void LogInformation(string message, params object[] args)
         {
             if (IdentityModelEventSource.Logger.IsEnabled())
```
brentschmaltz: Although, Logger.WriteInformation checks IsEnabled and EventLevel, should we add a condition, so we don't make the call?
```cs
if (IdentityModelEventSource.Logger.IsEnabled() && IdentityModelEventSource.Logger.LogLevel >= EventLevel.Information)
{
}
```

src/Microsoft.IdentityModel.Logging/LogHelper.cs
brentschmaltz: @bgavrilMS are you suggesting the IIdentityLogger should have a ShowPII property?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
brockallen: I've not read all of this code, nor do I know the design. I assumed maybe there was an options instance somewhere and maybe that'd be where you'd hang a per-instance logger. But if not, then maybe there's nothing to be done.

AndersAbel: As a library author using MIM, I am *very* careful to not alter any global state/config in MIM, which would be the case with a static property. Consider this case:

```
App ----------------- Microsoft.IdentityModel
  |                      /
  |                     /
  +------Library-------/
```

An app references Microsoft.IdentityModel both directly and indirectly through a Library that has a dependency. For any config/setting that is per-instance, the library can create an own instance of the MIM object and do whatever it wants without affecting the behaviour seen by the applications direct usage. But if the property is static there are really just two bad options:

1. State in library documentation that application needs to do certain settings (i.e. assign a logger) to get intended functionality.
2. Set the static property from the Library and thus make the direct reference (that might have existed before) behave differently for the App.

src/Microsoft.IdentityModel.Logging/IIdentityLogger.cs
sruke: So logger.Log(message) expects a pre-formatted message that is scrubbed off of PII. 

The way this handled in Wilson is -
```csharp
if (Logger.IsEnabled(EventLevel.Verbose)) // if given log level is enabled proceed to compute message
{
     LogEntry entry = WriteEntry(EventLevel.Verbose, null, message, args) // message and args are passed separately
     Logger.Log(entry);
}
```

[WriteEntry](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/586bb1a47777860b0a517f9399e2b216136ffb8c/src/Microsoft.IdentityModel.Logging/LogHelper.cs#:~:text=private%20static%20LogEntry%20WriteEntry(EventLevel%20level%2C%20Exception%20innerException%2C%20string%20message%2C%20params%20object%5B%5D%20args)) is responsible for generating a LogEntry

```csharp
private static LogEntry WriteEntry(EventLevel level, Exception innerException, string message, params object[] args)
{
      LogEntry entry = new LogEntry();
      entry.EventLevel = level;
      entry.Message = args != null ? FormatInvariant(message, args) : message;
      return entry;
}
```

LogEntry in turn invokes [FormatInvariant](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/586bb1a47777860b0a517f9399e2b216136ffb8c/src/Microsoft.IdentityModel.Logging/LogHelper.cs#:~:text=public%20static%20string%20FormatInvariant(string%20format%2C%20params%20object%5B%5D%20args)) which handles the processing of the expensive message based on the value of IdentityModelEventSource.ShowPII

```csharp
public static string FormatInvariant(string format, params object[] args)
 {
       if (!IdentityModelEventSource.ShowPII) // replace arguments with their type
           return string.Format(CultureInfo.InvariantCulture, format, args.Select(RemovePII).ToArray()); 

       return string.Format(CultureInfo.InvariantCulture, format, args); // format expensive message
}
```

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: If message == null and args == null 
    entry.Message will be null
if message == null and args != null
    entry.Message will be string,Empty (as returned by FormatInvariant(..))

Added a check in WriteEntry method.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Only downside to that is (IdentityModelEventSource.Logger.LogLevel >= EventLevel.Informational) condition would be evaluated twice for all messages whose eventLevel >= Informational. Once in LogHelper and again in IdentityModelEventSource. 

It may be worth making the change for LogInformation and LogVerbose as by default LogLevel is set to EventLevel.Warning.

src/Microsoft.IdentityModel.Logging.Abstractions/Microsoft.IdentityModel.Logging.Abstractions.csproj
sruke: Yup, he meant M.IM.Abstractions.

src/Microsoft.IdentityModel.Logging.Abstractions/LogEntry.cs
sruke: Added a remark [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1823/files#diff-18eba56cd8e596f57f24882d588e069abfa792062ca4c586e261c52ace5de62e)

src/Microsoft.IdentityModel.Logging.Abstractions/IIdentityLogger.cs
sruke: [System.Diagnostics.Tracing](https://www.nuget.org/packages/System.Diagnostics.Tracing/) has minimal dependencies similar to [System.Collections](https://www.nuget.org/packages/System.Collections/) that TelemetryClient depends on. This will be kept as is unless there is a good argument against it.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Updated

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Updated. Please review the method one more time.

sruke:
> We need to expose the concept of PII in our logs. The app developer needs to select the level of PII they are confortable with when they define the `IIdentityLogger` interface. We could use the simple "is PII" approach both SAL and MSAL have today, or we use the more complex approach MISE has.
> 
> Here's a prototype showing some of my comments - with 1 PII level (best to review from bottom up or just put it in a console app and run it) https://gist.github.com/bgavrilMS/16d44dca6370588db0b95fc370e73557
> 
> Would be good to iterate on the design to understand how we can introduce MISE's PII levels.

@bgavrilMS, can we revert the changes requested option for now and evaluate if IIdentityLogger can be adopted as it is?

src/Microsoft.IdentityModel.Logging/IIdentityLogger.cs
bgavrilMS: This works, just make sure that message and args are not evaluated beforehand.

```csharp
if (Logger.IsEnabled(EventLevel.Verbose)) // if given log level is enabled proceed to compute message
{
     message = "Here is an expensive to compute message with arg0 {0} and arg1 {1} ";
     arg0 = ExpensiveOperation0();
     arg1 = ExpensiveOperation1();

     LogEntry entry = WriteEntry(EventLevel.Verbose, null, message, new[] {arg0, arg1}) // message and args are passed separately
     Logger.Log(entry);
}
```

src/Microsoft.IdentityModel.Logging/LogHelper.cs
bgavrilMS: I was suggesting this originally, but then @sruke did point out that separating PII concept from Logging concept is better long term and easier to integrate. So I think it's fine that IIdentitityLogger remains de-coupled from PII and each of our libs does PII in their own way (SAL via the static, MSAL via a user config). 

Ultimately we want to get to a point where the app developer can configure: 

- an IIdenittyLogger 
- a bool PII (simple, i.e. I don't want PII in my logs) or a log scrubber (fine grained control of PII)

src/Microsoft.IdentityModel.Logging/LogHelper.cs
brentschmaltz: @AndersAbel i agree with your analysis about statics.

Currently all the objects in our runtime use a single instance of the current logging object.
There is no way to have two instances of JsonWebTokenHandler use a different logger or even with different PII settings.
As @brockallen suggested, we could develop a model where our TokenHandlers by default use the 'global' but will accept an instance of 'ILogger'.

What this work is about is developing a way for users to tell M.IM: "use this logger".
In this way, we hope that users can have an easy way to pipe all logs to a single location through multiple layers.
We can supply an adaptor that allows easy usage of asp.net ILogger.

buildConfiguration.xml
```diff
@@ -19,6 +19,8 @@
       <project name="Microsoft.IdentityModel.KeyVaultExtensions" />
       <project name="Microsoft.IdentityModel.ManagedKeyVaultSecurityKey" />
       <project name="Microsoft.IdentityModel.Validators" />
+      <project name="Microsoft.IdentityModel.Abstractions" />
+      <project name="Microsoft.IdentityModel.TestExtensions" />
```
GeoK, line 23: Do we want to release this package?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -371,5 +418,50 @@ public static object MarkAsNonPII(object arg)
...
+                // if PII is turned off and 'innerException' is a System exception only display the exception type
+                if (!IdentityModelEventSource.ShowPII && !LogHelper.IsCustomException(innerException))
+                    message = string.Format(CultureInfo.InvariantCulture, "Message: {0}, InnerException: {1}", message, innerException.GetType());
+                else // otherwise it's safe to display the entire exception message
+                    message = string.Format(CultureInfo.InvariantCulture, "Message: {0}, InnerException: {1}", message, innerException.Message);
+            }
+
+            LogEntry entry = new LogEntry();
+
+            // Logs basic information (library version, DateTime, whether PII is ON/OFF) once before any log messages are written.
```
GeoK, line 443: Technically not correct as multiple threads might hit this code path in parallel and write a few duplicate messages. Not sure how important it is in this context. Using any sorts of locks on the hot path should be also avoided.
Consider moving HeaderWritten up to ln 448.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -37,6 +39,36 @@ namespace Microsoft.IdentityModel.Logging
...
     {
+        /// <summary>
+        /// Gets or sets a logger to which logs will be written to.
+        /// </summary>
+        public static IIdentityLogger Logger { get; set; } = NullIdentityModelLogger.Instance;
+
+        /// <summary>
+        /// Indicates whether or the log message header (contains library version, date/time, and PII debugging information) has been written.
+        /// </summary>
+        private static bool HeaderWritten { get; set; } = false;
```
GeoK:
```suggestion
        private static bool _headerWritten { get; set; } = false;
```

src/Microsoft.IdentityModel.Abstractions/IIdentityLogger.cs
sruke: My understanding is that logger extension methods is a traditional logging approach and Logger message provided by .Net Core is more optimized.
* With the logger extension methods all arguments are boxed into objects. LoggerMessage avoids this by using static Action fields.
* With logger extensions methods, the message template is parsed every time a message is written and with LoggerMessage the message template is only parsed once at the time it is defined.

LoggerAdapter that will be written for ASP.Net Core's ILogger will use traditional logger extension methods so changes would be needed.

Moving to .Net Core's high-performance logging would be a significant change for an existing library -
* Wilson would take a direct dependency on Microsoft.Extensions.Logging
* All log messages need to be redefined to follow the below pattern.
```csharp
public static class LoggerExtensions
{
      private static readonly Action<ILogger, string, string, Exception?> debugMessage  =
                LoggerMessage.Define<string, string>(
                    LogLevel.Debug,
                    IDX0001,
                    "TestKey {Arg1} TestValue {Arg2}");

     public static void TestRequest(
                ILogger logger,
                string key,
                string value,
                Exception? ex) => debugMessage(
                    logger,
                    testKey,
                    testValue,
                    ex);
}
```

Adopting high performance logging directly would make sense.

buildConfiguration.xml
sruke: Microsoft.IdentityModel.Abstractions needs to be released. Not sure about TestExtensions. It was copied from Tim's Telemetry PR.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Thank you for catching it. I'll modify it.

bgavrilMS (approved):

buildConfiguration.xml
GeoK: Adding TestExtensions to buildConfiguration.xml will result in packing/releasing it. Not sure if that's something that we want to do.

TimHannMSFT: I believe it's needed to be released if we want folks to be able to consume it correct? the whole conceit of TestExentions was so folks could more easily reference it to write unit tests. If you think there's a better way to expose that I'm open to that.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
TimHannMSFT: what's the argument for a property vs a field here?

TimHannMSFT: perhaps it could be a property but you can only set it to true? Are there scenarios where anyone should be setting this to false?

TimHannMSFT:
> //------------------------------------------------------------------------------

no changes? Revert?

---
Refers to: src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs:1 in 9a3671b. [](commit_id = 9a3671bdad96512dcfbf162b5a443c3f6db2f28e, deletion_comment = False)

TimHannMSFT:
> //------------------------------------------------------------------------------

no changes, revert?

---
Refers to: src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs:1 in 9a3671b. [](commit_id = 9a3671bdad96512dcfbf162b5a443c3f6db2f28e, deletion_comment = False)

TimHannMSFT (approved):

src/Microsoft.IdentityModel.Abstractions/IIdentityLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.IdentityModel.Abstractions
+{
+    /// <summary>
+    /// Interface for Logging.
```
jmprieur:
```suggestion
    /// Interface that needs to be implemented by classes providing logging in
    /// Microsoft identity libraries
```

src/Microsoft.IdentityModel.Abstractions/IIdentityLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// </summary>
+    public interface IIdentityLogger
+    {
+        /// <summary>
+        /// Checks to see if logging is enabled at given <paramref name="eventLevel"/>.
+        /// </summary>
+        /// <param name="eventLevel">Log level of an Event.</param>
+        bool IsEnabled(EventLevel eventLevel);
+
```
jmprieur, line 42: Would it make sense to have IsEnabled(EventLevel, PIILevel) ?

jmprieur (approved):
LGTM
Thanks @sruke

buildConfiguration.xml
jennyf19: Do we have docs and samples on how to use TestExtensions?

src/Microsoft.IdentityModel.Abstractions/IIdentityLogger.cs
sruke: Closed is an offline discussion that we would like to keep PII separate from logger interface. This would make adoption of more granular PII levels easier in the future.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Code Analyzer doesn't allow me to create a field on its own. It has to be encapsulated in a property. It is only set and always set to true before logging the first message.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Toggling _headerWritten value right after verifying header hasn't been written yet.

src/Microsoft.IdentityModel.Abstractions/IIdentityLogger.cs
sruke: Implemented suggestion. Thank you.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Updated property name.

brentschmaltz (approved):
Please wait for JM

PR #1824, opened 2022-03-16
Incorrect logging: "ValidFrom: 'System.DateTime', Current time: 'System.DateTime'."

ivarne:
I got an unhelpful error message in my project, and googling the message sendt me here.
image

> JwtCookie was not authenticated. Failure message: IDX10222: Lifetime validation failed. The token is not yet valid. ValidFrom: 'System.DateTime', Current time: 'System.DateTime'.

It seems somewhat likely that this is because the wrong string conversion is used after the log formatter casts the parameters to `Object`. Please consider this an issue with some research, as I have not yet spent the time to actually test this change locally.


jennyf19 (approved):

jennyf19:
@brentschmaltz

brentschmaltz:
@ivarne thanks for posting this.
@sruke please look into this and see where else we may have this issue.

ivarne:
Hi, I discussed this with a colleague that claimed that there was no way to "accidentally" cast a `DateTime` to `object` and get the generic `object.ToString` method that returns the name of the type of the object. We tested a little, and could not figure out a way to get the `ValidateLifetime` function to create a message like the one I saw, so it seems likely that this suggested fix will not change the result, so I'm closing this. Sorry for suggesting a fix without proper testing.

PR #1825, opened 2022-03-18
Configuration validator

ciaozhang:
Introduce an ConfigurationValidator class that can be used to apply validation rules to a retrieved OpenIdConnectConfiguration.

Create a new exception ConfigurationValidationException, and wrap/throw this exception in the ConfigurationValidationResult if the metadata violates the rules.


src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -240,7 +240,10 @@ internal static bool IsRecoverableConfiguration(TokenValidationParameters valida
             {
                 // Inform the user that the LKG is expired.
                 if (!validationParameters.ConfigurationManager.IsLastKnownGoodValid)
+                {
                     LogHelper.LogInformation(TokenLogMessages.IDX10263);
+                    return false;
```
ciaozhang, line 245: add return false if LKG is not valid

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/JsonWebKeySetBadX509Data.json
```diff
@@ -1 +1,26 @@
...
\ No newline at end of file
+{
```
brentschmaltz, line 1: nice!

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -142,12 +160,25 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                         _currentConfiguration = await _configRetriever.GetConfigurationAsync(MetadataAddress, _docRetriever, CancellationToken.None).ConfigureAwait(false);
                         _lastRefresh = now;
                         _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval);
+                        if (_configValidator != null)
+                        {
+                            ConfigurationValidationResult result = _configValidator.Validate(_currentConfiguration);
+                            if (!result.Succeeded)
+                            {
+                                throw new ConfigurationValidationException(LogMessages.IDX20810, result.Exception);
```
brentschmaltz: you should use LogHelper.LogExceptionMessage(...)

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -142,12 +160,25 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
                     }
                     catch (Exception ex)
                     {
                         _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
                         if (_currentConfiguration == null) // Throw an exception if there's no configuration to return.
                             throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX20803, (MetadataAddress ?? "null")), ex));
+                        else if (ex is ConfigurationValidationException)
+                        {
+                            _currentConfiguration = null;
+                            throw LogHelper.LogExceptionMessage(ex);
```
brentschmaltz: not sure we should throw here.
this clause is not needed if we use the logic above.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -142,12 +160,25 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                         _currentConfiguration = await _configRetriever.GetConfigurationAsync(MetadataAddress, _docRetriever, CancellationToken.None).ConfigureAwait(false);
```
brentschmaltz: I am thinking not to throw if bad config is received as _currentConfiguration may be fine.
throwing may also interfere with LKG.
```cs
try
{
    var configuration = await _configRetriever.GetConfigurationAsync(
        MetadataAddress,
       _docRetriever, 
       CancellationToken.None).ConfigureAwait(false);

    if (_configValidator != null && !_configValidator.Validate(currentConfiguration).Succeeded)
        LogHelper.LogWarning(LogMessages.IDX20810, result.Exception);

}

**Review by brentschmaltz — COMMENTED** [review]
I am thinking we shouldn't throw on Validation error as LKG or _currentConfiguration may be usable.

File: src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs (hunk 5 above)
Comment (TimHannMSFT) [reply src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs]: I'm assuming the example above should be:

...
   if (_configValidator != null && !_configValidator.Validate(configuration).Succeeded)
...

right (instead of checking if _currentConfiguration is invalid which is the stale one?

Seems fine to me. Would have to be sure not to update _syncAfter or make it try again soon at least right? Something like 5min+jitter?

I think continuing to send back the stale config instead of failing is ok here should the stale version be validated as well? If _configValidator.Validate() fails for _currentConfiguration && configuration seems like it would be worth throwing then, no?

Comment (brentschmaltz) [reply src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs]: @TimHannMSFT that is what we do today, in the case we couldn't get new metadata.
In this case, i am suggesting we just skip any invalid metadata.

File: src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationValidationResult.cs [hunk 6]
```diff
@@ -0,0 +1,48 @@
...
+{
+    /// <summary>
+    /// Represents the result of validation a <see cref="IConfigurationValidator{T}"/>.
+    /// </summary>
+    public class ConfigurationValidationResult
+    {
+        /// <summary>
+        /// Gets or sets the <see cref="ConfigurationValidationException"/> that occurred during validation of the SignedHttpRequest.
+        /// </summary>
+        public ConfigurationValidationException Exception { get; set; }
```
brentschmaltz: Since we decided not to throw, do we need an Exception here or just a ErrorMessage (string)?

ciaozhang: changed to error message

src/Microsoft.IdentityModel.Protocols/Exceptions/ConfigurationValidationException.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Runtime.Serialization;
+
+namespace Microsoft.IdentityModel.Protocols
+{
+
+    /// <summary>
+    /// This exception is thrown when an OpenIdConnect protocol handler encounters a protocol error.
+    /// </summary>
+    [Serializable]
+    public class ConfigurationValidationException : Exception
```
brentschmaltz: Is there a reason to have this class anymore?

ciaozhang: Removed

brentschmaltz (approved):

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -139,9 +157,18 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                     {
                         // Don't use the individual CT here, this is a shared operation that shouldn't be affected by an individual's cancellation.
                         // The transport should have it's own timeouts, etc..
-                        _currentConfiguration = await _configRetriever.GetConfigurationAsync(MetadataAddress, _docRetriever, CancellationToken.None).ConfigureAwait(false);
+                        var configuration = await _configRetriever.GetConfigurationAsync(MetadataAddress, _docRetriever, CancellationToken.None).ConfigureAwait(false);
                         _lastRefresh = now;
```
dannybtsai, line 161: Does it make more sense to set the _currentConfiguration only if:
1. no validator is provided
2. validation succeeds, or validation fails but _currentConfiguration is null


                        _lastRefresh = now;
                        _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval);
                        if (_configValidator != null)
                        {
                            if (!result.Succeeded)
                            {
                                LogHelper.LogWarning(LogMessages.IDX20810, result.Exception);
                                if (_currentConfiguration == null)
                            }
                            else
                            {
                            }
                        }
                        else
                        {
                        }


 #Closed

PR #1833, opened 2022-04-08
added the SecurityToken.RawToken property to support more token types including MSA

dannybtsai:
Currently the classes deriving from the SecurityToken have their own properties for setting the raw token values including:

1. JsonWebToken.EncodedToken
2. JwtSecurityToken.RawDara

With the addition of the MsaToken(and other types to be supported), a new property, RawToken, is now added to the SecurityToken class to capture the raw value of a token.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/SecurityToken.cs
+++ b/src/Microsoft.IdentityModel.Tokens/SecurityToken.cs
@@ -68,6 +68,6 @@ namespace Microsoft.IdentityModel.Tokens
         /// <summary>
         /// Gets the original raw data of this instance when it was created.
         /// </summary>
-        public string RawToken { get; protected internal set; }
+        public string RawToken { get; set; }
     }
 }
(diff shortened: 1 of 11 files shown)


src/Microsoft.IdentityModel.Tokens/SecurityToken.cs
```diff
@@ -64,5 +64,10 @@ public abstract class SecurityToken
         /// This must be overridden to get the time when this <see cref="SecurityToken"/> is no longer Valid.
         /// </summary>
         public abstract DateTime ValidTo { get; }
+
+        /// <summary>
+        /// Gets the original raw data of this instance when it was created.
+        /// </summary>
+        public string RawToken { get; protected internal set; }
```
brentschmaltz: A custom SecurityToken will have to expose an internal method so the custom SecurityTokenHandler will be able to set it.
Could we code this as:
```cs
public virtual string RawToken { get; } = string.Empty;
```
or
```cs
public string RawToken { get; set; }
```
Then each SecurityTokenHandler will know how to set the RawToken, they can have their own contract.
With the virtual we will not be able to set the value in the ctor.

dannybtsai: This makes sense, changed to a public setter as suggested so it is consistent with, for example, the SigningKey property.

brentschmaltz (changes requested):
@dannybtsai i spoke with Shankar and this is not needed right now.
Can we hold off?

brentschmaltz:
We found out we do not need this right now.

Issue #1426, opened 2020-05-27
Add net5.0 target


GeoK:
Add net5.0 target and optimize hot-paths to leverage new .net features and System.Text.Json.

brentschmaltz:
We added 6.0 target

PR #1834, opened 2022-04-08
ECDH-ES Key Exchange Provider

### LINKED ISSUE #1726: Support ECDH-ES Key Agreement
Currently there's no support for ECDH-ES as noted in this [issue ](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1666)

Given the increase in usage of this algorithm we should support it instead of leaving it to users to implement via extension. 

We should support:
ECDH-ES+A128KW
ECDH-ES+A192KW
ECDH-ES+A256KW

Direct or symmetric key wrap

cesarpayan:
Creating PR for feedback. PR will be 'blocked' until perf/stress testing is complete.


src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
```diff
@@ -0,0 +1,275 @@
...
+        /// <param name="privateKey">The <see cref="ECDsaSecurityKey"/> that will be used for cryptographic operations and represents the private key.</param>
+        /// <param name="publicKey">The <see cref="ECDsaSecurityKey"/> that will be used for cryptographic operations and represents the public key.</param>
+        /// <param name="alg">alg header parameter value.</param>
+        /// <param name="enc">enc header parameter value.</param>
+        /// </summary>
+        public EcdhKeyExchangeProvider(ECDsaSecurityKey privateKey, ECDsaSecurityKey publicKey, string alg, string enc)
+        {
+            if (privateKey == null)
+                throw LogHelper.LogArgumentNullException(nameof(privateKey));
+            if (publicKey is null)
```
brentschmaltz: we should have a new-line separating clauses.

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
```diff
@@ -0,0 +1,275 @@
...
+        private string _algorithmId;
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="EcdhKeyExchangeProvider"/> used for CEKs
+        /// <param name="privateKey">The <see cref="ECDsaSecurityKey"/> that will be used for cryptographic operations and represents the private key.</param>
+        /// <param name="publicKey">The <see cref="ECDsaSecurityKey"/> that will be used for cryptographic operations and represents the public key.</param>
+        /// <param name="alg">alg header parameter value.</param>
+        /// <param name="enc">enc header parameter value.</param>
+        /// </summary>
+        public EcdhKeyExchangeProvider(ECDsaSecurityKey privateKey, ECDsaSecurityKey publicKey, string alg, string enc)
```
brentschmaltz: These two constructors do very similar work.
We could do two things
1. have a single ctor that takes a SecurityKey, then check that the keys are ECDsaSecurityKey OR a JsonWebKey that can be converted to a ECDsaSecurityKey.
2. have initialize methods that perform the common work.

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
```diff
@@ -0,0 +1,275 @@
...
+        /// represented as a base64url-encoded string.</param>
+        /// <param name="apv">Agreement PartyVInfo (optional). When used, the PartyUInfo value contains information about the recipient,
+        /// represented as a base64url-encoded string.</param>
+        /// <returns>Returns <see cref="SecurityKey"/> that represents the key generated</returns>
+        public SecurityKey GenerateKdf(string apu = null, string apv = null)
+        {
+            //The "apu" and "apv" values MUST be distinct when used (per rfc7518 section 4.6.2) https://datatracker.ietf.org/doc/html/rfc7518#section-4.6.2
+            if (!string.IsNullOrEmpty(apu)
+                && !string.IsNullOrEmpty(apv)
+                && apu.Equals(apv, StringComparison.InvariantCulture))
```
brentschmaltz: We should not specify culture here, we want OrdinalCaseSensitive, which is the default.

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        private static byte[] Concat(params byte[][] arrays)
+        {
+            int length = 0;
+            foreach (byte[] arr in arrays)
+                length += arr.Length;
+
+            byte[] output = new byte[length];
+            int x = 0;
+            foreach (byte[] arr in arrays)
```
brentschmaltz, line 223: Can we use ArrayCopy as we do here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/039d09b7bc4c1671e1c4093d5a62d307c2f96935/src/Microsoft.IdentityModel.Tokens/Encryption/SymmetricKeyWrapProvider.cs#L280

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
```diff
@@ -0,0 +1,275 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.IO;
+using System.Security.Cryptography;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+#if NET472
```
brentschmaltz: We need to add an additional target that can be used from .Net core.
We could add 5.0 and close this issue.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1426

cesarpayan: Question: Should I make the additional target part of this PR?

Comment:
There are two referenced classes that have potential conflicts with versioning in this file:
- [ECParameters](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecparameters?view=net-6.0) is available in all Core versions, and .NET Framework 4.7 onwards
- [ECDiffieHellman](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecdiffiehellman?view=net-6.0) is available in Core 2.1 onwards, and .NET Framework 3.5 onwards

Would it look something like?
`#if NET472_OR_GREATER || NETSTANDARD2_1_OR_GREATER`

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
cesarpayan: Changed in a52ecf360f5d0463192c760898765e2a8542ee68

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
cesarpayan: Fixed in a52ecf360f5d0463192c760898765e2a8542ee68

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
cesarpayan: Changed in a52ecf360f5d0463192c760898765e2a8542ee68
Taking the first suggestion

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
cesarpayan: Fixed in a52ecf360f5d0463192c760898765e2a8542ee68

cesarpayan (commented):
Addressing comments

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
```diff
@@ -56,51 +52,39 @@ public class EcdhKeyExchangeProvider //todo: rename to KeyExchangeProvider, or E
         /// <summary>
         /// Initializes a new instance of <see cref="EcdhKeyExchangeProvider"/> used for CEKs
         /// <param name="privateKey">The <see cref="ECDsaSecurityKey"/> that will be used for cryptographic operations and represents the private key.</param>
-        /// <param name="publicKey">The <see cref="ECDsaSecurityKey"/> that will be used for cryptographic operations and represents the public key.</param>
+        /// <param name="publicKey">The <see cref="SecurityKey"/> that will be used for cryptographic operations and represents the public key.</param>
         /// <param name="alg">alg header parameter value.</param>
         /// <param name="enc">enc header parameter value.</param>
         /// </summary>
-        public EcdhKeyExchangeProvider(ECDsaSecurityKey privateKey, ECDsaSecurityKey publicKey, string alg, string enc)
+        public EcdhKeyExchangeProvider(ECDsaSecurityKey privateKey, SecurityKey publicKey, string alg, string enc)
```
brentschmaltz: An ECDsaSecurityKey private key could be represented by a JsonWebKey, unlikely since JsonWebKey is normally associated with public metadata.
That said, it is possible so both parameters could be SecurityKey.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -263,6 +263,7 @@ internal static class LogMessages
         // Crypto Errors
         public const string IDX11000 = "IDX11000: Cannot create EcdhKeyExchangeProvider. '{0}'\'s Curve '{1}' does not match with '{2}'\'s curve '{3}'.";
         public const string IDX11001 = "IDX11001: Cannot generate KDF. '{0}':'{1}' and '{2}':'{3}' must be different.";
```
brentschmaltz, line 267: Perhaps a new error message for IDX1101 since the error is in the ctor:
```cs
public const string IDX11003 = "IDX11003: Cannot create the EcdhKeyExchangeProvider. the 'publicKey' cannot be converted to an ECDsaSecurityKey. publicKey type is: '{0}'.";
```

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
```diff
@@ -56,51 +52,39 @@ public class EcdhKeyExchangeProvider //todo: rename to KeyExchangeProvider, or E
...
-        /// </summary>
-        public EcdhKeyExchangeProvider(ECDsaSecurityKey privateKey, JsonWebKey publicKey, string alg, string enc)
-        {
-            if (privateKey == null)
-                throw LogHelper.LogArgumentNullException(nameof(privateKey));
             if (publicKey is null)
                 throw LogHelper.LogArgumentNullException(nameof(publicKey));
 
-            ValidateAlgAndEnc(alg, enc);
+            if (publicKey is ECDsaSecurityKey ecdsaKey)
```
brentschmaltz: You should call [TryConvertToECDsaSecurityKey](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/050297acc5b1dacba8f706459af6da8803762453/src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs#L344)

The reason is related to we have some logic in [ECDsaAdapter ](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/050297acc5b1dacba8f706459af6da8803762453/src/Microsoft.IdentityModel.Tokens/ECDsaAdapter.cs#L73)that can create an ECDsa object by examining the runtime, regardless of the 'target' specified by the application. Even though an application has targeted say .NET462, the actual runtime is most likely .NET48

You could write this as:
```cs
if (!TryConvertToECDsaSecurityKey(publicKey, out ecdaPublickey)
    throw LogHelper.LogArgumentException<ArgumentException>(
                    nameof(publicKey),
                    LogHelper.FormatInvariant(LogMessages.IDX11003, LogHelper.MarkAsNonPII(publicKey.GetType())));

    // then see if this works... if so, then GetEcParamsFromJwk would not be needed.
    _ecParamsPublic = ecdsaPublickey.ECDSA.ExportParameters(false);
```

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
```diff
@@ -56,51 +52,39 @@ public class EcdhKeyExchangeProvider //todo: rename to KeyExchangeProvider, or E
...
+            }
+            else if (publicKey is JsonWebKey jwk)
+            {
+                _ecParamsPublic = GetEcParamsFromJwk(jwk);
+            }
+            else
+            {
+                throw LogHelper.LogArgumentException<ArgumentException>(
+                    nameof(publicKey),
+                    LogHelper.FormatInvariant(LogMessages.IDX11001, LogHelper.MarkAsNonPII(nameof(publicKey))));
```
brentschmaltz: change error message, see comment in LogMessages.cs.

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
```diff
@@ -116,7 +100,7 @@ public SecurityKey GenerateKdf(string apu = null, string apv = null)
             //The "apu" and "apv" values MUST be distinct when used (per rfc7518 section 4.6.2) https://datatracker.ietf.org/doc/html/rfc7518#section-4.6.2
             if (!string.IsNullOrEmpty(apu)
                 && !string.IsNullOrEmpty(apv)
-                && apu.Equals(apv, StringComparison.InvariantCulture))
+                && apu.Equals(apv, StringComparison.Ordinal))
```
brentschmaltz: We found out that adding 'StringComparison.Ordinal', adds a performance penalty.
We should just leave it out.

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
cesarpayan: Per PR Meeting, we'll add .NET Core 3.1

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -755,13 +755,14 @@ private static string EncryptTokenPrivate(string innerJwt, EncryptingCredentials
                 throw LogHelper.LogExceptionMessage(new ArgumentException(TokenLogMessages.IDX10620));
```
dannybtsai, line 897: This is a copy from the JwtSecurityTokenHandler.EncryptToken() but missing a null check on encryptingCredentials.
Suggest adding the GetCryptoProviderFactory() to the base class TokenHandler:

```
        protected static CryptoProviderFactory GetCryptoProviderFactory(EncryptingCredentials encryptingCredentials)
        {
            if (encryptingCredentials == null)
                throw LogHelper.LogArgumentNullException(nameof(encryptingCredentials));

            var cryptoProviderFactory = encryptingCredentials.CryptoProviderFactory ?? encryptingCredentials.Key.CryptoProviderFactory;

            if (cryptoProviderFactory == null)
                throw LogHelper.LogExceptionMessage(new ArgumentException(TokenLogMessages.IDX10620));

            return cryptoProviderFactory;
        }
```

And both JsonWebTokenHandler and JwtSecurityTokenHandler get the factory in a consistent manner:

`           var cryptoProviderFactory = GetCryptoProviderFactory(encryptingCredentials);
`

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -830,6 +835,24 @@ internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken
...
+                        jwtToken.TryGetHeaderValue(JwtHeaderParameterNames.Apu, out string apu);
+                        jwtToken.TryGetHeaderValue(JwtHeaderParameterNames.Apv, out string apv);
+                        SecurityKey kdf = ecdhKeyExchangeProvider.GenerateKdf(apu, apv);
+                        var kwp = key.CryptoProviderFactory.CreateKeyWrapProviderForUnwrap(kdf, ecdhKeyExchangeProvider.GetEncryptionAlgorithm());
+                        var unwrappedKey = kwp.UnwrapKey(Base64UrlEncoder.DecodeBytes(jwtToken.EncryptedKey));
+                        unwrappedKeys.Add(new SymmetricSecurityKey(unwrappedKey));
+                    }
+                    else
+#endif
                     if (key.CryptoProviderFactory.IsSupportedAlgorithm(jwtToken.Alg, key))
```
dannybtsai, line 997: >CryptoProviderFactory [](http://example.com/codeflow?start=28&length=21)

Could this be null?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -886,7 +916,7 @@ protected virtual SecurityKey ResolveTokenDecryptionKey(string token, JsonWebTok
             {
                 if (validationParameters.TokenDecryptionKey != null)
                 {
-                    if (string.Equals(validationParameters.TokenDecryptionKey.KeyId, jwtToken.X5t, validationParameters.TokenDecryptionKey is X509SecurityKey ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
+                    if (string.Equals(validationParameters.TokenDecryptionKey.KeyId, jwtToken.X5t, stringComparison))
```
dannybtsai, line 1054: >stringComparison [](http://example.com/codeflow?start=99&length=16)

This is different than the original logic, is that correct? And it is differnt than the key collection below.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -866,17 +889,24 @@ protected virtual SecurityKey ResolveTokenDecryptionKey(string token, JsonWebTok
             if (validationParameters == null)
                 throw LogHelper.LogArgumentNullException(nameof(validationParameters));
 
+            StringComparison stringComparison = (validationParameters.TokenDecryptionKey is X509SecurityKey
+                || validationParameters.TokenDecryptionKey is ECDsaSecurityKey)
+                ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
             if (!string.IsNullOrEmpty(jwtToken.Kid))
```
dannybtsai, line 1034: This is repeated and I got lost easily (there are two versions), suggest replacing with calls like the following:

        private static StringComparison GetCaseRuleIfX509(SecurityKey key) => key is X509SecurityKey ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
        private static StringComparison GetCaseRuleIfECDsa(SecurityKey key) => (key is X509SecurityKey || key is ECDsaSecurityKey) ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -292,6 +301,26 @@ internal static SecurityKey GetSecurityKey(EncryptingCredentials encryptingCrede
...
+            {
+                // on decryption we get the public key from the EPK value see: https://datatracker.ietf.org/doc/html/rfc7518#appendix-C
+                string apu = null, apv = null;
+                if (additionalHeaderClaims != null && additionalHeaderClaims.Count > 0)
+                {
+                    if (additionalHeaderClaims.TryGetValue(JwtHeaderParameterNames.Apu, out object objApu))
+                        apu = objApu.ToString();
+                    if (additionalHeaderClaims.TryGetValue(JwtHeaderParameterNames.Apv, out object objApv))
+                        apv = objApv.ToString();
+                }
```
dannybtsai, line 320: objApu and objApv can be null.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -292,6 +301,26 @@ internal static SecurityKey GetSecurityKey(EncryptingCredentials encryptingCrede
...
+                EcdhKeyExchangeProvider ecdhKeyExchangeProvider = new EcdhKeyExchangeProvider(encryptingCredentials.Key as ECDsaSecurityKey, encryptingCredentials.JsonWebKey, encryptingCredentials.Alg, encryptingCredentials.Enc);
+                SecurityKey kdf = ecdhKeyExchangeProvider.GenerateKdf(apu, apv);
+                kwProvider = cryptoProviderFactory.CreateKeyWrapProvider(kdf, ecdhKeyExchangeProvider.GetEncryptionAlgorithm());
+                SecurityKey cek = new SymmetricSecurityKey(JwtTokenUtilities.GenerateKeyBytes(ecdhKeyExchangeProvider.KeyDataLen * 2));
```
dannybtsai: >ecdhKeyExchangeProvider.KeyDataLen * 2 [](http://example.com/codeflow?start=94&length=38)

If ecdhKeyExchangeProvider.KeyDataLen is size in bits, what is the logic of 2x? This seems like a magic number and is not clear.

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
```diff
@@ -109,6 +109,11 @@ public string Enc
             private set => _enc = string.IsNullOrEmpty(value) ? throw LogHelper.LogArgumentNullException("enc") : value;
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
```
dannybtsai, line 114: Missing comments.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -333,6 +333,32 @@ internal RSAParameters CreateRsaParameters()
             };
         }
 
+#if NET472
+        internal ECParameters CreateEcParameters()
+        {
+            //todo: update log msgs
+            if (string.IsNullOrEmpty(Crv))
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10700, LogHelper.MarkAsNonPII(_className), LogHelper.MarkAsNonPII("Crv"))));
```
dannybtsai: >"Crv" [](http://example.com/codeflow?start=181&length=5)

nameof(Crv)

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -333,6 +333,32 @@ internal RSAParameters CreateRsaParameters()
...
+
+            if (string.IsNullOrEmpty(X))
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10700, LogHelper.MarkAsNonPII(_className), LogHelper.MarkAsNonPII("X"))));
```
dannybtsai: >"X" [](http://example.com/codeflow?start=181&length=3)

nameof(X)

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -333,6 +333,32 @@ internal RSAParameters CreateRsaParameters()
...
+
+            if (string.IsNullOrEmpty(Y))
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10700, LogHelper.MarkAsNonPII(_className), LogHelper.MarkAsNonPII("Y"))));
```
dannybtsai: >"Y" [](http://example.com/codeflow?start=181&length=3)

nameof(Y)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: After inspecting the code paths into these two methods, I've realized that in neither case it is possible for these two methods to receive encyptingCredentials with a null value (each code path has a previous check for null value. As these two methods are private, I'll remove the check for null value in JwtSecurityTokenHandler.EncryptToken()

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: AFAIK, no. I don't think it is possible to have a SecurityKey with CryptoProviderFactory being null (there's a default being assigned when one isn't provided)

brentschmaltz: SecurityKey.CryptoProviderFactory cannot be null.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -292,6 +301,26 @@ internal static SecurityKey GetSecurityKey(EncryptingCredentials encryptingCrede
...
             }
+#if NET472
+            else if (SupportedAlgorithms.EcdsaWrapAlgorithms.Contains(encryptingCredentials.Alg))
+            {
+                // on decryption we get the public key from the EPK value see: https://datatracker.ietf.org/doc/html/rfc7518#appendix-C
+                string apu = null, apv = null;
+                if (additionalHeaderClaims != null && additionalHeaderClaims.Count > 0)
+                {
+                    if (additionalHeaderClaims.TryGetValue(JwtHeaderParameterNames.Apu, out object objApu))
+                        apu = objApu.ToString();
```
brentschmaltz: add a new line after 312 to separate clauses.

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
```diff
@@ -109,6 +109,11 @@ public string Enc
...
+        public JsonWebKey JsonWebKey { get; set; }
```
brentschmaltz: For ECDH, this can be a JsonWebKey OR a ECDsaSecurityKey.
Suggest a property

```cs
public SecurityKey KeyExchangePublicKey {get;set;}
```

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: Not really, it is the same logic (if validationParameters.TokenDecryptionKey is x509 it'll use StringComparison.OrdinalIgnoreCase. The change is that I am pre-calculating which string comparison it is being used in order to save extra lines of code, combined with using functions as you suggested above, it should be clearer. Changes in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
cesarpayan: That's for the size for the wrapped key. Changing it to match the style/logic for AesCbcHmac below in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
cesarpayan: Added in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: Adding it in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
cesarpayan: Added in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
cesarpayan: Added clause for null and a unit test case for that scenario in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.Tokens/EncryptingCredentials.cs
cesarpayan: Changed in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
cesarpayan: Changed to SecurityKey and necessary changes below to accommodate to it in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
cesarpayan: I am refactoring this code in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202) as now the ctor receives both keys as security keys as suggested above, to avoid duplicate code.
Adding it with log Message IDX11002 (previous id had been removed as I removed the code path that lead to it).
I am still checking if the security key is ECDsaSecurityKey or JsonWebKey as TryConvertToECDsaSecurityKey takes in a JsonWebKey.

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
cesarpayan: Changed to IDX11002 in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
cesarpayan: Removed StringComparison.Ordinal and added a supression message with the performance penalty as justification in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
cesarpayan: Removing this piece of code as is no longer used in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
cesarpayan: Removing this piece of code as is no longer used in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
cesarpayan: Removing this piece of code as is no longer used in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
cesarpayan: Since I'm removing the code path that lead to IDX1102, I am modifying the error message and using IDX1102 for that error case. Changes in [1595e40](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/1595e4097fc1b5c0f5b7a00fccf7f878bdd9c202)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -830,6 +842,24 @@ internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken
             {
                 try
                 {
+#if NET472
```
brentschmaltz: We need to add a CORE 3.1 target

src/Microsoft.IdentityModel.Tokens/Encryption/EcdhKeyExchangeProvider.cs
cesarpayan: Added in [f6f8794](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/f6f8794f6309d9962e52064dd47858458ee10130)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: Added in [f6f8794](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1834/commits/f6f8794f6309d9962e52064dd47858458ee10130)

PR #1842, opened 2022-04-29
Adding netstandard1.3 target to Abstractions assembly

### LINKED ISSUE #1849: [Feature Request] Add .netstandard1.3 target to Abstractions assembly for MSAL to consume IIdentityLogger
MSAL targets .netstandard1.3 but Wilson doesn't. Evaluate adding .netstandard1.3 to Abstractions assembly alone.

sruke:
MSAL needs to take a dependency on Microsoft.IdentityModel.Abstractions assembly to implement IIdentityLogger interface.
MSAL targets netstandard1.3 but Wilson doesn't.

This PR adds a target to netstandard1.3 to Abstractions assembly alone in Wilson to support MSAL.


jennyf19 (changes requested):
:clock1:

jennyf19:
I think we decided to not go this route correct? If so, I would link to an issue w/the findings and close.

sruke:
netstandard1.3 is no longer supported. MSAL will look into first upgrading to netstandard2.0 and resume their integration with IIdentityLogger.

PR #1843, opened 2022-05-05
Custom log level Enum to remove dependency on EventLevel

### LINKED ISSUE #1845: [Bug - Preview version] IIdentityLogger does not log information created before serviceProvider is available
Only LogLevel.Trace works to log the first line (header) of the Wilson logs. This is because, for all the other log levels, the logger instance is not created yet. The logger instance in ASP.NET Core version requires an IServiceProvider, which is created in Id. Web [here](https://github.com/AzureAD/microsoft-identity-web/blob/jennyf/logger/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs#L186), but the static instances in Wilson have already been initialized by ASP.NET Core, so the NullIdentityModelLogger is getting created, which sets `IsEnabled` to false. Logging gets enabled later, but the initial messages are not capture.

FYI: @sruke @brentschmaltz @jmprieur

### LINKED ISSUE #1848: [Bug - PreviewVersion] Remove EventLevel dependency in IIdentityLogger for wider adoption in SDKs

### LINKED ISSUE #1850: [Feature Request] Improve Log Message Formatting
ConfigurationManager is printing verbose assembly information which can create a log of noise. JwtSecurityTokenHandler does the same.

It creates a lot of noise, not sure if that’s necessary, but maybe it’s there for historical reasons, and do we need this information or a reduced version of it?

cb586d3a-4aa0-4e28-9a69-5508798bd7a7

### LINKED ISSUE #1851: [Bug - PreviewVersion] Header Information is not always displayed
Only LogLevel.Trace works to log the first line (header) of the Wilson logs.

Log level for header information is set to EventLevel.LogAlways. This level doesn't have an equivalent in ILogger's LogLevel enum and is currently mapped to LogLevel.Trace in classes that implement IIdentityLogger. That's why these logs are only seen at Trace level. ILogger requires that each message has a log level defined and doesn't allow a message to be displayed always.

Legacy behavior should remain the same.
Option 1: Replace LogAlways with Information. Con - Missing version info at higher log levels
Option 2: Prefix first Wilson and SAL message with header. Pro - helpful for IcMs.

### LINKED ISSUE #1857: [Bug] Error message for IDX21308 contains a typo
Error message for [IDX21308](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/039d09b7bc4c1671e1c4093d5a62d307c2f96935/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/LogMessages.cs#L52) contains a typo.

Currently
"IDX21308: 'Azp' claim exists in the 'id_token' but 'ciient_id' is null. Cannot validate the 'azp' claim."

Should be
"IDX21308: 'Azp' claim exists in the 'id_token' but 'client_id' is null. Cannot validate the 'azp' claim."

sruke:
Microsoft.IdentityModel.Abstractions currently depends on System.Diagnostics.Tracing.EventLevel. This PR removes that dependency and replaces it with a custom LogSeverityLevel Enum. 

This was done so that SDKs that consume this assembly won't run into compatibility issues.

### DIFF
--- a/src/Microsoft.IdentityModel.Abstractions/EventLogLevel.cs
+++ b/src/Microsoft.IdentityModel.Abstractions/EventLogLevel.cs
@@ -28,24 +28,42 @@
 namespace Microsoft.IdentityModel.Abstractions
 {
     /// <summary>
-    /// A minimalistic <see cref="IIdentityLogger"/> implementation that is disabled by default and doesn't log.
+    /// Defines Event Log Levels.
     /// </summary>
-    public sealed class NullIdentityModelLogger : IIdentityLogger
+    public enum EventLogLevel
     {
         /// <summary>
-        /// Default instance of <see cref="NullIdentityModelLogger"/>.
+        /// No level filtering is done on this log level. Log messages of all levels will be logged.
         /// </summary>
-        public static NullIdentityModelLogger Instance { get; } = new NullIdentityModelLogger();
+        LogAlways = 0,
 
-        private NullIdentityModelLogger() { }
+        /// <summary>
+        /// Logs that describe an unrecoverable application or system crash, or a catastrophic failure that requires
+        /// immediate attention.
+        /// </summary>
+        Critical = 1,
 
-        /// <inheritdoc/>
-        public bool IsEnabled(EventLogLevel eventLogLevel) => false;
+        /// <summary>
+        /// Logs that highlight when the current flow of execution is stopped due to a failure. These should indicate a
+        /// failure in the current activity, not an application-wide failure.
+        /// </summary>
+        Error = 2,
 
-        /// <inheritdoc/>
-        public void Log(LogEntry entry)
-        {
-            // no-op
-        }
+        /// <summary>
+        /// Logs that highlight an abnormal or unexpected event in the application flow, but do not otherwise cause the
+        /// application execution to stop.
+        /// </summary>
+        Warning = 3,
+
+        /// <summary>
+        /// Logs that track the general flow of the application. These logs should have long-term value.
+        /// </summary>
+        Informational = 4,
+
+        /// <summary>
+        /// Logs that are used for interactive investigation during development. These logs should primarily contain
+        /// information useful for debugging and have no long-term value.
+        /// </summary>
+        Verbose = 5
     }
 }
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -29,7 +29,6 @@ using System;
 using System.Diagnostics.Tracing;
 using System.Globalization;
 using System.Linq;
-using System.Reflection;
 using Microsoft.IdentityModel.Abstractions;
 
 namespace Microsoft.IdentityModel.Logging
@@ -47,27 +46,17 @@ namespace Microsoft.IdentityModel.Logging
         /// <summary>
         /// Indicates whether the log message header (contains library version, date/time, and PII debugging information) has been written.
         /// </summary>
-        private static bool _isHeaderWritten { get; set; } = false;
-
-        /// <summary>
-        /// The log message that indicates the current library version.
-        /// </summary>
-        private static string _versionLogMessage = "Microsoft.IdentityModel version: {0}. ";
-
-        /// <summary>
-        /// The log message that indicates the date.
-        /// </summary>
-        private static string _dateLogMessage = "Date: {0}. ";
+        private static bool _isHeaderWritten = false;
 
         /// <summary>
         /// The log message that is shown when PII is off.
         /// </summary>
-        private static string _piiOffLogMessage = "PII (personally identifiable information) logging is currently turned off. Set IdentityModelEventSource.ShowPII to 'true' to view the full details of exceptions. ";
+        private static string _piiOffLogMessage = "PII logging is OFF. See https://aka.ms/IdentityModel/PII for details. ";
 
         /// <summary>
-        /// The log message that is shown when PII is off.
+        /// The log message that is shown when PII is on.
         /// </summary>
-        private static string _piiOnLogMessage = "PII (personally identifiable information) logging is currently turned on. Set IdentityModelEventSource.ShowPII to 'false' to hide PII from log messages. ";
+        private static string _piiOnLogMessage = "PII logging is ON, do not use in production. See https://aka.ms/IdentityModel/PII for details. ";
 
         /// <summary>
         /// Logs an exception using the event source logger and returns new <see cref="ArgumentNullException"/> exception.
@@ -286,8 +275,9 @@ namespace Microsoft.IdentityModel.Logging
             if (IdentityModelEventSource.Logger.IsEnabled() && IdentityModelEventSource.Logger.LogLevel >= eventLevel)
                 IdentityModelEventSource.Logger.Write(eventLevel, exception.InnerException, exception.Message);
 
-            if (Logger.IsEnabled(eventLevel))
-                Logger.Log(WriteEntry(eventLevel, exception.InnerException, exception.Message, null));
+            EventLogLevel eventLogLevel = Enum.IsDefined(typeof(EventLogLevel), (int)eventLevel) ? (EventLogLevel)eventLevel : EventLogLevel.Error;
+            if (Logger.IsEnabled(eventLogLevel))
+                Logger.Log(WriteEntry((EventLogLevel)eventLevel, exception.InnerException, exception.Message, null));
 
             return exception;
         }
@@ -302,8 +292,8 @@ namespace Microsoft.IdentityModel.Logging
             if (IdentityModelEventSource.Logger.IsEnabled() && IdentityModelEventSource.Logger.LogLevel >= EventLevel.Informational)
                 IdentityModelEventSource.Logger.WriteInformation(message, args);
 
-            if (Logger.IsEnabled(EventLevel.Informational))
-                Logger.Log(WriteEntry(EventLevel.Informational, null, message, args));
+            if (Enum.IsDefined(typeof(EventLogLevel), (int)EventLevel.Informational) && Logger.IsEnabled((EventLogLevel)EventLevel.Informational))
+                Logger.Log(WriteEntry((EventLogLevel)EventLevel.Informational, null, message, args));
         }
 
         /// <summary>
@@ -316,8 +306,8 @@ namespace Microsoft.IdentityModel.Logging
             if (IdentityModelEventSource.Logger.IsEnabled())
                 IdentityModelEventSource.Logger.WriteVerbose(message, args);
 
-            if (Logger.IsEnabled(EventLevel.Verbose))
-                Logger.Log(WriteEntry(EventLevel.Verbose, null, message, args));
+            if (Enum.IsDefined(typeof(EventLogLevel), (int)EventLevel.Verbose) && Logger.IsEnabled((EventLogLevel)EventLevel.Verbose))
+                Logger.Log(WriteEntry((EventLogLevel)EventLevel.Verbose, null, message, args));
         }
 
         /// <summary>
@@ -330,8 +320,8 @@ namespace Microsoft.IdentityModel.Logging
             if (IdentityModelEventSource.Logger.IsEnabled())
                 IdentityModelEventSource.Logger.WriteWarning(message, args);
 
-            if (Logger.IsEnabled(EventLevel.Warning))
-                Logger.Log(WriteEntry(EventLevel.Warning, null, message, args));
+            if (Enum.IsDefined(typeof(EventLogLevel), (int)EventLevel.Warning) && Logger.IsEnabled((EventLogLevel)EventLevel.Warning))
+                Logger.Log(WriteEntry((EventLogLevel)EventLevel.Warning, null, message, args));
         }
 
         /// <summary>
@@ -354,8 +344,9 @@ namespace Microsoft.IdentityModel.Logging
             if (IdentityModelEventSource.Logger.IsEnabled() && IdentityModelEventSource.Logger.LogLevel >= eventLevel)
                 IdentityModelEventSource.Logger.Write(eventLevel, innerException, message);
 
-            if (Logger.IsEnabled(eventLevel))
-                Logger.Log(WriteEntry(eventLevel, innerException, message, null));
+            EventLogLevel eventLogLevel = Enum.IsDefined(typeof(EventLogLevel), (int)eventLevel) ? (EventLogLevel)eventLevel : EventLogLevel.Error;
+            if (Logger.IsEnabled(eventLogLevel))
+                Logger.Log(WriteEntry((EventLogLevel)eventLevel, innerException, message, null));
 
             if (innerException != null) 
                 if (string.IsNullOrEmpty(argumentName))
@@ -422,11 +413,11 @@ namespace Microsoft.IdentityModel.Logging
         /// <summary>
         /// Creates a <see cref="LogEntry"/> by using the provided event level, exception argument, string argument and arguments list.
         /// </summary>
-        /// <param name="level"><see cref="EventLevel"/></param>
+        /// <param name="eventLogLevel"><see cref="EventLogLevel"/></param>
         /// <param name="innerException"><see cref="Exception"/></param>
         /// <param name="message">The log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
-        private static LogEntry WriteEntry(EventLevel level, Exception innerException, string message, params object[] args)
+        private static LogEntry WriteEntry(EventLogLevel eventLogLevel, Exception innerException, string message, params object[] args)
         {
             if (string.IsNullOrEmpty(message))
                 return null;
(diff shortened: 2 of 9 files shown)


src/Microsoft.IdentityModel.Abstractions/EventLogLevel.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// <summary>
+    /// Defines Event Log Levels.
+    /// </summary>
+    public enum EventLogLevel
+    {
+        /// <summary>
+        /// No level filtering is done on this log level. Log messages of all levels will be logged.
+        /// </summary>
+        LogAlways = 0,
+
```
sruke, line 39: LogAlways doesn't have an exact mapping to AspNetCore's [LogLevel](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=dotnet-plat-ext-6.0) enum. We might have to consider avoiding logging messages at EventLevel.LogAlways. Today only the [header information](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Logging/LogHelper.cs#:~:text=headerEntry.EventLevel%20%3D%20EventLevel.LogAlways%3B) is logged at this level.  #Resolved

src/Microsoft.IdentityModel.Abstractions/EventLogLevel.cs
```diff
@@ -0,0 +1,69 @@
...
+//
+//------------------------------------------------------------------------------
+
+namespace Microsoft.IdentityModel.Abstractions
+{
+    /// <summary>
+    /// Defines Event Log Levels.
+    /// </summary>
+    public enum EventLogLevel
+    {
```
sruke, line 34: EventLogLevel enum has the same values as [EventLevel](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventlevel?view=net-6.0) enum. #Resolved

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -286,8 +286,8 @@ public static Exception LogExceptionMessage(EventLevel eventLevel, Exception exc
             if (IdentityModelEventSource.Logger.IsEnabled() && IdentityModelEventSource.Logger.LogLevel >= eventLevel)
                 IdentityModelEventSource.Logger.Write(eventLevel, exception.InnerException, exception.Message);
 
-            if (Logger.IsEnabled(eventLevel))
-                Logger.Log(WriteEntry(eventLevel, exception.InnerException, exception.Message, null));
+            if (Logger.IsEnabled((EventLogLevel)eventLevel))
+                Logger.Log(WriteEntry((EventLogLevel)eventLevel, exception.InnerException, exception.Message, null));
```
sruke, line 281: Performing a direct explicit cast without checks. Let me know if Enum.TryParse(..) or Enum.IsDefinied(..) checks are preferable to make sure a mapping exists.
Enum.IsDefined check will look like -
```csharp
 if (Enum.IsDefined(typeof(EventLogLevel), (int)eventLevel) && Logger.IsEnabled((EventLogLevel)eventLevel)).
```

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -446,7 +446,7 @@ private static LogEntry WriteEntry(EventLevel level, Exception innerException, s
                 _isHeaderWritten = true;
 
                 LogEntry headerEntry = new LogEntry();
-                headerEntry.EventLevel = EventLevel.LogAlways;
+                headerEntry.EventLogLevel = EventLogLevel.Informational;
```
sruke: @jennyf19, I know we would like to see the header information at all log levels. Unfortunately, there is no equivalent to EventLevel.LogAlways in AspNetCore's LogLevel enum. Set the EventLogLevel to Informational as that is our recommended LogLevel. Thoughts on how this can be improved?

src/Microsoft.IdentityModel.Abstractions/EventLogLevel.cs
jennyf19: So that means there might be an issue because I'm only finding the header when setting Trace, but really, based on the comment here, the header should always be getting written.

src/Microsoft.IdentityModel.Abstractions/EventLogLevel.cs
jennyf19: Do we want to have None as an option? And if so, will the header info still print?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: Can it be not tied to a log level? Would it always be written in that case? That can also be a separate PR.

src/Microsoft.IdentityModel.Abstractions/EventLogLevel.cs
sruke: Yes there is an issue with the translation from EventLevel to ILogger's LogLevel. ILogger always expects a LogLevel and there isn't an option to always log a given message.
Thank you for capturing it as one of the bugs [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1845).

src/Microsoft.IdentityModel.Abstractions/EventLogLevel.cs
sruke: Goal is to have a custom Enum equivalent to [EventLevel](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventlevel?view=net-6.0) enum. As EventLevel doesn't have a None option we don't need to add one here.

When an AspNetCore app sets LogLevel to None then no log messages are written. Its equivalent to disabling the ILogger. None of our log messages including the header information will be written to the ILogger in this case as IsEnabled(..) will return false.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Each message has to be tied to a log level :( If we were to set it to Critical it might be color coded in red (not sure) and customers may not like it so settled for Information.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
TimHannMSFT: I don't think the int definition check would be costly and will give a much better devEx if anything ever goes wrong. Would be interesting to benchmark it though: [https://benchmarkdotnet.org/](https://benchmarkdotnet.org/)

src/Microsoft.IdentityModel.Logging/LogHelper.cs
TimHannMSFT: Where are we doing the mapping between EventLevel/EventLogLevel and LogLevel ([https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel))?

TimHannMSFT: I think I get it a bit more than this morning in the meeting, seems like we should just collapse critcal and LogAlways in the mapping to LogLevel, I think the worry about displaying red shouldn't be too concerning, LogAlways is already something that says, "People should always see this" right?

sruke: > Where are we doing the mapping between EventLevel/EventLogLevel and LogLevel (https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel)?

Mapping from EventLevel/EventLogLevel to LogLevel is done in the Identity Libraries/SDKs that implement IIdentityLogger.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
sruke: Agreed, I ran the benchmark and it takes 76ns on an average to evaluate the condition when IsDefined check is added. I'll add the check and run the performance tests between dev and this topic branch to compare the performance impact.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -286,8 +276,8 @@ public static Exception LogExceptionMessage(EventLevel eventLevel, Exception exc
             if (IdentityModelEventSource.Logger.IsEnabled() && IdentityModelEventSource.Logger.LogLevel >= eventLevel)
                 IdentityModelEventSource.Logger.Write(eventLevel, exception.InnerException, exception.Message);
 
-            if (Logger.IsEnabled(eventLevel))
-                Logger.Log(WriteEntry(eventLevel, exception.InnerException, exception.Message, null));
+            if (Enum.IsDefined(typeof(EventLogLevel), (int)eventLevel) && Logger.IsEnabled((EventLogLevel)eventLevel))
```
TimHannMSFT: >if (Enum.IsDefined(typeof(EventLogLevel), (int)eventLevel) && Logger.IsEnabled((EventLogLevel)eventLevel)) [](http://example.com/codeflow?start=12&length=106)

Here and below should be logging at critical or warning? Currently if the cast fails nothing is logged and flow continues, that seems incorrect to me.

TimHannMSFT:
> using System.Reflection;

looks as though this may no longer be needed.

---
Refers to: src/Microsoft.IdentityModel.Logging/LogHelper.cs:32 in 3ef77ab. [](commit_id = 3ef77abfccdbbae89b174c7d8e304dcb0b18bb87, deletion_comment = False)

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -47,27 +47,17 @@ public class LogHelper
...
-        /// The log message that indicates the date.
-        /// </summary>
-        private static string _dateLogMessage = "Date: {0}. ";
+        private static bool _isHeaderWritten = false;
 
         /// <summary>
         /// The log message that is shown when PII is off.
         /// </summary>
-        private static string _piiOffLogMessage = "PII (personally identifiable information) logging is currently turned off. Set IdentityModelEventSource.ShowPII to 'true' to view the full details of exceptions. ";
+        private static string _piiOffLogMessage = "PII logging is OFF. See https://aka.ms/IdentityModel/PII for details. ";
```
TimHannMSFT, line 54: >  [](http://example.com/codeflow?start=120&length=1)

Q: Why does _piiOffLogMessage and _piiOnLogMessage end with a space?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -440,26 +430,23 @@ private static LogEntry WriteEntry(EventLevel level, Exception innerException, s
...
-                headerEntry.Message = string.Format(CultureInfo.InvariantCulture, _versionLogMessage, typeof(IdentityModelEventSource).GetTypeInfo().Assembly.GetName().Version.ToString());
-                Logger.Log(headerEntry);
-
-                headerEntry.Message = string.Format(CultureInfo.InvariantCulture, _dateLogMessage, DateTime.UtcNow);
-                Logger.Log(headerEntry);
-
-                headerEntry.Message = IdentityModelEventSource.ShowPII ? _piiOnLogMessage : _piiOffLogMessage;
-                Logger.Log(headerEntry);
+                headerEntry.EventLogLevel = EventLogLevel.LogAlways;
+                headerEntry.Message = headerMessage;
```
TimHannMSFT, line 448: Seems like this never gets logged? It doesn't show up in the test case either.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -440,26 +430,23 @@ private static LogEntry WriteEntry(EventLevel level, Exception innerException, s
                     message = string.Format(CultureInfo.InvariantCulture, "Message: {0}, InnerException: {1}. ", message, innerException.Message);
             }
 
+            message = args == null ? message : FormatInvariant(message, args);
+
             // Logs basic information (library version, DateTime, whether PII is ON/OFF) once before any log messages are written.
             if (!_isHeaderWritten)
             {
                 _isHeaderWritten = true;
```
TimHannMSFT: >_isHeaderWritten = true; [](http://example.com/codeflow?start=16&length=24)

should this get set after the Logger.Log call is made? Seems like better to err or the side of double logging this as opposed to missing it right? I might be missing what prevents it but seems like even this way there's a potential for double logging in race conditions.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -440,26 +430,21 @@ private static LogEntry WriteEntry(EventLevel level, Exception innerException, s
...
+                string headerMessage = string.Format(CultureInfo.InvariantCulture, "Microsoft.IdentityModel Version: {0}. Date {1}. {2}",
+                    typeof(IdentityModelEventSource).Assembly.GetName().Version.ToString(), DateTime.UtcNow, IdentityModelEventSource.ShowPII ? _piiOnLogMessage : _piiOffLogMessage);
```
TimHannMSFT: nit: IMO much more readable on multiple line:

                string headerMessage = string.Format(
                    CultureInfo.InvariantCulture,
                    "Microsoft.IdentityModel Version: {0}. Date {1}. {2}",
                    typeof(IdentityModelEventSource).Assembly.GetName().Version.ToString(),
                    DateTime.UtcNow,
                    IdentityModelEventSource.ShowPII ? _piiOnLogMessage : _piiOffLogMessage);

test/Microsoft.IdentityModel.Logging.Tests/LoggerTests.cs
```diff
@@ -237,7 +238,38 @@ public void PrepareMessageWithNoArguments()
 
              var exception = LogHelper.LogExceptionMessage(new ArgumentException("This is the first parameter '{0}'. This is the second parameter '{1}'."));
         }
-    }
 
+        [Fact]
+        public void EventLevelToEventLogLevelMapping()
```
TimHannMSFT, line 243: >EventLevelToEventLogLevelMapping [](http://example.com/codeflow?start=20&length=32)

validate the header is logged?

TimHannMSFT (changes requested):
:clock1:

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -302,8 +291,8 @@ public static void LogInformation(string message, params object[] args)
             if (IdentityModelEventSource.Logger.IsEnabled() && IdentityModelEventSource.Logger.LogLevel >= EventLevel.Informational)
                 IdentityModelEventSource.Logger.WriteInformation(message, args);
 
-            if (Logger.IsEnabled(EventLevel.Informational))
-                Logger.Log(WriteEntry(EventLevel.Informational, null, message, args));
+            if (Enum.IsDefined(typeof(EventLogLevel), (int)EventLevel.Informational) && Logger.IsEnabled((EventLogLevel)EventLevel.Informational))
```
jmprieur, line 295: Is there an opportunity to extract the test in a function which would take the EvenLevel?
It seems to be used lines 294, 308, 322, 346
Or would this decrease the performance?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -101,7 +101,6 @@ public class JwtSecurityTokenHandler : SecurityTokenHandler
         /// </summary>
         static JwtSecurityTokenHandler()
         {
-            LogHelper.LogVerbose("Assembly version info: " + LogHelper.MarkAsNonPII(typeof(JwtSecurityTokenHandler).AssemblyQualifiedName));
```
jmprieur, line 104: Did we move this somewhere else? Or do we say: not needed? Is it line 438 of LogHelper?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -57,7 +57,6 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
         /// </summary>
         static ConfigurationManager()
         {
-            LogHelper.LogVerbose("Assembly version info: " + LogHelper.MarkAsNonPII(typeof(ConfigurationManager<T>).AssemblyQualifiedName));
```
jmprieur, line 60: Did we move this somewhere else? Or do we say: not needed? ? Is it line 438 of LogHelper?

jmprieur (approved):
LGTM
Thanks @sruke

TimHannMSFT (approved):
couple nits left it sounds like you're addressing but either way, this looks good. Thanks for this @sruke!

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/LogMessages.cs
```diff
@@ -49,7 +49,7 @@ internal static class LogMessages
         internal const string IDX21305 = "IDX21305: OpenIdConnectProtocolValidationContext.ProtocolMessage.Code is null, there is no 'code' in the OpenIdConnect Response to validate.";
         internal const string IDX21306 = "IDX21306: The 'c_hash' claim was not a string in the 'id_token', but a 'code' was in the OpenIdConnectMessage, 'id_token': '{0}'.";
         internal const string IDX21307 = "IDX21307: The 'c_hash' claim was not found in the id_token, but a 'code' was in the OpenIdConnectMessage, id_token: '{0}'";
-        internal const string IDX21308 = "IDX21308: 'Azp' claim exists in the 'id_token' but 'ciient_id' is null. Cannot validate the 'azp' claim.";
+        internal const string IDX21308 = "IDX21308: 'Azp' claim exists in the 'id_token' but 'client_id' is null. Cannot validate the 'azp' claim.";
```
jennyf19: >Azp [](http://example.com/codeflow?start=53&length=3)

nit: should this be `azp`?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
jennyf19: I think we said this wasn't needed and info is captured elsewhere now.

test/Microsoft.IdentityModel.Logging.Tests/LoggerTests.cs
```diff
@@ -237,7 +238,39 @@ public void PrepareMessageWithNoArguments()
...
+
+            var verboseMessage = "Test verbose Message. {0}";
+            LogHelper.LogVerbose(verboseMessage, LogHelper.MarkAsNonPII(arg));
+
+            Assert.True(logger.ContainsLogOfSpecificLevel(string.Format(verboseMessage, arg), EventLogLevel.Verbose));
+
+            var warnMessage = "Warn Message. {0}";
+            LogHelper.LogWarning(warnMessage, LogHelper.MarkAsNonPII(arg));
+
+            Assert.True(logger.ContainsLogOfSpecificLevel(string.Format(warnMessage, arg), EventLogLevel.Warning));
```
jennyf19, line 266: could you use [Theory] and [InlineData()] to capture some of the variables so you can use one Assert.True() for all the log messages & have less code duplication.

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/LogMessages.cs
TimHannMSFT: We have an error starting with a lowercase quoted literal in IDX21335 and per spec the claim names should be case sensitive so makes sense to change since we're here

https://www.rfc-editor.org/rfc/rfc7519#section-10.1.1

PR #1862, opened 2022-05-27
Introducing IdentityLoggerAdapter that writes to ILogger

### LINKED ISSUE #1847: [Feature Request] Integrate IIdentityLogger with Id.Web for AspNetCore to see Wilson logs
Scope:
AspNetCore only

Create a class that implements IIdentityLogger to publish Wilson logs to AspNetCore's ILogger. This is so that AspNetCore can onboard and will no longer have to write a EventListener to read Wilson logs.

Proposal is [here](https://microsoft-my.sharepoint-df.com/:w:/p/srangavajhul/ETms1FahTcpPhjmXSGmb89YBfFTQEkL0sEYKgr-78x6AhA?e=JTXW3i)

sruke:
What is changing?
Introduced a new assembly Microsoft.IdentityModel.LoggingExtensions and a new class IdentityLoggerAdapter that implements IIdentityLogger interface and writes to Microsoft.Extensions.Logging.ILogger. 

What does it mean for customers?
This will allow Identity Libraries like Id.Web and other SDKs to use the logger adapter out of the box and receive Wilson logs by setting -
LogHelper.Logger = new IdentityLoggerAdapter(myILoggerInstance);

An event listener will no longer be needed. 

NOTE: myILoggerInstance must be of the type Micosoft.Extensions.Logging.ILogger


src/Microsoft.IdentityModel.LoggingExtensions/Microsoft.IdentityModel.LoggingExtensions.csproj
```diff
@@ -0,0 +1,34 @@
...
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <AssemblyName>Microsoft.IdentityModel.LoggingExtensions</AssemblyName>
+    <Description>A package containing logging extensions to support writing Identity logs to Microsoft.Extensions.Logging.ILogger.</Description>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageId>Microsoft.IdentityModel.LoggingExtensions</PackageId>
+    <PackageTags>.NET;Windows;Authentication;Identity;Extensions;Logging</PackageTags>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
sruke, line 11: As the assembly only targets netstandard2.0 I was unable to create a test project due to compatibility issues.
* xunit.running.visualstudio doesn't support netstandard2.0
* Adding additional targets in test projects break compatibility with LoggingExtensions assembly as it only supports netstandard2.0

TimHannMSFT:
> What is changing? Introduced a new assembly Microsoft.IdentityModel.LoggingExtensions and a new class IdentityLoggerAdapter that implements IIdentityLogger interface and writes to Microsoft.Extensions.Logging.ILogger.
> 
> What does it mean for customers? This will allow Identity Libraries like Id.Web and other SDKs to use the logger adapter out of the box and receive Wilson logs by setting - LogHelper.Logger = new IdentityLoggerAdapter(myILoggerInstance);
> 
> An event listener will no longer be needed.
> 
> NOTE: myILoggerInstance must be of the type Micosoft.Extensions.Logging.ILogger

I like this format for PR notes, what's changing, what does it mean for people consuming the library.

TimHannMSFT:
> [HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\StrongName]

Looks like lots of special characters were added here. Was the update done in a simple text editor? Might want to try base notepad or notepadd++ I'm not sure what impact the extra characters would have

---
Refers to: build/strongNameBypass.reg:5 in 6b925ec. [](commit_id = 6b925ecc3778a7bf4e60833f157b6096fba8f80b, deletion_comment = False)

src/Microsoft.IdentityModel.LoggingExtensions/IdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,116 @@
...
+    public class IdentityLoggerAdapter : IIdentityLogger
+    {
+        private readonly ILogger _logger;
+
+        /// <summary>
+        /// Instantiates <see cref="IdentityLoggerAdapter"/> using <paramref name="logger"/>.
+        /// </summary>
+        /// <param name="logger">An <see cref="ILogger"/> instance to which identity log messages are written.</param>
+
+#pragma warning disable CS3001 // Argument type is not CLS-compliant
```
TimHannMSFT, line 46: For consistency, let's use the GlobalSuppressions pattern and target it to just the types you need.

test/Microsoft.IdentityModel.Logging.Tests/LoggerTests.cs
```diff
@@ -265,5 +267,67 @@ public void EventLevelToEventLogLevelMapping()
...
+            get
+            {
+                var theoryData = new TheoryData<LoggerTheoryData>();
+
+                theoryData.Add(new LoggerTheoryData
+                {
+                    TestId = "NullLoggerNoMessage",
+                    Logger = null,
+                    ShouldMessageBeLogged = false
+                });
```
TimHannMSFT: Is this test actually doing anything? It just skips over the if statement and passes correct? Seems like this only exercises test code.

TimHannMSFT:
LGTM with a couple comments, remember you'll have to update the signing task for this new package once you're ready for builds with signing.

TimHannMSFT (approved):
:shipit:

build/dependencies.props
```diff
@@ -1,5 +1,6 @@
 <Project>
   <PropertyGroup>
+    <AspNetCoreMinSupportedVersion>2.1.1</AspNetCoreMinSupportedVersion>
```
jennyf19, line 3: >2.1.1< [](http://example.com/codeflow?start=35&length=6)

Why this version?

sruke:
> > �[�H�K�E�Y�_�L�O�C�A�L�_�M�A�C�H�I�N�E�\�S�O�F�T�W�A�R�E�\�M�i�c�r�o�s�o�f�t�\�S�t�r�o�n�g�N�a�m�e�]�
> 
> Looks like lots of special characters were added here. Was the update done in a simple text editor? Might want to try base notepad or notepadd++ I'm not sure what impact the extra characters would have
> 
> Refers to: build/strongNameBypass.reg:5 in [6b925ec](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/6b925ecc3778a7bf4e60833f157b6096fba8f80b). [](commit_id = [6b925ec](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/6b925ecc3778a7bf4e60833f157b6096fba8f80b), deletion_comment = False)

The special characters could be appearing only in CodeFlow. They aren't seen when viewing the file in Notepad or in GitHub.

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.LoggingExtensions/IdentityLoggerAdapter.cs
sruke: I attempted to use the Global Suppressions pattern, but it isn't working for this particular type. 
Intellisense doesn't auto recommend suppressing the error by adding to the Gobal Suppressions file.
The Code Analyzer doesn't recognize it either when I add the error to the suppressions file manually.

test/Microsoft.IdentityModel.Logging.Tests/LoggerTests.cs
sruke: True, it doesn't add a value. Removed it from the test data. It was left there to show that no message can be written if logger is set to null.

build/dependencies.props
sruke: Due to compatibility issues

PR #1864, opened 2022-06-02
LKG with AadIssuerValidator

ciaozhang:
Enable LKG feature when IssuerValidator is AadIssuerValidator:
1) use the effective authority to retrieve metadata to make sure the atomic
2) introduce the internal flag ValidateWithLKG to validate the token with LKG configuration


src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -187,12 +195,40 @@ public static AadIssuerValidator GetAadIssuerValidator(string aadAuthority)
             return GetAadIssuerValidator(aadAuthority, null);
         }
 
+        /// <summary>
+        /// Get the effective authority based on the token version.
+        /// </summary>
+        /// <param name="authority">The authority to query for metadata.</param>
+        /// <param name="securityToken">Received security token.</param>
+        /// <returns></returns>
+        public static string GetEffectiveAuthorityFromToken(string authority, SecurityToken securityToken)
```
ciaozhang: Not sure here is the best place for this method.  #Closed

dannybtsai:
> 

A value from a Dictionary can be null (tid & tenantId).

---
Refers to: src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs:293 in 6e3a1f3. [](commit_id = 6e3a1f332ba54db8c9d1a8eb0e950ba50fb49f8b, deletion_comment = False)

dannybtsai:
>                 if (jwtSecurityToken.Payload.TryGetValue(AadIssuerValidatorConstants.Tid, out object tid))

Payload can be null according to the comment.

---
In reply to: [1145635332](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1145635332) [](http://example.com/codeflow?ancestors=1145635332)

---
In reply to: [1145635332](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1145635332) [](http://example.com/codeflow?ancestors=1145635332)

---
Refers to: src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs:288 in 6e3a1f3. [](commit_id = 6e3a1f332ba54db8c9d1a8eb0e950ba50fb49f8b, deletion_comment = False)

dannybtsai:
>         private static string GetTenantIdFromIss(string iss)

nit: rename to GetTenantIdFromIssuer

---
In reply to: [1145636626](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1145636626) [](http://example.com/codeflow?ancestors=1145636626)

---
In reply to: [1145636626](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1145636626) [](http://example.com/codeflow?ancestors=1145636626)

---
Refers to: src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs:318 in 6e3a1f3. [](commit_id = 6e3a1f332ba54db8c9d1a8eb0e950ba50fb49f8b, deletion_comment = False)

dannybtsai:
>             if (validIssuerTemplate.Contains("{tenantid}"))

define a constant for {tenantid} as it appears multiple times

---
In reply to: [1146157331](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1146157331) [](http://example.com/codeflow?ancestors=1146157331)

---
In reply to: [1146157331](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1146157331) [](http://example.com/codeflow?ancestors=1146157331)

---
In reply to: [1146157331](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1146157331) [](http://example.com/codeflow?ancestors=1146157331)

---
Refers to: src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs:259 in 6e3a1f3. [](commit_id = 6e3a1f332ba54db8c9d1a8eb0e950ba50fb49f8b, deletion_comment = False)

dannybtsai:
>         }

Just a suggestion that the try/catch is not needed here (as we are not passing null or empty argument), and the if block can be simplified to oneline:

return validIssuerTemplate.Replace("{tenantid}", tenantId) == actualIssuer;

See [https://stackoverflow.com/questions/34404042/should-i-use-string-contains-before-string-replace](https://stackoverflow.com/questions/34404042/should-i-use-string-contains-before-string-replace) for calling Contains() and Replace().

---
In reply to: [1146194851](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1146194851) [](http://example.com/codeflow?ancestors=1146194851)

---
In reply to: [1146194851](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1146194851) [](http://example.com/codeflow?ancestors=1146194851)

---
Refers to: src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs:278 in 6e3a1f3. [](commit_id = 6e3a1f332ba54db8c9d1a8eb0e950ba50fb49f8b, deletion_comment = False)

dannybtsai:
>                 AadAuthorityV1 = CreateV1Authority(AadAuthorityV2);

move this line (and the one in the else) to before the if statement

---
In reply to: [1146205895](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1146205895) [](http://example.com/codeflow?ancestors=1146205895)

---
In reply to: [1146205895](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1146205895) [](http://example.com/codeflow?ancestors=1146205895)

---
In reply to: [1146205895](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1146205895) [](http://example.com/codeflow?ancestors=1146205895)

---
Refers to: src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs:59 in 6e3a1f3. [](commit_id = 6e3a1f332ba54db8c9d1a8eb0e950ba50fb49f8b, deletion_comment = False)

dannybtsai:
>             }

1. I think GetEffectiveAuthorityFromToken()  should use AadAuthorityV1 and AadAuthorityV2 instead of constructing its own authority string. This guarantees the correctness of effective authority because these two properties are verified in unit tests. But they don't need to be set in the constructor as only one will actually be used. Consider changing to something like the following:

```
        internal AadIssuerValidator(
            HttpClient httpClient,
            string aadAuthority)
        {
            HttpClient = httpClient;
            AadAuthority = aadAuthority.TrimEnd('/');
            IsV2Authority = AadAuthority.Contains(V2EndpointSuffix);
        }

        private string _aadAuthorityV1 = null;
        internal string AadAuthorityV1
        {
            get
            {
                if (_aadAuthorityV1 == null)
                    _aadAuthorityV1 = IsV2Authority ? CreateV1Authority(AadAuthorityV2) : AadAuthority;

                return _aadAuthorityV1;
            }
        }

        private string _aadAuthorityV2 = null;
        internal string AadAuthorityV2
        {
            get
            {
                if (_aadAuthorityV2 == null)
                    _aadAuthorityV2 = IsV2Authority ? AadAuthority : AadAuthority + V2EndpointSuffix;

                return _aadAuthorityV2;
            }
        }
```

2. IsV2Authority, AadAuthorityV1 and AadAuthorityV2 should be readonly, no setters.

---
In reply to: [1146216389](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1146216389) [](http://example.com/codeflow?ancestors=1146216389)

---
In reply to: [1146216389](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1146216389) [](http://example.com/codeflow?ancestors=1146216389)

---
Refers to: src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs:60 in 6e3a1f3. [](commit_id = 6e3a1f332ba54db8c9d1a8eb0e950ba50fb49f8b, deletion_comment = False)

ciaozhang:
[Build](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=931442&view=logs&j=92a4755a-6173-591e-5c07-8dbe1165e937&t=a41535f0-83e8-520a-008a-bdfc6edd6174&l=539) failed:
System.IO.FileLoadException : Could not load file or assembly 'Microsoft.IdentityModel.Validators, Version=0.0.1.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. Strong name signature could not be verified.  The assembly may have been tampered with, or it was delay signed but not fully signed with the correct private key. (Exception from HRESULT: 0x80131045)

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
dannybtsai: This needs to be changed to non-static if it is going to reference AadAuthorityV1 and AadAuthorityV2. Since there are unit tests verifying the values of AadAuthorityV1 and AadAuthorityV2, this method should use them instead of setting effectiveAuthority separately. Otherwise the unit tests should be modified. So this is the right place based on the initial implementation, IMO.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1192,7 +1193,10 @@ private async Task<TokenValidationResult> ValidateTokenAsync(string token, JsonW
 
                         // Only try to re-validate using the newly obtained config if it doesn't reference equal the previously used configuration.
                         if (lastConfig != currentConfiguration)
-                            return decryptedJwt != null ? ValidateJWE(outerToken, decryptedJwt, validationParameters, currentConfiguration) : ValidateJWS(token, validationParameters, currentConfiguration); ;
+                        {
+                            validationParameters.ValidateIssuerWithLKG = false;
```
ciaozhang: Might need rename this flag to ValidateWithLKG since we might need this flag for [Design a signaling mechanism where LKG will signal the application layer when used.](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1923689) #Resolved

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -120,22 +124,26 @@ public class AadIssuerValidator
...
-                if (securityToken.Issuer.EndsWith("v2.0", StringComparison.OrdinalIgnoreCase))
+                string AadIssuer;
+                if (validationParameters.ValidateIssuerWithLKG)
                 {
-                    if (AadIssuerV2 == null)
-                        AadIssuerV2 = CreateConfigManager(AadAuthorityV2).GetConfigurationAsync().ConfigureAwait(false).GetAwaiter().GetResult().Issuer;
-
-                    if (IsValidIssuer(AadIssuerV2, tenantId, issuer))
-                        return issuer;
+                    AadIssuer = validationParameters.ConfigurationManager.LastKnownGoodConfiguration.Issuer;
```
brentschmaltz: We can't change the logic here to use the ConfigurationManager in TokenValidationParameters. The issuer is obtained the first time a token is validated. The logic is to use the v1 or v2 endpoint and then adjust for common, which has a '{tenant}' value in the 'issuer' string.

It may make sense to have LKG and Refresh logic in AadIssuerValidator.
The flag you added to TokenValidationParameters.ValidateIssuerWithLKG, would signal to AadIssuerValidator to use its LKG value.
We will also need a way to instruct AadIssuerValidator to refresh, this should be throttled just as ConfigurationManager is.
Currently AadIssuerValidator does not have a cache that limits entries so caching ConfigurationManagers would need to use the LRUCache.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
RojaEnnam: This method need not be public.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -120,22 +124,26 @@ public class AadIssuerValidator
...
                 }
                 else
                 {
-                    if (AadIssuerV1 == null)
-                        AadIssuerV1 = CreateConfigManager(AadAuthorityV1).GetConfigurationAsync().ConfigureAwait(false).GetAwaiter().GetResult().Issuer;
-
-                    if (IsValidIssuer(AadIssuerV1, tenantId, issuer))
-                        return issuer;
+                    if (validationParameters.ConfigurationManager != null)
```
RojaEnnam: We cannot use validationParameters.ConfigurationManager because the authority will be the one that user provided instead of v1/v2. so, we might not get the metadata we need to validate current token. That is why we were creating a configmanger for v1/v2 authorities.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
RojaEnnam: LKG also need to be for both v1 and v2. Thinking what if GetConfigurationAsync gets both the v1 and v2 config's every time and we can have additional api's Getconfigv1 and Getconfigv2 that call GetConfigurationAsync internally and return respective config. That way throttling refresh is achieved by GetConfigurationAsync. 

Each AadIssuerValidator can have only one ConfigManager, as it is per authority correct?

If validator does LKG and refresh logic, like IsRecoverableException in JsonWebHandler, what signals the validator?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1192,7 +1193,10 @@ private async Task<TokenValidationResult> ValidateTokenAsync(string token, JsonW
```
dannybtsai, line 1229: Since you are at it, can you change this to use await instead of GetAwaiter (same as the call above)? The method can throw an exception so a try/catch is needed here.Because this is used multiple times, consider wrap it in a helper method so it is consistent.

dannybtsai:
>             }

See my comments in JsonWebTokenHandler.

---
Refers to: src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs:908 in 6e3a1f3. [](commit_id = 6e3a1f332ba54db8c9d1a8eb0e950ba50fb49f8b, deletion_comment = False)

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -943,8 +944,11 @@ private ClaimsPrincipal ValidateToken(string token, JwtSecurityToken outerToken,
```
dannybtsai, line 947: See the comment in JsonWebTokenHandler.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
dannybtsai: Does it make sense to change the logic of using the v1 or v2 endpoint in SAL to start with? SAL should be the one that decides the V1/V2, not Wilson.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
RojaEnnam: This is being used by SAL. Makes sense to be public.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
RojaEnnam: If ConfigurationManager has effective authority, this makes sense.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
RojaEnnam: AadIssuerValidator can be used independently without SAL as well.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -120,22 +160,26 @@ public class AadIssuerValidator
...
+                    if (IsV2Authority)
+                        configurationManager = ConfigurationManagerV2;
+                    else
+                        configurationManager = ConfigurationManagerV1;
                 }
+
+                if (validationParameters.ValidateIssuerWithLKG)
+                    AadIssuer = configurationManager.LastKnownGoodConfiguration.Issuer;
+                else
+                    AadIssuer = configurationManager.GetBaseConfigurationAsync(CancellationToken.None).GetAwaiter().GetResult().Issuer;
```
RojaEnnam: Should we do this only when AadIssuer is null?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -120,22 +160,26 @@ public class AadIssuerValidator
...
-                    if (IsValidIssuer(AadIssuerV1, tenantId, issuer))
-                        return issuer;
+                    if (IsV2Authority)
+                        configurationManager = ConfigurationManagerV2;
+                    else
+                        configurationManager = ConfigurationManagerV1;
                 }
+
+                if (validationParameters.ValidateIssuerWithLKG)
+                    AadIssuer = configurationManager.LastKnownGoodConfiguration.Issuer;
```
RojaEnnam: This LKG can be v1 or v2, it would fail if the LKG is v1 and current token is v2 and viceversa.

ciaozhang: it will check the token type and then decide to use configurationManagerv1 or configurationManagerv2

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
RojaEnnam: After thinking more about it, it makes sense for AadIssuerValidator to have LKGv1/v2 of its own. If issuer is invalid it could try respective LKG first and later try respective configmanagerv1/v2 and pull in fresh config.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
brentschmaltz: I am not seeing the need for this method.
The AadIssuerValidator cannot depend on external runtimes performing logic.
It must run independently, the flag on TokenValidationParameters that it is OK to use LKG, needs to be honored.

ciaozhang:
>         }

removed the try catch bolck

---
In reply to: [1146194851](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1864#issuecomment-1146194851) [](http://example.com/codeflow?ancestors=1146194851,1146194851)

---
Refers to: src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs:278 in 6e3a1f3. [](commit_id = 6e3a1f332ba54db8c9d1a8eb0e950ba50fb49f8b, deletion_comment = False)

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -120,21 +171,21 @@ public class AadIssuerValidator
...
-                    if (AadIssuerV1 == null)
-                        AadIssuerV1 = CreateConfigManager(AadAuthorityV1).GetConfigurationAsync().ConfigureAwait(false).GetAwaiter().GetResult().Issuer;
-
-                    if (IsValidIssuer(AadIssuerV1, tenantId, issuer))
-                        return issuer;
+                    aadIssuer = effectiveConfigurationManager.LastKnownGoodConfiguration.Issuer;
+                    if (IsValidIssuer(aadIssuer, tenantId, issuer))
+                        return aadIssuer;
+                    else
+                        effectiveConfigurationManager.RequestRefresh();
```
ciaozhang: RequestRefresh here if validate with LKG failed #Resolved

RojaEnnam: Why not try GetConfiguration again after refresh, like in JsonWebTokenHandler?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -120,21 +171,21 @@ public class AadIssuerValidator
...
-                        AadIssuerV2 = CreateConfigManager(AadAuthorityV2).GetConfigurationAsync().ConfigureAwait(false).GetAwaiter().GetResult().Issuer;
-
-                    if (IsValidIssuer(AadIssuerV2, tenantId, issuer))
-                        return issuer;
+                    // todo: set LKG only token is valid
+                    effectiveConfigurationManager.LastKnownGoodConfiguration = new OpenIdConnectConfiguration() { Issuer = aadIssuer };
+                    return issuer;
                 }
-                else
+                else if (validationParameters.ValidateWithLKG && effectiveConfigurationManager.LastKnownGoodConfiguration != null)
```
RojaEnnam: Why not check for LKG validity?

RojaEnnam: The configmanager based on which ValidateWithLKG is set is different from what's created here, we don't need this flag anymore correct?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
RojaEnnam: No longer relevant to current code.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
ciaozhang: I request refresh the metadata here if validate with LKG failed and the handler will validate token with refreshed metadata for the third try.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
ciaozhang: AadIssuerValidator's configurationManager(v1 or v2) is synced with tvp.ConfigurationManager by validateWithLKG, they share same LKG lifetime. We already checked the LKG validity in handler.

RojaEnnam: Since they are separate instances, I don't understand how they are synced.

RojaEnnam: Do we want to support LKG only when coming through JsonWebHandler or when AadIssuerValidator is used separately as well? 

Either way to ensure LKG is not stale we should also check validationParameters.ConfigurationManager.IsLastKnownGoodValid and perhaps also check if LKG==currentconfig to skip LKG altogether. IMO we shouldn't rely on things JsonWebTokenHandler is doing correctly.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -723,13 +723,23 @@ public string RoleClaimType
...
 
+        /// <summary>
+        /// Gets or sets a boolean to control if the issuer will be validated with LKG configuration during token validation.
+        /// </summary>
+        /// <remarks>
+        /// This boolean only applies to when <see cref="IssuerValidator"/> is AadIssuerValidator".
+        /// The default is <c>false</c>.
+        /// </remarks>
+        [DefaultValue(false)]
+        internal bool ValidateWithLKG { get; set; }
```
brentschmaltz: For 3p's to access this will need to be public.

test/Microsoft.IdentityModel.TestUtils/Default.cs
```diff
@@ -1239,6 +1259,11 @@ public static TokenValidationParameters JWECompressionTokenValidationParameters
             }
         }
 #endif
+        public static string Tid
```
brentschmaltz: This does not quite make sense.
The Default class is for the value of claims, not the type of the claim.

In AadPayloadString just use "tid' as the claim name and return "tenantId' here.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -120,21 +171,21 @@ public class AadIssuerValidator
 
             try
             {
-                if (securityToken.Issuer.EndsWith("v2.0", StringComparison.OrdinalIgnoreCase))
+                var effectiveConfigurationManager = GetEffectiveConfigurationManager();
```
brentschmaltz: It would be easier to understand if you just wrote 
var effectiveConfigurationManager = IsV2Authority ? ConfigurationManagerV2 : ConfigurationManagerV1; #Pending

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
ciaozhang: Changed it to public

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
ciaozhang: changed to check iss from token rahter than the isV2Authority

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
ciaozhang: Added RefeshBeforeValidating flag

test/Microsoft.IdentityModel.TestUtils/Default.cs
ciaozhang: removed the Tid

jennyf19 (approved):
:shipit:

RojaEnnam (approved):

PR #1866, opened 2022-06-06
Support for ECDH-ES

brentschmaltz:
Added support for ECDH-ES key exchange for ECDsa keys.


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -889,13 +896,14 @@ private static string EncryptTokenPrivate(string innerJwt, EncryptingCredentials
...
             byte[] wrappedKey = null;
-            SecurityKey securityKey = JwtTokenUtilities.GetSecurityKey(encryptingCredentials, cryptoProviderFactory, out wrappedKey);
+            SecurityKey securityKey = JwtTokenUtilities.GetSecurityKey(encryptingCredentials, cryptoProviderFactory, additionalHeaderClaims, out wrappedKey);
 
             using (var encryptionProvider = cryptoProviderFactory.CreateAuthenticatedEncryptionProvider(securityKey, encryptingCredentials.Enc))
             {
                 if (encryptionProvider == null)
                     throw LogHelper.LogExceptionMessage(new SecurityTokenEncryptionFailedException(LogMessages.IDX14103));
 
+                // we will have to make changes to creteing the JWE header and add the epk value as per https://datatracker.ietf.org/doc/html/rfc7518#section-4.1 and see example here: https://datatracker.ietf.org/doc/html/rfc7518#appendix-C
```
cesarpayan: Do we still need this comment? I believe we moved away from considering this change when we used the additional header claims above to create the KDF and CEK.

cesarpayan (approved):
Left and a question on a comment. Other than that, it looks good to me.

PR #1868, opened 2022-06-07
Make 'cty' claim optional

### LINKED ISSUE #1861: The "cty" header is always added breaking compatibility with certain services
6.18

**Where is the issue?**
 * [x] M.IM.JsonWebTokens

The app is in production and I have upgraded to a new version of Microsoft.IdentityModel.

**Expected behavior**
The `cty` header addition for `JsonWebTokenHandler.CreateToken` is controlled by some flag.

**Actual behavior**
The `cty` header is added unconditionally.

**Possible solution**
Since #1588, there seems to be no way to control the addition of the `cty` header, and there are services (for example Apple ID token service) that require the produced token to contain only specified headers. The header set works just fine in the versions up to and including 6.17. After #1588 was completed, there appears to be no way to control the `cty` header presence.

RojaEnnam:


brentschmaltz:
@RojaEnnam do we need this for JwtSecurityTokenHandler as well?

RojaEnnam:
Yes we need it for JwtSecurityTokenHandler as well but JwtHeader has 'cty' as a public property, not entirely sure how to address it. Thought we can go ahead with JsonWebTokenHandler as we know how to fix.

brentschmaltz:
What is the effect if SecurityTokenDescriptor.SetDefaultCtyClaim = true with the JwtSecurityTokenHandler?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -600,7 +605,8 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
             if (encryptingCredentials == null && additionalHeaderClaims != null && additionalHeaderClaims.Count > 0)
                 header.Merge(JObject.FromObject(additionalHeaderClaims));
 
-            header.Merge(JObject.FromObject(AddCtyClaimDefaultValue(additionalInnerHeaderClaims)));
+            if (additionalInnerHeaderClaims != null && additionalInnerHeaderClaims.Count > 0)
```
RojaEnnam, line 606: Similar to line 605, should we check if encryptcredentials is null and add inner headers if only so.

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
```diff
@@ -119,5 +119,13 @@ public class SecurityTokenDescriptor
...
         /// </summary>
         public ClaimsIdentity Subject { get; set; }
+
+        /// <summary>
+        /// Gets or sets a bool that controls if token creation will set default 'cty' if not specified.
+        /// <remarks>
+        /// Applies to only JWT tokens.
+        /// </remarks>
+        /// </summary>
+        public bool SetDefaultCtyClaim { get; set; } = true;
```
brentschmaltz: Since the 'Cty' header claim is related to Encryption, if we put the property on EncryptingCredentials, we may not need the additional public overloads that pass the bool: 'setDefaultCtyClaim'.

RojaEnnam: Makes sense. Moved it to EncryptingCredentials.

brentschmaltz (approved):

PR #1870, opened 2022-06-08
[Bugfix] fix casing for parameter "DateTimeProperties" => "dateTimeProperties"

TimHannMSFT:
parameters should always start with a lowercase letter then camel case to match existing code
  and expectations for .NET libraries.

### DIFF
--- a/src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
+++ b/src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
@@ -63,7 +63,7 @@ namespace Microsoft.IdentityModel.Abstractions
         /// <param name="stringProperties">Key value pair of strings to long with the event.</param>
         /// <param name="longProperties">Key value pair of longs to long with the event.</param>
         /// <param name="boolProperties">Key value pair of bools to long with the event.</param>
-        /// <param name="DateTimeProperties">Key value pair of DateTimes to long with the event.</param>
+        /// <param name="dateTimeProperties">Key value pair of DateTimes to long with the event.</param>
         /// <param name="doubleProperties">Key value pair of doubles to long with the event.</param>
         /// <param name="guidProperties">Key value pair of Guids to long with the event.</param>
         void TrackEvent(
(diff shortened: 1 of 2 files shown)


TimHannMSFT:
Want to clean up this interface before we publicly ship and are locked into this. I think this was based on a copy-paste error I missed.

src/Microsoft.IdentityModel.Abstractions/ITelemetryClient.cs
```diff
@@ -63,15 +63,15 @@ public interface ITelemetryClient
...
         /// <param name="doubleProperties">Key value pair of doubles to long with the event.</param>
         /// <param name="guidProperties">Key value pair of Guids to long with the event.</param>
         void TrackEvent(
             string eventName,
             IDictionary<string, string> stringProperties = null,
             IDictionary<string, long> longProperties = null,
             IDictionary<string, bool> boolProperties = null,
-            IDictionary<string, DateTime> DateTimeProperties = null,
+            IDictionary<string, DateTime> dateTimeProperties = null,
             IDictionary<string, double> doubleProperties = null,
```
sruke, line 75: [src/Microsoft.IdentityModel.Abstractions/NullTelemetryClient.cs](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/f3ddfe0dcb40ec5bde3e324b4246984753a0e121/src/Microsoft.IdentityModel.Abstractions/NullTelemetryClient.cs) would also need the same change.

TimHannMSFT: Good catch thanks Sruthi

TimHannMSFT:
> Want to clean up this interface before we publicly ship and are locked into this. I think this was based on a copy-paste error I missed.

Forgot that this isn't actually a breaking change to fix since the compiler only cares about method signature types. Still best to fix.

sruke (approved):
LGTM!

dannybtsai (approved):

PR #1874, opened 2022-06-10
fixed a test failure because the event queue task needs more time to …

dannybtsai:
This change is to increase the wait time in the CaompactCache() unit test to ensure that the event queue thread can be started to perform the cache compaction action.

### DIFF
--- a/test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
+++ b/test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
@@ -105,10 +105,8 @@ namespace Microsoft.IdentityModel.Tokens.Tests
             AddItemsToCache(cache, 3 * capacity, expiredInSeconds);
 
             // allows the event queue task to be started to compact the cache
-            Thread.Sleep(100);
-
-            // wait until the event queue is empty
             cache.WaitForProcessing();
+            Thread.Sleep(waitInMiliSeconds);
 
             // the number of cached items should be below the capacity after compaction
             if (cache.MapCount > capacity)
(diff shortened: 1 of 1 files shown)


test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
```diff
@@ -105,10 +105,8 @@ public void CompactCache()
             AddItemsToCache(cache, 3 * capacity, expiredInSeconds);
 
             // allows the event queue task to be started to compact the cache
-            Thread.Sleep(100);
-
-            // wait until the event queue is empty
             cache.WaitForProcessing();
+            Thread.Sleep(waitInMiliSeconds);
```
TimHannMSFT: These tests in general have an awful lot of Thread.Sleep calls

Is there something that prevents us from hooking into the LRU cache and getting an event based hook or at least an override for expiry time? Sleeps of 2 seconds here, 5 seconds in the above test really start to add up, especially if we're running tests serially.

dannybtsai: That is a valid point, while it is not straightforward to predict a thread action, it is a bad practice to hardcode a wait time because it increases the build time. This should be fixed now.

test/Microsoft.IdentityModel.Tokens.Tests/EventBasedLRUCacheTests.cs
```diff
@@ -104,13 +104,17 @@ public void CompactCache()
...
 
-            // allows the event queue task to be started to compact the cache
-            Thread.Sleep(100);
-
-            // wait until the event queue is empty
+            // allows the compaction event to be processed
             cache.WaitForProcessing();
 
-            // the number of cached items should be below the capacity after compaction
+            // keep checking the cache size, up to 1 min, until it is reduced
```
TimHannMSFT, line 110: "1 min" should be "1 second" right .1 sec * 10?

TimHannMSFT (approved):

PR #1879, opened 2022-06-15
Improve logging for tokens

### LINKED ISSUE #1880: [Bug] Signature shouldn't be logged along with token

sruke:
Tokens shouldn't be logged with signatures when they are in encoded or encrypted formats.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -331,7 +331,6 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// <param name="rawData">the original token.</param>
         private void Decode(string[] tokenParts, string rawData)
         {
-            LogHelper.LogInformation(LogMessages.IDX14106, rawData);
             try
             {
                 Header = JObject.Parse(Base64UrlEncoder.Decode(tokenParts[0]));
--- a/src/Microsoft.IdentityModel.JsonWebTokens/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/LogMessages.cs
@@ -46,7 +46,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         internal const string IDX14102 = "IDX14102: Unable to decode the header '{0}' as Base64Url encoded string. jwtEncodedString: '{1}'.";
         internal const string IDX14103 = "IDX14103: Failed to create the token encryption provider.";
         internal const string IDX14105 = "IDX14105: Header.Cty != null, assuming JWS. Cty: '{0}'.";
-        internal const string IDX14106 = "IDX14106: Decoding token: '{0}' into header, payload and signature.";
+        // internal const string IDX14106 = "IDX14106:";
         internal const string IDX14107 = "IDX14107: Token string does not match the token formats: JWE (header.encryptedKey.iv.ciphertext.tag) or JWS (header.payload.signature)";
         internal const string IDX14111 = "IDX14111: JWT: '{0}' must have three segments (JWS) or five segments (JWE).";
         internal const string IDX14112 = "IDX14112: Only a single 'Actor' is supported. Found second claim of type: '{0}', value: '{1}'";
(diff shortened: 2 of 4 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -331,7 +331,6 @@ public override SecurityKey SigningKey
         /// <param name="rawData">the original token.</param>
         private void Decode(string[] tokenParts, string rawData)
         {
-            LogHelper.LogInformation(LogMessages.IDX14106, rawData);
```
sruke, line 334: If Decode(..) method is unable to decode/deserialize the header or payload then exceptions with rawData are logged [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/9170aa6db120342ab65c9ea7e045b291e8fef908/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L340:~:text=throw%20LogHelper.LogExceptionMessage(new%20ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14102%2C%20tokenParts%5B0%5D%2C%20rawData)%2C%20ex))%3B) and [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/9170aa6db120342ab65c9ea7e045b291e8fef908/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L340:~:text=throw%20LogHelper.LogExceptionMessage(new%20ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14101%2C%20tokenParts%5B1%5D%2C%20EncodedToken)%2C%20ex))%3B).

sruke: For DecodeJWS(..), is there value in logging the encoded header and payload in [DecodeJws(string[] tokenParts)](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/9170aa6db120342ab65c9ea7e045b291e8fef908/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#:~:text=private%20void%20DecodeJws(string%5B%5D%20tokenParts))?

sruke: For DecodeJWE(..), according to the JWE documentation the token is encrypted so there is no value in logging the token. Is my understanding correct?

src/System.IdentityModel.Tokens.Jwt/LogMessages.cs
```diff
@@ -50,7 +50,6 @@ internal static class LogMessages
         internal const string IDX12713 = "IDX12713: Creating actor value using actor.BootstrapContext(as string)";
         internal const string IDX12714 = "IDX12714: Creating actor value using actor.BootstrapContext.rawData";
         internal const string IDX12715 = "IDX12715: Creating actor value by writing the JwtSecurityToken created from actor.BootstrapContext";
```
jennyf19, line 52: I've seen something, but maybe it was just in MISE, where we comment out the removed log message, instead of deleting it, to have a record of it in the code. At least the #'s used.

jennyf19 (approved):
LGTM, but would like to have @brentschmaltz  or @GeoK review as well.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
TimHannMSFT: This seems largely right to me. On the other hand logging the encrypted token seems less risky since presumably someone would have to have to key to access the PII we're attempting to protect. Seems plausible there might be some cases where the full token doesn't decrypt but the header (which I understand has to be unencrypted) would still provide diagnostically useful info.

src/Microsoft.IdentityModel.JsonWebTokens/LogMessages.cs
```diff
@@ -46,7 +46,6 @@ internal static class LogMessages
         internal const string IDX14102 = "IDX14102: Unable to decode the header '{0}' as Base64Url encoded string. jwtEncodedString: '{1}'.";
         internal const string IDX14103 = "IDX14103: Failed to create the token encryption provider.";
         internal const string IDX14105 = "IDX14105: Header.Cty != null, assuming JWS. Cty: '{0}'.";
-        internal const string IDX14106 = "IDX14106: Decoding token: '{0}' into header, payload and signature.";
```
TimHannMSFT, line 49: >IDX14106 [](http://example.com/codeflow?start=30&length=8)

Like @jennyf19 mentioned. Best practice is to comment out. (e.g. see below: // internal const string IDX14115 = "IDX14115:";)

This ensures that there's no confusion with future logging/documentation where something might give out of date instructions based on error Id (and we're not very limited in integers to use :D)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
TimHannMSFT: Ah, looks like we log on error (ln340) which seems good enough to my mind.

TimHannMSFT (approved):
:shipit:

PR #1885, opened 2022-06-17
update release to yaml

jennyf19:
Build pipeline can be found [here](https://identitydivision.visualstudio.com/IDDP/_build?definitionId=1733&_a=summary).

TODO:
- Uncomment the symbol publishing task
- Use dev branch instead of this branch
- Try to condense the build to fit under 1 hr time limit in ADO (try now it is timing out)

### DIFF
--- a/build/releaseBuild.yml
+++ b/build/releaseBuild.yml
@@ -1,0 +1,223 @@
+trigger: none
+pr: none
+
+name: $(TeamProject)_$(Build.DefinitionName)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Rev:.r)
+
+# Create a daily midnight build for release builds on master to ensure our release builds function
+schedules:
+- cron: "0 0 * * *"
+  displayName: Daily midnight build
+  branches:
+    include:
+    - dev
+
+variables:
+  - group: Middleware
+  - name: BuildPlatform
+    value: 'any cpu'
+  - name: BuildConfiguration
+    value: 'Release'
+
+jobs:
+- job: build
+  pool:
+    name: MwWilson1EsHostedPool
+    demands:
+    - msbuild
+    - DotNetFramework
+    - visualstudio
+  timeoutInMinutes: 360
+
+  steps:
+  - task: UseDotNet@2
+    displayName: 'Use .NET Core sdk 2.1.818'
+    inputs:
+      version: 2.1.818
+      installationPath: 'c:\Program Files\dotnet'
+
+  - task: UseDotNet@2
+    displayName: 'Use .NET Core sdk 2.1.x'
+    inputs:
+      version: 2.1.x
+      installationPath: $(Agent.ToolsDirectory)/dotnet
+
+  - task: UseDotNet@2
+    displayName: 'Use .Net Core SDK 6'
+    inputs:
+      version: 6.0.x
+
+  - task: DotNetCoreCLI@2
+    displayName: 'dotnet --list-sdks '
+    inputs:
+      command: custom
+      custom: '--list-sdks '
+
+  - task: CmdLine@1
+    displayName: 'Run VerifyResourceUsage.pl'
+    inputs:
+      filename: perl
+      arguments: '$(Agent.BuildDirectory)\s\src\VerifyResourceUsage.pl'
+
+  - powershell: |
+      regedit /s .\build\strongNameBypass.reg
+      regedit /s .\build\strongNameBypass2.reg
+    displayName: 'Strong Name Bypass'
+
+  - task: PowerShell@2
+    displayName: 'Update Assembly Info'
+    inputs:
+      targetType: filePath
+      filePath: ./updateAssemblyInfo.ps1
+      arguments: '-packageType $(BuildConfiguration)'
+
+  - task: PowerShell@2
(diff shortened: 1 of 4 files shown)


build/releaseBuild.yml
```diff
@@ -0,0 +1,227 @@
...
+             "toolName": "sign",
+             "toolVersion": "1.0"
+         }
+     ]
+    SessionTimeout: 20
+
+- task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
+  displayName: 'Component Detection'
+
+# - task: PublishSymbols@2
```
jennyf19: I will uncomment this task section before merging to dev.

build/releaseBuild.yml
```diff
@@ -0,0 +1,227 @@
...
+    optionsXS: 0
+    optionsHMENABLE: 0
+
+- task: securedevelopmentteam.vss-secure-development-tools.build-task-credscan.CredScan@2
+  displayName: 'Run CredScan'
+  inputs:
+    suppressionsFile: 'build/credscan-exclusion.json'
+    debugMode: false
+
+- task: securedevelopmentteam.vss-secure-development-tools.build-task-fxcop.FxCop@2
```
brentschmaltz: We can get rid of this task.

jennyf19: removed.

jmprieur (approved):
LGTM
Thanks @jennyf19

build/releaseBuild.yml
```diff
@@ -0,0 +1,223 @@
...
+  pool:
+    name: MwWilson1EsHostedPool
+    demands:
+    - msbuild
+    - DotNetFramework
+    - visualstudio
+  timeoutInMinutes: 360
+
+  steps:
+  - task: UseDotNet@2
```
cesarpayan, line 32: Out of curiosity: Why do we need two use .Net Core sdk 2.1? I see one as .818 and the other as .x.

build/releaseBuild.yml
```diff
@@ -0,0 +1,223 @@
...
+      version: 2.1.818
+      installationPath: 'c:\Program Files\dotnet'
+
+  - task: UseDotNet@2
+    displayName: 'Use .NET Core sdk 2.1.x'
+    inputs:
+      version: 2.1.x
+      installationPath: $(Agent.ToolsDirectory)/dotnet
+
+  - task: UseDotNet@2
```
cesarpayan, line 44: Is installationPath not needed on this use .net core?

build/releaseBuild.yml
```diff
@@ -0,0 +1,223 @@
...
+        -adoFeedName: $(AdoFeedName)
+        -adoAllowPrerelease: "false"
+        -apiCompatRoot: $(ApiCompatRootWilson)
+        -nugetPackageProviderVersion: "2.8.5.208"
+
+  - task: MSBuild@1
+    displayName: Build
+    inputs:
+      solution: wilson.sln
+      msbuildVersion: '17.0.0'
```
cesarpayan, line 92: Question: any specific reason to not use latest instead of version 17?

build/releaseBuild.yml
```diff
@@ -0,0 +1,223 @@
...
+  - group: Middleware
+  - name: BuildPlatform
+    value: 'any cpu'
+  - name: BuildConfiguration
+    value: 'Release'
+
+jobs:
+- job: build
+  pool:
+    name: MwWilson1EsHostedPool
```
cesarpayan, line 24: I ran into an issue before because the host was an older version of Windows, and I am trying to figure out how that is set in here. Are the agent host specifications in MwWilson1EsHostedPool?

build/releaseBuild.yml
jennyf19: I was getting some build issues on one of the tests, requiring .net core sdk 2.x, and it wasn't picking up the .818 version. It still uses the .818 version but it's now picked up by those tests.

build/releaseBuild.yml
jennyf19: No specific reason, this is what was given by the build when the yaml was created.

build/releaseBuild.yml
jennyf19: Yes, it's specified in this new 1ES pool that @jmprieur created.

build/releaseBuild.yml
jennyf19: SDK 2.x is required by the compliance tooling.

build/releaseBuild.yml
jennyf19: This is what we use in Id. Web, so I went with this path and it resolved the build issue.

cesarpayan (approved):

RojaEnnam:
@jennyf19 if we need to make changes to the pipeline in future, do we need to update this yaml file and publish PR?
(Right now we just edit the pipeline task and save it). Thanks for this change and also appreciate adding steps to set this up in our DevEx one note.

RojaEnnam (approved):

jennyf19:
> if we need to make changes to the pipeline in future, do we need to update this yaml file

Yes, that's correct. You can edit the yaml directly in ADO also.

PR #1890, opened 2022-06-24
update newtonsoft per governance report

jennyf19:


jmprieur (approved):
LGTM
This means that Wilson is not using the right KeyVault SDK, though. It should use Azure.Identity (not the old Azure SDK that is now deprecated).

@brentschmaltz: are these assemblies used?

TimHannMSFT (approved):

jennyf19:
> LGTM This means that Wilson is not using the right KeyVault SDK, though. It should use Azure.Identity (not the old Azure SDK that is now deprecated).
> 
> @brentschmaltz: are these assemblies used?

I did not want to spend more time on it, so did more of a quick fix.

jmprieur:
> I did not want to spend more time on it, so did more of a quick fix.

That makes sense @jennyf19: Moving to the latest Azure SDK leveraging Azure.Identity would be much bigger work and probably a breaking change

PR #1895, opened 2022-06-28
change method to protected virtual

### LINKED ISSUE #1894: Change to protected virutal method ClaimsIdentity CreateclaimsIdentity(JsonWebToken, TokenValidationVarameters, Issuer)
Change to protected virtual method ClaimsIdentity CreateclaimsIdentity(JsonWebToken, TokenValidationVarameters, Issuer) to allow SAL to override method so that it can do claims mapping.

Why?
Microsoft.IdentityModel has two assemblies to manipulate JWT tokens:

System.IdentityModel.Tokens.Jwt, which is the legacy assembly. It defines [JwtSecurityTokenHandler](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs) class to manipulate JWT tokens.
Microsoft.IdentityModel.JsonWebTokens, which defines the JsonWebToken class and [JsonWebTokenHandler](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs), more modern, and more efficient.
When using JwtSecurityTokenHandler, the short named claims (oid, tid), used to be transformed into the long named claims (with a namespace). With JsonWebTokenHandler this is no longer the case, but when you migrate your application from using JwtSecurityTokenHandler to JsonWebTokenHandler (or use a framework that does), you will only get original claims sent by the IdP. This is more efficient, and occupies less space, but might trigger a lot of changes in your application. In order to make it easier for people to migrate without changing their app too much, this PR offers extensibility to re-add the claims mapping.

cesarpayan:
Change to protected virtual method ClaimsIdentity CreateclaimsIdentity(JsonWebToken, TokenValidationVarameters, Issuer).

**Why?**
Microsoft.IdentityModel has two assemblies to manipulate JWT tokens:
- System.IdentityModel.Tokens.Jwt, which is the legacy assembly. It defines [JwtSecurityTokenHandler](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs) class to manipulate JWT tokens.
- Microsoft.IdentityModel.JsonWebTokens, which defines the JsonWebToken class and [JsonWebTokenHandler](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs), more modern, and more efficient.

When using `JwtSecurityTokenHandler`, the short named claims (oid, tid), used to be transformed into the long named claims (with a namespace). With JsonWebTokenHandler this is no longer the case, but when you migrate your application from using JwtSecurityTokenHandler to JsonWebTokenHandler (or use a framework that does), you will only get original claims sent by the IdP. This is more efficient, and occupies less space, but might trigger a lot of changes in your application. In order to make it easier for people to migrate without changing their app too much, this PR offers extensibility to re-add the claims mapping.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -654,7 +654,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         private static StringComparison GetStringComparisonRuleIf509OrECDsa(SecurityKey securityKey) => (securityKey is X509SecurityKey
                             || securityKey is ECDsaSecurityKey)
                             ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
-        
+
         /// <summary>
         /// Creates a <see cref="ClaimsIdentity"/> from a <see cref="JsonWebToken"/>.
         /// </summary>
@@ -663,38 +663,43 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// <returns>A <see cref="ClaimsIdentity"/> containing the <see cref="JsonWebToken.Claims"/>.</returns>
         protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, TokenValidationParameters validationParameters)
         {
-            if (jwtToken == null)
-                throw LogHelper.LogArgumentNullException(nameof(jwtToken));
+            _ = jwtToken ?? throw LogHelper.LogArgumentNullException(nameof(jwtToken));
 
-            if (validationParameters == null)
-                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
+            return CreateClaimsIdentityPrivate(jwtToken, validationParameters, GetActualIssuer(jwtToken));
+        }
 
-            var actualIssuer = jwtToken.Issuer;
+        /// <summary>
+        /// Creates a <see cref="ClaimsIdentity"/> from a <see cref="JsonWebToken"/> with the specified issuer.
+        /// </summary>
+        /// <param name="jwtToken">The <see cref="JsonWebToken"/> to use as a <see cref="Claim"/> source.</param>
+        /// <param name="validationParameters">Contains parameters for validating the token.</param>
+        /// <param name="issuer">Specifies the issuer for the <see cref="ClaimsIdentity"/>.</param>
+        /// <returns>A <see cref="ClaimsIdentity"/> containing the <see cref="JsonWebToken.Claims"/>.</returns>
+        protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, TokenValidationParameters validationParameters, string issuer)
+        {
+            _ = jwtToken ?? throw LogHelper.LogArgumentNullException(nameof(jwtToken));
+
+            if (string.IsNullOrWhiteSpace(issuer))
+                issuer = GetActualIssuer(jwtToken);
+
+            return CreateClaimsIdentityPrivate(jwtToken, validationParameters, issuer);
+        }
+
+        private static string GetActualIssuer(JsonWebToken jwtToken)
+        {
+            string actualIssuer = jwtToken.Issuer;
             if (string.IsNullOrWhiteSpace(actualIssuer))
             {
                 LogHelper.LogVerbose(TokenLogMessages.IDX10244, ClaimsIdentity.DefaultIssuer);
                 actualIssuer = ClaimsIdentity.DefaultIssuer;
             }
 
-            return CreateClaimsIdentity(jwtToken, validationParameters, actualIssuer);
+            return actualIssuer;
         }
 
-        /// <summary>
-        /// 
-        /// </summary>
-        /// <param name="jwtToken"></param>
-        /// <param name="validationParameters"></param>
-        /// <param name="issuer"></param>
-        /// <returns></returns>
-        protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, TokenValidationParameters validationParameters, string issuer)
+        private ClaimsIdentity CreateClaimsIdentityPrivate(JsonWebToken jwtToken, TokenValidationParameters validationParameters, string issuer)
         {
-
-            if (jwtToken == null)
-                throw LogHelper.LogArgumentNullException(nameof(jwtToken));
-
-            if (validationParameters == null)
-                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
-
+            _ = validationParameters ?? throw LogHelper.LogArgumentNullException(nameof(validationParameters));
 
             ClaimsIdentity identity = validationParameters.CreateClaimsIdentity(jwtToken, issuer);
             foreach (Claim jwtClaim in jwtToken.Claims)
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -679,9 +679,24 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             return CreateClaimsIdentity(jwtToken, validationParameters, actualIssuer);
         }
 
-        private ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, TokenValidationParameters validationParameters, string actualIssuer)
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="jwtToken"></param>
+        /// <param name="validationParameters"></param>
+        /// <param name="issuer"></param>
+        /// <returns></returns>
+        protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, TokenValidationParameters validationParameters, string issuer)
         {
-            ClaimsIdentity identity = validationParameters.CreateClaimsIdentity(jwtToken, actualIssuer);
+
+            if (jwtToken == null)
+                throw LogHelper.LogArgumentNullException(nameof(jwtToken));
+
+            if (validationParameters == null)
+                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
+
+
+            ClaimsIdentity identity = validationParameters.CreateClaimsIdentity(jwtToken, issuer);
             foreach (Claim jwtClaim in jwtToken.Claims)
             {
                 string claimType = jwtClaim.Type;
@@ -693,17 +708,17 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     if (CanReadToken(jwtClaim.Value))
                     {
                         JsonWebToken actor = ReadToken(jwtClaim.Value) as JsonWebToken;
-                        identity.Actor = CreateClaimsIdentity(actor, validationParameters, actualIssuer);
+                        identity.Actor = CreateClaimsIdentity(actor, validationParameters, issuer);
                     }
                 }
 
                 if (jwtClaim.Properties.Count == 0)
                 {
-                    identity.AddClaim(new Claim(claimType, jwtClaim.Value, jwtClaim.ValueType, actualIssuer, actualIssuer, identity));
+                    identity.AddClaim(new Claim(claimType, jwtClaim.Value, jwtClaim.ValueType, issuer, issuer, identity));
                 }
                 else
                 {
-                    Claim claim = new Claim(claimType, jwtClaim.Value, jwtClaim.ValueType, actualIssuer, actualIssuer, identity);
+                    Claim claim = new Claim(claimType, jwtClaim.Value, jwtClaim.ValueType, issuer, issuer, identity);
 
                     foreach (var kv in jwtClaim.Properties)
                         claim.Properties[kv.Key] = kv.Value;
--- a/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
+++ b/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
@@ -55,11 +55,30 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
             Assert.Throws<ArgumentNullException>(() => handler.CreateToken("Payload", Default.SymmetricEncryptingCredentials, (Dictionary<string, object>) null));
         }
 
-        [Fact]
-        public void ValidateTokenValidationResult()
+        [Theory, MemberData(nameof(TokenValidationClaimsTheoryData))]
+        public void ValidateTokenValidationResult(JsonWebTokenTheoryData theoryData)
         {
             TestUtilities.WriteHeader($"{this}.ValidateTokenValidationResult");
+            var tokenValidationResult = theoryData.TokenHandler.ValidateToken(theoryData.AccessToken, theoryData.ValidationParameters);
+            Assert.Equal(tokenValidationResult.Claims, TokenUtilities.CreateDictionaryFromClaims(tokenValidationResult.ClaimsIdentity.Claims));
+        }
 
+        [Theory, MemberData(nameof(TokenValidationClaimsTheoryData))]
+        public void ValidateTokenDerivedHandlerValidationResult(JsonWebTokenTheoryData theoryData)
+        {
+            TestUtilities.WriteHeader($"{this}.ValidateTokenDerivedHandlerValidationResult");
+            var derivedJsonWebTokenHandler = new DerivedJsonWebTokenHandler(); 
+            var tokenValidationResult = theoryData.TokenHandler.ValidateToken(theoryData.AccessToken, theoryData.ValidationParameters);
+            var tokenValidationDerivedResult = derivedJsonWebTokenHandler.ValidateToken(theoryData.AccessToken, theoryData.ValidationParameters);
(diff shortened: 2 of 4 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -703,9 +703,24 @@ protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, Tok
             return CreateClaimsIdentity(jwtToken, validationParameters, actualIssuer);
         }
 
-        private ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, TokenValidationParameters validationParameters, string actualIssuer)
+        /// <summary>
+        /// 
+        /// </summary>
```
jennyf19, line 673: add the summary from 683, but add how this one is different (includes issuer)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -703,9 +703,24 @@ protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, Tok
...
+        /// <param name="jwtToken"></param>
+        /// <param name="validationParameters"></param>
```
jennyf19: all the xml comments need to be filled out. You can copy the ones above

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -703,9 +703,24 @@ protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, Tok
...
+        /// <param name="issuer"></param>
+        /// <returns></returns>
```
jennyf19: same for return value.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -703,9 +703,24 @@ protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, Tok
...
+        protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, TokenValidationParameters validationParameters, string issuer)
```
jennyf19, line 678: >string issuer) [](http://example.com/codeflow?start=133&length=14)

nit: if issuer can be null, then you could add it to the original method at the end. We were discussing last week if this is a breaking change or not. We do this a lot in ID web to keep API surface smaller and helps with testing.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -703,9 +703,24 @@ protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, Tok
...
-            ClaimsIdentity identity = validationParameters.CreateClaimsIdentity(jwtToken, actualIssuer);
+
+            if (jwtToken == null)
+                throw LogHelper.LogArgumentNullException(nameof(jwtToken));
+
+            if (validationParameters == null)
+                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
+
+
+            ClaimsIdentity identity = validationParameters.CreateClaimsIdentity(jwtToken, issuer);
```
jennyf19, line 704: >issuer) [](http://example.com/codeflow?start=90&length=7)

Will CreateClaimsIdentity handle a null issuer?

jennyf19:
Any tests to update or add?

jennyf19 (changes requested):
:clock1:

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: It doesn't throw, but I think it leads to unwanted scenarios (leaving the issuer null in the claims). I'm adding a null or empty check. @brentschmaltz is there a reason to not have it? I can't find one.
Alternatively, we could have a check similar to the other method protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, TokenValidationParameters validationParameters) or just merge the methods together as Jenny mentioend above and have issuer as a default parameter to avoid breaking changes.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: Leaving this pending, on the question below around whether or not issuer could be null.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: Added in a3b8ef8

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: Added in a3b8ef8

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: Added in a3b8ef8

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -79,11 +79,30 @@ public void CreateTokenThrowsNullArgumentException()
...
         {
             TestUtilities.WriteHeader($"{this}.ValidateTokenValidationResult");
+            var tokenValidationResult = theoryData.TokenHandler.ValidateToken(theoryData.AccessToken, theoryData.ValidationParameters);
+            Assert.Equal(tokenValidationResult.Claims, TokenUtilities.CreateDictionaryFromClaims(tokenValidationResult.ClaimsIdentity.Claims));
+        }
+
+        [Theory, MemberData(nameof(TokenValidationClaimsTheoryData))]
+        public void ValidateTokenDerivedHandlerValidationResult(JsonWebTokenTheoryData theoryData)
+        {
+            TestUtilities.WriteHeader($"{this}.ValidateTokenValidationResult");
```
TimHannMSFT: >ValidateTokenValidationResult [](http://example.com/codeflow?start=47&length=29)

nit: ValidateTokenDerivedHandlerValidationResult

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -79,11 +79,30 @@ public void CreateTokenThrowsNullArgumentException()
...
+            var derivedJsonWebTokenHandler = new DerivedJsonWebTokenHandler(); 
+            var tokenValidationResult = theoryData.TokenHandler.ValidateToken(theoryData.AccessToken, theoryData.ValidationParameters);
+            var tokenValidationDerivedResult = derivedJsonWebTokenHandler.ValidateToken(theoryData.AccessToken, theoryData.ValidationParameters);
+            Assert.Equal(tokenValidationResult.Claims, TokenUtilities.CreateDictionaryFromClaims(tokenValidationResult.ClaimsIdentity.Claims));
+            Assert.Equal(tokenValidationDerivedResult.Claims, TokenUtilities.CreateDictionaryFromClaims(tokenValidationDerivedResult.ClaimsIdentity.Claims));
+            Assert.Equal(tokenValidationDerivedResult.Claims, tokenValidationDerivedResult.Claims);
```
TimHannMSFT: >Assert.Equal(tokenValidationDerivedResult.Claims, tokenValidationDerivedResult.Claims); [](http://example.com/codeflow?start=12&length=87)

copy and paste error? Seems like this likely isn't checking what it's supposed to be.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -3434,6 +3473,21 @@ public override AuthenticatedEncryptionResult Encrypt(byte[] plaintext, byte[] a
...
+
+    public class DerivedJsonWebTokenHandler : JsonWebTokenHandler
+    {
+        /// <summary>
+        /// Creates a <see cref="ClaimsIdentity"/> from a <see cref="JsonWebToken"/>.
+        /// </summary>
+        /// <param name="jwtToken">The <see cref="JsonWebToken"/> to use as a <see cref="Claim"/> source.</param>
+        /// <param name="validationParameters">Contains parameters for validating the token.</param>
+        /// <param name="issuer">Specifies the issuer for the <see cref="ClaimsIdentity"/>.</param>
+        /// <returns></returns>
```
TimHannMSFT: >returns [](http://example.com/codeflow?start=13&length=7)

nit: <returns>A <see cref="ClaimsIdentity"/> containing the <see cref="JsonWebToken.Claims"/>.</returns>

TimHannMSFT:
>             }

imo, move this into the newly virtual method then just have this call in, you can remove the TVP validation here too since you'll do it in the called method.

Seems like that would minimize code duplication without breaking anyone who's already derived this.

---
Refers to: src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs:701 in 32a47a6. [](commit_id = 32a47a6d142ec108c5779cfde947f97660b754c0, deletion_comment = False)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
TimHannMSFT: Seems to be we'd just set the default and log like we do in the other method.

TimHannMSFT:
Marking as approved with comments. I think all the right stuff is here, I left some comments on what I recommend you update before merging.

TimHannMSFT (approved):
:shipit:

TimHannMSFT:
Are there any updates to Changelog.md which are needed to express to customers what value this brings or are we mainly intending this to be something for our own consumption? @jmprieur do you have an opinion about how we should express these changes in the release notes if the latter is the case?

jmprieur:
> Are there any updates to Changelog.md which are needed to express to customers what value this brings or are we mainly intending this to be something for our own consumption? @jmprieur do you have an opinion about how we should express these changes in the release notes if the latter is the case?

@TimHannMSFT @cesarpayan 
Here is the customer value:

Microsoft.IdentityModel has two assemblies to manipulate JWT tokens:
- System.IdentityModel.Tokens.Jwt, which is the legacy assembly. It defines [JwtSecurityTokenHandler](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs) class to manipulate JWT tokens.
- Microsoft.IdentityModel.JsonWebTokens, which defines the JsonWebToken class and [JsonWebTokenHandler](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs), more modern, and more efficient.

When using `JwtSecurityTokenHandler`, the short named claims (oid, tid), used to be transformed into the long named claims (with a namespace). With JsonWebTokenHandler this is no longer the case, but when you migrate your application from using JwtSecurityTokenHandler to JsonWebTokenHandler (or use a framework that does), you will only get original claims sent by the IdP. This is more efficient, and occupies less space, but might trigger a lot of changes in your application. In order to make it easier for people to migrate without changing their app too much, this PR offers extensibility to re-add the claims mapping.

jmprieur:
@cesarpayan : why isn't this PR linked to the GitHub issue it fixes?

cesarpayan:
> 

We had the issue in ADO, I'll create an issue with the wilson portion to be able to reference it before I update the PR. Thanks for pointing it out,

cesarpayan:
> @cesarpayan : why isn't this PR linked to the GitHub issue it fixes?

@jmprieur 
We had the issue in ADO, ~~I'll create an issue with the wilson portion to be able to reference it before I update the PR. Thanks for pointing it out,~~ Edit: There's an existing issue already linked in here #1894 which I just updated with the explanation you've given above. Does it still show up as unlinked for you?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: > > issuer)
> 
> Will CreateClaimsIdentity handle a null issuer?

CreateClaimsIdentity handles a null issuer, and we have unit tests that takes that into account. We grab a default value in the other method if it is needed.

> Seems to be we'd just set the default and log like we do in the other method.

I'll mimic the other method's behavior.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: I've refactored a bit the calls and it should be clearer now that it is calling the private method. I 'created' a second virtual instead of changing the original private method that was there with the logic, with both virtual methods calling it.

jennyf19 (approved):
:shipit: Thanks @cesarpayan

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -666,22 +666,47 @@ protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, Tok
...
+        /// <param name="validationParameters">Contains parameters for validating the token.</param>
+        /// <param name="issuer">Specifies the issuer for the <see cref="ClaimsIdentity"/>.</param>
+        /// <returns>A <see cref="ClaimsIdentity"/> containing the <see cref="JsonWebToken.Claims"/>.</returns>
+        protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, TokenValidationParameters validationParameters, string issuer)
+        {
+            if (jwtToken == null)
+                throw LogHelper.LogArgumentNullException(nameof(jwtToken));
+
+            if (string.IsNullOrWhiteSpace(issuer))
+            {
```
brentschmaltz: we don't need brackets for single if statements.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -666,22 +666,47 @@ protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, Tok
...
+        /// <summary>
+        /// Creates a <see cref="ClaimsIdentity"/> from a <see cref="JsonWebToken"/> with the specified issuer.
+        /// </summary>
+        /// <param name="jwtToken">The <see cref="JsonWebToken"/> to use as a <see cref="Claim"/> source.</param>
+        /// <param name="validationParameters">Contains parameters for validating the token.</param>
+        /// <param name="issuer">Specifies the issuer for the <see cref="ClaimsIdentity"/>.</param>
+        /// <returns>A <see cref="ClaimsIdentity"/> containing the <see cref="JsonWebToken.Claims"/>.</returns>
+        protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, TokenValidationParameters validationParameters, string issuer)
+        {
+            if (jwtToken == null)
```
brentschmaltz: sometimes we write this as:

_ = jwtToken ?? throw LogHelper.LogArgumentNullException(nameof(jwtToken));

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: @cesarpayan there has to be an issuer.
If the user has turned off 'validate issuer' and there is no issuer, then the ClaimsIdentity.DefaultIssuer should be used.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: Changed in 5a4c742

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
cesarpayan: Changed in 5a4c742

TimHannMSFT:
Just finished my review as you merged, LGTM :)

---
In reply to: [1196054666](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1895#issuecomment-1196054666) [](http://example.com/codeflow?ancestors=1196054666,1196054666)

PR #1902, opened 2022-07-01
updated to 13.0.2, added pragmas

brentschmaltz:
Snapped to version 13.0.2.
The majority of the changes we #nullable enable as 13.0.2 uses nullable refs.
The other changes were suppressing warnings.

The changes in the source is the result of internal changes.
To see what those are you can compare these changes to the newtonsoft repo.


src/Microsoft.IdentityModel.Tokens/opensource/json/DefaultJsonNameTable.cs
```diff
@@ -77,7 +80,12 @@ public override string Get(char[] key, int start, int length)
             hashCode -= hashCode >> 5;
 
             // make sure index is evaluated before accessing _entries, otherwise potential race condition causing IndexOutOfRangeException
-            var index = hashCode & _mask;
+#if NET20 || NET35 || NET40
```
jennyf19: >NET35 || NET40 [](http://example.com/codeflow?start=13&length=14)

do we need these?

src/Microsoft.IdentityModel.Tokens/opensource/json/JsonTextWriter.Async.cs
```diff
@@ -131,7 +134,34 @@ internal async Task DoCloseAsync(CancellationToken cancellationToken)
...
+        {
+            if (_writeBuffer != null)
+            {
+                BufferUtils.ReturnBuffer(_arrayPool, _writeBuffer);
+                _writeBuffer = null;
+            }
+
+            if (CloseOutput && _writer != null)
+            {
+#if HAVE_ASYNC_DISPOABLE
```
jennyf19: >DISPOABLE [](http://example.com/codeflow?start=15&length=9)

spelling: DISPOSABLE  -> missing an S

src/Microsoft.IdentityModel.Tokens/opensource/json/JsonValidatingReader.cs
```diff
@@ -41,6 +41,8 @@
 
 #endif
 
+#nullable disable
```
jennyf19, line 44: remove?

src/Microsoft.IdentityModel.Tokens/opensource/json/SerializationBinder.cs
```diff
@@ -1,5 +1,5 @@
 ﻿
-#if (DOTNET || PORTABLE40 || PORTABLE || NET_CORE)
+#if (DOTNET || PORTABLE40 || PORTABLE)
```
jennyf19: >PORTABLE40 || PORTABLE [](http://example.com/codeflow?start=15&length=22)

do we need these?

src/Microsoft.IdentityModel.Tokens/opensource/json/Bson/BsonBinaryType.cs
```diff
@@ -25,6 +25,8 @@
 
 using System;
 
+#nullable disable
```
jennyf19, line 28: needed?

src/Microsoft.IdentityModel.Tokens/opensource/json/Bson/BsonBinaryWriter.cs
```diff
@@ -29,6 +29,8 @@
 using System.Text;
 using Microsoft.IdentityModel.Json.Utilities;
 
+#nullable disable
```
jennyf19, line 32: >nullable disable [](http://example.com/codeflow?start=1&length=16)

?

jennyf19:
> using System;

missing license info

---
Refers to: src/Microsoft.IdentityModel.Tokens/opensource/json/Linq/JsonSelectSettings.cs:1 in 6f7d57b. [](commit_id = 6f7d57b334d4e1eccde228fa7501ced8deabc315, deletion_comment = False)

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.Tokens/opensource/json/JsonTextWriter.Async.cs
jennyf19: https://github.com/JamesNK/Newtonsoft.Json/pull/2698

brentschmaltz: I changed it, to be consistent with the repo, but the define in M.IM.Tokens.csproj was also misspelled.
So, i changed them both.

src/Microsoft.IdentityModel.Tokens/opensource/json/SerializationBinder.cs
brentschmaltz: We use NET_CORE in our test code, i don't think it is related to the product code, but i will add it back.
PORTABLE40 and PORTABLE are not defined in our builds, and are in many locations in the newtonsoft source.

src/Microsoft.IdentityModel.Tokens/opensource/json/DefaultJsonNameTable.cs
brentschmaltz: We don't need any of those defines

TimHannMSFT:
> 

This and other files seem to be changed only in adding a trailing new line. Is that necessary?

---
Refers to: src/Microsoft.IdentityModel.Tokens/opensource/json/ConstructorHandling.cs:44 in 9934704. [](commit_id = 993470408518521410b888f972f474d8a43f4b68, deletion_comment = False)

src/Microsoft.IdentityModel.Tokens/opensource/json/DefaultJsonNameTable.cs
```diff
@@ -53,12 +55,13 @@ public DefaultJsonNameTable()
...
         /// <summary>
-        /// Gets the string containing the same characters as the specified range of characters in the given array.
+        /// Gets a string containing the same characters as the specified range of characters in the given array.
         /// </summary>
         /// <param name="key">The character array containing the name to find.</param>
         /// <param name="start">The zero-based index into the array specifying the first character of the name.</param>
         /// <param name="length">The number of characters in the name.</param>
-        public override string Get(char[] key, int start, int length)
+        /// <returns>A string containing the same characters as the specified range of characters in the given array.</returns>
+        public override string? Get(char[] key, int start, int length)
```
TimHannMSFT, line 64: >string? [](http://example.com/codeflow?start=24&length=7)

huh, TIL about nullable reference types... [https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types) #Resolved

src/Microsoft.IdentityModel.Tokens/opensource/json/Bson/BsonBinaryType.cs
TimHannMSFT, line 28: >disable [](http://example.com/codeflow?start=10&length=7)

Do you know why this is required here and elsewhere? This seems to be needed where nullable is enabled by default but then in the /json/ dir, there's #endable tags which make it seem as though it's disabled by default.

TimHannMSFT (approved):
:shipit:

TimHannMSFT:
Reviewed all the files for the first 3rd then spot checked the rest. LGTM

src/Microsoft.IdentityModel.Tokens/opensource/json/Bson/BsonBinaryType.cs
brentschmaltz: That code is in the 13.0.2 branch and not in our last snapshot.

src/Microsoft.IdentityModel.Tokens/opensource/json/DefaultJsonNameTable.cs
brentschmaltz: That is the code in the 13.0.2 branch.

src/Microsoft.IdentityModel.Tokens/opensource/json/DefaultJsonNameTable.cs
dannybtsai: Both are in the System.Threading and apply to all platforms we support.

src/Microsoft.IdentityModel.Tokens/opensource/json/DefaultJsonNameTable.cs
```diff
@@ -119,7 +128,7 @@ public string Add(string key)
             hashCode -= hashCode >> 5;
             for (Entry entry = _entries[hashCode & _mask]; entry != null; entry = entry.Next)
             {
-                if (entry.HashCode == hashCode && entry.Value.Equals(key))
+                if (entry.HashCode == hashCode && entry.Value.Equals(key, StringComparison.Ordinal))
```
dannybtsai, line 127: >StringComparison.Ordinal [](http://example.com/codeflow?start=74&length=24)

The should not be necessary, the default is StringComparison.Ordinal already.

src/Microsoft.IdentityModel.Tokens/opensource/json/DefaultJsonNameTable.cs
```diff
@@ -158,7 +167,12 @@ private void Grow()
                 }
             }
             _entries = newEntries;
-            _mask = newMask;
+
+#if NET20 || NET35 || NET40
```
dannybtsai, line 167: Like read, we probably don't need this, maybe just use Volatile.Write()?

PR #1915, opened 2022-07-24
update license info to be more consice, as the longer license info is…

jennyf19:
… not longer required when a license file is present.

@brentschmaltz I think it was thrown around to update the license headers, lmk when is a good time to merge so it's less invasive.

### DIFF
--- a/test/Microsoft.IdentityModel.TestUtils/TestStub.cs
+++ b/test/Microsoft.IdentityModel.TestUtils/TestStub.cs
@@ -1,5 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.-------------------
+// Licensed under the MIT License.
 
 using System;
 using Xunit;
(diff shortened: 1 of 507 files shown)


test/Microsoft.IdentityModel.TestUtils/TestStub.cs
```diff
@@ -1,8 +1,5 @@
-//-------------------------------------------------------------------------------------------------
```
jmprieur, line 1: Interesting that it didn't have the same copyright format than the others :-)

jmprieur (approved):
LGTM
Thanks @jennyf19.

FYI, I've done a smoke test by running: 
`git diff jennyf/updateLicense..dev --numstat > diff.txt`

and verified that each file had exactly 26 lines replaced by 2 lines.
This was the case for all except test/Microsoft.IdentityModel.TestUtils/TestStub.cs.

All is good.

test/Microsoft.IdentityModel.TestUtils/TestStub.cs
jennyf19: Yes. lol.

PR #1916, opened 2022-07-26
Update readme and STO tag info

jennyf19:


jmprieur (approved):
LGTM
Thanks @jennyf19 

Very good catch!

PR #1931, opened 2022-08-22
Use TVP.ConfigManager in AadIssuerValidator

RojaEnnam:
Use TVP.ConfigManager in AadIssuerValidator to support static configuration.


RojaEnnam:
Moved it to https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1932 for changing target branch.

PR #1932, opened 2022-08-22
Use TVP.ConfigManager in AadIssuerValidator

RojaEnnam:
Use TVP.ConfigManager in AadIssuerValidator to support static configuration


src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -261,8 +261,16 @@ private static bool IsValidIssuer(string validIssuerTemplate, string tenantId, s
...
+            if (validationParameters.ConfigurationManager != null)
+            {
+                if (IsV2Authority)
+                    ConfigurationManagerV2 = validationParameters.ConfigurationManager;
+                else
+                    ConfigurationManagerV1 = validationParameters.ConfigurationManager;
+            }
+
             return (securityToken.Issuer.EndsWith(V2EndpointSuffix, StringComparison.OrdinalIgnoreCase)) ? ConfigurationManagerV2 : ConfigurationManagerV1;
         }
```
dannybtsai, line 275: The logic seems a bit confusing and the result could be incorrect.
Line 268 uses IsV2Authority to determine V1 or V2, but line 274 uses Issuer.EndsWith(V2EndpointSuffix).

If IsV2Authority  is true, then ConfigurationManagerV2 = validationParameters.ConfigurationManager.
If Issuer.EndsWith(V2EndpointSuffix) is false, then ConfigurationManagerV1 will be returned, instead of validationParameters.ConfigurationManager.

It seems we should use either IsV2Authority or Issuer.EndsWith(V2EndpointSuffix), not both. And Issuer.EndsWith(V2EndpointSuffix) is probably the one to use.
Maybe do something like this:

            var isV2 = securityToken.Issuer.EndsWith(V2EndpointSuffix, StringComparison.OrdinalIgnoreCase);
            if (validationParameters.ConfigurationManager != null)
            {
                if (isV2)
                    ConfigurationManagerV2 = validationParameters.ConfigurationManager;
                else
                    ConfigurationManagerV1 = validationParameters.ConfigurationManager;

                return validationParameters.ConfigurationManager;
            }
            return isV2 ? ConfigurationManagerV2 : ConfigurationManagerV1;


Or if we want to always use the TVP.ConfigurationManager if it is a StaticConfigurationManager:

            var configManagerType = validationParameters.ConfigurationManager.GetType();
            if (validationParameters.ConfigurationManager != null && configManagerType.IsGenericType && configManagerType.GetGenericTypeDefinition() == typeof(StaticConfigurationManager<>))
                return validationParameters.ConfigurationManager;

             return (securityToken.Issuer.EndsWith(V2EndpointSuffix, StringComparison.OrdinalIgnoreCase)) ? ConfigurationManagerV2 : ConfigurationManagerV1;

RojaEnnam:
Found an alternative solution.

PR #1934, opened 2022-08-24
Use System.Text.Json when processing JsonWebTokens for Net461+

brentschmaltz:
This PR replaces the use of an internal version of Newtonsoft with System.Text.Json for JsonWebTokens.
Validation uses ArrayPool to minimize the memory usage.

A Delegate has been added to TokenValidationParameters to transform a token prior to signature validation.


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1435,31 +1458,27 @@ private static JsonWebToken ValidateSignature(string token, JsonWebToken jwtToke
...
+                        if (ValidateSignature(jwtToken, key, validationParameters))
+#endif
                         {
-                            LogHelper.LogInformation(TokenLogMessages.IDX10242, token);
+                            LogHelper.LogInformation(TokenLogMessages.IDX10242, jwtToken.EncodedToken);
                             jwtToken.SigningKey = key;
                             return jwtToken;
                         }
                     }
+#pragma warning disable CA1031 // Do not catch general exception types
```
TimHannMSFT: >#pragma warning disable CA1031 [](http://example.com/codeflow?start=0&length=30)

nit: Looks like you're suppressing this in GlobalSuppressions and here, pattern seems to be just to do it in Global elsewhere for the library.

TimHannMSFT:
>                     keysAttempted,

this is the public key from config correct? We should be able to log this as nonPII and it would seem to be very helpful for diagnostics.

---
Refers to: src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs:1533 in f9193f8. [](commit_id = f9193f8d350e5a0d324bc7d14dd643b92286a62e, deletion_comment = False)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -67,434 +113,658 @@ public JsonWebToken(string header, string payload)
...
+            Dot1 = encodedJson.IndexOf('.');
+            if (Dot1 == -1)
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14100, encodedJson)));
+
+            if (Dot1 == encodedJson.Length)
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14100, encodedJson)));
+
+            Dot2 = encodedJson.IndexOf('.', Dot1 + 1);
+            if (Dot2 == -1)
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14100, encodedJson)));
```
TimHannMSFT: consider including a hint to differentiate these if PII isn't enabled. Currently we'd need someone to enable PII to differentiate from logs why their token is invalid which could delay investigations.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -67,434 +113,658 @@ public JsonWebToken(string header, string payload)
...
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14307, encodedJson)));
+
+                HeaderAsciiBytes = Encoding.ASCII.GetBytes(_hChars);
+                try
+                {
+                    Header = new JsonClaimSet(Base64UrlEncoder.UnsafeDecode(_hChars));
+                }
+                catch (Exception ex)
+                {
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14102, encodedJson.Substring(0, Dot1), encodedJson), ex));
```
TimHannMSFT, line 524: >throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14102, encodedJson.Substring(0, Dot1), encodedJson), ex)); [](http://example.com/codeflow?start=20&length=157)

Do headers ever include PII? I wouldn't expect them to. Seems like the header can be logged in all cases.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -67,434 +113,658 @@ public JsonWebToken(string header, string payload)
...
+                // header cannot be empty
+                if (_hChars.Length == 0)
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14307, encodedJson)));
+
+                HeaderAsciiBytes = Encoding.ASCII.GetBytes(_hChars);
+                try
+                {
+                    Header = new JsonClaimSet(Base64UrlEncoder.UnsafeDecode(_hChars));
+                }
+                catch (Exception ex)
```
TimHannMSFT, line 522: >Exception [](http://example.com/codeflow?start=23&length=9)

Are there cases where this isn't a FormatException or JsonReaderException? Otherwise it would be more unexpected I would think and may indicate something more serious than a bad argument.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -67,434 +113,658 @@ public JsonWebToken(string header, string payload)
...
+
+                Dot4 = encodedJson.IndexOf('.', Dot3 + 1);
+
+                // JWE needs to have 4 dots
+                if (Dot4 == -1 || Dot4 == encodedJson.Length)
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14100, encodedJson)));
+
+                // too many dots...
+                if (encodedJson.IndexOf('.', Dot4 + 1) != -1)
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14100, encodedJson)));
```
TimHannMSFT: same comment here, would be nice to differentiate from PII scrubbed logs which error case we hit for quicker troubleshooting. At least differentiating between encrypted and non-encrypted would be helpful.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -67,434 +113,658 @@ public JsonWebToken(string header, string payload)
...
+                    Payload = new JsonClaimSet(JwtTokenUtilities.GetJsonDocumentFromBase64UrlEncodedString(encodedJson, Dot1 + 1, Dot2 - Dot1 - 1));
+#endif
+                }
+                catch (Exception ex)
+                {
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14101, encodedJson.Substring(Dot2, Dot2 - Dot1), encodedJson), ex));
+                }
             }
-            catch (Exception ex)
+            else
```
TimHannMSFT, line 483: >else [](http://example.com/codeflow?start=12&length=4)

This implies a JWE right? Might be helpful to have a comment outlining the format so people don't have to leave for the spec

>      BASE64URL(UTF8(JWE Protected Header)) || '.' ||
>      BASE64URL(JWE Encrypted Key) || '.' ||
>      BASE64URL(JWE Initialization Vector) || '.' ||
>      BASE64URL(JWE Ciphertext) || '.' ||
>      BASE64URL(JWE Authentication Tag)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -67,434 +113,658 @@ public JsonWebToken(string header, string payload)
...
+                // JWE needs to have 4 dots
+                if (Dot4 == -1 || Dot4 == encodedJson.Length)
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14100, encodedJson)));
+
+                // too many dots...
+                if (encodedJson.IndexOf('.', Dot4 + 1) != -1)
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14100, encodedJson)));
+
+                // right number of dots for JWE
+                _hChars = encodedJson.ToCharArray(0, Dot1);
```
TimHannMSFT, line 511: >_hChars [](http://example.com/codeflow?start=16&length=7)

Why are _hChars only set #if NET45 for JWS but always for JWE?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -67,434 +113,658 @@ public JsonWebToken(string header, string payload)
...
+                try
+                {
+                    Header = new JsonClaimSet(Base64UrlEncoder.UnsafeDecode(_hChars));
+                }
+                catch (Exception ex)
+                {
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14102, encodedJson.Substring(0, Dot1), encodedJson), ex));
+                }
+
+                // dir does not have any key bytes
```
TimHannMSFT, line 527: >dir [](http://example.com/codeflow?start=19&length=3)

What does dir mean here? Seems synonymous with token/JWE

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -67,434 +113,658 @@ public JsonWebToken(string header, string payload)
...
+                {
+                    EncryptedKeyBytes = Base64UrlEncoder.UnsafeDecode(encryptedKeyBytes);
+                    _encryptedKey = encodedJson.Substring(Dot1 + 1, Dot2 - Dot1 - 1);
+                }
+                else
+                {
+                    _encryptedKey = string.Empty;
+                }
+
+                char[] ivChars = encodedJson.ToCharArray(Dot2 + 1, Dot3 - Dot2 - 1);
```
TimHannMSFT: >ivChars [](http://example.com/codeflow?start=23&length=7)

super nit: ivChars is a rough name since it sort of seems like IV=4 so 4th part, at least that's where my mind first went. something like initVectorChars is only slightly longer and much clearer.

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -329,6 +329,10 @@ public virtual void ValidateAsymmetricSecurityKeySize(SecurityKey key, string al
         /// <remarks>Verify is thread safe.</remarks>
         public override bool Verify(byte[] input, byte[] signature)
         {
+            // The reason this method doesn't call through to: Verify(input, 0, input.Length, signature, 0, signature.Length)
+            // Is because this method's contract is to check the entire signature, if the signature was truncated and signature.Length
+            // was passed, the signature may verify.
```
TimHannMSFT, line 334: Does this imply a truncated signature may pass verification? Wouldn't the odds of that be similar to random signature verifying (i.e. vanishingly small)?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: I will make the change to be consistent and put the suppressions in GolbalSuppressions.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz: @TimHannMSFT this should be PII as the token is logged.
I will make a change to strip off the signature.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz: @TimHannMSFT It's hard to say as the spec doesn't specify.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz: @TimHannMSFT 'dir' refers to the method used to obtain the 'Content Encryption Key' (CEK).
see JWE: https://www.rfc-editor.org/rfc/rfc7516

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz: @TimHannMSFT because we do not need to calculate them for call graphs that use System.Text.Json.
We use ArrayPools that are not available in 4.5.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz: @TimHannMSFT i think in order to understand how this all works, one needs to read the spec.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -165,23 +171,59 @@ internal static string DecompressToken(byte[] tokenBytes, string algorithm)
...
+                            jsonWebToken.CipherTextBytes,
+                            jsonWebToken.HeaderAsciiBytes,
+                            jsonWebToken.InitializationVectorBytes,
+                            jsonWebToken.AuthenticationTagBytes);
+
+                        zipAlgorithm = jsonWebToken.Zip;
+                        decryptionSucceeded = true;
+                        break;
+                    }
+                    // JwtSecurityToken from JwtSecurityTokenHandler
```
dannybtsai, line 203: This is same as the JwtSecurityToken block. Can the properties of decryptionParameters be set in the calling method, JsonWebTokenhandler.DecrytTOken() line 765 so the code does not need to be duplicated (like at JwtSecurityHandler line 1720)?

src/Microsoft.IdentityModel.JsonWebTokens/LogMessages.cs
```diff
@@ -39,11 +38,17 @@ internal static class LogMessages
...
-#pragma warning restore 1591
+        internal const string IDX14305 = "IDX14305: Unable to convert the '{0}' json property to the following type: '{1}'. Property type was: '{2}'. Value: '{3}'.";
+        internal const string IDX14306 = "IDX14306: JWE Ciphertext cannot be an empty string. jwtEncodedString: '{0}'.";
+        internal const string IDX14307 = "IDX14307: JWE header is missing. jwtEncodedString: '{0}'.";
+        internal const string IDX14308 = "IDX14308: JWE initialization vector is missing. jwtEncodedString: '{0}'.";
+        internal const string IDX14309 = "IDX14309: Unable to decode the initialization vector as Base64Url encoded string. jwtEncodedString: '{0}'.";
+        internal const string IDX14310 = "IDX14310: JWE authentication tag is missing. jwtEncodedString: '{0}'.";
+        internal const string IDX14311 = "IDX14311: Unable to decode the authentication tag as Base64Url encoded string. jwtEncodedString: '{0}'.";
+        internal const string IDX14312 = "IDX14312: Unable to decode the cipher text Base64Url encoded string. jwtEncodedString: '{0}'.";
```
RojaEnnam: nit: cipher test "as"?

src/Microsoft.IdentityModel.JsonWebTokens/Properties/AssemblyInfo.cs
```diff
@@ -16,3 +16,7 @@
...
 [assembly: InternalsVisibleTo("System.IdentityModel.Tokens.Jwt, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
+[assembly: InternalsVisibleTo("Microsoft.IdentityModel.S2S, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
RojaEnnam, line 19: For my information, what are these friend assemblies for?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -249,8 +269,76 @@ private unsafe static byte[] UnsafeDecode(string str)
...
         }
+
+        internal static unsafe byte[] UnsafeDecode(char[] str)
+        {
+            int mod = str.Length % 4;
+            if (mod == 1)
+                throw LogHelper.LogExceptionMessage(new FormatException(LogHelper.FormatInvariant(LogMessages.IDX10400, str)));
+
+            bool needReplace = false;
+            int decodedLength = str.Length + (4 - mod) % 4;
```
RojaEnnam, line 281: nit: Consider adding summary for what this method is doing and comments for some of the logic.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -0,0 +1,425 @@
...
+    {
+        private const uint IntA = 'A';
+        private const uint IntZ = 'Z';
+        private const uint Inta = 'a';
+        private const uint Intz = 'z';
+        private const uint Int0 = '0';
+        private const uint Int9 = '9';
+        private const uint IntEq = '=';
+        private const uint IntPlus = '+';
+        private const uint Intminus = '-';
```
RojaEnnam: nit: Capital M in Minus.

src/Microsoft.IdentityModel.Tokens/RsaCryptoServiceProviderProxy.cs
```diff
@@ -195,6 +195,64 @@ public bool VerifyData(byte[] input, object hash, byte[] signature)
             return _rsa.VerifyData(input, hash, signature);
         }
 
+#if NET461 || NET472 || NETSTANDARD2_0
+        /// <summary>
+        /// Verifies that a digital signature is valid by determining the hash value in the signature using the provided public key and comparing it to the hash value of the provided data.
+        /// </summary>
+        /// <param name="input">The input byte array.</param>
+        /// <param name="offset"></param>
```
RojaEnnam, line 203: Need to fill 'offset' and 'length' param tags.

src/Microsoft.IdentityModel.Tokens/SecurityToken.cs
```diff
@@ -10,6 +12,11 @@ namespace Microsoft.IdentityModel.Tokens
     /// </summary>
     public abstract class SecurityToken
     {
+        internal virtual IEnumerable<Claim> CreateClaims(string issuer)
+        {
+            return new List<Claim>();
+        }
```
RojaEnnam, line 18: Does this need to be cleaned up?

src/Microsoft.IdentityModel.Tokens/SignatureProvider.cs
```diff
@@ -103,6 +103,28 @@ internal int Release()
...
         public abstract bool Verify(byte[] input, byte[] signature);
 
+        /// <summary>
+        /// Verifies that a signature created over the 'input' matches the signature. Using <see cref="SecurityKey"/> and 'algorithm' passed to <see cref="SignatureProvider( SecurityKey, string )"/>.
+        /// </summary>
+        /// <param name="input">The bytes to verify.</param>
+        /// <param name="inputOffset">offset in to input bytes to caculate hash.</param>
+        /// <param name="inputLength">number of bytes of signature to use.</param>
+        /// <param name="signature">signature to compare against.</param>
+        /// <param name="signatureOffset"></param>
```
RojaEnnam, line 113: Need to fill param tags.

RojaEnnam:
@brentschmaltz can you please add below friend assembly for Microsoft.IdentityModel.Validators. This will allow access to AadIssuerValidator.ConfigurationManagerV1 property in S2S tests.

[assembly: InternalsVisibleTo("Microsoft.IdentityModel.S2S.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
TimHannMSFT: I was making a slightly different point I think. I agree the encoded json here should be scrubbed for PII, even without the signature the claims will often contain pii so need to be scrubbed.

My point was that if we see IDX14100 in the logs without PII scrubbing off, we have no way of knowing what was wrong with the token formatting. Maybe this is fine but I think it offers some diagnostic value to know that it didn't have any dots vs it was ended with the only period, vs it only had one dot but other text (and likewise, since we use the same error code for JWEs we can't distinguish those either).

It's possible this wouldn't help but I think it's worth the small work to add a suffix since it might save a back and forth with a customer getting logs. to narrow down some error case. Your call though, not something I'd block on.

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -0,0 +1,414 @@
...
+        {
+            if (_claims != null)
+                return _claims;
+
+            _claims = new List<Claim>();
+            foreach (JsonProperty property in RootElement.EnumerateObject())
+            {
+                if (property.Value.ValueKind == JsonValueKind.Array)
+                    foreach (JsonElement jsonElement in property.Value.EnumerateArray())
+                       _claims.Add(CreateClaimFromJsonElement(property.Name, issuer, jsonElement));
```
dannybtsai: >Add [](http://example.com/codeflow?start=31&length=3)

CreateClaimFromJsonElement() can return null, should null be skipped?

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -0,0 +1,414 @@
...
+
+        internal JsonElement RootElement { get; }
+
+        internal IList<Claim> Claims(string issuer)
+        {
+            if (_claims != null)
+                return _claims;
+
+            _claims = new List<Claim>();
+            foreach (JsonProperty property in RootElement.EnumerateObject())
```
dannybtsai: Can this be changed to:

_claims = CreateClaims(string issuer);

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -0,0 +1,414 @@
...
+namespace Microsoft.IdentityModel.JsonWebTokens
+{
+    /// <summary>
+    /// This class provides an abstraction over the json parser for net461+ using System.Text.Json.
+    /// <see cref="JsonWebToken"/> will delegate to this class to get values.
+    /// </summary>
+    internal class JsonClaimSet
+    {
+        private IList<Claim> _claims;
+        private static Type _typeofDateTime = typeof(DateTime);
```
dannybtsai: >_typeofDateTime = typeof(DateTime) [](http://example.com/codeflow?start=28&length=34)

Is this meant to improve performance? Since typeof(DateTIme) is resolved at compile-time, I don't think it will make a difference compared to just using typeof() directly.

TimHannMSFT:
>                         //new KeyValuePair<string, List<object>>("CertificateValidator", new List<object>{(string)null, X509CertificateValidator.None, X509CertificateValidatorEx.None}),

nit: not yours but, remove dead code?

---
Refers to: test/Microsoft.IdentityModel.Tokens.Tests/TokenValidationParametersTests.cs:153 in f9193f8. [](commit_id = f9193f8d350e5a0d324bc7d14dd643b92286a62e, deletion_comment = False)

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -0,0 +1,414 @@
...
+            return string.Empty;
+        }
+
+        internal DateTime GetDateTime(string key)
+        {
+            if (!RootElement.TryGetProperty(key, out JsonElement jsonElement))
+                return DateTime.MinValue;
+
+            return EpochTime.DateTime(Convert.ToInt64(Math.Truncate(Convert.ToDouble(ParseTimeValue(key, jsonElement), CultureInfo.InvariantCulture))));
+        }
```
dannybtsai, line 229: I think this can be simplified to:

return EpochTime.DateTime(ParseTimeValue(key, jsonElement));

It should not need the conversion. If ParseTimeValue() has converted the json to a long why convert the value to double and back?

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -0,0 +1,414 @@
...
+
+                if (jsonElement.TryGetInt32(out int retValInt))
+                    return retValInt;
+
+                if (jsonElement.TryGetDecimal(out decimal retValDecimal))
+                    return (long)retValDecimal;
+            }
+
+            if (jsonElement.ValueKind == JsonValueKind.String)
+            {
```
dannybtsai, line 397: nit: Maybe check for String type first then Number? By default a DateTime is created according to the ISO 8601-1:2019 format, which seems to be string format like "2019-07-26T16:59:57-05:00".

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestValidationParameters.cs
```diff
@@ -178,6 +180,23 @@ public TimeSpan SignedHttpRequestLifetime
...
+                return _tokenHandler;
+            }
+
+            set
+            {
+                _tokenHandler = value ?? throw LogHelper.LogArgumentNullException(nameof(value));
+            }
+        }
+
+
```
ciaozhang: nit: extra line

src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
```diff
@@ -36,7 +60,36 @@ public class TokenValidationResult
...
+                if (_claimsIdentity == null)
+                    _claimsIdentity = CreateClaimsIdentity();
+
+                return _claimsIdentity;
+            }
+            set
+            {
+                _claimsIdentity = value ?? throw LogHelper.LogArgumentNullException(nameof(value));
+            }
+        }
```
dannybtsai, line 76: Could this have threading issue?

src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
```diff
@@ -36,7 +60,36 @@ public class TokenValidationResult
...
+        public ClaimsIdentity ClaimsIdentity
+        {
+            get
+            {
+                if (_claimsIdentity == null)
+                    _claimsIdentity = CreateClaimsIdentity();
+
+                return _claimsIdentity;
+            }
+            set
```
dannybtsai, line 72: >set [](http://example.com/codeflow?start=12&length=3)

So a ClaimsIdentity can be created via CreateClaimsIdentity() and be set directly?

src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
```diff
@@ -242,15 +246,74 @@ public override bool Verify(byte[] input, byte[] signature)
         /// <exception cref="ObjectDisposedException"><see cref="Dispose(bool)"/> has been called.</exception>
         /// <exception cref="InvalidOperationException">If the internal <see cref="KeyedHashAlgorithm"/> is null. This can occur if a derived type deletes it or does not create it.</exception>
         public bool Verify(byte[] input, byte[] signature, int length)
+        {
+            if (signature == null || signature.Length == 0)
+                throw LogHelper.LogArgumentNullException(nameof(signature));
+
+            if (input == null || input.Length == 0)
+                throw LogHelper.LogArgumentNullException(nameof(input));
+
```
dannybtsai, line 252: nit: if the length check has been moved to the other Verify(), these two can be skipped too as all of them are being checked.

src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
TimHannMSFT: This is for back-compat

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -360,6 +364,102 @@ public override bool Verify(byte[] input, byte[] signature)
...
+                    // Having the logic here, handles EC and RSA. We can revisit when we start using spans in 3.1+.
+                    byte[] signatureBytes = new byte[signatureLength];
+                    Array.Copy(signature, 0, signatureBytes, 0, signatureLength);
+                    return asym.Verify(input, inputOffset, inputLength, signatureBytes);
+                }
+            }
+            catch
+            {
+                CryptoProviderCache?.TryRemove(this);
+                Dispose(true);
```
dannybtsai, line 451: Is this needed since the Dispose() method will remove this from the cache?

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -360,6 +364,102 @@ public override bool Verify(byte[] input, byte[] signature)
...
+                        LogHelper.MarkAsNonPII(nameof(signatureLength)),
+                        LogHelper.MarkAsNonPII(nameof(signature)),
+                        LogHelper.MarkAsNonPII(signatureOffset),
+                        LogHelper.MarkAsNonPII(signatureLength),
+                        LogHelper.MarkAsNonPII(signature.Length))));
+
+            if (_disposed)
+            {
+                CryptoProviderCache?.TryRemove(this);
+                throw LogHelper.LogExceptionMessage(new ObjectDisposedException(GetType().ToString()));
```
dannybtsai, line 429: Is this needed since the Dispose() method will remove this from the cache?

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -360,6 +364,102 @@ public override bool Verify(byte[] input, byte[] signature)
...
+            }
+
+            AsymmetricAdapter asym = null;
+            try
+            {
+                asym = _asymmetricAdapterObjectPool.Allocate();
+                if (signature.Length == signatureLength)
```
dannybtsai, line 436: The Alloc() call is inside the try block and will cause the provider to be disposed if an exception occurs. On the other hand, a similar call in the SymmetricSignatureProvider is outside of a try block. Is this the intended behavior?

src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
```diff
@@ -262,7 +325,7 @@ public bool Verify(byte[] input, byte[] signature, int length)
             KeyedHashAlgorithm keyedHashAlgorithm = GetKeyedHashAlgorithm(GetKeyBytes(Key), Algorithm);
             try
```
dannybtsai, line 323: The Alloc() call is inside the try block and will cause the provider to be disposed if an exception occurs. On the other hand, a similar call in the SymmetricSignatureProvider is outside of a try block. Is this the intended behavior?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz: We can address this when we have a model that allows users to examine exceptions.
Given where we are in the stack, the next handler may succeed, so we don't want to throw back to the application.

src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
brentschmaltz: Fixed

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
brentschmaltz: Fixed

RojaEnnam (approved):

PR #1945, opened 2022-10-06
Logging Improvement for issuer/audience validation

### LINKED ISSUE #1922: [Bug] IDX10205 shows up as an error message.

### LINKED ISSUE #1946: [Feature Request] Log errors caused by policy failures as Information or just throw exception without logging

sruke:
An exception is logged as an error message and thrown which can create noise in the logs when multiple policies are being evaluated but only one is applicable for the incoming token validation request. 

To address this, **LogAllPolicyFailuresAsError** property has been introduced in the TokenValidationParameters class and set to true by default to remain backward compatible. 

When this property is set to **false** then exceptions are thrown but not logged and the invoking API will continue to have control over how to respond to the exception.

This change is applicable to issuer validation and audience validation.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
@@ -210,6 +210,7 @@ namespace Microsoft.IdentityModel.Tokens
             ClockSkew = other.ClockSkew;
             ConfigurationManager = other.ConfigurationManager;
             CryptoProviderFactory = other.CryptoProviderFactory;
+            DebugId = other.DebugId;
             IgnoreTrailingSlashWhenValidatingAudience = other.IgnoreTrailingSlashWhenValidatingAudience;
             IssuerSigningKey = other.IssuerSigningKey;
             IssuerSigningKeyResolver = other.IssuerSigningKeyResolver;
@@ -252,6 +253,7 @@ namespace Microsoft.IdentityModel.Tokens
             ValidIssuer = other.ValidIssuer;
             ValidIssuers = other.ValidIssuers;
             ValidTypes = other.ValidTypes;
+            LogAllPolicyFailuresAsError = other.LogAllPolicyFailuresAsError;
         }
 
         /// <summary>
@@ -270,6 +272,7 @@ namespace Microsoft.IdentityModel.Tokens
             ValidateIssuerSigningKey = false;
             ValidateLifetime = true;
             ValidateTokenReplay = false;
+            LogAllPolicyFailuresAsError = true;
         }
 
         /// <summary>
@@ -408,6 +411,11 @@ namespace Microsoft.IdentityModel.Tokens
         /// Users can override the default <see cref="CryptoProviderFactory"/> with this property. This factory will be used for creating signature providers.
         /// </summary>
         public CryptoProviderFactory CryptoProviderFactory { get; set; }
+
+        /// <summary>
+        /// Gets or sets a string that helps with setting breakpoints when debugging.
+        /// </summary>
+        public string DebugId { get; set; }
 
         /// <summary>
         /// Gets or sets a boolean that controls if a '/' is significant at the end of the audience.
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -409,6 +412,11 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
         /// </summary>
         public CryptoProviderFactory CryptoProviderFactory { get; set; }
 
+        /// <summary>
+        /// Gets or sets a string that helps with setting breakpoints when debugging.
+        /// </summary>
+        public string DebugId { get; set; }
```
jmprieur, line 418: Who is the audience of this property? is it for the library maintainers (us?), or also for the customer?
Is it used with Conditional breakpoints.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -825,5 +833,16 @@ public string RoleClaimType
         /// The default is <c>null</c>.
         /// </summary>
         public IEnumerable<string> ValidTypes { get; set; }
+
+        /// <summary>
+        /// Gets or sets a <see cref="bool"/> that will decide if messages are logged as error, when a policy evaluates to <c>false</c> during token validation.
+        /// Default value is <c>true</c>.
```
jmprieur:
```suggestion
        /// Default value is <c>true</c> for backward compatibility of the behavior.
```

jmprieur (commented):
LGTM
I've asked a question.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
TimHannMSFT: mostly for developers, used for conditional breakpoints as I understand @brentschmaltz requested this.

sruke: Right, its for us developers who maintain the library to set conditional breakpoints while debugging across products.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -825,5 +833,16 @@ public string RoleClaimType
...
         /// </summary>
         public IEnumerable<string> ValidTypes { get; set; }
+
+        /// <summary>
+        /// Gets or sets a <see cref="bool"/> that will decide if messages are logged as error, when a policy evaluates to <c>false</c> during token validation.
+        /// Default value is <c>true</c> for backward compatibility of the behavior.
+        /// If set to <c>false</c>, exceptions are logged as Information and then thrown.
+        /// </summary>
+        /// <remarks>
+        /// When multiple polices are defined, all of them are tried until one succeeds and setting this property to false will reduce the noise in the logs.
```
TimHannMSFT, line 843: >false [](http://example.com/codeflow?start=117&length=5)

supernit: <c>false</c>

to match pattern

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -825,5 +833,16 @@ public string RoleClaimType
...
+        /// <summary>
+        /// Gets or sets a <see cref="bool"/> that will decide if messages are logged as error, when a policy evaluates to <c>false</c> during token validation.
+        /// Default value is <c>true</c> for backward compatibility of the behavior.
+        /// If set to <c>false</c>, exceptions are logged as Information and then thrown.
+        /// </summary>
+        /// <remarks>
+        /// When multiple polices are defined, all of them are tried until one succeeds and setting this property to false will reduce the noise in the logs. 
+        /// </remarks>
+        [DefaultValue(true)]
+        public bool VerboseErrorLogs { get; set; }
```
TimHannMSFT: >VerboseErrorLogs [](http://example.com/codeflow?start=20&length=16)

nit: To me this name is a bit ambiguous. Does it mean that the error messages will be more verbose if enabled? Does it mean errors will be logged at the Vebose level?

Consider something like 'LogAllPolicyFailuresAsError'

TimHannMSFT (approved):
:shipit:

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
sruke: Great suggestion, thank you.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
sruke: Updated.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
sruke: Updated.

PR #1952, opened 2022-10-20
Add additional comments to clarify token is not validated.

brentschmaltz:
Added comments to JsonWebTokenHandler and JwtSecurityTokenHandler to emphasize that ReadToken does not validate the token.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -1086,6 +1086,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// <exception cref="ArgumentException">'token.Length' is greater than <see cref="TokenHandler.MaximumTokenSizeInBytes"/>.</exception>
         /// <remarks><para>If the <paramref name="token"/> is in JWE Compact Serialization format, only the protected header will be deserialized.</para>
         /// This method is unable to decrypt the payload. Use <see cref="ValidateToken(string, TokenValidationParameters)"/>to obtain the payload.</remarks>
+        /// <remarks>The token is NOT validated and no security decisions should be made about the contents.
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1086,6 +1086,8 @@ protected virtual SecurityKey ResolveTokenDecryptionKey(string token, JsonWebTok
         /// <exception cref="ArgumentException">'token.Length' is greater than <see cref="TokenHandler.MaximumTokenSizeInBytes"/>.</exception>
         /// <remarks><para>If the <paramref name="token"/> is in JWE Compact Serialization format, only the protected header will be deserialized.</para>
         /// This method is unable to decrypt the payload. Use <see cref="ValidateToken(string, TokenValidationParameters)"/>to obtain the payload.</remarks>
+        /// <remarks>The token is NOT validated and no security decisions should be made about the contents.
```
TimHannMSFT: is it to spec to have multiple remarks method headers? I would think these should be merged into one remark.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -757,6 +757,8 @@ private IEnumerable<Claim> OutboundClaimTypeTransform(IEnumerable<Claim> claims)
         /// <exception cref="ArgumentException"><see cref="CanReadToken(string)"/></exception>
         /// <remarks><para>If the <paramref name="token"/> is in JWE Compact Serialization format, only the protected header will be deserialized.</para>
         /// This method is unable to decrypt the payload. Use <see cref="ValidateToken(string, TokenValidationParameters, out SecurityToken)"/>to obtain the payload.</remarks>
+        /// <remarks>The token is NOT validated and no security decisions should be made about the contents.
```
TimHannMSFT: same note here re multiple remarks blocks.

TimHannMSFT (approved):
LGTM. I'd say it makes sese to clean up the multiple blocks but that's not blocking.

ciaozhang (approved):

PR #1954, opened 2022-10-27
Add Instance property bag that is cleared when Clone is called.

brentschmaltz:
TokenValidationParameters can be used to pass per authentication request information to lower levels.
The pattern is at the top of the call graph to call Clone, which creates an empty InstancePropertyBag.
After Clone is called, users can add to InstancePropertyBag and be assured that the items are associated with the current call.

TokenValidationParameters.IsClone property indicates if Clone was called.
For types derived from TokenValidationParameters set IsClone = true in the Clone method rather than the constructor, otherwise ever new instance will report true for IsClone.

TokenValidationParameters.PropertyBag should be thought of as a container for static data that will be available for all requests.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
@@ -365,7 +365,10 @@ namespace Microsoft.IdentityModel.Tokens
         /// <remarks>This is a shallow Clone.</remarks>
         public virtual TokenValidationParameters Clone()
         {
-            return new TokenValidationParameters(this);
+            return new(this)
+            {
+                IsClone = true
+            };
         }
 
         /// <summary>
@@ -450,6 +453,17 @@ namespace Microsoft.IdentityModel.Tokens
         /// priority.
         /// </remarks>
         public IssuerSigningKeyValidatorUsingConfiguration IssuerSigningKeyValidatorUsingConfiguration { get; set; }
+
+        /// <summary>
+        /// Gets a <see cref="IDictionary{String, Object}"/> that is unique to this instance.
+        /// Calling <see cref="Clone"/> will result in a new instance of this IDictionary.
+        /// </summary>
+        public IDictionary<string, object> InstancePropertyBag { get; } = new Dictionary<string, object>();
+
+        /// <summary>
+        /// Gets a value indicating if <see cref="Clone"/> was called to obtain this instance.
+        /// </summary>
+        public bool IsClone { get; protected set; } = false;
 
         /// <summary>
         /// Gets or sets the <see cref="SecurityKey"/> that is to be used for signature validation.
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -365,7 +365,10 @@ public TimeSpan ClockSkew
         /// <remarks>This is a shallow Clone.</remarks>
         public virtual TokenValidationParameters Clone()
         {
-            return new TokenValidationParameters(this);
+            return new(this)
+            {
+                IsClone = true
```
jennyf19, line 370: nit: add a ';' at the end. Do you not see any errors or warnings in the IDE?

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -451,6 +454,17 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
         /// </remarks>
         public IssuerSigningKeyValidatorUsingConfiguration IssuerSigningKeyValidatorUsingConfiguration { get; set; }
 
+        /// <summary>
+        /// Gets a <see cref="IDictionary{String, Object}"/> that is unique to this instance.
+        /// Calling <see cref="Clone"/> will result in a new instance of this IDictionary.
+        /// </summary>
+        public IDictionary<string, object> InstancePropertyBag { get; } = new Dictionary<string, object>();
```
TimHannMSFT, line 461: >get [](http://example.com/codeflow?start=65&length=3)

What do you think about throwing an exception (or at least logging a warning) if someone calls get when IsClone == false? That's never valid correct?

test/Microsoft.IdentityModel.Tokens.Tests/TokenValidationParametersTests.cs
```diff
@@ -185,6 +200,31 @@ public void GetSets()
             Assert.Null(validationParameters.SignatureValidator);
         }
 
+        [Fact]
```
TimHannMSFT, line 203: >Fact [](http://example.com/codeflow?start=9&length=4)

Just to check my understanding, now that this is in place, it will be leveraged in SAL and that's where we should have tests to ensure properties get added to the instanced copy vs the root/normal property bag, correct?

TimHannMSFT (approved):
:shipit:

test/Microsoft.IdentityModel.Tokens.Tests/TokenValidationParametersTests.cs
brentschmaltz: Yes

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
brentschmaltz: Good for further discussion.

PR #1964, opened 2022-11-11
Replace temporary string.Split calls with manual counting

### LINKED ISSUE #1955: [Performance] Low hanging fruit in Tokens.JWT
Recently while comparing a benchmark that didn't use auth and one that used jwt auth I noticed the allocations were extremely high when using jwt. Digging into it I found many places where allocations could be improved in System.IdentityModel.Tokens.Jwt, I'll focus on the largest allocation I saw for this issue.

Our code first checks if we the token can be read via
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b7a4fdd6c55b1a017469c15e4f4ffeecf9f577f6/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L278
Then it proceeds to call `ValidateToken`
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b7a4fdd6c55b1a017469c15e4f4ffeecf9f577f6/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L835

Both these methods call `string.Split` on the provided token, check how many values are in the array, and then throw away the results.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b7a4fdd6c55b1a017469c15e4f4ffeecf9f577f6/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L292-L293

Trying out a quick test to count the `'.'` characters in a string (then adding 1 to count the segments) shows we can do it allocation free and almost 6 times faster:

<details>
<summary>Benchmark code</summary>

```
[MemoryDiagnoser]
public class SplitCount
{
    private string JwtToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJodHRwOi8vc2NoZW1hcy54bWxzb2FwLm9yZy93cy8yMDA1LzA1L2lkZW50aXR5L2NsYWltcy9uYW1laWRlbnRpZmllciI6InRlc3QiLCJodHRwOi8vc2NoZW1hcy54bWxzb2FwLm9yZy93cy8yMDA1LzA1L2lkZW50aXR5L2NsYWltcy9lbWFpbGFkZHJlc3MiOiJ0ZXN0QHRlc3QuY29tIiwic3ViIjoidGVzdEB0ZXN0LmNvbSIsImV4cCI6MjAxMzUzMjk3NiwiaXNzIjoiVGVzdCIsImF1ZCI6InRlc3QifQ.tQ-s9uAokahLfeKY-SKnvYQWDU4HrFCRV-0Ae7LAabY";

    [Benchmark]
    public int Split()
    {
        var tokenParts = JwtToken.Split(new char[] { '.' }, 6);
        return tokenParts.Length;
    }

    [Benchmark]
    public int ManualCount()
    {
        var count = 1;
        var span = JwtToken.AsSpan();
        while (span.Length > 0)
        {
            var index = span.IndexOf('.');
            if (index < 0)
            {
                break;
            }
            count++;
            span = span.Slice(index);
            if (count == 6)
            {
                break;
            }
        }
        return count;
    }
}
```

</details>

|      Method |      Mean |    Error |   StdDev |   Gen0 | Allocated |
|------------ |----------:|---------:|---------:|-------:|----------:|
|       Split | 193.57 ns | 1.845 ns | 1.725 ns | 0.1183 |     928 B |
| ManualCount |  35.83 ns | 0.155 ns | 0.130 ns |      - |         - |

And since there are at least 2 calls to `string.Split` per request in the scenario I'm looking at, it would be even bigger savings!

## Proposal

1. Create an internal helper method to count the segments in a token instead of using `string.Split`, could even try to vectorize 😄 
2. Then call this method from everywhere that `string.Split` is used just for the length of the output array
3. ???
4. Profit!

BrennanConroy:
Closes https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1955

I mainly focused on changing the `string.Split` calls, but saw a couple places where duplicate `Base64UrlEncode` calls were being made and updated those as well.

I saw about ~12% reduction in allocation rate in a web app I'm testing with.

RPS didn't really change sadly. Looking at CPU traces it looks like RPS wins could be achieved by converting more of the code base to System.Text.Json similar to https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1934


BrennanConroy:
@brentschmaltz

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -520,6 +520,34 @@ internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, Token
...
+        /// Counts the number of Jwt Token segments.
+        /// </summary>
+        /// <param name="token">The Jwt Token.</param>
+        /// <param name="maxCount">The maximum number of segments to count up to.</param>
+        /// <returns>The number of segments up to <paramref name="maxCount"/>.</returns>
+        internal static int CountJwtTokenPart(string token, int maxCount)
+        {
+            var count = 1;
+            var index = 0;
+            token.IndexOf('.', index);
```
lechu445: Result of the operation is not used. It looks this line can be removed.

lechu445:
relates https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1772

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
BrennanConroy:
```suggestion
```

BrennanConroy:
@brentschmaltz this should be a purely positive change, it even touches a code path in the JsonWebTokenHandler class which is what was updated in your recent change.

lechu445 (approved):

lechu445 (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -520,6 +520,33 @@ internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, Token
             return null;
         }
 
+        /// <summary>
+        /// Counts the number of Jwt Token segments.
+        /// </summary>
+        /// <param name="token">The Jwt Token.</param>
+        /// <param name="maxCount">The maximum number of segments to count up to.</param>
+        /// <returns>The number of segments up to <paramref name="maxCount"/>.</returns>
+        internal static int CountJwtTokenPart(string token, int maxCount)
```
eerhardt, line 510: Note, on `net8.0`, this method could take advantage of the new `MemoryExtensions.Count` method, which can take advantage of SIMD operations to make the count even faster. See https://github.com/dotnet/runtime/issues/59466.

brentschmaltz:
@BrennanConroy sorry this hung around so long, if you can scope this to JwtSecurityTokenHandler we would love to take the fix.
We made changes to JsonWebTokenHandler in a different PR.

BrennanConroy:
> We made changes to JsonWebTokenHandler in a different PR.

I can remove the change to that file, but I don't see any recent changes to it
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/85fa86af743e2b1a0078a9ecd956f34ee703acfc/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L86

lechu445:
@BrennanConroy it looks you need to resolve conflicts on your branch

jennyf19:
@BrennanConroy - can you use dev7x as the topic branch?

BrennanConroy:
Your force-push removed a bug fix. I don't know where though.

The current code looks something like
`return EncodedString.Substring(0, EncodedString.LastIndexOf('.'));`
But if `LastIndexOf` returns -1 this will throw.

BrennanConroy:
Oh maybe here:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/68898befd832684372d8c8289fdc60725488f022/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L849

Although maybe that specific spot is guaranteed to always have `.`'s

keegan-caruso:
@BrennanConroy can you clarify?

BrennanConroy:
I think it's fine, the EncodedToken in JsonWebToken looks like it must have `.`'s based on the ctor.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
jennyf19: will we do that in this PR w/some Ifdefs?

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
eerhardt: We would need to target `net8.0`, which we aren't doing:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/ddd4a4c01d2c05cb65323c5bb49e81579f2322b3/build/targets.props#L3

It would also involve using the 8.0 SDK to build. So I would say we should do that in a different PR.

keegan-caruso (approved):

eerhardt (approved):

PR #1967, opened 2022-11-23
Optimize LKG feature for better resilience

ciaozhang:
Add jitter for metadata refresh to avoid spike traffic.
Update default to 24H for LKG lifetime in AadIssuerValidator for 1P.
Update LKG logic: refresh metadata before validate with LKG.
Mark the exception when fetching the metadata as non-PII.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -1224,8 +1224,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             {
                 if (tokenValidationResult.IsValid)
                 {
-                    // Set current configuration as LKG if it exists and has not already been set as the LKG.
-                    if (currentConfiguration != null && !ReferenceEquals(currentConfiguration, validationParameters.ConfigurationManager.LastKnownGoodConfiguration))
+                    // Set current configuration as LKG if it exists.
+                    if (currentConfiguration != null)
                         validationParameters.ConfigurationManager.LastKnownGoodConfiguration = currentConfiguration;
 
                     return tokenValidationResult;
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -1234,15 +1234,6 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 // we want to make sure that the clause for SecurityTokenUnableToValidateException is hit so that the ValidationFailure is checked
                 else if (TokenUtilities.IsRecoverableException(tokenValidationResult.Exception))
                 {
-                    if (TokenUtilities.IsRecoverableConfiguration(validationParameters, currentConfiguration, out currentConfiguration))
-                    {
-                        validationParameters.ValidateWithLKG = true;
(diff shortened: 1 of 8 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -134,7 +134,8 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                         // The transport should have it's own timeouts, etc..
                         var configuration = await _configRetriever.GetConfigurationAsync(MetadataAddress, _docRetriever, CancellationToken.None).ConfigureAwait(false);
                         _lastRefresh = now;
-                        _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval);
+                        // Add 1 hour jitter to avoid spike traffic to ESTS.
```
brentschmaltz: Change comment ESTS -> IdentityProvider.
This is not just for Microsoft.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1234,15 +1234,6 @@ private async Task<TokenValidationResult> ValidateTokenAsync(JsonWebToken jsonWe
                 // we want to make sure that the clause for SecurityTokenUnableToValidateException is hit so that the ValidationFailure is checked
                 else if (TokenUtilities.IsRecoverableException(tokenValidationResult.Exception))
                 {
-                    if (TokenUtilities.IsRecoverableConfiguration(validationParameters, currentConfiguration, out currentConfiguration))
```
brentschmaltz, line 1237: I think the logic for setting LastKnownGood (line 1288) should be just
if (currentConfiguration != null) 
No need to check ReferenceEquals.

Then the logic here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/013a42271866c02574fff4c74cb9ca7d08fc090e/src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs#L75

Should be modified so that when LKG is set, the timer starts.
This means that LKG will be used for only the number of hours AFTER it was last set.
The current logic could result in LKG being valid longer.
Not a big deal, but it's non deterministic.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -904,16 +904,6 @@ private ClaimsPrincipal ValidateToken(string token, JwtSecurityToken outerToken,
                 }
                 else if (TokenUtilities.IsRecoverableException(exceptionThrown.SourceException))
                 {
-                    if (TokenUtilities.IsRecoverableConfiguration(validationParameters, currentConfiguration, out currentConfiguration))
```
brentschmaltz, line 907: Same comment as JsonWebTokenHandler, no need for ReferenceEquals check.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
ciaozhang: updated the logic

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
ciaozhang: updated the logic

brentschmaltz (approved):

PR #1969, opened 2022-11-25
Bundling the exception as an inner exception of the ArgumentException.

### LINKED ISSUE #1965: [Feature Request]  Let ReadToken exceptions trickle up to ValidateTokenAsync
When I have let's say an assembly binding problem on my .net framework web app during validation of a JWT. Instead of getting an exception telling me it couldn't find the right assembly with the correct strong name the app instead sent me a IDX14100 error message `JWT is not well formed: '{0}', there are no dots (.).\nThe token needs to be in JWS or JWE Compact Serialization Format. `
This lead to a wild goose chase trying to validate the tokens were well formed (hidden characters?). Then lead to me having to debug the nugget package in order to find what the real exception was and which DLL were faulty.

The problem come from this function : https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b7a4fdd6c55b1a017469c15e4f4ffeecf9f577f6/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L1173 which is used inside ValidateTokenAsync and doesn't let ValidateTokenAsync recover the real exception.

I would like to have a way to recover the initial exception either through a logging setting or through the Exception property of TokenValidationResult.

Debuguing the internal dlls. Trying to mount the project on my visual studio resulted in strong name signature error. Debuguing using microsoft symbols can be tedious due to async throwing me into execution context and the debuging experience of external library can be tedious to the uninitiated (of which I'm one unfortunately).

**Additional context**
It seems the try catch masquing the original Exception was added in the rewrite to use System.Text.Json and not in the original code.

Esurnir:
This will embed the exception caught from the constructor as an inner exception of the ArgumentException logged by ReadToken as suggested in the issue.

Fixes https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1965.

Unfortunately I'm having trouble building the projet locally and running the tests (the build process mention not finding Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests\obj\Debug\netcoreapp2.1\Microsoft.I
dentityModel.Protocols.SignedHttpRequest.Tests.GeneratedMSBuildEditorConfig.editorconfig for instance). So I'm having trouble figuring out if I need to fix the test or add a new one to make the PR work.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -1174,11 +1174,11 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 {
                     jsonWebToken = new JsonWebToken(token);
                 }
-                catch (Exception)
+                catch (Exception ex)
                 {
                     return new TokenValidationResult
                     {
-                        Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14100, token))),
+                        Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14100, token, ex))),
                         IsValid = false
                     };
                 }


brentschmaltz:
@Esurnir sometimes our tests check if an inner exception is expected.
What problems are you experiencing?

brentschmaltz (approved):

PR #1975, opened 2022-11-30
Handle escaped datetime conversion.

brentschmaltz:
.net 6.0 can throw IndexOutOfRangeException exceptions when calling TryGetDateTime.
In this case, catch the exception and return the claim as a string.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
@@ -78,10 +78,17 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             // Json.net recognized DateTime by default.
             if (jsonElement.ValueKind == JsonValueKind.String)
             {
-                if (jsonElement.TryGetDateTime(out DateTime dateTimeValue))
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -78,10 +78,17 @@ private static Claim CreateClaimFromJsonElement(string key, string issuer, JsonE
...
-                    return new Claim(key, dateTimeValue.ToUniversalTime().ToString("o", CultureInfo.InvariantCulture), ClaimValueTypes.DateTime, issuer, issuer);
-                else
+                try
+                {
+                    if (jsonElement.TryGetDateTime(out DateTime dateTimeValue))
+                        return new Claim(key, dateTimeValue.ToUniversalTime().ToString("o", CultureInfo.InvariantCulture), ClaimValueTypes.DateTime, issuer, issuer);
+                    else
+                        return new Claim(key, jsonElement.ToString(), ClaimValueTypes.String, issuer, issuer);
+                }
+                catch(IndexOutOfRangeException)
```
TimHannMSFT, line 88: 💖

TimHannMSFT (approved):
LGTM

dannybtsai (approved):

PR #1978, opened 2022-12-01
Remove PII scrubbing from Metadata logging

TimHannMSFT:
Before this fix we're scrubbing the MetadataAddress on the Configuration object when logging. These addresses are public and are helpful in diagnosing issues. We should not scrub the address and always log it. This change marks MetadataAddress as NonPII.


jmprieur (approved):
LGTM
Thanks @TimHannMSFT 

is there a related work item?

#supportability

ciaozhang (approved):

PR #1981, opened 2022-12-13
corrected links to datatracker.ietf.org

brentschmaltz:
The links to datatracker.ietf.org used '.' instead of '-'.


jmprieur (approved):
LGTM
Thanks @brentschmaltz

Issue #1985, opened 2022-12-14
[Bug] Dependency System.Text.Json 4.7.2 is affected by CVE-2021-26701 Security Vulnerability


nmaric02-quest:
Using latest version:
System.IdentityModel.Tokens.Jwt 6.25.1

with following transitive dependencies:
Microsoft.IdentityModel.JsonWebTokens 6.25.1
Microsoft.IdentityModel.Tokens 6.25.1

The problem is dependency on System.Text.Json, 4.7.2
This version is affected by CVE-2021-26701 (https://github.com/advisories/GHSA-ghhp-997w-qr28) through transitive dependency on  System.Text.Encodings.Web 4.7.1. 

The issue is fixed in subsequent releases of System.Text.Json - (>= 5.0.2)

Suggested fix:
Update dependency to the latest version of System.Text.Json

jennyf19:
Closing, as I'm assuming this is not an issue anymore with 7x? If I'm wrong, please re-open and we will prioritize. @nmaric02-quest

PR #1997, opened 2023-01-11
Update System.Text.Encodings.Web

TimHannMSFT:
#1985 

Suggested fix is to upgrade System.Text.Json but this could impact other consumers who require the lowest secure version. Therefore updating the dependency which has the issue.


jmprieur (approved):
LGTM
Thanks @TimHannMSFT we have the same in IdWeb

PR #1998, opened 2023-01-12
Added TokenEndpoint to BaseConfiguration

dannybtsai:
Add the TokenEndpoint to the BaseConfiguration class since both OIDC and WsFed have the property.


sruke (approved):
LGTM!

PR #2002, opened 2023-01-18
Add TVP to check if Token Identifier can be logged

### LINKED ISSUE #2001: [Feature Request] Create a TokenValidationParameter that will check if token identifier claim needs to be logged

sruke:
What is changing?
Introducing a new boolean TokenValidationParameter LogTokenId. By default, it set to true.
Why?
It will allow the library to log a token identifier if LogTokenId is set to true.

**Note**: Today, no action will be taken in Wilson based on the value of LogTokenId.


dannybtsai:
Should Wilson log any message when the TokenValidationParameter.LogTokenId is true, or this is only for SAL to use?

sruke:
> Should Wilson log any message when the TokenValidationParameter.LogTokenId is true, or this is only for SAL to use?

Right now, only the TokenValidationParameter (TVP) is getting introduced in Wilson. No action will be taken in Wilson based on the value. I will create a GitHub issue to log the Token Identifier in Wilson based on the TVP value.

dannybtsai (approved):

TimHannMSFT (approved):

PR #2007, opened 2023-01-24
Xiao/LKG configuration cache

ciaozhang:
Introduce a LKG configuration cache to store each valid base configuration instead of a single entry of configuration.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -1230,8 +1230,6 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                     return tokenValidationResult;
                 }
-                // using 'GetType()' instead of 'is' as SecurityTokenUnableToValidException (and others) extend SecurityTokenInvalidSignatureException
-                // we want to make sure that the clause for SecurityTokenUnableToValidateException is hit so that the ValidationFailure is checked
                 else if (TokenUtilities.IsRecoverableException(tokenValidationResult.Exception))
                 {
                     // If we were still unable to validate, attempt to refresh the configuration and validate using it
@@ -1258,14 +1256,22 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                         }
                     }
 
-                    if (TokenUtilities.IsRecoverableConfiguration(validationParameters, currentConfiguration, out currentConfiguration))
+                    if (validationParameters.ConfigurationManager.UseLastKnownGoodConfiguration)
                     {
                         validationParameters.RefreshBeforeValidation = false;
                         validationParameters.ValidateWithLKG = true;
-                        tokenValidationResult = ValidateToken(jsonWebToken, validationParameters, currentConfiguration);
+                        var recoverableException = tokenValidationResult.Exception;
 
-                        if (tokenValidationResult.IsValid)
-                            return tokenValidationResult;
+                        foreach (BaseConfiguration lkgConfiguration in validationParameters.ConfigurationManager.GetValidLkgConfigurations())
+                        {
+                            if (!lkgConfiguration.Equals(currentConfiguration) && TokenUtilities.IsRecoverableConfiguration(jsonWebToken.Kid, currentConfiguration, lkgConfiguration, recoverableException))
+                            {
+                                tokenValidationResult = ValidateToken(jsonWebToken, validationParameters, lkgConfiguration);
+
+                                if (tokenValidationResult.IsValid)
+                                    return tokenValidationResult;
+                            }
+                        }
                     }
                 }
             }
@@ -1534,16 +1540,13 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                 if (!validationParameters.ValidateSignatureLast)
                 {
-                    InternalValidators.ValidateLifetimeAndIssuerAfterSignatureNotValidatedJwt(
+                    InternalValidators.ValidateAfterSignatureFailed(
                         jwtToken,
                         notBefore,
                         expires,
-                        jwtToken.Kid,
+                        jwtToken.Audiences,
                         validationParameters,
-                        configuration,
-                        exceptionStrings,
-                        numKeysInTokenValidationParameters,
-                        numKeysInConfiguration);
+                        configuration);
                 }
             }
 
--- a/src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
+++ b/src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
@@ -1,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Comparison class for a <see cref="BaseConfiguration"/>.
+    /// </summary>
+    internal class BaseConfigurationComparer : IEqualityComparer<BaseConfiguration>
+    {
+        public bool Equals(BaseConfiguration config1, BaseConfiguration config2)
+        {
+            if (config1 == null && config2 == null)
+                return true;
+            else if (config1 == null || config2 == null)
+                return false;
+            else if (config1.Issuer == config2.Issuer && config1.SigningKeys.Count == config2.SigningKeys.Count
+                     && !config1.SigningKeys.Select(x => x.InternalId).Except(config2.SigningKeys.Select(x => x.InternalId)).Any())
+                return true;
+            else
+                return false;
+        }
+
+        public int GetHashCode(BaseConfiguration config)
+        {
+            int defaultHash = string.Empty.GetHashCode();
+            int hashCode = defaultHash;
+            hashCode ^= string.IsNullOrEmpty(config.Issuer) ? defaultHash : config.Issuer.GetHashCode();
+            foreach(string internalId in config.SigningKeys.Select(x => x.InternalId))
+            {
+                hashCode ^= internalId.GetHashCode();
+            }
+
+            return hashCode;
+        }
+    }
+}
--- a/src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
@@ -3,6 +3,8 @@
 
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.IdentityModel.Logging;
@@ -19,6 +21,8 @@ namespace Microsoft.IdentityModel.Tokens
         private TimeSpan _lastKnownGoodLifetime = DefaultLastKnownGoodConfigurationLifetime;
         private BaseConfiguration _lastKnownGoodConfiguration;
         private DateTime? _lastKnownGoodConfigFirstUse = null;
+
+        internal EventBasedLRUCache<BaseConfiguration, DateTime> _lastKnownGoodConfigurationCache;
 
         /// <summary>
         /// Gets or sets the <see cref="TimeSpan"/> that controls how often an automatic metadata refresh should occur.
@@ -64,6 +68,15 @@ namespace Microsoft.IdentityModel.Tokens
         }
 
         /// <summary>
+        /// Gets all valid last known good configurations.
+        /// </summary>
+        /// <returns>A collection of all valid last known good configurations.</returns>
+        internal ICollection<BaseConfiguration> GetValidLkgConfigurations()
+        {
+            return _lastKnownGoodConfigurationCache.ToArray().Where(x => x.Value.Value > DateTime.UtcNow).Select(x => x.Key).ToArray();
+        }
+
+        /// <summary>
         /// The last known good configuration or LKG (a configuration retrieved in the past that we were able to successfully validate a token against).
         /// </summary>
         public BaseConfiguration LastKnownGoodConfiguration
@@ -76,6 +89,9 @@ namespace Microsoft.IdentityModel.Tokens
             {
                 _lastKnownGoodConfiguration = value ?? throw LogHelper.LogArgumentNullException(nameof(value));
                 _lastKnownGoodConfigFirstUse = DateTime.UtcNow;
+
+                // LRU cache will remove the expired configuration
+                _lastKnownGoodConfigurationCache.SetValue(_lastKnownGoodConfiguration, DateTime.UtcNow + LastKnownGoodLifetime, DateTime.UtcNow + LastKnownGoodLifetime);
             }
         }
 
--- a/src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
@@ -187,43 +187,42 @@ namespace Microsoft.IdentityModel.Tokens
         /// <returns><c>true</c> if the exception is certain types of exceptions otherwise, <c>false</c>.</returns>
         internal static bool IsRecoverableException(Exception exception)
         {
-            // using 'GetType()' instead of 'is' as SecurityTokenUnableToValidException (and others) extend SecurityTokenInvalidSignatureException
-            // we want to make sure that the clause for SecurityTokenUnableToValidateException is hit so that the ValidationFailure is checked
-            return exception.GetType().Equals(typeof(SecurityTokenInvalidSignatureException))
-                   || exception is SecurityTokenInvalidSigningKeyException
-                   || exception is SecurityTokenInvalidIssuerException
-                   // we should not try to revalidate with the LKG or request a refresh if the token has an invalid lifetime
-                   || (exception as SecurityTokenUnableToValidateException)?.ValidationFailure != ValidationFailure.InvalidLifetime
-                   || exception is SecurityTokenSignatureKeyNotFoundException;
+            return   exception is SecurityTokenInvalidSignatureException
+                  || exception is SecurityTokenInvalidIssuerException
+                  || exception is SecurityTokenSignatureKeyNotFoundException;
         }
 
         /// <summary>
         /// Check whether the given configuration is recoverable by LKG.
         /// </summary>
-        /// <param name="validationParameters">The <see cref="TokenValidationParameters"/> to be used for validation.</param>
-        /// <param name="configuration">The <see cref="BaseConfiguration"/> to check.</param>
-        /// <param name="currentConfiguration">The updated <see cref="BaseConfiguration"/>.</param>
+        /// <param name="kid">The kid from token."/></param>
+        /// <param name="currentConfiguration">The <see cref="BaseConfiguration"/> to check.</param>
+        /// <param name="lkgConfiguration">The LKG exception to check.</param>
+        /// <param name="currentException">The exception to check.</param>
         /// <returns><c>true</c> if the configuration is recoverable otherwise, <c>false</c>.</returns>
-        internal static bool IsRecoverableConfiguration(TokenValidationParameters validationParameters, BaseConfiguration configuration, out BaseConfiguration currentConfiguration)
+        internal static bool IsRecoverableConfiguration(string kid, BaseConfiguration currentConfiguration, BaseConfiguration lkgConfiguration, Exception currentException)
         {
-            bool isRecoverableConfiguration = (validationParameters.ConfigurationManager.UseLastKnownGoodConfiguration
-                && validationParameters.ConfigurationManager.LastKnownGoodConfiguration != null
-                && !ReferenceEquals(configuration, validationParameters.ConfigurationManager.LastKnownGoodConfiguration));
+            Lazy<bool> isRecoverableSigningKey = new Lazy<bool>(() => lkgConfiguration.SigningKeys.Any(signingKey => signingKey.KeyId == kid));
 
-            currentConfiguration = configuration;
-            if (isRecoverableConfiguration)
+            if (currentException is SecurityTokenInvalidIssuerException)
             {
-                // Inform the user that the LKG is expired.
-                if (!validationParameters.ConfigurationManager.IsLastKnownGoodValid)
-                {
-                    LogHelper.LogInformation(TokenLogMessages.IDX10263);
-                    return false;
-                }
-                else
-                    currentConfiguration = validationParameters.ConfigurationManager.LastKnownGoodConfiguration;
+                return currentConfiguration.Issuer != lkgConfiguration.Issuer;
+            }
+            else if (currentException is SecurityTokenSignatureKeyNotFoundException)
+            {
+                return isRecoverableSigningKey.Value;
+            }
+            else if (currentException is SecurityTokenInvalidSignatureException)
+            {
+                SecurityKey currentSigningKey = currentConfiguration.SigningKeys.FirstOrDefault(x => x.KeyId == kid);
+                if (currentSigningKey == null)
+                    return isRecoverableSigningKey.Value;
+
+                SecurityKey lkgSigningKey = lkgConfiguration.SigningKeys.FirstOrDefault(signingKey => signingKey.KeyId == kid);
+                return lkgSigningKey != null && currentSigningKey.InternalId != lkgSigningKey.InternalId;
             }
 
-            return isRecoverableConfiguration;
+            return false;
         }
     }
 }
--- a/test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
+++ b/test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
@@ -141,11 +141,12 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests
         {
             TestUtilities.WriteHeader($"{this}.GetSets", "GetSets", true);
 
+            int ExpectedPropertyCount = 7;
             var configManager = new ConfigurationManager<OpenIdConnectConfiguration>("OpenIdConnectMetadata.json", new OpenIdConnectConfigurationRetriever(), new FileDocumentRetriever());
             Type type = typeof(ConfigurationManager<OpenIdConnectConfiguration>);
             PropertyInfo[] properties = type.GetProperties();
-            if (properties.Length != 7)
-                Assert.True(false, "Number of properties has changed from 7 to: " + properties.Length + ", adjust tests");
+            if (properties.Length != ExpectedPropertyCount)
+                Assert.True(false, $"Number of properties has changed from {ExpectedPropertyCount} to: " + properties.Length + ", adjust tests");
 
             var defaultAutomaticRefreshInterval = ConfigurationManager<OpenIdConnectConfiguration>.DefaultAutomaticRefreshInterval;
             var defaultRefreshInterval = ConfigurationManager<OpenIdConnectConfiguration>.DefaultRefreshInterval;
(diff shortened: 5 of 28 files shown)


ciaozhang:
This PR is not ready for reivew.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
```diff
@@ -0,0 +1,32 @@
...
+            else if (config1.Issuer == config2.Issuer && config1.SigningKeys.Count == config2.SigningKeys.Count
+                     && !config1.SigningKeys.Select(x => x.InternalId).Except(config1.SigningKeys.Select(x => x.InternalId)).Any())
+                return true;
+            else
+                return false;
+        }
+
+        public int GetHashCode(BaseConfiguration config)
+        {
+            return (config.Issuer + string.Join("", config.SigningKeys.Select(x => x.InternalId).OrderBy(x => x.FirstOrDefault()))).GetHashCode();
```
TimHannMSFT: this might work for the prototype but we shouldn't be doing string operations to produce the hash code in the release version. Can you at least comment what you plan to do here?

src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
```diff
@@ -0,0 +1,32 @@
...
+    internal class BaseConfigurationComparer : IEqualityComparer<BaseConfiguration>
+    {
+        public bool Equals(BaseConfiguration config1, BaseConfiguration config2)
+        {
+            if (config1 == null && config2 == null)
+                return true;
+            else if (config1 == null || config2 == null)
+                return false;
+            else if (config1.Issuer == config2.Issuer && config1.SigningKeys.Count == config2.SigningKeys.Count
+                     && !config1.SigningKeys.Select(x => x.InternalId).Except(config1.SigningKeys.Select(x => x.InternalId)).Any())
```
TimHannMSFT: Is the InternalId some generated hash of the crypto artifact? How can we be sure this isn't duplicated with a valid and invalid artifact in config1 vs config2? #Resolved

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -76,9 +93,39 @@ public BaseConfiguration LastKnownGoodConfiguration
             {
                 _lastKnownGoodConfiguration = value ?? throw LogHelper.LogArgumentNullException(nameof(value));
                 _lastKnownGoodConfigFirstUse = DateTime.UtcNow;
+
+                if (_lkgConfigurationCache == null)
+                    _lkgConfigurationCache = new Dictionary<BaseConfiguration, DateTime>(BaseConfigurationComparer);
+                
+                _lkgConfigurationCache[_lastKnownGoodConfiguration] = DateTime.UtcNow + LastKnownGoodLifetime;
```
TimHannMSFT: >_lkgConfigurationCache [](http://example.com/codeflow?start=16&length=22)

Is there any sane upper bound we can add on this cache? Currently it's only bounded by the LastKnownGoodLifetime correct? This goes back to the complexity functionality tradeoff but seems reasonable to have a high upper bound here which can be overridden to prevent runaway cache in some error case where ESTS starts doing something weird with metadata publishing.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -76,9 +93,39 @@ public BaseConfiguration LastKnownGoodConfiguration
...
+
+                //remove expired configuration to avoid memory leak 
+                foreach (KeyValuePair<BaseConfiguration, DateTime> lkgConfiguration in _lkgConfigurationCache)
```
TimHannMSFT: >foreach [](http://example.com/codeflow?start=16&length=7)

Does this work? My memory of modifying a collection while iterating over it was that it will throw an exception.

[https://stackoverflow.com/questions/2347269/modifying-net-dictionary-while-enumerating-through-it](https://stackoverflow.com/questions/2347269/modifying-net-dictionary-while-enumerating-through-it)

Generally you'll want to create a collection and remove once finished with iteration.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -76,9 +93,39 @@ public BaseConfiguration LastKnownGoodConfiguration
...
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <returns></returns>
+        public IDictionary<BaseConfiguration, DateTime> GetValidLkgConfiguraitonFromCache()
+        {
+            if (_lkgConfigurationCache == null)
+                return null;
+
+            foreach (KeyValuePair<BaseConfiguration, DateTime> lkgConfiguration in _lkgConfigurationCache)
```
TimHannMSFT: >foreach [](http://example.com/codeflow?start=12&length=7)

same comment regarding removing while iterating, consider a private helper since you're doing this in two places.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
ciaozhang: We could set the upper limit for the cache of lavage our EventBaseLruCache.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
ciaozhang: Here to use issuer and sorted signingKeys.InternalId to generate a hashcode.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
ciaozhang: Is the InternalId some generated hash of the crypto artifact?
Yes

src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
TimHannMSFT: take a look at something like: [https://stackoverflow.com/questions/59375124/how-to-use-system-hashcode-combine-with-more-than-8-values](https://stackoverflow.com/questions/59375124/how-to-use-system-hashcode-combine-with-more-than-8-values)

We shouldn't be doing string operations in the GetHashCode method.

ciaozhang: Update the GetHashcode method with "^".

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1258,14 +1258,22 @@ private async Task<TokenValidationResult> ValidateTokenAsync(JsonWebToken jsonWe
                         }
                     }
 
-                    if (TokenUtilities.IsRecoverableConfiguration(validationParameters, currentConfiguration, out currentConfiguration))
+                    if (TokenUtilities.ShouldValidateWithLKG(validationParameters))
                     {
                         validationParameters.RefreshBeforeValidation = false;
                         validationParameters.ValidateWithLKG = true;
-                        tokenValidationResult = await ValidateTokenAsync(jsonWebToken, validationParameters, currentConfiguration).ConfigureAwait(false);
+                        var recoverableException = tokenValidationResult.Exception;
```
brentschmaltz, line 1263: Do we need to create this local variable 'recoverableException', it is only used once.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
brentschmaltz: The InternalId is based off this spec: https://www.rfc-editor.org/rfc/rfc7638

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
ciaozhang: It is make sure we are recovering from the current configuration. Otherwise, each time we loop through the cache might get different exception.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -63,6 +80,25 @@ public virtual Task<BaseConfiguration> GetBaseConfigurationAsync(CancellationTok
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Gets all valid last known good configurations from the cache.
+        /// </summary>
+        /// <returns>A collection of all valid last known good configurations.</returns>
+        internal ICollection<BaseConfiguration> GetValidLkgConfiguraitonFromCache()
```
brentschmaltz: Suggest dropping the postfix: 'Cache'

TimHannMSFT: Makes sense, a bit superfluous I suppose, LKG will always be a cache of some sort.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -330,6 +331,43 @@ public void ResetLastKnownGoodLifetime()
             TestUtilities.AssertFailIfErrors(context);
         }
 
+        [Fact]
+        public void TestConfigurationComparer()
+        {
+            TestUtilities.WriteHeader($"{this}.TestConfigurationComparer", "TestConfigurationComparer", true);
+            var context = new CompareContext();
+
+            var config1 = new OpenIdConnectConfiguration() { TokenEndpoint = Default.Issuer + "oauth/token", Issuer = Default.Issuer };
```
TimHannMSFT: >TokenEndpoint = Default.Issuer + "oauth/token" [](http://example.com/codeflow?start=61&length=46)

here and below this will produce: "[http://Default.Issuer.comoauth/token](http://Default.Issuer.comoauth/token)"

add in the '/' for a more valid URI.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -330,6 +331,43 @@ public void ResetLastKnownGoodLifetime()
...
+            var config2 = new OpenIdConnectConfiguration() { TokenEndpoint = Default.Issuer + "oauth/token", Issuer = Default.Issuer };
+            config2.SigningKeys.Add(KeyingMaterial.DefaultRsaSecurityKey1);
+            config2.SigningKeys.Add(KeyingMaterial.DefaultX509Key_2048);
+            config2.SigningKeys.Add(KeyingMaterial.DefaultRsaSecurityKey2);
+
+            var config3 = new OpenIdConnectConfiguration() { TokenEndpoint = Default.Issuer + "oauth/token", Issuer = Default.Issuer };
+            config3.SigningKeys.Add(KeyingMaterial.DefaultRsaSecurityKey1);
+
+            var config4 = new OpenIdConnectConfiguration() { TokenEndpoint = Default.Issuer + "oauth/token", Issuer = Default.Issuer + "1" };
+            config3.SigningKeys.Add(KeyingMaterial.DefaultRsaSecurityKey1);
```
TimHannMSFT: >config3 [](http://example.com/codeflow?start=12&length=7)

Is this supposed to be config4?

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
TimHannMSFT: >config1 [](http://example.com/codeflow?start=16&length=7)

Please add comments to be explicit about the test cases:

Config,
Config with same keys but different order
Config with overlapping key
Config with overlapping key but different issuer

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -330,6 +331,43 @@ public void ResetLastKnownGoodLifetime()
             TestUtilities.AssertFailIfErrors(context);
         }
 
+        [Fact]
+        public void TestConfigurationComparer()
```
TimHannMSFT, line 371: >TestConfigurationComparer [](http://example.com/codeflow?start=20&length=25)

Test case where token endpoint differs but is considered duplicate because it's not part of the hash check?

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -63,6 +80,25 @@ public virtual Task<BaseConfiguration> GetBaseConfigurationAsync(CancellationTok
...
+                return null;
+
+            var expiredLkgConfiguration = _lkgConfigurationCache.Where(x => x.Value < DateTime.UtcNow).ToArray();
+
+            foreach (KeyValuePair<BaseConfiguration, DateTime> lkgConfiguration in expiredLkgConfiguration)
+            {
+                _lkgConfigurationCache.Remove(lkgConfiguration.Key);
+            }
+
+            return _lkgConfigurationCache.Any() ? _lkgConfigurationCache.Keys : null;
```
TimHannMSFT: >null [](http://example.com/codeflow?start=80&length=4)

I think, as a developer, I'd expect this to return an empty set in this case not null. null seems to be the case where there's no cache (is that possible?) empty should be when there's nothing valid (e.g. all expired) IMO.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
TimHannMSFT: Does anything real-world prevent duplicate keys being added? Here it looks like this Config results in SigningKeys collection with duplicate keys, that's something we'd never want in practice, correct?

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -330,6 +331,43 @@ public void ResetLastKnownGoodLifetime()
...
+            var configurationManager = new MockConfigurationManager<OpenIdConnectConfiguration>(config1, config1);
+            IdentityComparer.AreEqual(configurationManager.GetValidLkgConfiguraitonFromCache().Count, 1, context);
+
+            configurationManager.LastKnownGoodConfiguration = config2;
+            IdentityComparer.AreEqual(configurationManager.GetValidLkgConfiguraitonFromCache().Count, 1, context);
+
+            configurationManager.LastKnownGoodConfiguration = config3;
+            IdentityComparer.AreEqual(configurationManager.GetValidLkgConfiguraitonFromCache().Count, 2, context);
+
+            configurationManager.LastKnownGoodConfiguration = config4;
```
TimHannMSFT: >LastKnownGoodConfiguration [](http://example.com/codeflow?start=33&length=26)

because of the testing bug, we're adding a config with no keys to the cache, would we ever want to actually do this? Seems like something should prevent this, I don't see what the value would be. Perhaps there should be a warning log too.

test/Microsoft.IdentityModel.TestUtils/MockConfigurationManager.cs
```diff
@@ -74,6 +74,30 @@ public MockConfigurationManager(T configuration, T lkgConfiguration, T refreshed
...
+            LastKnownGoodLifetime = lkgLifetime;
+        }
+
+        /// <summary>
+        /// Initializes an new instance of <see cref="MockConfigurationManager{T}"/> with a Configuration instance and a LKG Configuration instance.
+        /// </summary>
+        /// <param name="configuration">Configuration of type OpenIdConnectConfiguration or WsFederationConfiguration.</param>
+        /// <param name="lkgConfiguration">Configuration of type OpenIdConnectConfiguration or WsFederationConfiguration.</param>
+        /// <param name="lkgLifetime">The LKG configuration lifetime.</param>
+        public MockConfigurationManager(T configuration, T lkgConfiguration, TimeSpan lkgLifetime) : this(configuration, lkgLifetime)
```
TimHannMSFT, line 99: >T [](http://example.com/codeflow?start=57&length=1)

BaseConfiguration?

TimHannMSFT:
>     /// <typeparam name="T">must be a class.</typeparam>

Not part of your change but why do we need to be so lax here? Wouldn't we always want T to be of type BaseConfiguration?

---
Refers to: test/Microsoft.IdentityModel.TestUtils/MockConfigurationManager.cs:17 in 32e63b8. [](commit_id = 32e63b88faf74433cc2ca7f8899836b64ad687c6, deletion_comment = False)

src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
TimHannMSFT: Are you doing force pushes here? (I see this changed but it still shows as the first iteration) Once you push a PR, please just push new changes as new commits in case anyone has checked out your branch and is attached to its change history.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
TimHannMSFT: Also consider a test case which covers duplicate keys with same kid

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -330,6 +331,43 @@ public void ResetLastKnownGoodLifetime()
             TestUtilities.AssertFailIfErrors(context);
         }
```
TimHannMSFT, line 368: >} [](http://example.com/codeflow?start=8&length=1)

Unit tests for aging out the LKG entries?

TimHannMSFT (changes requested):
:clock1:

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
ciaozhang: updated

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
ciaozhang: Added test

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
ciaozhang: We have ConfigurationValidtor to make sure there is at least 1 key in the key set.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -76,6 +111,24 @@ public BaseConfiguration LastKnownGoodConfiguration
             {
                 _lastKnownGoodConfiguration = value ?? throw LogHelper.LogArgumentNullException(nameof(value));
                 _lastKnownGoodConfigFirstUse = DateTime.UtcNow;
+
+                if (_lkgConfigurationCache == null)
+                    _lkgConfigurationCache = new EventBasedLRUCache<BaseConfiguration, DateTime>(LastKnownGoodConfigurationSizeLimit, TaskCreationOptions.None, BaseConfigurationComparer, true, maintainLRU: true);
```
ciaozhang: Adopt the EventBaseLRUCache here. @brentschmaltz @dannybtsai 
What is a reasonable value for default cache size limit?

test/Microsoft.IdentityModel.TestUtils/MockConfigurationManager.cs
ciaozhang: the type should be derived from BaseConfiguration

src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
```diff
@@ -0,0 +1,40 @@
...
```
RojaEnnam: This should be config2?
config1.SigningKeys.Select(x => x.InternalId).Except(**config2**.SigningKeys.Select(x => x.InternalId)).Any())

ciaozhang: Good catch!

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -190,7 +190,6 @@ internal static bool IsRecoverableException(Exception exception)
             // using 'GetType()' instead of 'is' as SecurityTokenUnableToValidException (and others) extend SecurityTokenInvalidSignatureException
             // we want to make sure that the clause for SecurityTokenUnableToValidateException is hit so that the ValidationFailure is checked
             return exception.GetType().Equals(typeof(SecurityTokenInvalidSignatureException))
-                   || exception is SecurityTokenInvalidSigningKeyException
```
ciaozhang, line 193: TBD: SecurityTokenInvalidSigningKeyException is recoverable exception or not.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
brentschmaltz: 10 - 20

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -63,6 +86,18 @@ public virtual Task<BaseConfiguration> GetBaseConfigurationAsync(CancellationTok
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Gets all valid last known good configurations.
+        /// </summary>
+        /// <returns>A collection of all valid last known good configurations.</returns>
+        internal ICollection<BaseConfiguration> GetValidLkgConfiguraitons()
```
brentschmaltz: Please rename: GetValidLkgConfiguraitons => GetValidLkgConfigurations.

'i' is in the wrong place.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -20,6 +22,10 @@ public abstract class BaseConfigurationManager
         private BaseConfiguration _lastKnownGoodConfiguration;
         private DateTime? _lastKnownGoodConfigFirstUse = null;
 
+        private EventBasedLRUCache<BaseConfiguration, DateTime> _lkgConfigurationCache = null;
```
brentschmaltz: be consistent with naming.
_lkgConfigurationCache vs. _lastKnownGoodConfigFirstUse vs. _lastKnownGoodConfiguration vs. _lastKnownGoodConfigurationSizeLimit

_lkgConfiguration is my choice.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -76,6 +111,24 @@ public BaseConfiguration LastKnownGoodConfiguration
             {
                 _lastKnownGoodConfiguration = value ?? throw LogHelper.LogArgumentNullException(nameof(value));
                 _lastKnownGoodConfigFirstUse = DateTime.UtcNow;
+
+                if (_lkgConfigurationCache == null)
```
brentschmaltz: I don't see a reason to delay the creation of the cache, might as well new it up in the ctor or assign the private variable.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
brentschmaltz: I don't think we should set maintainLRU to true.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -63,6 +86,18 @@ public virtual Task<BaseConfiguration> GetBaseConfigurationAsync(CancellationTok
...
+        /// <summary>
+        /// Gets all valid last known good configurations.
+        /// </summary>
+        /// <returns>A collection of all valid last known good configurations.</returns>
+        internal ICollection<BaseConfiguration> GetValidLkgConfiguraitons()
+        {
+            if (_lkgConfigurationCache == null)
+                return null;
+
+            return _lkgConfigurationCache.ToArray().Where(x => x.Value.Value > DateTime.UtcNow).Select(x => x.Key).ToArray();
```
brentschmaltz: The EventBasedLRUCache will keep the cache the right size and remove expired items in the background.
Could we just return _LkgConfigurationCache.ToArray() assuming all items in the cache should be tried?

If we make the other change to instantiate the cache in the ctor, no need to check for null.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1258,14 +1258,22 @@ private async Task<TokenValidationResult> ValidateTokenAsync(JsonWebToken jsonWe
...
                         validationParameters.RefreshBeforeValidation = false;
                         validationParameters.ValidateWithLKG = true;
-                        tokenValidationResult = ValidateToken(jsonWebToken, validationParameters, currentConfiguration);
+                        var recoverableException = tokenValidationResult.Exception;
 
-                        if (tokenValidationResult.IsValid)
-                            return tokenValidationResult;
+                        foreach (BaseConfiguration lkgConfiguration in validationParameters.ConfigurationManager.GetValidLkgConfiguraitons())
+                        {
+                            if (!lkgConfiguration.Equals(currentConfiguration) && TokenUtilities.IsRecoverableConfiguration(jsonWebToken.Kid, currentConfiguration, lkgConfiguration, recoverableException))
```
brentschmaltz, line 1267: We checked above if the error is recoverable.
Could we just try all the configurations in the cache?

We only add configurations to the cache if they are different.
We could have a situation where we have configurations that are different only by issuer with the same keys, but i think it is ok as we are in a temporary situation and keeping the logic simple is a good choice.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1258,14 +1258,22 @@ private async Task<TokenValidationResult> ValidateTokenAsync(JsonWebToken jsonWe
                         }
                     }
 
-                    if (TokenUtilities.IsRecoverableConfiguration(validationParameters, currentConfiguration, out currentConfiguration))
+                    if (TokenUtilities.ShouldValidateWithLKG(validationParameters))
```
brentschmaltz: If we make the change in BaseConfigurationManager.GetValidLkgConfiguraitons(), TokenUtilities.ShouldValidateWithLKG is not needed as we assume all entries in the lkgCache should be tried.
If there are none, an empty array will be returned.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
ciaozhang: Going to use _lastKnownGoodConfigurationCache and it is consistent with other public property.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
ciaozhang: Use defalut value(false) for maintainLRU

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
brentschmaltz: I don't think SecurityTokenInvalidSigningKeyException is recoverable.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -200,30 +199,39 @@ internal static bool IsRecoverableException(Exception exception)
...
+        /// <param name="lkgConfiguration">The LKG exception to check.</param>
+        /// <param name="currentException">The exception to check.</param>
         /// <returns><c>true</c> if the configuration is recoverable otherwise, <c>false</c>.</returns>
-        internal static bool IsRecoverableConfiguration(TokenValidationParameters validationParameters, BaseConfiguration configuration, out BaseConfiguration currentConfiguration)
+        internal static bool IsRecoverableConfiguration(string kid, BaseConfiguration currentConfiguration, BaseConfiguration lkgConfiguration, Exception currentException)
         {
-            bool isRecoverableConfiguration = (validationParameters.ConfigurationManager.UseLastKnownGoodConfiguration
-                && validationParameters.ConfigurationManager.LastKnownGoodConfiguration != null
-                && !ReferenceEquals(configuration, validationParameters.ConfigurationManager.LastKnownGoodConfiguration));
+            Lazy<bool> isRecoverableIssuer = new Lazy<bool>(() => currentConfiguration.Issuer != lkgConfiguration.Issuer);
```
brentschmaltz: I am not understanding the need for this Lazy.
On line 214, why not just write: return (currentConfiguration.Issuer != lkgConfiguration.Issuer)?

While the code may seem cleaner, Lazy introduces a new object creation.

brentschmaltz (approved):

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
ciaozhang: resolved

Issue #1743, opened 2021-10-27
Increase the minimum size of SymmetricSecurityKey for HS256


ackh:
`SymmetricSecurityKey` instances with a length of at least 16 bytes (128 bit) that are used together with `SecurityAlgorithms.HmacSha256Signature` are accepted when creating new JWTs using `JwtSecurityTokenHandler`. An example of such a valid key is as follows:

````csharp
SymmetricSecurityKey key = new(new byte[] { 16, 172, 97, 77, 34, 201, 23, 99, 198, 56, 241, 0, 23, 97, 79, 14 });
````

Keys with less than 16 bytes result in the following exception

````
System.ArgumentOutOfRangeException: IDX10653: The encryption algorithm 'System.String' requires a key size of at least 'System.Int32' bits. Key 'Microsoft.IdentityModel.Tokens.SymmetricSecurityKey', is of size: 'System.Int32'. (Parameter 'key')
````

However, [RFC 7518](https://datatracker.ietf.org/doc/html/rfc7518#section-3.2) states the following about the key length:

>  A key of the same size as the hash output (for instance, 256 bits for "HS256") or larger MUST be used with this algorithm.

As far as i understand, the current implementation violates that statement because the statement requires at least 32 bytes, not just 16.

Am I missing something or is that a potential security issue?

As a side note: the exception message does not print the actual size of the key but simply `System.Int32` which isn't helpful

brentschmaltz:
@ackh thanks for noticing this, we will look into this.
About the error message, when PII is on, we hid all variables in the message.
We are working on relaxing this for some messages that don't contain any pii.

aussieklutz:
Just a further note on this, HMAC signing keys should right pad out to the block size with /0 bytes, as per [RFC 2104](https://www.ietf.org/rfc/rfc2104.txt). The following issue, raised against the JwtBearerHandler demonstrates this point: https://github.com/dotnet/aspnetcore/issues/41024

brentschmaltz:
@cesarpayan can you please look into this.

aussieklutz:
[...]

PR #2008, opened 2023-01-27
Validate size of signature against algorithm considering JWE

brentschmaltz:
SymmetricSignatures are now ensuring the size of the signature matches the algorithm.
Authenticated Encryption needed special processing as only the first half of the signature is considered for Aes128CbcHmacSha256, Aes128CbcHmacSha384 and Aes128CbcHmacSha512.

### DIFF
--- a/test/Microsoft.IdentityModel.Tokens.Tests/TamperedTokenTests.cs
+++ b/test/Microsoft.IdentityModel.Tokens.Tests/TamperedTokenTests.cs
@@ -1,0 +1,143 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.IdentityModel.Tokens.Jwt;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.TestUtils;
+using Xunit;
+
+using JwtRegisteredClaimNames = Microsoft.IdentityModel.JsonWebTokens.JwtRegisteredClaimNames;
+
+namespace Microsoft.IdentityModel.Tokens.Tests
+{
+    public class TamperedTokenTests
+    {
+        [Theory, MemberData(nameof(JwtSignatureTruncationTheoryData))]
+        public async Task JwtSignatureTruncation(ValidateTokenTheoryData theoryData)
+        {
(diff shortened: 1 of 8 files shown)


TimHannMSFT:
Seems like this also addresses #1743, is that correct?

brentschmaltz:
@TimHannMSFT had not considered #1743

TimHannMSFT:
My mistake, I think I got the key size check and the signature size check confused.

---
In reply to: [1407145465](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2008#issuecomment-1407145465) [](http://example.com/codeflow?ancestors=1407145465,1407145465)

test/Microsoft.IdentityModel.Tokens.Tests/TamperedTokenTests.cs
```diff
@@ -0,0 +1,143 @@
...
+using Xunit;
+
+using JwtRegisteredClaimNames = Microsoft.IdentityModel.JsonWebTokens.JwtRegisteredClaimNames;
+
+namespace Microsoft.IdentityModel.Tokens.Tests
+{
+    public class TamperedTokenTests
+    {
+        [Theory, MemberData(nameof(JwtSignatureTruncationTheoryData))]
+        public async Task JwtSignatureTruncation(ValidateTokenTheoryData theoryData)
```
TimHannMSFT, line 19: >JwtSignatureTruncation [](http://example.com/codeflow?start=26&length=22)

Looks like we're doing some similar testing in SignatureProviderTests::SignatureTrunctation. What's the difference? Looks like that also did symmetric hmac.

Can you help me understand why that didn't catch the issue? Should that test case be updated in any way for protection on the RSA/ECDSA cases?

TimHannMSFT (approved):
:shipit:

7409355636 (changes requested):
Help mail all my checks and money to maggie mccann 1138 rose street portsmouth ohio 45662

test/Microsoft.IdentityModel.Tokens.Tests/TamperedTokenTests.cs
brentschmaltz: @TimHannMSFT internally RSA/ECDSA require the signature bytes to be the correct size, so any truncation will be caught by the .net runtimes.

Signature provider tests missed this because they were being told an incorrect expected size from the JsonWebTokenHandler.

PR #2010, opened 2023-02-01
Xiao/Enrich metadata failure message during metadata refresh interval

### LINKED ISSUE #1858: [FeatureRequest] Cache the previous "richer error details" in ConfigurationManager.GetConfigurationAsync.
This feature request is a follow-up from [my previous feature request](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1714) (thanks for addressing that one!).

Consider a scenario where during token validation, we want to retrieve the `OpenIdConnectConfiguration` for a tenant-specific authority containing a tenant ID that does not exist by calling a URI like `https://login.windows.net/aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa/.well-known/openid-configuration`. Thanks to [this pull request](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1776), the underlying `HttpDocumentRetriever` now throws an `IOException` whose `Data` property contains the HttpStatusCode and full ResponseContent from the authority.

In `ConfigurationManager.GetConfigurationAsync`, there are currently two places where an IDX20803 exception can be thrown: at Line 176 and at Line 186 (see the below code snippet). The first time that `ConfigurationManager.GetConfigurationAsync` gets called, `_syncAfter` is `DateTimeOffset.MinValue`, so the IDX20803 exception at Line 176 gets thrown (which contains the inner exception) and `_syncAfter` is set to be (now + 5 minutes). But if `ConfigurationManager.GetConfigurationAsync` is called again with the same authority 2 minutes later, then the IDX20803 exception at Line 186 gets thrown instead, which doesn't contain an inner exception. Consequently, since the thrown exception is missing the rich details from its inner exception, our service is unable to distinguish if the error was due to client misconfiguration (e.g.: tenant not found) or some transient error (e.g.: networking issue).

**In this scenario, can the previous "richer error details" be cached so that if the 5 minute refresh interval hasn't elapsed yet, then an exception with better diagnostic information can be quickly thrown instead of a generic exception that is missing the richer error details?**

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/433f26b30b9b45adae9a9e7286edebf9222ba831/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs#L153-L186

ciaozhang:
Enrich the fetch metadata failure details from its inner exception. That enabled us to distinguish if the error was due to client misconfiguration (e.g.: tenant not found) or some transient error (e.g.: networking issue) during the 5 mins of metadata refresh interval.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -27,7 +27,7 @@ namespace Microsoft.IdentityModel.Protocols
         private readonly IConfigurationRetriever<T> _configRetriever;
         private readonly IConfigurationValidator<T> _configValidator;
         private T _currentConfiguration;
-        private Exception _getConfigurationException;
+        private Exception _fetchMetadataFailure;
 
         /// <summary>
         /// Static initializer for a new object. Static initializers run before the first instance of the type is created.
@@ -149,7 +149,7 @@ namespace Microsoft.IdentityModel.Protocols
                     }
                     catch (Exception ex)
                     {
-                        _getConfigurationException = ex;
+                        _fetchMetadataFailure = ex;
                         _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
                         if (_currentConfiguration == null) // Throw an exception if there's no configuration to return.
                             throw LogHelper.LogExceptionMessage(
@@ -168,7 +168,7 @@ namespace Microsoft.IdentityModel.Protocols
                 else
                     throw LogHelper.LogExceptionMessage(
                         new InvalidOperationException(
-                            LogHelper.FormatInvariant(LogMessages.IDX20803, LogHelper.MarkAsNonPII(MetadataAddress ?? "null"), LogHelper.MarkAsNonPII(_getConfigurationException)), _getConfigurationException));
+                            LogHelper.FormatInvariant(LogMessages.IDX20803, LogHelper.MarkAsNonPII(MetadataAddress ?? "null"), LogHelper.MarkAsNonPII(_fetchMetadataFailure)), _fetchMetadataFailure));
             }
             finally
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -148,6 +149,7 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                     }
                     catch (Exception ex)
                     {
+                        _fetchMetadataFailure = ex;
```
TimHannMSFT, line 152: What requires this to be a variable at the class level instead of just a local variable? Seems like there's an oppertunity here for race conditions and I'm not clear on the advantage.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -100,6 +101,41 @@ public void Defaults()
...
 
+        [Fact]
+        public void FetchMetadataFailureTest()
+        {
+            var context = new CompareContext($"{this}.FetchMetadataFailureTest");
+
+            var documentRetriever = new HttpDocumentRetriever(HttpResponseMessageUtils.SetupHttpClientThatReturns("OpenIdConnectMetadata.json", HttpStatusCode.NotFound));
+            var configManager = new ConfigurationManager<OpenIdConnectConfiguration>("OpenIdConnectMetadata.json", new OpenIdConnectConfigurationRetriever(), documentRetriever);
+
+            //First time to fetch metadata
```
TimHannMSFT: nit: Generally the style is to add a space after the // for comments. (one other occurance below as well)

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
ciaozhang: If ConfigurationManager.GetConfigurationAsync is called again before the syncAfter, then the IDX20803 exception at Line 167 gets thrown instead. We need the exception at the class level so that we can return it as the inner exception.

TimHannMSFT: Ok, this seems reasonable to me. Makes sense we wouldn't just want to have the exception in logs the first time from a usibility/servicability standpoint.

TimHannMSFT (approved):
LGTM

PR #2023, opened 2023-02-14
Add encryption keys to base configuration

RojaEnnam:
Adds encryption keys to base configuration and leverages configuration while finding the keys during token decryption.


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -939,7 +944,43 @@ private static string EncryptTokenPrivate(string innerJwt, EncryptingCredentials
...
-        internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken, TokenValidationParameters validationParameters)
+        private static SecurityKey ResolveTokenDecryptionKeyFromConfig(JsonWebToken jwtToken, BaseConfiguration configuration)
+        {
+            if (jwtToken == null)
+                throw LogHelper.LogArgumentNullException(nameof(jwtToken));
+
+            if (!string.IsNullOrEmpty(jwtToken.Kid))
+            {
+                if (configuration.TokenDecryptionKeys != null)
+                {
```
TimHannMSFT: Combine if statemets?
```suggestion
            if (!string.IsNullOrEmpty(jwtToken.Kid) && configuration.TokenDecryptionKeys != null)
            {
```

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -939,7 +944,43 @@ private static string EncryptTokenPrivate(string innerJwt, EncryptingCredentials
...
+                    {
+                        if (key != null && string.Equals(key.KeyId, jwtToken.Kid, GetStringComparisonRuleIf509OrECDsa(key)))
+                            return key;
+                    }
+                }
+            }
+
+            if (!string.IsNullOrEmpty(jwtToken.X5t))
+            {
+                if (configuration.TokenDecryptionKeys != null)
```
TimHannMSFT: Combine?
```suggestion
            if (!string.IsNullOrEmpty(jwtToken.X5t) && configuration.TokenDecryptionKeys != null)
```

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -962,7 +1006,11 @@ internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken
             // 2. ResolveTokenDecryptionKey returned null
```
TimHannMSFT, line 1011: Update with config case

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -939,7 +944,43 @@ private static string EncryptTokenPrivate(string innerJwt, EncryptingCredentials
...
+                    }
+                }
+            }
+
+            return null;
+        }
+
+        internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken, TokenValidationParameters validationParameters, BaseConfiguration configuration)
         {
             IEnumerable<SecurityKey> keys = null;
```
TimHannMSFT, line 980: It would help my reading through to have a bit more comments here. Maybe, e.g.

// First we check to see if the caller has set a custom decryption resolver on TVP for the call, then we'll check to see if they've set some static decryption keys on TVP and finally we'll see if the metadata config contains a key.

As an aside, if they set all of these we'll only take the resolver logic, likewise if they set in TVP and metadata, metadata is ignored. There's some pretty complex logic here, how discoverable is it if someone set multiple which one is getting overriden from looking at logs etc?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -950,8 +991,11 @@ internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken
             else
             {
                 var key = ResolveTokenDecryptionKey(jwtToken.EncodedToken, jwtToken, validationParameters);
+                if (key == null && configuration != null)
+                    key = ResolveTokenDecryptionKeyFromConfig(jwtToken, configuration);
+
```
TimHannMSFT: Seems like we should have some testing to ensure the order of override remains the same e.g.

Prefer delegate override, then static TVP, then metadata.

Is there a reason we don't accept TVP set and metadata set? i.e. why not have both in the collection?

TimHannMSFT (changes requested):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: Addressed

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: Addressed

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: Updated comments

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
RojaEnnam: Added below comments to describe the logic and log messages to know where the key was found. 

Since we are already finding the key that matches kid, that is our key wherever it is found. so IMO we need not continue to check others(eg: config).

// First we check to see if the caller has set a custom decryption resolver on TVP for the call, if so any keys set on TVP and keys in Configuration are ignored.
 // If no custom decryption resolver set, we'll check to see if they've set some static decryption keys on TVP. If a key found, we ignore configuration.
 // If no key found in TVP, we'll check the configuration.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
```diff
@@ -305,6 +305,12 @@ public OpenIdConnectConfiguration(string json)
         [JsonIgnore]
         public override ICollection<SecurityKey> SigningKeys { get; } = new Collection<SecurityKey>();
 
+        /// <summary>
+        /// Gets the <see cref="ICollection{SecurityKey}"/> that the IdentityProvider indicates are to be used for decrypting tokens.
+        /// </summary>
+        [JsonIgnore]
+        public override ICollection<SecurityKey> TokenDecryptionKeys { get; } = new Collection<SecurityKey>();
```
TimHannMSFT: @RojaEnnam  to review if this can be removed per discussion in standup?

RojaEnnam: This is added to address the problem with serialization and ignore TokenDecryptionKeys while doing so, similar to what we did with SigningKeys

RojaEnnam: Removed it.

TimHannMSFT (approved):
LGTM

PR #2024, opened 2023-02-15
updated fix for controlling depth.

brentschmaltz:
Picks up a fix for controlling depth of parsing.


TimHannMSFT:
Since we fork newtonsoft do we normally pull in their tests too are do we assume we'd catch a regression here another way --OR-- are we simply saying that since since we don't intend to touch this code we're relying on the newtonsoft tests running in their repo?

Other than that LGTM. I think generally if we have the code in our repo we should have the tests in our repo too.

TimHannMSFT (approved):

brentschmaltz:
@TimHannMSFT i didn't see any test code for NewtonSoft in our branch.

PR #2025, opened 2023-02-16
Add BootstrapRefreshInterval into ConfigurationManager

ciaozhang:
Add a new timer for failing to fetch metadata at bootstrap where we will retry more aggressively til we get the first config (30s)


dannybtsai (commented):
nit: move line 174-176 to after 169, and remove 178-180 since they are dup of 160.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -150,15 +151,23 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+                        }
                         else
+                        {
+                            _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
                             LogHelper.LogExceptionMessage(
                                 new InvalidOperationException(
                                     LogHelper.FormatInvariant(LogMessages.IDX20806, LogHelper.MarkAsNonPII(MetadataAddress ?? "null"), LogHelper.MarkAsNonPII(ex)), ex));
+                        }
                     }
                 }
```
dannybtsai, line 218: nit: move line 174-176 to after 169, and remove 178-180 since they are dup of 160.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -50,6 +50,11 @@ public TimeSpan AutomaticRefreshInterval
         /// </summary>
         public static readonly TimeSpan DefaultRefreshInterval = new TimeSpan(0, 0, 5, 0);
 
+        /// <summary>
+        /// 30 seconds is the time interval that must pass for <see cref="RequestRefresh"/> before obtain the first configuration.
+        /// </summary>
+        public TimeSpan BootstrapRefreshInterval { get; } = new TimeSpan(0, 0, 30, 0);
```
dannybtsai: TimeSpan(0, 0, 30, 0) is 30 min, not 30 seconds.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -150,15 +151,23 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                     catch (Exception ex)
                     {
                         _fetchMetadataFailure = ex;
-                        _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
-                        if (_currentConfiguration == null) // Throw an exception if there's no configuration to return.
+
+                        if (_currentConfiguration == null)
+                        {
+                            _syncAfter = DateTimeUtil.Add(now.UtcDateTime, new[] { AutomaticRefreshInterval, RefreshInterval, BootstrapRefreshInterval }.Min());
```
dannybtsai: If RefreshInterval < BootstrapRefreshInterval, then the Min() will be less than BootstrapRefreshInterval.

For example, if RefreshInterval = 5 seconds, then _syncAfter will be now + 5 sec.

Should the logic be Math.Min(Math.Min(AutomaticRefreshInterval, RefreshInterval), BootstrapRefreshInterval)?

dannybtsai: Maybe add more tests to make sure the result is expected.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -150,15 +151,23 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
                             throw LogHelper.LogExceptionMessage(
                                 new InvalidOperationException(
                                     LogHelper.FormatInvariant(LogMessages.IDX20803, LogHelper.MarkAsNonPII(MetadataAddress ?? "null"), LogHelper.MarkAsNonPII(ex)), ex));
+                        }
                         else
+                        {
+                            _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
                             LogHelper.LogExceptionMessage(
                                 new InvalidOperationException(
                                     LogHelper.FormatInvariant(LogMessages.IDX20806, LogHelper.MarkAsNonPII(MetadataAddress ?? "null"), LogHelper.MarkAsNonPII(ex)), ex));
```
dannybtsai: Replace all "null" with Utility.Null for consistency.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
dannybtsai: If this is to allow retrieving metadata (again) in less than 5 min at the startup when the first attempt fails, can't users just set the RefreshInterval to 30 seconds instead of adding the new BootstrapRefreshInterval?

dannybtsai: Should we do this instead?
_syncAfter = now.UtcDateTime + BootstrapRefreshInterval;

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
ciaozhang: Change to TimeSpan(0,0,30)

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
ciaozhang: Use Utility.null

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
ciaozhang: resolved

TimHannMSFT:
@ciaozhang @dannybtsai please confirm we will only do this aggressive retry in the case that we get a host not reachable error. If we ever get a response from the IdP we must not perform this aggressive retry. This is ONLY if there is no network and this retry is intended to retry until outbound calls are possible. If the IdP is giving any response then we need to fall back to the same configurable backoff (e.g. 5min default IIRC)

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -185,20 +184,27 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+                        if (_currentConfiguration == null)
+                        {
+                            _syncAfter = DateTimeUtil.Add(now.UtcDateTime, BootstrapRefreshInterval);
+
+                            // Throw an exception if there's no configuration to return.
                             throw LogHelper.LogExceptionMessage(
                                 new InvalidOperationException(
-                                    LogHelper.FormatInvariant(LogMessages.IDX20803, LogHelper.MarkAsNonPII(MetadataAddress ?? "null"), LogHelper.MarkAsNonPII(ex)), ex));
+                                    LogHelper.FormatInvariant(LogMessages.IDX20803, LogHelper.MarkAsNonPII(MetadataAddress ?? Utility.Null), LogHelper.MarkAsNonPII(ex)), ex));
+                        }
```
TimHannMSFT: This won't work. We can't do the `BootstrapRefreshInterval` for any exception it can only be if the IdP doesn't return any response (where we'll infer it never reached the IdP). If, e.g. the IdP simply returns metadata that can't be parse this code as written would DDOS the IdP potentially which isn't acceptable.

TimHannMSFT (changes requested):
Need to only perform aggressive retry in the case where we don't receive a response from the IDP. If we receive any sort of response code we should use existing retry logic.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
dannybtsai: This has been updated to retry up to 4 times with the BootstrapRefreshInterval, then fall back to the regular retry logic afterwards.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
TimHannMSFT: Danny is this still relevant? Did you address it?

dannybtsai: I have addressed it in the updates.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -185,20 +188,38 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+                    {
+                        _fetchMetadataFailure = ex;
 
+                        // HttpDocumentRetriever throws an IOException on non-successful status code including HttpStatusCode.RequestTimeout and HttpStatusCode.ServiceUnavailable
+                        if (_currentConfiguration == null)
+                        {
+                            if (_bootStrapRetryCount++ < BootstrapRetryCount)
+                            {
+                                var statusCode = ex.Data[HttpDocumentRetriever.StatusCode];
+                                if (statusCode != null && (statusCode.Equals(HttpStatusCode.RequestTimeout) || statusCode.Equals(HttpStatusCode.ServiceUnavailable)))
```
TimHannMSFT: >HttpStatusCode.RequestTimeout [](http://example.com/codeflow?start=77&length=29)

I didn't realize this was the HTTP Code Request Timeout 408. I don't think we should special case that code.

The intent of the change should be if there's a network issue suspected, we should retry aggressively for a bit, otherwise let's just use the default.

I think this is adding additional complexity and not actually addressing the root thing we're targeting.

I would think, EITHER:
-Let's just have this be the behavior regardless of the exception type (retry for 2 min @30sec each)
-Let's just wait on this PR and find a better way for the exception handling to only trigger for network issues.

This shouldn't in principle block LKG, they're unrelated.

brentschmaltz: One might argue that we might just try for 2 mins, 30 sec apart for any reason.

brentschmaltz: I think the algorithm could be the automaticRefresh is 30 sec until 4 attempts OR a configuration (valid or invalid) is obtained.

RojaEnnam: +1 for having better way of recognising network issues or have 2min retrying at bootstrap.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -54,6 +54,16 @@ public TimeSpan AutomaticRefreshInterval
         /// </summary>
         public static readonly TimeSpan DefaultRefreshInterval = new TimeSpan(0, 0, 5, 0);
 
+        /// <summary>
+        /// 30 seconds is the time interval that must pass for <see cref="RequestRefresh"/> before obtain the first configuration.
```
RojaEnnam, line 63: nit:obtaining

brentschmaltz:
We replaced this with: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2052

PR #2027, opened 2023-02-21
Update Wilson logs with aka.ms pointers to known wikis

TimHannMSFT:
We currently have two wiki pages relating to common exceptions customers see:
[IDX10503](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/SecurityTokenInvalidSignatureException)
[IDX10511](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/SecurityTokenInvalidSignatureException:-IDX10511)
We should add aka.ms links to these wiki pages in the actual log message text (and we should continue to do this for other common exceptions as we add wiki pages for them).

Considered trying to have more human readable aka.ms links but if we retire these links we likely want to rethink the error messages and vice versa.


ciaozhang (approved):
LGTM

PR #2041, opened 2023-03-24
Get rid of "stale 'now' ", In case IdentityModel ConfigurationManager takes long time to get Idp Metadata

RyanLiu99:


RyanLiu99:
This is to address "stale now" issue.

This meta data is cached for x seconds to reduce the call to Idp server to get meta data.  But to make the cache works, a global lock is imposed cross requests.  Cache start time is set before calling to Id server, before waiting for the lock, not after the call returns. I call it a “stale now”. This is  a problem when call to Idp server take longer then x seconds.  Ocz x seconds has been changes from 30s to 5m from version 6.8 to version 6.27, make it less sever.

When the call to IdentiyProvider has issue, take much longer time, all requests are blocked at this lock. 

Once the call eventually returned from Idp server, SOME requests piling at the lock can just take cached data and  keep going. But if the call to Id server take longer than cache time x seconds, those requests come later than stale now + x seconds, still need call Id server again even there is fresh cached data. The global lock is blocking 2nd half requests and new requests again.  And also impose more stress on already stressed Idp server since it does not wait x seconds and call it right way again. This can get worse and worse.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -127,16 +127,16 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
             await _refreshLock.WaitAsync(cancel).ConfigureAwait(false);
             try
             {
-                if (_syncAfter <= now)
+                if (_syncAfter <= DateTimeOffset.UtcNow)
```
brentschmaltz, line 162: @RyanLiu99 can you removed the local variable "now">

brentschmaltz (changes requested):
nit: remove local variable: 'now'.

brentschmaltz (approved):

brentschmaltz:
@RyanLiu99 for some reason this is showing up with a conflict, can you rebase on top of dev.

RyanLiu99:
@brentschmaltz   I just merged dev branch into my RyanLiu-patch-1, no conflicts at all. I also merged up-to-date dev branch into RyanLiu-patch-1 in my folk, no issues either.

 Can you try again? If there still issue, I will rebase.

brentschmaltz (approved):

brentschmaltz:
@RyanLiu99 something odd is going on, git is blocking this PR.
Perhaps start with a new branch?
Last resort, i can do a manual merge, but would like to see your name on the PR.

RyanLiu99:
Ok, @brentschmaltz  I'll do it .

RyanLiu99:
A new PR created.  I will close this one.  @brentschmaltz 
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2063
Sorry take a while, I was travelling.

PR #2043, opened 2023-03-28
fix order of log parameter according to log message

WolfspiritM:
Fixes order of parameters.

The Logmessage for IDX12721 expects Issuer first and then Audience.
This switches the parameters around to match that.


ciaozhang (approved):
LGTM, thanks.

PR #2048, opened 2023-04-04
create the configuration cache in the BaseConfigurationManager class

dannybtsai:
In BaseConfigurationManager, the LKG configuration cache (_lastKnownGoodConfigurationCache) is internal but is never initialized within the BaseConfigurationManager class itself. Instead, this field is initialized by the derived class ConfigurationManager. Other derived classes cannot access this field at all and therefore have no way to initialize it (ConfigurationManager can access it because it is in an assembly that is a friend of Microsoft.IdentityModel.Tokens, where BaseConfigurationManager is defined). This caused NRE at runtime because the cache is null.

This issue is fixed by adding a new constructor to BaseConfigurationManager that initializes the cache. However, this constructor cannot use the class LastKnownGoodConfigurationCacheOptions to pass in parameters for the cache (which is what ConfigurationManager does), because LastKnownGoodConfigurationCacheOptions is defined in the assembly Microsoft.IdentityModel.Protocols and this assembly cannot be referenced from Microsoft.IdentityModel.Tokens (the assembly where BaseConfigurationManager is defined) without creating a circular reference.

The solution is to create a copy of the LastKnownGoodConfigurationCacheOptions class in Microsoft.IdentityModel.Tokens, which is named LKGConfigurationCacheOptions. The original LastKnownGoodConfigurationCacheOptions class cannot be removed because it is already exposed as a public API, so an implicit conversion from LastKnownGoodConfigurationCacheOptions to LKGConfigurationCacheOptions is defined to make it easier for users who are using LastKnownGoodConfigurationCacheOptions to pass those options to the BaseConfigurationManager constructor.


ciaozhang (approved):

src/Microsoft.IdentityModel.Protocols/Configuration/LastKnownGoodConfigurationCacheOptions.cs
```diff
@@ -46,5 +46,34 @@ public int LastKnownGoodConfigurationSizeLimit
                 _lastKnownGoodConfigurationSizeLimit = (value > 0) ? value : throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(value)));
             }
         }
+
+        /// <summary>
+        /// Defines an implicit conversion of a <see cref="LastKnownGoodConfigurationCacheOptions"/> object to an <see cref="LKGConfigurationCacheOptions"/> object.
+        /// See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/user-defined-conversion-operators for more details.
+        /// </summary>
+        /// <param name="options">the <see cref="LastKnownGoodConfigurationCacheOptions"/> object to be converted</param>
+        public static implicit operator LKGConfigurationCacheOptions(LastKnownGoodConfigurationCacheOptions options)
```
brentschmaltz: Why do we need this method?
See below.

src/Microsoft.IdentityModel.Protocols/Configuration/LastKnownGoodConfigurationCacheOptions.cs
```diff
@@ -46,5 +46,34 @@ public int LastKnownGoodConfigurationSizeLimit
...
+
+            return options.ToLKGConfigurationCacheOptions();
+        }
+
+        /// <summary>
+        /// Converts this to a <see cref="LKGConfigurationCacheOptions"/> object.
+        /// Rule CA2225 requires an alternative member named ToLKGConfigurationCacheOptions(), see https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2225.
+        /// </summary>
+        /// <returns>an <see cref="LKGConfigurationCacheOptions"/> object</returns>
+        public LKGConfigurationCacheOptions ToLKGConfigurationCacheOptions()
```
brentschmaltz: We could have LastKnownGoodConfigurationCacheOptions derive from LKGConfigurationCacheOptions then the conversion would be automatic.

LastKnownGoodConfigurationCacheOptions becomes an empty class with all properties in LKGConfigurationCacheOptions

brentschmaltz (changes requested):
Should LastKnownGoodConfigurationCacheOptions derive from LKGConfigurationCacheOptions

src/Microsoft.IdentityModel.Protocols/Configuration/LastKnownGoodConfigurationCacheOptions.cs
dannybtsai: Could this cause binary compatibility issues? The property, for example Microsoft.IdentityModel.Protocols.Configuration.LastKnownGoodConfigurationCacheOptions.LastKnownGoodConfigurationSizeLimit, is now in Microsoft.IdentityModel.Tokens.Configuration.LKGConfigurationCacheOptions.LastKnownGoodConfigurationSizeLimit.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -59,6 +55,36 @@ public TimeSpan AutomaticRefreshInterval
...
+                true);
+        }
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="options">The event queue task creation option.</param>
+        /// <param name="removeExpiredValues">Whether or not to remove expired items.</param>
+        /// <param name="taskCreationOptions">The event queue task creation option, default to None instead of LongRunning as LongRunning will always start a task on a new thread instead of ThreadPool.</param>
+        public BaseConfigurationManager(LKGConfigurationCacheOptions options, bool removeExpiredValues = true, TaskCreationOptions taskCreationOptions = TaskCreationOptions.None)
```
brentschmaltz: would it make sense to add two properties to LKGConfigurationCacheOptions.
bool RemoveExpiredValues {default true}
TaskCreationOptions TaskCreationOptions {default TaskCreationOptions.None}

We could then cascade from the default ctor as BaseConfigurationManager():this(new LKGConfigurationCacheOptions())

dannybtsai: God suggestion, thanks.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -59,6 +55,34 @@ public TimeSpan AutomaticRefreshInterval
         /// </summary>
         public static readonly TimeSpan DefaultRefreshInterval = new TimeSpan(0, 0, 5, 0);
 
+        /// <summary>
+        /// The default constructor.
+        /// </summary>
+        public BaseConfigurationManager()
```
brentschmaltz: I think you can write :this(new LKGConfigurationCacheOptions())

dannybtsai: Yeah, good suggestion, thanks.

brentschmaltz (approved):

fowl2:
Apologies for the drive by, but in future perhaps [type forwarding](https://learn.microsoft.com/en-us/dotnet/standard/assembly/type-forwarding) could be considered. :)

PR #2052, opened 2023-04-06
Add BootstrapRefreshInterval

ciaozhang:
Add a new timer with exponential backoff for failing to fetch metadata at bootstrap til we get the first config (30s)

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -29,12 +29,12 @@ namespace Microsoft.IdentityModel.Protocols
         private readonly IConfigurationValidator<T> _configValidator;
         private T _currentConfiguration;
         private Exception _fetchMetadataFailure;
-        private int _bootstrapRefreshRetryCount = 0;
+        private TimeSpan _bootstrapRefreshInterval = TimeSpan.FromSeconds(1);
 
-        /// <summary>
-        /// Static initializer for a new object. Static initializers run before the first instance of the type is created.
-        /// </summary>
-        static ConfigurationManager()
+    /// <summary>
+    /// Static initializer for a new object. Static initializers run before the first instance of the type is created.
+    /// </summary>
+    static ConfigurationManager()
         {               
         }
 
@@ -191,21 +191,32 @@ namespace Microsoft.IdentityModel.Protocols
                     catch (Exception ex)
                     {
                         _fetchMetadataFailure = ex;
-                        _bootstrapRefreshRetryCount = (_bootstrapRefreshRetryCount < int.MaxValue) ? _bootstrapRefreshRetryCount + 1 : int.MaxValue;
-
-                        if (_bootstrapRefreshRetryCount <= BootstrapRefreshMaxAttempt)
-                            _syncAfter = DateTimeUtil.Add(now.UtcDateTime, BootstrapRefreshInterval);
-                        else
-                            _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
 
                         if (_currentConfiguration == null) // Throw an exception if there's no configuration to return.
+                        {
+                            if (_bootstrapRefreshInterval < DefaultRefreshInterval)
+                            {
+                                // Adopt exponential backoff for bootstrap refresh interval if it is not longer than 5 minutes.
+                                _bootstrapRefreshInterval += _bootstrapRefreshInterval;
+                                _syncAfter = DateTimeUtil.Add(now.UtcDateTime, _bootstrapRefreshInterval);
+                            }
+                            else
+                            {
+                                _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
+                            }
+
                             throw LogHelper.LogExceptionMessage(
                                 new InvalidOperationException(
                                     LogHelper.FormatInvariant(LogMessages.IDX20803, LogHelper.MarkAsNonPII(MetadataAddress ?? "null"), LogHelper.MarkAsNonPII(ex)), ex));
+                        } 
                         else
+                        {
+                            _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
+
                             LogHelper.LogExceptionMessage(
                                 new InvalidOperationException(
                                     LogHelper.FormatInvariant(LogMessages.IDX20806, LogHelper.MarkAsNonPII(MetadataAddress ?? "null"), LogHelper.MarkAsNonPII(ex)), ex));
+                        }
                     }
                 }
 
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -196,9 +196,9 @@ namespace Microsoft.IdentityModel.Protocols
                         {
                             if (_bootstrapRefreshInterval < DefaultRefreshInterval)
                             {
-                                // Adopt exponential backoff for bootstrap refresh interval if it is not longer than 5 minutes.
+                                // Adopt exponential backoff for bootstrap refresh interval with a jitter if it is not longer than 5 minutes.
                                 _bootstrapRefreshInterval += _bootstrapRefreshInterval;
-                                _syncAfter = DateTimeUtil.Add(now.UtcDateTime, _bootstrapRefreshInterval);
+                                _syncAfter = DateTimeUtil.Add(now.UtcDateTime, _bootstrapRefreshInterval + TimeSpan.FromMilliseconds(new Random().Next((int)(_bootstrapRefreshInterval.TotalMilliseconds / 2))));
                             }
                             else
                             {
@@ -207,7 +207,7 @@ namespace Microsoft.IdentityModel.Protocols
 
                             throw LogHelper.LogExceptionMessage(
                                 new InvalidOperationException(
-                                    LogHelper.FormatInvariant(LogMessages.IDX20803, LogHelper.MarkAsNonPII(MetadataAddress ?? "null"), LogHelper.MarkAsNonPII(ex)), ex));
+                                    LogHelper.FormatInvariant(LogMessages.IDX20803, LogHelper.MarkAsNonPII(MetadataAddress ?? "null"), LogHelper.MarkAsNonPII(_syncAfter), LogHelper.MarkAsNonPII(ex)), ex));
                         } 
                         else
                         {
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -194,11 +194,12 @@ namespace Microsoft.IdentityModel.Protocols
 
                         if (_currentConfiguration == null) // Throw an exception if there's no configuration to return.
                         {
-                            if (_bootstrapRefreshInterval < DefaultRefreshInterval)
+                            if (_bootstrapRefreshInterval < RefreshInterval)
                             {
-                                // Adopt exponential backoff for bootstrap refresh interval with a jitter if it is not longer than 5 minutes.
+                                // Adopt exponential backoff for bootstrap refresh interval with a decorrelated jitter if it is not longer than the refresh interval.
+                                TimeSpan _bootstrapRefreshIntervalWithJitter = TimeSpan.FromSeconds(new Random().Next((int)_bootstrapRefreshInterval.TotalSeconds));
                                 _bootstrapRefreshInterval += _bootstrapRefreshInterval;
-                                _syncAfter = DateTimeUtil.Add(now.UtcDateTime, _bootstrapRefreshInterval + TimeSpan.FromMilliseconds(new Random().Next((int)(_bootstrapRefreshInterval.TotalMilliseconds / 2))));
+                                _syncAfter = DateTimeUtil.Add(now.UtcDateTime, _bootstrapRefreshIntervalWithJitter);
                             }
                             else
                             {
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -44,6 +44,16 @@ public TimeSpan AutomaticRefreshInterval
...
 
+        /// <summary>
+        /// 30 seconds is the time interval that must pass for <see cref="RequestRefresh"/> before obtain the first configuration.
+        /// </summary>
+        public TimeSpan BootstrapRefreshInterval { get; } = new TimeSpan(0, 0, 30);
+
+        /// <summary>
+        /// 4 is maximum number of attempts to refresh metadata with the<see cref="BootstrapRefreshInterval"/>.
+        /// </summary>
+        public int BootstrapRefreshMaxAttempt { get; internal set; } = 4;
```
TimHannMSFT: IMO we should have min/max controls here. Shouldn't let folks retry more frequently than 30s, shouldn't let bootstrap retries run for more than 10 retries. @brentschmaltz do you have any strong feelings about limits on this?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -190,7 +191,13 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                     catch (Exception ex)
                     {
                         _fetchMetadataFailure = ex;
-                        _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
+                        _bootstrapRefreshRetryCount = (_bootstrapRefreshRetryCount < int.MaxValue) ? _bootstrapRefreshRetryCount + 1 : int.MaxValue;
```
TimHannMSFT: Perhaps we should use https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked.increment

I think so long as we set a reasonable upper bound on the max retires this check isn't necessary. Just increment and move on, if someone has incremented to int max here even with only 1 sec of wait time which we'd never want to go below...that's at least 68 years and no one is going to have that kind of uptime (besides we'd all be retired by then :D)

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -190,7 +191,13 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+
+                        if (_bootstrapRefreshRetryCount <= BootstrapRefreshMaxAttempt)
+                            _syncAfter = DateTimeUtil.Add(now.UtcDateTime, BootstrapRefreshInterval);
```
TimHannMSFT: If we come in here we should log which retry # we're on at a fairly high log level, maybe max-1

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -136,12 +136,60 @@ public void FetchMetadataFailureTest()
...
+
+            // First time to fetch metadata
+            try
+            {
+                var configuration = configManager.GetConfigurationAsync().Result;
+            }
+            catch (Exception firstFetchMetadataFailure)
+            {
+                // Refresh interval is BootstrapRefreshInterval
+                var syncAfter = configManager.GetType().GetField("_syncAfter", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(configManager);
```
TimHannMSFT, line 155: Can we just expose an internal getter here for the tests? Using reflection seems unnecessarily brittle to me. If someone updates this we'll have to wait til runtime to catch the break.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -44,6 +44,16 @@ public TimeSpan AutomaticRefreshInterval
             }
         }
 
+        /// <summary>
+        /// 30 seconds is the time interval that must pass for <see cref="RequestRefresh"/> before obtain the first configuration.
+        /// </summary>
+        public TimeSpan BootstrapRefreshInterval { get; } = new TimeSpan(0, 0, 30);
```
TimHannMSFT: If we keep these public need @jmprieur to sign off. I think we can likely keep internal for first release and then add the ability to configure this in the future. If we make them settable publicly we should have a plan to quickly release a way of setting them via config.

TimHannMSFT (changes requested):
This looks pretty good. I think we need range validation on the new public controls. Given these are new configurable settings they should be exposed via the config layer too. Perhaps we can keep internal for now and then expose publicly when we're able to add via config?

TimHannMSFT:
In general would also be good to ensure the description contains not only what we're doing (I think you cover that pretty well) but why we're doing the work (i.e. because some partners don't have network at bootstrap and thus would, before this fix, be stuck not able to validate tokens for 5 minutes)

TimHannMSFT:
As discussed in standup, let's remove the public getter/setter for now and not have any new public parts AND let's switch to an exponential backoff with jitter up to 5 min

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
ciaozhang: Adopted exponential backoff

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
ciaozhang: Adoped exponential backoff

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -190,15 +191,32 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                     catch (Exception ex)
                     {
                         _fetchMetadataFailure = ex;
-                        _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
+
                         if (_currentConfiguration == null) // Throw an exception if there's no configuration to return.
+                        {
+                            if (_bootstrapRefreshInterval < DefaultRefreshInterval)
```
TimHannMSFT: This evaluation needs to be done AFTER we calculate the time for sync after. e.g. It's possible 1_bootstrapRefreshInterval` here will be 256 seconds which is less than 5 minutes but THEN we will double it on line 200 resulting in >8 minutes of wait time so it will go from ~4min of wait to ~8 min of wait, then back down to 5 minutes of wait. DefaultRefreshInterval should always be the max for the failure condition retry

TimHannMSFT: Additionally, when this is done in the correct spot, I think it should be keying off he MinimulmRefreshInterval, not the default. Reason being we don't want this more aggressive strategy to be LESS aggressive than the configured default. We seem to currently allow folks to set this to retry every second which seems a bit dangerous but at least that won't be the default.

@brentschmaltz do you think we should look into a higher min refresh time? Allowing people to refresh every second on error seems awfully agressive. NOTE: we don't have to close on that here but want to flag it.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -190,15 +191,32 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+                            {
+                                // Adopt exponential backoff for bootstrap refresh interval with a jitter if it is not longer than 5 minutes.
+                                _bootstrapRefreshInterval += _bootstrapRefreshInterval;
+                                _syncAfter = DateTimeUtil.Add(now.UtcDateTime, _bootstrapRefreshInterval + TimeSpan.FromMilliseconds(new Random().Next((int)(_bootstrapRefreshInterval.TotalMilliseconds / 2))));
```
TimHannMSFT: Have a look at the link @victorm-hernandez shared: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

It seems to suggest the "Full Jitter" or "Decorrelated Jitter" approach. I think what you have is something close to "Equal Jitter" which is mostly okay but let's make sure to cap it at the `RefreshInterval` to avoid cases where the more aggressive mode actually waits LONGER in between calls than the steady state.

TimHannMSFT (changes requested):
Would like to see us cap out at the RefreshInterval with exponential backoff til then. Current implementation has strange, unintuitive behavior from a client perspective.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -190,15 +191,32 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
                     {
                         _fetchMetadataFailure = ex;
-                        _syncAfter = DateTimeUtil.Add(now.UtcDateTime, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
+
                         if (_currentConfiguration == null) // Throw an exception if there's no configuration to return.
+                        {
+                            if (_bootstrapRefreshInterval < DefaultRefreshInterval)
+                            {
+                                // Adopt exponential backoff for bootstrap refresh interval with a jitter if it is not longer than 5 minutes.
+                                _bootstrapRefreshInterval += _bootstrapRefreshInterval;
```
victorm-hernandez, line 201: >_bootstrapRefreshInterval [](http://example.com/codeflow?start=61&length=25)

This increment is not enough to be called exponential. You can do something like Math.Pow(2, retryAttempt) #Closed

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
victorm-hernandez: >_bootstrapRefreshInterval [](http://example.com/codeflow?start=32&length=25)

We should check the type of exception, only transient HTTP errors should be included here not every single exception.  #WontFix

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
victorm-hernandez: >_bootstrapRefreshInterval.TotalMilliseconds [](http://example.com/codeflow?start=157&length=43)

The jitter could be increased as well, up to 3 secs maybe?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
victorm-hernandez, line 201: >_bootstrapRefreshInterval [](http://example.com/codeflow?start=32&length=25)

This value is never reset, we should set it back to the default value upon successful retrieval of the config.  #Closed

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -136,6 +136,54 @@ public void FetchMetadataFailureTest()
             TestUtilities.AssertFailIfErrors(context);
         }
 
+        [Fact]
+        public void BootstrapRefreshIntervalTest()
+        {
+            var context = new CompareContext($"{this}.BootstrapRefreshIntervalTest");
+
+            var documentRetriever = new HttpDocumentRetriever(HttpResponseMessageUtils.SetupHttpClientThatReturns("OpenIdConnectMetadata.json", HttpStatusCode.NotFound));
```
victorm-hernandez, line 144: Not found is not a transient error. We should not retry if this happens. This means that the URL we are using is wrong.  #WontFix

victorm-hernandez: We can have a collection of HTTP status codes that are transient and those that are permanent to decide if a retry is required. 

In addition we should also have a list of .NET exceptions for scenarios like timeouts.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
TimHannMSFT: Isn't this identical? Doubling each time is the same as (2 ** retry attempt) if we start at 1

2 ** 0 == 1 == 1
2 ** 1 == 2 == 1+1
2 ** 2 == 4 == 2+2
2 * 3 == 8 == 4+4
...

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -190,15 +191,32 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                     catch (Exception ex)
```
victorm-hernandez, line 191: >Exception [](http://example.com/codeflow?start=27&length=9)

The change here is limiting when the next refresh can be requested rather than performing another request. 

Have we discussed the possibility of doing the retry in here?  #WontFix

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
victorm-hernandez: Nevermind this, Tim pointed out this is actually equivalent. Since you are duplicating the value each time.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
TimHannMSFT: IIRC the initial fetch already has a retry so that would be multiple levels of retries. That's not to say we shouldn't do this but my preference would be to have that as a separate change.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
TimHannMSFT: I think we only bootstrap once. Once we get a config we use that stale config forever and only retry every `RefreshInterval`

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
TimHannMSFT: Agree we should take into consideration retriable vs non-retriable errors but IMO can wait til a subsequent interation.

TimHannMSFT (approved):

victorm-hernandez (approved):
:shipit:

PR #2061, opened 2023-04-25
Resolve a signing key using both KeyId and X5t in a consistent manner

dannybtsai:
An issuer signing key is resolved using the values of KeyId or X5t (of a X509SecurityKey) of a signing key. However, for a X509SecurityKey, the X5t value is not being compared consistently in the JwtTokenUtilities.ResolveTokenSigningKey() method. For example, the method checks if the validationParameters.IssuerSigningKey is a X509SecurityKey and compares the X5t value if it is, but the same logic is not applied to the validationParameters.IssuerSigningKeys.

The changes are meant to ensure consistency and reduce the number of places where values are compared.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Globalization;
+using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -441,29 +442,20 @@ private static long ParseTimeValue(JToken jToken, string claimName)
...
-            return ResolveTokenSigningKey(kid, x5t, validationParameters);
+        /// <summary>
+        /// Returns a <see cref="SecurityKey"/> to use when validating the signature of a token.
+        /// </summary>
+        /// <param name="kid">The <see cref="string"/> kid field of the token being validated</param>
+        /// <param name="x5t">The <see cref="string"/> x5t field of the token being validated</param>
+        /// <param name="configuration">The <see cref="BaseConfiguration"/> that will be used along with the <see cref="TokenValidationParameters"/> to resolve the signing key</param>
+        /// <returns>Returns a <see cref="SecurityKey"/> to use for signature validation.</returns>
+        /// <remarks>If key fails to resolve, then null is returned</remarks>
+        internal static SecurityKey ResolveTokenSigningKeyUsingConfiguration(string kid, string x5t, BaseConfiguration configuration)
```
brentschmaltz: Do we need this method or could we just call on line 445, ResolveTokenSigningKey(kid, x5t, configuration?.SigningKeys) ?? ...
Same idea for ResolveTokenSigningKeyUsingValidationParameters.

brentschmaltz (approved):

brentschmaltz (changes requested):
Code looks good. 
Build is failing, need some change.

brentschmaltz (approved):

PR #2063, opened 2023-04-25
Use up-to-date "now", in case take long time to get Metadata

RyanLiu99:
This is to address "stale now" issue.

This meta data is cached for x seconds to reduce the call to Idp server to get meta data. But to make the cache works, a global lock is imposed cross requests. Cache start time is set before calling to Id server, before waiting for the lock, not after the call returns. I call it a “stale now”. This is a problem when call to Idp server take longer then x seconds. Ocz x seconds has been changes from 30s to 5m from version 6.8 to version 6.27, make it less sever.

When the call to IdentiyProvider has issue, take much longer time, all requests are blocked at this lock.

Once the call eventually returned from Idp server, SOME requests piling at the lock can just take cached data and keep going. But if the call to Id server take longer than cache time x seconds, those requests come later than stale now + x seconds, still need call Id server again even there is fresh cached data. The global lock is blocking 2nd half requests and new requests again. And also impose more stress on already stressed Idp server since it does not wait x seconds and call it right way again. This can get worse and worse.

cc @brentschmaltz


brentschmaltz (approved):

brentschmaltz:
@RyanLiu99 thanks for this work.
We finally got it in, look for the change in the next release.

PR #2064, opened 2023-04-25
Xiao/DisarmSecurityArtifact

ciaozhang:
Remove the signatures from JWS and authentication tag from JWE when PII is on.
1. For SecurityToken: A new interface is introduced: Microsoft.IdentityModel.Logging. ISafeLogSecurityArtifact with a single method UnsafeToString(). SecurityToken will implementations (JsonWebToken, SAML, etc.) the interface and will provide an implementation of UnsafeToString().
2. For type of object that is not able modified (such as ‘string’): Struct SecurityArtifact can be used to wrap the object and provide a callback to Log appropriately.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -119,6 +119,9 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             {
                 throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14302, payload), ex));
             }
+
+            _encodedHeader = header;
+            _encodedPayload = payload;
         }
 
         internal string ActualIssuer { get; set; }
@@ -401,9 +404,6 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// </summary>
         public bool IsSigned { get; internal set; }
 
-        /// <summary>
-        ///
-        /// </summary>
         internal JsonClaimSet Payload { get; set; }
 
         /// <summary>
@@ -583,17 +583,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             EncodedToken = encodedJson;
         }
 
-        /// <summary>
-        /// Serializes the <see cref="JsonWebToken"/>.
-        /// </summary>
-        /// <returns>A string containing the header and payload in JSON format.</returns>
-        public override string ToString()
-        {
-            if (!IsEncrypted)
-                return Header.ToString() + "." + Payload.ToString();
-            else
-                return Header.ToString()  + "." + EncryptedKey + "." + InitializationVector + "." + Ciphertext;
-        }
+        /// <inheritdoc/>
+        public override string UnsafeToString() => EncodedToken;
 
 #if NET45
         /// <summary>
@@ -905,6 +896,15 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                 return _sub;
             }
+        }
+
+        /// <summary>
+        /// Returns the encoded token without signature or authentication tag.
+        /// </summary>
+        /// <returns>Encoded token string  without signature or authentication tag.</returns>
+        public override string ToString()
+        {
+            return EncodedToken.Substring(0, EncodedToken.LastIndexOf("."));
         }
 
         /// <summary>
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -901,7 +901,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// <summary>
         /// Returns the encoded token without signature or authentication tag.
         /// </summary>
-        /// <returns>Encoded token string  without signature or authentication tag.</returns>
+        /// <returns>Encoded token string without signature or authentication tag.</returns>
         public override string ToString()
         {
             return EncodedToken.Substring(0, EncodedToken.LastIndexOf("."));
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -583,6 +583,18 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             EncodedToken = encodedJson;
         }
 
+        /// <summary>
+        /// Serializes the <see cref="JsonWebToken"/>.
+        /// </summary>
+        /// <returns>A string containing the header and payload in JSON format.</returns>
+        public override string ToString()
+        {
+            if (!IsEncrypted)
+                return Header.ToString() + "." + Payload.ToString();
+            else
+                return Header.ToString()  + "." + EncryptedKey + "." + InitializationVector + "." + Ciphertext;
+        }
+
 #if NET45
         /// <summary>
         ///
@@ -590,7 +602,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         internal byte[] SignatureBytes { get; set; }
 #endif
 
-#region Claims
+        #region Claims
         /// <summary>
         /// Gets the 'value' of the 'actort' claim the payload.
         /// </summary>
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
+using System.Net;
 using System.Security.Claims;
 using System.Text;
 using System.Text.RegularExpressions;
@@ -317,10 +319,10 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         }
 
         /// <summary>
-        /// Creates a JWS (Json Web Signature) or a JWE (Json Web Encryption).
+        /// Creates a JWS(Json Web Signature).
         /// </summary>
         /// <param name="tokenDescriptor">A <see cref="SecurityTokenDescriptor"/> that contains details of contents of the token.</param>
-        /// <returns>A JWS in Compact Serialization Format or a JWE if <see cref="SecurityTokenDescriptor.EncryptingCredentials"/> is specified.</returns>
+        /// <returns>A JWS in Compact Serialization Format.</returns>
         public virtual string CreateToken(SecurityTokenDescriptor tokenDescriptor)
         {
             if (tokenDescriptor == null)
@@ -1201,7 +1203,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (unwrappedKeys.Count > 0 && exceptionStrings.Length == 0)
                 return unwrappedKeys;
             else
-                throw LogHelper.LogExceptionMessage(new SecurityTokenKeyWrapException(LogHelper.FormatInvariant(TokenLogMessages.IDX10618, keysAttempted, exceptionStrings, jwtToken)));
+                throw LogHelper.LogExceptionMessage(new SecurityTokenKeyWrapException(LogHelper.FormatInvariant(TokenLogMessages.IDX10618, keysAttempted, exceptionStrings, jwtToken.ToString())));
         }
 
         /// <summary>
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Net;
 using System.Security.Claims;
@@ -1203,7 +1202,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (unwrappedKeys.Count > 0 && exceptionStrings.Length == 0)
                 return unwrappedKeys;
             else
-                throw LogHelper.LogExceptionMessage(new SecurityTokenKeyWrapException(LogHelper.FormatInvariant(TokenLogMessages.IDX10618, keysAttempted, exceptionStrings, jwtToken.ToString())));
+                throw LogHelper.LogExceptionMessage(new SecurityTokenKeyWrapException(LogHelper.FormatInvariant(TokenLogMessages.IDX10618, keysAttempted, exceptionStrings, jwtToken)));
         }
 
         /// <summary>
@@ -1402,11 +1401,11 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             {
                 var securityToken = validationParameters.TokenReader(token, validationParameters);
                 if (securityToken == null)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenMalformedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10510, token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenMalformedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10510, LogHelper.MarkAsSecurityArtifact(token, JwtTokenUtilities.SafeLogJwtToken))));
 
                 jsonWebToken = securityToken as JsonWebToken;
                 if (jsonWebToken == null)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenMalformedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10509, typeof(JsonWebToken), securityToken.GetType(), token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenMalformedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10509, typeof(JsonWebToken), securityToken.GetType(), LogHelper.MarkAsSecurityArtifact(token, JwtTokenUtilities.SafeLogJwtToken))));
             }
             else
             {
@@ -1419,7 +1418,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 {
                     return new TokenValidationResult
                     {
-                        Exception = LogHelper.LogExceptionMessage(new SecurityTokenMalformedException(LogHelper.FormatInvariant(LogMessages.IDX14100, token), ex)),
+                        Exception = LogHelper.LogExceptionMessage(new SecurityTokenMalformedException(LogHelper.FormatInvariant(LogMessages.IDX14100, LogHelper.MarkAsSecurityArtifact(token, JwtTokenUtilities.SafeLogJwtToken), ex))),
                         IsValid = false
                     };
                 }
@@ -1538,7 +1537,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                 if (validationParameters.SignatureValidator != null || validationParameters.SignatureValidatorUsingConfiguration != null)
                 {
-                    var validatedToken = ValidateSignatureUsingDelegates(jsonWebToken.EncodedToken, validationParameters, configuration);
+                    var validatedToken = ValidateSignatureUsingDelegates(jsonWebToken, validationParameters, configuration);
                     tokenValidationResult = await ValidateTokenPayloadAsync(validatedToken, validationParameters, configuration).ConfigureAwait(false);
                     Validators.ValidateIssuerSecurityKey(validatedToken.SigningKey, validatedToken, validationParameters, configuration);
                 }
@@ -1607,32 +1606,32 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             }
         }
 
-        private static JsonWebToken ValidateSignatureUsingDelegates(string token, TokenValidationParameters validationParameters, BaseConfiguration configuration)
+        private static JsonWebToken ValidateSignatureUsingDelegates(JsonWebToken jsonWebToken, TokenValidationParameters validationParameters, BaseConfiguration configuration)
         {
             if (validationParameters.SignatureValidatorUsingConfiguration != null)
             {
-                var validatedToken = validationParameters.SignatureValidatorUsingConfiguration(token, validationParameters, configuration);
+                var validatedToken = validationParameters.SignatureValidatorUsingConfiguration(jsonWebToken.EncodedToken, validationParameters, configuration);
                 if (validatedToken == null)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10505, token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10505, jsonWebToken)));
 
                 if (!(validatedToken is JsonWebToken validatedJsonWebToken))
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10506, LogHelper.MarkAsNonPII(typeof(JsonWebToken)), LogHelper.MarkAsNonPII(validatedToken.GetType()), token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10506, LogHelper.MarkAsNonPII(typeof(JsonWebToken)), LogHelper.MarkAsNonPII(validatedToken.GetType()), jsonWebToken)));
 
                 return validatedJsonWebToken;
             }
             else if (validationParameters.SignatureValidator != null)
             {
-                var validatedToken = validationParameters.SignatureValidator(token, validationParameters);
+                var validatedToken = validationParameters.SignatureValidator(jsonWebToken.EncodedToken, validationParameters);
                 if (validatedToken == null)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10505, token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10505, jsonWebToken)));
 
                 if (!(validatedToken is JsonWebToken validatedJsonWebToken))
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10506, LogHelper.MarkAsNonPII(typeof(JsonWebToken)), LogHelper.MarkAsNonPII(validatedToken.GetType()), token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10506, LogHelper.MarkAsNonPII(typeof(JsonWebToken)), LogHelper.MarkAsNonPII(validatedToken.GetType()), jsonWebToken)));
 
                 return validatedJsonWebToken;
             }
 
-            throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10505, token)));
+            throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10505, jsonWebToken)));
         }
 
         private static JsonWebToken ValidateSignatureAndIssuerSecurityKey(JsonWebToken jsonWebToken, TokenValidationParameters validationParameters, BaseConfiguration configuration)
@@ -1736,7 +1735,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                         if (ValidateSignature(jwtToken, key, validationParameters))
 #endif
                         {
-                            LogHelper.LogInformation(TokenLogMessages.IDX10242, jwtToken.EncodedToken);
+                            LogHelper.LogInformation(TokenLogMessages.IDX10242, jwtToken);
                             jwtToken.SigningKey = key;
                             return jwtToken;
                         }
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -25,6 +25,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// </summary>
     public class JwtTokenUtilities
     {
+        private string _unrecognizedEncodedToken = "UnrecognizedEncodedToken";
+
         /// <summary>
         /// Regex that is used to figure out if a token is in JWS format.
         /// </summary>
@@ -434,15 +436,15 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (obj == null)
                 return string.Empty;
 
-            // not a string
+            // not a string, we do not know how to sanitize so we return a String which represents the object instance
             if (!(obj is string token))
                 return obj.ToString();
 
             int lastDot = token.LastIndexOf(".");
 
-            // no dots, maybe not a JWT
+            // no dots, not a JWT, we do not know how to sanitize so we return UnrecognizedEncodedToken
             if (lastDot == -1)
-                return typeof(string).ToString();
+                return _unrecognizedEncodedToken;
 
             return token.Substring(0, lastDot);
         }
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -250,13 +250,13 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         private static void ValidateDecryption(JwtTokenDecryptionParameters decryptionParameters, bool decryptionSucceeded, bool algorithmNotSupportedByCryptoProvider, StringBuilder exceptionStrings, StringBuilder keysAttempted)
         {
             if (!decryptionSucceeded && keysAttempted.Length > 0)
-                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10603, keysAttempted, exceptionStrings, PrepareSecurityArtifact(decryptionParameters.EncodedToken))));
+                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10603, keysAttempted, exceptionStrings, LogHelper.MarkAsSecurityArtifact(decryptionParameters.EncodedToken, SafeLogJwtToken))));
 
             if (!decryptionSucceeded && algorithmNotSupportedByCryptoProvider)
                 throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10619, LogHelper.MarkAsNonPII(decryptionParameters.Alg), LogHelper.MarkAsNonPII(decryptionParameters.Enc))));
 
             if (!decryptionSucceeded)
-                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10609, PrepareSecurityArtifact(decryptionParameters.EncodedToken))));
+                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10609, LogHelper.MarkAsSecurityArtifact(decryptionParameters.EncodedToken, SafeLogJwtToken))));
         }
 
         private static byte[] DecryptToken(CryptoProviderFactory cryptoProviderFactory, SecurityKey key, string encAlg, byte[] ciphertext, byte[] headerAscii, byte[] initializationVector, byte[] authenticationTag)
@@ -429,15 +429,22 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             throw LogHelper.LogExceptionMessage(new FormatException(LogHelper.FormatInvariant(LogMessages.IDX14300, LogHelper.MarkAsNonPII(claimName), jToken.ToString(), LogHelper.MarkAsNonPII(typeof(long)))));
         }
 
-        internal static string PrepareSecurityArtifact(string securityArtifact)
+        internal static string SafeLogJwtToken(object obj)
         {
-            if (IdentityModelEventSource.ShowPII)
-            {
-                if (!string.IsNullOrEmpty(securityArtifact) && (RegexJws.IsMatch(securityArtifact) || RegexJwe.IsMatch(securityArtifact)))
-                    return securityArtifact.Substring(0, securityArtifact.LastIndexOf("."));
-            }
+            if (obj == null)
+                return string.Empty;
 
-            return securityArtifact;
+            // not a string
+            if (!(obj is string token))
+                return obj.ToString();
+
+            int lastDot = token.LastIndexOf(".");
+
+            // no dots, maybe not a JWT
+            if (lastDot == -1)
+                return typeof(string).ToString();
+
+            return token.Substring(0, lastDot);
         }
 
         /// <summary>
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -438,8 +438,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
             // not a string, we do not know how to sanitize so we return a String which represents the object instance
             if (!(obj is string token))
-                return obj.ToString();
-
+                return obj.GetType().ToString();
+ 
             int lastDot = token.LastIndexOf(".");
 
             // no dots, not a JWT, we do not know how to sanitize so we return UnrecognizedEncodedToken
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -250,13 +250,13 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         private static void ValidateDecryption(JwtTokenDecryptionParameters decryptionParameters, bool decryptionSucceeded, bool algorithmNotSupportedByCryptoProvider, StringBuilder exceptionStrings, StringBuilder keysAttempted)
         {
             if (!decryptionSucceeded && keysAttempted.Length > 0)
-                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10603, keysAttempted, exceptionStrings, decryptionParameters.EncodedToken)));
+                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10603, keysAttempted, exceptionStrings, PrepareSecurityArtifact(decryptionParameters.EncodedToken))));
 
             if (!decryptionSucceeded && algorithmNotSupportedByCryptoProvider)
                 throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10619, LogHelper.MarkAsNonPII(decryptionParameters.Alg), LogHelper.MarkAsNonPII(decryptionParameters.Enc))));
 
             if (!decryptionSucceeded)
-                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10609, decryptionParameters.EncodedToken)));
+                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10609, PrepareSecurityArtifact(decryptionParameters.EncodedToken))));
         }
 
         private static byte[] DecryptToken(CryptoProviderFactory cryptoProviderFactory, SecurityKey key, string encAlg, byte[] ciphertext, byte[] headerAscii, byte[] initializationVector, byte[] authenticationTag)
@@ -427,6 +427,17 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             }
 
             throw LogHelper.LogExceptionMessage(new FormatException(LogHelper.FormatInvariant(LogMessages.IDX14300, LogHelper.MarkAsNonPII(claimName), jToken.ToString(), LogHelper.MarkAsNonPII(typeof(long)))));
+        }
+
+        internal static string PrepareSecurityArtifact(string securityArtifact)
+        {
+            if (IdentityModelEventSource.ShowPII)
+            {
+                if (!string.IsNullOrEmpty(securityArtifact) && (RegexJws.IsMatch(securityArtifact) || RegexJwe.IsMatch(securityArtifact)))
+                    return securityArtifact.Substring(0, securityArtifact.LastIndexOf("."));
+            }
+
+            return securityArtifact;
         }
 
         /// <summary>
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -25,7 +25,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// </summary>
     public class JwtTokenUtilities
     {
-        private string _unrecognizedEncodedToken = "UnrecognizedEncodedToken";
+        private const string _unrecognizedEncodedToken = "UnrecognizedEncodedToken";
 
         /// <summary>
         /// Regex that is used to figure out if a token is in JWS format.
--- a/src/Microsoft.IdentityModel.Logging/ISafeLogSecurityArtifact.cs
+++ b/src/Microsoft.IdentityModel.Logging/ISafeLogSecurityArtifact.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// Interface that provides an unsafe method to log a security artifact.
+    /// </summary>
+    public interface ISafeLogSecurityArtifact
+    {
+        /// <summary>
+        /// Returns a string that represents the complete security artifact.
+        /// This may include sensitive information and should only be used for logging purposes.
+        /// </summary>
+        string UnsafeToString();
+    }
+}
+
--- a/src/Microsoft.IdentityModel.Logging/ISafeLogSecurityArtifact.cs
+++ b/src/Microsoft.IdentityModel.Logging/ISafeLogSecurityArtifact.cs
@@ -6,11 +6,14 @@ namespace Microsoft.IdentityModel.Logging
     /// <summary>
     /// Interface that provides an unsafe method to log a security artifact.
     /// </summary>
+    /// <remarks>
+    /// SecurityToken and encoded token are considered as SecurityArtifacts.
+    /// </remarks>
     public interface ISafeLogSecurityArtifact
     {
         /// <summary>
         /// Returns a string that represents the complete security artifact.
-        /// This may include sensitive information and should only be used for logging purposes.
+        /// This may include sensitive information and should only be used for debugging purposes.
         /// </summary>
         string UnsafeToString();
     }
--- a/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
+++ b/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
@@ -36,6 +36,11 @@ namespace Microsoft.IdentityModel.Logging
         public static bool ShowPII { get; set; } = false;
 
         /// <summary>
+        /// Flag which indicates whether or not armed security artifact is shown in logs when <see cref="ShowPII"/> is set to true. False by default.
+        /// </summary>
+        public static bool LogCompleteSecurityArtifact { get; set; } = false;
+
+        /// <summary>
         /// String that is used in place of any arguments to log messages if the 'ShowPII' flag is set to false.
         /// </summary>
         public static string HiddenPIIString { get; } = "[PII of type '{0}' is hidden. For more details, see https://aka.ms/IdentityModel/PII.]";
--- a/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
+++ b/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
@@ -36,7 +36,7 @@ namespace Microsoft.IdentityModel.Logging
         public static bool ShowPII { get; set; } = false;
 
         /// <summary>
-        /// Flag which indicates whether or not armed security artifact is shown in logs when <see cref="ShowPII"/> is set to true. False by default.
+        /// Flag which indicates whether or not complete <see cref="SecurityArtifact"/> is shown in logs when <see cref="ShowPII"/> is set to true. False by default.
         /// </summary>
         public static bool LogCompleteSecurityArtifact { get; set; } = false;
 
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -2,10 +2,12 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Data;
 using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.Tracing;
 using System.Globalization;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Microsoft.IdentityModel.Abstractions;
 
 namespace Microsoft.IdentityModel.Logging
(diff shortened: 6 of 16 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -579,14 +579,26 @@ private void ReadToken(string encodedJson)
...
+        /// <summary>
+        /// Serializes the <see cref="JsonWebToken"/>.
+        /// </summary>
+        /// <returns>A string containing the header and payload in JSON format.</returns>
+        public override string ToString()
+        {
+            if (!IsEncrypted)
+                return Header.ToString() + "." + Payload.ToString();
+            else
+                return Header.ToString()  + "." + EncryptedKey + "." + InitializationVector + "." + Ciphertext;
```
TimHannMSFT: ####                 return Header.ToString()  + "." + EncryptedKey + "." + InitializationVector + "." + Ciphertext;

---

Have you evaluated StringBuilder in this case? Usually when you're concatonating more than ~3 strings StringBuilder wins. #Resolved

TimHannMSFT:
@ciaozhang can you add more details to the description as well as a section which can be copied directly into the release notes? I'm happy to help construct if you need help. @jmprieur would also likely be able to help.

---
In reply to: [1524185285](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2064#issuecomment-1524185285) [](http://example.com/codeflow?ancestors=1524185285)

---
In reply to: [1524185285](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2064#issuecomment-1524185285) [](http://example.com/codeflow?ancestors=1524185285)

---
In reply to: [1524185285](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2064#issuecomment-1524185285) [](http://example.com/codeflow?ancestors=1524185285)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
TimHannMSFT: For three string parts:
> |                          Method |       Mean |     Error |    StdDev |     Median |   Gen0 | Allocated |
|-------------------------------- |-----------:|----------:|----------:|-----------:|-------:|----------:|
| ConcatStringsUsingStringBuilder | 27.4693 ns | 1.1695 ns | 3.2793 ns | 26.5547 ns | 0.0242 |     152 B |
|          ConcatStringsUsingPlus |  0.0000 ns | 0.0000 ns | 0.0000 ns |  0.0000 ns |      - |         - |

Interesting, seems like just plussing them together still wins at 7....Benchmark.net isn't able to distinguish a difference between that and an empty method.
> |                           Method |       Mean |     Error |    StdDev |   Gen0 | Allocated |
|--------------------------------- |-----------:|----------:|----------:|-------:|----------:|
| ConcatStringsUsingStringBuilder7 | 61.3046 ns | 1.2456 ns | 1.4828 ns | 0.0459 |     288 B |
|          ConcatStringsUsingPlus7 |  0.0000 ns | 0.0000 ns | 0.0000 ns |      - |         - |

I forced in a ToString() to make sure it was actually sampling but got a similar result so seems like your code is good here.
> |                           Method |     Mean |    Error |    StdDev |   Median |   Gen0 | Allocated |
|--------------------------------- |---------:|---------:|----------:|---------:|-------:|----------:|
| ConcatStringsUsingStringBuilder7 | 97.53 ns | 3.743 ns | 10.370 ns | 93.10 ns | 0.0790 |     496 B |
|          ConcatStringsUsingPlus7 | 12.99 ns | 0.601 ns |  1.752 ns | 12.45 ns | 0.0217 |     136 B |

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -579,14 +579,26 @@ private void ReadToken(string encodedJson)
...
         }
 
+        /// <summary>
+        /// Serializes the <see cref="JsonWebToken"/>.
+        /// </summary>
+        /// <returns>A string containing the header and payload in JSON format.</returns>
+        public override string ToString()
+        {
+            if (!IsEncrypted)
+                return Header.ToString() + "." + Payload.ToString();
```
brentschmaltz: ####                 return Header.ToString() + "." + Payload.ToString();

---

In most cases EncodedHeader and EncodedPayload is what you want here.
If you fix this bug: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2065 then you will always get proper value. #Resolved

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1064,7 +1066,7 @@ internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken
             if (unwrappedKeys.Count > 0 && exceptionStrings.Length == 0)
                 return unwrappedKeys;
             else
-                throw LogHelper.LogExceptionMessage(new SecurityTokenKeyWrapException(LogHelper.FormatInvariant(TokenLogMessages.IDX10618, keysAttempted, exceptionStrings, jwtToken)));
+                throw LogHelper.LogExceptionMessage(new SecurityTokenKeyWrapException(LogHelper.FormatInvariant(TokenLogMessages.IDX10618, keysAttempted, exceptionStrings, jwtToken.ToString())));
```
TimHannMSFT: ####                 throw LogHelper.LogExceptionMessage(new SecurityTokenKeyWrapException(LogHelper.FormatInvariant(TokenLogMessages.IDX10618, keysAttempted, exceptionStrings, jwtToken.ToString())));

---

Does this change do anything? I believe ToString is called by default anyways in string.Format which this eventually calls. #Resolved

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1240,11 +1242,11 @@ private static TokenValidationResult ReadToken(string token, TokenValidationPara
             {
                 var securityToken = validationParameters.TokenReader(token, validationParameters);
                 if (securityToken == null)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10510, token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10510, JwtTokenUtilities.PrepareSecurityArtifact(token))));
```
TimHannMSFT: ####                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10510, JwtTokenUtilities.PrepareSecurityArtifact(token))));

---

I would prefer we followed the pattern used for PII. Something like, `LogHelper.MarkSecurityArtifact`. This has a couple advantages in my mind
1. It reuses a pattern which is already established in code so it makes it easier to understand what's going on.
2. It allows log level to be taken into account before performing any string operations. We can win from a performance standpoint if we're only doing these string operations (or regEx searches) when we actually go to log. If there's a log informational with a token. Code as written would still do the string manipulation even if the caller had only enabled error logs. #Resolved

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// An internal structure that is used to mark an argument as NonPII.
+    /// Arguments wrapped with a NonPII structure will be considered as NonPII in the message logging process.
+    /// </summary>
+    internal struct SecurityArtifact
```
TimHannMSFT: ####     internal struct SecurityArtifact

---

Not used, but should be! #Resolved

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        /// Argument wrapped with a <see cref="NonPII"/> structure is considered as NonPII in the message logging process.
```
TimHannMSFT: ####         /// Argument wrapped with a <see cref="NonPII"/> structure is considered as NonPII in the message logging process.

---

Copy/paste artifacts left over, please clean up #Resolved

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -0,0 +1,35 @@
...
+            Argument = argument;
+        }
+
+        /// <summary>
+        /// Returns a string that represents the <see cref="Argument"/>.
+        /// </summary>
+        /// <returns><c>Null</c> if the <see cref="Argument"/> is <c>null</c>, otherwise calls <see cref="System.ValueType.ToString()"/> method of the <see cref="Argument"/>.</returns>
+        public override string ToString()
+        {
+            return Argument?.ToString() ?? "Null";
```
TimHannMSFT: ####             return Argument?.ToString() ?? "Null";

---

nit: We mix "Null" and "null" throughout the Wilson code base. We should standarize to just one or the other. I'd prefer "null" since that's the casing of in .NET. #Resolved

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -346,7 +346,7 @@ protected virtual void ValidateIdToken(OpenIdConnectProtocolValidationContext va
                 }
                 catch (Exception ex)
                 {
-                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21313, validationContext.ValidatedIdToken), ex));
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21313, validationContext.ValidatedIdToken.ToString()), ex));
```
TimHannMSFT: Again, it's better to wait until the last minute to do string operations for performance reasons. If our logging helper methods aren't able to handle this we should update the helper/utility methods not each instance where we call.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1141,7 +1141,7 @@ private ClaimsPrincipal ValidateTokenPayload(JwtSecurityToken jwtToken, TokenVal
             if (validationParameters.SaveSigninToken)
                 identity.BootstrapContext = jwtToken.RawData;
 
-            LogHelper.LogInformation(TokenLogMessages.IDX10241, jwtToken.RawData);
+            LogHelper.LogInformation(TokenLogMessages.IDX10241, JwtTokenUtilities.PrepareSecurityArtifact(jwtToken.RawData));
```
brentschmaltz: ####             LogHelper.LogInformation(TokenLogMessages.IDX10241, JwtTokenUtilities.PrepareSecurityArtifact(jwtToken.RawData));

---

Could this be SecurityToken.ToString()? #Resolved

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1151,7 +1151,7 @@ private ClaimsPrincipal CreateClaimsPrincipalFromToken(JwtSecurityToken jwtToken
...
```
brentschmaltz: ####             LogHelper.LogInformation(TokenLogMessages.IDX10241, JwtTokenUtilities.PrepareSecurityArtifact(jwtToken.RawData));

---

Could this be SecurityToken.ToString()? #Resolved

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -356,33 +356,33 @@ protected virtual void ValidateIdToken(OpenIdConnectProtocolValidationContext va
...
 
                 if (!idToken.Payload.Iat.HasValue)
-                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, LogHelper.MarkAsNonPII(JwtRegisteredClaimNames.Iat.ToLowerInvariant()), idToken)));
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, LogHelper.MarkAsNonPII(JwtRegisteredClaimNames.Iat.ToLowerInvariant()), idToken.ToString())));
 
                 if (idToken.Payload.Iss == null)
-                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, LogHelper.MarkAsNonPII(JwtRegisteredClaimNames.Iss.ToLowerInvariant()), idToken)));
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, LogHelper.MarkAsNonPII(JwtRegisteredClaimNames.Iss.ToLowerInvariant()), idToken.ToString())));
```
TimHannMSFT: Same comment here about agressive ToString() calls. Best to avoid. I'm curious if this has any functional difference at runtime.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1291,7 +1291,7 @@ private JwtSecurityToken ValidateSignature(string token, JwtSecurityToken jwtTok
             if (string.IsNullOrEmpty(jwtToken.RawSignature))
             {
                 if (validationParameters.RequireSignedTokens)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10504, token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10504, JwtTokenUtilities.PrepareSecurityArtifact(token))));
```
brentschmaltz: ####                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10504, JwtTokenUtilities.PrepareSecurityArtifact(token))));

---

Could you use the JwtSecurityToken passed in the parameter? #Resolved

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1347,7 +1347,7 @@ private JwtSecurityToken ValidateSignature(string token, JwtSecurityToken jwtTok
                     {
                         if (ValidateSignature(encodedBytes, signatureBytes, key, jwtToken.Header.Alg, jwtToken, validationParameters))
                         {
-                            LogHelper.LogInformation(TokenLogMessages.IDX10242, token);
+                            LogHelper.LogInformation(TokenLogMessages.IDX10242, JwtTokenUtilities.PrepareSecurityArtifact(token));
```
brentschmaltz: ####                             LogHelper.LogInformation(TokenLogMessages.IDX10242, JwtTokenUtilities.PrepareSecurityArtifact(token));

---

Could you use the JwtSecurityToken passed in the parameter? #Resolved

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1573,7 +1575,7 @@ private static JsonWebToken ValidateSignature(JsonWebToken jwtToken, TokenValida
                         if (ValidateSignature(jwtToken, key, validationParameters))
 #endif
                         {
-                            LogHelper.LogInformation(TokenLogMessages.IDX10242, jwtToken.EncodedToken);
+                            LogHelper.LogInformation(TokenLogMessages.IDX10242, JwtTokenUtilities.PrepareSecurityArtifact(jwtToken.EncodedToken));
```
brentschmaltz: ####                             LogHelper.LogInformation(TokenLogMessages.IDX10242, JwtTokenUtilities.PrepareSecurityArtifact(jwtToken.EncodedToken));

---

Could you use jwtToken.ToString() #Resolved

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -356,33 +356,33 @@ protected virtual void ValidateIdToken(OpenIdConnectProtocolValidationContext va
...
 
                 // sub is required in OpenID spec; but we don't want to block valid idTokens provided by some identity providers
                 if (RequireSub && (string.IsNullOrWhiteSpace(idToken.Payload.Sub)))
-                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, LogHelper.MarkAsNonPII(JwtRegisteredClaimNames.Sub.ToLowerInvariant()), idToken)));
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolException(LogHelper.FormatInvariant(LogMessages.IDX21314, LogHelper.MarkAsNonPII(JwtRegisteredClaimNames.Sub.ToLowerInvariant()), idToken.ToString())));
```
TimHannMSFT: A couple points about the ToString() usage in addtion to what I've said above:
1. Ideally we have one way people mark tokens at log time. Not great devEx to have one way if it's a string, one way if it's a `JwtSecurityToken`, etc. Easier to write a method once to take that cognative burden from the dev making the change.
2. Do you know why we have to rebuild this string? Shouldn't we store the base64 string for the token (at least in some cases) when it's passed in the constructor? @brentschmaltz do you know why this is the behavior?
3. Somewhat related to (2) above ideally we'd be disarming the string in the same way as for other parts of the code. That we have multiple places rebuilding tokens from their parts in code is a signal we may have more code duplication than we need.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
TimHannMSFT: IMO neither of those is the ideal way. We want to delay string operations as log as possible. Especially for informational logs which may not ever actually see the light of day. Those would just be wasted CPU cycles.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
TimHannMSFT: Would be nice if the method header described how these were different. Coming to this cold, it's not clear why both would exist.

TimHannMSFT (changes requested):
Thanks for the changes so far Xiao.

You mentioned to me you also had two questions for Brent/Team? Can you please add those as comments in the PR and we can resolve there?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
victorm-hernandez: @TimHannMSFT , I am curious, what did you use to do your benchmark? I would like to use that tool.

About the string concatenation at hand, I found that in these type of scenarios where the string concatenation happens in a single line, the compiler adds the required optimizations such that there are no immutable strings waste here.

In this case the optimization looks like
``` csharp
        string[] array = new string[7];
        array[0] = header.ToString();
        array[1] = ".";
        array[2] = EncryptedKey;
        array[3] = ".";
        array[4] = InitializationVector;
        array[5] = ".";
        array[6] = Ciphertext;
        return string.Concat(array);
```

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
sruke: That's correct. LogHelper.FormatInvariant internally invokes string.Format where ToString is called by default

victorm-hernandez:
@ciaozhang minor observation, the title of the PR is very descriptive but the commit message is not. Let's try to keep our commit messages readable and with enough context for anyone to understand a change. 

---
In reply to: [1526562219](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2064#issuecomment-1526562219) [](http://example.com/codeflow?ancestors=1526562219)

---
In reply to: [1526562219](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2064#issuecomment-1526562219) [](http://example.com/codeflow?ancestors=1526562219)

---
In reply to: [1526562219](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2064#issuecomment-1526562219) [](http://example.com/codeflow?ancestors=1526562219)

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -431,6 +431,17 @@ private static long ParseTimeValue(JToken jToken, string claimName)
             throw LogHelper.LogExceptionMessage(new FormatException(LogHelper.FormatInvariant(LogMessages.IDX14300, LogHelper.MarkAsNonPII(claimName), jToken.ToString(), LogHelper.MarkAsNonPII(typeof(long)))));
         }
 
+        internal static string PrepareSecurityArtifact(string securityArtifact)
+        {
+            if (IdentityModelEventSource.ShowPII)
```
victorm-hernandez: >ShowPII [](http://example.com/codeflow?start=41&length=7)

####             if (IdentityModelEventSource.ShowPII)

---

[BUG] Isn't this the other way around? ShowPII = true means we dont modify the security artifact, otherwise we remove the signature. 

If you agree we should negate this if statement. #Resolved

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JwtTokenUtilitiesTests.cs
```diff
@@ -13,6 +16,52 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
 {
     public class JwtTokenUtilitiesTests
     {
+        // Used for formatting a message for testing with one parameter.
+        private const string TestMessageOneParam = "This is the parameter: '{0}'.";
+
+        [Fact]
+        public void LogExceptionsWithStringJwe_JsonWebTokenHandler()
```
victorm-hernandez: >LogExceptionsWithStringJwe_JsonWebTokenHandler [](http://example.com/codeflow?start=20&length=46)

####         public void LogExceptionsWithStringJwe_JsonWebTokenHandler()

---

You don't validate the negative case, when ShowPII is set to false. This could be done on the same test. #Closed

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
TimHannMSFT: https://github.com/dotnet/BenchmarkDotNet

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -431,6 +431,17 @@ private static long ParseTimeValue(JToken jToken, string claimName)
             throw LogHelper.LogExceptionMessage(new FormatException(LogHelper.FormatInvariant(LogMessages.IDX14300, LogHelper.MarkAsNonPII(claimName), jToken.ToString(), LogHelper.MarkAsNonPII(typeof(long)))));
         }
 
+        internal static string PrepareSecurityArtifact(string securityArtifact)
```
victorm-hernandez: >PrepareSecurityArtifact [](http://example.com/codeflow?start=31&length=23)

####         internal static string PrepareSecurityArtifact(string securityArtifact)

---

nit: In my opinion this method doesn't describe very well what its done, may I suggest something on the line of SanitizeForLog or RemovePII #Resolved

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
dannybtsai: There is a Null constant defined in the Utility class;

public const string Null = "null";

We should use it.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
TimHannMSFT: I think part of the reason for confusion here is there's a lot of conflation of logic going on. This is set to only execute if PII is enabled because that's the only time the string will be logged (otherwise it's scrubbed). This can lead to difficult to debug problems down the line though. Also why I was favoring/suggesting following the pattern of using the decoration pattern (e.g. MarkAsNonPII) so the logic to scrub log lines is consolidated and not spread out which will make it easier to understand.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
victorm-hernandez: In addition it seems this method is better suited for the Microsoft.IdentityModel.Logging.LogHelper class.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
victorm-hernandez: Thanks!

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <param name="argument">An argument that is considered as NonPII.</param>
+        public SecurityArtifact(object argument)
+        {
+            Argument = argument;
+        }
+
+        /// <summary>
+        /// Returns a string that represents the <see cref="Argument"/>.
+        /// </summary>
+        /// <returns><c>Null</c> if the <see cref="Argument"/> is <c>null</c>, otherwise calls <see cref="System.ValueType.ToString()"/> method of the <see cref="Argument"/>.</returns>
```
TimHannMSFT: ####         /// <returns><c>Null</c> if the <see cref="Argument"/> is <c>null</c>, otherwise calls <see cref="System.ValueType.ToString()"/> method of the <see cref="Argument"/>.</returns>

---

supernit: we use
```
<c>null</c>
```
 and
```
<see langword="null" />
```

We should standardize. I prefer the later because I think it produces better help syntax and I believe it's what .NET libraries do as well.
 #Resolved

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
victorm-hernandez: I completely agree on the decorator pattern, not using it will most likely require additional work to adapt to the other helper methods. 

If I understood your explanation we only cleanup the string when show PII is enabled so we we never leak the signature in the logs? If that is the case we shouldnt even have two execution flows and always strip the value out.

This method belongs to the LogHelper as well.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
victorm-hernandez: Given that this has no change on the output, and that changing the type of the parameter from JsonWebToken to string will prevent the LogHelper class to be able to handle this type in specific like it does for NonPII or Exception, I would not change it.

victorm-hernandez: Following the same train of thought, this change could be implemented entirely on the LogHelper.FormatInvariant by checking for the type of the object to be SecurityToken and if it is sanitize the output in place.

victorm-hernandez (dismissed):
:clock1:

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
ciaozhang: Changed to  "null"

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
dannybtsai: @TimHannMSFT/@victorm-hernandez  StringBuilder should be used in for loops or dynamics cases. I believe the compiler generates code similar to the string.Concat method where it calculates the total length of all strings first, allocate a memory once then copy all values. It is efficient because it does not need to re-allocate the buffer.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1238,13 +1240,13 @@ private static TokenValidationResult ReadToken(string token, TokenValidationPara
             JsonWebToken jsonWebToken = null;
             if (validationParameters.TokenReader != null)
             {
-                var securityToken = validationParameters.TokenReader(token, validationParameters);
+                var securityToken = validationParameters.TokenReader(token, validationParameters);
```
jmprieur: ####                 var securityToken = validationParameters.TokenReader(token, validationParameters); 

---

extra space #Closed

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -579,14 +582,41 @@ private void ReadToken(string encodedJson)
...
+                sb.Append(EncodedPayload);
+            }
+            else
+            {
+                sb.Append(EncodedHeader);
+                sb.Append('.');
+                sb.Append(EncryptedKey);
+                sb.Append('.');
+                sb.Append(InitializationVector);
+                sb.Append('.');
```
jmprieur: ####                 sb.Append('.');

---

There are 4 dots in a JWT? I thought there were only 3? #Closed

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -431,6 +431,17 @@ private static long ParseTimeValue(JToken jToken, string claimName)
...
+        {
+            if (IdentityModelEventSource.ShowPII)
+            {
+                if (!string.IsNullOrEmpty(securityArtifact) && (RegexJws.IsMatch(securityArtifact) || RegexJwe.IsMatch(securityArtifact)))
+                    return securityArtifact.Substring(0, securityArtifact.LastIndexOf("."));
```
jmprieur: ####                     return securityArtifact.Substring(0, securityArtifact.LastIndexOf("."));

---

ah! back to the question 3 or 4? #Resolved

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
```diff
@@ -35,6 +35,11 @@ private IdentityModelEventSource()
         /// </summary>
         public static bool ShowPII { get; set; } = false;
 
+        /// <summary>
+        /// Flag which indicates whether or not armed security artifact is shown in logs when <see cref="ShowPII"/> is set to true. False by default.
+        /// </summary>
+        public static bool ShowArmedSecurityArtifact { get; set; } = false;
```
jmprieur: ####         public static bool ShowArmedSecurityArtifact { get; set; } = false;

---

What is the end to end for libraries using it?
How does it integrate with the IIdentityLogger? #Resolved

src/Microsoft.IdentityModel.Logging/NonPII.cs
```diff
@@ -26,10 +26,10 @@ public NonPII(object argument)
         /// <summary>
         /// Returns a string that represents the <see cref="Argument"/>.
         /// </summary>
-        /// <returns><c>Null</c> if the <see cref="Argument"/> is <c>null</c>, otherwise calls <see cref="System.ValueType.ToString()"/> method of the <see cref="Argument"/>.</returns>
+        /// <returns><c>Null</c> if the <see cref="Argument"/> is <see langword="null"/>, otherwise calls <see cref="System.ValueType.ToString()"/> method of the <see cref="Argument"/>.</returns>
         public override string ToString()
         {
-            return Argument?.ToString() ?? "Null";
+            return Argument?.ToString() ?? "null";
```
jmprieur: ####             return Argument?.ToString() ?? "null";

---

Isn't it going to mess-up existing logs? #Resolved

jmprieur (commented):
Thanks @ciaozhang 
I've asked questions.

Not sure why this is so complicated?
why not just:

if (PII)
{
 write token;
}
else
{
 write token.SubString(0, token.LastIndexOf('.'));
}

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
ciaozhang: There are 5 segments for JWE and 3 segments for a JWS.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -3,7 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
```
victorm-hernandez: >IO [](http://example.com/codeflow?start=13&length=2)

#### using System.IO;

---

Do you really need these 2 namespaces? From the code it seems there is no need. Could you confirm? #Closed

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -358,9 +360,42 @@ public static string FormatInvariant(string format, params object[] args)
...
+
+        /// <summary>
+        /// Regex that is used to figure out if a token is in JWE format.
+        /// </summary>
+        public static readonly Regex RegexJwe = new Regex(JweCompactSerializationRegex, RegexOptions.Compiled | RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(100));
+
+        /// <summary>
+        /// JWS - Token format: 'header.payload.signature'. Signature is optional, but '.' is required.
+        /// </summary>
+        public const string JsonCompactSerializationRegex = @"^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*$";
```
victorm-hernandez: ####         public const string JsonCompactSerializationRegex = @"^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*$";

---

All of this code is never referenced and is outside of concern of this namespace, lets remove the regex code. #Closed

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -358,9 +360,42 @@ public static string FormatInvariant(string format, params object[] args)
...
 
-            return string.Format(CultureInfo.InvariantCulture, format, args);
+        /// <summary>
+        /// JWE - Token format: 'protectedheader.encryptedkey.iv.cyphertext.authenticationtag'.
+        /// </summary>
+        public const string JweCompactSerializationRegex = @"^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+$";
+
+        private static string SanitizeSecurityArtifact(object arg)
+        {
+            if (arg is ISafeLogSecurityArtifact)
```
victorm-hernandez: >ISafeLogSecurityArtifact [](http://example.com/codeflow?start=23&length=24)

####             if (arg is ISafeLogSecurityArtifact)

---

If you are planning to use the ISafeLogSecurityArtifact interface, there is no need for the SecurityArtifact type and the related helper methods to wrap objects with this type. 
 #Resolved

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1238,13 +1240,13 @@ private static TokenValidationResult ReadToken(string token, TokenValidationPara
             JsonWebToken jsonWebToken = null;
             if (validationParameters.TokenReader != null)
             {
-                var securityToken = validationParameters.TokenReader(token, validationParameters);
+                var securityToken = validationParameters.TokenReader(token, validationParameters); 
                 if (securityToken == null)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10510, token)));
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10510, LogHelper.MarkAsSecurityArtifact(token))));
```
victorm-hernandez: >MarkAsSecurityArtifact [](http://example.com/codeflow?start=162&length=22)

####                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10510, LogHelper.MarkAsSecurityArtifact(token))));

---

This method wraps the argument in a SecurityArtifact object, yet this type is not used anywhere on the LogHelper. Since the current implementation relies on the ISafeLogSecurityArtifact interface there is no need for this helper method and the SecurityArtifact type. 

 #Resolved

victorm-hernandez (dismissed):
:clock1:

src/Microsoft.IdentityModel.Logging/ISafeLogSecurityArtifact.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// Interface that defines methods to log security artifact.
+    /// </summary>
+    public interface ISafeLogSecurityArtifact
+    {
+        /// <summary>
+        /// Returns a string that represents the disarmed security artifact.
```
henrik-me: >disarmed [](http://example.com/codeflow?start=49&length=8)

####         /// Returns a string that represents the disarmed security artifact.

---

what is a disarmed security artifact? and what is an armed security artifact? ... and also we are explaining the method names with words that are even harder to understand. Can we pls. simplify the message and also, can we simplify the names of these 2 methods - it's not logical to me what these 2 things are doing? #Resolved

src/Microsoft.IdentityModel.Logging/LogHelper.cs
henrik-me: method could have the argument be of type SecurityArtifact instead. right now you are typecasting 3 places. and every path has 2 type casts

src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
```diff
@@ -469,6 +469,18 @@ public override string ToString()
                 return Header.SerializeToJson() + ".";
         }
 
+        /// <inheritdoc/>
+        public override string SecureToString()
```
henrik-me: >SecureToString [](http://example.com/codeflow?start=31&length=14)

####         public override string SecureToString()

---

Why not simply call these what they do? 
NoSignatureToString() and WithSignatureToString() #Resolved

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -358,9 +360,42 @@ public static string FormatInvariant(string format, params object[] args)
...
-                return string.Format(CultureInfo.InvariantCulture, format, args.Select(RemovePII).ToArray()); 
+                return string.Format(CultureInfo.InvariantCulture, format, args.Select(RemovePII).ToArray());
+            else
+                return string.Format(CultureInfo.InvariantCulture, format, args.Select(SanitizeSecurityArtifact).ToArray());
+        }
+
+        /// <summary>
+        /// Regex that is used to figure out if a token is in JWS format.
+        /// </summary>
+        public static readonly Regex RegexJws = new Regex(JsonCompactSerializationRegex, RegexOptions.Compiled | RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(100));
```
brentschmaltz: ####         public static readonly Regex RegexJws = new Regex(JsonCompactSerializationRegex, RegexOptions.Compiled | RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(100));

---

These should not be public. #Resolved

src/Microsoft.IdentityModel.Logging/NonPII.cs
brentschmaltz: +1 no need to make this change from Null => null.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
brentschmaltz: For that matter perhaps a property: LogHelper.LogTokensWithSignature.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
ciaozhang: Removed

src/Microsoft.IdentityModel.Logging/NonPII.cs
ciaozhang: Changed to "Null"

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
ciaozhang: We only implement the interface for the object derived from SecurityToken. Here we log the token as string that the LogHelper cannot scrub the signature for a string.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
brentschmaltz: Several places in our call graphs log tokens as strings, since they are not ISafeLogSecurityArtifact, ToString() will be called and the entire token with signature will be logged.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -431,6 +431,12 @@ private static long ParseTimeValue(JToken jToken, string claimName)
             throw LogHelper.LogExceptionMessage(new FormatException(LogHelper.FormatInvariant(LogMessages.IDX14300, LogHelper.MarkAsNonPII(claimName), jToken.ToString(), LogHelper.MarkAsNonPII(typeof(long)))));
         }
 
+        internal static string SafeLogJwtToken(object obj)
+        {
+            string token = obj as string;
```
brentschmaltz: ####             string token = obj as string;

---

Suggested logic here

```cs
if (obj == null)
    return string.Empty;

// not a string
if (token == null)
    return obj.ToString()

int lastDot = token.LastIndexOf(".");

// no dots, maybe not a JWT
if (lastDot == -1)
    return typeof(string).ToString();

return token.Substring(0,lastDot);

``` #Resolved

File: src/Microsoft.IdentityModel.Logging/ISafeLogSecurityArtifact.cs [hunk 33]
```diff
@@ -0,0 +1,18 @@
...
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// Interface that provides an unsafe method to log a security artifact.
+    /// </summary>
+    public interface ISafeLogSecurityArtifact
+    {
+        /// <summary>
+        /// Returns a string that represents the complete security artifact.
+        /// This may include sensitive information and should only be used for logging purposes.
```
victorm-hernandez: >logging [](http://example.com/codeflow?start=79&length=7)

We should not encourage the use of this for logging.

This should be used for exceptional debugging cases as it will reveal security tokens in plain text. 

 #Resolved

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -891,6 +894,15 @@ public string Subject
             }
         }
 
+        /// <summary>
+        /// Returns the encoded token without signature or authentication tag.
+        /// </summary>
+        /// <returns>Encoded token string  without signature or authentication tag.</returns>
```
victorm-hernandez: >   [](http://example.com/codeflow?start=41&length=2)

####         /// <returns>Encoded token string  without signature or authentication tag.</returns>

---

nit: extra space #Closed

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// An internal structure that is used to mark an argument as SecurityArtifact.
+    /// Arguments wrapped with a SecurityArtifact structure will be considered as SecurityArtifact in the message logging process.
```
victorm-hernandez: >Arguments [](http://example.com/codeflow?start=8&length=9)

####     /// Arguments wrapped with a SecurityArtifact structure will be considered as SecurityArtifact in the message logging process.

---

This may be the right place to clarify that this is meant for the scenario where the argument is not a class but a simple string. If the argument is a string there is no way to add an interface that can be used by the LogHelper.

Instead we wrap the string with this class and pass a delegate that knows how to sanitize or not the string. #Closed

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <returns><c>Null</c> if the <see cref="Argument"/> is <see langword="null"/>, otherwise calls <see cref="System.ValueType.ToString()"/> method of the <see cref="Argument"/>.</returns>
+        public override string ToString()
+        {
+            if (_callback != null && Argument != null)
+                return _callback(Argument);
+
+            return Argument?.ToString() ?? "null";
+        }
+
+        public string UnsafeToString()
```
victorm-hernandez, line 52: >UnsafeToString [](http://example.com/codeflow?start=22&length=14)

XML docs missing. #Resolved

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -511,13 +511,13 @@ protected virtual void ValidateCHash(OpenIdConnectProtocolValidationContext vali
             object cHashClaim;
             if (!validationContext.ValidatedIdToken.Payload.TryGetValue(JwtRegisteredClaimNames.CHash, out cHashClaim))
             {
-                throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidCHashException(LogHelper.FormatInvariant(LogMessages.IDX21307, validationContext.ValidatedIdToken)));
+                throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidCHashException(LogHelper.FormatInvariant(LogMessages.IDX21307, validationContext.ValidatedIdToken.ToString())));
```
victorm-hernandez: >ToString [](http://example.com/codeflow?start=182&length=8)

Why do you convert to a string here? Isn't this the opposite of what we want? And if you make it a string, should not you use the MarkAsSecurityArtifact?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
```diff
@@ -469,6 +469,9 @@ public override string ToString()
                 return Header.SerializeToJson() + ".";
         }
 
+        /// <inheritdoc/>
+        public override string UnsafeToString() => RawData;
```
victorm-hernandez, line 474: >UnsafeToString [](http://example.com/codeflow?start=31&length=14)

Lets update the Saml2SecurityToken as well to have parity on this feature. #Resolved

victorm-hernandez (changes requested):
:clock1:

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
ciaozhang: Done

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
ciaozhang: Fixed

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
ciaozhang: Added

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
ciaozhang: Removed ToString()

src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
brentschmaltz: @victorm-hernandez are you with tackling SAML after the JWT's.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
brentschmaltz: +1

src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
brentschmaltz: @victorm-hernandez can we open a new item for Saml as removing the signature from Saml tokens is a bit more complicated that JWT's and will need some thought.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -429,6 +429,24 @@ private static long ParseTimeValue(JToken jToken, string claimName)
             throw LogHelper.LogExceptionMessage(new FormatException(LogHelper.FormatInvariant(LogMessages.IDX14300, LogHelper.MarkAsNonPII(claimName), jToken.ToString(), LogHelper.MarkAsNonPII(typeof(long)))));
         }
 
+        internal static string SafeLogJwtToken(object obj)
```
victorm-hernandez, line 434: >object [](http://example.com/codeflow?start=47&length=6)

I thought this was meant for strings only, for any other object type we should be able to use the provided interface #Closed

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -429,6 +429,24 @@ private static long ParseTimeValue(JToken jToken, string claimName)
...
+        {
+            if (obj == null)
+                return string.Empty;
+
+            // not a string
+            if (!(obj is string token))
+                return obj.ToString();
```
victorm-hernandez: This doesn't seem safe at all. By doing this we are giving the impression of safety while we are not really doing any changes. #Closed

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -429,6 +429,24 @@ private static long ParseTimeValue(JToken jToken, string claimName)
...
+
+            int lastDot = token.LastIndexOf(".");
+
+            // no dots, maybe not a JWT
+            if (lastDot == -1)
```
victorm-hernandez, line 446: >lastDot [](http://example.com/codeflow?start=16&length=7)

Since this is meant to parse JWT token strings we should throw an argument exception here.  #Closed

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
```diff
@@ -35,6 +35,11 @@ private IdentityModelEventSource()
         /// </summary>
         public static bool ShowPII { get; set; } = false;
 
+        /// <summary>
+        /// Flag which indicates whether or not armed security artifact is shown in logs when <see cref="ShowPII"/> is set to true. False by default.
```
victorm-hernandez: >true [](http://example.com/codeflow?start=126&length=4)

It is not clear for someone without context what a security artifact is. Let's provide a concrete example, security tokens  #Closed

src/Microsoft.IdentityModel.Logging/ISafeLogSecurityArtifact.cs
victorm-hernandez: @ciaozhang

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
victorm-hernandez: @ciaozhang

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
victorm-hernandez: @ciaozhang

src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
victorm-hernandez: Agreed @brentschmaltz , lets do this in another PR. @ciaozhang could you please create the work item and leave it here.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
ciaozhang: It is the callback func we used in SecurityArtifact struct. The reason we take obj as the type of parameter is LogHelper treat all arguments as object.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
ciaozhang: We will throw in the ReadToken() method if the string token is not an encoded JWT token.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
ciaozhang: WI created:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2090

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
ciaozhang: Updated the XML.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
ciaozhang: We only mark string as Security artifact.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
victorm-hernandez: I see, still regardless of the parameter type this is meant to be consumed by using string version of tokens correct? 

how do you justify the code for non strings that you have here?

src/Microsoft.IdentityModel.Logging/ISafeLogSecurityArtifact.cs
victorm-hernandez: Still waiting for this change @ciaozhang

src/System.IdentityModel.Tokens.Jwt/JwtSecurityToken.cs
victorm-hernandez: Created ADO work item as well
Product Backlog Item 2595096: Replicate the logic to "disarm security artifact" from JWT tokens to SAML tokens

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
victorm-hernandez: What I mean here is that if I call your method I should be able to assume that your output is now safe. In the case of non strings you cannot guarantee that, the only safe choice is to return null here or string.empty.

victorm-hernandez: Since this is really an exceptional case and should never happen you may consider throwing.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
victorm-hernandez: This should be safe regardless of the caller, for non strings, non tokens we should return null.

src/Microsoft.IdentityModel.Logging/ISafeLogSecurityArtifact.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// Interface that provides an unsafe method to log a security artifact.
```
victorm-hernandez, line 7: >Interface [](http://example.com/codeflow?start=8&length=9)

You may want to bring the description you have for the SecurityArtifact class to the interface, this would address the questions that Henrik had here.  #Resolved

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -393,6 +406,17 @@ public static object MarkAsNonPII(object arg)
             return new NonPII(arg);
         }
 
+        /// <summary>
+        /// Marks a log message argument (<paramref name="arg"/>) as SecurityArtifact.
+        /// </summary>
+        /// <param name="arg">A log message argument to be marked as SecurityArtifact.</param>
+        /// <param name="callback">A callback function to log the security artifact safely.</param>
+        /// <returns>An argument marked as SecurityArtifact.</returns>
+        public static object MarkAsSecurityArtifact(object arg, Func<object, string> callback)
```
victorm-hernandez, line 415: Why do we receive an object here, this should receive and return a string.  #Resolved

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
ciaozhang: It will use the default ToString() method return the GetType().ToString()

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
ciaozhang: return "UncognizedEncodedString"

victorm-hernandez (approved):
:shipit:

TimHannMSFT (approved):
Approving based on Victor's approval and seeing that the new direction is more consistent. Thanks for the changes @ciaozhang!

PR #2066, opened 2023-04-27
Fix AotCompatibilityTests

eerhardt:
- Use Path.Combine so it works on non-Windows
- Increase the timeout to 3 minutes, since slower machines are seeing this test take around 1 minute
- Use the plain TreatWarningsAsErrors property


eerhardt:
FYI @brentschmaltz

brentschmaltz (approved):

PR #2072, opened 2023-05-05
Enforce key sizes when creating HMAC

brentschmaltz:
This is a fix to ensure when creating a JWS (signed JsonWebToken), the key size is of sufficient size.
The key size must be > the size Hash algorithm returns.
HS256 requires 256 bits, HS384 requires 384 bits, HS512 requires 512 bits.

Allowance is made for authenticated encryption which uses a smaller key for the AuthenticationTag.

If the SymmetricKey is smaller than the required size IdentityModel will throw an ArgumentOutOfRangeException which is the same exception is the SymmetricKey is smaller than the minimum key size.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
+++ b/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
@@ -445,88 +445,58 @@ namespace Microsoft.IdentityModel.Tokens
             {
                 case SecurityAlgorithms.Aes128CbcHmacSha256:
                 {
-                        if (keyBytes.Length < 16)
-                            throw LogHelper.LogExceptionMessage(
-                                new InvalidOperationException(
-                                    LogHelper.FormatInvariant(LogMessages.IDX10720,
-                                    LogHelper.MarkAsNonPII(algorithm),
-                                    LogHelper.MarkAsNonPII("128"),
-                                    LogHelper.MarkAsNonPII(keyBytes.Length * 8))));
-
-                        return new HMACSHA256(keyBytes);
+                    ValidateKeySize(keyBytes, algorithm, 16);
+                    return new HMACSHA256(keyBytes);
                 }
 
                 case SecurityAlgorithms.Aes192CbcHmacSha384:
                 {
-                    if (keyBytes.Length < 24)
-                        throw LogHelper.LogExceptionMessage(
-                            new InvalidOperationException(
-                                LogHelper.FormatInvariant(LogMessages.IDX10720,
-                                LogHelper.MarkAsNonPII(algorithm),
-                                LogHelper.MarkAsNonPII("192"),
-                                LogHelper.MarkAsNonPII(keyBytes.Length * 8))));
-
+                    ValidateKeySize(keyBytes, algorithm, 24);
                     return new HMACSHA384(keyBytes);
                 }
 
                 case SecurityAlgorithms.Aes256CbcHmacSha512:
                 {
-                    if (keyBytes.Length < 32)
-                        throw LogHelper.LogExceptionMessage(
-                            new InvalidOperationException(
-                                LogHelper.FormatInvariant(LogMessages.IDX10720,
-                                LogHelper.MarkAsNonPII(algorithm),
-                                LogHelper.MarkAsNonPII("256"),
-                                LogHelper.MarkAsNonPII(keyBytes.Length * 8))));
-
+                    ValidateKeySize(keyBytes, algorithm, 32);
                     return new HMACSHA512(keyBytes);
                 }
 
                 case SecurityAlgorithms.HmacSha256Signature:
                 case SecurityAlgorithms.HmacSha256:
                 {
-                        if (keyBytes.Length < 32)
-                        throw LogHelper.LogExceptionMessage(
-                            new InvalidOperationException(
-                                LogHelper.FormatInvariant(LogMessages.IDX10720,
-                                LogHelper.MarkAsNonPII(algorithm),
-                                LogHelper.MarkAsNonPII("256"),
-                                LogHelper.MarkAsNonPII(keyBytes.Length * 8))));
-
+                    ValidateKeySize(keyBytes, algorithm, 32);
                     return new HMACSHA256(keyBytes);
                 }
 
                 case SecurityAlgorithms.HmacSha384Signature:
                 case SecurityAlgorithms.HmacSha384:
                 {
-                    if (keyBytes.Length < 48)
-                        throw LogHelper.LogExceptionMessage(
-                            new InvalidOperationException(
-                                LogHelper.FormatInvariant(LogMessages.IDX10720,
-                                LogHelper.MarkAsNonPII(algorithm),
-                                LogHelper.MarkAsNonPII("384"),
-                                LogHelper.MarkAsNonPII(keyBytes.Length * 8))));
-
+                    ValidateKeySize(keyBytes, algorithm, 48);
                     return new HMACSHA384(keyBytes);
                 }
 
                 case SecurityAlgorithms.HmacSha512Signature:
                 case SecurityAlgorithms.HmacSha512:
                 {
-                    if (keyBytes.Length < 64)
-                        throw LogHelper.LogExceptionMessage(
-                            new InvalidOperationException(
-                                LogHelper.FormatInvariant(LogMessages.IDX10720,
-                                LogHelper.MarkAsNonPII(algorithm),
-                                LogHelper.MarkAsNonPII("512"),
-                                LogHelper.MarkAsNonPII(keyBytes.Length * 8))));
-
+                    ValidateKeySize(keyBytes, algorithm, 64);
                     return new HMACSHA512(keyBytes);
                 }
 
                 default:
                     throw LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant(LogMessages.IDX10666, LogHelper.MarkAsNonPII(algorithm))));
             }
+        }
+
+        private static void ValidateKeySize(byte[] keyBytes, string algorithm, int expectedNumberOfBytes)
+        {
+            if (keyBytes.Length < expectedNumberOfBytes)
+                throw LogHelper.LogExceptionMessage(
+                    new ArgumentOutOfRangeException(
+                        nameof(keyBytes),
+                        LogHelper.FormatInvariant(LogMessages.IDX10720,
+                            LogHelper.MarkAsNonPII(algorithm),
+                            LogHelper.MarkAsNonPII(expectedNumberOfBytes * 8),
+                            LogHelper.MarkAsNonPII(keyBytes.Length * 8))));
         }
 
         private SignatureProvider CreateSignatureProvider(SecurityKey key, string algorithm, bool willCreateSignatures, bool cacheProvider)
--- a/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
+++ b/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
@@ -428,24 +428,101 @@ namespace Microsoft.IdentityModel.Tokens
             if (CustomCryptoProvider != null && CustomCryptoProvider.IsSupportedAlgorithm(algorithm, keyBytes))
             {
                 if (!(CustomCryptoProvider.Create(algorithm, keyBytes) is KeyedHashAlgorithm keyedHashAlgorithm))
-                    throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10647, LogHelper.MarkAsNonPII(algorithm), LogHelper.MarkAsNonPII(typeof(KeyedHashAlgorithm)))));
+                    throw LogHelper.LogExceptionMessage(
+                        new InvalidOperationException(
+                            LogHelper.FormatInvariant(
+                                LogMessages.IDX10647,
+                                LogHelper.MarkAsNonPII(algorithm),
+                                LogHelper.MarkAsNonPII(typeof(KeyedHashAlgorithm)))));
 
                 return keyedHashAlgorithm;
             }
 
+            // In the case of Aes128CbcHmacSha256, Aes192CbcHmacSha384, Aes256CbcHmacSha512 which are Authenticated Encryption algorithms
+            // SymmetricSignatureProvider will get passed a key with 1/2 the minimum keysize expected size for the HashAlgorithm. 16 bytes for SHA256, instead of 32 bytes.
+            // see: https://datatracker.ietf.org/doc/html/rfc7518#section-5.2.2.1
             switch (algorithm)
             {
+                case SecurityAlgorithms.Aes128CbcHmacSha256:
+                {
+                        if (keyBytes.Length < 16)
+                            throw LogHelper.LogExceptionMessage(
+                                new InvalidOperationException(
+                                    LogHelper.FormatInvariant(LogMessages.IDX10720,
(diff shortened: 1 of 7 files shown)


src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -428,24 +428,101 @@ public virtual KeyedHashAlgorithm CreateKeyedHashAlgorithm(byte[] keyBytes, stri
...
+                            new InvalidOperationException(
+                                LogHelper.FormatInvariant(LogMessages.IDX10720,
+                                LogHelper.MarkAsNonPII(algorithm),
+                                LogHelper.MarkAsNonPII("192"),
+                                LogHelper.MarkAsNonPII(keyBytes.Length * 8))));
+
+                    return new HMACSHA384(keyBytes);
+                }
+
+                case SecurityAlgorithms.Aes256CbcHmacSha512:
```
dannybtsai, line 458: Does it make sense to move this to the SecurityAlgorithms.HmacSha256Signature case?

TimHannMSFT:
@brentschmaltz can you include a section in the description for what we should include in the release notes for this? It's technically a behavior change so I want to call it out and not just list a PR. Fine to just be a sentence or two.

TimHannMSFT:
Sign Microsoft.IdentityModel.Abstractions seems to have had a signing issue on the CI build. I've kicked off a new one.
image


---
In reply to: [1536789416](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2072#issuecomment-1536789416) [](http://example.com/codeflow?ancestors=1536789416)

TimHannMSFT:
>         }

I believe this is now unused right? Should be removed.

---
Refers to: src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs:396 in 8d4445f. [](commit_id = 8d4445fb103f4428e67ebed7ad31bc4bf5aeed22, deletion_comment = False)

TimHannMSFT:
>             throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10652, LogHelper.MarkAsNonPII(algorithm)), nameof(algorithm)));

So we no longer throw this exception now correct? Probably at least need to call this out in case someone was depending on that in the past?

---
Refers to: src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs:395 in 8d4445f. [](commit_id = 8d4445fb103f4428e67ebed7ad31bc4bf5aeed22, deletion_comment = False)

TimHannMSFT:
>             }

If the algorithm isn't supported here we just keep going? That seems somewhat odd. Why wouldn't we try to throw with `IDX10652` like we did in the `GetHmacAlgorithm` call?

---
Refers to: src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs:548 in 8d4445f. [](commit_id = 8d4445fb103f4428e67ebed7ad31bc4bf5aeed22, deletion_comment = False)

TimHannMSFT:
>             throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10652, LogHelper.MarkAsNonPII(algorithm)), nameof(algorithm)));

Eyeballing it, seems like now we'll throw an exception from `AuthenticationEncryptionProvider(SecurityKey, string)` constructor.

Seems like it will be `IDX10668` instead which I think is fine.

We don't seem to have any negative unit tests for bad algorithm here.

---
In reply to: [1536847843](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2072#issuecomment-1536847843) [](http://example.com/codeflow?ancestors=1536847843)

---
Refers to: src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs:395 in 8d4445f. [](commit_id = 8d4445fb103f4428e67ebed7ad31bc4bf5aeed22, deletion_comment = False)

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -428,24 +428,101 @@ public virtual KeyedHashAlgorithm CreateKeyedHashAlgorithm(byte[] keyBytes, stri
...
+                                LogHelper.FormatInvariant(LogMessages.IDX10720,
+                                LogHelper.MarkAsNonPII(algorithm),
+                                LogHelper.MarkAsNonPII("512"),
+                                LogHelper.MarkAsNonPII(keyBytes.Length * 8))));
+
                     return new HMACSHA512(keyBytes);
+                }
 
                 default:
                     throw LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant(LogMessages.IDX10666, LogHelper.MarkAsNonPII(algorithm))));
```
TimHannMSFT, line 486: >new NotSupportedException(LogHelper.FormatInvariant(LogMessages.IDX10666, LogHelper.MarkAsNonPII(algorithm)))); [](http://example.com/codeflow?start=56&length=111)

Are there cases where we used to throw an `ArgumentException` for the algorithm but now the call flow will reach here and we'll throw a `NotSupportedException`?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -428,24 +428,101 @@ public virtual KeyedHashAlgorithm CreateKeyedHashAlgorithm(byte[] keyBytes, stri
...
             {
+                case SecurityAlgorithms.Aes128CbcHmacSha256:
+                {
+                        if (keyBytes.Length < 16)
+                            throw LogHelper.LogExceptionMessage(
+                                new InvalidOperationException(
+                                    LogHelper.FormatInvariant(LogMessages.IDX10720,
+                                    LogHelper.MarkAsNonPII(algorithm),
+                                    LogHelper.MarkAsNonPII("128"),
+                                    LogHelper.MarkAsNonPII(keyBytes.Length * 8))));
```
TimHannMSFT: nit: Create something like:

```
private void ThrowIfKeysizeTooSmall(string alg, int keyBytesLength, int minKeySize)
{
...
}
```

to avoid code duplication and prevent typos if this gets extended. #Resolved

dannybtsai: ThrowIfKeySizeTooSmall(keyBytes.Length, 64, algorithm);

        private static void ThrowIfKeySizeTooSmall(int keyLengthInBytes, int minSizeInBytes, string algorithm)
        {
            if (keyLengthInBytes < minSizeInBytes)
            {
                    new InvalidOperationException(
                        LogHelper.FormatInvariant(LogMessages.IDX10720,
                        LogHelper.MarkAsNonPII(algorithm),
                        LogHelper.MarkAsNonPII(minSizeInBytes * 8),
                        LogHelper.MarkAsNonPII(keyLengthInBytes * 8))));
            }
        }

test/Microsoft.IdentityModel.Tokens.Tests/SignatureProviderTests.cs
```diff
@@ -621,14 +621,14 @@ public static TheoryData<SignatureProviderTheoryData> SymmetricVerifySingatureSi
                         ExpectedException = EE.ArgumentException("IDX10719:"),
                         RawBytes= new byte[32],
                         Signature = new byte[32],
-                        SigningSignatureProvider = new SymmetricSignatureProvider(KEY.SymmetricSecurityKey2_256, ALG.HmacSha384),
```
TimHannMSFT, line 624: >SigningSignatureProvider = new SymmetricSignatureProvider(KEY.SymmetricSecurityKey2_256, ALG.HmacSha384), [](http://example.com/codeflow?start=24&length=105)

So this call used to throw `ArgumentException` but now throws `InvalidOperationException`. Let's make sure to call this out in release notes.

I know Karel would say we should bump the major version for this. We want to at least make it explicit when we release though since we know some partners will consider this breaking.

TimHannMSFT:
>                         ExpectedException = EE.ArgumentException("IDX10719:"),

The way we validate this theory data seems broken. `SymmetricVerify1Tests` for example passes whether or not `ExpectedException` is present or not. If I comment the `ExpectedException` line on these three cases the tests still pass, that seems like we're likely not validating in the way we plan to, no?

`SymmetricVerify2Tests` and `SymmetricVerify3Tests` do require this as the exception *IS* thrown in those cases. I haven't pulled the current version to see what the behavior is there.

---
Refers to: test/Microsoft.IdentityModel.Tokens.Tests/SignatureProviderTests.cs:614 in 8d4445f. [](commit_id = 8d4445fb103f4428e67ebed7ad31bc4bf5aeed22, deletion_comment = False)

TimHannMSFT:
Seems like retrying worked...

---
In reply to: [1536789416](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2072#issuecomment-1536789416) [](http://example.com/codeflow?ancestors=1536789416,1536789416)

TimHannMSFT (changes requested):
IMO it's too confusing to have unit tests that specify an "ExpectedException" but they can pass if it's not thrown. I think everything else I see as optional to fix (thought I think it's worthwhile) and I think we need to at least note in release notes the changes in exception behavior.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: I thought about that but wanted to keep them separate.

test/Microsoft.IdentityModel.Tokens.Tests/SignatureProviderTests.cs
brentschmaltz: @TimHannMSFT it will still throw argument exception when the algorithm is not known.
A new exception will be thrown InvalidOperationException when the key size is not correct.
I will review this to see if we can throw an existing exception.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: @TimHannMSFT not in this call graph.
I will review to see if we can use an existing exception.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: I made a helper function.

test/Microsoft.IdentityModel.Tokens.Tests/SignatureProviderTests.cs
```diff
@@ -603,7 +603,7 @@ public void SymmetricVerify3Tests(SignatureProviderTheoryData theoryData)
             TestUtilities.AssertFailIfErrors(context);
         }
 
-        public static TheoryData<SignatureProviderTheoryData> SymmetricVerifySingatureSizeTheoryData
+        public static TheoryData<SignatureProviderTheoryData> SymmetricVerifySignatureSizeTheoryData
```
TimHannMSFT, line 606: >Signature [](http://example.com/codeflow?start=77&length=9)

Good catch #Resolved

test/Microsoft.IdentityModel.Tokens.Tests/SignatureProviderTests.cs
TimHannMSFT: Not sure I follow. When I run these tests they're not throwing any exceptions and the test doesn't seem to care (at least `SymmetricVerify1Tests` e.g. that the Theory data contains an `ExpectedException`.

That test has Verify return `false` and since no exception is thrown it's never checked. This is what's happening in dev too though so I guess it's no worse than what's checked it. This seems like confusing test behavior to me and...SEEMS to allow for a case where someone sets an expected exception in the theory data and tests silently succeed which seems problematic. In any case, given that part isn't a behavior change I don't want to hold this up.

TimHannMSFT (approved):
I think the main thing I was blocking on was something I assumed changed from dev, that `TheoryData` containing `ExpectedException` wasn't checked. That just seems to be how these tests work for some of the usages so won't block this. I think it makes sense to address that in a future change though once this security issue is addressed. Thanks for these changes @brentschmaltz . Still some suggestions outstanding but nothing I see as blocking. 

I would still like to see a section in the description which can be leveraged for the release notes.

ciaozhang (approved):

brentschmaltz:
@TimHannMSFT i added additional comments for release notes.
We should address the test cases that you feel are not working.
Can you please open issues?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -428,30 +428,77 @@ public virtual KeyedHashAlgorithm CreateKeyedHashAlgorithm(byte[] keyBytes, stri
...
+        private static void ValidateKeySize(byte[] keyBytes, string algorithm, int expectedNumberOfBytes)
+        {
+            if (keyBytes.Length < expectedNumberOfBytes)
+                throw LogHelper.LogExceptionMessage(
+                    new ArgumentOutOfRangeException(
+                        nameof(keyBytes),
+                        LogHelper.FormatInvariant(LogMessages.IDX10720,
+                            LogHelper.MarkAsNonPII(algorithm),
+                            LogHelper.MarkAsNonPII(expectedNumberOfBytes * 8),
+                            LogHelper.MarkAsNonPII(keyBytes.Length * 8))));
```
henrik-me, line 499: nit: why waste cpu time multiplying by 8? How about clarifying the message instead? Also is 8 always the right number? (message could use length as the terminology rather than bits/bytes)

niemyjski:
This is a breaking change, semver should have been bumped.

brentschmaltz:
@niemyjski we thought of this as a bug.

niemyjski:
Bug or not, it brought down our dev environment and authentication failed.

brentschmaltz:
@niemyjski where you able to work around the issue?
Do you need a fix?

niemyjski:
I downgraded for now. We will need to inspect all environments and update deployments and do it when it least impacts customers. There were no documented requirements for key sizes when we first generated them. It would be awesome if this was documented someplace...

AaronAtDuo:
This hit a customer of ours recently, the sudden enforcement of key size broke an application.

FWIW, the HMAC spec accounts for keys that are too short, by appending null bytes onto the end
```
(1) append zeros to the end of K to create a B byte string
        (e.g., if K is of length 20 bytes and B=64, then K will be
         appended with 44 zero bytes 0x00)
```
so I'm not sure the library really needs to prevent this, unless it wants to be very prescriptive about key strength?

Liebeck:
Uff. Updating to 6.30.1 caused a lot of issues for us :-|

`System.ArgumentOutOfRangeException: IDX10720: Unable to create KeyedHashAlgorithm for algorithm 'http://www.w3.org/2001/04/xmldsig-more#hmac-sha256', the key size must be greater than: '256' bits, key has '160' bits. (Parameter 'keyBytes')`

The changelogs should have been more clear, maybe with a "If you are doing X, this will beak" and some recommendations

If anyone else is having the same issue:

Pin the NuGet version to 6.30.0 ` <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="[6.30.0]" />`

VitaliyMF:
For everyone who needs to keep backward compatibility with 'short' keys (and old JWTs generated with these short keys): https://github.com/dotnet/aspnetcore/issues/52369#issuecomment-1872286954

PR #2076, opened 2023-05-10
Introduce ConfigurationValidationException

ciaozhang:
ConfigurationManager won't cache configuration if config is invalid and ConfigurationValidationException


sruke:
@ciaozhang, consider creating a GitHub issue for this PR with a summary of the issue and why this change is needed. This will help with updating the release notes with relevant information.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -173,11 +173,10 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
                         {
                             ConfigurationValidationResult result = _configValidator.Validate(configuration);
                             if (!result.Succeeded)
-                                LogHelper.LogWarning(LogMessages.IDX20810, result.ErrorMessage);
+                                throw LogHelper.LogExceptionMessage(new ConfigurationInvalidException(LogHelper.FormatInvariant(LogMessages.IDX20810, result.ErrorMessage)));
```
sruke: My understanding is result.ErrorMessage can be safely marked as NonPII to surface the reason for why the configuration validation failed.  #ByDesign

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -506,7 +509,8 @@ public static TheoryData<ConfigurationManagerTheoryData<OpenIdConnectConfigurati
                     ConfigurationRetreiver = new OpenIdConnectConfigurationRetriever(),
                     ConfigurationValidator = openIdConnectConfigurationValidator2,
                     DocumentRetriever = new FileDocumentRetriever(),
-                    ExpectedErrorMessage = "IDX21818: ",
+                    //ExpectedErrorMessage = "IDX21818: ",
```
sruke: Consider deleting this line.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
brentschmaltz: I don't think we should throw here unless _currentConfiguration is null.

brentschmaltz: Only set: _lastRefresh _syncAfter _currentConfiguration if configurationValidator was successful (if one was set).

dannybtsai (commented):
Don't use hard-coded "null", use Microsoft.IdentityModel.Tokens.Utility.Null instead.

dannybtsai (commented):
Why renaming this class? The original ConfigurationValidationException is thrown when a configuration fails to validate, which seems right to me.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationInvalidException.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.IdentityModel.Protocols.Configuration
+{
+    /// <summary>
+    /// This exception is thrown when retrieved configuration is not valid.
+    /// </summary>
+    [Serializable]
+    public class ConfigurationInvalidException : Exception
```
victorm-hernandez: >ConfigurationInvalidException [](http://example.com/codeflow?start=17&length=29)

nit: InvalidConfigurationException ? #Resolved

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationInvalidException.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// Initializes a new instance of the <see cref="ConfigurationInvalidException"/> class.
+        /// </summary>
+        public ConfigurationInvalidException()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ConfigurationInvalidException"/> class.
+        /// </summary>
+        /// <param name="message">Addtional information to be included in the exception and displayed to user.</param>
```
victorm-hernandez: >Addtional [](http://example.com/codeflow?start=34&length=9)

nit: typo in both constructors #Resolved

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationInvalidException.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <param name="innerException">A <see cref="Exception"/> that represents the root cause of the exception.</param>
+        public ConfigurationInvalidException(string message, Exception innerException)
+            : base(message, innerException)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ConfigurationInvalidException"/> class.
+        /// </summary>
+        /// <param name="info">the <see cref="SerializationInfo"/> that holds the serialized object data.</param>
```
victorm-hernandez: >t [](http://example.com/codeflow?start=31&length=1)

nit: upper case T #Resolved

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
victorm-hernandez: The exception thrown will not bubble up outside of this method since the code is within a try catch block.

The catch block below has the behavior you described Brent, it will throw if the _currentConfiguration == null.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -462,22 +463,24 @@ public void ValidateOpenIdConnectConfigurationTests(ConfigurationManagerTheoryDa
...
 
             try
             {
-                //create a listener and enable it for logs
-                var listener = TestUtils.SampleListener.CreateLoggerListener(EventLevel.Warning);
-
-                var configurationManager = new ConfigurationManager<OpenIdConnectConfiguration>(theoryData.MetadataAddress, theoryData.ConfigurationRetreiver, theoryData.DocumentRetriever, theoryData.ConfigurationValidator);
-                var configuration = configurationManager.GetConfigurationAsync().Result;
-
-                if (!string.IsNullOrEmpty(theoryData.ExpectedErrorMessage) && !listener.TraceBuffer.Contains(theoryData.ExpectedErrorMessage))
```
victorm-hernandez, line 474: >IsNullOrEmpty [](http://example.com/codeflow?start=28&length=13)

The only indication that this happen if the config is not initially null is from the logs. Lets keep this check around.

Let's test both scenarios, one where the initial configuration is set and only logs show and one where no configuration is set and an exception is thrown with the proper details.

victorm-hernandez (changes requested):
:clock1:

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationInvalidException.cs
ciaozhang: rename the exception

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationInvalidException.cs
ciaozhang: Fixed

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
ciaozhang: Updated tests

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -462,22 +465,34 @@ public void ValidateOpenIdConnectConfigurationTests(ConfigurationManagerTheoryDa
...
                 if (!string.IsNullOrEmpty(theoryData.ExpectedErrorMessage) && !listener.TraceBuffer.Contains(theoryData.ExpectedErrorMessage))
                     context.AddDiff($"Expected exception to contain: '{theoryData.ExpectedErrorMessage}'.{Environment.NewLine}Log is:{Environment.NewLine}'{listener.TraceBuffer}'");
 
+                theoryData.ExpectedException.ProcessNoException(context);
             }
-            catch (Exception ex)
+            catch (AggregateException ex)
             {
-                theoryData.ExpectedException.ProcessException(ex, context);
+                // this should throw, because last configuration retrieved was null
```
victorm-hernandez, line 488: >// this should throw, because last configuration retrieved was null [](http://example.com/codeflow?start=16&length=67)

This seems was not cleaned up, you are setting the _currentConfiguration on line 472, so this should never throw.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -462,22 +465,34 @@ public void ValidateOpenIdConnectConfigurationTests(ConfigurationManagerTheoryDa
...
+                Assert.Throws<AggregateException>(() => configuration = configurationManager.GetConfigurationAsync().Result);
```
victorm-hernandez, line 489: >Assert [](http://example.com/codeflow?start=16&length=6)

Let's add a check here, if the theoryData.PresetCurrentConfiguration is set to true, throw since this should never be the case.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -516,7 +531,18 @@ public static TheoryData<ConfigurationManagerTheoryData<OpenIdConnectConfigurati
                     ConfigurationRetreiver = new OpenIdConnectConfigurationRetriever(),
```
victorm-hernandez, line 531: >ConfigurationRetreiver [](http://example.com/codeflow?start=20&length=22)

nit: unrelated to the change, but there is a typo here

victorm-hernandez (approved):
:shipit:

brentschmaltz (changes requested):
We are throwing a new exception that does not derive from a previously thrown exception, we need to derive InvalidConfigurationException  from the exception that was previously thrown.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -31,10 +31,10 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
...
         private TimeSpan _bootstrapRefreshInterval = TimeSpan.FromSeconds(1);
 
-    /// <summary>
-    /// Static initializer for a new object. Static initializers run before the first instance of the type is created.
-    /// </summary>
-    static ConfigurationManager()
+        /// <summary>
+        /// Static initializer for a new object. Static initializers run before the first instance of the type is created.
+        /// </summary>
+        static ConfigurationManager()
```
brentschmaltz: I know this was here before, i cannot see a reason to for this.
Can we remove it? #Resolved

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -166,18 +166,17 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
                         {
                             ConfigurationValidationResult result = _configValidator.Validate(configuration);
                             if (!result.Succeeded)
-                                LogHelper.LogWarning(LogMessages.IDX20810, result.ErrorMessage);
+                                throw LogHelper.LogExceptionMessage(new InvalidConfigurationException(LogHelper.FormatInvariant(LogMessages.IDX20810, result.ErrorMessage)));
                         }
 
+                        _lastRefresh = DateTimeOffset.UtcNow;
+                        // Add 1 hour jitter to avoid spike traffic to IdentityProvider.
+                        _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval + TimeSpan.FromMinutes(new Random().Next(60)));
```
brentschmaltz: The jitter value should be based on the AutomaticRefreshInterval. #Resolved

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
ciaozhang: removed

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
TimHannMSFT: This comment thread seems to have forked from Sruthi's initial observation which I beleive is still valid. It looks like we control this error message and already scrub it. It should be marked NonPII so it's acutally usable for investigations.

```
var convertKeyInfos = string.Join("\n", openIdConnectConfiguration.JsonWebKeySet.Keys.Where(key => !string.IsNullOrEmpty(key.ConvertKeyInfo)).Select(key => key.Kid.ToString() + ": " + key.ConvertKeyInfo));
                return new ConfigurationValidationResult
                {
                    ErrorMessage = LogHelper.FormatInvariant(
                        LogMessages.IDX21818,
                        LogHelper.MarkAsNonPII(MinimumNumberOfKeys),
                        LogHelper.MarkAsNonPII(numberOfValidKeys),
                        string.IsNullOrEmpty(convertKeyInfos) ? "None" : convertKeyInfos),
                    Succeeded = false
                };
```
Looks like we're doing the wrong thing with how we create this error message sometimes too, I've created a separate PBI for that and assigned to Franco (https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Middleware/Backlog%20items/?workitem=2596631)

TimHannMSFT: @brentschmaltz @victorm-hernandez it looks like your side channel on this is resolved right? The exception handling is how Brent described?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -166,18 +159,18 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
                             if (!result.Succeeded)
-                                LogHelper.LogWarning(LogMessages.IDX20810, result.ErrorMessage);
+                                throw LogHelper.LogExceptionMessage(new InvalidConfigurationException(LogHelper.FormatInvariant(LogMessages.IDX20810, result.ErrorMessage)));
                         }
 
+                        _lastRefresh = DateTimeOffset.UtcNow;
+                        // Add up to 1 hour jitter to avoid spike traffic to IdentityProvider.
+                        _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval +
+                            TimeSpan.FromSeconds(new Random().Next((int)Math.Min(AutomaticRefreshInterval.TotalSeconds / 10, TimeSpan.FromSeconds(600).TotalSeconds))));
                         _currentConfiguration = configuration;
```
TimHannMSFT: It seems to me we should use the same jitter strategy we used for `_bootstrapRefreshIntervalWithJitter` in a recent PR. Another possibility is we should update that jitter strategy to match this. In any case we shouldn't have jitter behave one way in one section and another way in another section.

Here we seem to be saying Interval + 0-Interval/10 and in the other case we're saying 0-Interval.

Further, we shoudln't use "magic numbers". Since we're picking (random?) numbers here let's at least comment on the reasoning to make revisiting easier--this may be moot if we follow the older pattern.

I'm not necessarily against having a floor (i.e. the min range vs constant value) but in that case the constant value should be something declared up top not just inline here.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -166,18 +159,18 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
                         var configuration = await _configRetriever.GetConfigurationAsync(MetadataAddress, _docRetriever, CancellationToken.None).ConfigureAwait(false);
-                        _lastRefresh = DateTimeOffset.UtcNow;
-                        // Add 1 hour jitter to avoid spike traffic to IdentityProvider.
-                        _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval + TimeSpan.FromMinutes(new Random().Next(60)));
                         if (_configValidator != null)
                         {
                             ConfigurationValidationResult result = _configValidator.Validate(configuration);
                             if (!result.Succeeded)
-                                LogHelper.LogWarning(LogMessages.IDX20810, result.ErrorMessage);
+                                throw LogHelper.LogExceptionMessage(new InvalidConfigurationException(LogHelper.FormatInvariant(LogMessages.IDX20810, result.ErrorMessage)));
```
TimHannMSFT, line 166: This should also be marked as NonPII IMO. Otherwise the logs will usually not be very useful.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
TimHannMSFT: As in the prior comment, we should unify the way we do jitter across our libraries. I'm okay if you open another PBI to address this but this is different than the other jitter method you added in this same PR. Consider leveraging a helper function so folks adding jitter don't have to think about it and it's easy to have a common way we do jitter everywhere.w

Also, as in the other comment. Why 60 minutes? This should either be directly based on the value you're jittering on or should be a constant declared up top.

TimHannMSFT (approved):
Please either unify jitter or cut a PBI to unify jitter before checking in. We shouldn't proliferate different ways of doing this in our code without a plan to address.

Approving with the expectation you'll open an issue before merging. (same goes for pending error message marking as PII but that's probably easier to just fix).

ciaozhang:
PBI created: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2596832

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
brentschmaltz: Let's add a random amount in the interval [0, 5%] of AutomaticRefreshInterval.

ciaozhang: Updated to 0~5%

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
ciaozhang: Will be fixed in:
https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Middleware/Backlog%20items/?workitem=2596631

brentschmaltz (approved):

PR #2082, opened 2023-05-17
Westinm/claims mapping

### LINKED ISSUE #2044: [Feature Request] Add support for Mapping Claims to JsonWebTokenHandler
Trying to upgrade ASP.NET's `AddAuthentication().AddJwtBearer()` implementation from `JwtSecurityTokenHandler` to `JsonWebTokenHandler`, it appears a missing feature is the `MapInboundClaims` feature that `JwtSecurityTokenHandler` has.

This means if a token with payload like:

```json
{
  "sub": "1234567890",
  "unique_name": "MyTestUser",
  "iat": 1516239022
}
```

is issued using the new `JsonWebTokenHandler`, the `unique_name` claim isn't set on the resulting `ClaimsIdentity` as illustrated in the following repro program:

```C#

const string token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwidW5pcXVlX25hbWUiOiJNeVRlc3RVc2VyIiwiaWF0IjoxNTE2MjM5MDIyfQ.RdBPDWUy92hIn9MjxdzZjkn5yeFObc9-cq5xMaDFuag";

string symmetricKeyEncoded_256 = "Vbxq2mlbGJw8XH+ZoYBnUHmHga8/o/IduvU/Tht70iE=";
byte[] symmetricKeyBytes_256 = Convert.FromBase64String(symmetricKeyEncoded_256);
SymmetricSecurityKey symmetricSecurityKey_256 = new SymmetricSecurityKey(symmetricKeyBytes_256) { KeyId = "symmetricKeyBytes_256" };

var validationParameters = new TokenValidationParameters
{
    ValidateIssuerSigningKey = true,
    RequireSignedTokens = false,
    IssuerSigningKey = symmetricSecurityKey_256,
    ValidateIssuer = false,
    ValidateAudience = false,
    ValidateLifetime = false,
    ValidAlgorithms = new List<string> { SecurityAlgorithms.HmacSha256 }
};

JwtSecurityTokenHandler oldHandler = new JwtSecurityTokenHandler();
TokenValidationResult result = await oldHandler.ValidateTokenAsync(token, validationParameters);

Console.WriteLine(result.ClaimsIdentity.Name); // prints 'MyTestUser'

JsonWebTokenHandler newHandler = new JsonWebTokenHandler();
result = await newHandler.ValidateTokenAsync(token, validationParameters);

Console.WriteLine(result.ClaimsIdentity.Name ?? "<null>"); // should print 'MyTestUser' but instead prints '<null>'
```

`JsonWebTokenHandler` should be able to map the `unique_name` claim to the `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name` claim, so the `ClaimsIdentity` works properly, like it does with `JwtSecurityTokenHandler`.

ASP.NET could implement this functionality itself, but that seems inefficient, and it wouldn't help anyone else moving to the `JsonWebTokenHandler` API.

**Additional context**
N/A

cc @brentschmaltz

westin-m:
Map claims in JsonWebTokenHandler. Move the ClaimTypeMapping class into TokenUtilities.


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -39,6 +52,26 @@ public Type TokenType
             get { return typeof(JsonWebToken); }
         }
 
+        /// <summary>
+        /// Gets or sets the property name of <see cref="Claim.Properties"/> the will contain the original JSON claim 'name' if a mapping occurred when the <see cref="Claim"/>(s) were created.
+        /// </summary>
+        /// <exception cref="ArgumentException">If <see cref="string"/>.IsNullOrWhiteSpace('value') is true.</exception>
+        public static string ShortClaimTypeProperty
```
westin-m, line 72: There are now some duplicate members between JsonWebTokenHandler and JwtSecurityTokenHandler. Is this a concern/ should I work to move these things up to TokenHandler?

westin-m:
I wasn't sure what, if anything, from JwtSecurityTokenHandler/JsonWebTokenHendler should be moved up to TokenHandler; let me know your thoughts.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: @westin-m 

TokenHandler is the base class for different types of tokens (SAML, SAML2, JWT).
ShortClaimTypeProperty is specific to JWT, so we should use a common place where JsonWebTokenHandler and JwtSecurityTokenHandler can access.

S.IM.Tokens.JwtSecurityTokenHandler takes a dependency on M.IM.JsonWebTokenHandler, so the common code can be in  M.IM.JsonWebTokens.

westin-m: @brentschmaltz 
Done, see the latest push. Removing ShortClaimTypeProperty from JwtSecurityTokenHandler messes with [this test](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/2b1b78bf38a5739719e95b7045d81902a2cf09d5/test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs#LL857C29-L857C29); should I change it also reference JsonWebToken?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -57,6 +59,26 @@ public class JsonWebToken : SecurityToken
         private string _x5t;
         private string _zip;
 
+        /// <summary>
+        /// Gets or sets the property name of <see cref="Claim.Properties"/> the will contain the original JSON claim 'name' if a mapping occurred when the <see cref="Claim"/>(s) were created.
```
jennyf19: >the [](http://example.com/codeflow?start=77&length=3)

this or that?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -22,6 +23,139 @@
 
 namespace Microsoft.IdentityModel.JsonWebTokens
 {
+
+    internal static class ClaimTypeMapping
+    {
+        // This is the short to long mapping.
+        // key      is the long  claim type
```
jennyf19: >       [](http://example.com/codeflow?start=14&length=6)

remove extra spaces here and in line 31. Remove line 26.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -57,6 +59,26 @@ public class JsonWebToken : SecurityToken
...
+        /// </summary>
+        /// <exception cref="ArgumentException">If <see cref="string"/>.IsNullOrWhiteSpace('value') is true.</exception>
+        public static string ShortClaimTypeProperty
```
brentschmaltz: We should have one of these properties in JwtSecurityTokenHandler and JsonWebTokenHandler having a static shared between the two mean that changing JsonWebToken.ShortClaimTypeProperty will affect JwtSecurityToken, that is not desirable.

Leave the property on JwtSecurityTokenHandler and add another to JsonWebTokenHandler.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -23,6 +23,16 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// </summary>
     public class JsonWebTokenHandler : TokenHandler
     {
+        /// <summary>
+        /// Default claim type mapping for inbound claims.
+        /// </summary>
+        public readonly static IDictionary<string, string> DefaultInboundClaimTypeMap = ClaimTypeMapping.InboundClaimTypeMap;
```
brentschmaltz: This sets DefaultInboundClaimTypeMap to a reference of ClaimTypeMapping.InboundClaimTypeMap.
Should change the code here and in JwtSecurityTokenHandler. With only JwtSecurityTokenHandler there was no problem.
```cs
public readonly static IDictionary<string, string> DefaultInboundClaimTypeMap = new Dictionary<string,string>(ClaimTypeMapping.InboundClaimTypeMap);
```

The following code will result in JsonWebTokenHandler.DefaultInboundClaimTypeMap with size 0.
```cs
var jwtClaimsMapping = JwtSecurityTokenHandler.DefaultInboundClaimTypeMap;
jwtClaimsMapping.Clear();
var jsonClaimsMapping = JsonWebTokenHandler.DefaultInboundClaimTypeMap;
```

brentschmaltz (changes requested):
Be careful about shared statics.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: Make sure there are tests to ensure these are independent to guard against regressions.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
westin-m: Each handler will have their own version of this property

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
westin-m: that

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
westin-m: Both Handlers use `new` to construct their Dictionaries, and a test checks for independence.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -23,6 +23,19 @@ namespace Microsoft.IdentityModel.JsonWebTokens
...
+
+        /// <summary>
+        /// Default claim type mapping for inbound claims.
+        /// </summary>
+        public readonly static IDictionary<string, string> DefaultInboundClaimTypeMap = new Dictionary<string, string>(ClaimTypeMapping.InboundClaimTypeMap);
+
+        /// <summary>
+        /// Default value for the flag that determines whether or not the InboundClaimTypeMap is used.
+        /// </summary>
+        public readonly static bool DefaultMapInboundClaims = false;
```
keegan-caruso: Why is this readonly?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -23,6 +23,19 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// </summary>
     public class JsonWebTokenHandler : TokenHandler
     {
+        private const string _namespace = "http://schemas.xmlsoap.org/ws/2005/05/identity/claimproperties";
+        private static string _shortClaimType = _namespace + "/ShortTypeName";
+
+        /// <summary>
+        /// Default claim type mapping for inbound claims.
+        /// </summary>
+        public readonly static IDictionary<string, string> DefaultInboundClaimTypeMap = new Dictionary<string, string>(ClaimTypeMapping.InboundClaimTypeMap);
```
keegan-caruso: Why is this readonly?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -39,7 +39,7 @@ public class JwtSecurityTokenHandler : SecurityTokenHandler
         /// <summary>
         /// Default claim type mapping for inbound claims.
         /// </summary>
-        public static IDictionary<string, string> DefaultInboundClaimTypeMap = ClaimTypeMapping.InboundClaimTypeMap;
+        public readonly static IDictionary<string, string> DefaultInboundClaimTypeMap = new Dictionary<string, string>(ClaimTypeMapping.InboundClaimTypeMap);
```
keegan-caruso: Why are we changing this to readonly?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JwtTokenUtilitiesTests.cs
```diff
@@ -13,6 +14,26 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
 {
     public class JwtTokenUtilitiesTests
     {
+        [Fact]
+        public void ClaimTypeMappingIsIndependent()
+        {
+            // Each handler should have its own instance of th ClaimTypeMap
```
keegan-caruso: nit typo: "th"

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -680,9 +713,62 @@ protected virtual ClaimsIdentity CreateClaimsIdentity(JsonWebToken jwtToken, Tok
             if (string.IsNullOrWhiteSpace(issuer))
                 issuer = GetActualIssuer(jwtToken);
 
+            if (DefaultMapInboundClaims)
```
keegan-caruso: this branch looks like it is missing new test coverage?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -22,6 +23,139 @@
 
 namespace Microsoft.IdentityModel.JsonWebTokens
 {
+
+    internal static class ClaimTypeMapping
```
keegan-caruso: can we not have two classes in this file?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -14,6 +14,7 @@
 using Microsoft.IdentityModel.Logging;
 using Microsoft.IdentityModel.Tokens;
 using TokenLogMessages = Microsoft.IdentityModel.Tokens.LogMessages;
+using System.Security.Claims;
```
keegan-caruso: nit: this should be directly before system.security.cryptography

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
westin-m: good point, moved to its own

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
westin-m: thanks for catching that. The theory data wasn't working how I thought, so there's a new test.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
westin-m: removed readonly, unecessary change

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
westin-m: removed readonly and suppressed the error

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
westin-m: removed readonly and suppressed the error

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -23,13 +23,54 @@ namespace Microsoft.IdentityModel.JsonWebTokens
...
         /// Gets the Base64Url encoded string representation of the following JWT header: 
         /// { <see cref="JwtHeaderParameterNames.Alg"/>, <see cref="SecurityAlgorithms.None"/> }.
         /// </summary>
         /// <return>The Base64Url encoded string representation of the unsigned JWT header.</return>
         public const string Base64UrlEncodedUnsignedJWSHeader = "eyJhbGciOiJub25lIn0";
 
+        /// <summary>
+        /// Static initializer for a new object. Static initializers run before the first instance of the type is created.
+        /// </summary>
+        static JsonWebTokenHandler()
```
brentschmaltz: Why do we need this static ctor?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JwtTokenUtilitiesTests.cs
```diff
@@ -13,6 +14,26 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
...
+        public void ClaimTypeMappingIsIndependent()
+        {
+            // Each handler should have its own instance of the ClaimTypeMap
+            var jwtClaimsMapping = JwtSecurityTokenHandler.DefaultInboundClaimTypeMap;
+            var jsonClaimsMapping = JsonWebTokenHandler.DefaultInboundClaimTypeMap;
+
+            Assert.NotEmpty(jwtClaimsMapping);
+            Assert.NotEmpty(jsonClaimsMapping);
+
+            Assert.Equal(jwtClaimsMapping, jsonClaimsMapping);
```
brentschmaltz, line 27: What does Assert.Equal do here?
Our model to check if two dictionaries are the same it to use IdentityComparer.[AreStringDictionariesEqual](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/2b1b78bf38a5739719e95b7045d81902a2cf09d5/test/Microsoft.IdentityModel.TestUtils/IdentityComparer.cs#L736).

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2575,6 +2575,37 @@ public void ValidateJWE(JwtTheoryData theoryData)
             TestUtilities.AssertFailIfErrors(context);
         }
 
+        [Fact]
+        public async Task ValidateJsonWebTokenClaimMapping()
```
brentschmaltz, line 2582: We need a test that creates a JWT with every mapped claim and then checks that the result of validation from the JwtSecurityTokenHandler and JsonWebTokenHandler are the same, both in the mapped and unmapped case.
We also need a test that shows if the JwtSecurityTokenHandler has mapping OFF and the JsonWebTokenHandler has mapping ON, the test shows the claims are different.

Have a look at these sets of [claims](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/2b1b78bf38a5739719e95b7045d81902a2cf09d5/test/Microsoft.IdentityModel.TestUtils/Default.cs#L504).

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2575,6 +2575,37 @@ public void ValidateJWE(JwtTheoryData theoryData)
...
+        {
+            const string token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwidW5pcXVlX25hbWUiOiJNeVRlc3RVc2VyIiwiaWF0IjoxNTE2MjM5MDIyfQ.RdBPDWUy92hIn9MjxdzZjkn5yeFObc9-cq5xMaDFuag";
+
+            string symmetricKeyEncoded_256 = "Vbxq2mlbGJw8XH+ZoYBnUHmHga8/o/IduvU/Tht70iE=";
```
brentschmaltz: Instead of creating a key, use KeyingMaterial.
There are a number of keys available for example: [DefaultSymmetricSecurityKey_256 ](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/2b1b78bf38a5739719e95b7045d81902a2cf09d5/test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs#L232)

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2575,6 +2575,37 @@ public void ValidateJWE(JwtTheoryData theoryData)
             TestUtilities.AssertFailIfErrors(context);
         }
 
+        [Fact]
+        public async Task ValidateJsonWebTokenClaimMapping()
+        {
+            const string token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwidW5pcXVlX25hbWUiOiJNeVRlc3RVc2VyIiwiaWF0IjoxNTE2MjM5MDIyfQ.RdBPDWUy92hIn9MjxdzZjkn5yeFObc9-cq5xMaDFuag";
```
brentschmaltz: It is best not to have hard coded strings as validation will fail when the token is expired, rather create a new token using code similar shown [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/2b1b78bf38a5739719e95b7045d81902a2cf09d5/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs#L110).

Also think about using these [sets ](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/2b1b78bf38a5739719e95b7045d81902a2cf09d5/test/Microsoft.IdentityModel.TestUtils/Default.cs#L504)of claims to validate the mapping works as expected.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -23,13 +23,54 @@ namespace Microsoft.IdentityModel.JsonWebTokens
...
+
+        /// <summary>
+        /// Default claim type mapping for outbound claims.
+        /// </summary>
+        public static IDictionary<string, string> DefaultOutboundClaimTypeMap = new Dictionary<string, string>(ClaimTypeMapping.OutboundClaimTypeMap);
+
+        /// <summary>
+        /// Default value for the flag that determines whether or not the InboundClaimTypeMap is used.
+        /// </summary>
+        public static bool DefaultMapInboundClaims = false;
```
eerhardt, line 39: We typically don't declare mutable static fields. This should be a property instead.

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/field

> his interpretation immediately implies that all fields must be private.

We exclude constant and static read-only fields from this strict restriction, because such fields, almost by definition, are never required to change.

eerhardt: Same for the above 2 static fields.

westin-m: Thanks for pointing this out, Eric. I will leave the default values, but include properties to get/set the private members.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2575,6 +2576,90 @@ public void ValidateJWE(JwtTheoryData theoryData)
             TestUtilities.AssertFailIfErrors(context);
         }
 
+        // Test creates a JWT with every mapped claim and then checks that the result of validation from the
+        // JwtSecurityTokenHandler and JsonWebTokenHandler are the same, both in the mapped and unmapped case.
+        [Fact]
+        public async Task ValidateJsonWebTokenClaimMapping()
+        {
+            var jwtTokenHandler = new JsonWebTokenHandler() { MapInboundClaims = false };
```
brentschmaltz: change name: jwtTokenHandler -> jsonWebTokenHandler

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2575,6 +2576,90 @@ public void ValidateJWE(JwtTheoryData theoryData)
...
+                ValidAudience = Default.Audience,
+                ValidIssuer = Default.Issuer,
+                ValidateIssuerSigningKey = true,
+                IssuerSigningKey = KeyingMaterial.DefaultSymmetricSigningCreds_256_Sha2.Key,
+                TokenDecryptionKey = KeyingMaterial.DefaultX509Key_2048,
+                AlgorithmValidator = ValidationDelegates.AlgorithmValidatorBuilder(true),
+                RequireExpirationTime = false,
+            };
+
+            JwtSecurityTokenHandler securityTokenHandler = new JwtSecurityTokenHandler() { MapInboundClaims = false };
```
brentschmaltz: change name: securityTokenHandler -> jwtSecurityTokenHandler

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2575,6 +2576,90 @@ public void ValidateJWE(JwtTheoryData theoryData)
...
+
+            TokenValidationResult jwtValidationResult = await jwtTokenHandler.ValidateTokenAsync(accessToken, validationParameters);
```
brentschmaltz: name change: jwtValidationResult -> jsonValidationResult
securityValidationResult -> jwtValidationResult

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2575,6 +2576,90 @@ public void ValidateJWE(JwtTheoryData theoryData)
...
+
+            var context = new CompareContext
+            {
+                PropertiesToIgnoreWhenComparing = new Dictionary<Type, List<string>>
+                {
+                    { typeof(TokenValidationResult),  new List<string> { "SecurityToken", "TokenType" } }
+                }
+            };
+
+            Assert.True(IdentityComparer.AreEqual(jwtValidationResult, securityValidationResult, context));
```
brentschmaltz: Instead of calling Assert.True, use [TokenUtilities.AssertFailIfErrors ](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/02518f7f8b6829be3dbb2786a0d1a644c1549aa1/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs#L48).
Assert will throw after the first issue, when we use AssertFailIfErrors, then the test will complete with all issues.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2575,6 +2576,90 @@ public void ValidateJWE(JwtTheoryData theoryData)
...
+        }
+
+        // Test shows if the JwtSecurityTokenHandler has mapping OFF and 
+        // the JsonWebTokenHandler has mapping ON,the claims are different.
+        [Fact]
+        public async Task ValidateDifferentClaimsBetweenHandlers()
+        {
+            var jwtTokenHandler = new JsonWebTokenHandler() { MapInboundClaims = true };
```
brentschmaltz: same name changes as above.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2575,6 +2576,90 @@ public void ValidateJWE(JwtTheoryData theoryData)
...
+                AlgorithmValidator = ValidationDelegates.AlgorithmValidatorBuilder(true),
+                RequireExpirationTime = false,
+            };
+
+            JwtSecurityTokenHandler securityTokenHandler = new JwtSecurityTokenHandler() { MapInboundClaims = false };
+
+            TokenValidationResult jwtValidationResult = await jwtTokenHandler.ValidateTokenAsync(accessToken, validationParameters);
+            TokenValidationResult securityValidationResult = await securityTokenHandler.ValidateTokenAsync(accessToken, validationParameters);
+
+            Assert.False(IdentityComparer.AreEqual(jwtValidationResult.Claims, securityValidationResult.Claims, CompareContext.Default));
```
brentschmaltz: Same as above, use AssertFailIfErrors

brentschmaltz (changes requested):
Please squash to a single commit.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -23,13 +23,39 @@ namespace Microsoft.IdentityModel.JsonWebTokens
...
     {
+        private IDictionary<string, string> _inboundClaimTypeMap;
+        private const string _namespace = "http://schemas.xmlsoap.org/ws/2005/05/identity/claimproperties";
+        private static string _shortClaimType = _namespace + "/ShortTypeName";
+        private bool _mapInboundClaims = DefaultMapInboundClaims;
+
+        /// <summary>
+        /// Default claim type mapping for inbound claims.
+        /// </summary>
+        public static IDictionary<string, string> DefaultInboundClaimTypeMap = new Dictionary<string, string>(ClaimTypeMapping.InboundClaimTypeMap);
```
dannybtsai, line 34: This can be changed by anyone but not thread-safe, is that the intention? Is there a test to confirm the behavior?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -23,13 +23,39 @@ namespace Microsoft.IdentityModel.JsonWebTokens
...
         /// </summary>
         /// <return>The Base64Url encoded string representation of the unsigned JWT header.</return>
         public const string Base64UrlEncodedUnsignedJWSHeader = "eyJhbGciOiJub25lIn0";
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="JsonWebTokenHandler"/> class.
+        /// </summary>
+        public JsonWebTokenHandler()
+        {
+            if (_mapInboundClaims)
```
dannybtsai, line 53: DefaultMapInboundClaims is public and not thread-safe.

ThreadA sets DefaultMapInboundClaims  = true
ThreadB sets DefaultMapInboundClaims = false
ThreadA calls the default constructor, but the result is not expected
ThreadB  calls the default constructor with the expected result.

The behavior needs to be clearly documented.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: Danny and i discussed this offline.
The idea is that in asp.net startup, this gives a uses the ability to turn off mapping in one place.
It is not meant to be used more than once.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
brentschmaltz: See above, this is OK.

brentschmaltz (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -23,13 +23,39 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// </summary>
     public class JsonWebTokenHandler : TokenHandler
     {
+        private IDictionary<string, string> _inboundClaimTypeMap;
+        private const string _namespace = "http://schemas.xmlsoap.org/ws/2005/05/identity/claimproperties";
+        private static string _shortClaimType = _namespace + "/ShortTypeName";
```
eerhardt, line 28: `private const _shortClaimType = _namespace + "/ShortTypeName";

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
eerhardt: I think you want to make these properties as well. Exposing public static fields isn't recommended.

Issue #2081, opened 2023-05-17
[Feature Request] JsonWebTokenHandler: Add support for creation of A256KW / A256GCM EncryptionCredentials 


anblanco:
**Is your feature request related to a problem? Please describe.**
I'm working on code that parses a JWT provided by Google Play Integrity. 

The [token format](https://developer.android.com/google/play/integrity/verdict#token-format) is 
* JWE uses A256KW for alg and A256GCM for enc
* JWS uses ES256

I am able to successfully validate these tokens when using JsonWebTokenHandler.

However, if I try to create a token of this format the an exception is thrown 
```
SecurityKey encryptionKey = new SymmetricSecurityKey(symmetricKeyBytes);
SecurityKey signingKey = new ECDsaSecurityKey(ecdsa);

var jsonWebTokenHandler = new JsonWebTokenHandler();

// A256GCM Encryption
var encryptingCredentials = new EncryptingCredentials(
  key: encryptionKey,
  alg: SecurityAlgorithms.Aes256KW,
  enc: SecurityAlgorithms.Aes256Gcm);

// ES256 Signing
var signingCredentials = new SigningCredentials(
  key: signingKey,
  algorithm: SecurityAlgorithms.EcdsaSha256);

string jwe = jsonWebTokenHandler.CreateToken(payload, signingCredentials, encryptingCredentials);

Exception:
Microsoft.IdentityModel.Tokens.SecurityTokenEncryptionFailedException: IDX10617: Encryption failed. Keywrap is only supported for: 'A128CBC-HS256', 'A192CBC-HS384' and 'A256CBC-HS512'. The content encryption specified is: 'A256GCM'.

Stack Trace: 
JwtTokenUtilities.GetSecurityKey(EncryptingCredentials encryptingCredentials, CryptoProviderFactory cryptoProviderFactory, IDictionary`2 additionalHeaderClaims, Byte[]& wrappedKey)
JsonWebTokenHandler.EncryptTokenPrivate(String innerJwt, EncryptingCredentials encryptingCredentials, String compressionAlgorithm, IDictionary`2 additionalHeaderClaims, String tokenType)
JsonWebTokenHandler.CreateTokenPrivate(String payload, SigningCredentials signingCredentials, EncryptingCredentials encryptingCredentials, String compressionAlgorithm, IDictionary`2 additionalHeaderClaims, IDictionary`2 additionalInnerHeaderClaims, String tokenType)
[...]

PR #1606, opened 2021-03-03
Support decryption using AesGcm

RojaEnnam:
Includes AesGcm decryption implementation using Bcrypt library and reference test from the [spec](https://tools.ietf.org/html/rfc7516#appendix-A.1).

PR #2083, opened 2023-05-18
Add Windows support for AESGCM EncryptingCredentials

anblanco:
This commit adds Windows support for specifying AESGCM algorithms for `EncryptingCredentials`. 

The motivation is described in #2081

Notably #1606 first added support for AESGCM Decryption, but throws a `NotSupportedException` when attempting to create a new `EncryptingCredentials`. It also adds a testing class that supports Encryption / Decryption, but I don't have the full context on why encryption was excluded from production code given that decryption works as expected. Alternatively I wouldn't mind overloading `CryptoProvider` / `AuthenticatedEncryptionProvider` with the Windows-specific Interop for my use case (.NET Framework), but these types are made internal to the library. 

Uses the existing Windows Interop for OS Support of the encryption algorithm.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -300,6 +300,27 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             return key;
         }
 
+        /// <summary>
+        /// Generates key bytes.
+        /// </summary>
+        public static byte[] GenerateAesGcmKeyBytes(int sizeInBits)
+        {
+            byte[] key = null;
+            if (sizeInBits != 128 && sizeInBits != 192 && sizeInBits != 256)
+                throw LogHelper.LogExceptionMessage(new ArgumentException(TokenLogMessages.IDX10402, nameof(sizeInBits)));
+
+            using (var aes = Aes.Create())
+            { 
+                int sizeInBytes = sizeInBits >> 3;
+                key = new byte[sizeInBytes];
+                aes.KeySize = sizeInBits;
+                aes.GenerateKey();
+                Array.Copy(aes.Key, key, sizeInBytes);
+            }
+
+            return key;
+        }
+
         internal static SecurityKey GetSecurityKey(
             EncryptingCredentials encryptingCredentials,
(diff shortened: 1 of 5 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -362,9 +383,17 @@ public static byte[] GenerateKeyBytes(int sizeInBits)
...
+                // only 128, 192 and 256 AesGcm for CEK algorithm
+                else if(SecurityAlgorithms.Aes128Gcm.Equals(encryptingCredentials.Enc))
+                    securityKey = new SymmetricSecurityKey(GenerateAesGcmKeyBytes(128));
+                else if (SecurityAlgorithms.Aes192Gcm.Equals(encryptingCredentials.Enc))
+                    securityKey = new SymmetricSecurityKey(GenerateAesGcmKeyBytes(192));
+                else if (SecurityAlgorithms.Aes256Gcm.Equals(encryptingCredentials.Enc))
+                    securityKey = new SymmetricSecurityKey(GenerateAesGcmKeyBytes(256));
                 else
                     throw LogHelper.LogExceptionMessage(
-                        new SecurityTokenEncryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10617, LogHelper.MarkAsNonPII(SecurityAlgorithms.Aes128CbcHmacSha256), LogHelper.MarkAsNonPII(SecurityAlgorithms.Aes192CbcHmacSha384), LogHelper.MarkAsNonPII(SecurityAlgorithms.Aes256CbcHmacSha512), LogHelper.MarkAsNonPII(encryptingCredentials.Enc))));
```
anblanco, line 367: Even if I provide a custom `CryptoProviderFactory` with my own implementation of `AuthenticatedEncryptionProvider`, this code still throws and prevents me from creating tokens of type AESGCM

One alternate approach would be to have two separate PRs:

1. Enable support for AES-GCM with custom CryptoProviderFactory. This would allow consumers of the API to plug-in their own AuthenticatedEncryptionProvider (e.g. Windows Interop / .NET AES GCM) 

2. A separate PR that adds support for AES-GCM Encryption. Later .NET APIs have a cross-platform API to call, but folks on .NET Framework would need to use Windows Interop

PR #2087, opened 2023-05-24
Remove unnecessary NuGet package dependencies.

### LINKED ISSUE #2086: Avoid referencing System.Text.Encoding
See https://github.com/dotnet/runtime/issues/86671 for the full context. 

TL;DR: `Microsoft.IdentityModel.JsonWebTokens` 6.25.1 transitively depends on a vulnerable package `System.Private.Uri` 4.3.0. This can be avoided if `Microsoft.IdentityModel.JsonWebTokens` no longer depends on `System.Text.Encoding` 4.3.0.

teo-tsirpanis:
Fixes #2086. The removed packages are for legacy target frameworks and are not needed anymore; their functionality is built-in.

No code was changed, if CI passes it should be OK.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -6,19 +6,8 @@
     <MicrosoftCSharpVersion>4.5.0</MicrosoftCSharpVersion>
     <MicrosoftSourceLinkGitHubVersion>1.0.0</MicrosoftSourceLinkGitHubVersion>
     <NetStandardVersion>2.0.3</NetStandardVersion>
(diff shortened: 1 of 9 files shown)


jennyf19:
@teo-tsirpanis Thank you for the contribution, but I think we have this fixed now in the dev7x branch. Do you want to confirm and/or do a PR there if we missed something? thanks again.

teo-tsirpanis:
Looks good, thanks!

PR #2097, opened 2023-06-01
remove net45 target

westin-m:
Remove the net45, 452 targets from wilson


ViktorHofer:
Note that net461 is already out-of-support as well: https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/

PR #2100, opened 2023-06-02
(feature): Add ActiveTokenEndpoint to WsFederationConfiguration

victorm-hernandez:
Added support to the WsFederationMetadataSerializer to read the SecurityTokenSerivceEndpoint element from WS-Federation metadata and exposed it via the WsFederationConfiguration object as ActiveTokenEndpoint.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
@@ -318,6 +318,12 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         public override string TokenEndpoint { get; set; }
 
         /// <summary>
+        /// This base class property is not used in OpenIdConnect. 
+        /// </summary>
+        [JsonIgnore]
+        public override string ActiveTokenEndpoint { get; set; }
+
+        /// <summary>
         /// Gets the collection of 'token_endpoint_auth_methods_supported'.
         /// </summary>
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore, NullValueHandling = NullValueHandling.Ignore, PropertyName = OpenIdProviderMetadataNames.TokenEndpointAuthMethodsSupported, Required = Required.Default)]
--- a/src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
@@ -1,0 +1,155 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using Microsoft.IdentityModel.Tokens;
+using Microsoft.IdentityModel.Xml;
+using static Microsoft.IdentityModel.Logging.LogHelper;
+
+namespace Microsoft.IdentityModel.Protocols.WsFederation
+{
+    /// <summary>
+    /// Defines a class for validating the WsFederationConfiguration.
+    /// </summary>
+    public class WsFederationConfigurationValidator : IConfigurationValidator<WsFederationConfiguration>
+    {
+        /// <summary>
+        /// Validates a WsFederationConfiguration.
+        /// </summary>
+        /// <param name="configuration">WsFederationConfiguration to validate</param>
+        /// <returns>A <see cref="ConfigurationValidationResult"/> that contains validation result.</returns>
+        /// <exception cref="ArgumentNullException">If the provided configuration is null</exception>
+        public ConfigurationValidationResult Validate(WsFederationConfiguration configuration)
+        {
+            if (configuration == null)
+                throw LogArgumentNullException(nameof(configuration));
+
+            if (string.IsNullOrWhiteSpace(configuration.Issuer))
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22700,
+                    Succeeded = false
+                };
+            }
+
+            if (configuration.Signature == null)
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22701,
+                    Succeeded = false
+                };
+            }
+
+            if (configuration.Signature.KeyInfo == null)
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22702,
+                    Succeeded = false
+                };
+            }
+
+            if (string.IsNullOrWhiteSpace(configuration.Signature.SignatureValue))
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22703,
+                    Succeeded = false
+                };
+            }
+
+            if (string.IsNullOrWhiteSpace(configuration.Signature.SignedInfo.SignatureMethod))
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22704,
+                    Succeeded = false
+                };
+            }
+
+            if (configuration.Signature.SignedInfo.References == null || configuration.Signature.SignedInfo.References.Count == 0)
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22705,
+                    Succeeded = false
+                };
+            }
+
+            if (string.IsNullOrWhiteSpace(configuration.ActiveTokenEndpoint))
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22706,
+                    Succeeded = false
+                };
+            }
+
+            if (!Uri.IsWellFormedUriString(configuration.ActiveTokenEndpoint, UriKind.Absolute))
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22707,
+                    Succeeded = false
+                };
+            }
+
+            if (string.IsNullOrWhiteSpace(configuration.TokenEndpoint))
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22708,
+                    Succeeded = false
+                };
+            }
+
+            if (!Uri.IsWellFormedUriString(configuration.TokenEndpoint, UriKind.Absolute))
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22709,
+                    Succeeded = false
+                };
+            }
+
+            if (configuration.SigningKeys == null || configuration.SigningKeys.Count == 0)
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22710,
+                    Succeeded = false
+                };
+            }
+
+            // We know the key used to sign the doc is part of the token signing keys as per the spec.
+            // http://docs.oasis-open.org/wsfed/federation/v1.2/os/ws-federation-1.2-spec-os.html#_Toc223174958:~:text=%3C/fed%3ATargetScopes%20%3E-,3.1.15%20%5BSignature%5D%20Property,-The%20OPTIONAL%20%5BSignature
+            // If the metadata is for a token issuer then the key used to sign issued tokens SHOULD
+            // be used to sign this document.  This means that if a <fed:TokenSigningKey> is specified,
+            // it SHOULD be used to sign this document.
+            foreach (SecurityKey key in configuration.SigningKeys)
+            {
+                try
+                {
+                    configuration.Signature.Verify(key, key.CryptoProviderFactory);
+
+                    return new ConfigurationValidationResult
+                    {
+                        Succeeded = true
+                    };
+                }
+                catch (XmlValidationException)
+                {
+                    // Ignore
+                }
+            }
+            
+            return new ConfigurationValidationResult
+            {
+                ErrorMessage = LogMessages.IDX22711,
+                Succeeded = false
+            };
+        }
+    }
+}
--- a/src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
 using Microsoft.IdentityModel.Tokens;
 using Microsoft.IdentityModel.Xml;
 using static Microsoft.IdentityModel.Logging.LogHelper;
@@ -123,13 +125,51 @@ namespace Microsoft.IdentityModel.Protocols.WsFederation
                 };
             }
 
+            // Get the thumbprint of the cert used to sign the metadata
+            string signingKeyId = string.Empty;
+            var signatureX509Data = configuration.Signature.KeyInfo.X509Data.GetEnumerator();
+
+            if (signatureX509Data.MoveNext())
+            {
+                var signatureCertData = signatureX509Data.Current.Certificates.GetEnumerator();
+                if (signatureCertData.MoveNext())
+                {
+                    X509Certificate2 cert = null;
+
+                    try
+                    {
+                        cert = new X509Certificate2(Convert.FromBase64String(signatureCertData.Current));
+                        signingKeyId = cert.Thumbprint;
+                    }
+                    catch (CryptographicException)
+                    {
+                        return new ConfigurationValidationResult
+                        {
+                            ErrorMessage = LogMessages.IDX22711,
+                            Succeeded = false
+                        };
+                    }
+                    finally
+                    {
+                        if (cert != null)
+                        {
+                            ((IDisposable)cert).Dispose();
+                        }
+                    }
+                }
+            }
+
             // We know the key used to sign the doc is part of the token signing keys as per the spec.
             // http://docs.oasis-open.org/wsfed/federation/v1.2/os/ws-federation-1.2-spec-os.html#_Toc223174958:~:text=%3C/fed%3ATargetScopes%20%3E-,3.1.15%20%5BSignature%5D%20Property,-The%20OPTIONAL%20%5BSignature
             // If the metadata is for a token issuer then the key used to sign issued tokens SHOULD
             // be used to sign this document.  This means that if a <fed:TokenSigningKey> is specified,
             // it SHOULD be used to sign this document.
+
             foreach (SecurityKey key in configuration.SigningKeys)
             {
+                if (signingKeyId != key.KeyId)
+                    continue;
+
                 try
                 {
                     configuration.Signature.Verify(key, key.CryptoProviderFactory);
@@ -141,10 +181,10 @@ namespace Microsoft.IdentityModel.Protocols.WsFederation
                 }
                 catch (XmlValidationException)
                 {
-                    // Ignore
+                    break;
                 }
             }
-            
+
             return new ConfigurationValidationResult
             {
                 ErrorMessage = LogMessages.IDX22711,
--- a/src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
@@ -141,7 +141,7 @@ namespace Microsoft.IdentityModel.Protocols.WsFederation
             if (signatureX509Data.MoveNext())
             {
                 var signatureCertData = signatureX509Data.Current.Certificates.GetEnumerator();
-                if (signatureCertData.MoveNext())
+                if (signatureCertData.MoveNext() && !string.IsNullOrWhiteSpace(signatureCertData.Current))
                 {
                     X509Certificate2 cert = null;
 
--- a/src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
@@ -19,7 +19,7 @@ namespace Microsoft.IdentityModel.Protocols.WsFederation
         /// Validates a WsFederationConfiguration.
         /// </summary>
         /// <param name="configuration">WsFederationConfiguration to validate</param>
-        /// <returns>A <see cref="ConfigurationValidationResult"/> that contains validation result.</returns>
+        /// <returns>A <see cref="ConfigurationValidationResult"/> containing the validation result.</returns>
         /// <exception cref="ArgumentNullException">If the provided configuration is null</exception>
         public ConfigurationValidationResult Validate(WsFederationConfiguration configuration)
         {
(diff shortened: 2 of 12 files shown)


src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
```diff
@@ -248,6 +258,75 @@ protected virtual string ReadPassiveRequestorEndpoint(XmlReader reader)
...
+                {
+                    reader.ReadOuterXml();
+                }
+            }
+
+            // </EndpointReference>
+            reader.MoveToContent();
+            reader.ReadEndElement();
+
+            // </PassiveRequestorEndpoint>
```
brentschmaltz: Should this comment be: // SecurityTokenServiceEndpoint #Resolved

src/Microsoft.IdentityModel.Tokens/BaseConfiguration.cs
```diff
@@ -27,11 +27,17 @@ public virtual ICollection<SecurityKey> SigningKeys
...
         /// Gets or sets the token endpoint specified via the metadata endpoint.
-        /// This can be the fed:SecurityTokenServiceType in WS-Federation, http://docs.oasis-open.org/wsfed/federation/v1.2/os/ws-federation-1.2-spec-os.html#:~:text=fed%3ASecurityTokenSerivceEndpoint
+        /// This can be the fed:PassiveRequestorEndpoint in WS-Federation, https://docs.oasis-open.org/wsfed/federation/v1.2/os/ws-federation-1.2-spec-os.html#:~:text=fed%3ASecurityTokenServiceType/fed%3APassiveRequestorEndpoint
         /// Or the token_endpoint in the OIDC metadata.
         /// </summary>
         public virtual string TokenEndpoint { get; set; }
 
+        /// <summary>
+        /// Gets or sets the token endpoint specified via the metadata endpoint.
+        /// This can is the fed:SecurityTokenServiceType in WS-Federation, http://docs.oasis-open.org/wsfed/federation/v1.2/os/ws-federation-1.2-spec-os.html#:~:text=fed%3ASecurityTokenSerivceEndpoint
```
brentschmaltz: 'This can is' => 'This is' #Resolved

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
```diff
@@ -248,6 +258,75 @@ protected virtual string ReadPassiveRequestorEndpoint(XmlReader reader)
...
+
+            reader.ReadStartElement(WsAddressing.Elements.EndpointReference, WsAddressing.Namespace);
+            reader.MoveToContent();
+
+            if (reader.IsEmptyElement)
+                throw XmlUtil.LogReadException(LogMessages.IDX22814);
+
+            string tokenEndpoint = null;
+
+            while (reader.IsStartElement())
```
brentschmaltz, line 291: If the reader is not positioned on a StartElement what will happen? #Resolved

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
```diff
@@ -317,6 +317,12 @@ public OpenIdConnectConfiguration(string json)
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore, NullValueHandling = NullValueHandling.Ignore, PropertyName = OpenIdProviderMetadataNames.TokenEndpoint, Required = Required.Default)]
         public override string TokenEndpoint { get; set; }
 
+        /// <summary>
+        /// This base class property is not used in OpenIdConnect. 
+        /// </summary>
+        [JsonIgnore]
+        public override string ActiveTokenEndpoint { get; set; }
```
brentschmaltz, line 324: We will modify this in our 7.x release when we will use the utf8Reader and set the ActiveTokenEndpoint to the "token_endpoint" and PassiveTokenEndpoint to "authorization_endpoint".

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationConstants.cs
```diff
@@ -5,6 +5,7 @@ namespace Microsoft.IdentityModel.Protocols.WsFederation
 {
     /// <summary>
     /// Constants for WsFederation.
+    /// As defined on the http://docs.oasis-open.org/wsfed/federation/v1.2/os/ws-federation-1.2-spec-os.html
```
brentschmaltz: 'As defined on the' => 'As defined in: ' #Resolved

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
victorm-hernandez: Yes, updating

src/Microsoft.IdentityModel.Tokens/BaseConfiguration.cs
victorm-hernandez: Fixed

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationConstants.cs
victorm-hernandez: done

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
victorm-hernandez: Since we know that the XML is not malformed, the only scenario where the reader is not on a start element is if the endpointreference has no inner tags and ends right there.  

<wsa:EndpointReference xmlns:wsa=""[http://www.w3.org/2005/08/addressing](http://www.w3.org/2005/08/addressing)""></wsa:EndpointReference>

In that case the MoveToContent below will have no effect and finish reading the SecurityTokenServiceEndpoint tag.

The return value will be null for tokenEndpoint.

victorm-hernandez: The configuration validator class has a check for null on this URI and would catch that problem.

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
```diff
@@ -0,0 +1,195 @@
...
+    /// <summary>
+    /// Defines a class for validating the WsFederationConfiguration.
+    /// </summary>
+    public class WsFederationConfigurationValidator : IConfigurationValidator<WsFederationConfiguration>
+    {
+        /// <summary>
+        /// Validates a WsFederationConfiguration.
+        /// </summary>
+        /// <param name="configuration">WsFederationConfiguration to validate</param>
+        /// <returns>A <see cref="ConfigurationValidationResult"/> that contains validation result.</returns>
```
FuPingFranco: nit: "that contains validation result" -> "containing the validation result" #Resolved

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
```diff
@@ -248,6 +258,75 @@ protected virtual string ReadPassiveRequestorEndpoint(XmlReader reader)
             return tokenEndpoint;
         }
 
+        /// <summary>
+        /// Read fed:SecurityTokenServiceEndpoint element from metadata XML.
```
FuPingFranco, line 262: nit: "Read" -> "Reads"

FuPingFranco (approved):
Looks great!

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
```diff
@@ -0,0 +1,195 @@
...
+                }
+                catch (XmlValidationException)
+                {
+                    break;
+                }
+            }
+
+            return new ConfigurationValidationResult
+            {
+                ErrorMessage = LogMessages.IDX22711,
```
sruke: CryptographicException and signature verification failure log the same message. Consider logging a clearer log message in both places to indicate the reason for signature validation failure.  #Resolved

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
TimHannMSFT: Are we keeping track of the changes we're intending to make/are in the process of making somewhere? We'll definitely want to make sure we have really clear release notes when we go to 7.x so it's clear who's impacted.

This comment is more an FYI for the community then? There's no action for this PR correct?

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
```diff
@@ -0,0 +1,195 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.IdentityModel.Tokens;
+using Microsoft.IdentityModel.Xml;
+using static Microsoft.IdentityModel.Logging.LogHelper;
```
TimHannMSFT, line 9: TIL about using static #Pending

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
```diff
@@ -0,0 +1,195 @@
...
+                throw LogArgumentNullException(nameof(configuration));
+
+            if (string.IsNullOrWhiteSpace(configuration.Issuer))
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22700,
+                    Succeeded = false
+                };
+            }
```
TimHannMSFT, line 36: Consider doing all these checks up front and use StringBuilder to build out an error message. In the current model a developer may face one issue, update, face another issue, update...etc. Getting all the errors up front allows for more actionable errors/fewer iterations of debugging. Also less code maintenance costs IMO since if we decide to add another check it doesn't require another log message.  #WontFix

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
```diff
@@ -0,0 +1,195 @@
...
+            // it SHOULD be used to sign this document.
+
+            foreach (SecurityKey key in configuration.SigningKeys)
+            {
+                if (signingKeyId != key.KeyId)
+                    continue;
+
+                try
+                {
+                    configuration.Signature.Verify(key, key.CryptoProviderFactory);
```
TimHannMSFT, line 184: Looks like this can throw `ArgumentNullException` too. That would be a bit odd since in other cases where there's unexpected input it's returned as a failed `ConfigurationValidationResult`. Is this intentional? #Resolved

src/Microsoft.IdentityModel.Protocols.WsFederation/LogMessages.cs
```diff
@@ -23,17 +23,33 @@ internal static class LogMessages
...
+        internal const string IDX22803 = "IDX22803: Token reference address is missing in PassiveRequestorEndpoint in metadata file.";
         internal const string IDX22804 = "IDX22804: Security token type role descriptor is expected.";
         internal const string IDX22806 = "IDX22806: Key descriptor for signing is missing in security token service type RoleDescriptor.";
         internal const string IDX22807 = "IDX22807: Token endpoint is missing in security token service type RoleDescriptor.";
         internal const string IDX22808 = "IDX22808: 'Use' attribute is missing in KeyDescriptor.";
         internal const string IDX22810 = "IDX22810: 'Issuer' value is missing in wsfederationconfiguration.";
         internal const string IDX22811 = "IDX22811: 'TokenEndpoint' value is missing in wsfederationconfiguration.";
         internal const string IDX22812 = "IDX22812: Element: '{0}' was an empty element. 'TokenEndpoint' value is missing in wsfederationconfiguration.";
+        internal const string IDX22813 = "IDX22813: 'ActiveTokenEndpoint' is missing in 'SecurityTokenServiceTypeRoleDescriptor'.";
```
TimHannMSFT, line 37: nit: [consistency] this is in single quotes here and SecurityTokenServiceEndpoint is not in the next log. Also IDX22806, IDX22807 talk about "token service type RoleDescriptor". I think the way you have it is more clear probably but good to have consistency too for ease of understanding. #Resolved

src/Microsoft.IdentityModel.Tokens/BaseConfiguration.cs
TimHannMSFT: Seems like it's not yet fixed?

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
TimHannMSFT: Looks like this was marked resolved but also not updated in the branch so far as I can tell. @victorm-hernandez do you have local changes which still will be persisted?

TimHannMSFT (approved):
Did not complete looking through all the testing but the core changes LGTM, left a few comments where I believe some improvements could be considered.

src/Microsoft.IdentityModel.Protocols.WsFederation/LogMessages.cs
victorm-hernandez: Good point, will update.

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
victorm-hernandez: You are right, to avoid any confusion I will mark elements as resolved once I sent another iteration.

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
```diff
@@ -191,6 +195,9 @@ protected virtual SecurityTokenServiceTypeRoleDescriptor ReadSecurityTokenServic
             if (string.IsNullOrEmpty(roleDescriptor.TokenEndpoint))
                 LogHelper.LogWarning(LogMessages.IDX22807);
 
+            if (string.IsNullOrEmpty(roleDescriptor.ActiveTokenEndpoint))
```
dannybtsai, line 198: This implies that roleDescriptor.ActiveTokenEndpoint can be null or empty (not an error), but ReadSecurityTokenServiceEndpoint method throws LogMessages.IDX22812 exception if the element is empty. #Resolved

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
```diff
@@ -248,6 +258,75 @@ protected virtual string ReadPassiveRequestorEndpoint(XmlReader reader)
...
+                    // <Address>
+                    XmlUtil.CheckReaderOnEntry(reader, WsAddressing.Elements.Address, WsAddressing.Namespace);
+
+                    if (reader.IsEmptyElement)
+                        throw XmlUtil.LogReadException(LogMessages.IDX22812, WsAddressing.Elements.Address);
+
+                    reader.ReadStartElement(WsAddressing.Elements.Address, WsAddressing.Namespace);
+                    reader.MoveToContent();
+
+                    tokenEndpoint = Trim(reader.ReadContentAsString());
```
dannybtsai, line 304: Do we need to catch exception from ReadContentAsString? #Resolved

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
```diff
@@ -0,0 +1,195 @@
...
+            if (string.IsNullOrWhiteSpace(configuration.Signature.SignatureValue))
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22703,
+                    Succeeded = false
+                };
+            }
+
+            if (string.IsNullOrWhiteSpace(configuration.Signature.SignedInfo.SignatureMethod))
```
dannybtsai: null check SignedInfo #Resolved

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
```diff
@@ -0,0 +1,195 @@
...
+            {
+                return new ConfigurationValidationResult
+                {
+                    ErrorMessage = LogMessages.IDX22704,
+                    Succeeded = false
+                };
+            }
+
+            if (configuration.Signature.SignedInfo.References == null || configuration.Signature.SignedInfo.References.Count == 0)
```
dannybtsai: null check SignedInfo #Resolved

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
```diff
@@ -0,0 +1,195 @@
...
+                try
+                {
+                    configuration.Signature.Verify(key, key.CryptoProviderFactory);
+
+                    return new ConfigurationValidationResult
+                    {
+                        Succeeded = true
+                    };
+                }
+                catch (XmlValidationException)
```
dannybtsai, line 191: 1. Should catch other types of exceptions too?
2. Should break or continue? #Pending

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
victorm-hernandez: The current code assume some of the checks before it. While is not a bad idea it requires a refactor of the whole method. If you dont mind I would like to postpone this improvement.

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
victorm-hernandez: Could you clarify?

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
victorm-hernandez: Good point, added IDX22712

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
victorm-hernandez: This is correct. I understand how this can be confusing. 

This inconsistency is there to avoid a breaking change ActiveTokenEndpoint is a new property that may not be used today by consumers of our library, if they chose to populate this class in any way but our metadata serializer this value may be null. So we cannot really enforce this new property without introducing a breaking change. 

The story is different for the configuration validator. This is a new class that no one is using so it can properly validate that the field exists and has the correct value.

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
victorm-hernandez: good catch

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
victorm-hernandez: This is a good catch. 

On the our scenario this is not a real problem. If we deserialize the configuration using the metadataSerializer class the security key always use the default crypto provider and this value is never null.

Having said that, it is possible to modify this property before calling the validator. We need to handle that scenario.

I will update the logic to check if there is a crypto provider before attempting to verify.

victorm-hernandez:
thanks for the review

---
In reply to: [1470906811](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2100#issuecomment-1470906811) [](http://example.com/codeflow?ancestors=1470906811,1470906811)

victorm-hernandez:
thanks for the review

---
In reply to: [1472717066](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2100#issuecomment-1472717066) [](http://example.com/codeflow?ancestors=1472717066,1472717066)

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
victorm-hernandez: Although the only exception expected is this, the cryptoprovider may potentially throw any other exception.

We could eat the exception and try to continue however given that we checked for the key thumbprint we do not expect any other key to be valid. We do not expect multiple entries with the same thumbprint and different crypto provider factory.

In my opinion the best thing to do is to let that exception bubble up. 

The other alternative I see would be to update the ConfigurationValidationResult object to include exception information, given that that object is shared across ConfigurationValidators I would prefer to not change the pattern.

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs
victorm-hernandez: There is no need, the entry point or this flow has a try/catch block that handles all exceptions

sruke (approved):
LGTM!

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
RojaEnnam: We do want to try rest of the keys in the list if one fails correct? In that case, like Danny suggested, it should be continue.

src/Microsoft.IdentityModel.Protocols.WsFederation/SecurityTokenServiceTypeRoleDescriptor.cs
```diff
@@ -21,13 +21,23 @@ public List<KeyInfo> KeyInfos
         } = new List<KeyInfo>();
 
         /// <summary>
-        /// Token endpoint
+        /// Passive Requestor Token endpoint
+        /// fed:PassiveRequestorEndpoint, https://docs.oasis-open.org/wsfed/federation/v1.2/os/ws-federation-1.2-spec-os.html#:~:text=fed%3ASecurityTokenServiceType/fed%3APassiveRequestorEndpoint
```
RojaEnnam, line 25: Can we add a line or link on what passive requestor is.

src/Microsoft.IdentityModel.Protocols.WsFederation/SecurityTokenServiceTypeRoleDescriptor.cs
```diff
@@ -21,13 +21,23 @@ public List<KeyInfo> KeyInfos
...
         /// </summary>
         public string TokenEndpoint
         {
             get;
             set;
         }
+
+        /// <summary>
+        /// Active Requestor Token Endpoint
```
RojaEnnam, line 34: Same as above, can we add some description on what active requestor is.

PR #2106, opened 2023-06-08
IJsonWebToken and it's usage in OIDC protocol validator

keegan-caruso:


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
```diff
@@ -46,5 +46,11 @@ public class OpenIdConnectProtocolValidationContext
         /// This id_token is assumed to have audience, issuer, lifetime and signature validated.
         /// </summary>
         public JwtSecurityToken ValidatedIdToken { get; set; }
+
+        /// <summary>
+        /// The id_token is assumed to have audience, issuer, lifetime and signature validated.
+        /// If this property is set, then it will be used instead of the <see cref="ValidatedIdToken"/>.
+        /// </summary>
+        public IJsonWebToken ValidatedIdTokenJwt { get; set; }
```
jennyf19, line 54: This is the opt in? What if both are set? Maybe a bit more context in the XML on why I would use one over the other would help. @jmprieur what are your thoughts from a PM perspective.

brentschmaltz: @jennyf19 we can control what Asp.net code does so that will be easier. 
We could throw, if both are set as that is ambiguous and most likely a programming error.
We could enforce an order: IJsonWebToken -> JwtSecurityToken

I don't like ambiguity, so i would vote to throw.

brentschmaltz, line 54: Suggest a name change: ValidateIdTokenJwt -> ValidatedJsonWebToken.
The name is derived from the return value IJsonWebToken.

jennyf19:
>             {

not this PR, but do we have test coverage for this?

---
Refers to: src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs:680 in bd50ede. [](commit_id = bd50ede9182eebde31d47066e69e9831ff1ee7df, deletion_comment = False)

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -562,17 +682,35 @@ protected virtual void ValidateAtHash(OpenIdConnectProtocolValidationContext val
...
+            if (validationContext.ValidatedIdTokenJwt != null)
+            {
+                if (!validationContext.ValidatedIdTokenJwt.TryGetPayloadValue(JsonWebTokens.JwtRegisteredClaimNames.AtHash, out object atHashClaim))
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidAtHashException(LogHelper.FormatInvariant(LogMessages.IDX21312, validationContext.ValidatedIdToken)));
+
+                atHash = atHashClaim as string;
+                if (atHash == null)
+                    throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidAtHashException(LogHelper.FormatInvariant(LogMessages.IDX21311, validationContext.ValidatedIdToken)));
+
+                var idToken = validationContext.ValidatedIdTokenJwt;
```
jennyf19, line 697: >idToken [](http://example.com/codeflow?start=20&length=7)

null check needed?

jennyf19:
>                         // expected type of inner excpetion is: System.Text.Json.JsonReaderException, but that is internal to System.Text.Json so we cannot

nit: exception here and line 604

---
Refers to: test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectProtocolValidatorTests.cs:590 in bd50ede. [](commit_id = bd50ede9182eebde31d47066e69e9831ff1ee7df, deletion_comment = False)

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
GeoK: Would we break developers who are using derived OpenIdConnectProtocolValidator (with virtuals overridden that look only at JwtSecurityToken)?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -21,7 +21,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// <summary>
     /// A <see cref="SecurityToken"/> designed for representing a JSON Web Token (JWT). 
     /// </summary>
-    public class JsonWebToken : SecurityToken
+    public class JsonWebToken : SecurityToken, IJsonWebToken
```
GeoK, line 24: Would be great to get more context on why JwtSecurityToken also doesn't implement IJsonWebToken

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -290,24 +301,25 @@ public virtual void ValidateUserInfoResponse(OpenIdConnectProtocolValidationCont
...
             try
             {
                 // if user info response is a jwt token
-                var handler = new JwtSecurityTokenHandler();
+                var handler = new JsonWebTokenHandler();
                 if (handler.CanReadToken(validationContext.UserInfoEndpointResponse))
                 {
-                    var token = handler.ReadToken(validationContext.UserInfoEndpointResponse) as JwtSecurityToken;
-                    sub = token.Payload.Sub;
+                    var token = handler.ReadToken(validationContext.UserInfoEndpointResponse) as JsonWebToken;
```
brentschmaltz, line 314: It is faster to write: new JsonWebToken(validationContext.UserInfoEndpointResponse), than use the JsonWebTokenHandler.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -508,21 +606,43 @@ protected virtual void ValidateCHash(OpenIdConnectProtocolValidationContext vali
...
-            if (!validationContext.ValidatedIdToken.Payload.TryGetValue(JwtRegisteredClaimNames.CHash, out cHashClaim))
-            {
-                throw LogHelper.LogExceptionMessage(new OpenIdConnectProtocolInvalidCHashException(LogHelper.FormatInvariant(LogMessages.IDX21307, validationContext.ValidatedIdToken)));
-            }
+            string alg;
+            string chash;
 
-            var chash = cHashClaim as string;
-            if (chash == null)
+            if (validationContext.ValidatedIdTokenJwt != null)
```
brentschmaltz, line 612: It may be cleaner to have this method transform the representation of the JWT into a normalized form then call an internal method that has a single format.

That transformation method may apply to different methods: ValidateCHash, ValidateAtHash, ...

src/Microsoft.IdentityModel.Tokens/IJsonWebToken.cs
```diff
@@ -0,0 +1,215 @@
...
+        /// <remarks>
+        /// <para>
+        /// See: https://datatracker.ietf.org/doc/html/rfc7516#section-2
+        /// </para>
+        /// For encrypted tokens {JWE}, this represents the JWT that was encrypted.
+        /// <para>
+        /// If the JWT is not encrypted, this value will be null.
+        /// </para>
+        /// </remarks>
+        public IJsonWebToken InnerToken { get; }
```
brentschmaltz, line 109: If you are shooting for lexicographic order, InnerToken is before Issuer.

I am understanding why we need this property.
How does it help us?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
keegan-caruso: should be covered! see line 688, we check that validationContext.ValidatedIdTokenJwt is not null

src/Microsoft.IdentityModel.Tokens/IJsonWebToken.cs
keegan-caruso: this is for getting the algorithm off of the inner token

It is needed for ValidateAtHash and ValidateCHash

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
keegan-caruso: How header parameters and claims are retrieved are a little bit different. It can be done fairly straight forward with an adapter over JwtSecurityToken where the adapter implements the interface. I don't see the need for us to produce that adapter at this time. Do you see it differently?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidationContext.cs
keegan-caruso: +1 to George's point, if we throw on both being set we might break implementations.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
GeoK: It crossed my mind that it might be easier to work only with IJsonWebToken in OIDCProtocolValidator but I assumed that there must be a tradeoff there :)

PR #2113, opened 2023-06-11
Update comments for SecurityToken.ReadToken method

mgblackwater:
Updated the comments for the SecurityToken.ReadToken method to reflect the actual return type, which is SecurityToken instead of JwtSecurityToken.


brentschmaltz:
@mgblackwater i think it is ok to that the comments use JwtSecurityToken, this gives users knowledge of the actual type returned.

Issue #2118, opened 2023-06-22
[Bug] Unix systems (such as macOS and Linux) aren't building NET6.0 TFM 


GregDomzalski:
**Which version of Microsoft.IdentityModel are you using?**
Note that to get help, you need to run the latest version. 
Latest dev / 6.31.0

**Where is the issue?**
 * [ ] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.ManagedKeyVaultSecurityKey
 * [ ] M.IM.Protocols
 * [ ] M.IM.Protocols.OpenIdConnect
 * [ ] M.IM.Protocols.SignedHttpRequest
 * [ ] M.IM.Protocols.WsFederation
 * [ ] M.IM.TestExtensions
 * [ ] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [ ] S.IM.Tokens.Jwt
 * [x] Other (please describe)

Project build system: `build/targets.props`

**Is this a new or an existing app?**
N/A

**Repro**

On macOS after cloning the repo:
```shell
./build.sh
```

**Expected behavior**
Both netstandard2.0 and net6.0 TFMs are present in the generated NuGet packages.

**Actual behavior**
Only netstandard2.0 is present. net6.0 is missing.

**Possible solution**
Add `net6.0` TFM to `SrcStandardTargets` in `targets.props` and `TestOnlyCoreTargets` in `targetsText.props`

**Additional context / logs / screenshots / links to code**
I'm getting ready to submit additional bug reports and fix PRs that have to do with using ECDH + JWE. ECDH is not available in netstandard2.0. Since I develop on a mac, I must develop against the net6.0 and not the netfx TFMs.

jennyf19:
@GregDomzalski does it repro on .NET7 or .NET8-preview?

GregDomzalski:
I repro'd it with .NET6 and .NET7 SDKs. I have not tried any of the 8 previews, but I suspect the same issue will be present. I'd be happy to try that anyways - just in case - if you'd like.

PR #2119, opened 2023-06-22
Build NET6.0 TFM on Unix based systems

GregDomzalski:
This PR addresses #2118 

Unix systems such as Linux and macOS are capable of running any Core/5+ version of .NET. There are sections of code (specifically ECDH support) that will only build in recent versions of .NET Framework 4.x and .NET 6.0 and are not present in the `netstandard2.0` TFM.

A straightforward fix would be to add the `net6.0` TFM to the `SrcStandardTargets`.

If `SrcStandardTargets` wants to remain `netstandard` TFMs only, then an alternate solution would be to have `SrcTargets` conditionally include the `net4*` targets when building on a Windows host.

*What about the test targets?* I'm already having issues running the tests on macOS using `netcoreapp`. Targeting .NET6 also introduces some macOS-specific oddities with .NET 6's implementation of X509 certificates. I've encountered these errors on my test projects as well, where well-formed certs and other materials are now throwing 'malformed' exceptions. I honestly don't know how to proceed there without either diagnosing the problem in .NET's crypto implementation, or by regenerating all of the test certificates using OpenSSL instead of the Windows-based APIs. I would honestly be fine developing/building on macOS and then running the tests in a VM.

### DIFF
--- a/build/targets.props
+++ b/build/targets.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
     <SrcTargets>net45;net461;net462;net472;netstandard2.0;net6.0</SrcTargets>
-    <SrcStandardTargets>netstandard2.0</SrcStandardTargets>
+    <SrcStandardTargets>netstandard2.0;net6.0</SrcStandardTargets>
   </PropertyGroup>
 </Project>


jmprieur:
~@GregDomzalski : any reason why you closed the PR?~
Sorry for the noise. I just saw Brent's comment in the issue #2120

PR #2121, opened 2023-06-22
add net8 target

westin-m:
remove hintpath referencing net45


test/Microsoft.IdentityModel.AotCompatibility.Tests/Microsoft.IdentityModel.AotCompatibility.Tests.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <!-- This test only needs to run on .NET -->
-    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
```
brentschmaltz: Why is this net6.0 and net8.0, where others are just net8.0

brentschmaltz (approved):

PR #2122, opened 2023-06-26
conditional compilation for net 8

### LINKED ISSUE #2111: [Feature Request] Conditional compilation for net8.0
As of #2108, net8.0 is a required build target. Add a condition to the build props that targets net8.0 if an environment variable is set, making it opt-in.

westin-m:
opt in for targeting net 8 through env variable

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -42,8 +42,9 @@
     <NoWarn>$(NoWarn);SYSLIB0050</NoWarn>
     <NoWarn>$(NoWarn);SYSLIB0051</NoWarn>
   </PropertyGroup>
-
-  <ItemGroup>
+  
+  <ItemGroup Condition="'$(TargetFramework)' != 'net8.0'">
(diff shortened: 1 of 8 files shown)


build/common.props
```diff
@@ -42,8 +42,15 @@
     <NoWarn>$(NoWarn);SYSLIB0050</NoWarn>
     <NoWarn>$(NoWarn);SYSLIB0051</NoWarn>
   </PropertyGroup>
-
-  <ItemGroup>
+  
+  <PropertyGroup>
+    <EnableMicrosoftExtensionsConfigurationBinderSourceGenerator>
+      true
```
jmprieur: Does Wilson use `Configuration.Bind`?
This is for Id.Web, not sure for Wilson!

jennyf19 (approved):
LGTM....will you do something similar in Id Web?

jmprieur (approved):
Thanks @westin-m 
I checked, this build on a machine where  .NET 8 is not installed!

build/common.props
keegan-caruso: same question, I don't see why this is needed

PR #2131, opened 2023-07-10
Possible fixes to address perf issue in 6.31

jennyf19:


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -904,7 +906,11 @@ public string Subject
         /// <returns>Encoded token string without signature or authentication tag.</returns>
         public override string ToString()
         {
-            return EncodedToken.Substring(0, EncodedToken.LastIndexOf("."));
+            int lastDot = EncodedToken.LastIndexOf('.');
```
henrik-me, line 908: >int lastDot = EncodedToken.LastIndexOf('.'); [](http://example.com/codeflow?start=12&length=44)

can we store lastDot and only recalculated when encodedtoken has been updated?

jennyf19: As discussed, leaving as-is.

henrik-me: I think we should change and optimize this code a bit? Can't we leverage what has already been calculated in Readtoken above? IsSigned? and dot2? What if this is not a JWS but a JWE?

jmprieur: - ReadToken is not called in all the constructors.
- removing after the last dot works both for JWT and JWEs (after 2nd dot would not work for JWE)
- Why would we want to optimize the ToString() which is called only in exceptions.

jennyf19: `JWT` -> you mean JWS in this case.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -56,6 +56,7 @@ public class JsonWebToken : SecurityToken
         private DateTime? _validTo;
         private string _x5t;
         private string _zip;
+        private int _lastDot = -1;
```
jmprieur: I would do `= -2`

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -904,7 +906,14 @@ public string Subject
         /// <returns>Encoded token string without signature or authentication tag.</returns>
         public override string ToString()
         {
-            return EncodedToken.Substring(0, EncodedToken.LastIndexOf("."));
+            if (_lastDot < 0)
```
jmprieur: and here `== -2`

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -904,7 +906,14 @@ public string Subject
...
+            {
+                _lastDot = EncodedToken.LastIndexOf('.');
```
jmprieur: as if there is no dot, _lastDot will be -1.

jmprieur (commented):
Suggested one change for Henrik's optimization
(as `_lastDot = EncodedToken.LastIndexOf('.');` will return -1
if there is no dot, so we would re-run it again next time.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
GeoK: JWTs without any dots would fail on a number of places in this and other classes. Not sure about the value for optimizing these cases.

GeoK (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
jennyf19: fixed

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
henrik-me: it's better :) nit: from original comment "and only recalculated when encodedtoken has been updated"

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
henrik-me: my original ask was to persist the string returned instead of creating it every time. we have a ton of things we try to only calculate once here.

jmprieur (approved):
LGTM
Thanks @jennyf19

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -365,15 +365,15 @@ public static string FormatInvariant(string format, params object[] args)
...
         }
 
-        private static string SanitizeSecurityArtifact(object arg)
+        private static object SanitizeSecurityArtifact(object arg)
         {
             if (arg == null)
                 return "null";
 
-            if (arg is ISafeLogSecurityArtifact && IdentityModelEventSource.LogCompleteSecurityArtifact)
+            if (IdentityModelEventSource.LogCompleteSecurityArtifact && arg is ISafeLogSecurityArtifact)
```
dannybtsai, line 373: Use in-line is operator:

            if (IdentityModelEventSource.LogCompleteSecurityArtifact && arg is ISafeLogSecurityArtifact safeLogSecurityArtifact)
                return safeLogSecurityArtifact.UnsafeToString();

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -120,8 +120,9 @@ public JsonWebToken(string header, string payload)
                 throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14302, payload), ex));
             }
 
-            _encodedHeader = header;
-            _encodedPayload = payload;
+            _encodedHeader = Base64UrlEncoder.Encode(header);
```
dannybtsai, line 123: This is different than the logic in the EncodedHeader where `encodedHeader = EncodedToken.Substring(0, Dot1)`

        public string EncodedHeader
        {
            get
            {
                // TODO - need to account for JWE
                if (_encodedHeader == null)
                {
                    if (EncodedToken != null)
                        _encodedHeader = EncodedToken.Substring(0, Dot1);
                    else
                        _encodedHeader = string.Empty;
                }

                return _encodedHeader;
            }
        }

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -120,8 +120,9 @@ public JsonWebToken(string header, string payload)
...
+            _encodedPayload = Base64UrlEncoder.Encode(payload);
```
dannybtsai, line 124: In the EncodedPayload property, the _encodedPayload is set to:

`_encodedPayload = IsEncrypted ? string.Empty : EncodedToken.Substring(Dot1 + 1, Dot2 - Dot1 - 1)`

PR #2139, opened 2023-07-12
Initial addition of Nullable property to Abstractions Project.

FuPingFranco:
Initial PR to Abstractions project to enable nullable across the solution allowing variables to have null as a value, which can be useful in cases where a value may not be present or when a value is optional. It also allows for the use of null-forgiving operator, which suppresses nullable warnings for an expression.

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -31,7 +31,7 @@
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(SrcTargets)</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">$(SrcStandardTargets)</TargetFrameworks>
     <NetStandardImplicitPackageVersion>$(NetStandardVersion)</NetStandardImplicitPackageVersion>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>10</LangVersion>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
--- a/build/common.props
+++ b/build/common.props
@@ -31,7 +31,7 @@
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(SrcTargets)</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">$(SrcStandardTargets)</TargetFrameworks>
     <NetStandardImplicitPackageVersion>$(NetStandardVersion)</NetStandardImplicitPackageVersion>
-    <LangVersion>10</LangVersion>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
--- a/src/Microsoft.IdentityModel.Abstractions/NullTelemetryClient.cs
+++ b/src/Microsoft.IdentityModel.Abstractions/NullTelemetryClient.cs
@@ -25,7 +25,9 @@ namespace Microsoft.IdentityModel.Abstractions
         /// <remarks>
         /// Private constructor to prevent the default constructor being exposed.
         /// </remarks>
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
         private NullTelemetryClient() { }
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
 
         /// <inheritdoc />
         public bool IsEnabled() => false;
@@ -48,12 +50,12 @@ namespace Microsoft.IdentityModel.Abstractions
         /// <inheritdoc/>
         public void TrackEvent(
             string eventName,
-            IDictionary<string, string> stringProperties = null,
-            IDictionary<string, long> longProperties = null,
-            IDictionary<string, bool> boolProperties = null,
-            IDictionary<string, DateTime> dateTimeProperties = null,
-            IDictionary<string, double> doubleProperties = null,
-            IDictionary<string, Guid> guidProperties = null)
+            IDictionary<string, string>? stringProperties = null,
+            IDictionary<string, long>? longProperties = null,
+            IDictionary<string, bool>? boolProperties = null,
+            IDictionary<string, DateTime>? dateTimeProperties = null,
+            IDictionary<string, double>? doubleProperties = null,
+            IDictionary<string, Guid>? guidProperties = null)
         {
             // no-op
         }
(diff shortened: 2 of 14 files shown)


build/common.props
```diff
@@ -31,7 +31,7 @@
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(SrcTargets)</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">$(SrcStandardTargets)</TargetFrameworks>
     <NetStandardImplicitPackageVersion>$(NetStandardVersion)</NetStandardImplicitPackageVersion>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>10</LangVersion>
```
jmprieur: you could use **latest** instead of 10.

westin-m:
```suggestion
    <LangVersion>latest</LangVersion>
```
This could be better than choosing a static version.

src/Microsoft.IdentityModel.Abstractions/NullTelemetryClient.cs
```diff
@@ -25,7 +25,9 @@ public class NullTelemetryClient : ITelemetryClient
         /// <remarks>
         /// Private constructor to prevent the default constructor being exposed.
         /// </remarks>
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
```
jmprieur, line 28: Is it the ClientId?

jmprieur (commented):
LGTM
thanks @FuPingFranco

src/Microsoft.IdentityModel.Abstractions/NullTelemetryClient.cs
FuPingFranco: Yes, we should never have a null clientId in my opinion. What are you're thoughts on this?

build/common.props
FuPingFranco: Yes, I forgot about our discussion this morning. Thanks for reminding me Jean-Marc!

FuPingFranco: Commited Thanks Westin!!

src/Microsoft.IdentityModel.Abstractions/NullTelemetryClient.cs
jennyf19: What if we do have one?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/GlobalSuppressions.cs
```diff
@@ -12,3 +12,4 @@
 [assembly: SuppressMessage("Globalization", "CA1307:Specify StringComparison", Justification = "Vendored component", Scope = "module")]
 [assembly: SuppressMessage("Design", "CA1001:Types that own disposable fields should be disposable", Justification = "Breaking change", Scope = "type", Target = "~T:Microsoft.IdentityModel.Protocols.SignedHttpRequest.SignedHttpRequestHandler")]
 [assembly: SuppressMessage("Usage", "CA2227:Collection properties should be read only", Justification = "Previously released as read / write", Scope = "member", Target = "~P:Microsoft.IdentityModel.Protocols.SignedHttpRequest.SignedHttpRequestInvalidNonceClaimException.PropertyBag")]
+[assembly: SuppressMessage("Globalization", "CA1305:Specify IFormatProvider", Justification = "Breaking Change", Scope = "member", Target = "~T:Microsoft.IdentityModel.Protocols.SignedHttpRequest.SignedHttpRequestHandler)")]
```
jennyf19, line 15: >[assembly: SuppressMessage( [](http://example.com/codeflow?start=0&length=27)

why is this needed?
 #Resolved

FuPingFranco: There's a couple of methods from the Microsoft.IdentityModel.Protocols.SignedHttpRequest project that were throwing error code [CA1305](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1305). After some investigation, it essentially was that we should try and define a certain format for a StringBuilder obj that's being created from headers information within the methods so, doing that did not seem to be very useful in those cases as they deal with headers, and we shouldn't format those. 

I opted for suppresing this error instead as it is a currently working functionality.

IFormatProvider doc: https://learn.microsoft.com/en-us/dotnet/api/system.iformatprovider?view=net-7.0

src/Microsoft.IdentityModel.Abstractions/NullTelemetryClient.cs
FuPingFranco: I am not entirely sure what it would happen on the kusto/db side however, not having a clientID for telemetry purposes would basically defeat its purpose I believe.

jmprieur: Then in that case add a constructor taking the ClientId?
It's the null telemetry client though. Meaning it probably does nothing?

FuPingFranco: Correct, It doesn't do anything. I think just disabling the alert in here should suffice so I can move on to the other projects. What do you think? @jmprieur

keegan-caruso: We need to be mindful here.

If there is a case that ClientId can be null we should annotate clientId property that it can be null. Otherwise we are not respecting the contract and we defeat the purpose of using the nullable checks. This could results in a NRE.

If the contract should instead be that the clientId cannot be null, we need to change the NullTelemetryClient to always set the property.

jmprieur: for example by changing the constructor to take it as a parameter

FuPingFranco: I agree we have to be mindful of this to respect the current contract however, this particular property is tricky because it is not very suitable to use the nullable property as it is written today, changing the signature of the constructor would change its current functionality and we'll have to use some sort of default value in order for that to be viable with is not a good approach in my opinion. I synced with Keegan over this and seems that disabling the alert on this property made the most sense for now.

jennyf19: @keegan-caruso do you sign-off on the PR? Thanks for the update @FuPingFranco

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/GlobalSuppressions.cs
jennyf19: sounds good. thanks for the addtional context.

jennyf19 (approved):
:shipit:

jmprieur (approved):
LGTM
Thanks @FuPingFranco

keegan-caruso (approved):

PR #2161, opened 2023-07-26
Avoid unnecessary boxing in LogHelper

stephentoub:


jennyf19 (approved):
Logging is a huge issue for us, and we are looking to re-do it all-up.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -291,7 +289,7 @@ public static void LogVerbose(string message, params object[] args)
             if (IdentityModelEventSource.Logger.IsEnabled())
                 IdentityModelEventSource.Logger.WriteVerbose(message, args);
 
-            if (Enum.IsDefined(typeof(EventLogLevel), (int)EventLevel.Verbose) && Logger.IsEnabled((EventLogLevel)EventLevel.Verbose))
+            if (Logger.IsEnabled(EventLogLevel.Verbose))
                 Logger.Log(WriteEntry((EventLogLevel)EventLevel.Verbose, null, message, args));
```
keegan-caruso: Can we directly use EventLogLevel.Verbose here as well?

stephentoub: Yup, I missed that one, thanks (it won't change the code gen, but it'll certainly be cleaner)

stephentoub:
```suggestion
                Logger.Log(WriteEntry(EventLogLevel.Verbose, null, message, args));
```

keegan-caruso (approved):

PR #2162, opened 2023-07-26
Reduce allocation with Base64UrlEncoder

stephentoub:
NOTE: This requires https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2158 before it will successfully build and run.  It depends on the System.Memory package reference added there for downlevel builds.

- Stop calling ToCharArray to produce inputs to it; instead pass in a ReadOnlyMemory sliced appropriately
- Move the char[] s_base64Table into a u8 span
- Use IndexOfAny to determine whether replacements are needed in UnsafeDecode
- Stackalloc a span or rent a char[] from the ArrayPool when a temporary is needed in UnsafeDecode

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -17,7 +17,6 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// </summary>
     public class JsonWebToken : SecurityToken
     {
-        private char[] _hChars;
         private ClaimsIdentity _claimsIdentity;
         private bool _wasClaimsIdentitySet;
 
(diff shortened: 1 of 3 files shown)


brentschmaltz (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -470,25 +469,39 @@ private void ReadToken(string encodedJson)
                     throw LogHelper.LogExceptionMessage(new SecurityTokenMalformedException(LogHelper.FormatInvariant(LogMessages.IDX14310, encodedJson)));
 
                 // right number of dots for JWE
-                _hChars = encodedJson.ToCharArray(0, Dot1);
+                ReadOnlyMemory<char> hChars = encodedJson.AsMemory(0, Dot1);
```
eerhardt, line 472: (nit) what does `h` mean in `hChars`? "header"?

eerhardt, line 472: Why `ReadOnlyMemory<char>` and not `ReadOnlySpan<char>`?

eerhardt: Oh, I think I just found the answer in UnsafeDecode.

`MemoryMarshal.TryGetArray(str`

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -470,25 +469,39 @@ private void ReadToken(string encodedJson)
...
                     throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX14102, encodedJson.Substring(0, Dot1), encodedJson), ex));
                 }
 
                 // dir does not have any key bytes
-                char[] encryptedKeyBytes = encodedJson.ToCharArray(Dot1 + 1, Dot2 - Dot1 - 1);
-                if (encryptedKeyBytes.Length != 0)
+                ReadOnlyMemory<char> encryptedKeyBytes = encodedJson.AsMemory(Dot1 + 1, Dot2 - Dot1 - 1);
+                if (!encryptedKeyBytes.IsEmpty)
                 {
                     EncryptedKeyBytes = Base64UrlEncoder.UnsafeDecode(encryptedKeyBytes);
```
eerhardt, line 506: This is a comment for existing code, but all the other calls to `Base64UrlEncoder.UnsafeDecode` have a try-catch around them, but this one doesn't. Do we know why?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -168,106 +161,100 @@ internal static string EncodeString(string str)
...
-                    fixed (char* dest = decodedString)
+                    int pos;
+                    while ((pos = charsSpan.IndexOfAny(base64UrlCharacter62, base64UrlCharacter63)) >= 0)
                     {
-                        Buffer.MemoryCopy(src, dest, str.Length * 2, str.Length * 2);
-
-                        dest[str.Length] = base64PadCharacter;
-                        if (str.Length + 2 == decodedLength)
-                            dest[str.Length + 1] = base64PadCharacter;
+                        charsSpan[pos] = charsSpan[pos] == base64UrlCharacter62 ? base64Character62 : base64Character63;
```
eerhardt, line 213: Should we slice charsSpan here, so we don't keep searching the beginning over and over?

eerhardt (approved):
This looks like a nice win. LGTM

keegan-caruso (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
stephentoub: No idea, I just kept the original naming :)

stephentoub: Exactly.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
stephentoub: Ooh, yes, definitely.  Nice catch. I'll submit a PR to fix that.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
eerhardt: cc @brentschmaltz

Issue #2163, opened 2023-07-26
Calls to LogHelper incur cost even when logging is disabled


stephentoub:
Many calls to LogHelper are incurring allocation and other costs even when logging is disabled.  More than half of all the calls to LogInformation, for example, pass a parameter that will cause the call site to allocate an array for the params array argument.  Other call sites box value type arguments to be passed in that params array as an object. Etc.

jennyf19:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/releases/tag/7.0.0-preview2

PR #2164, opened 2023-07-26
Guard logging calls to avoid unnecessary work at call site

stephentoub:
(Ignore the first commit. It is a copy of the commit from https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2161, which this PR then builds on.)

Many LogHelper.Log calls are doing work at the call site, such as allocating params arrays, boxing structs, formatting strings, and so on, even when the data will be thrown away because logging (or logging for that verbosity level) isn't enabled.  This adds an IsEnabled method to LogHelper and uses it at any call site where there's such work to be avoided.

Fixes https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2163

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -9,6 +9,7 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.IdentityModel.Abstractions;
 using Microsoft.IdentityModel.Logging;
 using Microsoft.IdentityModel.Tokens;
 using Newtonsoft.Json;
@@ -158,7 +159,9 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
             if (token.Length > MaximumTokenSizeInBytes)
             {
-                LogHelper.LogInformation(TokenLogMessages.IDX10209, LogHelper.MarkAsNonPII(token.Length), LogHelper.MarkAsNonPII(MaximumTokenSizeInBytes));
+                if (LogHelper.IsEnabled(EventLogLevel.Informational))
+                    LogHelper.LogInformation(TokenLogMessages.IDX10209, LogHelper.MarkAsNonPII(token.Length), LogHelper.MarkAsNonPII(MaximumTokenSizeInBytes));
+
                 return false;
             }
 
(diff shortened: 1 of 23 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -159,7 +160,9 @@ public virtual bool CanReadToken(string token)
 
             if (token.Length > MaximumTokenSizeInBytes)
             {
-                LogHelper.LogInformation(TokenLogMessages.IDX10209, LogHelper.MarkAsNonPII(token.Length), LogHelper.MarkAsNonPII(MaximumTokenSizeInBytes));
+                if (LogHelper.IsEnabled(EventLogLevel.Informational))
```
jennyf19, line 162: nice catch

jennyf19 (approved):
This is great. We should the same in our other library as well, I think we have a similar issue there.

keegan-caruso:
@stephentoub merge conflict from #2161

stephentoub:
> merge conflict from https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2161

Yup, thanks, fixed.  I'd submitted this with the commit from that one, just deleted that first commit here with a rebase.

jennyf19:
@stephentoub looks like there are still merge conflicts?

stephentoub:
> looks like there are still merge conflicts?

Not "still", but "again" :)  I'll fix them again.

jennyf19:
> > looks like there are still merge conflicts?
> 
> Not "still", but "again" :) I'll fix them again.

Hard to keep up with you!

brentschmaltz (approved):

stephentoub:
Conflicts resolved.

jennyf19:
@keegan-caruso  and @brentschmaltz the tests via the GitHub action is not working properly, or there is a test failure not bubbling up nicely. We will have to test this one locally for now.

keegan-caruso:
@jennyf19 I don't understand, when I pull this one locally the test run hangs

jennyf19:
> @jennyf19 I don't understand, when I pull this one locally the test run hangs

That makes sense...But that test doesn't normally hang?

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -110,27 +111,29 @@ public virtual KeyInfo ReadKeyInfo(XmlReader reader)
...
 
                         // </KeyValue>
                         reader.ReadEndElement();
                     }
                     else
                     {
                         // Skip the element since it is not one of  <RetrievalMethod>, <X509Data>, <KeyValue>
-                        LogHelper.LogWarning(LogMessages.IDX30300, reader.ReadOuterXml());
+                        if (LogHelper.IsEnabled(EventLogLevel.Warning))
+                            LogHelper.LogWarning(LogMessages.IDX30300, reader.ReadOuterXml());
```
BrennanConroy, line 143: Does this change behavior? `reader.ReadOuterXml()` isn't called unless logging is on now.

Should it call `reader.Skip()` in an `else`?

stephentoub: Ugh, yes, it does, and should, or at the very least it should store the result into a local outside of the if. Good spot.

jennyf19 (approved):

PR #2167, opened 2023-07-26
Remove some Task allocations from JsonWebTokenHandler

stephentoub:
These methods are all private and all of the call sites directly await them.  They can be ValueTasks to avoid allocating on the synchronously-completing path (which I see happening reasonably frequently in traces).


dosper7:
Just out of curiosity and for learning purposes, can you share one of those traces? To have an idea of how you check this things? Thanks.

westin-m (approved):

keegan-caruso (approved):

stephentoub:
> Just out of curiosity and for learning purposes, can you share one of those traces? To have an idea of how you check this things? Thanks.

Sorry, @dosper7, I don't have them available. But it wasn't anything fancy, just a `dotnet trace` taken of a system interacting with JWTs.  Actually, two traces, one for CPU and one for allocation. For this PR and others related to allocation, I just looked through the list at each type being allocated, looked to see where they were coming from, and looked to see whether they were easily avoidable.  I typically start by looking at the strings and arrays, as there's often low-hanging fruit there and they're also often at the top of the most-allocated list, and also look for things that stand out as not belonging, e.g. when value types show up in the allocation list, that means boxing, which is often avoidable.

PR #2169, opened 2023-07-26
Remove closure from JsonWebTokenHandler.ValidateSignature

stephentoub:
The Any call on an exceptional path is closing over the jwtToken parameter, which means a display class gets allocated at the beginning of the method, whether that code path is taken or not.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -1760,7 +1760,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             {
                 if (kidMatched)
                 {
-                    var isKidInTVP = keysInTokenValidationParameters.Any(x => x.KeyId.Equals(jwtToken.Kid));
+                    JsonWebToken localJwtToken = jwtToken; // avoid closure on non-exceptional path
+                    var isKidInTVP = keysInTokenValidationParameters.Any(x => x.KeyId.Equals(localJwtToken.Kid));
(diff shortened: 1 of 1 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1760,7 +1760,8 @@ private static JsonWebToken ValidateSignature(JsonWebToken jwtToken, TokenValida
             {
                 if (kidMatched)
                 {
-                    var isKidInTVP = keysInTokenValidationParameters.Any(x => x.KeyId.Equals(jwtToken.Kid));
+                    JsonWebToken localJwtToken = jwtToken; // avoid closure on non-exceptional path
```
eerhardt, line 1763: This is so tricky to catch both when coding and in review. Is there a way a static analyzer could catch this? It would probably be super noisy.

I assume the best way to find these today is to look at the allocations in a trace?

stephentoub: > I assume the best way to find these today is to look at the allocations in a trace?

They certainly jump out at you there. "What is <>__DisplayClass_52 and why is it being allocated 18,000 times?"

In other projects, every now and then I'll also load up all the relevant assemblies in ILSpy, search for DisplayClass, and just flip through all of them to see whether they make sense.

Of course, the real solution is to avoid the patterns that lead to this in the first place, e.g. don't use `Any(x => ...)` and you won't close over anything, or use APIs that allow the TState to be passed in. And if you can avoid closing over stuff, protect it by using `static` on the lambda so they don't then regress.

keegan-caruso (approved):

PR #2171, opened 2023-07-26
Avoid allocating exceptional StringBuilder until there's a need

stephentoub:


jennyf19 (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -1193,16 +1193,16 @@ internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken
...
-                keysAttempted.AppendLine(key.ToString());
+                (keysAttempted ??= new StringBuilder()).AppendLine(key.ToString());
             }
 
-            if (unwrappedKeys.Count > 0 && exceptionStrings.Length == 0)
+            if (unwrappedKeys.Count > 0 && exceptionStrings is null)
                 return unwrappedKeys;
             else
-                throw LogHelper.LogExceptionMessage(new SecurityTokenKeyWrapException(LogHelper.FormatInvariant(TokenLogMessages.IDX10618, keysAttempted, exceptionStrings, jwtToken)));
+                throw LogHelper.LogExceptionMessage(new SecurityTokenKeyWrapException(LogHelper.FormatInvariant(TokenLogMessages.IDX10618, (object)keysAttempted ?? "", exceptionStrings, jwtToken)));
```
keegan-caruso: does exceptionStrings need the same null check?

stephentoub: Yes, it should ideally have it. Without it, it'll end up saying "null" rather than "".

stephentoub:
```suggestion
                throw LogHelper.LogExceptionMessage(new SecurityTokenKeyWrapException(LogHelper.FormatInvariant(TokenLogMessages.IDX10618, (object)keysAttempted ?? "", (object)exceptionStrings ?? "", jwtToken)));
```

PR #2174, opened 2023-07-26
The first step in replacing newtonsoft with System.Text.Json

brentschmaltz:
Removed our internal newtonsoft, all our code uses the public newtonsoft.
This will ease the work to use System.Json.Text on a class-by-class basis.
The use of newtonsoft will be removed in all src code when we ship, but may remain in test code.

Removed tests that were designed to ensure that changes from 4.x -> 5.x where switched from System.IdentityModel.Tokens to Microsoft.IdentityModel.Tokens were not breaking, we do not need this code anymore.

Removed some old performance code that was moved to its own repo.
Cleaned up 'usings'.
Set CLSCompliant to false in all test code (Assemblyinfo.cs) so #ifdef's could be removed in code.


jennyf19 (approved):
:shipit:

build/common.props
```diff
@@ -45,6 +45,7 @@
   
   <ItemGroup Condition="'$(TargetFramework)' != 'net8.0'">
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="$(MicrosoftSourceLinkGitHubVersion)" PrivateAssets="All"/>
+    <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftVersion)" />
```
eerhardt, line 52: Does this code build if you are targeting `net8.0`? How does this PackageReference come in that case?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/Properties/AssemblyInfo.cs
```diff
@@ -5,7 +5,7 @@
 using System.Runtime.InteropServices;
 using Xunit;
 
-[assembly: CLSCompliant(true)]
+[assembly: CLSCompliant(false)]
```
eerhardt, line 8: Why are these CLSCompliant attributes changing?

brentschmaltz: We don't need to be CLSCompliant in test.

build/common.props
brentschmaltz: I will have to fix that.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/Properties/AssemblyInfo.cs
eerhardt: I would just remove these lines then. If you don't have a CLSCompliant attribute, the assembly is not CLSCompliant.

westin-m (approved):

PR #2177, opened 2023-07-26
Francofung/adding nullable post rebase

FuPingFranco:
Second PR for logging and KeyVaultExtension dll's with the goal to enable nullable across the solution. This allows variables to have null as a value, which can be useful in cases where a value may not be present or when a value is optional. It also allows for the use of null-forgiving operator, which suppresses nullable warnings for an expression that has been checked for nulls.


keegan-caruso:
@FuPingFranco can you target dev7 instead? And rebase as well

PR #2180, opened 2023-07-27
Remove several Lazy-related objects from every TokenValidationResult

stephentoub:
Creating a TokenValidationResult is also creating a `Lazy<>`, a `LazyHelper` internal to the `Lazy<>`, and a `Func<>` delegate due to the lambda passed to the lazy closing over `this`.

It's not clear to me why `Lazy<>` was being used here. There's other lazily-initialized state on the same type that's not using `Lazy<>`, and I went back and looked at the PR/issue that added this and there's no discussion of it.  Is there some kind of thread-safety guarantee it was trying to provide?  Or was this just done this way because that's how the dev happened to write it?

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
@@ -70,14 +70,12 @@ namespace Microsoft.IdentityModel.Tokens
                 if (!_hasIsValidOrExceptionBeenRead)
                     LogHelper.LogWarning(LogMessages.IDX10109);
 
-                IDictionary<string, object> claims = _claims;
-                if (claims is null && ClaimsIdentity is { } ci)
+                if (_claims is null && ClaimsIdentity is { } ci)
                 {
                     Interlocked.CompareExchange(ref _claims, TokenUtilities.CreateDictionaryFromClaims(ci.Claims), null);
-                    claims = _claims;
                 }
 
-                return claims;
+                return _claims;
             }
         }
 
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Security.Claims;
+using System.Threading;
 using Microsoft.IdentityModel.Logging;
 
 namespace Microsoft.IdentityModel.Tokens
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
```diff
@@ -56,7 +54,7 @@ internal TokenValidationResult(SecurityToken securityToken, TokenHandler tokenHa
                 if (!_hasIsValidOrExceptionBeenRead)
                     LogHelper.LogWarning(LogMessages.IDX10109);
 
-                return _claims.Value;
+                return _claims ??= TokenUtilities.CreateDictionaryFromClaims(ClaimsIdentity?.Claims);
```
brentschmaltz: The current code is thread safe, we should maintain that contract.

stephentoub:
@brentschmaltz, based on our offline conversation, I revamped it.  Please take a look and let me know if this matches your expectations for what's synchronized and how.

src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
```diff
@@ -14,20 +16,33 @@ namespace Microsoft.IdentityModel.Tokens
...
+        // Fields lazily initialized in a thread-safe manner. _claimsIdentity is protected by the _claimsIdentitySyncObj
+        // lock, and since null is a valid initialized value, _claimsIdentityInitialized tracks whether or not it's valid.
+        // _claims is constructed by reading the data from the ClaimsIdentity and is synchronized using Interlockeds
+        // to ensure only one dictionary is published in the face of concurrent access (but if there's a race condition,
+        // multiple dictionaries could be constructed, with only one published for all to see). Simiarly, _propertyBag
+        // is initalized with Interlocked to ensure only a single instance is published in the face of concurrent use.
+        // _claimsIdentityInitialized only ever transitions from false to true, and is volatile to reads/writes are not
+        // reordered relative to the other operations. The rest of the objects are not because the .NET memory model
+        // guarantees object writes are store releases and that reads won't be introduced.
+        private volatile bool _claimsIdentityInitialized;
```
keegan-caruso, line 35: apologies if I'm missing something here, but why does this need to be volatile?

stephentoub: If it's not volatile, then the write to set it to true can effectively move to before the write that initializes the thing being represented, e.g.
```C#
_value = value;
_valueInitialized = true;
```
could be executed as if it were:
```C#
_valueInitialized = true;
_value = value;
```
Then another thread reading _valueInitialized could see it as true before _value was fully set.  This won't happen on x86/x64, which have a strong memory model, but it could happen on ARM which has a weaker memory model.

jennyf19 (approved):

src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
```diff
@@ -67,27 +88,73 @@ public ClaimsIdentity ClaimsIdentity
...
+            }
+        }
 
-            return null;
+        /// <summary>Gets the object to use in <see cref="ClaimsIdentity"/> for double-checked locking.</summary>
+        private object ClaimsIdentitySyncObj
+        {
+            get
+            {
+                object syncObj = _claimsIdentitySyncObj;
```
brentschmaltz, line 147: Is this so the lock object is created in a lazy manner?

stephentoub: Yes

src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
```diff
@@ -67,27 +88,73 @@ public ClaimsIdentity ClaimsIdentity
...
         }
 
         /// <summary>
-        /// This call is for JWTs, SamlTokenHandler will set the ClaimsPrincipal.
+        /// Gets or sets the <see cref="_claimsIdentity"/> without synchronization. All accesses must either
+        /// be protected or used when the caller knows access is serialized.
         /// </summary>
-        /// <returns></returns>
-        private ClaimsIdentity CreateClaimsIdentity()
+        internal ClaimsIdentity ClaimsIdentityNoLocking
```
brentschmaltz, line 115: Curious as to the thinking behind this property.

src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
```diff
@@ -56,7 +70,14 @@ internal TokenValidationResult(SecurityToken securityToken, TokenHandler tokenHa
                 if (!_hasIsValidOrExceptionBeenRead)
                     LogHelper.LogWarning(LogMessages.IDX10109);
 
-                return _claims.Value;
+                IDictionary<string, object> claims = _claims;
+                if (claims is null && ClaimsIdentity is { } ci)
```
brentschmaltz: I don't understand why we need the assignment to local variable claims or the need to call ClaimsIdentity.

src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
stephentoub: ValidateJWEAsync creates a TokenValidationResult and explicitly sets the ClaimsIdentity.  There's no other consumer at that point, and it shouldn't have to pay for the allocation of the lock object or the synchronization on that object in order to set the object.  That same code also accesses the ClaimsIdentity of another instance, which it again knows no one else is currently using, and so it similarly needn't have to lock or allocate.

If that code can go away, ideally because rather than creating a new TokenValidationResult it instead just repopulates the existing one, then you might remove this as well and just inline its contents into ClaimsIdentity.  But as it stands, there are multiple consumers that want the same get/set operations.

src/Microsoft.IdentityModel.Tokens/TokenValidationResult.cs
stephentoub: > or the need to call ClaimsIdentity

It calls ClaimsIdentity to get the ClaimsIdentity instance off of which it accesses Claims below.

> why we need the assignment to local variable claims

This isn't strictly necessary.  It's a holdover from when I had _claims as volatile and was minimizing the number of times I touched it on the fast path.  We can remove the local if you'd prefer.

brentschmaltz (approved):

PR #2181, opened 2023-07-27
Remove object[] allocation from InMemoryCryptoProviderCache.GetCacheKeyPrivate

stephentoub:
There's a params object[] being allocated. On .NET 6+, using an interpolated string will avoid that. On < .NET 6, it'll be exactly what it was before (InvariantCulture isn't needed as nothing here is affected by culture).


jennyf19 (approved):
reviewed as a team. approved

PR #2185, opened 2023-07-27
Better thread safety for JsonClaimSet Claims and JsonWebToken Audiences

### LINKED ISSUE #2186: [Bug] Better thread safety for JsonClaimSet Claims and JsonWebToken Audiences

keegan-caruso:
#2186

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet45.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet45.cs
@@ -21,6 +21,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     internal class JsonClaimSet45
     {
         IList<Claim> _claims;
+        private readonly object _claimsLock = new object();
 
         internal JsonClaimSet45()
         {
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -650,28 +651,35 @@ public IEnumerable<string> Audiences
...
-                    }
-#else
-                    if (Payload.TryGetValue(JwtRegisteredClaimNames.Aud, out JsonElement audiences))
+                    lock (_audiencesLock)
                     {
-                        if (audiences.ValueKind == JsonValueKind.String)
-                            _audiences = new List<string> { audiences.GetString() };
-
-                        if (audiences.ValueKind == JsonValueKind.Array)
+                        if (_audiences == null)
```
jennyf19, line 656: do you need to check again for null? see line 652

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet45.cs
```diff
@@ -122,35 +123,46 @@ internal IList<Claim> CreateClaims(string issuer)
...
-                    _claims.Add(new Claim(entry.Key, claimValue, ClaimValueTypes.String, issuer, issuer));
-                    continue;
-                }
-
-                var jtoken = entry.Value;
-                if (jtoken != null)
-                {
-                    AddClaimsFromJToken(_claims, entry.Key, jtoken, issuer);
-                    continue;
+                    if (_claims == null)
```
jennyf19, line 130: two null checks? Are they needed?

jennyf19 (approved):

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet45.cs
keegan-caruso: yes, this is a standard double checked lock

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
keegan-caruso: see other comment, this is a standard double checked lock pattern

brentschmaltz: Yes it is necessary, only new up the list if its not null.

brentschmaltz (approved):

PR #2188, opened 2023-07-28
Guard logging calls to avoid unnecessary work at call site (#2164)

brentschmaltz:
This is Stephen's PR originally committed into dev7x.
Many LogHelper.Log calls are doing work at the call site, such as allocating params arrays, boxing structs, formatting strings, and so on, even when the data will be thrown away because logging (or logging for that verbosity level) isn't enabled.  This adds an IsEnabled method to LogHelper and uses it at any call site where there's such work to be avoided.

### DIFF
--- a/src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
+++ b/src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics.Tracing;
 using System.IO;
+using Microsoft.IdentityModel.Abstractions;
 
 namespace Microsoft.IdentityModel.Logging
 {
@@ -31,7 +32,7 @@ namespace Microsoft.IdentityModel.Logging
                 _streamWriter = new StreamWriter(fileStream);
                 _streamWriter.AutoFlush = true;
             }
-            catch (Exception ex)
+            catch (Exception ex) when (LogHelper.IsEnabled(EventLogLevel.Error))
             {
                 LogHelper.LogExceptionMessage(new InvalidOperationException(LogMessages.MIML10001, ex));
                 throw;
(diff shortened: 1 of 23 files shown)


keegan-caruso (approved):

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
```diff
@@ -31,7 +32,7 @@ public TextWriterEventListener()
                 _streamWriter = new StreamWriter(fileStream);
                 _streamWriter.AutoFlush = true;
             }
-            catch (Exception ex)
+            catch (Exception ex) when (LogHelper.IsEnabled(EventLogLevel.Error))
```
jennyf19, line 35: @stephentoub will this exception still be thrown even with logging not enabled?

stephentoub: Yes. If logging is disabled, that when filter will make it behave as if the catch block didn't exist at all, and the original exception will propagate. It doesn't change the observed behavior and makes it cheaper when logging isn't enabled as we're then avoiding catching the exception only to immediately rethrow it.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -110,27 +111,41 @@ public virtual KeyInfo ReadKeyInfo(XmlReader reader)
...
+                            if (LogHelper.IsEnabled(EventLogLevel.Warning))
+                            {
+                                LogHelper.LogWarning(LogMessages.IDX30300, reader.ReadOuterXml());
                             }
                             else
                             {
-                                // Skip the element since it is not an <RSAKeyValue>
-                                LogHelper.LogWarning(LogMessages.IDX30300, reader.ReadOuterXml());
+                                reader.Skip();
                             }
```
henrik-me, line 132: looks like changed behavior. @keegan-caruso @brentschmaltz @jennyf19 assuming no tests are failing, however would be good to assess this change for any potential impact (perhaps not relevant as the reader.ReadEndElement does this implicitly already.

Same comment below.

PR #2190, opened 2023-07-28
Fix replacement loop in Base64UrlEncoder.UnsafeDecode

stephentoub:
It was doing unnecessary work by starting the search from the beginning each time.


eerhardt (approved):
Thanks!

jennyf19 (approved):

PR #2194, opened 2023-07-28
Fix casing of Properties directory in updateAssemblyInfo.ps1 script

brentschmaltz:


brentschmaltz:
cherry-picked from dev.

PR #2203, opened 2023-08-03
Adding nullable into extensions dir

FuPingFranco:
Added Nullables into extensions dir.

### DIFF
--- a/src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSecurityKey.cs
+++ b/src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSecurityKey.cs
@@ -17,7 +17,7 @@ namespace Microsoft.IdentityModel.KeyVaultExtensions
     public class KeyVaultSecurityKey : SecurityKey
     {
         private int? _keySize;
-        private string _keyId;
+        private string? _keyId;
 
         /// <summary>
         /// The authentication callback delegate which is to be implemented by the client code.
@@ -58,14 +58,14 @@ namespace Microsoft.IdentityModel.KeyVaultExtensions
         /// <summary>
         /// The authentication callback delegate that retrieves an access token for the KeyVault.
         /// </summary>
-        public AuthenticationCallback Callback { get; protected set; }
+        public AuthenticationCallback? Callback { get; protected set; }
 
         /// <summary>
         /// The uniform resource identifier of the security key.
         /// </summary>
         public override string KeyId
         {
-            get => _keyId;
+            get => _keyId!;
             set
             {
                 if (string.IsNullOrEmpty(value))
@@ -90,7 +90,7 @@ namespace Microsoft.IdentityModel.KeyVaultExtensions
                 if (!_keySize.HasValue)
                     Initialize();
 
-                return _keySize.Value;
+                return _keySize!.Value;
             }
         }
 
@@ -99,7 +99,7 @@ namespace Microsoft.IdentityModel.KeyVaultExtensions
         /// </summary>
         private void Initialize()
         {
-            using (var client = new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(Callback)))
+            using (var client = new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(Callback!)))
             {
                 var bundle = client.GetKeyAsync(_keyId, CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
                 _keySize = new BitArray(bundle.Key.N).Length;
--- a/src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSignatureProvider.cs
@@ -41,11 +41,11 @@ namespace Microsoft.IdentityModel.KeyVaultExtensions
         /// <param name="algorithm">The signature algorithm to apply.</param>
         /// <param name="willCreateSignatures">Whether this <see cref="KeyVaultSignatureProvider"/> is required to create signatures then set this to true.</param>
         /// <param name="client">A mock <see cref="IKeyVaultClient"/> used for testing purposes.</param>
-        internal KeyVaultSignatureProvider(SecurityKey key, string algorithm, bool willCreateSignatures, IKeyVaultClient client)
+        internal KeyVaultSignatureProvider(SecurityKey key, string algorithm, bool willCreateSignatures, IKeyVaultClient? client)
             : base(key, algorithm)
         {
             _key = key as KeyVaultSecurityKey ?? throw LogHelper.LogArgumentNullException(nameof(key));
-            _client = client ?? new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(_key.Callback));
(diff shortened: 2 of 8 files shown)


src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSignatureProvider.cs
```diff
@@ -41,11 +41,11 @@ public KeyVaultSignatureProvider(SecurityKey key, string algorithm, bool willCre
         /// <param name="algorithm">The signature algorithm to apply.</param>
         /// <param name="willCreateSignatures">Whether this <see cref="KeyVaultSignatureProvider"/> is required to create signatures then set this to true.</param>
         /// <param name="client">A mock <see cref="IKeyVaultClient"/> used for testing purposes.</param>
-        internal KeyVaultSignatureProvider(SecurityKey key, string algorithm, bool willCreateSignatures, IKeyVaultClient client)
+        internal KeyVaultSignatureProvider(SecurityKey key, string algorithm, bool willCreateSignatures, IKeyVaultClient? client)
             : base(key, algorithm)
         {
             _key = key as KeyVaultSecurityKey ?? throw LogHelper.LogArgumentNullException(nameof(key));
-            _client = client ?? new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(_key.Callback));
+            _client = client ?? new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(_key.Callback!));
```
MZOLN, line 48: This seems to me like it could result in a NRE?

Could we handle that case? I am not sure what would be appropriate - does a KeyVaultClient without any callback make sense? We should not pass the callback if it is optional.

In case it is required I`d think that we should still throw ArgumentNullException on the Callback property then, because NRE are hard to debug with a misleading callstack. ANE would be a clear pointer to where the issue lay.

FuPingFranco: Great point. In this case when we try to use the Callback from _key, callback is set on the the src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSecurityKey.cs  where we check it is null at the constructor.
        public KeyVaultSecurityKey(string keyIdentifier, AuthenticationCallback callback)
        {
            Callback = callback ?? throw LogHelper.LogArgumentNullException(nameof(callback));
            KeyId = keyIdentifier;
        }

I am not sure if this is enough. I do also like your suggestion on checking callback in here as well.

jennyf19: comment says this is a mock for test purposes and the method is internal.

src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSecurityKey.cs
```diff
@@ -90,7 +90,7 @@ public override int KeySize
                 if (!_keySize.HasValue)
                     Initialize();
 
-                return _keySize.Value;
+                return _keySize!.Value;
```
jennyf19, line 93: are we sure this will never be null?

FuPingFranco: It's more along the lines of it can be null based on what I saw on the Key Id:

public override string KeyId
        {
            get => _keyId!;
            set
            {
                if (string.IsNullOrEmpty(value))
                    throw LogHelper.LogArgumentNullException(nameof(value));
                else if (StringComparer.Ordinal.Equals(_keyId, value))
                    return;

                _keyId = value;

                // Reset the properties so they can be retrieved from Azure KeyVault the next time they are accessed.
                _keySize = null;
            }

jennyf19: maybe something like this would work:
` return _keySize != null ? _keySize.Value : 0;`

src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSignatureProvider.cs
MZOLN: I understand, still we could make someone`s life easier.

src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSecurityKey.cs
MZOLN: Yes I think that would work. On the other hand is there really a need for _keySize to be nullable in the first place if we expose it through non nullable property?

Looking at the code it`s hard for me to see the reason. It would have been better to handle the nullable on the assignment rather than here.

jmprieur:
@FuPingFranco 
The work should be done in the src only, in the dependency order (from the top to the bottom)

image

src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSecurityKey.cs
jmprieur, line 93: Once initialize is used, keySize would not be null? unless getKeyAsync throws

jmprieur (approved):
LGTM
Thanks @FuPingFranco

FuPingFranco:
> @FuPingFranco The work should be done in the src only, in the dependency order (from the top to the bottom)
> 
> image

Yes, this PR was intended for the logging extensions as depicted on the code map however, I did some experimenting and didn't see any major impact on this PR by adding the remaining dll's in the extensions dir.

jmprieur:
That's good news, @FuPingFranco

Maybe we want to have a list of checkboxes (so that we know what is done, and what is coming next)?
- [x] Abstractions
- [x] Logging
- [x] LoggingExtensions
- [ ] Tokens

etc ...
?

src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSecurityKey.cs
FuPingFranco: I thought on this approach of setting a default KeySize but didn't want to define a default value as it might change its behavior somewhere down the call stack.

FuPingFranco: That is my understanding as well.

FuPingFranco:
> That's good news, @FuPingFranco
> 
> Maybe we want to have a list of checkboxes (so that we know what is done, and what is coming next)?
> 
> * [x]  Abstractions
> * [x]  Logging
> * [x]  LoggingExtensions
> * [ ]  Tokens
> 
> etc ... ?

I agree, I will create one on ADO to keep track of this. Thanks Jean-Marc!

keegan-caruso (approved):

jennyf19 (approved):

src/Microsoft.IdentityModel.KeyVaultExtensions/KeyVaultSignatureProvider.cs
FuPingFranco: Given that this is not used other than for test purposes. I am resolving this as it is not high priority on overall work. Thanks for the feedback. It will be definitely good to keep it in mind for my next few PR's as a complete this task.

Issue #1293, opened 2019-11-18
RSA-OAEP-256 not in SupportedAlgorithms.IsSupportedRsaAlgorithm()


Cristallix:
I'm trying to decrypt a JWE + JWS token and here is the first part of the token :

`eyJ6aXAiOiJERUYiLCJlbmMiOiJBMjU2Q0JDLUhTNTEyIiwiYWxnIjoiUlNBLU9BRVAtMjU2In0`

which after a base64ToString is giving : 

`{"zip":"DEF","enc":"A256CBC-HS512","alg":"RSA-OAEP-256"}`

The problem is that when I'm trying to decrypt the token it goes down to the method _SupportedAlgorithms.IsSupportedRsaAlgorithm()_ but RSA-OAEP-256 isn't listed and I can't decrypt my token. Do you plan to support it any time soon or I'm missing something ?

Thanks !

GeoK:
Hi @Cristallix - RSA-OAEP-256 is not currently supported.
I don't see a reason why we shouldn't support it in net461 and netstandard targets.

The team will triage the issue soon, but for now I'm marking it as an _enhancement proposal_.

shjin404:
Yes, we want to use this as well given SHA1 is now discouraged and people are recommended to use SHA2 family. 

@brentschmaltz Minor, but there is a bug in the exception message as well that it shows 'System.String' as Algorithm. It should say No support for Algorithm: 'RSA-OAEP-256', ...

```
Microsoft.IdentityModel.Tokens.SecurityTokenEncryptionFailedException
  HResult=0x80131500
  Message=IDX10615: Encryption failed. No support for: Algorithm: 'System.String', SecurityKey: 'Microsoft.IdentityModel.Tokens.RsaSecurityKey'.
  Source=Microsoft.IdentityModel.JsonWebTokens
  StackTrace:
   at Microsoft.IdentityModel.JsonWebTokens.JsonWebTokenHandler.EncryptTokenPrivate(String innerJwt, EncryptingCredentials encryptingCredentials, String compressionAlgorithm, IDictionary`2 additionalHeaderClaims) in D:\a\1\s\src\Microsoft.IdentityModel.JsonWebTokens\JsonWebTokenHandler.cs:line 856
[...]

Issue #1231, opened 2019-08-07
System.IdentityModel.Tokens.Jwt & Microsoft.IdentityModel.JsonWebTokens 5.5.0 are unsigned


williamBurgeson:
I have dropped System.IdentityModel.Tokens.Jwt into a project I am working on, and the latest default version was 5.5.0. When I pushed to our company's main repo (Git on bitbucket) I was told that the build failed with and error, which included the following info:-
...
Installing Microsoft.IdentityModel.JsonWebTokens 5.5.0.
[12:04:35]  Installing System.IdentityModel.Tokens.Jwt 5.5.0.
[12:04:35]C:\BuildAgent\work\...\[proj path].csproj : error NU3004: Package 'Microsoft.IdentityModel.JsonWebTokens 5.5.0' from source 'https://api.nuget.org/v3/index.json': This repository indicated that all its packages are repository signed; however, this package is unsigned.
[12:04:35]C:\BuildAgent\work\...\[proj path].csproj : error NU3004: Package 'System.IdentityModel.Tokens.Jwt 5.5.0' from source 'https://api.nuget.org/v3/index.json': This repository indicated that all its packages are repository signed; however, this package is unsigned.
...

I looked in another company project, which referenced IdentityServer4.AspNetIdentity version 2.5.0, which didn't have this problem. I could see the following chain of dependency:
IdentityServer4.AspNetIdentity@2.5.0 ->
IdentityServer4@2.5.0 ->
System.IdentityModel.Tokens.Jwt@5.3.0 ->
Microsoft.IdentityModel.JsonWebTokens@5.3.0

I replaced the nuget reference to System.IdentityModel.Tokens.Jwt to version 5.3.0 and this resolved the problem. Note that this also reverted Microsoft.IdentityModel.JsonWebTokens to 5.3.0.

This proves to me that the 5.5.0 version of these components do not have a valid signature if the git repo is set to declare that all nuget packages are signed.

Note that some related packages, Microsoft.IdentityModel.Logging and Microsoft.IdentityModel.Tokens appear to be OK at 5.5.0 without any error occurring.

<img width="398" alt="RangeBoard Nuget Dependencies" src="https://user-images.githubusercontent.com/1300727/62644484-1ef71480-b942-11e9-89f1-be915f6a8227.PNG">

[...]

PR #2207, opened 2023-08-07
Add support for RSA-OAEP 256 (#1293)

karoberts:
This PR adds support for RSA-OAEP-256, as mentioned in RFC 7518 (https://datatracker.ietf.org/doc/html/rfc7518#section-4.1) and fixes Issue 1293

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
+++ b/src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
@@ -31,6 +31,7 @@ namespace Microsoft.IdentityModel.Tokens
         public const string Aes256KW = "A256KW";
         public const string RsaPKCS1 = "RSA1_5";
         public const string RsaOAEP = "RSA-OAEP";
+        public const string RsaOAEP256 = "RSA-OAEP-256";
 
         // See: https://www.w3.org/TR/xmlenc-core1/#sec-Exclusive-Canonicalization
         public const string ExclusiveC14n = "http://www.w3.org/2001/10/xml-exc-c14n#";
--- a/test/Microsoft.IdentityModel.KeyVaultExtensions.Tests/KeyVaultKeyWrapProviderTests.cs
+++ b/test/Microsoft.IdentityModel.KeyVaultExtensions.Tests/KeyVaultKeyWrapProviderTests.cs
@@ -59,6 +59,12 @@ namespace Microsoft.IdentityModel.KeyVaultExtensions.Tests
                     ExpectedException = ExpectedException.NoExceptionExpected,
                     TestId = nameof(SecurityAlgorithms.RsaOAEP),
                 },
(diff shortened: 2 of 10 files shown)


karoberts:

jennyf19:
@brentschmaltz can you review when you have time? #1231

src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
```diff
@@ -31,6 +31,9 @@ public static class SecurityAlgorithms
         public const string Aes256KW = "A256KW";
         public const string RsaPKCS1 = "RSA1_5";
         public const string RsaOAEP = "RSA-OAEP";
+#if NET461 || NET462 || NET472 || NETSTANDARD2_0 || NET6_0_OR_GREATER
```
trukhanov: Is condition needed? Maybe just add this constant with no conditions.

test/Microsoft.IdentityModel.KeyVaultExtensions.Tests/KeyVaultKeyWrapProviderTests.cs
```diff
@@ -59,6 +59,14 @@ public static TheoryData<KeyWrapProviderTheoryData> DisposeProviderTheoryData
                     ExpectedException = ExpectedException.NoExceptionExpected,
                     TestId = nameof(SecurityAlgorithms.RsaOAEP),
                 },
+#if NET461 || NET462 || NET472 || NETSTANDARD2_0 || NET6_0_OR_GREATER
```
trukhanov:
```suggestion
#if NET461_OR_GREATER || NETSTANDARD2_0 || NET6_0_OR_GREATER
```
Can simplify.

src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs
karoberts: Yes, thanks for pointing that out.  I think that the removal of 4.5.2 support happened just after I put in this PR.  I'll remove this since it should always be included now.

karoberts:
I removed all the ifdefs since they were effectively "> net45", and no longer needed

brentschmaltz:
@karoberts sorry to get to this so late, but we would like to take this.
We can build on your work OR you can rebase on top of dev.

karoberts:
> @karoberts sorry to get to this so late, but we would like to take this. We can build on your work OR you can rebase on top of dev.

Sounds great!  I have rebased on top of latest dev.

brentschmaltz:
@karoberts Awesome, we are trying to get this into IdentityModel 8 that is coordinated with .net 9.

brentschmaltz (approved):

brentschmaltz:
@karoberts Thanks.

Issue #2148, opened 2023-07-19
[Bug] Investigate and correctly fix test issues in #2107


jennyf19:
PR #2107 brought in test failures in some classes, which were fixed in #2147 as a workaround, but a more permanent fix needs to be found. 
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2147

This was only impacting tests.

brentschmaltz:
was fixed in 7.x branch.

PR #2208, opened 2023-08-07
JsonWebKey and JsonWebKeySet using System.Text.Json for reading and writing

brentschmaltz:
Reading and Writing JsonWebKey and JsonWebKeySet types now use System.Text.Json.Utf8JsonReaders/Writers for serialization.
JsonWebKey6x and JsonWebKeySet6x were copied from the 6x branch and Deserialization and Roundtrips were compared against what was generated using Newtonsoft.
Differences were captured and will be used to created release notes.

JsonWebKeySerializer and JsonWebKeySetSerializer use JsonSerializerPrimitives as will other types that need serialization.
JsonSerializationPrimitives will provide a robust error message with the type being serialized, the property in play and the reason for the error.

DataSets.cs was modified a considerable amount to make it easier to navigate.
JsonWebKeyTests were modified to use the ctor that takes a TheoryData object, rather than multiple parameters.
JsonSerializerPrimitivesTests was written to understand how failures would be handled, error messages and the difference between Newtonsoft and System.Text.Json.

Microsoft.IdentityModel.Tokens has two additional references to Newtonsoft, those will be removed with modifications that will update OIDC to use System.Text.Json.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Json/EncodedJsonWebKeyParameterNames.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Json/EncodedJsonWebKeyParameterNames.cs
@@ -1,0 +1,271 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Text.Encodings.Web;
+using System.Text.Json;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// JsonEncodedText used with utf8Writer when writing property names for JsonWebKey.
+    /// Common names are initialized on startup, lazy for others.
+    /// </summary>
+    internal static class EncodedJsonWebKeyParameterNames
+    {
+        internal static bool _algSet;
+        internal static JsonEncodedText _alg;
+        internal static bool _crvSet;
+        internal static JsonEncodedText _crv;
+        internal static bool _dSet;
+        internal static JsonEncodedText _d;
+        internal static bool _dpSet;
+        internal static JsonEncodedText _dp;
+        internal static bool _dqSet;
+        internal static JsonEncodedText _dq;
+        internal static bool _kSet;
+        internal static JsonEncodedText _k;
+        internal static bool _keyopsSet;
+        internal static JsonEncodedText _keyOps;
+        internal static bool _othSet;
+        internal static JsonEncodedText _oth;
+        internal static bool _pSet;
+        internal static JsonEncodedText _p;
+        internal static bool _qSet;
+        internal static JsonEncodedText _q;
+        internal static bool _qiSet;
+        internal static JsonEncodedText _qi;
+        internal static bool _x5tS256Set;
+        internal static JsonEncodedText _x5tS256;
+        internal static bool _x5uSet;
+        internal static JsonEncodedText _x5u;
+        internal static bool _xSet;
+        internal static JsonEncodedText _x;
+        internal static bool _ySet;
+        internal static JsonEncodedText _y;
+
+        public static JsonEncodedText Alg
+        {
+            get
+            {
+                if (!_algSet)
+                {
+                    _alg = JsonEncodedText.Encode(JsonWebKeyParameterNames.Alg, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+                    _algSet = true;
+                }
+
+                return _alg;
+            }
+        }
+
+        public static JsonEncodedText Crv
+        {
+            get
+            {
+                if (!_crvSet)
+                {
+                    _crv = JsonEncodedText.Encode(JsonWebKeyParameterNames.Crv, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+                    _crvSet = true;
+                }
+
+                return _crv;
+            }
+        }
+
+        public static JsonEncodedText D
+        {
+            get
+            {
+                if (!_dSet)
+                {
+                    _d = JsonEncodedText.Encode(JsonWebKeyParameterNames.D, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+                    _dSet = true;
+                }
+
+                return _d;
+            }
+        }
+
+        public static JsonEncodedText DP
+        {
+            get
+            {
+                if (!_dpSet)
+                {
+                    _dp = JsonEncodedText.Encode(JsonWebKeyParameterNames.DP, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+                    _dpSet = true;
+                }
+
+                return _dp;
+            }
+        }
+
+        public static JsonEncodedText DQ
+        {
+            get
+            {
+                if (!_dqSet)
+                {
+                    _dq = JsonEncodedText.Encode(JsonWebKeyParameterNames.DQ, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+                    _dqSet = true;
+                }
+
+                return _dq;
+            }
+        }
+
+        public static readonly JsonEncodedText E = JsonEncodedText.Encode(JsonWebKeyParameterNames.E, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+
+        public static JsonEncodedText K
+        {
+            get
+            {
+                if (!_kSet)
+                {
+                    _k = JsonEncodedText.Encode(JsonWebKeyParameterNames.K, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+                    _kSet = true;
+                }
+
+                return _k;
+            }
+        }
+
+        public static JsonEncodedText KeyOps
+        {
+            get
+            {
+                if (!_keyopsSet)
+                {
+                    _keyOps = JsonEncodedText.Encode(JsonWebKeyParameterNames.KeyOps, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+                    _keyopsSet = true;
+                }
+
+                return _keyOps;
+            }
+        }
+
+        public static readonly JsonEncodedText Keys = JsonEncodedText.Encode(JsonWebKeyParameterNames.Keys, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+
+        public static readonly JsonEncodedText Kid = JsonEncodedText.Encode(JsonWebKeyParameterNames.Kid, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+
+        public static readonly JsonEncodedText Kty = JsonEncodedText.Encode(JsonWebKeyParameterNames.Kty, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+
+        public static readonly JsonEncodedText N = JsonEncodedText.Encode(JsonWebKeyParameterNames.N, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+
+        public static JsonEncodedText Oth
+        {
+            get
+            {
+                if (!_othSet)
+                {
+                    _oth = JsonEncodedText.Encode(JsonWebKeyParameterNames.Oth, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+                    _othSet = true;
+                }
+
+                return _oth;
+            }
+        }
+        
+        public static JsonEncodedText P
+        {
+            get
+            {
+                if (!_pSet)
+                {
+                    _p = JsonEncodedText.Encode(JsonWebKeyParameterNames.P, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+                    _pSet = true;
+                }
+
+                return _p;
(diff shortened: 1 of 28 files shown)


src/Microsoft.IdentityModel.Tokens/Json/EncodedJsonWebKeyParameterNames.cs
```diff
@@ -0,0 +1,271 @@
...
+        internal static bool _xSet;
+        internal static JsonEncodedText _x;
+        internal static bool _ySet;
+        internal static JsonEncodedText _y;
+
+        public static JsonEncodedText Alg
+        {
+            get
+            {
+                if (!_algSet)
```
keegan-caruso, line 50: I don't understand, why the lazy initialization? Why not just set all of these up directly?

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -60,194 +61,267 @@ internal static Exception CreateJsonReaderExceptionInvalidType(ref Utf8JsonReade
...
-                bool retVal = reader.GetBoolean();
-                if (advanceReader)
-                    reader.Read();
-
-                return retVal;
+                return reader.Read();
             }
+            catch (JsonException ex)
+            {
+                throw new JsonException(ex.Message, ex);
```
keegan-caruso, line 132: Do we need the try catch at all in this case?

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -60,194 +61,267 @@ internal static Exception CreateJsonReaderExceptionInvalidType(ref Utf8JsonReade
...
+            Debug.Assert(false, "expected to read a number, but none of the Utf8JsonReader.TryGet... methods returned true.");
+
+            return ReadJsonElement(ref reader);
+        }
+
+        internal static void WriteAdditionalData(ref Utf8JsonWriter writer, IDictionary<string, object> additionalData)
+        {
+            if (additionalData.Count > 0)
+            {
+                foreach (string key in additionalData.Keys)
```
keegan-caruso: nit: since you need both the key and value in every iteration it is slightly more efficient to iterate over the KVPs directly and use the key and value off of that

src/Microsoft.IdentityModel.Tokens/Json/EncodedJsonWebKeyParameterNames.cs
```diff
@@ -0,0 +1,271 @@
...
+                {
+                    _alg = JsonEncodedText.Encode(JsonWebKeyParameterNames.Alg, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
```
jennyf19, line 52: >JsonEncodedText.Encode( [](http://example.com/codeflow?start=27&length=23)

nit: I think this is clear, but could remove the duplicate code and have a method `GetEncodedText` which takes the jsonWebKeyParamName and returns the JsonEncodedText.

src/Microsoft.IdentityModel.Tokens/JsonWebKeyParameterNames.cs
```diff
@@ -24,16 +27,41 @@ public static class JsonWebKeyParameterNames
...
         public const string X5c = "x5c";
         public const string X5t = "x5t";
         public const string X5tS256 = "x5t#S256";
         public const string X5u = "x5u";
-        public const string X = "x";
         public const string Y = "y";
-#pragma warning restore 1591
     }
+
+    public static class JsonWebKeyParameterUtf8Bytes
```
keegan-caruso: could this be internal?

test/Microsoft.IdentityModel.TestUtils/IdentityComparer.cs
```diff
@@ -33,54 +33,55 @@ namespace Microsoft.IdentityModel.TestUtils
 {
     public class IdentityComparer
     {
+        // Dictionary of types and the validation function.
```
jennyf19, line 36: >Dictio [](http://example.com/codeflow?start=11&length=6)

wondering if this issue: [https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2148](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2148) will be fixed now?

test/Microsoft.IdentityModel.Tokens.Tests/Json/JsonWebKey6x.cs
```diff
@@ -0,0 +1,482 @@
...
+    /// <summary>
+    /// Represents a JSON Web Key as defined in https://datatracker.ietf.org/doc/html/rfc7517.
+    /// This is the original JsonWebKey in the 6x branch.
+    /// Used for ensuring backcompat.
+    /// </summary>
+    [JsonObject]
+    public class JsonWebKey6x : SecurityKey6x
+    {
+        private string _kid;
+        private const string _className = "Microsoft.IdentityModel.Tokens.JsonWebKey6x6x";
```
jennyf19: >6x6x [](http://example.com/codeflow?start=84&length=4)

`6x6x`? or just `6x`

jennyf19 (approved):
:shipit:

keegan-caruso (approved):

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
brentschmaltz: @keegan-caruso Yep, make this fix.
Thanks.

src/Microsoft.IdentityModel.Tokens/JsonWebKeyParameterNames.cs
brentschmaltz: Yes, will change, we may want to make it public later.

test/Microsoft.IdentityModel.Tokens.Tests/Json/JsonWebKey6x.cs
brentschmaltz: Fixed

keegan-caruso (approved):

src/Microsoft.IdentityModel.Tokens/Json/JsonWebKeySetSerializer.cs
```diff
@@ -0,0 +1,140 @@
...
+            using (MemoryStream memoryStream = new MemoryStream())
+            {
+                Utf8JsonWriter writer = null;
+                try
+                {
+                    // writing strings without escaping is as we know this is a utf8 encoding
+                    writer = new Utf8JsonWriter(memoryStream, new JsonWriterOptions { Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping });
+                    Write(ref writer, jsonWebKeySet);
+                    writer.Flush();
+                    return Encoding.UTF8.GetString(memoryStream.ToArray());
```
stephentoub, line 105: This shouldn't be using ToArray. At a minimum it can use GetBuffer() to get the MemoryStream's underlying buffer directly, rather than allocating and copying, e.g.
```C#
return Encoding.UTF8.GetString(memoryStream.GetBuffer(), 0, memoryStream.Length);
```
However, it would also be nice to avoid the MemoryStream and Utf8JsonWriter allocations as well. Could this instead just use JsonSerializer? It handles all the pooling for you.  The task then I think would just be writing a JsonConverter for JsonWebKeySet, which should mostly be copy/paste of this code into its Write method.

src/Microsoft.IdentityModel.Tokens/JsonWebKeyParameterNames.cs
```diff
@@ -24,16 +27,41 @@ public static class JsonWebKeyParameterNames
...
+        public static readonly byte[] P = Encoding.UTF8.GetBytes("p");
+        public static readonly byte[] Q = Encoding.UTF8.GetBytes("q");
+        public static readonly byte[] QI = Encoding.UTF8.GetBytes("qi");
+        public static readonly byte[] Use = Encoding.UTF8.GetBytes("use");
+        public static readonly byte[] X5c = Encoding.UTF8.GetBytes("x5c");
+        public static readonly byte[] X5t = Encoding.UTF8.GetBytes("x5t");
+        public static readonly byte[] X5tS256 = Encoding.UTF8.GetBytes("x5t#S256");
+        public static readonly byte[] X5u = Encoding.UTF8.GetBytes("x5u");
+        public static readonly byte[] X = Encoding.UTF8.GetBytes("x");
+        public static readonly byte[] Y = Encoding.UTF8.GetBytes("y");
```
stephentoub, line 64: These can all use `u8` instead and avoid doing the UTF8 encoding at run-time. In fact, the only usage appears to be as arguments to ValueTextEquals, which has span-based overloads, so these don't even need to be arrays, e.g. replace
```C#
public static readonly byte[] X = Encoding.UTF8.GetBytes("x");
```
with
```C#
public static ReadOnlySpan<byte> X => "x"u8;
```
or do away with this class entirely and just use `"x"u8` at the call site.

src/Microsoft.IdentityModel.Tokens/Json/EncodedJsonWebKeyParameterNames.cs
```diff
@@ -0,0 +1,271 @@
...
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// JsonEncodedText used with utf8Writer when writing property names for JsonWebKey.
+    /// Common names are initialized on startup, lazy for others.
+    /// </summary>
+    internal static class EncodedJsonWebKeyParameterNames
+    {
+        internal static bool _algSet;
```
stephentoub, line 15: These bools should all be volatile

src/Microsoft.IdentityModel.Tokens/Json/EncodedJsonWebKeyParameterNames.cs
```diff
@@ -0,0 +1,271 @@
...
+            }
+        }
+
+        public static readonly JsonEncodedText Keys = JsonEncodedText.Encode(JsonWebKeyParameterNames.Keys, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+
+        public static readonly JsonEncodedText Kid = JsonEncodedText.Encode(JsonWebKeyParameterNames.Kid, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+
+        public static readonly JsonEncodedText Kty = JsonEncodedText.Encode(JsonWebKeyParameterNames.Kty, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+
+        public static readonly JsonEncodedText N = JsonEncodedText.Encode(JsonWebKeyParameterNames.N, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
```
stephentoub, line 152: What's the rhyme/reason for why some of these are lazy and some aren't? Are the prompt ones always used and the lazy ones rarely used?

src/Microsoft.IdentityModel.Tokens/Json/EncodedJsonWebKeyParameterNames.cs
```diff
@@ -0,0 +1,271 @@
...
+            get
+            {
+                if (!_ySet)
+                {
+                    _y = JsonEncodedText.Encode(JsonWebKeyParameterNames.Y, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+                    _ySet = true;
+                }
+
+                return _y;
+            }
```
stephentoub, line 268: You could simplify this file with a helper like:
```C#
private static JsonEncodedText Get(ref bool initialized, ref JsonEncodedText field, string name)
{
    if (!Volatile.Read(ref initialized)
    {
        field = JsonEncodedText.Encode(name, JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
        Volatile.Write(ref initialized, true);
    }
    return field;
}
```
and then your properties all become one-liners, e.g.
```C#
public static JsonEncodedText Y => Get(ref _ySet, ref _y, JsonWebKeyParameterNames.Y);
```
(and then you don't need to make the bool fields volatile as it's then handled by the Get helper).

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -60,194 +61,267 @@ internal static Exception CreateJsonReaderExceptionInvalidType(ref Utf8JsonReade
...
+            try
             {
-                bool retVal = reader.GetBoolean();
-                if (advanceReader)
-                    reader.Read();
-
-                return retVal;
+                return reader.Read();
             }
+            catch (JsonException ex)
```
stephentoub, line 130: What is the purpose of this catch block?

PR #2210, opened 2023-08-08
Adding nullable to logging project

FuPingFranco:
Adding Nullable to Logging project

### DIFF
--- a/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
+++ b/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
@@ -93,7 +93,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <param name="message">The log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
         [NonEvent]
-        public void WriteAlways(string message, params object[] args)
+        public void WriteAlways(string message, params object[]? args)
         {
             if (IsEnabled())
             {
@@ -124,7 +124,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <param name="message">The log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
         [NonEvent]
-        public void WriteVerbose(string message, params object[] args)
+        public void WriteVerbose(string message, params object[]? args)
         {
             if (IsEnabled() && LogLevel >= EventLevel.Verbose)
             {
@@ -155,7 +155,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <param name="message">The log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
         [NonEvent]
-        public void WriteInformation(string message, params object[] args)
+        public void WriteInformation(string message, params object[]? args)
         {
             if (IsEnabled() && LogLevel >= EventLevel.Informational)
             {
@@ -186,7 +186,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <param name="message">The log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
         [NonEvent]
-        public void WriteWarning(string message, params object[] args)
+        public void WriteWarning(string message, params object[]? args)
         {
             if (args != null)
                 WriteWarning(FormatInvariant(message, args));
@@ -214,7 +214,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <param name="message">The log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
         [NonEvent]
-        public void WriteError(string message, params object[] args)
+        public void WriteError(string message, params object[]? args)
         {
             if (IsEnabled() && LogLevel >= EventLevel.Error)
             {
@@ -245,7 +245,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <param name="message">The log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
         [NonEvent]
-        public void WriteCritical(string message, params object[] args)
+        public void WriteCritical(string message, params object[]? args)
         {
             if (IsEnabled() && LogLevel >= EventLevel.Critical)
             {
@@ -263,9 +263,9 @@ namespace Microsoft.IdentityModel.Logging
         /// <param name="innerException"><see cref="Exception"/></param>
         /// <param name="message">The log message.</param>
         [NonEvent]
-        public void Write(EventLevel level, Exception innerException, string message)
+        public void Write(EventLevel level, Exception? innerException, string message)
         {
-            Write(level, innerException, message, null!);
+            Write(level, innerException, message, null);
         }
 
         /// <summary>
@@ -276,7 +276,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <param name="message">The log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
         [NonEvent]
-        public void Write(EventLevel level, Exception innerException, string message, params object[] args)
+        public void Write(EventLevel level, Exception? innerException, string message, params object[]? args)
         {
             if (innerException != null)
             {
--- a/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
+++ b/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
@@ -265,7 +265,7 @@ namespace Microsoft.IdentityModel.Logging
         [NonEvent]
         public void Write(EventLevel level, Exception innerException, string message)
         {
-            Write(level, innerException, message, null);
+            Write(level, innerException, message, null!);
         }
 
         /// <summary>
@@ -291,7 +291,7 @@ namespace Microsoft.IdentityModel.Logging
             if (!HeaderWritten)
             {
                 // Obtain the current library version dynamically.
-                WriteAlways(string.Format(CultureInfo.InvariantCulture, _versionLogMessage, typeof(IdentityModelEventSource).GetTypeInfo().Assembly.GetName().Version.ToString()));
+                WriteAlways(string.Format(CultureInfo.InvariantCulture, _versionLogMessage, typeof(IdentityModelEventSource).GetTypeInfo().Assembly!.GetName().Version!.ToString()));
                 WriteAlways(string.Format(CultureInfo.InvariantCulture, _dateLogMessage, DateTime.UtcNow));
                 if (ShowPII) 
                     WriteAlways(_piiOnLogMessage);
--- a/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
+++ b/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
@@ -291,7 +291,7 @@ namespace Microsoft.IdentityModel.Logging
             if (!HeaderWritten)
             {
                 // Obtain the current library version dynamically.
-                WriteAlways(string.Format(CultureInfo.InvariantCulture, _versionLogMessage, typeof(IdentityModelEventSource).GetTypeInfo().Assembly!.GetName().Version!.ToString()));
+                WriteAlways(string.Format(CultureInfo.InvariantCulture, _versionLogMessage, typeof(IdentityModelEventSource).GetTypeInfo().Assembly.GetName().Version?.ToString()));
                 WriteAlways(string.Format(CultureInfo.InvariantCulture, _dateLogMessage, DateTime.UtcNow));
                 if (ShowPII) 
                     WriteAlways(_piiOnLogMessage);
--- a/src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
+++ b/src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
@@ -44,7 +44,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <summary>
         /// Get the string that represents the client version.
         /// </summary>
-        public static string ClientVer => typeof(IdentityModelTelemetryUtil).GetTypeInfo().Assembly!.GetName().Version!.ToString();
+        public static string ClientVer => typeof(IdentityModelTelemetryUtil).GetTypeInfo().Assembly.GetName().Version?.ToString() ?? "Version Unavailable";
 
         /// <summary>
         /// Adds a key and its value to the collection of telemetry data.
--- a/src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
+++ b/src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
@@ -44,7 +44,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <summary>
         /// Get the string that represents the client version.
         /// </summary>
-        public static string ClientVer => typeof(IdentityModelTelemetryUtil).GetTypeInfo().Assembly.GetName().Version.ToString();
+        public static string ClientVer => typeof(IdentityModelTelemetryUtil).GetTypeInfo().Assembly!.GetName().Version!.ToString();
 
         /// <summary>
         /// Adds a key and its value to the collection of telemetry data.
--- a/src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
+++ b/src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
@@ -20,7 +20,7 @@ namespace Microsoft.IdentityModel.Logging
         internal static readonly ConcurrentDictionary<string, string> telemetryData = new ConcurrentDictionary<string, string>()
         {
             [skuTelemetry] = ClientSku,
-            [versionTelemetry] = ClientVer
+            [versionTelemetry] = ClientVer ?? String.Empty
         };
 
         /// <summary>
@@ -44,7 +44,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <summary>
         /// Get the string that represents the client version.
         /// </summary>
-        public static string ClientVer => typeof(IdentityModelTelemetryUtil).GetTypeInfo().Assembly.GetName().Version?.ToString() ?? "Version Unavailable";
+        public static string? ClientVer => typeof(IdentityModelTelemetryUtil).GetTypeInfo().Assembly.GetName().Version?.ToString();
 
         /// <summary>
         /// Adds a key and its value to the collection of telemetry data.
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -262,7 +262,7 @@ namespace Microsoft.IdentityModel.Logging
         public static Exception LogExceptionMessage(EventLevel eventLevel, Exception exception)
         {
             if (exception == null)
-                throw new ArgumentNullException(nameof(exception));
+                throw new ArgumentNullException(nameof(exception), "null exception was passed to logger");
 
             if (IdentityModelEventSource.Logger.IsEnabled(eventLevel, EventKeywords.All))
                 IdentityModelEventSource.Logger.Write(eventLevel, exception.InnerException, exception.Message);
@@ -356,7 +356,7 @@ namespace Microsoft.IdentityModel.Logging
             }
 
             if(value is null)
-                throw new ArgumentException(nameof(value));
+                throw new ArgumentException("Newly created Exception is null");
 
             return value;
         }
@@ -449,7 +449,7 @@ namespace Microsoft.IdentityModel.Logging
         private static LogEntry WriteEntry(EventLogLevel eventLogLevel, Exception? innerException, string message, params object[]? args)
         {
             if (string.IsNullOrEmpty(message))
-                throw new ArgumentNullException(nameof(message));
+                throw new ArgumentNullException(nameof(message), "null message was passed to new entry");
 
             if (innerException != null)
             {
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -265,7 +265,7 @@ namespace Microsoft.IdentityModel.Logging
                 return null!;
 
             if (IdentityModelEventSource.Logger.IsEnabled(eventLevel, EventKeywords.All))
-                IdentityModelEventSource.Logger.Write(eventLevel, exception.InnerException!, exception.Message);
+                IdentityModelEventSource.Logger.Write(eventLevel, exception.InnerException, exception.Message);
 
             EventLogLevel eventLogLevel = EventLevelToEventLogLevel(eventLevel);
             if (Logger.IsEnabled(eventLogLevel))
@@ -333,7 +333,7 @@ namespace Microsoft.IdentityModel.Logging
                 message = format;
 
             if (IdentityModelEventSource.Logger.IsEnabled(eventLevel, EventKeywords.All))
-                IdentityModelEventSource.Logger.Write(eventLevel, innerException!, message);
+                IdentityModelEventSource.Logger.Write(eventLevel, innerException, message);
 
             EventLogLevel eventLogLevel = EventLevelToEventLogLevel(eventLevel);
             if (Logger.IsEnabled(eventLogLevel))
@@ -458,7 +458,7 @@ namespace Microsoft.IdentityModel.Logging
             // Prefix header (library version, DateTime, whether PII is ON/OFF) to the first message logged by Wilson.
             if (!_isHeaderWritten)
             {
-                string headerMessage = string.Format(CultureInfo.InvariantCulture!, "Microsoft.IdentityModel Version: {0}. Date {1}. {2}",
+                string headerMessage = string.Format(CultureInfo.InvariantCulture, "Microsoft.IdentityModel Version: {0}. Date {1}. {2}",
                     typeof(IdentityModelEventSource).Assembly!.GetName().Version!.ToString(),
                     DateTime.UtcNow,
                     IdentityModelEventSource.ShowPII ? _piiOnLogMessage : _piiOffLogMessage);
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -459,7 +459,7 @@ namespace Microsoft.IdentityModel.Logging
             if (!_isHeaderWritten)
             {
                 string headerMessage = string.Format(CultureInfo.InvariantCulture, "Microsoft.IdentityModel Version: {0}. Date {1}. {2}",
-                    typeof(IdentityModelEventSource).Assembly!.GetName().Version!.ToString(),
+                    typeof(IdentityModelEventSource).Assembly.GetName().Version?.ToString(),
                     DateTime.UtcNow,
                     IdentityModelEventSource.ShowPII ? _piiOnLogMessage : _piiOffLogMessage);
 
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -227,7 +227,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <param name="innerException">the inner <see cref="Exception"/> to be added to the outer exception.</param>
         /// <param name="format">Format string of the log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
-        public static T LogException<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] T>(EventLevel eventLevel, Exception innerException, string format, params object[] args) where T : Exception
+        public static T LogException<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] T>(EventLevel eventLevel, Exception? innerException, string format, params object[]? args) where T : Exception
         {
             return LogExceptionImpl<T>(eventLevel, null, innerException, format, args);
         }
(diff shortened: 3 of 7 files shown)


src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
```diff
@@ -265,7 +265,7 @@ public void WriteCritical(string message, params object[] args)
         [NonEvent]
         public void Write(EventLevel level, Exception innerException, string message)
         {
-            Write(level, innerException, message, null);
+            Write(level, innerException, message, null!);
```
keegan-caruso: `null!` hints that something isn't right here.

Why did we end up with this solution?

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
```diff
@@ -291,7 +291,7 @@ public void Write(EventLevel level, Exception innerException, string message, pa
             if (!HeaderWritten)
             {
                 // Obtain the current library version dynamically.
-                WriteAlways(string.Format(CultureInfo.InvariantCulture, _versionLogMessage, typeof(IdentityModelEventSource).GetTypeInfo().Assembly.GetName().Version.ToString()));
+                WriteAlways(string.Format(CultureInfo.InvariantCulture, _versionLogMessage, typeof(IdentityModelEventSource).GetTypeInfo().Assembly!.GetName().Version!.ToString()));
```
keegan-caruso: I don't understand, why is Assembly! and Version! required here?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -44,7 +44,7 @@ public static class IdentityModelTelemetryUtil
         /// <summary>
         /// Get the string that represents the client version.
         /// </summary>
-        public static string ClientVer => typeof(IdentityModelTelemetryUtil).GetTypeInfo().Assembly.GetName().Version.ToString();
+        public static string ClientVer => typeof(IdentityModelTelemetryUtil).GetTypeInfo().Assembly!.GetName().Version!.ToString();
```
keegan-caruso: same comment, why is this required?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -262,10 +262,10 @@ public static Exception LogExceptionMessage(Exception exception)
         public static Exception LogExceptionMessage(EventLevel eventLevel, Exception exception)
         {
             if (exception == null)
-                return null;
+                return null!;
```
keegan-caruso: this doesn't seem right, why not change the return type to `Exception?`

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -262,10 +262,10 @@ public static Exception LogExceptionMessage(Exception exception)
...
 
             if (IdentityModelEventSource.Logger.IsEnabled(eventLevel, EventKeywords.All))
-                IdentityModelEventSource.Logger.Write(eventLevel, exception.InnerException, exception.Message);
+                IdentityModelEventSource.Logger.Write(eventLevel, exception.InnerException!, exception.Message);
```
keegan-caruso: how do we know innerexception is not null here?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -333,22 +333,22 @@ public static void LogWarning(string message, params object[] args)
                 message = format;
 
             if (IdentityModelEventSource.Logger.IsEnabled(eventLevel, EventKeywords.All))
-                IdentityModelEventSource.Logger.Write(eventLevel, innerException, message);
+                IdentityModelEventSource.Logger.Write(eventLevel, innerException!, message);
```
keegan-caruso: same question, how do we know innerException is not null?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -333,22 +333,22 @@ public static void LogWarning(string message, params object[] args)
...
 
             EventLogLevel eventLogLevel = EventLevelToEventLogLevel(eventLevel);
             if (Logger.IsEnabled(eventLogLevel))
                 Logger.Log(WriteEntry(eventLogLevel, innerException, message, null));
 
             if (innerException != null) 
                 if (string.IsNullOrEmpty(argumentName))
-                    return (T)Activator.CreateInstance(typeof(T), message, innerException);
+                    return (T)Activator.CreateInstance(typeof(T), message, innerException)!;
```
keegan-caruso: for all of these, Activator can return null though, why are we suppressing that here?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -383,12 +383,12 @@ private static object SanitizeSecurityArtifact(object arg)
...
 
             if (IdentityModelEventSource.LogCompleteSecurityArtifact && arg is ISafeLogSecurityArtifact)
-                return (arg as ISafeLogSecurityArtifact).UnsafeToString();
+                return ((ISafeLogSecurityArtifact)arg).UnsafeToString();
 
             return arg;
         }
 
-        private static string RemovePII(object arg)
+        private static string? RemovePII(object arg)
```
keegan-caruso: this doesn't look like it can return null?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -401,7 +401,7 @@ private static string RemovePII(object arg)
 
         internal static bool IsCustomException(Exception ex)
         {
-            return ex.GetType().FullName.StartsWith("Microsoft.IdentityModel.", StringComparison.Ordinal);
+            return ex.GetType().FullName!.StartsWith("Microsoft.IdentityModel.", StringComparison.Ordinal);
```
keegan-caruso, line 414: why is this needed?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -436,10 +436,10 @@ public static object MarkAsSecurityArtifact(object arg, Func<object, string> cal
         /// <param name="innerException"><see cref="Exception"/></param>
         /// <param name="message">The log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
-        private static LogEntry WriteEntry(EventLogLevel eventLogLevel, Exception innerException, string message, params object[] args)
+        private static LogEntry WriteEntry(EventLogLevel eventLogLevel, Exception? innerException, string message, params object[]? args)
         {
             if (string.IsNullOrEmpty(message))
-                return null;
+                return null!;
```
keegan-caruso: I don't understand

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -458,8 +458,8 @@ private static LogEntry WriteEntry(EventLogLevel eventLogLevel, Exception innerE
             // Prefix header (library version, DateTime, whether PII is ON/OFF) to the first message logged by Wilson.
             if (!_isHeaderWritten)
             {
-                string headerMessage = string.Format(CultureInfo.InvariantCulture, "Microsoft.IdentityModel Version: {0}. Date {1}. {2}",
-                    typeof(IdentityModelEventSource).Assembly.GetName().Version.ToString(),
+                string headerMessage = string.Format(CultureInfo.InvariantCulture!, "Microsoft.IdentityModel Version: {0}. Date {1}. {2}",
```
keegan-caruso: why is this needed?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -458,8 +458,8 @@ private static LogEntry WriteEntry(EventLogLevel eventLogLevel, Exception innerE
...
+                    typeof(IdentityModelEventSource).Assembly!.GetName().Version!.ToString(),
```
keegan-caruso: why are these needed?

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
```diff
@@ -91,7 +91,7 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
 
             for (int i = 0; i < eventData.Payload.Count; i++)
             {
-                _streamWriter.WriteLine(eventData.Payload[i].ToString());
+                _streamWriter.WriteLine(eventData.Payload[i]!.ToString());
```
keegan-caruso, line 96: is this safe?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jmprieur: the change is rather in Logger.Write to accept null

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
jennyf19: I'd think that you would need:
`public void Write(EventLevel level, Exception innerException, string message, params object[]? args)` 
and then 
`public void WriteAlways(string message, params object[]? args)`

which already has a null check for null `args`

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
jennyf19: I'm not sure `!` is needed for either. Do you get warnings without?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: You need to return null here, so adding `null!` is not correct because we are returning null. Only add `!` when you know for sure the value will never be null.

Instead, note that the return type can be null instead:
`public static Exception? LogExceptionMessage(EventLevel eventLevel, Exception exception)`

and bubble it up:
`public static Exception? LogExceptionMessage(Exception exception)`

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: +1, also, I don't think this is needed.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: I don't think so. The args can be null, but I don't think we'll get any nullable warnings with the code as-is.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: I don't think `!` is needed on any of them.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: Correct...why was this added as a possible null return?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: The issue, I think, is this will return null, so adding a null return type `LogEntry?` means things like this 
```csharp
if (Logger.IsEnabled(eventLogLevel))
                Logger.Log(WriteEntry(eventLogLevel, exception.InnerException, exception.Message, null));
``` 
could have null references.

I'm wondering why we return null here anyway if message is empty? We could remove the null check on the message, looks like the only place where it's not passed is here, and we can check for null before:
`Logger.Log(WriteEntry(eventLogLevel, exception.InnerException, exception.Message ?? string.Empty, null));` and pass in an empty string instead. 

Not sure the right thing to do here, but it's a bit tricky.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: +1 should not be needed

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
jennyf19: why do you think this is needed @FuPingFranco ? Are you getting a warning in the IDE?

jennyf19 (changes requested):
left a lot of questions and some suggestions. The nullables do not look to be added in the correct places.

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
FuPingFranco: @keegan-caruso I believe it is safe, null checks for  eventData and Payload are checked on lines 83 and 86 respectively.
@jennyf19  Yes, I get warnings for error code CS8602-Dereference of a possibly null reference. (happens when we set nullables but it doesn't know we already performed null checks previous to using it this way)

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: IDE throws error code CS8602-Dereference of a possibly null reference. Version is not nullable in System.Reflection so it causes noise

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: It was throwing an error before but seems to not be the case anymore. Will remove thanks!

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: IDE throws error code CS8602-Dereference of a possibly null reference. FullName is not nullable in System so it causes noise

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: It used in line 375 as a return value

 if (!IdentityModelEventSource.ShowPII)
                return string.Format(CultureInfo.InvariantCulture, format, args.Select(RemovePII).ToArray()); 

IDE throws error code CS8603-Possible null reference return

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: @keegan-caruso Yes Activator can return null because it's been handled, but IDE throws error code CS8603-Possible null reference return.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: @jmprieur  Is correct, I took off the warning during development to deal with another error and forgot to take it off to add the nullable arg to the logger.Write

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: Same as the next comment

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
FuPingFranco: @jennyf19  Yeap, without it, IDE throws error code CS8602-Dereference of a possibly null reference. Version is not nullable in System.Reflection so it causes noise

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
FuPingFranco: IDE throws error code CS8602-Dereference of a possibly null reference. Version is not nullable in System.Reflection so it causes noise

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
FuPingFranco: Great catch, I muted off this error because I started working on a different as it was very repetitive from lines 306-328. I'll update this.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: I agree with returning Exception? but it leads to error code (CS8597- Thrown value may be null )
in all instances where we have something like "throw LogExceptionMessage(new WsFederationException(LogMessages.IDX22902)"

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: @jennyf19 that is a solution however, an empty string and a Null string are not equal so, I opted to keep the current flow by using !, not the best solution or a solution at all but, given the deadline I opted for keeping the current flow untouched.

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
jennyf19: Can you resolve the comments which are fixed? thank you

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
jennyf19: I had Build on, not Build + IntelliSense, so I didn't see this one. You can actually do `GetTypeInfo().Assembly.GetName().Version?.ToString()));` instead

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
jennyf19: should be `GetTypeInfo().Assembly.GetName().Version?.ToString();` instead

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: should be `return (T?)Activator.CreateInstance(typeof(T), message);`

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: should be `return arg.ToString()!; // the null check is done in FormatInvariant`

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
jennyf19: should be `_streamWriter.WriteLine(eventData!.Payload[i]?.ToString());`

jennyf19:
> left a lot of questions and some suggestions. The nullables do not look to be added in the correct places.

@FuPingFranco I think addressed most of the comments in this [PR](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2215), so please take a look, merge when you want and resolve comments which are addressed. thanks!

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: Made arg nullable

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: Made arg nullable

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
FuPingFranco: Changed implementation, pls review again for correctness. Thanks!

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
FuPingFranco: @jennyf19 added condition instead as per Marcin recommendation.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
jennyf19: I would declare this as null: `public static string? ClientVer => ...` and then earlier handle the null by passing an empty string. We wouldn't want to mess up with the values sent to eSTS. Maybe something like `[versionTelemetry] = ClientVer ?? string.Empy`

FuPingFranco: Makes sense, I updated it to reflect your suggestion. Thanks Jenny!

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: can we do `message = string.Empty;`?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: why not `exception = exception ?? throw new ArgumentNullException(nameof(exception));` for example?

jennyf19:
>         public static string FormatInvariant(string format, params object[] args)

I'm surprised you don't need `string?` here as format can be null. 

---
Refers to: src/Microsoft.IdentityModel.Logging/LogHelper.cs:366 in ed96099. [](commit_id = ed960993b38122ecbd186eed8b764c8ddb9fe26f, deletion_comment = False)

jennyf19:
Left a few comments. still needs a bit of work.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
keegan-caruso: returning null is still incorrect.

I would argue that passing null is a usage error of the API.

It would be a behavioral breaking change, but what about throwing an argument null exception if passed null? We would need to include it in the breaking change list, but it is targeting the 7x branch.

Thoughts @jennyf19  and @FuPingFranco ?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
keegan-caruso: this is still incorrect. We either need to actually check that the object created is not null, or we need to change the return type to T?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
keegan-caruso: this one still needs to be figured out, we can't return null here.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: Is this still an issue @FuPingFranco ?

FuPingFranco: @jennyf19 Yes, this method cannot return null as is right now. For some reason when you help me look into it with the outdated branch, it was returning nulls then as I see it in this [PR](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2215).

@keegan-caruso I like the idea of checking for nulls here more than changing the return type but, not sure what would be appropriate return/throw if it is null. Any thoughts?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
brentschmaltz: @keegan-caruso i agree with you about return null, that seems to be a programming mistake and can lead to a null ref at any time.

How about we add a debug assert if the exception passed in is null and return new ArgumentNullException("null exception was passed to logger");

@jennyf19  not sure where you are suggesting we have this code.
why not exception = exception ?? throw new ArgumentNullException(nameof(exception)); for example?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: @jennyf19 After synching with Keegan, we changed the implementation slightly which took care of this. Resolving this comment Thank you both!

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: Will be retuning new ArgumentNullException and adding it to the breaking changes. Resolving this Thread. Thank you both!

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: I agree, I have made the change to thow an ArgumentNullException post null check and will include it to the coming breaking changes. Resolving this thread, thanks everyone!

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -383,7 +393,7 @@ private static object SanitizeSecurityArtifact(object arg)
                 return "null";
 
             if (IdentityModelEventSource.LogCompleteSecurityArtifact && arg is ISafeLogSecurityArtifact)
-                return (arg as ISafeLogSecurityArtifact).UnsafeToString();
+                return ((ISafeLogSecurityArtifact)arg).UnsafeToString();
```
jmprieur: why this change?

jmprieur (approved):
LGTM
I just left a question

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: @jmprieur Using 'as' leads to error CS8602- Dereference of a possibly null reference after we enable nullables. This is because when using 'as' it will never return an InvalidCastExeception if it happens, instead it will ansign and return null which is not right in this case. An alternative is to declare arg as a nullable which is still not correct as we never return null from this method as shown in line 393 where we return a string named "null" if the argument is ever null.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -339,16 +339,26 @@ public static void LogWarning(string message, params object[] args)
...
                 if (string.IsNullOrEmpty(argumentName))
-                    return (T)Activator.CreateInstance(typeof(T), message);
+                    value = Activator.CreateInstance(typeof(T), message) as T;
                 else
-                    return (T)Activator.CreateInstance(typeof(T), argumentName, message);
+                    value = Activator.CreateInstance(typeof(T), argumentName, message) as T;
+            }
+
+            if(value is null)
+                throw new ArgumentException("Newly created Exception is null");
```
keegan-caruso, line 359: we need a better error message, this isn't actionable.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -262,7 +262,7 @@ public static Exception LogExceptionMessage(Exception exception)
         public static Exception LogExceptionMessage(EventLevel eventLevel, Exception exception)
         {
             if (exception == null)
-                return null;
+                throw new ArgumentNullException(nameof(exception), "null exception was passed to logger");
```
keegan-caruso: does the message add any value in this case? or can we just pass the param name?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
keegan-caruso: suggestion: use pattern matching to clean this up

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: @keegan-caruso We could just pass the param name, but the I think the message would help more quickly identify why and where this ArgumentNullException happened.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: @keegan-caruso I can elaborate more on this error message however, because 'value' is created local to this method, there isn't a specific param or inner exception I can include as part of the ArgumentException. Do you have any suggestions to improve on this?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
keegan-caruso: "why" is irrelevant in this case. It is a ANE, it should be self-describing.

"where" should be clear from a stack trace. If we need more we should use an IDX prefaced log message.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -436,10 +446,10 @@ public static object MarkAsSecurityArtifact(object arg, Func<object, string> cal
         /// <param name="innerException"><see cref="Exception"/></param>
         /// <param name="message">The log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
-        private static LogEntry WriteEntry(EventLogLevel eventLogLevel, Exception innerException, string message, params object[] args)
+        private static LogEntry WriteEntry(EventLogLevel eventLogLevel, Exception? innerException, string message, params object[]? args)
         {
             if (string.IsNullOrEmpty(message))
-                return null;
+                throw new ArgumentNullException(nameof(message), "null message was passed to new entry");
```
keegan-caruso: same comment as above

src/Microsoft.IdentityModel.Logging/LogHelper.cs
keegan-caruso: either build up the message based on the path taken above or list out all of the params even if they are potentially null/unused? (while adhering to PII awareness)

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: Fair enough, I will remove the exception message. Thanks Keegan!

src/Microsoft.IdentityModel.Logging/LogHelper.cs
FuPingFranco: Same reply as above.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
keegan-caruso: at a minimum we could at least call out the type?

FuPingFranco: I am not sure which one will make the most sense in here. @jennyf19 @jmprieur do you any suggestions on improving this message?

jennyf19:
As discussed, closing this PR as we are on a tight ship schedule.

PR #2211, opened 2023-08-08
Bash version of updateAssemblyInfo script

mthalman:
This defines a Bash version of the existing [updateAssemblyInfo.ps1 script](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/updateAssemblyInfo.ps1). Since there are already existing Bash versions of the build and test scripts, it made sense to follow that pattern here as well.

This accommodates .NET's source-build requirements which disallow the dependency on PowerShell because it's not buildable from source by Linux distros. The implementation in this script is necessary for .NET's source-build in order to set the assembly versions correctly (see https://github.com/dotnet/source-build/issues/3565).

### DIFF
--- a/updateAssemblyInfo.sh
+++ b/updateAssemblyInfo.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+set -euo pipefail
+
+scriptroot=$(cd -P "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+
+packageType=${1:-preview}
+
+date=$(date '+%y%m%d%H%M%S')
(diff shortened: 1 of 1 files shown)


jennyf19:
@mthalman Thanks for the contribution...could you target dev7 branch instead? Thx.

mthalman:
Converted to a draft. I need to make some more changes so that this will run in Alpine Linux.

keegan-caruso (approved):

jennyf19 (approved):

jennyf19:
Thanks @mthalman

jennyf19:

PR #2213, opened 2023-08-09
Adding nullable to Token project

FuPingFranco:
Adding nullable to base Token project

### DIFF
--- a/src/Microsoft.IdentityModel.TestExtensions/TestTokenCreator.cs
+++ b/src/Microsoft.IdentityModel.TestExtensions/TestTokenCreator.cs
@@ -258,9 +258,9 @@ namespace Microsoft.IdentityModel.TestExtensions
             var tokenDescriptor = CreateTokenDescriptorWithInstanceOverrides();
 
             // Remove the default value and null out any class specific override.
-            tokenDescriptor?.Subject?.RemoveClaim(
+            tokenDescriptor.Subject?.RemoveClaim(
                 tokenDescriptor.Subject.FindFirst(JwtRegisteredClaimNames.Iss));
-            tokenDescriptor!.Issuer = null;
+            tokenDescriptor.Issuer = null;
             return CreateToken(tokenDescriptor);
         }
 
@@ -273,9 +273,9 @@ namespace Microsoft.IdentityModel.TestExtensions
             var tokenDescriptor = CreateTokenDescriptorWithInstanceOverrides();
 
             // Remove the default value and null out any class specific override.
-            tokenDescriptor?.Subject?.RemoveClaim(
+            tokenDescriptor.Subject?.RemoveClaim(
                 tokenDescriptor.Subject.FindFirst(JwtRegisteredClaimNames.Aud));
-            tokenDescriptor!.Audience = null;
+            tokenDescriptor.Audience = null;
             return CreateToken(tokenDescriptor);
         }
 
@@ -288,9 +288,9 @@ namespace Microsoft.IdentityModel.TestExtensions
             var tokenDescriptor = CreateTokenDescriptorWithInstanceOverrides();
 
             // Remove the default value.
-            tokenDescriptor?.Subject?.RemoveClaim(
+            tokenDescriptor.Subject?.RemoveClaim(
                 tokenDescriptor.Subject.FindFirst(JwtRegisteredClaimNames.Iat));
-            return CreateToken(tokenDescriptor!);
+            return CreateToken(tokenDescriptor);
         }
 
         /// <summary>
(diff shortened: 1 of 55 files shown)


jennyf19:
@FuPingFranco make sure to pull from dev7 and then move to your branch and do a `merge dev7` to get it updated & let VS resolve the merge conflicts for you.

FuPingFranco:
> @FuPingFranco make sure to pull from dev7 and then move to your branch and do a `merge dev7` to get it updated & let VS resolve the merge conflicts for you.

Yeap, I didn't see there was changes on dev7 before submitting the PR. I've merged and resolved conflicts on the last commit

src/Microsoft.IdentityModel.TestExtensions/TestTokenCreator.cs
```diff
@@ -258,9 +258,9 @@ public string CreateTokenWithMissingIssuer()
             var tokenDescriptor = CreateTokenDescriptorWithInstanceOverrides();
 
             // Remove the default value and null out any class specific override.
-            tokenDescriptor.Subject.RemoveClaim(
+            tokenDescriptor?.Subject?.RemoveClaim(
```
keegan-caruso: this doesn't make sense. we handle the null case then 2 lines down we state it is not null?

Can CreateTokenDescriptorWithInstanceOverrides return null?

src/Microsoft.IdentityModel.TestExtensions/TestTokenCreator.cs
```diff
@@ -288,9 +288,9 @@ public string CreateTokenWithMissingIssuedAt()
             var tokenDescriptor = CreateTokenDescriptorWithInstanceOverrides();
 
             // Remove the default value.
-            tokenDescriptor.Subject.RemoveClaim(
+            tokenDescriptor?.Subject?.RemoveClaim(
                 tokenDescriptor.Subject.FindFirst(JwtRegisteredClaimNames.Iat));
-            return CreateToken(tokenDescriptor);
+            return CreateToken(tokenDescriptor!);
```
keegan-caruso: why do we need to say this is not null here but not at the other call sites for CreateToken?

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -35,9 +35,9 @@ internal AsymmetricAdapter(SecurityKey key, string algorithm, bool requirePrivat
         {
         }
 
-        internal AsymmetricAdapter(SecurityKey key, string algorithm, HashAlgorithm hashAlgorithm, bool requirePrivateKey)
+        internal AsymmetricAdapter(SecurityKey key, string algorithm, HashAlgorithm? hashAlgorithm, bool requirePrivateKey)
         {
-            HashAlgorithm = hashAlgorithm;
+            HashAlgorithm = hashAlgorithm!;
```
keegan-caruso: I'm confused, the parameter explicitly allows for null, how can we say it is not null here?

src/Microsoft.IdentityModel.TestExtensions/TestTokenCreator.cs
jennyf19: only needed here:

`tokenDescriptor.Subject?.RemoveClaim(`

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
jennyf19: The property needs to be null:

`private HashAlgorithm? HashAlgorithm { get; set; }`

jennyf19: See lines 26-30, the delegate used by the public method are set correctly depending on the context.

jennyf19:
Let's slow down a bit. This PR is pretty massive and we need to go line by line and make sure we are doing the right thing. Please do another pass on the changes, as it seems like we have too many `!` where we shouldn't have them. 

---
In reply to: [1672154548](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2213#issuecomment-1672154548) [](http://example.com/codeflow?ancestors=1672154548,1672154548)

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -124,14 +126,16 @@ private int EventQueuePollingInterval
...
             TaskCreationOptions options = TaskCreationOptions.None,
-            IEqualityComparer<TKey> comparer = null,
+            IEqualityComparer<TKey>? comparer = null,
             bool removeExpiredValues = false,
             int removeExpiredValuesIntervalInSeconds = 300,
             bool maintainLRU = false)
         {
             _capacity = capacity > 0 ? capacity : throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(capacity)));
             _options = options;
+#pragma warning disable CS8714 // The type cannot be used as type parameter in the generic type or method. Nullability of type argument doesn't match 'notnull' constraint.
```
keegan-caruso, line 136: can you elaborate?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -493,7 +497,7 @@ public bool TryGetValue(TKey key, out TValue value)
 
             if (!_map.TryGetValue(key, out var cacheItem))
             {
-                value = default;
+                value = default!;
```
keegan-caruso, line 500: default of TValue could be null though right?

PR #2215, opened 2023-08-10
fix nullables

jennyf19:
@FuPingFranco I think I've fixed the issues in your branch, if you want to take a look. go ahead and merge if you agree.

### DIFF
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -2,12 +2,10 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Data;
 using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.Tracing;
 using System.Globalization;
 using System.Linq;
-using System.Text.RegularExpressions;
 using Microsoft.IdentityModel.Abstractions;
 
 namespace Microsoft.IdentityModel.Logging
@@ -334,14 +332,14 @@ namespace Microsoft.IdentityModel.Logging
 
             if (innerException != null) 
                 if (string.IsNullOrEmpty(argumentName))
-                    return (T)Activator.CreateInstance(typeof(T), message, innerException)!;
+                    return (T?)Activator.CreateInstance(typeof(T), message, innerException);
                 else
-                    return (T)Activator.CreateInstance(typeof(T), argumentName, message, innerException)!;
+                    return (T?)Activator.CreateInstance(typeof(T), argumentName, message, innerException);
             else
                 if (string.IsNullOrEmpty(argumentName))
-                    return (T)Activator.CreateInstance(typeof(T), message)!;
+                    return (T?)Activator.CreateInstance(typeof(T), message);
                 else
-                    return (T)Activator.CreateInstance(typeof(T), argumentName, message)!;
+                    return (T?)Activator.CreateInstance(typeof(T), argumentName, message);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -375,20 +373,22 @@ private static object SanitizeSecurityArtifact(object arg)
...
 
-        private static string? RemovePII(object arg)
+        private static string RemovePII(object arg)
         {
             if (arg is Exception ex && IsCustomException(ex))
                 return ex.ToString();
 
             if (arg is NonPII)
-                return arg!.ToString();
+                return arg.ToString()!; // the null check is done in FormatInvariant
```
MZOLN, line 382: I find those situations tricky. Although it is flow correct it might become fragile for change during refactoring as moving it around might change the dependency, but good one.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -375,20 +373,22 @@ private static object SanitizeSecurityArtifact(object arg)
...
 
-            return string.Format(CultureInfo.InvariantCulture, IdentityModelEventSource.HiddenPIIString, arg?.GetType().ToString() ?? "Null");
+            return string.Format(CultureInfo.InvariantCulture, IdentityModelEventSource.HiddenPIIString, arg.GetType().ToString() ?? "Null");
         }
 
         internal static bool IsCustomException(Exception ex)
         {
-            return ex!.GetType().FullName!.StartsWith("Microsoft.IdentityModel.", StringComparison.Ordinal);
+            var fullName = ex.GetType().FullName;
```
MZOLN, line 389: Although I like this better, this could still throw an ArgumentNullException and we would spend decades finding why. Could we handle the ex being null case? e.g. return string empty or null or throw ANE as appropriate?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -446,7 +446,7 @@ public static object MarkAsSecurityArtifact(object arg, Func<object, string> cal
             if (!_isHeaderWritten)
             {
                 string headerMessage = string.Format(CultureInfo.InvariantCulture, "Microsoft.IdentityModel Version: {0}. Date {1}. {2}",
-                    typeof(IdentityModelEventSource).Assembly!.GetName().Version!.ToString(),
+                    typeof(IdentityModelEventSource).Assembly.GetName().Version?.ToString(),
```
MZOLN, line 449: there should never be the case where Version is null.

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
```diff
@@ -90,7 +90,7 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
 
             for (int i = 0; i < eventData.Payload.Count; i++)
             {
-                _streamWriter.WriteLine(eventData!.Payload[i]!.ToString());
+                _streamWriter.WriteLine(eventData!.Payload[i]?.ToString());
```
MZOLN, line 93: Although it might look simpler and nicer it is a bit tricky for two reasons:
1. you have to know that streamewriter.WriteLine() allows arguments to be null
2. writing a null would write a new empty - line - this can affect the logs -> empty lines do add up if the data stream is big it could become something like \r\n\r\n\r\n\r\n in extreme cases being pretty long and big and this being an event listener for logging, could end up being the case
2. it is allways quicker to evaluate a boolean expression than call a function, thus a code that would take more lines of code if (eventData == null) continue;  if (eventData.Payload[i] == null) continue; would still be more performant and would produce less costly logs than nullables over here.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: true but `GetName()` can return null.

src/Microsoft.IdentityModel.Logging/TextWriterEventListener.cs
jennyf19: feel free to propose a PR w/the changes you are suggesting.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
jennyf19: Would be more helpful to have you review Franco's main PRs for this work.

PR #2216, opened 2023-08-10
initial commit for a few files

jennyf19:


src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -230,7 +230,9 @@ private static byte[] SignatureFunctionNotFound(byte[] _)
 
         private byte[] SignWithECDsa(byte[] bytes)
         {
-            return ECDsa!.SignHash(HashAlgorithm.ComputeHash(bytes));
+            if (HashAlgorithm != null)
```
MZOLN, line 233: good change

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -230,7 +230,9 @@ private static byte[] SignatureFunctionNotFound(byte[] _)
...
+                return ECDsa!.SignHash(HashAlgorithm.ComputeHash(bytes));
```
MZOLN, line 234: Are we sure that ECDsa in this case wont be null? It will end up with a NullReferenceException if it is so, might be as well worth checking that and throwing another InvalidOperationEception that would tell the developer that he needs to set it up?

From what I`m seeing in the original code, if i`m correct, if the developer initializes with RSA but then calls ECDsa a NRE would be thrown and then we would have to be guessing why (hard to say withuot source code). An IOE like you have above would be more informative.

This would apply to all those crypto methods.

jennyf19: That's what I originally thought, but they are set by the delegates. They will not be null, only HashAlgorithm could be null. I left a comment in Franco's original PR around this.

PR #2218, opened 2023-08-10
Make CryptoProviderFactory disposable

### LINKED ISSUE #2209: [Bug] CryptoProviderFactory.cs should implement IDisposable and dispose the CryptoProviderCache.
6.32.2

**Where is the issue?**

 * [x] M.IM.Tokens

This is captured by a test and prevent a partner from upgrading.

```csharp 
public class CryptoProviderFactory  // does not implement IDisposable
...
        public CryptoProviderFactory()
        {
            CryptoProviderCache = new InMemoryCryptoProviderCache() { CryptoProviderFactory = this };
        }
...
```
whereas `InMemoryCryptoProvider` is IDisposable.

**Expected behavior**
Implement IDisposable and dispose the in memory crypto provider cache.

**Actual behavior**
CryptoProviderFactory is not disposable whereas `InMemoryCryptoProvider` is IDisposable.

Note: This needs to be fixed in 6.x and in 7.x

keegan-caruso:


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs
```diff
@@ -704,5 +705,28 @@ protected virtual void ValidateState(OpenIdConnectProtocolValidationContext vali
...
         }
+
+        /// <inheritdoc/>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
+            {
+                if (disposing)
+                {
+                    // This is safe to dispose. What get's disposed on it is the cache,
```
westin-m:
```suggestion
                    // This is safe to dispose. What gets disposed on it is the cache,
```

jennyf19 (approved):
will this go in dev7 as well?

westin-m (approved):

brentschmaltz:
@keegan-caruso we can put dispose on CryptoProviderFactory, normal use the user never creates it, and we don't ever dispose of it.
The only time it is created is for extensibility.

jennyf19 (approved):

jennyf19:
closing PR for now...don't have confirmation from the partner if the issue will be fixed.

PR #2220, opened 2023-08-11
Use an OrdinalIgnoreCase HashSet instead of ToUpperInvariant on each property name

stephentoub:


brentschmaltz (approved):

jennyf19 (approved):

brentschmaltz:
@stephentoub i will make this change across the entire code base.
Thanks for the help.

PR #2221, opened 2023-08-11
Fix IdentityModel.Tokens dependencies

eerhardt:
IdentityModel.Tokens is still depending on System.Text.* packages in net6.0, but these are unnecessary.

The reason is a type-o in the .csproj - we need a `.`.


stephentoub (approved):

jennyf19 (approved):

ViktorHofer (commented):
Thanks for the fix. Not the first time this happened, unfortunately.

PR #2214, opened 2023-08-09
First cut at OpenIdConnectConfiguration and SignedHttpRequest

brentschmaltz:
These changes remove the use of Newtonsoft from OpenIdConnectConfiguration and OpenIdConnectMessage completely.
SignedHttpRequestHandler still has some dependency on usage in JsonWebTokenHandler. That will be next.

The serialization of OpenIdConnectConfiguration follows the pattern of JsonWebKey and JsonWebKeySet.
SignedHttpRequestHandler changes were mostly related to using Utf8JsonReader/Writers all the through the stack rather than creating temporary strings that need to be transformed back utf8bytes.
This led to a number of changes in the SignedHttpRequestHandler tests.

The tests that checked for a null payload were no longer needed as the temporary dictionary is not used anymore.

PR #2222, opened 2023-08-11
Avoid unnecessary ToArrays in JsonKeyWebSerializer

stephentoub:
These methods should likely be switched to just use JsonSerializer and let it handle pooling of Utf8JsonWriters and streams and the like. But for now, we can at least avoid an interim byte[] that's not needed.


brentschmaltz:
@stephentoub why would we switch them to use JsonSerializer?

brentschmaltz (approved):

stephentoub:
> why would we switch them to use JsonSerializer?

We should measure, but this function is doing effectively what JsonSerializer would do, except JsonSerializer is already optimized to pool all of the intermediate objects (a MemoryStream and its underlying byte[], a Utf8JsonWriter, etc.).

jennyf19 (approved):

brentschmaltz:
@stephentoub i pulled these into https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2214

PR #2233, opened 2023-08-17
Removal of dependency on Newtonsoft

brentschmaltz:
This PR removes the use of Newtonsoft from IdentityModel by using Utf8JsonReader/Writer
JsonWebTokens were previously populated with JsonDocument.Parse, which resulted Claims and Values being parsed from  a JsonElement each time properties were accessed.
This PR introduces a new model of reading with a Utf8JsonReader and creating a collection of values that have been deserialized from Json. This will improve performance when accessing either a System.Claim or a C# property. There is more work to do as it will be helpful if the JsonElement is remembered when parsing as when creating a Claim, the serialized Json is needed in the claim.

A first cut at improving creation of JWT's when a reduced number of string -> Utf8bytes -> string + string -> Utf8bytes transformations were taking place. The goal is to use buffer pooling and write into the buffer (Span) once, perform all work (signature, compression, encryption) on the buffer.

Standardized roundtrip serialization was improved with the goal of providing a deterministic model for 'additional' data found on some objects.


src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1127,15 +1130,19 @@ internal virtual async Task<SecurityKey> ResolvePopKeyFromJkuAsync(string jkuSet
             // If there are multiple keys in the referenced JWK Set document, a "kid" member MUST also be included
             // with the referenced key's JWK also containing the same "kid" value.
             // https://datatracker.ietf.org/doc/html/rfc7800#section-3.5
-            else if (cnf.TryGetValue(ConfirmationClaimTypes.Kid, StringComparison.Ordinal, out var kid))
+            else if (!string.IsNullOrEmpty(cnf.Kid))
```
jennyf19, line 1133: do you need a null check on cnf?

src/Microsoft.IdentityModel.TestExtensions/Microsoft.IdentityModel.TestExtensions.csproj
```diff
@@ -19,6 +19,10 @@
     <ProjectReference Include="..\System.IdentityModel.Tokens.Jwt\System.IdentityModel.Tokens.Jwt.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftVersion)" />
```
jennyf19, line 23: >(NewtonsoftVersion) [](http://example.com/codeflow?start=58&length=19)

do we have to keep it for the tests?

src/System.IdentityModel.Tokens.Jwt/JsonClaimValueTypes.cs
```diff
@@ -1,29 +1,34 @@
...
     /// </summary>
     public static class JsonClaimValueTypes
     {
         /// <summary>
-        /// A URI that represents the JSON XML data type.
+        /// A value that indicates the <see cref="Claim.Value"/> is a Json object.
         /// </summary>
-        /// <remarks>When mapping json to .Net Claim(s), if the value was not a string (or an enumeration of strings), the ClaimValue will serialized using the current JSON serializer, a property will be added with the .Net type and the ClaimTypeValue will be set to 'JsonClaimValueType'.</remarks>
-        public const string Json = Microsoft.IdentityModel.JsonWebTokens.JsonClaimValueTypes.Json;
+        /// <remarks>When creating a <see cref="Claim"/> from Json to if the value was not a simple type {String, Null, True, False, Number}
```
jennyf19: >to [](http://example.com/codeflow?start=67&length=2)

nit; remove "to"? same for the other xml comments

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonClaimSetTests.cs
```diff
@@ -34,14 +34,15 @@ public class JsonClaimSetTests
         };
 
         [Theory, MemberData(nameof(ClaimSetTestCases))]
+
         public void ClaimSetGetValueTests(JsonClaimSetTheoryData theoryData)
```
jennyf19: nit remove line

build/releaseBuild.yml
```diff
@@ -105,7 +105,8 @@ jobs:
       command: 'custom'
       projects: 'wilson.sln'
       custom: 'msbuild'
-      arguments: '/r:True /p:Configuration=$(BuildConfiguration) /p:Platform="Any CPU" /verbosity:m /p:SourceLinkCreate=true /p:RunApiCompat=true'
+      arguments: '/r:True /p:Configuration=$(BuildConfiguration) /p:Platform="Any CPU" /verbosity:m /p:SourceLinkCreate=true'
+#      arguments: '/r:True /p:Configuration=$(BuildConfiguration) /p:Platform="Any CPU" /verbosity:m /p:SourceLinkCreate=true /p:RunApiCompat=true'
```
eerhardt, line 109: Why is ApiCompat disabled?

jennyf19: We just need to update the baseline, but mostly disabled so we could get a build to test in our entire stack.

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -18,121 +19,134 @@ namespace Microsoft.IdentityModel.JsonWebTokens
...
         }
-
-        internal JsonClaimSet(byte[] jsonBytes) : this(JsonDocument.Parse(jsonBytes))
+        internal JsonClaimSet(byte[] jsonUtf8Bytes)
         {
+            _jsonClaims = JwtTokenUtilities.CreateClaimsDictionary(jsonUtf8Bytes, jsonUtf8Bytes.Length);
         }
 
-        internal JsonClaimSet(string json) : this(JsonDocument.Parse(json))
+        internal JsonClaimSet(string json) : this(Encoding.UTF8.GetBytes(json))
```
eerhardt: Is this one of the places we could use the ArrayPool so we don't need to allocate a new `byte[]` just to turn it into a Dictionary and throw the byte array away?

jennyf19:
>                 var claimFromGetClaim = jsonWebToken.GetClaim(claim.Type);

getting this error locally:

   str1 != str2, StringComparison: 'Ordinal'
[http://www.w3.org/2001/XMLSchema#double](http://www.w3.org/2001/XMLSchema#double)
[http://www.w3.org/2001/XMLSchema#integer32](http://www.w3.org/2001/XMLSchema#integer32)

---
Refers to: test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs:78 in 40be4d0. [](commit_id = 40be4d0ca81e928de1f84f2313277cbc2258a55c, deletion_comment = False)

jennyf19:
>         public void TryGetClaim()

getting this error locally:

str1 != str2, StringComparison: 'Ordinal'
[http://www.w3.org/2001/XMLSchema#double](http://www.w3.org/2001/XMLSchema#double)
[http://www.w3.org/2001/XMLSchema#integer32](http://www.w3.org/2001/XMLSchema#integer32)

---
Refers to: test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs:160 in 40be4d0. [](commit_id = 40be4d0ca81e928de1f84f2313277cbc2258a55c, deletion_comment = False)

jennyf19:
>         public void GetClaim()

str1 != str2, StringComparison: 'Ordinal'
[http://www.w3.org/2001/XMLSchema#double](http://www.w3.org/2001/XMLSchema#double)
[http://www.w3.org/2001/XMLSchema#integer32](http://www.w3.org/2001/XMLSchema#integer32)

---
Refers to: test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs:134 in 40be4d0. [](commit_id = 40be4d0ca81e928de1f84f2313277cbc2258a55c, deletion_comment = False)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -5,18 +5,17 @@
...
 
-
 namespace Microsoft.IdentityModel.JsonWebTokens
 {
     /// <summary>
     /// A <see cref="SecurityToken"/> designed for representing a JSON Web Token (JWT). 
     /// </summary>
     public class JsonWebToken : SecurityToken
     {
+        internal object _audienceLock = new();
```
jennyf19: >_audienceLock [](http://example.com/codeflow?start=24&length=13)

this needs to be used in the locking below, otherwise we get a null ref

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -613,18 +612,16 @@ public IEnumerable<string> Audiences
             {
                 if (_audiences == null)
                 {
-                    _audiences = new List<string>();
-
-                    if (Payload.TryGetValue(JwtRegisteredClaimNames.Aud, out JsonElement audiences))
+                    lock (_audiences)
```
jennyf19: >_audiences [](http://example.com/codeflow?start=26&length=10)

should be `_audienceLock`

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -613,18 +612,16 @@ public IEnumerable<string> Audiences
...
-                        if (audiences.ValueKind == JsonValueKind.String)
+                        if (_audiences == null)
                         {
-                            _audiences.Add(audiences.GetString());
-                        }
-                        else if (audiences.ValueKind == JsonValueKind.Array)
-                        {
-                            foreach (JsonElement jsonElement in audiences.EnumerateArray())
-                                _audiences.Add(jsonElement.ToString());
+                            _audiences = new List<string>();
```
keegan-caruso: I believe you have a race condition here.

You need to wait until the collection is fully initialized before assigning the reference.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -337,49 +286,213 @@ public virtual string CreateToken(string payload, SigningCredentials signingCred
...
+                        }
+                    }
+
+                    if (!typeWritten)
+                        writer.WriteString(JwtHeaderUtf8Bytes.Typ, string.IsNullOrEmpty(tokenType) ? JwtConstants.HeaderType : tokenType);
+
+                    writer.WriteEndObject();
+                    writer.Flush();
+
+                    return memoryStream.ToArray();
```
keegan-caruso, line 370: here and below, can we use the memory streams buffer instead of allocating a new array?

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -328,125 +284,317 @@ internal static string ReadString(ref Utf8JsonReader reader, string propertyName
...
+                if (DateTime.TryParse(originalString, out DateTime dateTimeValue))
+                {
+                    dateTimeValue = dateTimeValue.ToUniversalTime();
+                    string dtUniversal = dateTimeValue.ToString("o", CultureInfo.InvariantCulture);
+                    if (dtUniversal.Equals(originalString, StringComparison.Ordinal))
+                        return dateTimeValue;
+                }
+            }
+            catch(Exception)
+            { }
```
keegan-caruso, line 316: do we need to log or something?

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -328,125 +284,317 @@ internal static string ReadString(ref Utf8JsonReader reader, string propertyName
...
         }
 
-        internal static void WriteObject(ref Utf8JsonWriter writer, string key, object obj)
+        /// <summary>
+        /// Writes an 'object' as a JsonProperty.
+        /// This was written to support what IdentityModel6x supported and is not meant to be a
+        /// general object serializer.
+        /// If a user needs to serialize a special value, then serialize the value into a JsonElement.
+        /// </summary>
+        /// <param name="writer"></param>
```
keegan-caruso, line 442: empty params

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -328,125 +284,317 @@ internal static string ReadString(ref Utf8JsonReader reader, string propertyName
...
 
-        internal static void WriteStrings(ref Utf8JsonWriter writer, ReadOnlySpan<byte> propertyName, IList<string> strings)
+        /// <summary>
+        /// Writes values into an array.
+        /// Assumes the writer.StartArray() has been called.
+        /// </summary>
+        /// <param name="writer"></param>
+        /// <param name="obj"></param>
+        /// <param name="depth">The current depth of recursive call for objects.
+        /// Maximum is 3.</param>
```
keegan-caruso: isn't max depth 2?

test/System.IdentityModel.Tokens.Jwt.Tests/JwtTestData.cs
```diff
@@ -316,8 +317,8 @@ public static TheoryData<JwtTheoryData> InvalidEncodedSegmentsData(string errorS
                 CanRead = true,
                 TestId = nameof(EncodedJwts.InvalidPayload),
                 Token = EncodedJwts.InvalidPayload,
-                ExpectedException = ExpectedException.ArgumentException(substringExpected: "IDX12723:", inner: typeof(JsonReaderException))
-            });
+                ExpectedException = new ExpectedException(typeof(ArgumentException), "IDX12723:", null, true)
+            }); ;
```
keegan-caruso: nit: extra semicolon

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz: Fixed, silly typo.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz: fixed.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz: @keegan-caruso nice catch, fixed.

src/System.IdentityModel.Tokens.Jwt/JsonClaimValueTypes.cs
brentschmaltz: @jennyf19 fixed, thanks.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonClaimSetTests.cs
brentschmaltz: Fixed.

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
brentschmaltz: @eerhardt, this method is not used anymore, it was helpful during the coding.
removed.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
brentschmaltz: @keegan-caruso , yea i kept going back and forth between 2 and 3.
fixed.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtTestData.cs
brentschmaltz: fixed

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
BrennanConroy: You could do that if you pass in the MemoryBuffer to this function. Since there is only 1 call site and it immediately uses and throws away the buffer after getting a string that's a nice easy allocation saving.

```csharp
using var MemoryStream memoryStream = new MemoryStream();
WriteJwsHeader(memoryStream, ...);
string header = Base64UrlEncoder.Encode(memoryStream.GetBuffer(), 0, (int)memoryStream.Length);

internal static void WriteJwsHeader(MemoryStream memoryStream, ...);
```

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -61,111 +73,71 @@ internal static Exception CreateJsonReaderExceptionInvalidType(ref Utf8JsonReade
...
+                    writer = new(memoryStream, new JsonWriterOptions { Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping });
+                    writer.WriteStartArray();
+
+                    foreach (string str in strings)
+                        writer.WriteStringValue(str);
+
+                    writer.WriteEndArray();
+                    writer.Flush();
+
+                    Utf8JsonReader reader = new(memoryStream.ToArray());
```
BrennanConroy:
```suggestion
                    Utf8JsonReader reader = new(memoryStream.GetBuffer().AsSpan(0, (int)memoryStream.Length));
```

brentschmaltz: @BrennanConroy thanks, i thought i got them all.

jennyf19 (approved):

jmprieur:
One breaking change in Wilson7 that we might want to fix (but in preview4) is: 

```csharp
jsonWebToken.TryGetPayload(JwtClaimsTypes.Roles, out roles) 
```
used to accept that roles is a `string[]`. I now needs to be a `IList<string>`

@brentschmaltz: 
- is it expected?
- do we agree that we would provide something in preview4? (not preview 3)?

jennyf19:
> One breaking change in Wilson7 that we might want to fix (but in preview4) is:
> 
> ```cs
> jsonWebToken.TryGetPayload(JwtClaimsTypes.Roles, out roles) 
> ```
> 
> used to accept that roles is a `string[]`. I now needs to be a `IList<string>`
> 
> @brentschmaltz:
> 
> * is it expected?
> * do we agree that we would provide something in preview4? (not preview 3)?

Thoughts @kevinchalet ?

kevinchalet:
> Thoughts @kevinchalet ?

Thanks for tagging me, @jennyf19!

I'm assuming `TryGetPayload` actually means `TryGetPayloadValue`. If so, OpenIddict uses it in 3 places: twice with `string` and once with `ImmutableDictionary<string, string[]>`. Chances things are broken with `string` are low, but with `ImmutableDictionary<string, string[]>`, it's not unreasonable so I'll give it a try 😄 

Was this PR included in the preview that shipped yesterday?

jennyf19:
> > Thoughts @kevinchalet ?
> 
> Thanks for tagging me, @jennyf19!
> 
> I'm assuming `TryGetPayload` actually means `TryGetPayloadValue`. If so, OpenIddict uses it in 3 places: twice with `string` and once with `ImmutableDictionary<string, string[]>`. Chances things are broken with `string` are low, but with `ImmutableDictionary<string, string[]>`, it's not unreasonable so I'll give it a try 😄
> 
> Was this PR included in the preview that shipped yesterday?

Yes, this is in preview3, which is out on NuGet as of Sunday (8/20) night. Would appreciate your feedback on this one.

kevinchalet:
> Would appreciate your feedback on this one.

Sadly, I'm seeing important regressions when moving to preview3. I'll open dedicated tickets.

kevinchalet:
> Sadly, I'm seeing important regressions when moving to preview3. I'll open dedicated tickets.

Here's the first one: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2244

kevinchalet:
And the second one: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2245.

kevinchalet:
And the third one: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2246.

It seems the migration to `System.Text.Json` has caused a lot of side effects/regressions. More agressive E2E testing would be a good idea. Having such bugs in the final version would be absolutely terrible 😅

jennyf19:
> And the third one: #2246.
> 
> It seems the migration to `System.Text.Json` has caused a lot of side effects/regressions. More agressive E2E testing would be a good idea. Having such bugs in the final version would be absolutely terrible 😅

We agree. Appreciate the bugs and all the thorough testing. We will improve our coverage as well.

Issue #2232, opened 2023-08-17
[Feature Request] Enable perf testing similar to aspnetcore 


jennyf19:
Enable perf testing on PRs using a config maintained within the repository.

Usage and examples can be found in the [crank documentation](https://github.com/dotnet/crank/tree/main/src/Microsoft.Crank.PullRequestBot).

jennyf19:
Included in [preview4 release](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/releases/tag/7.0.0-preview4)

PR #2235, opened 2023-08-18
Create microbenchmark for ValidateTokenAsync

westin-m:
Initial creation of microbenchmarks to be run with [crank](https://github.com/dotnet/crank/blob/main/docs/microbenchmarks.md). 

Included is a config file to run PR benchmarks.

Regarding the conditions added to the IsTrimmable items, those projects target frameworks that do not support it, and this breaks the benchmark app's build.


benchmark/MIcrosoft.IdentityModel.Benchmarks/MIcrosoft.IdentityModel.Benchmarks.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
keegan-caruso: typo in the file name, you have an extra capital

keegan-caruso: a couple files actually

benchmark/MIcrosoft.IdentityModel.Benchmarks/prbenchmarks.identitymodel.config.yml
```diff
@@ -0,0 +1,22 @@
...
+components:
+    jwt:
+        script: |
+            call dotnet build -c release --framework net8.0 .\src\Microsoft.IdentityModel.JsonWebTokens\Microsoft.IdentityModel.JsonWebTokens.csproj
+
+        arguments:
+            --application.options.outputFiles .\src\Microsoft.IdentityModel.JsonWebTokens\bin\release\net8.0\
+
+# default arguments that are always used on crank commands
+defaults: --config https://github.com/aspnet/Benchmarks/blob/main/build/azure.profile.yml?raw=true --application.framework net7.0
```
keegan-caruso: net7?

benchmark/MIcrosoft.IdentityModel.Benchmarks/identitymodel.benchmarks.yaml
```diff
@@ -0,0 +1,24 @@
+﻿jobs:
+  benchmarks:
+    source:
+      repository: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet
+      branchOrCommit: westinm/benchmark
```
sebastienros: Remember to change it before it's merged

benchmark/MIcrosoft.IdentityModel.Benchmarks/identitymodel.benchmarks.yaml
```diff
@@ -0,0 +1,24 @@
...
+    arguments: --job {{jobArg}} --filter {{filterArg}} --memory
+    options:
+      benchmarkDotNet: true
+    
+scenarios:
+  JsonWebTokenTests:
+    application:
+      job: benchmarks
+
+profiles:
```
sebastienros: Remove this. You should only use crank locally for debugging purposes, and it's already defined in public profile files, you should import of the aspnet ones instead (standard profiles in /scenarios).

benchmark/MIcrosoft.IdentityModel.Benchmarks/prbenchmarks.identitymodel.config.yml
sebastienros: Not the correct profile, use the one I suggested

benchmark/MIcrosoft.IdentityModel.Benchmarks/prbenchmarks.identitymodel.config.yml
```diff
@@ -0,0 +1,22 @@
...
+        script: |
+            call dotnet build -c release --framework net8.0 .\src\Microsoft.IdentityModel.JsonWebTokens\Microsoft.IdentityModel.JsonWebTokens.csproj
+
+        arguments:
+            --application.options.outputFiles .\src\Microsoft.IdentityModel.JsonWebTokens\bin\release\net8.0\
+
+# default arguments that are always used on crank commands
+defaults: --config https://github.com/aspnet/Benchmarks/blob/main/build/azure.profile.yml?raw=true --application.framework net7.0  
+
+# the first vaule is the defualt if none is specified
```
sebastienros: `u`s seems to be too fast on your keyboard

benchmark/MIcrosoft.IdentityModel.Benchmarks/JsonWebTokenTests.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.IdentityModel.Tokens;
+using Microsoft.IdentityModel.TestUtils;
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace Microsoft.IdentityModel.Benchmarks
+{
+    public class JsonWebTokenTests
+    {
+
```
brentschmaltz: extra line

benchmark/MIcrosoft.IdentityModel.Benchmarks/JsonWebTokenTests.cs
```diff
@@ -0,0 +1,43 @@
...
+            tokenDescriptor = new SecurityTokenDescriptor
+            {
+                Subject = new ClaimsIdentity(Default.PayloadClaims),
+                SigningCredentials = KeyingMaterial.JsonWebKeyRsa256SigningCredentials,
+            };
+            token = tokenHandler.CreateToken(tokenDescriptor);
+            validationParameters = new TokenValidationParameters()
+            {
+                ValidAudience = "http://Default.Audience.com",
+                ValidateLifetime = false,
```
brentschmaltz: We should validate lifetime.

benchmark/MIcrosoft.IdentityModel.Benchmarks/JsonWebTokenTests.cs
```diff
@@ -0,0 +1,43 @@
...
+                ValidIssuer = "http://Default.Issuer.com",
+                IssuerSigningKey = KeyingMaterial.JsonWebKeyRsa256SigningCredentials.Key,
+            };
+        }
+
+        [Benchmark]
+        public async Task<TokenValidationResult> ValidateTokenAsync() => await tokenHandler.ValidateTokenAsync(token, validationParameters).ConfigureAwait(false);
```
brentschmaltz: It would be great to compare JwtSecurityTokenHandler also.
Can we have more than one benchmark in these tests?

westin-m: Yes, you can have however many benchmarks in a single class.
I will add a benchmark for that handler.

jennyf19 (approved):

benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        [GlobalSetup(Targets = new[] { nameof(JsonWebTokenHandler) })]
+        public void JsonWebTokenSetup()
+        {
+            jsonWebTokenHandler = new JsonWebTokenHandler();
+        }
+
+        [Benchmark]
+        public async Task<TokenValidationResult> JsonWebTokenHandler() => await jsonWebTokenHandler.ValidateTokenAsync(jsonWebToken, validationParameters).ConfigureAwait(false);
```
brentschmaltz: @westin-m Is there a way to a descriptive name in the report.
When you are running this, you know its validating a token.
When several reports are in front of us, how do we know what action we are doing?
I see several such runs, reading a token, creating a token, validating a token, getting claims, getting properties, etc.

westin-m: The name in the report table will always be the method name of the benchmark, but there is a way to include more, even custom, columns. https://benchmarkdotnet.org/articles/configs/columns.html

As for several reports, it can depend on the context. In each of those several runs, they are likely best split into different classes and measured separately. I've not yet found a built-in way to easily label results such that they give the full context of what's being evaluated.

jennyf19: change the method name to something more specific for example: `ValidateAsync`

benchmark/Microsoft.IdentityModel.Benchmarks/CreateTokenTests.cs
```diff
@@ -0,0 +1,32 @@
...
+            jsonWebTokenHandler = new JsonWebTokenHandler();
+            tokenDescriptor = new SecurityTokenDescriptor
+            {
+                Subject = new ClaimsIdentity(Default.PayloadClaims),
+                SigningCredentials = KeyingMaterial.JsonWebKeyRsa256SigningCredentials,
+            };
+        }
+
+        [Benchmark]
+        public string CreateToken() => jsonWebTokenHandler.CreateToken(tokenDescriptor);
```
keegan-caruso: This name also needs updated, at a minimum needs to indicate it is coming from JsonWebTokenHandler

jennyf19:
#2232

benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        SecurityTokenDescriptor tokenDescriptor;
+        string jsonWebToken;
+        TokenValidationParameters validationParameters;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            tokenDescriptor = new SecurityTokenDescriptor
+            {
+                Subject = new ClaimsIdentity(Default.PayloadClaims),
```
brentschmaltz: I assume Default.PayloadClaims is a collection of System.Security.Claims.Claim.
We do a lot of work translating into a Dictionary<string,object>[see](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/af9abdb1cad7826bc1ffbc420562781306ebd954/src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs#L43)This as this is another degree of freedom and will make a difference in cpu and memory usage.

A minimum baseline would populate SecurityTokenDescriptor.Claims with a representative set of name/value pairs including "exp", "nbf", "iat". These values will need to be dynamic.

Then set JsonWebTokenHandler.SetDefaultTimesOnTokenCreation = false;

PR #2237, opened 2023-08-21
Avoid a few more allocations

stephentoub:
- JsonWebToken.Audiences can allocate a string[] of exactly the right size, saving on intermediate string[] allocations as well as a List allocation. If empty, it can avoid allocations completely.
- ValidateIssuerAsync can avoid a Task allocation in the common case where it completes synchronously.
- Also make some static fields readonly and change a few types to be concrete classes rather than interfaces; it helps the JIT generate better code.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
@@ -6,7 +6,6 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Security.Claims;
-using System.Text;
 using System.Text.Json;
 using Microsoft.IdentityModel.Logging;
 using Microsoft.IdentityModel.Tokens;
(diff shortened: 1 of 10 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -271,19 +270,20 @@ internal T GetValue<T>(string key, bool throwEx, out bool found)
 
             if (typeof(T) == typeof(IList<string>))
             {
-                List<string> list = new();
                 if (obj is IList iList)
                 {
-                    foreach (object item in iList)
-                        list.Add(item?.ToString());
+                    string[] arr = new string[iList.Count];
```
brentschmaltz, line 275: I didn't want to add null strings.

stephentoub: The existing code is adding null strings:
`list.Add(item?.ToString())`
will be null if either `item` is null or ToString returns null. You're saying the existing code is wrong? I was just maintaining that behavior.

brentschmaltz: @stephentoub let's stay with what we have, i think the existing code could have been improved and will focus on such issues later.
When validating the audience of a token, a null value is not useful.

Issue #2240, opened 2023-08-21
[Bug] Add nullable to Wctx in WsFederationMessage


jennyf19:
should be `public string? Wctx`

as `wctx` is optional according to spec: http://docs.oasis-open.org/wsfed/federation/v1.2/os/ws-federation-1.2-spec-os.html#_Toc223175023

jennyf19:
Included in [preview4 release](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/releases/tag/7.0.0-preview4)

PR #2241, opened 2023-08-21
add nullable to wctx

jennyf19:
#2240

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
+++ b/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
@@ -336,7 +336,7 @@ namespace Microsoft.IdentityModel.Protocols.WsFederation
         /// Gets or sets 'wattr'.
         /// </summary>
         [property: SuppressMessage("Microsoft.Naming", "CA1704")]
-        public string Wattr
+        public string? Wattr
         {
             get { return GetParameter(WsFederationConstants.WsFederationParameterNames.Wattr); }
             set { SetParameter(WsFederationConstants.WsFederationParameterNames.Wattr, value); }
@@ -346,7 +346,7 @@ namespace Microsoft.IdentityModel.Protocols.WsFederation
         /// Gets or sets 'wattrptr'.
         /// </summary>
         [property: SuppressMessage("Microsoft.Naming", "CA1704")]
-        public string Wattrptr
+        public string? Wattrptr
         {
             get { return GetParameter(WsFederationConstants.WsFederationParameterNames.Wattrptr); }
             set { SetParameter(WsFederationConstants.WsFederationParameterNames.Wattrptr, value); }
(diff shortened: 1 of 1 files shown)


src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
```diff
@@ -376,7 +376,7 @@ public string Wct
         /// Gets or sets 'wctx'.
         /// </summary>
         [property: SuppressMessage("Microsoft.Naming", "CA1704")]
-        public string Wctx
+        public string? Wctx
```
eerhardt, line 379: What about the other properties in this type? Should other ones be made `string?`?

jennyf19: Included the others marked as optional in the spec, but that only covers 3 more. @brentschmaltz what do you recommend?

jmprieur (approved):
LGTM
thanks @jennyf19

src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs
eerhardt: What about:

* Wattr
* Wattrptr
* Wauth

jennyf19: added and a few others, which seem to be optional in the spec.

brentschmaltz: @jennyf19 it really depends on what the WsFederationMessage is used for.
The spec specifies that wtrealm is required for a RequestSecurityToken message.
For a Pseudonym Request wtrealm is optional.

Worst that can happen is the request will fail.
I don't see what we are gaining by enforcing them to be not null.
Many times these message flow through a call graph where lower levels set values.

jennyf19: @eerhardt discussed with @brentschmaltz  and I think we are going to leave this as-is. As Brent points out, some of the properties can be optional or required depending on the context.

eerhardt: > some of the properties can be optional or required depending on the context.

Then my recommendation would be to make them all `string?`. This models how they can be used in the API - for example:

```C#
var m = new WsFederationMessage();
m.Wctx.ToString();  // This will result in a NullReferenceException, but the C# compiler won't warn me that `m.Wctx` could be null
```

It's better to make these properties `string?` if there is a chance they are `null`, than to say "some of these strings may be required in some contexts, so none of these strings can ever be null".

jennyf19: @brentschmaltz ^

brentschmaltz: We could mark them as nullable.
What i was saying is that we can't make the NOT nullable.

jennyf19: @eerhardt and @brentschmaltz marked them as nullable.

jmprieur (approved):

brentschmaltz (approved):

PR #2242, opened 2023-08-22
More strongly-type multiple fields and return types

stephentoub:
Some of these, like JsonClaimSet._jsonClaims, help to avoid allocation by ensuring that when the collection is enumerated, we're using its struct enumerator. In other cases, like with CreateClaimFromObject, we reduce the overhead of calls to the object.  And other cases are just good hygiene.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
@@ -22,10 +22,10 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
         internal static JsonClaimSet Empty { get; } = new JsonClaimSet("{}"u8.ToArray());
         internal object _claimsLock = new();
-        internal IDictionary<string, object> _jsonClaims;
-        private IList<Claim> _claims;
+        internal readonly Dictionary<string, object> _jsonClaims;
+        private List<Claim> _claims;
 
-        internal JsonClaimSet(IDictionary<string, object> jsonClaims)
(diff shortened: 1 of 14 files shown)


brentschmaltz (approved):

jennyf19 (approved):
:shipit: Thanks @stephentoub

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -43,16 +43,16 @@ internal IList<Claim> Claims(string issuer)
             return _claims;
         }
 
-        internal IList<Claim> CreateClaims(string issuer)
```
saviour123, line 46: @stephentoub  Notice you changed all Interfaces to Concrete Subtypes. Just curious what is the benefit or advantage of this.

stephentoub: Some of the advantages were noted in the PR description.

For example, if you have:
```C#
IList<T> list = ...;
foreach (T item in list) { ... }
```
that allocates an `IEnumerator<T>` object. If you instead have:
```C#
List<T> list = ...;
foreach (T item in list) { ... }
```
it doesn't, instead using the `List<T>.Enumerator` struct for enumeration.

But even just simple accesses, e.g.
```C#
IList<T> list = ...;
return list[2];
```
That involves interface dispatch, which is not only slower than a regular method invocation:
```C#
List<T> list = ...;
return list[2];
```
but without a technology like dynamic PGO, it can't be inlined.

jennyf19:

PR #2243, opened 2023-08-22
Use optimistic synchronization in JsonWebToken.Audiences

stephentoub:
I didn't see anything in Payload.TryGetValue that would require synchronization with a concurrent call to Payload.TryGetValue. And without that, and since concurrent access is not something to optimize for here, we can instead employ optimistic synchronization and just swap in the result with an interlocked. If there's a race condition and two threads do end up accessing Audiences on the same instance concurrently, they may both end up creating the array, but only one will be published for all to consume.

In addition to making Audiences cheaper to access, the primary benefit here is it saves an allocation per JsonWebToken, which no longer needs to construct the audiences lock object.


eerhardt (approved):
LGTM

westin-m (approved):

brentschmaltz (approved):

jennyf19:

PR #2248, opened 2023-08-22
Correctly handle claims of same type with multiple values

### LINKED ISSUE #2244: Claims with multiple values are incorrectly serialized in 7.0.0-preview3
There's a serious regression in 7.0.0-preview3 that affects the serialization of identities containing multiple claims of the same type:

```csharp
var identity = new ClaimsIdentity("Test");
identity.AddClaim(new Claim("a", "value1"));
identity.AddClaim(new Claim("a", "value2"));
identity.AddClaim(new Claim("a", "value3"));
identity.AddClaim(new Claim("a", "value4"));

var descriptor = new SecurityTokenDescriptor
{
    SigningCredentials = new SigningCredentials(new RsaSecurityKey(RSA.Create(2048)), SecurityAlgorithms.RsaSha256),
    Subject = identity
};

var token = handler.CreateToken(descriptor);
```

7.0.0-preview creates a token with the expected JSON payload:

```json
{
  "a": [
    "value1",
    "value2",
    "value3",
    "value4"
  ],
  "exp": 1692705694,
  "iat": 1692702094,
  "nbf": 1692702094
}
```

7.0.0-preview3 doesn't: two values are missing 😱 

```json
{
  "a": [
    "value1",
    "value2"
  ],
  "exp": 1692705632,
  "iat": 1692702032,
  "nbf": 1692702032
}
```

/cc @brentschmaltz @jennyf19 it's a serious regression that may have security implications as it can lead to incorrect authorization decisions. I suggest marking it as a priority bug and unlisting 7.0.0-preview3 until it's fixed 😃

keegan-caruso:
Fixes #2244

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
@@ -128,17 +128,17 @@ namespace Microsoft.IdentityModel.Tokens
                 // When the creating the JWT and a list is found, a JsonArray will be created.
                 if (payload.TryGetValue(claim.Type, out object existingValue))
                 {
-                    if (existingValue is not IList<object>)
+                    if (existingValue is IList<object> existingList)
+                    {
+                        existingList.Add(jsonClaimValue);
+                    }
+                    else
                     {
                         payload[claim.Type] = new List<object>
                         {
                             existingValue,
                             jsonClaimValue
-                        }; 
-                    }
-                    else if (existingValue is IList<object> existingList)
-                    {
-                        existingList.Add(jsonClaimValue);
+                        };
                     }
                 }
                 else
--- a/src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
@@ -136,6 +136,10 @@ namespace Microsoft.IdentityModel.Tokens
                             jsonClaimValue
                         }; 
                     }
+                    else if (existingValue is IList<object> existingList)
+                    {
+                        existingList.Add(jsonClaimValue);
+                    }
                 }
                 else
                 {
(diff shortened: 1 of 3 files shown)


test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -28,6 +28,40 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
 {
     public class JsonWebTokenHandlerTests
     {
+        [Fact]
```
jennyf19, line 31: good to have a test covering this now.

jennyf19 (approved):

jmprieur (approved):
LGTM
Thanks @keegan-caruso

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -136,6 +136,10 @@ internal class TokenUtilities
                             jsonClaimValue
                         }; 
                     }
+                    else if (existingValue is IList<object> existingList)
```
kevinchalet: You're basically doing the same check twice 😅 

Something like that might make more sense:

```csharp
if (existingValue is not IList<object> existingList)
{
    payload[claim.Type] = new List<object>
    {
        existingValue,
        jsonClaimValue
    }; 
}
else
{
    existingList.Add(jsonClaimValue);
}
```

(you can also invert the two branches to avoid the `not`)

keegan-caruso: fair point 😆 

I'll clean this up.

brentschmaltz (approved):

kevinchalet:
I'd like to test these changes but I don't see any mention of a nightly builds feed... tho' I'm sure you guys used to have such a thing in the past 😕

Am I missing something?

PR #2251, opened 2023-08-23
Fix description for JWK `X5tS256` field

polys:


jennyf19:
Thanks @polys - can you do this into dev7? Thank you.

polys:
Replaced with #2252

PR #2264, opened 2023-08-25
Adding support for some standard types when calling JsonWebToken.TryGetPayloadValue<T>

brentschmaltz:
Added to the support 'T' that were available in 6x.
If possible, Json will be mapped to one of the following types:
string[], List<string>,,Collection<string>, object[], List<object>, Collection<object>, int[], long[]
Dictionary<string, string>>, Dictionary<string, string[]>, Dictionary<string, List<string>>, Dictionary<string, Collection<string>>
Dictionary<string, object>, Dictionary<string, object[]>

With this change 7 will not be as general as 6x but will help with common scenarios.
A general solution will require a new api.

When calling TryGetPayloadValue<T> the priority is generally:
sting, string[], List<string>, Collection<string>, ...

Type.IsAssignable is used in a couple of places which helps in writing as this improves the types that can be written.
If this violates AOT, then we will have to reduce the scope of what can be written.

=================================
Pattern matching is applied when possible.
Json = {“p”:[1,2,3]}
TryGetPayLoadValue<string[]>(“p”) => string[]{“1”,”2”,”3”}

Json = {“p”:”value”}
TryGetPayLoadValue<string[]>(“p”) => string[]{“value”}

Json = {“p”:{”object”:”value”}}
TryGetPayLoadValue<string>(“p”) => “{“object”:”value”}”
This will always return a value if it was the property exists.

=====================================
We still write obj.ToString() when serializing, even if we don't understand the type, as users that are working with Newtosoft.JObject will benefit. We are open to changing this depending on feedback.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
@@ -17,6 +18,12 @@ namespace Microsoft.IdentityModel.Tokens.Json
     internal static class JsonSerializerPrimitives
     {
         internal const int MaxDepth = 2;
+        internal static Type typeofDictionary = typeof(IDictionary);
+        internal static Type typeofDictionaryString = typeof(Dictionary<string, string>);
+        internal static Type typeofDictionaryStrings = typeof(Dictionary<string, string[]>);
+        internal static Type typeofList = typeof(IList);
+        internal static Type typeofListString = typeof(List<string>);
+        internal static Type typeofArray = typeof(Array[]);
 
         /// <summary>
         /// Creates a JsonException that provides information on what went wrong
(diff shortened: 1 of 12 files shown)


jennyf19:
FYI @kevinchalet

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -119,6 +121,297 @@ public static JsonElement CreateJsonElement(string json)
...
+            else if (jsonElement.ValueKind == JsonValueKind.True)
+                return true;
+            else if (jsonElement.ValueKind == JsonValueKind.Number)
+            {
+                if (jsonElement.TryGetInt32(out int intValue))
+                    return intValue;
+                else if (jsonElement.TryGetInt64(out long longValue))
+                    return longValue;
+                else if (jsonElement.TryGetDecimal(out decimal decimalValue))
+                    return decimalValue;
```
jennyf19, line 144: nit: move this last as it's more specific and maybe less common compared to the other numeric types?

jmprieur:
@brentschmaltz 
I published a little test app published with native AoT and it got the same results as without. IsAssignable does not seem to be a problem.

Not for this PR, but should we raise an issue for this one:
controls_auds => Array: IDX11027: Cannot create return type: 'System.Array' from JsonElement.ValueType: 'Array'.
?

brentschmaltz:
I have an updated PR for these issues.

PR #2275, opened 2023-08-29
Remove System.Text.Encodings.Web references

### LINKED ISSUE #2277: Remove superfluous System.Text.Encodings.Web explicit references
**What?** We have some partners which are running into issues due to an explicit/forced usage of 4.7.2 version of System.Text.Encodings.Web. It appears that the explicit reference is not required for Wilson to build correctly.

**Why?** extra references which enforce higher than necessary versions limit partners' ability to leverage our library.

**To confirm:** would removing this reference open us back up to #1985 ?

TimHannMSFT:
The PR #1997 updated the references for System.Text.Encodings.Web.

This PR removes the references entirely because they were causing conflicts for users and are no longer in use.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Microsoft.IdentityModel.JsonWebTokens.csproj
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Microsoft.IdentityModel.JsonWebTokens.csproj
@@ -40,4 +40,9 @@
     <PackageReference Include="System.Text.Json" Version="$(SystemTextJson)" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net45' Or '$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.Memory" Version="4.5.0" />
+    <PackageReference Include="System.Buffers" Version="4.5.1" />
+  </ItemGroup>
+
 </Project>
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Microsoft.IdentityModel.JsonWebTokens.csproj
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Microsoft.IdentityModel.JsonWebTokens.csproj
@@ -36,7 +36,7 @@
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != 'NETCoreApp'">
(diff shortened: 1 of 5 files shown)


jmprieur (approved):
LGTM - great to see
Thanks @TimHannMSFT 

Do we also need to reference SystemText.Json explicitly?

TimHannMSFT:
From https://github.com/advisories/GHSA-ghhp-997w-qr28
> Microsoft is releasing this security advisory to provide information about a vulnerability in .NET 5.0, .NET Core 3.1, and .NET Core 2.1

Our `System.Text.Json` references are only needed in pre-netcore targets so we can avoid pulling in the encodings.web library for core targets by making SJT conditional to older targets.

src/Microsoft.IdentityModel.JsonWebTokens/Microsoft.IdentityModel.JsonWebTokens.csproj
```diff
@@ -29,12 +29,12 @@
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net461' Or '$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'net472' Or '$(TargetFramework)' == 'netstandard2.0' Or '$(TargetFramework)' == 'net6.0'">
     <PackageReference Include="System.Text.Encoding" Version="$(SystemTextEncoding)" />
-	<PackageReference Include="System.Text.Encodings.Web" Version="$(SystemTextEncodingsWeb)" />
-    <PackageReference Include="System.Text.Json" Version="$(SystemTextJson)" />
+
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net45' Or '$(TargetFramework)' == 'net461' Or '$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'net472'">
```
TimHannMSFT, line 35: Here and elsewhere, would 

```
<ItemGroup Condition="'$(TargetFrameworkIdentifier)' != 'NETCoreApp'">
```
be a more succinct way of expressing this? I'm not 100% clear on how that would interact with netstandard2.0.

I was going to update this but worried there may be some implications I didn't understand.

brentschmaltz: Wilson7 uses System.Text.Encodings.Web, so we will need to address that.

jennyf19:
going to try something else. closing for now.

PR #2283, opened 2023-08-30
Obsolete int claims, add long claims

### LINKED ISSUE #2266: [Feature Request] Deprecate int? JwtPayload.Exp, .Iat, .Nbf
These properties return int, EpochTime requires a long to correctly represent the datetime.

We should add new methods in 7
Expiration, IssuedAt, NotBefore of type long?.

And mark Exp, Iat, Nbf as obsolete.

westin-m:
Addresses #2266

### DIFF
--- a/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
@@ -350,9 +350,19 @@ namespace System.IdentityModel.Tokens.Jwt
         /// Gets the 'value' of the 'expiration' claim { exp, 'value' }.
         /// </summary>
         /// <remarks>If the 'expiration' claim is not found OR could not be converted to <see cref="Int32"/>, null is returned.</remarks>
+        [Obsolete("`int? JwtPayload.Exp` is deprecated and will be removed in a future release. Use `long? JwtPayload.Expiration` instead. For more information, see https://aka.ms/IdentityModel/7-breaking-changes")]
         public int? Exp
         {
             get => GetIntClaim(JwtRegisteredClaimNames.Exp);
+        }
+
+        /// <summary>
+        /// Gets the 'value' of the 'expiration' claim { exp, 'value' }.
+        /// </summary>
+        /// <remarks>If the 'expiration' claim is not found OR could not be converted to <see cref="long"/>, null is returned.</remarks>
+        public long? Expiration
+        {
+            get => GetLongClaim(JwtRegisteredClaimNames.Exp);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -350,11 +350,21 @@ public string CHash
         /// Gets the 'value' of the 'expiration' claim { exp, 'value' }.
         /// </summary>
         /// <remarks>If the 'expiration' claim is not found OR could not be converted to <see cref="Int32"/>, null is returned.</remarks>
+        [Obsolete("`int? JwtPayload.Exp` is deprecated and will be removed in a future release. Use `long? JwtPayload.Expiration` instead. For more information, see https://aka.ms/IdentityModel/7-breaking-changes")]
```
jmprieur, line 353: long? or DateTime / DateTimeOffset?

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -350,11 +350,21 @@ public string CHash
...
         public int? Exp
         {
             get => GetIntClaim(JwtRegisteredClaimNames.Exp);
         }
 
+        /// <summary>
+        /// Gets the 'value' of the 'expiration' claim { exp, 'value' }.
+        /// </summary>
+        /// <remarks>If the 'expiration' claim is not found OR could not be converted to <see cref="long"/>, null is returned.</remarks>
+        public long? Expiration
```
jmprieur, line 363: Don't we want to use a DateTime (or a DateTimeOffset)?

brentschmaltz: @jmprieur we have ValidFrom and ValidTo as DateTime.
This property was for users that wanted the epoch time.

jmprieur: Given it's in Wilon7 (which has breaking changes), can't we just return `long` then? instead of `int` ?

westin-m: @jmprieur we will be returning `long?` instead of `int?`, if I understand your question correctly

brentschmaltz (approved):

sruke:
Should EpochTime class support representing DateTime in long? It currently supports int. 

See https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/3ef57d97d048c1ca34a5bffdd9d94a2325e629fc/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L118 where DateTime is being read as an int.

EpochTime:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/2cf0feb2d50c5fe4c334d194294ccb1f6cb719af/src/Microsoft.IdentityModel.Tokens/EpochTime.cs#L25"

brentschmaltz:
@sruke EpochTime.GetIntDate returns a long.

sruke (approved):

PR #2284, opened 2023-08-31
Introduce maxdepth for reading and writing in JsonSerializerPrimitives

### LINKED ISSUE #2269: [Feature Request] JsonSerializationPrimitives needs to set a Maximum depth for recursion when deserializing/serializing
Wilson7 currently does not have a maximum level for deserializing a token where the json contains Arrays of Arrays.
For example {"a1":["a2",["a3",[.... ]]]}

We need to decide how may levels, the maximum size of a token is set, but that could be a lot of levels.

https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonreaderoptions.maxdepth?view=net-7.0 indicates 64 for reading
https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonwriteroptions.maxdepth?view=net-7.0 indicates 1000 for writing

keegan-caruso:
Fixes #2269

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
@@ -18,6 +18,13 @@ namespace Microsoft.IdentityModel.Tokens.Json
 {
     internal static class JsonSerializerPrimitives
     {
+        // This is not a general purpose JSON serializer. It is specifically
+        // made for the use in the IdentityModel libraries. As such, we can take a
+        // lower limit to both our reading and writing max depth.
+        // This number is the min between System.Text.Jsons default for
+        // writing and reading max depth.
+        const int MaxDepth = 64;
+
         /// <summary>
         /// Creates a JsonException that provides information on what went wrong
         /// </summary>
@@ -119,8 +126,14 @@ namespace Microsoft.IdentityModel.Tokens.Json
 #endif
         }
 
-        internal static object CreateObjectFromJsonElement(JsonElement jsonElement)
+        internal static object CreateObjectFromJsonElement(JsonElement jsonElement, int currentDepth)
         {
+            if (currentDepth >= MaxDepth)
+                throw new InvalidOperationException(LogHelper.FormatInvariant(
+                    LogMessages.IDX10815,
+                    LogHelper.MarkAsNonPII(currentDepth),
+                    LogHelper.MarkAsNonPII(MaxDepth)));
+
             if (jsonElement.ValueKind == JsonValueKind.String)
             {
                 if (DateTime.TryParse(jsonElement.GetString(), CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out DateTime dateTime))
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -169,9 +186,14 @@ internal static object CreateObjectFromJsonElement(JsonElement jsonElement)
                 foreach (JsonProperty property in jsonElement.EnumerateObject())
                     numItems++;
 
+                int index = 0;
```
keegan-caruso, line 187: small bugfix here

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -159,7 +172,11 @@ internal static object CreateObjectFromJsonElement(JsonElement jsonElement)
 
                 int index = 0;
                 foreach (JsonElement j in jsonElement.EnumerateArray())
-                    items[index++] = CreateObjectFromJsonElement(j);
+                {
+                    currentDepth++;
+                    items[index++] = CreateObjectFromJsonElement(j, currentDepth);
```
westin-m: Could you pass `currentDepth+1` , rather than incrementing then decrementing?

test/Microsoft.IdentityModel.Tokens.Tests/Json/JsonSerializerPrimitivesTests.cs
```diff
@@ -16,12 +17,64 @@ namespace Microsoft.IdentityModel.Tokens.Json.Tests
 {
     public class JsonSerializerPrimitivesTests
     {
+        [Fact]
+        public void ChechMaxDepthReading()
```
westin-m: Chech -> Check

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
brentschmaltz: @keegan-caruso thanks :-)

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -169,9 +186,14 @@ internal static object CreateObjectFromJsonElement(JsonElement jsonElement)
                 foreach (JsonProperty property in jsonElement.EnumerateObject())
                     numItems++;
 
+                int index = 0;
                 KeyValuePair<string, object>[] kvps = new KeyValuePair<string, object>[numItems];
                 foreach (JsonProperty property in jsonElement.EnumerateObject())
-                    kvps[numItems++] = new KeyValuePair<string, object>(property.Name, CreateObjectFromJsonElement(property.Value));
+                {
+                    currentDepth++;
```
brentschmaltz: Does mean that the Json can only have MaxDepth properties {"string":"value"} ... as the check is when the method is entered.
Should we differentiate between Object and Array?

keegan-caruso: what do you mean by:

Should we differentiate between Object and Array?

westin-m (approved):

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
brentschmaltz: I think the PR is fine, the currentDepth-- was under the comments and i missed it.
I agree with Westin that just passing currnetDepth+1 would be clearer.

brentschmaltz (approved):

PR #2285, opened 2023-08-31
Fixed some issues when adding some additional tests

brentschmaltz:
We were missing support for bool and ulong
DateTimes were not roundtripping correctly.
Added additional tests for TryGetPropertyValue.

Removed redundant tests.


src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -213,21 +213,23 @@ internal T GetValue<T>(string key, bool throwEx, out bool found)
...
+               return (T)((object)obj.ToString());
+            }
+            else if (typeof(T) == typeof(bool))
+            {
+                if (bool.TryParse(obj.ToString(), out bool value))
+                    return (T)(object)value;
             }
             else if (typeof(T) == typeof(int))
             {
-                if (objType == typeof(int))
```
jennyf19, line 220: why remove this?

jennyf19 (approved):

PR #2287, opened 2023-09-01
Change Task to ValueTask.

brentschmaltz:
We will have to coordinate this with other changes as Task is expected in some call graphs.


keegan-caruso:
if you want to go this route we have to revert these changes as well:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/5c9192c857e2b74d248466a32b6a562434d32ae0/src/Microsoft.IdentityModel.Tokens/Validators.cs#L218

brentschmaltz:
@keegan-caruso you make a good point, it will be easier to make a single change in AadIssuerValidator.

jmprieur:
Is it a breaking change for SAL? IdWeb?

westin-m (approved):

jennyf19 (approved):

PR #2296, opened 2023-09-06
Pulling in changes from dev7

brentschmaltz:
Rebased dev7 on top of dev.
All changes for 6x releases should be from the dev6x branch.


westin-m (approved):

jennyf19 (approved):
🥳

.github/workflows/dotnetcore.yml
```diff
@@ -0,0 +1,40 @@
+name: .NET Core
+
+on:
+  push:
+    branches:
+      - dev7
```
jennyf19, line 6: @brentschmaltz this will need to change to dev here and line 14

PR #2307, opened 2023-09-11
review preview from dev

jennyf19:


build/version.props
```diff
@@ -4,14 +4,4 @@
         <VersionPrefix>0.0.1</VersionPrefix>
         <VersionSuffix>preview</VersionSuffix>
     </PropertyGroup>
-
```
jmprieur, line 7: I don't understand. Why would we want to remove that?

`<Version>` is what matters for .NET pack

jmprieur (commented):
No sure about this one, @jennyf19

PR #2321, opened 2023-09-19
Fix build on macOS (Bash 3.2)

directhex:
# Fix build on macOS (Bash 3.2)

macOS ships with Bash 3.2, the last version under GPLv2, due to a corporate ban on GPLv3 code in macOS. The substring syntax used in updateAssemblyInfo.sh, with a negative value in the last field, is only supported in Bash 4.2+. This ugly version is more compatible across versions.

Use of $PWD is also replaced with `${scriptroot}`, which is already defined earlier in the script, and makes it easier to run the script outside a mandatory working directory.

### DIFF
--- a/updateAssemblyInfo.sh
+++ b/updateAssemblyInfo.sh
@@ -11,15 +11,15 @@ dateTimeStamp=$(echo $((10#${date:0:2}-19)))${date:2}
 
 commitSha=$(git rev-parse HEAD)
 
-assemblyVersion=$(sed -n 's/.*<assemblyVersion>\([^<]*\)<.*/\1/p' $PWD/buildConfiguration.xml)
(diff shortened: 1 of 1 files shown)


mthalman (approved):
Also wanna remove the unused `date` variable on line 8?

directhex:
That bit broke when I poked it too hard, so I would rather worry about that another day. Lots of repos to get through, to get the whole repo building again on Mac

brentschmaltz:
@directhex @mthalman not knowing enough about this, is this PR ready?

directhex:
IMHO it's fine. Seems to be passing your CI. Your call if you want extra eyes on it

mthalman:
It's good for me.

brentschmaltz (approved):

Issue #2328, opened 2023-09-26
[Bug] breaking change in 7: TryGetPayloadValue() changes the type of returned object (compared with Identity.Model 6.x) for the "aud" claim 


jmprieur:
**Which version of Microsoft.IdentityModel are you using?**
Note that to get help, you need to run the latest version. 

**Where is the issue?**
 * [ x ] M.IM.JsonWebTokens
 

**Is this a new or an existing app?**
Following-up the removal of NewtonSoft.json, the actual claim type  returned by TryGetPayloadValue when passing object, is different. For instance:
 
If we are trying to get the value of the audience `claim("aud")`, when there is a single audience, previously `subjectToken.TryGetPayloadValue("aud", out object claimValue)` would retum claimValue as a string. 
With Wilson 7.0.0, the claimValue retumed as a `IList`. 

It is important for customers to maintain the same type.

**Repro**

```csharp
subjectToken.TryGetPayloadValue("aud", out object claimValue);
Assert.True(claimValue is string);
```

**Expected behavior**
The claimValue for "aud" should be a string

**Actual behavior**
The claimValue for "aud" is now an IList<string> (with one element)

**Possible solution**
Return a string if there is only one audience.

**Workaround**

```csharp
subjectToken.TryGetPayloadValue("aud", out string claimValue);
Assert.True(claimValue is string);
```

brentschmaltz:
fixed

PR #2331, opened 2023-09-26
Revert to 6x behavior with GetPayloadClaim<object>("aud")

brentschmaltz:
In 6x, GetPayloadClaim<object>("aud") will return a single string when aud was a single value.
The changes in 7 resulted in a List<string> with one value, which broke some users.

This change returns to the same behavior as 6x.

Note: GetPayloadClaim<string>("aud") would return a string.

JsonWebToken.Audiences is a better approach to access the 'aud' values as during the deserialization of the JWT, the audience values are deserialized.

Addresses: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2328


jennyf19 (approved):

sruke (approved):
LGTM

PR #2339, opened 2023-09-29
[Log Scrubbing] Clean up log messages in Wilson

TimHannMSFT:
[Log Scrubbing] Clean up log messages in Wilson

## Description
- Fixing up logging messages in Wilson
- Adding unit tests for some logging methods

### DIFF
--- a/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
+++ b/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
@@ -46,6 +46,11 @@ namespace Microsoft.IdentityModel.Logging
         public static string HiddenPIIString { get; } = "[PII of type '{0}' is hidden. For more details, see https://aka.ms/IdentityModel/PII.]";
 
         /// <summary>
+        /// String that is used in place of any arguments to log messages if the 'LogCompleteSecurityArtifact' flag is set to false.
+        /// </summary>
+        public static string HiddenArtifactString { get; } = "[Security Artifact of type '{0}' is hidden. For more details, see https://aka.ms/IdentityModel/SecurityArtifactLogging.]";
+
+        /// <summary>
         /// Indicates whether or the log message header (contains library version, date/time, and PII debugging information) has been written.
         /// </summary>
         public static bool HeaderWritten { get; set; } = false;
--- a/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
+++ b/src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
@@ -48,7 +48,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <summary>
         /// String that is used in place of any arguments to log messages if the 'LogCompleteSecurityArtifact' flag is set to false.
         /// </summary>
-        public static string HiddenArtifactString { get; } = "[Security Artifact of type '{0}' is hidden. For more details, see https://aka.ms/IdentityModel/SecurityArtifactLogging.]";
+        public static string HiddenSecurityArtifactString { get; } = "[Security Artifact of type '{0}' is hidden. For more details, see https://aka.ms/IdentityModel/SecurityArtifactLogging.]";
 
         /// <summary>
         /// Indicates whether or the log message header (contains library version, date/time, and PII debugging information) has been written.
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -471,7 +471,7 @@ namespace Microsoft.IdentityModel.Logging
         /// <param name="callbackUnsafe">A callback function to log the security artifact without scrubbing.</param>
         /// <returns>An argument marked as SecurityArtifact.</returns>
         /// <exception cref="ArgumentNullException">if <paramref name="callbackUnsafe"/> is null.</exception>
-        public static object MarkAsUnsafeOnlySecurityArtifact(object arg, Func<object, string> callbackUnsafe)
+        public static object MarkAsUnsafeSecurityArtifact(object arg, Func<object, string> callbackUnsafe)
         {
             return new SecurityArtifact(arg, SecurityArtifact.UnknownSafeTokenCallback, callbackUnsafe);
         }
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -389,7 +389,7 @@ namespace Microsoft.IdentityModel.Logging
                 // We may later add a further flag which would log a best effort scrubbing of an artifact. E.g. JsonWebToken tries to remove the signature
                 // in the current implementation. Another flag may be added in the future to allow this middle path but for now, LogCompleteSecurityArtifact
                 // must be logged to emit any token part (other than specific claim values).
-                return string.Format(CultureInfo.InvariantCulture, IdentityModelEventSource.HiddenArtifactString, arg?.GetType().ToString() ?? "Null");
+                return string.Format(CultureInfo.InvariantCulture, IdentityModelEventSource.HiddenSecurityArtifactString, arg?.GetType().ToString() ?? "Null");
             }
 
             // If it's not a ISafeLogSecurityArtifact then just return the object which will be converted to string.
--- a/src/Microsoft.IdentityModel.Logging/LogHelper.cs
+++ b/src/Microsoft.IdentityModel.Logging/LogHelper.cs
@@ -384,7 +384,17 @@ namespace Microsoft.IdentityModel.Logging
 
             if (IdentityModelEventSource.LogCompleteSecurityArtifact && arg is ISafeLogSecurityArtifact)
                 return (arg as ISafeLogSecurityArtifact).UnsafeToString();
+            else if (arg is ISafeLogSecurityArtifact)
+            {
+                // We may later add a further flag which would log a best effort scrubbing of an artifact. E.g. JsonWebToken tries to remove the signature
+                // in the current implementation. Another flag may be added in the future to allow this middle path but for now, LogCompleteSecurityArtifact
+                // must be logged to emit any token part (other than specific claim values).
+                return string.Format(CultureInfo.InvariantCulture, IdentityModelEventSource.HiddenArtifactString, arg?.GetType().ToString() ?? "Null");
+            }
 
+            // If it's not a ISafeLogSecurityArtifact then just return the object which will be converted to string.
+            // It's possible a raw string will contain a security artifact and be exposed here but the alternative is to scrub all objects
+            // which defeats the purpose of the ShowPII flag.
             return arg;
         }
 
@@ -424,9 +434,46 @@ namespace Microsoft.IdentityModel.Logging
         /// <param name="arg">A log message argument to be marked as SecurityArtifact.</param>
         /// <param name="callback">A callback function to log the security artifact safely.</param>
         /// <returns>An argument marked as SecurityArtifact.</returns>
+        /// <remarks>
+        /// Since even the payload may sometimes contain security artifacts, naïve disarm algorithms such as removing signatures
+        /// will not work. For now the <paramref name="callback"/> will only be leveraged if
+        /// <see cref="IdentityModelEventSource.LogCompleteSecurityArtifact"/> is set and no unsafe callback is provided. Future changes
+        /// may introduce a support for best effort disarm logging.
+        /// </remarks>
         public static object MarkAsSecurityArtifact(object arg, Func<object, string> callback)
         {
             return new SecurityArtifact(arg, callback);
+        }
+
+        /// <summary>
+        /// Marks a log message argument (<paramref name="arg"/>) as SecurityArtifact.
+        /// </summary>
+        /// <param name="arg">A log message argument to be marked as SecurityArtifact.</param>
+        /// <param name="callback">A callback function to log the security artifact safely.</param>
+        /// <param name="callbackUnsafe">A callback function to log the security artifact without scrubbing.</param>
+        /// <returns>An argument marked as SecurityArtifact.</returns>
+        /// <exception cref="ArgumentNullException">if <paramref name="callback"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="callbackUnsafe"/> is null.</exception>
+        /// <remarks>
+        /// Since even the payload may sometimes contain security artifacts, naïve disarm algorithms such as removing signatures
+        /// will not work. For now the <paramref name="callback"/> is currently unused. Future changes
+        /// may introduce a support for best effort disarm logging which will leverage <paramref name="callback"/>.
+        /// </remarks>
+        public static object MarkAsSecurityArtifact(object arg, Func<object, string> callback, Func<object, string> callbackUnsafe)
+        {
+            return new SecurityArtifact(arg, callback, callbackUnsafe);
+        }
+
+        /// <summary>
+        /// Marks a log message argument (<paramref name="arg"/>) as SecurityArtifact.
+        /// </summary>
+        /// <param name="arg">A log message argument to be marked as SecurityArtifact.</param>
+        /// <param name="callbackUnsafe">A callback function to log the security artifact without scrubbing.</param>
+        /// <returns>An argument marked as SecurityArtifact.</returns>
+        /// <exception cref="ArgumentNullException">if <paramref name="callbackUnsafe"/> is null.</exception>
+        public static object MarkAsUnsafeOnlySecurityArtifact(object arg, Func<object, string> callbackUnsafe)
+        {
+            return new SecurityArtifact(arg, SecurityArtifact.UnknownSafeTokenCallback, callbackUnsafe);
         }
 
         /// <summary>
--- a/src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
+++ b/src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
@@ -14,44 +14,100 @@ namespace Microsoft.IdentityModel.Logging
     /// </remarks>
     internal struct SecurityArtifact : ISafeLogSecurityArtifact
     {
+        private const string _scrubbedArtifact = "#ScrubbedArtifact#";
+
         /// <summary>
         /// Argument wrapped with a <see cref="SecurityArtifact"/> structure is considered as SecurityArtifact in the message logging process.
         /// </summary>
         private object Argument { get; set; }
 
         /// <summary>
-        /// The ToString callback delegate that return a disarmed SecurityArtifact.
+        /// The ToString callback delegate that returns a disarmed SecurityArtifact.
         /// </summary>
-        private Func<object, string> _callback;
+        private Func<object, string> _disarmCallback;
+
+        /// <summary>
+        /// The ToString callback delegate that returns an unscrubbed SecurityArtifact.
+        /// </summary>
+        private Func<object, string> _callbackUnsafe;
+
+        /// <summary>
+        /// Creates an instance of <see cref="SecurityArtifact"/> that wraps the <paramref name="argument"/>.
+        /// </summary>
+        /// <param name="argument">An argument that is considered as SecurityArtifact.</param>
+        /// <param name="toStringCallback">A callback used to disarm the token.</param>
+        /// <remarks>
+        /// Since even the payload may sometimes contain security artifacts, naïve disarm algorithms such as removing signatures
+        /// will not work. For now the <paramref name="toStringCallback"/> will only be leveraged if
+        /// <see cref="IdentityModelEventSource.LogCompleteSecurityArtifact"/> is set and no unsafe callback is provided. Future changes
+        /// may introduce a support for best effort disarm logging.
+        /// </remarks>
+        public SecurityArtifact(object argument, Func<object, string> toStringCallback)
+        {
+            Argument = argument;
+            _disarmCallback = toStringCallback;
+        }
 
         /// <summary>
         /// Creates an instance of <see cref="SecurityArtifact"/> that wraps the <paramref name="argument"/>.
         /// </summary>
         /// <param name="argument">An argument that is considered as SecurityArtifact.</param>
         /// <param name="toStringCallback">A ToString callback.</param>
-        /// <exception cref="ArgumentNullException">if <paramref name="argument"/> is null.</exception>
-        public SecurityArtifact(object argument, Func<object, string> toStringCallback)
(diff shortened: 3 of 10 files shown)


TimHannMSFT:
@keegan-caruso raised a good point offline, this will be cherry picked into Wilson 7 once completed.

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
```diff
@@ -45,6 +45,11 @@ private IdentityModelEventSource()
         /// </summary>
         public static string HiddenPIIString { get; } = "[PII of type '{0}' is hidden. For more details, see https://aka.ms/IdentityModel/PII.]";
 
+        /// <summary>
+        /// String that is used in place of any arguments to log messages if the 'LogCompleteSecurityArtifact' flag is set to false.
+        /// </summary>
+        public static string HiddenArtifactString { get; } = "[Security Artifact of type '{0}' is hidden. For more details, see https://aka.ms/IdentityModel/SecurityArtifactLogging.]";
```
keegan-caruso: it is incredibly minor, but consider including a space after the link and after the period. I think that the period will be included in the link and wont work.

test/Microsoft.IdentityModel.Logging.Tests/LogHelperTests.cs
```diff
@@ -0,0 +1,402 @@
...
+
+namespace Microsoft.IdentityModel.Logging.Tests
+{
+    using System;
+    using Microsoft.IdentityModel.JsonWebTokens;
+    using Microsoft.IdentityModel.Logging;
+    using Microsoft.IdentityModel.Tokens;
+    using Xunit;
+
+    [Collection("Relying on ShowPII and LogLogCompleteSecurityArtifact")]
```
keegan-caruso: minor typo "LogLog"

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
TimHannMSFT: Here I'm matching existing style (see `HiddenPIIString`) Seems like visual studio link generation and github email both don't include the period in the link and it looks more readable to my eye so leaving for now. Feel free to open a new issue and we can address outside this PR if you think it will cause problems.

keegan-caruso (approved):
just minor nits, LGTM

test/Microsoft.IdentityModel.Logging.Tests/LogHelperTests.cs
TimHannMSFT: good catch, thanks, fixed 06473b3ad7c0e462cc7aaa1eb174c3ece68ee370

TimHannMSFT:
> @keegan-caruso raised a good point offline, this will be cherry picked into Wilson 7 once completed.

My knowledge is out of date, `dev` is Wilson 7 now, should be cherry picked into `dev6x`

henrik-me (approved):
:shipit:

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -14,44 +14,100 @@ namespace Microsoft.IdentityModel.Logging
...
+        /// <summary>
+        /// Creates an instance of <see cref="SecurityArtifact"/> that wraps the <paramref name="argument"/>.
+        /// </summary>
+        /// <param name="argument">An argument that is considered as SecurityArtifact.</param>
+        /// <param name="toStringCallback">A ToString callback.</param>
+        /// <param name="toStringCallbackUnsafe">A ToString callback which will return the unscrubbed artifact.</param>
+        /// <exception cref="ArgumentNullException">if <paramref name="toStringCallback"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="toStringCallbackUnsafe"/> is null.</exception>
+        /// <remarks>
+        /// Since even the payload may sometimes contain security artifacts, naïve disarm algorithms such as removing signatures
```
jennyf19: Looks like there is something wrong with this comment: naÃ¯ve

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -14,44 +14,100 @@ namespace Microsoft.IdentityModel.Logging
...
         /// </summary>
-        /// <returns><c>Null</c> if the <see cref="Argument"/> is <see langword="null"/>, otherwise calls <see cref="System.ValueType.ToString()"/> method of the <see cref="Argument"/>.</returns>
+        /// <returns><c>Null</c> if the <see cref="Argument"/> is <see langword="null"/>, otherwise calls the provided safe callback on <see cref="Argument"/>.</returns>
         public override string ToString()
         {
-            if (_callback != null && Argument != null)
-                return _callback(Argument);
-
-            return Argument?.ToString() ?? "null";
+            // Defense in depth, callers should never provide a null callback but publicly exposed method in LogHelper referencing this didn't check.
```
jennyf19: this comment doesn't make sense.

test/Microsoft.IdentityModel.Logging.Tests/LogHelperTests.cs
```diff
@@ -0,0 +1,404 @@
...
+{
+    using System;
+    using Microsoft.IdentityModel.Logging;
+    using Microsoft.IdentityModel.Tokens;
+    using Xunit;
+
+    // Leveraging Xunit collections: https://xunit.net/docs/running-tests-in-parallel#parallelism-in-test-frameworks
+    // Currently this does nothing but if we face issues with tests colliding trying to access ShowPII and LogCompleteSecurityArtifact
+    // this collection can be used to prevent race conditions.
+    [Collection("Relying on ShowPII and LogCompleteSecurityArtifact")]
```
jennyf19, line 14: why wouldn't services also have a similar issue in production? I guess I'm not following what the issue is.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -14,44 +14,100 @@ namespace Microsoft.IdentityModel.Logging
     /// </remarks>
     internal struct SecurityArtifact : ISafeLogSecurityArtifact
     {
+        private const string _scrubbedArtifact = "#ScrubbedArtifact#";
```
jennyf19, line 17: nit; Underscore is not needed.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -14,44 +14,100 @@ namespace Microsoft.IdentityModel.Logging
...
         /// Argument wrapped with a <see cref="SecurityArtifact"/> structure is considered as SecurityArtifact in the message logging process.
         /// </summary>
         private object Argument { get; set; }
 
         /// <summary>
-        /// The ToString callback delegate that return a disarmed SecurityArtifact.
+        /// The ToString callback delegate that returns a disarmed SecurityArtifact.
+        /// </summary>
+        private Func<object, string> _disarmCallback;
+
```
jennyf19: mark as readonly

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -14,44 +14,100 @@ namespace Microsoft.IdentityModel.Logging
...
+        /// <summary>
+        /// The ToString callback delegate that returns an unscrubbed SecurityArtifact.
         /// </summary>
-        private Func<object, string> _callback;
+        private Func<object, string> _callbackUnsafe;
```
jennyf19: mark as readonly

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -384,7 +384,17 @@ private static object SanitizeSecurityArtifact(object arg)
 
             if (IdentityModelEventSource.LogCompleteSecurityArtifact && arg is ISafeLogSecurityArtifact)
                 return (arg as ISafeLogSecurityArtifact).UnsafeToString();
+            else if (arg is ISafeLogSecurityArtifact)
+            {
+                // We may later add a further flag which would log a best effort scrubbing of an artifact. E.g. JsonWebToken tries to remove the signature
+                // in the current implementation. Another flag may be added in the future to allow this middle path but for now, LogCompleteSecurityArtifact
+                // must be logged to emit any token part (other than specific claim values).
+                return string.Format(CultureInfo.InvariantCulture, IdentityModelEventSource.HiddenArtifactString, arg?.GetType().ToString() ?? "Null");
```
jennyf19: not seeing tests for this condition.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -14,44 +14,100 @@ namespace Microsoft.IdentityModel.Logging
...
+            _disarmCallback = toStringCallback;
+            _callbackUnsafe = toStringCallbackUnsafe;
+        }
+
+        /// <summary>
+        /// A dummy callback which can be leveraged to return a standard scrubbed token in the case where expected token is unknown.
+        /// </summary>
+        /// <param name="_">Ignored token.</param>
+        /// <returns>The standard scrubbed token string.</returns>
+        public static string UnknownSafeTokenCallback(object _)
```
jennyf19, line 74: is there test coverage for this?

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -14,44 +14,100 @@ namespace Microsoft.IdentityModel.Logging
...
         }
 
         // <inheritdoc/>
         public string UnsafeToString()
         {
-            return Argument?.ToString() ?? "null";
+            if (_callbackUnsafe == null)
+                return ToString();
+            if (Argument == null)
+                return "null";
```
jennyf19: do we cover the return "null" test case?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
```diff
@@ -269,7 +269,9 @@ private ClaimsPrincipal ValidateToken(Saml2SecurityToken samlToken, string token
                 identity.BootstrapContext = samlToken.Assertion.CanonicalString;
 
             if (LogHelper.IsEnabled(EventLogLevel.Informational))
-                LogHelper.LogInformation(TokenLogMessages.IDX10241, token);
+                LogHelper.LogInformation(
```
jennyf19, line 272: names can be simplified (ex. `LogHelper` can be removed) and do we have tests verifying this behavior? same w/other similar areas

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
brentschmaltz: Suggest a name change to HiddenSecurityArtifactString

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -14,44 +14,100 @@ namespace Microsoft.IdentityModel.Logging
...
+        /// <exception cref="ArgumentNullException">if <paramref name="toStringCallbackUnsafe"/> is null.</exception>
+        /// <remarks>
+        /// Since even the payload may sometimes contain security artifacts, naïve disarm algorithms such as removing signatures
+        /// will not work. For now the <paramref name="toStringCallback"/> is currently unused. Future changes
+        /// may introduce a support for best effort disarm logging which will leverage <paramref name="toStringCallback"/>.
+        /// </remarks>
+        public SecurityArtifact(object argument, Func<object, string> toStringCallback, Func<object, string> toStringCallbackUnsafe)
+        {
+            if (toStringCallback == null)
+                throw new ArgumentNullException(nameof(toStringCallback));
```
brentschmaltz: We should be logging these exceptions using LogHelper.LogException.

Do we need to throw, could we have a 'no-opt' callback that just logs that the callback(s) was null and log that?
In SecurityArtifact.ToString() and UnsafeToString() if _disarmCallback or == null

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -14,44 +14,100 @@ namespace Microsoft.IdentityModel.Logging
...
+        /// <exception cref="ArgumentNullException">if <paramref name="toStringCallback"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="toStringCallbackUnsafe"/> is null.</exception>
+        /// <remarks>
+        /// Since even the payload may sometimes contain security artifacts, naïve disarm algorithms such as removing signatures
+        /// will not work. For now the <paramref name="toStringCallback"/> is currently unused. Future changes
+        /// may introduce a support for best effort disarm logging which will leverage <paramref name="toStringCallback"/>.
+        /// </remarks>
+        public SecurityArtifact(object argument, Func<object, string> toStringCallback, Func<object, string> toStringCallbackUnsafe)
+        {
+            if (toStringCallback == null)
```
brentschmaltz: We usually write
_ = toStringCallback ?? LogHelper.Log......;

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
```diff
@@ -14,44 +14,100 @@ namespace Microsoft.IdentityModel.Logging
...
+        /// Since even the payload may sometimes contain security artifacts, naïve disarm algorithms such as removing signatures
+        /// will not work. For now the <paramref name="toStringCallback"/> will only be leveraged if
+        /// <see cref="IdentityModelEventSource.LogCompleteSecurityArtifact"/> is set and no unsafe callback is provided. Future changes
+        /// may introduce a support for best effort disarm logging.
+        /// </remarks>
         public SecurityArtifact(object argument, Func<object, string> toStringCallback)
         {
             Argument = argument;
-            _callback = toStringCallback;
+            _disarmCallback = toStringCallback;
```
brentschmaltz, line 48: We throw below when toStringCallback is null but not here.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
brentschmaltz: This comment is specific to JWT tokens, whereas a SecurityArtifact is general.
We may want to write: For example, a JsonWebToken payload may contain security artifacts as should be hidden.

TimHannMSFT: Hmm not sure what the issue is here. Happy to address in a future PR to clarify the comment. Basically there's a behavior change related to not logging scrubbed tokens I'm trying to capture here. Not sure what's getting lost in translation though so will leave for a future commit.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
TimHannMSFT: Not sure how to clarify. There's a public API that doesn't validate the inputs, we can't add validation now otherwise that's a breaking change.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
TimHannMSFT: This is the style of the library at large, private variables are prepended with an underscore. Looks like we do both, so it's not consistent. Given the library doesn't have a consistent style there's nothing obvious to snap to, leaving for now. Also private so easy to change later.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
TimHannMSFT: LogSecurityArtifactTest covers several of these cases.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.cs
TimHannMSFT: Library doesn't have per log line testing. I've filled the testing gap for LogHelper to some extent but adding per log tests is outside the scope of this PR.

src/Microsoft.IdentityModel.Logging/IdentityModelEventSource.cs
TimHannMSFT: Sure, will include in next commit

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
TimHannMSFT: This is because this is linked to a public method so we can't change here. Ideally I'd expect us to check inputs on public methods. I did a best effort fix for the overload but perhaps keeping behavior consistent is more important here.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
TimHannMSFT: Added more clarity that this is for the JWT case. It's meant as an example of a naive algo

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
TimHannMSFT: I've removed the input checks to match the other methods which are already exposed publicly

test/Microsoft.IdentityModel.Logging.Tests/LogHelperTests.cs
TimHannMSFT: It's about race conditions in unit tests.

Imagine two unit tests are executing in parallel and both modifying the static properties controlling how logs are generated. Those two unit tests will not be able to reliably predict the format and so will be flaky. Forcing them to run serially will make them more reliable. A better design still is not not rely on static variables but that's already pretty baked into Wilson unfortunately.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
TimHannMSFT: It was previously un-reachable but I've removed input validation to allow for this case and added a unit test which leverages this.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
TimHannMSFT: I've updated to make this more allowing of bad instrumentation. If nothing's passed we'll use a default callback that just prints a generic string, "#ScrubbedArtifact#"

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
TimHannMSFT: Added a test for this case and refactored so that unsafe and safe callbacks use the same flow for this

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -424,11 +434,48 @@ public static object MarkAsNonPII(object arg)
...
+        }
+
+        /// <summary>
+        /// Marks a log message argument (<paramref name="arg"/>) as SecurityArtifact.
+        /// </summary>
+        /// <param name="arg">A log message argument to be marked as SecurityArtifact.</param>
+        /// <param name="callbackUnsafe">A callback function to log the security artifact without scrubbing.</param>
+        /// <returns>An argument marked as SecurityArtifact.</returns>
+        /// <exception cref="ArgumentNullException">if <paramref name="callbackUnsafe"/> is null.</exception>
+        public static object MarkAsUnsafeOnlySecurityArtifact(object arg, Func<object, string> callbackUnsafe)
```
brentschmaltz: Suggest name change: MarkAsUnsafeSecurityArtifact.

Should we consider returning ISafeLogSecurityArtifact returning an object is just awkard.

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
brentschmaltz, line 74: Suggest name change using SecurityArtifact rather than Token: SafeSecurityArtifactCallback.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
brentschmaltz, line 17: Should we have some info here, similar to our PII string with a link to a wiki explaining how to control this feature?

TimHannMSFT: Wrong level of abstraction, that's only relevant in the logging library.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
TimHannMSFT: This is part of an internal struct so we can change in the future.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
brentschmaltz: What will users see in the log?

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
brentschmaltz: @TimHannMSFT are we not going to use LogHelper to log the error?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
TimHannMSFT: Too late on the return value, this is already the contract the others used so I'm being consistent. I generally agree though.

I can update the name but I'm not sure it's an important distinction.

src/Microsoft.IdentityModel.Logging/SecurityArtifact.cs
brentschmaltz: Why not change it now?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
brentschmaltz: No reason to check if arg is null as that is checked on line 382

TimHannMSFT: technically it's the toString return value on the type not the arg. I agree this is unlikely to ever trigger though. This was an existing behavior, we can address outside this PR if important.

brentschmaltz: What i am saying is 'arg' will never get here if null, every object has a type, ToString() is on every type.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
brentschmaltz: "UnsafeOnly" is confusing to me.

PR #2340, opened 2023-09-29
Decouple JsonElements from JsonDocument.

brentschmaltz:
In some multi threaded situations the incorrect JsonElement could have been returned.
Each JsonProperty is stored in a dictionary as a <name, JsonElement>.
First cloning the JsonElement to detach from the JsonDocument.

The JsonDocument is now correctly disposed by the creator.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
@@ -25,8 +25,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
         internal JsonClaimSet(JsonDocument jsonDocument)
         {
-            RootElement = jsonDocument.RootElement.Clone();
-            jsonDocument.Dispose();
+            foreach (JsonProperty jsonProperty in jsonDocument.RootElement.EnumerateObject())
+                Elements[jsonProperty.Name] = jsonProperty.Value.Clone();
         }
 
         internal JsonClaimSet(byte[] jsonBytes) : this(JsonDocument.Parse(jsonBytes))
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -25,8 +25,8 @@ internal class JsonClaimSet
 
         internal JsonClaimSet(JsonDocument jsonDocument)
         {
-            RootElement = jsonDocument.RootElement.Clone();
-            jsonDocument.Dispose();
```
keegan-caruso, line 29: can you elaborate on why the Dispose is not needed anymore?

keegan-caruso: oh I see, they get disposed where they are created now, resolving

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
```diff
@@ -43,6 +44,117 @@ public class JsonWebTokenTests
             new Claim("dateTimeIso8061", dateTime.ToUniversalTime().ToString("o", CultureInfo.InvariantCulture), ClaimValueTypes.DateTime, "LOCAL AUTHORITY", "LOCAL AUTHORITY"),
         };
 
+#if !NET452
```
jennyf19, line 47: NET45? not NET452?

jennyf19 (approved):

keegan-caruso (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
brentschmaltz: @jennyf19 the test environment uses 452 as xunit is not available for 451.

PR #2343, opened 2023-10-02
benchmarks conditionally targets net8

### LINKED ISSUE #2342: [Bug] Microsoft.IdentityModel.Benchmarks should conditionally target .net8.
Currently Microsoft.IdentityModel.Benchmarks targets .net8 which causes the current release of VS to fail building.
We should use the logic in targets.props.
see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/9cd930c766b9840ab277b5fff9a71f778159e0be/build/targets.props#L3

westin-m:


jennyf19 (approved):
:shipit: thanks @westin-m

brentschmaltz (approved):

PR #2345, opened 2023-10-02
Add baseline package validation

westin-m:
Introduce the baseline package validator through msbuild. Remove ApiCompat props, powershell script, and task in build .yml file as this does that work. 

more info on the tool: https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/baseline-version-validator


build/common.props
```diff
@@ -35,6 +35,12 @@
     <LangVersion>11</LangVersion>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <PackageVersion>7.1.0</PackageVersion>
+    <EnablePackageValidation>true</EnablePackageValidation>
+    <PackageValidationBaselineVersion>7.0.2</PackageValidationBaselineVersion>
```
jmprieur: Shouldn't the baseline be 7.0.0? Otherwise how to find breaking changes compared to a major version

jmprieur (dismissed):

test/Microsoft.IdentityModel.Tokens.Tests/SignatureProviderTests.cs
```diff
@@ -939,14 +940,13 @@ private void SymmetricProviderDispose(string testId, SecurityKey securityKey, st
...
-        // Excluding OSX as SignatureTampering test is slow on OSX (~6 minutes)
-        // especially tests with IDs RS256 and ES256
-        [PlatformSpecific(TestPlatforms.Windows | TestPlatforms.Linux)]
-#endif
+
         [Theory, MemberData(nameof(SignatureTheoryData))]
         public void SignatureTampering(SignatureProviderTheoryData theoryData)
         {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                throw new PlatformNotSupportedException("OSX is excluded as the SignatureTampering test is slow (~6 minutes).");
```
keegan-caruso: this doesn't look right, before OSX platforms would skip, now they will fail on this test?

keegan-caruso (approved):

PR #2350, opened 2023-10-04
update build config version

westin-m:


buildConfiguration.xml
```diff
@@ -2,7 +2,7 @@
   <dotnetArchitecture>x64</dotnetArchitecture>
   <nugetVersion>3.5.0-rc-1285</nugetVersion>
   <runtimes>net45,net461,netstandard2.0</runtimes>
-  <assemblyVersion>6.32.4</assemblyVersion>
+  <assemblyVersion>6.33.0</assemblyVersion>
```
brentschmaltz: 6.33.0 has been released, should this be 6.33.1?

sruke: The patch part of the assemblyVersion is usually one higher than the released version. In this case, v6.33.0 was released so the assemblyVersion would be set to 6.33.1. This is done so that preview versions generated by our nightly builds are higher than the released version and can be shared for testing.

westin-m: fixed

westin-m: fixed

FuPingFranco (approved):
LGTM

PR #2362, opened 2023-10-15
Stop on error during packaging

sruke:
Throw an error if there was an error during packaging.

### DIFF
--- a/pack.ps1
+++ b/pack.ps1
@@ -69,12 +69,15 @@ foreach ($project in $buildConfiguration.SelectNodes("root/projects/src/project"
     $name = $project.name
     $projectPath = "$root\src\$name\$name.csproj"
 
+    Write-Host ">>> dotnet pack --no-build --no-restore -nodereuse:false -c $buildType -o $artifactsRoot -v q -s $projectPath"
     dotnet pack --no-build --no-restore -nodereuse:false -c $buildType -o $artifactsRoot -v q -s $projectPath
 
     if($LASTEXITCODE -ne 0)
     {
         throw "Error occurred while packaging project '$projectPath'!"
     }
+
+    Write-Host "Packaging for project '$projectPath' completed successfully."
 }
 
 Write-Host "============================"
--- a/pack.ps1
+++ b/pack.ps1
@@ -63,11 +63,17 @@ Write-Host "releaseVersion: " $releaseVersion;
 Write-Host "nugetPreview:   " $nugetPreview;
 
 CreateArtifactsFolder($artifactsRoot);
-foreach($project in $buildConfiguration.SelectNodes("root/projects/src/project"))
+
+foreach ($project in $buildConfiguration.SelectNodes("root/projects/src/project"))
 {
-	$name = $project.name;
-	Write-Host ">>> Start-Process -wait -NoNewWindow $dotnetexe 'pack' --no-build --no-restore -nodereuse:false -c $buildType -o $artifactsRoot -v q -s $root\src\$name\$name.csproj"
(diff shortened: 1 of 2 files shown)


pack.ps1
```diff
@@ -63,11 +63,21 @@ Write-Host "releaseVersion: " $releaseVersion;
...
 CreateArtifactsFolder($artifactsRoot);
-foreach($project in $buildConfiguration.SelectNodes("root/projects/src/project"))
+
+foreach ($project in $buildConfiguration.SelectNodes("root/projects/src/project"))
 {
-	$name = $project.name;
-	Write-Host ">>> Start-Process -wait -NoNewWindow $dotnetexe 'pack' --no-build --no-restore -nodereuse:false -c $buildType -o $artifactsRoot -v q -s $root\src\$name\$name.csproj"
-	Start-Process -Wait -PassThru -NoNewWindow $dotnetexe "pack --no-build --no-restore -nodereuse:false -c $buildType -o $artifactsRoot -v q -s $root\src\$name\$name.csproj"
+    $name = $project.name
+    $projectPath = "$root\src\$name\$name.csproj"
```
keegan-caruso: this won't work on linux/mac right?

How about [System.IO.Path]::Combine instead?

src/Microsoft.IdentityModel.Tokens/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/diagnostic-ids -->
+<Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:Microsoft.IdentityModel.Tokens.JsonWebKeyConverter.ConvertFromECDsaSecurityKey(Microsoft.IdentityModel.Tokens.ECDsaSecurityKey)</Target>
```
keegan-caruso, line 6: oh I think I see what the tool is trying to catch.

net461 implements netstandard 2.0

So this is a bit confusing. Is there any reason ConvertFromECDsaSecurityKey cannot also be included in net461?

pack.ps1
sruke: My understanding is this script is not used by Wilson Linux builds.

src/Microsoft.IdentityModel.Tokens/CompatibilitySuppressions.xml
sruke: I'm not aware of the history behind this decision. A [GitHub Issue](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2363) has been opened to fail gracefully if a API attempts to invoke this method on net461.

brentschmaltz: @keegan-caruso even though the api is available in netstandard 2.0, we found 461 and 462 did not support this api.
https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.eccurve?view=netstandard-2.0
Some types are not supported.

keegan-caruso: interesting!

brentschmaltz: @keegan-caruso i wrote 472 above, but i meant 462.
What puzzles me is why only an error for 461 and not 462?

brentschmaltz (approved):

pack.ps1
sruke: Updated to use Path.Combine.

keegan-caruso (approved):

PR #2376, opened 2023-10-21
Use source-generated Regex in JwtTokenUtilities

### LINKED ISSUE #2036: [Feature Request] Use GeneratedRegex in JwtTokenUtilities
When publishing an app for NativeAOT (see #2035), IL generation at runtime is disabled.

One usage of IL generation at runtime is `new Regex(regex, RegexOptions.Compiled)`. At startup, RegexOptions.Compiled instructs the Regex engine to generate IL given the regular expression, and use the generated IL, instead of interpreting the expression every time it is executed.

When you use a "Compiled" Regex in NativeAOT, the Regex engine ignores the `Compiled` option and instead falls back to interpreting the expression at runtime. This means your app runs slower when on NativeAOT.

The solution to this issue is to use the new Regex Source Generator:

* https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators
* https://devblogs.microsoft.com/dotnet/regular-expression-improvements-in-dotnet-7/#source-generation

However, this feature is only available when targeting `net7.0` and above. So we would need the following:

1. We would need to target `net7.0` in `Microsoft.IdentityModel.JsonWebTokens.csproj`.
    - This means updating any CI scripts and build machines to use the .NET 7 SDK.
2. JwtTokenUtilities would look like this:

```C#
    public partial class JwtTokenUtilities
    {
        private const int RegexMatchTimeoutMilliseconds = 100;
        /// <summary>
        /// Regex that is used to figure out if a token is in JWS format.
        /// </summary>
        public static Regex RegexJws = CreateJwsRegex();

        /// <summary>
        /// Regex that is used to figure out if a token is in JWE format.
        /// </summary>
        public static Regex RegexJwe = CreateJweRegex();

#if NET7_0_OR_GREATER
        [GeneratedRegex(JwtConstants.JsonCompactSerializationRegex, RegexOptions.CultureInvariant, RegexMatchTimeoutMilliseconds)]
        private static partial Regex CreateJwsRegex();
        [GeneratedRegex(JwtConstants.JweCompactSerializationRegex, RegexOptions.CultureInvariant, RegexMatchTimeoutMilliseconds)]
        private static partial Regex CreateJweRegex();
#else
        private static Regex CreateJwsRegex() => new Regex(JwtConstants.JsonCompactSerializationRegex, RegexOptions.Compiled | RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(RegexMatchTimeoutMilliseconds));
        private static Regex CreateJweRegex() => new Regex(JwtConstants.JweCompactSerializationRegex, RegexOptions.Compiled | RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(RegexMatchTimeoutMilliseconds));
#endif
```

Note that there are benefits to using the source generator even when not in NativeAOT mode. For example, your app starts faster because it doesn't need to generate IL at runtime.

Live with slower NativeAOT applications.

**Additional context**
The following benchmark shows `JsonWebTokenHandler.CanReadToken` performance is 2-3x faster on NativeAOT when using the source generated Regex:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net7.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="\git\azure-activedirectory-identitymodel-extensions-for-dotnet\src\Microsoft.IdentityModel.JsonWebTokens\Microsoft.IdentityModel.JsonWebTokens.csproj" />
    <PackageReference Include="BenchmarkDotNet" Version="0.13.5" />
  </ItemGroup>

</Project>
```

```C#
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Configs;
using BenchmarkDotNet.Jobs;
using BenchmarkDotNet.Running;
using Perfolizer.Horology;

var job = Job.Default
                    .WithWarmupCount(1) // 1 warmup is enough for our purpose
                    .WithIterationTime(TimeInterval.FromMilliseconds(250)) // the default is 0.5s per iteration, which is slighlty too much for us
                    .WithMinIterationCount(15)
                    .WithMaxIterationCount(20) // we don't want to run more that 20 iterations
                    .DontEnforcePowerPlan(); // make sure BDN does not try to enforce High Performance power plan on Windows

var config = DefaultConfig.Instance
    .AddJob(job.AsDefault()); // tell BDN that this are our default settings

BenchmarkSwitcher
    .FromAssembly(typeof(Program).Assembly)
    .Run(args, config);


public class JsonWebTokenHandlerBenchmark
{
    const string jwsToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1bmlxdWVfbmFtZSI6Ik15VGVzdFVzZXIiLCJzdWIiOiJNeVRlc3RVc2VyIiwianRpIjoiMWVhZjQ4NzYiLCJzY29wZSI6Im15YXBpOnNlY3JldHMiLCJhdWQiOlsiaHR0cDovL2xvY2FsaG9zdDo1MjA4MCIsImh0dHBzOi8vbG9jYWxob3N0OjQ0MzA4IiwiaHR0cDovL2xvY2FsaG9zdDo1Mjk0IiwiaHR0cHM6Ly9sb2NhbGhvc3Q6NzE3MyJdLCJuYmYiOjE2Nzg5OTg2MjUsImV4cCI6MTY4Njk0NzQyNSwiaWF0IjoxNjc4OTk4NjI2LCJpc3MiOiJkb3RuZXQtdXNlci1qd3RzIn0.Ol6TuW6OgkaLcuFWx5O4JGcGwqdG9IlvA84roUQPhc8";
    const string jweToken = "eyJhbGciOiJSUzI1NiIsImtpZCI6IlJzYVNlY3VyaXR5S2V5XzIwNDgiLCJ0eXAiOiJKV1QifQ.eyJlbWFpbCI6IkJvYkBjb250b3NvLmNvbSIsImdpdmVuX25hbWUiOiJCb2IiLCJpc3MiOiJodHRwOi8vRGVmYXVsdC5Jc3N1ZXIuY29tIiwiYXVkIjoiaHR0cDovL0RlZmF1bHQuQXVkaWVuY2UuY29tIiwibmJmIjoiMTQ4OTc3NTYxNyIsImV4cCI6IjE2MTYwMDYwMTcifQ.GcIi6FGp1JS5VF70_ULa8g6GTRos9Y7rUZvPAo4hm10bBNfGhdd5uXgsJspiQzS8vwJQyPlq8a_BpL9TVKQyFIRQMnoZWe90htmNWszNYbd7zbLJZ9AuiDqDzqzomEmgcfkIrJ0VfbER57U46XPnUZQNng2XgMXrXmIKUqEph_vLGXYRQ4ndfwtRrR6BxQFd1PS1T5KpEoUTusI4VEsMcutzfXUygLDiRKIcnLFA0kQpeoHllO4Nb_Sxv63GCb0d1076FfSEYtyRxF4YSCz1In-ee5dwEK8Mw3nHscu-1hn0Fe98RBs-4OrUzI0WcV8mq9IIB3i-U-CqCJEP_hVCiA";

    private readonly JsonWebTokenHandler handler = new JsonWebTokenHandler();

    [Benchmark]
    public bool JWS() => handler.CanReadToken(jwsToken);

    [Benchmark]
    public bool JWE() => handler.CanReadToken(jweToken);
}
```

### Existing Code Normal CoreCLR (not AOT'd)
`dotnet run -c Release -f net7.0`

| Method |     Mean |    Error |  StdDev |
|------- |---------:|---------:|--------:|
|    JWS | 662.5 ns | 10.21 ns | 9.05 ns |
|    JWE | 861.8 ns |  5.27 ns | 4.12 ns |

### Existing Code Publish AOT
`dotnet run -c Release -f net7.0 --runtimes nativeaot7.0`

| Method |     Mean |     Error |    StdDev |
|------- |---------:|----------:|----------:|
|    JWS | 1.798 us | 0.0222 us | 0.0197 us |
|    JWE | 2.376 us | 0.0106 us | 0.0099 us |

* Note the change in time units from nanoseconds to microseconds.

### Proposed Code Normal CoreCLR (not AOT'd)
`dotnet run -c Release -f net7.0`

| Method |     Mean |    Error |   StdDev |
|------- |---------:|---------:|---------:|
|    JWS | 598.0 ns |  6.53 ns |  6.11 ns |
|    JWE | 788.3 ns | 11.16 ns | 10.44 ns |

### Proposed Code Publish AOT
`dotnet run -c Release -f net7.0 --runtimes nativeaot7.0`

| Method |     Mean |    Error |   StdDev |
|------- |---------:|---------:|---------:|
|    JWS | 558.8 ns | 19.37 ns | 22.31 ns |
|    JWE | 702.2 ns |  5.12 ns |  4.54 ns |

ghost:
Fixes #2036.

I agree with @eerhardt on the fact that the current implementation already uses Regex's, so it's easy to improve it by using source-generated Regex's. If, later, we want to further improve performance by changing the implementation, we're free to do so.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -22,7 +22,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// <summary>
     /// A class which contains useful methods for processing JWT tokens.
     /// </summary>
-    public class JwtTokenUtilities
+    public partial class JwtTokenUtilities
     {
         private const int _regexMatchTimeoutMilliseconds = 100;
         private const string _unrecognizedEncodedToken = "UnrecognizedEncodedToken";
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -24,17 +24,28 @@ namespace Microsoft.IdentityModel.JsonWebTokens
     /// </summary>
     public class JwtTokenUtilities
     {
+        private const int _regexMatchTimeoutMilliseconds = 100;
         private const string _unrecognizedEncodedToken = "UnrecognizedEncodedToken";
 
         /// <summary>
         /// Regex that is used to figure out if a token is in JWS format.
         /// </summary>
-        public static Regex RegexJws = new Regex(JwtConstants.JsonCompactSerializationRegex, RegexOptions.Compiled | RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(100));
+        public static Regex RegexJws = CreateJwsRegex();
 
         /// <summary>
         /// Regex that is used to figure out if a token is in JWE format.
         /// </summary>
-        public static Regex RegexJwe = new Regex(JwtConstants.JweCompactSerializationRegex, RegexOptions.Compiled | RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(100));
+        public static Regex RegexJwe = CreateJweRegex();
+
+#if NET7_0_OR_GREATER
+        [GeneratedRegex(JwtConstants.JsonCompactSerializationRegex, RegexOptions.CultureInvariant, _regexMatchTimeoutMilliseconds)]
+        private static partial Regex CreateJwsRegex();
+        [GeneratedRegex(JwtConstants.JweCompactSerializationRegex, RegexOptions.CultureInvariant, _regexMatchTimeoutMilliseconds)]
+        private static partial Regex CreateJweRegex();
+#else
+        private static Regex CreateJwsRegex() => new Regex(JwtConstants.JsonCompactSerializationRegex, RegexOptions.Compiled | RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(_regexMatchTimeoutMilliseconds));
+        private static Regex CreateJweRegex() => new Regex(JwtConstants.JweCompactSerializationRegex, RegexOptions.Compiled | RegexOptions.CultureInvariant, TimeSpan.FromMilliseconds(_regexMatchTimeoutMilliseconds));
+#endif
 
         internal static IList<string> DefaultHeaderParameters = new List<string>()
         {


eerhardt (approved):
LGTM. Thanks for contributing this.

cc @jennyf19 @brentschmaltz

eerhardt:
FYI - this looks to have saved ~1MB in our NativeAOT benchmark app:

Before:
image

After:
image

Thanks again for the contribution!

PR #2378, opened 2023-10-21
Add support for enum in JsonClaimSet

### LINKED ISSUE #1941: [Bug] Regression in parsing of (lower case) enum values between 6.22.0 and 6.23.0 of Microsoft.IdentityModel.JsonWebTokens
Microsoft.IdentityModel.JsonWebTokens 6.23.0

**Where is the issue?**
 * [x] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [ ] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [ ] S.IM.Tokens.Jwt
 * Other (please describe)

The app is in production and I have upgraded to a new version of Microsoft.IdentityModel.

**Repro**

```csharp
enum TokenType {
	  Unknown
	  User,
	  Application
}

jsonWebToken.TryGetPayloadValue<TokenType>("type", out var tokenType )
```

Input token would have 

```json
"type": "user"
```

**Expected behavior**
It should support case insensitive parsing of enum values like with 6.22.0

**Actual behavior**
It returns the default value of the enum because it can't match the value anymore.

**Possible solution**

Maybe one of the `JsonSerializerOptions` `*CaseInsensitive` options.

[Version diff](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/compare/6.22.0...6.23.0) seems like there was a switch to System.Text.Json for some of the claim handling, but enums were not considered.

ghost:
Fixes #1941.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
@@ -253,6 +253,10 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 if (long.TryParse(obj.ToString(), out long value))
                     return (T)(object)value;
             }
+            else if(typeof(T).IsEnum)
+            {
+                return (T)Enum.Parse(typeof(T), obj.ToString(), ignoreCase: true);
+            }
             else if (typeof(T) == typeof(string[]))
             {
                 if (objType == typeof(string))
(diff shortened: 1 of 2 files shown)


joegoldman2:

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs
```diff
@@ -253,6 +253,10 @@ internal T GetValue<T>(string key, bool throwEx, out bool found)
                 if (long.TryParse(obj.ToString(), out long value))
                     return (T)(object)value;
             }
+            else if(typeof(T).IsEnum)
```
brentschmaltz: I think enum should be lower on the stack, below ICollection<string>

joegoldman2: Fair enough considering that enums won't be used so often. Done.

brentschmaltz (approved):

PR #2381, opened 2023-10-22
Allow JSON deserialization into JsonWebKeySet

### LINKED ISSUE #2326: Unable to deserialize JSON into JsonWebKeySet
Currently, it is not possible to deserialize a JWK Set into `Microsoft.IdentityModel.Tokens.JsonWebKeySet` without additionnal configuration.

This is because the `Keys` property is readonly.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/eeaa61ba4887ded80875ed5659aa8a7c04b1da96/src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs#L82

Steps to reproduce:
```cs
using System.Net.Http.Json;

var client = new HttpClient();
var response = await client.GetAsync("https://login.microsoftonline.com/common/discovery/v2.0/keys");
var jsonWebKeySet = await response.Content.ReadFromJsonAsync<JsonWebKeySet>();
var count = jsonWebKeySet!.Keys.Count; // 0
```

**Known workarounds**

Before .NET 8, we can use `JsonWebKeySet.Create(string)` to create an instance of `JsonWebKeySet` from the raw content:

```cs
using System.Net.Http.Json;

var client = new HttpClient();
var response = await client.GetAsync("https://login.microsoftonline.com/common/discovery/v2.0/keys");
var raw = await response.Content.ReadAsStringAsync(); // An temporary string is allocated here :(
var jsonWebKeySet = JsonWebKeySet.Create(raw);
var count = jsonWebKeySet!.Keys.Count; // 5
```

With .NET 8, we can use the new [`JsonObjectCreationHandling`](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonobjectcreationhandling) introduced, this way:

```cs
using System.Net.Http.Json;
using System.Text.Json.Serialization;
using System.Text.Json;

var client = new HttpClient();
var response = await client.GetAsync("https://login.microsoftonline.com/common/discovery/v2.0/keys");
var jsonWebKeySet = await response.Content.ReadFromJsonAsync<JsonWebKeySet>(new JsonSerializerOptions
{
    PreferredObjectCreationHandling = JsonObjectCreationHandling.Populate
});
var count = jsonWebKeySet!.Keys.Count; // 5
```

**Suggested solution**
 
We can add the attribute `JsonObjectCreationHandlingAttribute` with the value `JsonObjectCreationHandling.Populate` directly on the `Keys` property to allow the deserialization without additional configuration.
This attribute has been introduced in .NET 8 so it will probably be necessary to condition it with `if NET8_0_OR_GREATER`.

ghost:
Fixes #2326.

We'll probably have to wait for .NET 8 before we can merge this PR.

### DIFF
--- a/test/Microsoft.IdentityModel.Tokens.Tests/JsonWebKeySetTests.cs
+++ b/test/Microsoft.IdentityModel.Tokens.Tests/JsonWebKeySetTests.cs
@@ -417,6 +417,15 @@ namespace Microsoft.IdentityModel.Tokens.Json.Tests
             }
         }
 
+        [Fact]
+        public void JsonDeserialize()
+        {
+            var json = DataSets.JsonWebKeySetEvoString;
+            var jsonWebKeys = System.Text.Json.JsonSerializer.Deserialize<JsonWebKeySet>(json);
+            Assert.NotNull(jsonWebKeys);
+            Assert.NotEmpty(jsonWebKeys.Keys);
+        }
+
         private static X509SecurityKey CreateX509SecurityKey(JsonWebKeySet webKeySet, int keyIndex)
         {
             var webKey = (webKeySet.Keys as List<JsonWebKey>)[keyIndex];
--- a/test/Microsoft.IdentityModel.Tokens.Tests/JsonWebKeySetTests.cs
+++ b/test/Microsoft.IdentityModel.Tokens.Tests/JsonWebKeySetTests.cs
@@ -417,6 +417,7 @@ namespace Microsoft.IdentityModel.Tokens.Json.Tests
             }
         }
 
+#if NET8_0_OR_GREATER
         [Fact]
         public void JsonDeserialize()
         {
@@ -425,6 +426,7 @@ namespace Microsoft.IdentityModel.Tokens.Json.Tests
             Assert.NotNull(jsonWebKeys);
             Assert.NotEmpty(jsonWebKeys.Keys);
         }
+#endif
 
         private static X509SecurityKey CreateX509SecurityKey(JsonWebKeySet webKeySet, int keyIndex)
         {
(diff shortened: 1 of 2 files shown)


test/Microsoft.IdentityModel.Tokens.Tests/JsonWebKeySetTests.cs
```diff
@@ -417,6 +417,15 @@ public static TheoryData<JsonWebKeySetTheoryData> GetSigningKeysTheoryData
             }
         }
 
+        [Fact]
+        public void JsonDeserialize()
```
joegoldman2, line 421: I'm not really sure what can be the right strategy to run this test only if the TFM is `.net8.0`. I'm open to suggestions :)

joegoldman2, line 421: I'm not really sure what can be the right strategy to run this test only if the TFM is `net8.0`. I'm open to suggestions :)

ghost, line 422: I'm not really sure what can be the right strategy to run this test only if the TFM is `net8.0` and above, apart from putting a `#if NET8_0_OR_GREATER` around the test. I'm open to suggestions :)

joegoldman2:
@jennyf19 @brentschmaltz this PR is ready to be reviewed.

test/Microsoft.IdentityModel.Tokens.Tests/JsonWebKeySetTests.cs
brentschmaltz: @joegoldman2 yes you can put NET_8_OR_GREATER

joegoldman2: Perfect, done!

brentschmaltz (approved):

PR #2389, opened 2023-10-26
update bug template with default tags

jennyf19:


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #2394, opened 2023-11-06
Add NotNullWhen attribute to Microsoft.IdentityModel.Tokens.CollectionUtilities.IsNullOrEmpty extension method

fabriziobagala:
Added `NotNullWhen` attribute to indicate that the parameter is not `null` when the method returns `false`.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/CollectionUtilities.cs
+++ b/src/Microsoft.IdentityModel.Tokens/CollectionUtilities.cs
@@ -17,7 +17,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// <typeparam name="T">The type of the <paramref name="enumerable"/>.</typeparam>
         /// <param name="enumerable">The <see cref="IEnumerable{T}"/> to be checked.</param>
         /// <returns>True if <paramref name="enumerable"/> is null or empty, false otherwise.</returns>
-        public static bool IsNullOrEmpty<T>(this IEnumerable<T> enumerable)
+        public static bool IsNullOrEmpty<T>([NotNullWhen(false)] this IEnumerable<T> enumerable)
         {
             return enumerable == null || !enumerable.Any();
         }


brentschmaltz:
@fabriziobagala i am wondering how this helps.
the return value is bool, which is not nullable.

joegoldman2:
@brentschmaltz The `NotNullWhen` attribute is used to provide hints to the C# compiler about the nullability of a reference type when a specific condition is met. This attribute is typically used in combination with nullable reference types to help the compiler understand when a variable or expression is guaranteed to be non-null based on a condition.

Considering:

```cs
using System.Diagnostics.CodeAnalysis;

string? input = Console.ReadLine();
if (!IsNullOrEmpty(input))
{
    // At this point, the compiler knows that 'input' is not null thanks to NotNullWhen attribute.
    // So you can safely use it without triggering nullability warnings.
    Console.WriteLine(input.Length);
}

static bool IsNullOrEmpty([NotNullWhen(false)]string? input)
{
    return string.IsNullOrEmpty(input);
}
```

In this example, we apply `NotNullWhen(false)` to indicate that when the method returns false, `input` is guaranteed to be non-null. So the compiler will not trigger any nullability warnings. Otherwise, a `CS8602 Dereference of a possibly null reference` is triggered when accessing the `Length` property because the compiler doesn't know that you've validated the input just before.

fabriziobagala:
@brentschmaltz using the `NotNullWhen(false)` attribute means that when the method returns `false`, the `enumerable` parameter is not `null`. This allows you to write subsequent code that relies on this guarantee without the need for additional nullity checks.

brentschmaltz:
@fabriziobagala what happens if IEnumerable is null when passed in?

fabriziobagala:
@brentschmaltz this happens:

```csharp
IEnumerable<string> collection = default!;
Console.WriteLine(collection.IsNullOrEmpty()); // Prints 'true'

collection = Enumerable.Repeat("a", 3);
Console.WriteLine(collection.IsNullOrEmpty()); // Prints 'false'
```

fabriziobagala:
@brentschmaltz the goal of this MR is only to prevent the code from being filled with `CS8602` warnings.

keegan-caruso:
The feedback from the community was to make CollectionUtilities internal in the latest major release.

PR #2398, opened 2023-11-10
Re-enable JwtPayload and JsonWebToken to process a JWT where 'sub' claim is set as a Number instead of String.

### LINKED ISSUE #2325: [Bug] System.IdentityModel.Tokens.Jwt 7.* broke loose JWT parsing
**Which version of Microsoft.IdentityModel are you using?** `System.IdentityModel.Tokens.Jwt 7.0.0`

**Where is the issue?**
 * [ ] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [ ] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [x] S.IM.Tokens.Jwt
 * Other (please describe)

The issue appears in `System.IdentityModel.Tokens.Jwt 7.0.0` only. The bug was introduced in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2268 (https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/2df4ca87d984cf44c6d30f00d8156651b1d7aae2).

**Repro**

Try and parse a JWT where `sub` is a number instead of a string.

**Expected behavior**
`sub` and other "string" fields should convert a number into a string instead of throwing.

**Actual behavior**
The `JwtPayload.CreatePayload` method throws an exception because `sub` is expected to be a string: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/2df4ca87d984cf44c6d30f00d8156651b1d7aae2/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L115-L141

**Possible solution**
Implicitly convert a number to a string.

Version 6.* had "loose" parsing of JWT tokens. This changed in 7.* to a more strict parser that throws an exception when a number is encountered instead of a string. While this behavior is spec-compliant (https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.2), some backends written in loosly typed languages like JavaScript don't convert a number to a string before creating the JWT. As such, many tokens in the wild will be affected by this change to a stricter parser.

FuPingFranco:
## Description

This PR re-enables incoming Jwt's to set the 'sub' claim as a Number type in their payloads.

- Added a new method "ReadStringOrNumberAsString" in the JsonSerializerPrimitives. It will process the 'sub' claim that comes in either as String or Number and will always return it back as a string.
- Replaced 'sub' claim logic to leverage ReadNumberAsString method in JwtPayload.cs
- Replaced 'sub' claim logic to leverage ReadNumberAsString method in JsonWebToken.cs

Fixes #2325

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
@@ -89,7 +89,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     }
                     else if (reader.ValueTextEquals(JwtPayloadUtf8Bytes.Sub))
                     {
-                        _sub = JsonSerializerPrimitives.ReadString(ref reader, JwtRegisteredClaimNames.Sub, ClassName, true);
+                        _sub = JsonSerializerPrimitives.ReadAsString(ref reader, JwtRegisteredClaimNames.Sub, ClassName, true);
                         claims[JwtRegisteredClaimNames.Sub] = _sub;
                     }
                     else
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
@@ -89,7 +89,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     }
                     else if (reader.ValueTextEquals(JwtPayloadUtf8Bytes.Sub))
                     {
-                        _sub = JsonSerializerPrimitives.ReadNumberAsString(ref reader, JwtRegisteredClaimNames.Sub, ClassName, true);
+                        _sub = JsonSerializerPrimitives.ReadStringOrNumberAsString(ref reader, JwtRegisteredClaimNames.Sub, ClassName, true);
                         claims[JwtRegisteredClaimNames.Sub] = _sub;
                     }
                     else
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
@@ -89,7 +89,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     }
                     else if (reader.ValueTextEquals(JwtPayloadUtf8Bytes.Sub))
                     {
-                        _sub = JsonSerializerPrimitives.ReadAsString(ref reader, JwtRegisteredClaimNames.Sub, ClassName, true);
+                        _sub = JsonSerializerPrimitives.ReadNumberAsString(ref reader, JwtRegisteredClaimNames.Sub, ClassName, true);
                         claims[JwtRegisteredClaimNames.Sub] = _sub;
                     }
                     else
--- a/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
@@ -655,8 +655,8 @@ namespace Microsoft.IdentityModel.Tokens.Json
         /// <param name="propertyName">The property name that is being read.</param>
         /// <param name="className">The type that is being deserialized.</param>
         /// <param name="read">If true reader.Read() will be called.</param>
-        /// <returns></returns>
-        internal static string ReadNumberAsString(ref Utf8JsonReader reader, string propertyName, string className, bool read = false)
+        /// <returns>Value from reader as string.</returns>
+        internal static string ReadStringOrNumberAsString(ref Utf8JsonReader reader, string propertyName, string className, bool read = false)
         {
             if (read)
                 reader.Read();
@@ -804,7 +804,7 @@ namespace Microsoft.IdentityModel.Tokens.Json
         /// <param name="propertyName">The property name that is being read.</param>
         /// <param name="className">The type that is being deserialized.</param>
         /// <param name="read">If true reader.Read() will be called.</param>
-        /// <returns></returns>
+        /// <returns>Value from reader as an object.</returns>
         internal static object ReadPropertyValueAsObject(ref Utf8JsonReader reader, string propertyName, string className, bool read = false)
         {
             if (read)
--- a/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
@@ -648,6 +648,33 @@ namespace Microsoft.IdentityModel.Tokens.Json
             return reader.GetString();
         }
 
+        /// <summary>
+        /// This method is called to enable the 'sub' claim to deserialize it as a Number then returns it as a string.
+        /// </summary>
+        /// <param name="reader">the <see cref="Utf8JsonReader"/></param>
+        /// <param name="propertyName">the property name that is being read</param>
+        /// <param name="className">the type that is being deserialized</param>
+        /// <param name="read">if true reader.Read() will be called.</param>
+        /// <returns></returns>
+        internal static string ReadAsString(ref Utf8JsonReader reader, string propertyName, string className, bool read = false)
+        {
+            if (read)
+                reader.Read();
+
+            // returning null keeps the same logic as JsonSerialization.ReadObject
+            if (reader.TokenType == JsonTokenType.Null)
+                return null;
+
+            if (reader.TokenType == JsonTokenType.Number)
+                return ReadNumber(ref reader).ToString();
+
+            if (reader.TokenType != JsonTokenType.String || reader.TokenType != JsonTokenType.Number)
+                throw LogHelper.LogExceptionMessage(
+                    CreateJsonReaderException(ref reader, "JsonTokenType.String or JsonTokenType.Number", className, propertyName));
+
+            return reader.GetString();
+        }
+
         internal static object ReadStringAsObject(ref Utf8JsonReader reader, string propertyName, string className, bool read = false)
         {
             if (read)
--- a/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
@@ -649,14 +649,14 @@ namespace Microsoft.IdentityModel.Tokens.Json
         }
 
         /// <summary>
-        /// This method is called to enable the 'sub' claim to deserialize it as a Number then returns it as a string.
+        /// This method allows a JsonTokenType to be string or number but, it will always return it as a string.
         /// </summary>
-        /// <param name="reader">the <see cref="Utf8JsonReader"/></param>
-        /// <param name="propertyName">the property name that is being read</param>
-        /// <param name="className">the type that is being deserialized</param>
-        /// <param name="read">if true reader.Read() will be called.</param>
+        /// <param name="reader">The<see cref="Utf8JsonReader"/></param>
+        /// <param name="propertyName">The property name that is being read.</param>
+        /// <param name="className">The type that is being deserialized.</param>
+        /// <param name="read">If true reader.Read() will be called.</param>
         /// <returns></returns>
-        internal static string ReadAsString(ref Utf8JsonReader reader, string propertyName, string className, bool read = false)
+        internal static string ReadNumberAsString(ref Utf8JsonReader reader, string propertyName, string className, bool read = false)
         {
             if (read)
                 reader.Read();
@@ -668,7 +668,7 @@ namespace Microsoft.IdentityModel.Tokens.Json
             if (reader.TokenType == JsonTokenType.Number)
                 return ReadNumber(ref reader).ToString();
 
-            if (reader.TokenType != JsonTokenType.String || reader.TokenType != JsonTokenType.Number)
+            if (reader.TokenType != JsonTokenType.String)
                 throw LogHelper.LogExceptionMessage(
                     CreateJsonReaderException(ref reader, "JsonTokenType.String or JsonTokenType.Number", className, propertyName));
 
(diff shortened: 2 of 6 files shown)


src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -648,6 +648,33 @@ internal static string ReadString(ref Utf8JsonReader reader, string propertyName
...
 
+        /// <summary>
+        /// This method is called to enable the 'sub' claim to deserialize it as a Number then returns it as a string.
+        /// </summary>
+        /// <param name="reader">the <see cref="Utf8JsonReader"/></param>
+        /// <param name="propertyName">the property name that is being read</param>
+        /// <param name="className">the type that is being deserialized</param>
+        /// <param name="read">if true reader.Read() will be called.</param>
+        /// <returns></returns>
+        internal static string ReadAsString(ref Utf8JsonReader reader, string propertyName, string className, bool read = false)
```
ciaozhang: Suggest change name to ReadNumberAsString

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -648,6 +648,33 @@ internal static string ReadString(ref Utf8JsonReader reader, string propertyName
             return reader.GetString();
         }
 
+        /// <summary>
+        /// This method is called to enable the 'sub' claim to deserialize it as a Number then returns it as a string.
+        /// </summary>
+        /// <param name="reader">the <see cref="Utf8JsonReader"/></param>
+        /// <param name="propertyName">the property name that is being read</param>
+        /// <param name="className">the type that is being deserialized</param>
```
ciaozhang: nit: capitalize the first character and end with a period

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -648,6 +648,33 @@ internal static string ReadString(ref Utf8JsonReader reader, string propertyName
             return reader.GetString();
         }
 
+        /// <summary>
+        /// This method is called to enable the 'sub' claim to deserialize it as a Number then returns it as a string.
```
ciaozhang: nit: If it is for sub claim only, consider rename it to ReadSubClaimAsString or don't mention the claim in summary

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -648,6 +648,33 @@ internal static string ReadString(ref Utf8JsonReader reader, string propertyName
...
+                reader.Read();
+
+            // returning null keeps the same logic as JsonSerialization.ReadObject
+            if (reader.TokenType == JsonTokenType.Null)
+                return null;
+
+            if (reader.TokenType == JsonTokenType.Number)
+                return ReadNumber(ref reader).ToString();
+
+            if (reader.TokenType != JsonTokenType.String || reader.TokenType != JsonTokenType.Number)
```
keegan-caruso: check my logic, but we don't need to check if token type is not a number here because of the check directly above

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
```diff
@@ -603,6 +604,131 @@ public void TryGetPayloadValue(GetPayloadValueTheoryData theoryData)
...
 
+        [Theory, MemberData(nameof(GetPayloadSubClaimValueTheoryData), DisableDiscoveryEnumeration = true)]
+        public void GetPayloadSubClaimValue(GetPayloadValueTheoryData theoryData)
+        {
+            CompareContext context = TestUtilities.WriteHeader($"{this}.GetPayloadSubClaimValue", theoryData);
+            try
+            {
+                JsonWebToken jsonWebToken = new JsonWebToken(theoryData.Json);
+                string payload = Base64UrlEncoder.Decode(jsonWebToken.EncodedPayload);
+                MethodInfo method = typeof(JsonWebToken).GetMethod("GetPayloadValue");
```
keegan-caruso, line 615: we shouldn't need to use reflection....

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
```diff
@@ -89,7 +89,7 @@ internal JsonClaimSet CreatePayloadClaimSet(byte[] bytes, int length)
                     }
                     else if (reader.ValueTextEquals(JwtPayloadUtf8Bytes.Sub))
                     {
-                        _sub = JsonSerializerPrimitives.ReadString(ref reader, JwtRegisteredClaimNames.Sub, ClassName, true);
+                        _sub = JsonSerializerPrimitives.ReadAsString(ref reader, JwtRegisteredClaimNames.Sub, ClassName, true);
```
ciaozhang: could it be ReadNumber(ref reader).ToString() here instead of create a new parse method?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
FuPingFranco: Yes but, it is more consistent with the rest of tests in this class to write it this way. What do you think?

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
FuPingFranco: I don't think so, ReadNumber only checks reader it's a valid number and returns it back, Although, I see why ToString() might seem as a way to do it, there's additional logic that deals with Null as well as, additional exceptions thrown when dealing with Array, Object and Bool types. I think this approach encapsulates this work as are deviating from the oauth spec to fit our customers.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
FuPingFranco: I agree, I'll update the summary. Thanks Xiao!

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
FuPingFranco: You're right, I'll remove the additional check. Thanks Keegan!

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
FuPingFranco: Done. Thanks Xiao!

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
FuPingFranco: Done. Thanks Xiao!

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
brentschmaltz: @keegan-caruso this is a way i found to call generic methods specifying [T]. There may be other way.

brentschmaltz (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
keegan-caruso: Would have to frame it slightly differently. Could have had a func<object> on the theorydata that would retrieve the value instead of a type parameter. 

That would be a medium refactor. Lets consider it for later. Resolving for now.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -648,6 +648,33 @@ internal static string ReadString(ref Utf8JsonReader reader, string propertyName
...
 
+        /// <summary>
+        /// This method allows a JsonTokenType to be string or number but, it will always return it as a string.
+        /// </summary>
+        /// <param name="reader">The<see cref="Utf8JsonReader"/></param>
+        /// <param name="propertyName">The property name that is being read.</param>
+        /// <param name="className">The type that is being deserialized.</param>
+        /// <param name="read">If true reader.Read() will be called.</param>
+        /// <returns></returns>
+        internal static string ReadNumberAsString(ref Utf8JsonReader reader, string propertyName, string className, bool read = false)
```
keegan-caruso: This name isn't honest. Reads as either string or number isn't ReadNumberAsString

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -648,6 +648,33 @@ internal static string ReadString(ref Utf8JsonReader reader, string propertyName
...
         }
 
+        /// <summary>
+        /// This method allows a JsonTokenType to be string or number but, it will always return it as a string.
+        /// </summary>
+        /// <param name="reader">The<see cref="Utf8JsonReader"/></param>
+        /// <param name="propertyName">The property name that is being read.</param>
+        /// <param name="className">The type that is being deserialized.</param>
+        /// <param name="read">If true reader.Read() will be called.</param>
+        /// <returns></returns>
```
keegan-caruso: nit empty tag

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
```diff
@@ -603,6 +604,131 @@ public void TryGetPayloadValue(GetPayloadValueTheoryData theoryData)
...
+                    PropertyType = typeof(uint),
+                    PropertyValue = 540,
+                    Json = JsonUtilities.CreateUnsignedToken("sub", 540)
+                });
+
+                theoryData.Add(new GetPayloadValueTheoryData("SubjectAsUlong")
+                {
+                    PropertyName = "sub",
+                    PropertyType = typeof(ulong),
+                    PropertyValue = 642,
```
keegan-caruso, line 723: should this actually be a ulong?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
```diff
@@ -603,6 +604,131 @@ public void TryGetPayloadValue(GetPayloadValueTheoryData theoryData)
...
+                    PropertyType = typeof(int),
+                    PropertyValue = 42,
+                    Json = JsonUtilities.CreateUnsignedToken("sub", 42)
+                });
+
+                theoryData.Add(new GetPayloadValueTheoryData("SubjectAsUInt")
+                {
+                    PropertyName = "sub",
+                    PropertyType = typeof(uint),
+                    PropertyValue = 540,
```
keegan-caruso, line 715: should this actually be a uint?

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -114,31 +114,8 @@ internal static JwtPayload CreatePayload(byte[] bytes, int length)
                     }
                     else if (reader.ValueTextEquals(JwtPayloadUtf8Bytes.Sub))
                     {
-                        reader.Read();
```
keegan-caruso, line 117: I think this is failing the tests

FuPingFranco: Thanks Keegan!!

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
FuPingFranco: I had thoughts about this as well, but we can't really control what values are being passed to us so, it seemed appropriate to test this as well since, I am calling the current ReadNumber() method and those are valid numbers. Do you think we should create a different ReadNumber check excluding them?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
FuPingFranco: Same comment as above.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
keegan-caruso: This is testing if you can convert an int in the JWT into a ulong/uint. It isn't testing if you can round trip a ulong/uint.

Which was this trying to test?

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
FuPingFranco: I agree, how about ReadStringOrNumberAsString or ReadAsString?

keegan-caruso: it is not public api, so no strong opinions. As long as it is honest 😄

FuPingFranco: ReadSringOrNumberAsString it is lol

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
FuPingFranco: I was using this test to try different variations of the TryGet on the ReadNumber() method when we set a 'sub' claim as a number.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
FuPingFranco: Fixed. Thanks Keegan!

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -648,6 +648,33 @@ internal static string ReadString(ref Utf8JsonReader reader, string propertyName
             return reader.GetString();
         }
 
+        /// <summary>
+        /// This method allows a JsonTokenType to be string or number but, it will always return it as a string.
```
jennyf19, line 652: Do we want to add that this is to maintain compatibility with 6x?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
jennyf19: Nice to have more test coverage

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
FuPingFranco: Yes, also we used to allow 'sub' to be an array and that was failing an older test when I worked on this but, that is out of scope for this one. Hopefully, nobody is doing that.

jennyf19 (approved):
:shipit:

PR #2413, opened 2023-11-22
Decouple benchmark runs from test projects.

brentschmaltz:
Adjusted tests to use the same claims, keys and more realistic settings in runs.
Benchmark tests no longer depend on test projects for defaults as these can change when tests change.

Added simple way to debug through tests to verify settings.


jennyf19 (approved):
:shipit:

PR #2415, opened 2023-11-23
[Backport] Fix AadIssuerValidator's handling of trailing forward slashes

halter73:
This is a backport of #2361 since some customers are running into this issue when using the Microsoft.IdentityModel.Validators 6.x packages with ASP.NET Core 8.

This can happen even if developers are referencing the latest Microsoft.Identity.Web NuGet package but happen to be doing it from a net7.0 or earlier project that is then referenced by net8.0 project. See https://github.com/dotnet/aspnetcore/issues/52283#issuecomment-1823625562 for background

@jmprieur @jennyf19


jennyf19 (approved):
:shipit:

PR #2416, opened 2023-11-23
make CollectionUtilities internal

SimonCropp:
You should not extend types you dont own

also CollectionUtilities.IsNullOrEmpty  is not a safe general purpose method. if the IEnumerable is a forward only approach, then this method will consume the first value effectively creating a bug.


Dreamescaper:
Fixes #1722.

jmprieur:
This is a breaking change?

brentschmaltz:
@jmprieur @SimonCropp @Dreamescaper This is a breaking change.

SimonCropp:
yes it is a breaking change

brentschmaltz (commented):
This would be a breaking change.
We will have to wait for 8.x.

PR #2421, opened 2023-11-30
Adds an AppContext switch to control HMAC key size verification.

brentschmaltz:
Conforming with https://datatracker.ietf.org/doc/html/rfc7518#section-3.2 causes some users to fault.
The AppContext switch "Switch.Microsoft.IdentityModel.UnsafeRelaxHmacKeySizeValidation" can be used bypass this new check.

HMAC will still require a minimum of 128 bits.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
+++ b/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
@@ -19,7 +19,9 @@ namespace Microsoft.IdentityModel.Tokens
         private static object _cacheLock = new object();
         private static int _defaultSignatureProviderObjectPoolCacheSize = Environment.ProcessorCount * 4;
         private int _signatureProviderObjectPoolCacheSize = _defaultSignatureProviderObjectPoolCacheSize;
-
+#if !NET45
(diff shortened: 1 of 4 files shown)


jennyf19 (approved):

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -489,6 +491,10 @@ public virtual KeyedHashAlgorithm CreateKeyedHashAlgorithm(byte[] keyBytes, stri
 
         private static void ValidateKeySize(byte[] keyBytes, string algorithm, int expectedNumberOfBytes)
         {
+#if !NET45
+            if (AppContext.TryGetSwitch(_skipValidationOfHmacKeySizes, out bool skipValidationOfHmacKeySize) && skipValidationOfHmacKeySize)
```
keegan-caruso, line 495: should this skip or go back to the older thresholds?

keegan-caruso (approved):

nikcio:
@brentschmaltz I can see this is part of v6.35.0. Do you know if it will also become part of v7 at some point?

ajoaquincardozo:
@brentschmaltz coming back on this - are you aware if this change is planned to be included in version 7 or future releases?

PR #2424, opened 2023-12-14
Remove condition for net8.0, fixes pack.

brentschmaltz:
We need all versions built for api compat to work.


keegan-caruso (approved):

keegan-caruso:
There are a lot more TargetNet8 conditions throughout the code if you want to tackle those. Might cause more of the same problems

brentschmaltz:
@keegan-caruso some of the net8.0 ifdefs are needed.

PR #2427, opened 2023-12-15
Draft: Add Eddsa signature validation algorithms. fix #2426

### LINKED ISSUE #2426: Support EdDSA signature validations
Today:
EdDSA signed JWTs

Future:
The goal is for this list: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L62 to support an EdDSA signature strategy so that dependencies of this library understand how to handle EdDSA signed JWTs.

In this case the issue is sourced from `Microsoft.AspNetCore.Authorization.Authorize` attribute which attempts to use `System.IdentityModel.Tokens.Jwt` through the package `Microsoft.AspNetCore.Authentication.JwtBearer` (this repo) package for verifying the incoming tokens.

Related:
* #1970
* https://github.com/Authress/authress-sdk.cs/issues/20

If a PR is opened to add this support, will it be accepted and helped to be shepherded to completion or will a PR of this nature lay on apathetic ears?

wparad:
# Add Eddsa signature validation algorithms. fix #2426

Add in EdDSA almost everywhere. Still missing:
* [ ] Code to actually verify EdDSA algorithm along with the generation of the EdDSASecurity key class.

At this point, I'm looking for encouragement to continue this PR.

Thanks,
Warren

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -41,7 +41,12 @@ namespace Microsoft.IdentityModel.Tokens
             { SecurityAlgorithms.RsaSsaPssSha512, 1040 },
             { SecurityAlgorithms.RsaSsaPssSha256Signature, 528 },
             { SecurityAlgorithms.RsaSsaPssSha384Signature, 784 },
-            { SecurityAlgorithms.RsaSsaPssSha512Signature, 1040 }
+            { SecurityAlgorithms.RsaSsaPssSha512Signature, 1040 },
+            { SecurityAlgorithms.EddsaEd25519Signature, 256 },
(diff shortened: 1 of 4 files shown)


wparad:
@microsoft-github-policy-service agree company="Rhosys AG"

brentschmaltz:
@wparad we are discussing this internally.
Will have an answer soon.

brentschmaltz (changes requested):
We need to see test cases for all the additional algorithms.
It would be best to create a new file for theses tests.

wparad:
@brentschmaltz Not just the unit tests, the code for the implementation is missing. Can you give me some pointers on where to write that implementation. It's very hard to understand a lot of the nomenclature that is being used here.

Thanks!

PR #2436, opened 2024-01-04
Optmise IDX10503

ciaozhang:
What?
The keys tried during token validation are public keys used to validate the tokens and are listed on a public endpoint. They should be logged without PII to enable partner debugging more easily.

The text of the error should read instead "The token's kid is either missing or doesn't match any of the keys in the configuration" or similar to better reflect the flow as it's possible the token does have a kid, it just doesn't match the keys in config.

WI:
https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20AuthNZ%20-%20Customer%20trust/Backlog%20items/?workitem=2723627

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
@@ -83,7 +83,7 @@ namespace Microsoft.IdentityModel.Tokens
         // 10500 - SignatureValidation
         public const string IDX10500 = "IDX10500: Signature validation failed. No security keys were provided to validate the signature.";
         //public const string IDX10501 = "IDX10501: Signature validation failed. Unable to match key: \nkid: '{0}'. \nNumber of keys in TokenValidationParameters: '{1}'. \nNumber of keys in Configuration: '{2}'. \nExceptions caught:\n '{3}'. \ntoken: '{4}'.";
-        public const string IDX10503 = "IDX10503: Signature validation failed. Token does not have a kid. Keys tried: '{0}'. Number of keys in TokenValidationParameters: '{1}'. \nNumber of keys in Configuration: '{2}'. \nExceptions caught:\n '{3}'.\ntoken: '{4}'. See https://aka.ms/IDX10503 for details.";
+        public const string IDX10503 = "IDX10503: Signature validation failed. The token's kid is either missing or doesn't match any of the keys in the configuration and TokenValidationParameters. Keys tried: '{0}'. Number of keys in TokenValidationParameters: '{1}'. \nNumber of keys in Configuration: '{2}'. \nExceptions caught:\n '{3}'.\ntoken: '{4}'. See https://aka.ms/IDX10503 for details.";
         public const string IDX10504 = "IDX10504: Unable to validate signature, token does not have a signature: '{0}'.";
         public const string IDX10505 = "IDX10505: Signature validation failed. The user defined 'Delegate' specified on TokenValidationParameters returned null when validating token: '{0}'.";
         public const string IDX10506 = "IDX10506: Signature validation failed. The user defined 'Delegate' specified on TokenValidationParameters did not return a '{0}', but returned a '{1}' when validating token: '{2}'.";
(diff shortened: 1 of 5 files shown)


sruke (approved):
LGTM!

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -83,7 +83,7 @@ internal static class LogMessages
         // 10500 - SignatureValidation
         public const string IDX10500 = "IDX10500: Signature validation failed. No security keys were provided to validate the signature.";
         //public const string IDX10501 = "IDX10501: Signature validation failed. Unable to match key: \nkid: '{0}'. \nNumber of keys in TokenValidationParameters: '{1}'. \nNumber of keys in Configuration: '{2}'. \nExceptions caught:\n '{3}'. \ntoken: '{4}'.";
-        public const string IDX10503 = "IDX10503: Signature validation failed. Token does not have a kid. Keys tried: '{0}'. Number of keys in TokenValidationParameters: '{1}'. \nNumber of keys in Configuration: '{2}'. \nExceptions caught:\n '{3}'.\ntoken: '{4}'. See https://aka.ms/IDX10503 for details.";
+        public const string IDX10503 = "IDX10503: Signature validation failed. The token's kid is either missing or doesn't match any of the keys in the configuration and TokenValidationParameters. Keys tried: '{0}'. Number of keys in TokenValidationParameters: '{1}'. \nNumber of keys in Configuration: '{2}'. \nExceptions caught:\n '{3}'.\ntoken: '{4}'. See https://aka.ms/IDX10503 for details.";
```
brentschmaltz: Could you consider separate error messages.
2 - could be the same message, but with different id's for JsonWebTokenHandler, JwtSecurityTokenHandler.
Currently, the two handlers share the same error id, it would simplify debugging if the errors were distinguished.

Then, separate into two messages one when the kid is null and one when the kid is not null.
For example:

public const string IDX10503 = "IDX10503: Signature validation failed. The token's kid is missing. Keys tried: '{0}'. Number of keys in TokenValidationParameters: '{1}'. \nNumber of keys in Configuration: '{2}'. \nExceptions caught:\n '{3}'.\ntoken: '{4}'. See https://aka.ms/IDX10503 for details.";

public const string IDX10504 = "IDX10503: Signature validation failed. The token's kid is: '{0}', but did not match any keys in TokenValidationParameters or Configuration. Keys tried: '{1}'. Number of keys in TokenValidationParameters: '{2}'. \nNumber of keys in Configuration: '{3}'. \nExceptions caught:\n '{4}'.\ntoken: '{5}'. See https://aka.ms/IDX10504 for details.";

brentschmaltz (changes requested):
separate id's between JsonWebTokenHandler and JwtSecurityTokenHandler, have messages indicate that kid is found (and log) or empty.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
ciaozhang: Added a new message for kid is missing case @brentschmaltz pls take a look

jmprieur (approved):

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -83,7 +83,7 @@ internal static class LogMessages
         // 10500 - SignatureValidation
         public const string IDX10500 = "IDX10500: Signature validation failed. No security keys were provided to validate the signature.";
         //public const string IDX10501 = "IDX10501: Signature validation failed. Unable to match key: \nkid: '{0}'. \nNumber of keys in TokenValidationParameters: '{1}'. \nNumber of keys in Configuration: '{2}'. \nExceptions caught:\n '{3}'. \ntoken: '{4}'.";
-        public const string IDX10503 = "IDX10503: Signature validation failed. Token does not have a kid. Keys tried: '{0}'. Number of keys in TokenValidationParameters: '{1}'. \nNumber of keys in Configuration: '{2}'. \nExceptions caught:\n '{3}'.\ntoken: '{4}'. See https://aka.ms/IDX10503 for details.";
```
jmprieur, line 86: Do we remove the case of tokens without kid?

ciaozhang: Add a another message for no kid case

jennyf19 (approved):
:shipit:

PR #2437, opened 2024-01-04
Add CancellationToken for ValidateTokenAsync

RojaEnnam:
Adds CancellationToken to ValidateTokenAsync  API. #2451


ciaozhang (approved):

westin-m (approved):

westin-m (commented):
waiting for a solution to virtual method throwing an exception in abstract class

keegan-caruso:
pls rebase off of dev. I expect this to fail on baseline package validation.

keegan-caruso (changes requested):
will likely fail baseline package validation.

keegan-caruso:
I was wrong, it isn't failing, I guess because of the default implementation. But this is still breaking at runtime.

PR #2440, opened 2024-01-09
Add back support for Guid values in JwtPayload...

### LINKED ISSUE #2439: [Bug] JwtPayload no longer supports Guid claim values
Microsoft.IdentityModel.Tokens 7.0.3

**Where is the issue?**
 * [ ] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [X] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [ ] S.IM.Tokens.Jwt
 * Other (please describe)

The app is in production and I have upgraded to a new version of Microsoft.IdentityModel.*

**Repro**

```csharp

var payload = new JwtPayload();
payload.Add("appid", Guid.NewGuid());

var jwt = payload.SerializeToJson();
```

**Expected behavior**
```csharp
jwt == $"{{\"appid\":\"{guid}\"}}"
```

**Actual behavior**
```
System.ArgumentException: 'IDX11025: Cannot serialize object of type: 'System.Guid' into property: 'appid'.'
```

**Possible solution**
New serializer logic could serialize Guids as strings (as the Newtonsoft implementation did)...

KevinH-MS:
# Add back support for Guid values in JwtPayload...

Changes `JsonSerializerPrimitves` to serialize claim values of type `Guid` as `string` (rather than throwing an Exception)...

## Description

In versions of `Microsoft.IdentityModel.Tokens` < 7.0.0-preview4, claim values could have values of type `Guid` and would be serialized successfully. This appears to have been regressed with the rewrite of the serializer implementation.

I checked the round-trip behavior with older versions, and it seems that the serialized strings are subsequently deserialized as `string` (by `JwtPayload.Deserialize`), so I left the behavior asymmetric (rather than try-parsing all string values as `Guid`).

Fixes #2439

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
@@ -941,6 +941,8 @@ namespace Microsoft.IdentityModel.Tokens.Json
                 writer.WriteNumber(key, d);
             else if (obj is float f)
                 writer.WriteNumber(key, f);
+            else if (obj is Guid g)
+                writer.WriteString(key, g);
             else
                 throw LogHelper.LogExceptionMessage(
                     new ArgumentException(
--- a/test/Microsoft.IdentityModel.Tokens.Tests/Json/JsonSerializerPrimitivesTests.cs
+++ b/test/Microsoft.IdentityModel.Tokens.Tests/Json/JsonSerializerPrimitivesTests.cs
@@ -20,7 +20,7 @@ namespace Microsoft.IdentityModel.Tokens.Json.Tests
         [Fact]
         public void CheckMaxDepthReading()
         {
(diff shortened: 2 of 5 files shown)


KevinH-MS:
Tagging @brentschmaltz ... Seems like a difference between Newtonsoft <-> System.Text.Json. I went ahead and added a quirk for this since it seemed trivial, but let me know if this behavior was deliberately changed. 

In my situation, it'd be a lot easier to make the fix here as opposed to fixing and upgrading the library I'm calling that's adding the claim with the `Guid` value...seems like preserving the old behavior might save others some headache as well...

keegan-caruso (approved):
LGTM, thanks

jennyf19 (approved):
:shipit:

jennyf19:
@KevinH-MS appreciate the added tests.

PR #2449, opened 2024-01-22
Added benchmarks for AsymmetricAdapter signatures

brentschmaltz:
These benchmarks measure the AsymmetricAdapter performance.
This measure performance at a low level.
Higher level measurements will be added to ensure that performance is not regressed.


jennyf19 (approved):
:shipit:

PR #2453, opened 2024-01-23
Use invariant culture when formatting claims in JwtSecurityToken

### LINKED ISSUE #2409: [Bug] JwtPayload.Claim.Value should be culture invariant
System.IdentityModel.Tokens.Jwt 7.0.3

**Where is the issue?**
 * [ ] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [ ] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [X] S.IM.Tokens.Jwt
 * Other (please describe)

c. This is a new app or an experiment.

**Repro**
Step 1: Set computer localization to one of west european countries, for example "fr-FR", where decimal separator is comma `,`.
In these countries, decimal value `10.9` is written as `10,9`.
Step 2: Create a JWT with a decimal type claim in the Payload:
Step 3: Get the claim in the payload. The claim.value has a string value `10,9`

```csharp
var credentials = new SigningCredentials(securityKey, algorithm); //whatever signing credentials
var jwtHeader = new JwtHeader(credentials); //whatever headers
var payload = new JwtPayload();
payload.add("numericClaim", 10.9d); //<- here
var token = new JwtSecurityToken(jwtHeader, payload)
```

**Expected behavior**
At API level, we would like decimal and datetime data to be represented in invariant culture. 
DateTime is perfectly handled with invariant culture: line 522
But all decimal values are not: line 532
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/5c65369a4f7bb5df685731831fcfbf00fcc48fc2/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L522-L532

**Actual behavior**
Claim.value is string, from what i have obseved, when the claim is added in the payload with a simple ToString(). The claim value is stored with current culture of the computer (for example: fr-FR, which is not exterchangable with another computer en-US)
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-

Claim.Value stores "10,9", which is not expected at API/data exchange level. We would like all numeric values to be stored as standard iso format.

**Possible solution**
I think the behaviour can be fixed in these two lines, by adding a check on numberic types (decimal, double, float, ...) and use invariant culture version of string conversion:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/5c65369a4f7bb5df685731831fcfbf00fcc48fc2/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L532
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/5c65369a4f7bb5df685731831fcfbf00fcc48fc2/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L559

please find here a screenshot of debug info of a generated token:
image

keegan-caruso:
# Use invariant culture when formatting claims in JwtSecurityToken

## Description

Verified JSONWebToken handles this correctly already.

Fixes #2409

### DIFF
--- a/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
@@ -529,7 +529,14 @@ namespace System.IdentityModel.Tokens.Jwt
                             claims.Add(new Claim(keyValuePair.Key, "false", ClaimValueTypes.Boolean, issuer, issuer));
                     }
                     else if (keyValuePair.Value != null)
-                        claims.Add(new Claim(keyValuePair.Key, keyValuePair.Value.ToString(), GetClaimValueType(keyValuePair.Value), issuer, issuer));
+                    {
+                        var value = keyValuePair.Value;
+                        var claimValueType = GetClaimValueType(value);
+                        if (value is IConvertible convertible)
+                            claims.Add(new Claim(keyValuePair.Key, convertible.ToString(CultureInfo.InvariantCulture), claimValueType, issuer, issuer));
+                        else
+                            claims.Add(new Claim(keyValuePair.Key, value.ToString(), claimValueType, issuer, issuer));
+                    }
                 }
 
                 return claims;
--- a/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
@@ -532,8 +532,8 @@ namespace System.IdentityModel.Tokens.Jwt
                     {
                         var value = keyValuePair.Value;
                         var claimValueType = GetClaimValueType(value);
-                        if (value is IConvertible convertible)
-                            claims.Add(new Claim(keyValuePair.Key, convertible.ToString(CultureInfo.InvariantCulture), claimValueType, issuer, issuer));
+                        if (value is IFormattable formattable)
+                            claims.Add(new Claim(keyValuePair.Key, formattable.ToString(null, CultureInfo.InvariantCulture), claimValueType, issuer, issuer));
                         else
                             claims.Add(new Claim(keyValuePair.Key, value.ToString(), claimValueType, issuer, issuer));
                     }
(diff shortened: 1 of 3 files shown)


test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
```diff
@@ -1631,6 +1631,25 @@ public void EscapedClaims()
...
         }
+
+        [Fact]
+        public void DifferentCultureJsonWebToken()
+        {
+            string result = string.Empty;
+
+            var thread = new Thread(() =>
+            {
+                CultureInfo.CurrentCulture = new CultureInfo("fr-FR");
```
jennyf19, line 1642: 😁🥖

jennyf19 (approved):
:shipit:

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -529,7 +529,14 @@ public virtual IEnumerable<Claim> Claims
                             claims.Add(new Claim(keyValuePair.Key, "false", ClaimValueTypes.Boolean, issuer, issuer));
                     }
                     else if (keyValuePair.Value != null)
-                        claims.Add(new Claim(keyValuePair.Key, keyValuePair.Value.ToString(), GetClaimValueType(keyValuePair.Value), issuer, issuer));
+                    {
+                        var value = keyValuePair.Value;
+                        var claimValueType = GetClaimValueType(value);
+                        if (value is IConvertible convertible)
```
brentschmaltz: IConvertible documentation indicates the interface is not CLS-Compliant.
Is that a problem? Currently, System.Identitymodel.Tokens.Jwt is CLS-Compliant.

https://learn.microsoft.com/en-us/dotnet/api/system.iconvertible?view=net-8.0

keegan-caruso: fixed

PR #2458, opened 2024-01-24
Adds JsonWebToken Constructor that accepts Memory<char>

### LINKED ISSUE #2487: [Feature Request] Add a JsonWebToken Constructor that accepts ReadOnlyMemory representation of the encodedToken
The usage of String.Split to extract segments from the token is expensive as it creates new strings and leads to additional memory allocation.

- Introduce a JsonWebToken Constructor that accepts ReadOnlyMemory representation of the encodedToken. This would allow us to substitute string.Split with span.Slice which will avoid string creation.
- Update existing constructors to call into ReadTokens(ReadOnlyMemory<char>) to maintain a single code path and enable current implementations to take advantage of some of the improvements without making any changes.

sruke:
# Introduces a JsonWebToken Constructor that accepts ReadOnlyMemory<char> representation of the encodedToken

## Description
This PR
- Introduces a JsonWebToken constructor that accepts ReadOnlyMemory<char> representation of the encodedToken
- Updates existing JsonWebToken constructors to use string.AsMemory() to maintain a single code path
- ReadTokens(..) Uses Memory<T>.Span to extract the token segments such as the header, payload, encryptedKeyBytes etc.
- Refactors CreateClaimSet by moving ArrayPool usage into the JsonWebToken that reduces memory allocation by 4.6% during JWS creation.

**Performance Assessment on .NET 8:**
When using the preview version and opting for the constructor that accepts a string for the encoded token, the generation of a JWS yields a 3.95% decrease in memory usage, a 4.52% reduction in Gen0, and a 4.61% decrease in execution time compared to the latest release. The performance of the constructor that accepts ReadOnlyMemory is comparable to the constructor accepting a string for the encoded token, both for JWS and JWE.

Preview branch:
image

Dev (main) branch:
image

**Why?**
The new constructor allows a customer to leverage spans for slicing the encoded token directly from the Authorization Header eliminating additional String.Split operations on their end and can further improve performance.


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -513,6 +537,193 @@ internal void ReadToken(string encodedJson)
...
+                IsSigned = !(headerSpan.Length + secondSpan.Length + 2 == encodedJsonSpan.Length);
+                try
+                {
+                    Header = CreateClaimSet(headerSpan.ToString(), 0, headerSpan.Length, CreateHeaderClaimSet); // CreateClaimSet invokes Base64UrlEncoding.Decode which accepts only a string
+                }
+                catch (Exception ex)
+                {
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(
+                        LogMessages.IDX14102,
+                        LogHelper.MarkAsUnsafeSecurityArtifact(headerSpan.ToString(), t => t.ToString())), // TODO: Add an overload to LogHelper.MarkAsUnsafeSecurityArtifact that accepts span?
```
keegan-caruso: It is on the exception path, I'm not worried about it in this case.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -513,6 +537,193 @@ internal void ReadToken(string encodedJson)
...
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(
+                        LogMessages.IDX14102,
+                        LogHelper.MarkAsUnsafeSecurityArtifact(headerSpan.ToString(), t => t.ToString())),
+                        ex));
+                }
+
+                // dir does not have any key bytes
+                if (!secondSpan.IsEmpty)
+                {
+                    EncryptedKeyBytes = Base64UrlEncoder.UnsafeDecode(secondSpan.ToArray());
```
keegan-caruso: will be great when we can drop the ToArray

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -513,6 +537,193 @@ internal void ReadToken(string encodedJson)
...
+
+                HeaderAsciiBytes = headerAsciiBytes;
+
+                try
+                {
+                    // Discuss: Using ReadOnlySpan instead of ReadOnlyMemory
+                    // ReadOnlySpan is recommended for synchronous calls where as ReadOnlyMemory is recommended for async calls.
+                    // ReadOnlySpan performs better than ReadOnlyMemory.
+                    // https://learn.microsoft.com/en-us/dotnet/standard/memory-and-spans/memory-t-usage-guidelines
+                    Header = CreateHeaderClaimSet(Base64UrlEncoder.UnsafeDecode(headerSpan.ToArray()));  // TODO: Looking to update UnsafeDecode to accept ReadOnlySpan
```
keegan-caruso: the conversion makes sense

buildConfiguration.xml
```diff
@@ -2,7 +2,7 @@
   <dotnetArchitecture>x64</dotnetArchitecture>
   <nugetVersion>3.5.0-rc-1285</nugetVersion>
   <runtimes>net461,netstandard2.0</runtimes>
-  <assemblyVersion>7.1.2</assemblyVersion>
+  <assemblyVersion>7.2.1</assemblyVersion>
```
jennyf19: >7.2.1 [](http://example.com/codeflow?start=19&length=5)

should probably be a separate PR sooner.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -513,6 +537,193 @@ internal void ReadToken(string encodedJson)
...
+                }
+            }
+            else
+            {
+                // JWE: https://www.rfc-editor.org/rfc/rfc7516
+                // Format: https://www.rfc-editor.org/rfc/rfc7516#page-8
+                // empty payload for JWE's {encrypted tokens}.
+                Payload = new JsonClaimSet();
+
+                if (dotIndex == encodedJsonSpanCopy.Length) // TODO: Should this be encodedJsonSpanCopy.Length - 1? Using encodedJsonSpanCopy.Length doesn't add up
```
jennyf19: >Length [](http://example.com/codeflow?start=52&length=6)

what's the issue with .Length? 'dotIndex' is the first dot character, so I'd think `dotIndex == encodedJsonSpanCopy.Length` would work.

benchmark/Microsoft.IdentityModel.Benchmarks/EncodedJWTs.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright(c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.IdentityModel.Benchmarks
+{
+    internal readonly struct EncodedJWTs
+    {
+        internal static string Asymmetric_LocalSts => "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJpc3MiOiJodHRwOi8vR290Snd0LmNvbSIsImF1ZCI6Imh0dHA6Ly9Db250b3NvLmNvbSIsImh0dHA6Ly9zY2hlbWFzLnhtbHNvYXAub3JnL3dzLzIwMDUvMDUvaWRlbnRpdHkvY2xhaW1zL2NvdW50cnkiOiJVU0EiLCJodHRwOi8vc2NoZW1hcy54bWxzb2FwLm9yZy93cy8yMDA1LzA1L2lkZW50aXR5L2NsYWltcy9lbWFpbGFkZHJlc3MiOiJ1c2VyQGNvbnRvc28uY29tIiwiaHR0cDovL3NjaGVtYXMueG1sc29hcC5vcmcvd3MvMjAwNS8wNS9pZGVudGl0eS9jbGFpbXMvZ2l2ZW5uYW1lIjoiVG9ueSIsImh0dHA6Ly9zY2hlbWFzLnhtbHNvYXAub3JnL3dzLzIwMDUvMDUvaWRlbnRpdHkvY2xhaW1zL2hvbWVwaG9uZSI6IjU1NS4xMjEyIiwiaHR0cDovL3NjaGVtYXMubWljcm9zb2Z0LmNvbS93cy8yMDA4LzA2L2lkZW50aXR5L2NsYWltcy9yb2xlIjoiU2FsZXMiLCJzdWIiOiJib2IifQ.QW0Wfw-R5n3BHXE0vG-0giRFeB6W9oFrWJyFTaLI0qICDYx3yZ2eLXJ3zNFLVf3OG-MqytN5tqUdNfK1mRzeubqvdODHLFX36e1o3X8DR_YumyyQvgSeTJ0wwqT8PowbE3nbKfiX4TtJ4jffBelGKnL6vdx3AU2cwvLfSVp8ppA";
```
brentschmaltz: Could we build the JWT using standard claims, then we know its the same in other benchmarks.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -513,11 +537,192 @@ internal void ReadToken(string encodedJson)
...
+        /// Converts a span into an instance of <see cref="JsonWebToken"/>.
+        /// </summary>
+        /// <param name="encodedJsonSpan">A span representing a 'JSON Web Token' (JWT) in JWS or JWE Compact Serialization Format.</param>
+        /// <exception cref="SecurityTokenMalformedException">if <paramref name="encodedJsonSpan"/> is malformed, a valid JWT should have either 2 dots (JWS) or 4 dots (JWE).</exception>
+        /// <exception cref="SecurityTokenMalformedException">if <paramref name="encodedJsonSpan"/> does not have an non-empty authentication tag after the 4th dot for a JWE.</exception>
+        /// <exception cref="SecurityTokenMalformedException">if <paramref name="encodedJsonSpan"/> has more than 4 dots.</exception>
+        internal void ReadToken(ReadOnlySpan<char> encodedJsonSpan)
+        {
+            // JWT must have 2 dots for JWS or 4 dots for JWE (a.b.c.d.e)
+            ReadOnlySpan<char> encodedJsonSpanCopy = encodedJsonSpan;
```
brentschmaltz: Do we need these copies?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -513,11 +537,192 @@ internal void ReadToken(string encodedJson)
...
+                {
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(
+                        LogMessages.IDX14102,
+                        LogHelper.MarkAsUnsafeSecurityArtifact(headerSpan.ToString(), t => t.ToString())), // TODO: Add an overload to LogHelper.MarkAsUnsafeSecurityArtifact that accepts span?
+                        ex));
+                }
+
+                try
+                {
+                    Payload = CreateClaimSet(secondSpan, 0, secondSpan.Length, CreatePayloadClaimSet);
```
brentschmaltz: We should add another api to https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e7eb8b717aae2001ee4d40fc974adca6a33c4445/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs#L15 that takes a span.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -279,30 +278,21 @@ internal static unsafe byte[] UnsafeDecode(ReadOnlyMemory<char> str)
...
-                    {
-                        return Convert.FromBase64CharArray(segment.Array, segment.Offset, segment.Count);
-                    }
-                    else
-                    {
-                        bool gotString = MemoryMarshal.TryGetString(str, out string text, out int start, out int length);
-                        Debug.Assert(gotString, "Expected ReadOnlyMemory to wrap either array or string");
-                        return Convert.FromBase64String(text.Substring(start, length));
-                    }
+                    return Convert.FromBase64CharArray(strSpan.ToArray(), 0, strSpan.Length); // TODO: ToArray results in a heap allocation. Is there a better way?
```
keegan-caruso: perhaps: https://learn.microsoft.com/en-us/dotnet/api/system.convert.tryfrombase64string?view=net-8.0

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Span cannot be used as an argument to a generic type. See [this section by Adam Sitnik](https://adamsitnik.com/Span/#span-must-not-be-a-generic-type-argument).

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: The EncodedJsonSpanCopy is sliced multiple times and updated each time to extract different token parts, such as the header, payload, and authentication tag etc. The original EncodedJsonSpan is used at the end to store the underlying string into the EncodedToken property. At the moment, I cannot see a way around having a copy.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: ToArray() is no longer in use here.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Removed comments and replaced ReadyOnlyMemory with ReadOnlySpan where appropriate.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: My thinking is that the value of EncodedJsonSpanCopy.Length will fall outside the range of the string and EncodedJsonSpanCopy.Length - 1 would be more appropriate.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Removed copies where possible.

benchmark/Microsoft.IdentityModel.Benchmarks/EncodedJWTs.cs
sruke: Updated the benchmark to use standard claims

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Looking into SpanAction and how it can be used.

buildConfiguration.xml
sruke: This change is no logger applicable to dev. Will rebase.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
keegan-caruso: I don't understand. Why can't CreatePayloadClaimSet take a Span<byte>?

sruke: CreatePayloadClaimSet is an action that is performed on the decoded string. 
When you see [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blame/0b783fbda1c047c0905186a8136c407999389824/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L516) this action is of the form is a Func<byte[], int, JsonClaimSet>.
Even though, an overload can be added for CreatePayloadClaimSet to accept a span, it cannot be invoked by using `<Func<ReadOnlySpan<byte>, int, JsonClaimSet>`. It leads to a compilation error.

An alternative would be to invoke the method directly without using Func.

keegan-caruso: thanks for explaining, I think we need to refactor in a way we can pass the span.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -192,10 +221,10 @@ public string EncodedHeader
                 // TODO - need to account for JWE
                 if (_encodedHeader == null)
                 {
-                    if (EncodedToken != null)
-                        _encodedHeader = EncodedToken.Substring(0, Dot1);
+                    if (!_encodedTokenMemory.IsEmpty)
+                        _encodedHeader = _encodedTokenMemory.Span.Slice(0, Dot1).ToString();
```
sruke, line 225: Need to compare the performance of Span.ToString with String.SubString

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -279,30 +278,21 @@ internal static unsafe byte[] UnsafeDecode(ReadOnlyMemory<char> str)
...
-                    {
-                        return Convert.FromBase64CharArray(segment.Array, segment.Offset, segment.Count);
-                    }
-                    else
-                    {
-                        bool gotString = MemoryMarshal.TryGetString(str, out string text, out int start, out int length);
-                        Debug.Assert(gotString, "Expected ReadOnlyMemory to wrap either array or string");
-                        return Convert.FromBase64String(text.Substring(start, length));
-                    }
+                    return Convert.FromBase64CharArray(strSpan.ToArray(), 0, strSpan.Length);
```
sruke, line 283: Converting the span into a char Array irrespective of whether the underlying structure is an array or a string. Thoughts on this approach?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Refactored CreateClaimSet and updated CreateHeaderClaimSet and CreatePayloadClaimSet to accept spans

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
sruke: Convert.TryFromBase64String is only available starting from NET6.0.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -27,9 +28,13 @@ public partial class JsonWebToken : SecurityToken
         private string _encodedHeader;
         private string _encodedPayload;
         private string _encodedSignature;
+        private string _encodedToken;
         private string _encryptedKey;
         private string _initializationVector;
         private List<string> _audiences;
+#pragma warning disable CS0649 // Field 'JsonWebToken._encodedTokenMemory' is never assigned to, and will always have its default value
+        private ReadOnlyMemory<char> _encodedTokenMemory;
+#pragma warning restore CS0649 // Field 'JsonWebToken._encodedTokenMemory' is never assigned to, and will always have its default value
```
jennyf19: readonly

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -81,6 +86,30 @@ public JsonWebToken(string jwtEncodedString)
             ReadToken(jwtEncodedString);
         }
 
+#if NET6_0_OR_GREATER
+        /// <summary>
+        /// Initializes a new instance of <see cref="JsonWebToken"/> from a span in JWS or JWE Compact serialized format.
+        /// </summary>
+        /// <param name="encodedTokenMemory">A span containing the JSON Web Token serialized in JWS or JWE Compact format.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="encodedTokenMemory"/> is empty.</exception>
```
jennyf19: >is empty [](http://example.com/codeflow?start=90&length=8)

null or empty?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -81,6 +86,30 @@ public JsonWebToken(string jwtEncodedString)
...
+        /// <exception cref="ArgumentException"><paramref name="encodedTokenMemory"/> does not represent a valid JWS or JWE Compact serialization format.</exception>
+        /// <remarks>
+        /// See: https://datatracker.ietf.org/doc/html/rfc7519 (JWT)
+        /// See: https://datatracker.ietf.org/doc/html/rfc7515 (JWS)
+        /// See: https://datatracker.ietf.org/doc/html/rfc7516 (JWE)
+        /// <para>
+        /// The contents of the returned <see cref="JsonWebToken"/> have not been validated; the JSON Web Token is simply decoded. Validation can be performed using the methods in <see cref="JsonWebTokenHandler"/>.
+        /// </para>
+        /// </remarks>
+        public JsonWebToken(ReadOnlyMemory<char> encodedTokenMemory)
```
jennyf19, line 103: >ReadOnlyMemory [](http://example.com/codeflow?start=28&length=14)

this will help ensure the memory is not accidently modified, correct?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -192,10 +221,10 @@ public string EncodedHeader
                 // TODO - need to account for JWE
```
jennyf19, line 223: >JWE [](http://example.com/codeflow?start=46&length=3)

what is the plan for JWE? chcking the `alg` claim in the header and proceeding from there?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,14 +562,180 @@ internal void ReadToken(string encodedJson)
...
+            _encodedToken = encodedJson;
+        }
+
+#if NET6_0_OR_GREATER
+        /// <summary>
+        /// Converts a span into an instance of <see cref="JsonWebToken"/>.
+        /// </summary>
+        /// <param name="encodedTokenMemory">A span representing a 'JSON Web Token' (JWT) in JWS or JWE Compact Serialization Format.</param>
+        /// <exception cref="SecurityTokenMalformedException">if <paramref name="encodedTokenMemory"/> is malformed, a valid JWT should have either 2 dots (JWS) or 4 dots (JWE).</exception>
+        /// <exception cref="SecurityTokenMalformedException">if <paramref name="encodedTokenMemory"/> does not have an non-empty authentication tag after the 4th dot for a JWE.</exception>
```
jennyf19: >an [](http://example.com/codeflow?start=117&length=2)

a

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,14 +562,180 @@ internal void ReadToken(string encodedJson)
...
+                }
+            }
+            else
+            {
+                // JWE: https://www.rfc-editor.org/rfc/rfc7516
+                // Format: https://www.rfc-editor.org/rfc/rfc7516#page-8
+                // empty payload for JWE's {encrypted tokens}.
+                Payload = new JsonClaimSet();
+
+                if (Dot3 == encodedTokenSpan.Length) // TODO: Should this be encodedJsonSpan.Length - 1? Using encodedJsonSpan.Length doesn't add up
```
jennyf19: >Usin [](http://example.com/codeflow?start=105&length=4)

Is it because if `Dot3` is equal to `encodedTokenSpan.Length - 1`, it means that the third dot is the second-to-last character in the encoded token, which is an error?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,14 +562,180 @@ internal void ReadToken(string encodedJson)
...
+                ReadOnlySpan<char> cipherTextSpan = encodedTokenSpan.Slice(Dot3 + 1, Dot4 - Dot3 - 1);
+                if (cipherTextSpan.IsEmpty)
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogMessages.IDX14306));
+
+                try
+                {
+                    CipherTextBytes = Base64UrlEncoder.UnsafeDecode(cipherTextSpan);
+                }
+                catch (Exception ex)
+                {
```
jennyf19, line 564: >{ [](http://example.com/codeflow?start=16&length=1)

wonder if this can be broken down into smaller methods to test consolidated parts.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: This comment was already present. Need to check with Brent if this can be addressed separately.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Correct. ReadyOnlyMemory is recommended when working with memory that we don't intend to modify.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: ReadToken is a large method. Would be good to introduce smaller methods for readability and testing. However, it would add additional calls to the call graph.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -388,6 +406,68 @@ private static int ValidateAndGetOutputSize(string inputString, int offset, int
...
         }
 
+        /// <summary>
+        /// Validates the input span for decode operation.
+        /// </summary>
+        /// <param name="strSpan">String represented by a span to validate.</param>
+        /// <param name="offset">Index of char in <paramref name="strSpan"/> to start decode operation.</param>
+        /// <param name="length">Number of chars in <paramref name="strSpan"/> to decode, starting from offset.</param>
+        /// <returns>Size of the decoded bytes arrays.</returns>
+        internal static int ValidateAndGetOutputSize(ReadOnlySpan<char> strSpan, int offset, int length)
```
jennyf19, line 341: >internal [](http://example.com/codeflow?start=8&length=8)

this should be covered by it's own unit tests

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: My understanding is encodedTokenSpan.Length - 1 represents the last character and index of encodedTokenSpan.Length  would fall outside the range of the Span. That's why I'm wondering if this condition should be rewritten to `Dot3 == encodedTokenSpan.Length - 1`

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
jennyf19: run fine code coverage on one of readToken tests...I think it would be best to just unit test this to the max, and not test through other tests.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
jennyf19: does create an additional allocation. Could you use `Encoding.GetBytes` method to convert the `ReadOnlySpan<char>` to a `byte[]`. This would avoid the overhead of allocating and copying to a new `char[]`
for example:
```csharp
byte[] bytes = Encoding.UTF8.GetBytes(strSpan);
return Convert.FromBase64CharArray(bytes, 0, bytes.Length);
```
but I'm not sure.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
sruke: I will add unit tests once we align on the approach.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
jennyf19: looks like we need a bit more test coverage on this line too

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,14 +562,180 @@ internal void ReadToken(string encodedJson)
...
+                    Header = CreateHeaderClaimSet(Base64UrlEncoder.UnsafeDecode(headerSpan).AsSpan());
+                }
+                catch (Exception ex)
+                {
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(
+                        LogMessages.IDX14102,
+                        LogHelper.MarkAsUnsafeSecurityArtifact(headerSpan.ToString(), t => t.ToString())),
+                        ex));
+                }
```
jennyf19, line 515: tests needed for this

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,14 +562,180 @@ internal void ReadToken(string encodedJson)
...
+
+                // delegating retrieving encrypted Key to the getter on EncryptedKey
+                ReadOnlySpan<char> encryptedKeyBytes = encodedTokenSpan.Slice(Dot1 + 1, Dot2 - Dot1 - 1);
+                if (!encryptedKeyBytes.IsEmpty)
+                    EncryptedKeyBytes = Base64UrlEncoder.UnsafeDecode(encryptedKeyBytes);
+
+                ReadOnlySpan<char> initializationVectorSpan = encodedTokenSpan.Slice(Dot2 + 1, Dot3 - Dot2 - 1);
+                if (initializationVectorSpan.IsEmpty)
+                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogMessages.IDX14308));
```
jennyf19, line 531: > throw  [](http://example.com/codeflow?start=19&length=7)

missing tests on the exceptions here and below.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
keegan-caruso: ReadOnlyMemory is a value type

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
keegan-caruso: why are we not using: https://learn.microsoft.com/en-us/dotnet/api/system.convert.tryfrombase64chars?view=net-8.0

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,14 +562,180 @@ internal void ReadToken(string encodedJson)
...
 
         internal static JsonClaimSet CreateClaimSet(string rawString, int startIndex, int length, Func<byte[], int, JsonClaimSet> action)
         {
             return Base64UrlEncoding.Decode(rawString, startIndex, length, action);
         }
 
+        internal JsonClaimSet CreateClaimSet(ReadOnlySpan<char> strSpan, int startIndex, int length, bool createHeaderClaimSet)
+        {
+            int outputSize = Base64UrlEncoding.ValidateAndGetOutputSize(strSpan, startIndex, length);
+            byte[] output = ArrayPool<byte>.Shared.Rent(outputSize);
```
keegan-caruso, line 573: not all headers are that big, it might be worth checking size and doing a stackalloc on small sizes.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
keegan-caruso: we may want to talk to partners about expectations here.

I'm not convinced we should be using Memory<T> over Span<T> here.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,14 +562,180 @@ internal void ReadToken(string encodedJson)
...
+            if (Dot3 == -1)
+            {
+                // JWS has two dots
+                // JWS: https://www.rfc-editor.org/rfc/rfc7515
+                // Format: https://www.rfc-editor.org/rfc/rfc7515#page-7
+
+                IsSigned = !(Dot2 + 1 == encodedTokenSpan.Length);
+                try
+                {
+                    Header = CreateClaimSet(encodedTokenSpan, 0, Dot1, createHeaderClaimSet: true);
```
brentschmaltz, line 437: I question the value of adding a general CreateClaimSet, then passing a bool to control which ClaimSet to create.
This does not remove the need for CreateHeaderClaimSet of CreatePayloadClaimSet as those are called, but later.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,14 +562,180 @@ internal void ReadToken(string encodedJson)
...
+
+                // right number of dots for JWE (4)
+                byte[] headerAsciiBytes = new byte[headerSpan.Length];
+                Encoding.ASCII.GetBytes(headerSpan, headerAsciiBytes);
+
+                HeaderAsciiBytes = headerAsciiBytes;
+
+                try
+                {
+                    Header = CreateHeaderClaimSet(Base64UrlEncoder.UnsafeDecode(headerSpan).AsSpan());
```
brentschmaltz, line 507: Does this use the ArrayPool?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,14 +562,180 @@ internal void ReadToken(string encodedJson)
...
+
+            Dot1 = encodedTokenSpan.IndexOf("."); 
+            if (Dot1 == -1 || Dot1 == encodedTokenSpan.Length - 1)
+                throw LogHelper.LogExceptionMessage(new SecurityTokenMalformedException(LogMessages.IDX14100));
+
+            Dot2 = encodedTokenSpan.Slice(Dot1 + 1).IndexOf(".");
+            if (Dot2 == -1)
+                throw LogHelper.LogExceptionMessage(new SecurityTokenMalformedException(LogMessages.IDX14120));
+
+            Dot2 = Dot1 + Dot2 + 1;
```
brentschmaltz, line 589: Why modify Dot2 before calculating Dot3?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,14 +562,180 @@ internal void ReadToken(string encodedJson)
...
+#endif
 
         internal static JsonClaimSet CreateClaimSet(string rawString, int startIndex, int length, Func<byte[], int, JsonClaimSet> action)
         {
             return Base64UrlEncoding.Decode(rawString, startIndex, length, action);
         }
 
+        internal JsonClaimSet CreateClaimSet(ReadOnlySpan<char> strSpan, int startIndex, int length, bool createHeaderClaimSet)
+        {
+            int outputSize = Base64UrlEncoding.ValidateAndGetOutputSize(strSpan, startIndex, length);
```
brentschmaltz, line 572: If Base64UrlEncoding.Decode was modified to take a Span so we would not need this code here.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -510,14 +562,180 @@ internal void ReadToken(string encodedJson)
...
+        /// <param name="encodedTokenMemory">A span representing a 'JSON Web Token' (JWT) in JWS or JWE Compact Serialization Format.</param>
+        /// <exception cref="SecurityTokenMalformedException">if <paramref name="encodedTokenMemory"/> is malformed, a valid JWT should have either 2 dots (JWS) or 4 dots (JWE).</exception>
+        /// <exception cref="SecurityTokenMalformedException">if <paramref name="encodedTokenMemory"/> does not have an non-empty authentication tag after the 4th dot for a JWE.</exception>
+        /// <exception cref="SecurityTokenMalformedException">if <paramref name="encodedTokenMemory"/> has more than 4 dots.</exception>
+        internal void ReadToken(ReadOnlyMemory<char> encodedTokenMemory)
+        {
+            // JWT must have 2 dots for JWS or 4 dots for JWE (a.b.c.d.e)
+            ReadOnlySpan<char> encodedTokenSpan = encodedTokenMemory.Span;
+
+            Dot1 = encodedTokenSpan.IndexOf(".");
```
brentschmaltz: I believe it is more performant to check for '.', than ".".
char vs. string.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -27,9 +28,13 @@ public partial class JsonWebToken : SecurityToken
         private string _encodedHeader;
         private string _encodedPayload;
         private string _encodedSignature;
+        private string _encodedToken;
         private string _encryptedKey;
         private string _initializationVector;
         private List<string> _audiences;
+#pragma warning disable CS0649 // Field 'JsonWebToken._encodedTokenMemory' is never assigned to, and will always have its default value
+        private ReadOnlyMemory<char> _encodedTokenMemory;
```
brentschmaltz: Can we set this in the ctor and remove the #pragma?
We can also use the readonly keyword, this will stop accidental pointing to a new string.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
brentschmaltz, line 103: Could we add another constructor that takes a string, int so downlevel supported .net runtimes can take advantage of this improvement?

Something like:

JsonWebToken(string encodedJson, int offset)

brentschmaltz: We may not need this as string.AsMemory() is available downlevel in System.Memory.

sruke: Refactored the existing ctors to use string.AsMemory() extension and invoke ReadTokens(ReadyOnlyMemory<char> encodedToken) method.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Now that a single code path applies to all .NET versions supported by Wilson the warning is no longer applicable

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Replaced "." with '.'

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Removed method 1internal static JsonClaimSet CreateClaimSet(string rawString ...)` as it is no longer referenced

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: CreateClaimSet was kept to avoid code duplication while renting and returning an ArrayPool.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Updated

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Good idea. Set _encodedTokenMemory to be read-only!

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
sruke: Convert.TryFromBase64Chars is not available on netstandard2.0, net472 and its predecessors

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Base64UrlEncoder.UnsafeDecode(headerSpan) does use an ArrayPool for NET6.0 and greater.
I think the intent behind this question is to check if its possible to surface the ArrayPool for JWE similar to the work done for JWS. The implementation in this case is a bit more complex.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
keegan-caruso: can we profile to see the difference? if it is meaningful, it is worth the ifdef

sruke: Convert.TryFromBase64Chars is already in use under a preprocessor directive [on Ln239](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blame/6bea7839b1f3c1dfb14502a304ecc14b4782bad2/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs#L239)

keegan-caruso: ah, you are correct. thanks

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: The comment in the code is not in scope for the current PR.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
sruke: Added tests for this method

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: added test coverage for exception paths.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Regarding your comment on removing
[CreateClaimSet](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/74232825dce9009c840daa47b20e784cbcb7df3b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L568C9-L581C10)
method and adding an overload for Base64UrlEncoding.Decode that accepts a ReadyOnlySpan<char> I have implemented the suggest modification and saw a regression in memory allocation. It looks like the location of ArrayPool matters and it is more effective when it is rented within the JsonWebToken class and shows a reduction in memory allocation.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Keeping the method as it is.

benchmark/Microsoft.IdentityModel.Benchmarks/ReadTokenTests.cs
```diff
@@ -0,0 +1,69 @@
...
+                SigningCredentials = BenchmarkUtils.SigningCredentialsRsaSha256,
+                EncryptingCredentials = BenchmarkUtils.EncryptingCredentialsAes256Sha512,
+                TokenType = JwtHeaderParameterNames.Jwk,
+                Claims = BenchmarkUtils.Claims
+            };
+
+            _encryptedJWE = jsonWebTokenHandler.CreateToken(jweTokenDescriptor);
+        }
+
+        [Benchmark]
```
Saibamen:
```suggestion
        [Benchmark(Baseline = true)]
```

benchmark/Microsoft.IdentityModel.Benchmarks/ReadTokenTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            return new JsonWebToken(_encodedJWS);
+        }
+
+        [Benchmark]
+        public JsonWebToken ReadJWS_FromMemory()
+        {
+            return new JsonWebToken(_encodedJWS.AsMemory());
+        }
+
+        [Benchmark]
```
Saibamen:
```suggestion
        [Benchmark(Baseline = true)]
```

And move it to separate test class

Saibamen:
In `BenchmarkDotNet.Artifacts\results` folder, you have `*.md` file to copy and paste markdown table, no need to make screenshots

Saibamen:
`ReadJWE_FromString` has less allocations in current master branch

You can use profiling to search for more optimization: https://devblogs.microsoft.com/visualstudio/a-look-back-visual-studios-profiling-tool-advancements-in-2023/

benchmark/Microsoft.IdentityModel.Benchmarks/ReadTokenTests.cs
```diff
@@ -0,0 +1,69 @@
...
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Benchmarks
+{
+    // dotnet run -c release -f net8.0 --filter Microsoft.IdentityModel.Benchmarks.ReadTokenTests*
+
+    [Config(typeof(BenchmarkConfig))]
+    [HideColumns("Type", "Job", "WarmupCount", "LaunchCount")]
+    [MemoryDiagnoser]
+    public class ReadTokenTests
```
Saibamen:
```suggestion
    [MemoryDiagnoser]
    [RankColumn]
    public class ReadTokenTests
```

sruke:
> `ReadJWE_FromString` has less allocations in current master branch
> 
> You can use profiling to search for more optimization: https://devblogs.microsoft.com/visualstudio/a-look-back-visual-studios-profiling-tool-advancements-in-2023/

Yes I have noticed and working towards finding a way to reduce allocations using the profiler.

benchmark/Microsoft.IdentityModel.Benchmarks/ReadTokenTests.cs
sruke: The same [ReadTokens(..)](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/d5744855dcf94816a970dae70b99dcc82ffe324c/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L410C9-L410C73) method is invoked by all the JsonWebToken constructors. Making the string constructor the baseline and ranking it may not make sense.

The goal here is to maintain a single code path and ensure the existing constructors that accept a string are able to take advantage of the performance improvements from using spans.

benchmark/Microsoft.IdentityModel.Benchmarks/ReadTokenTests.cs
sruke: Separated JWS and JWE benchmarks.

Results for JWE:
``` ini

BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22631.3007), VM=Hyper-V
AMD EPYC 7763, 1 CPU, 16 logical and 8 physical cores
.NET SDK=8.0.200-preview.23624.5
  [Host]    : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2
  MediumRun : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2

LaunchCount=4  WarmupCount=10  

```
|             Method |     Mean |     Error |    StdDev |   Median |      P90 |      P95 |     P100 |   Gen0 | Allocated |
|------------------- |---------:|----------:|----------:|---------:|---------:|---------:|---------:|-------:|----------:|
| **ReadJWE_FromMemory** | **2.571 μs** | **0.0254 μs** | **0.0557 μs** | **2.547 μs** | **2.667 μs** | **2.671 μs** | **2.681 μs** | **0.1373** |   **2.29 KB** |
| **ReadJWE_FromString** | **2.622 μs** | **0.0183 μs** | **0.0410 μs** | **2.624 μs** | **2.668 μs** | **2.684 μs** | **2.704 μs** | **0.1373** |   **2.29 KB** |

and JWS:
``` ini

BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22631.3007), VM=Hyper-V
AMD EPYC 7763, 1 CPU, 16 logical and 8 physical cores
.NET SDK=8.0.200-preview.23624.5
  [Host]    : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2
  MediumRun : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2

LaunchCount=4  WarmupCount=10  

```
|             Method |     Mean |     Error |    StdDev |   Median |      P90 |      P95 |     P100 | Rank |   Gen0 | Allocated |
|------------------- |---------:|----------:|----------:|---------:|---------:|---------:|---------:|-----:|-------:|----------:|
| **ReadJWS_FromMemory** | **3.276 μs** | **0.0143 μs** | **0.0320 μs** | **3.278 μs** | **3.312 μs** | **3.333 μs** | **3.364 μs** |    **1** | **0.1602** |   **2.67 KB** |
| **ReadJWS_FromString** | **3.253 μs** | **0.0223 μs** | **0.0485 μs** | **3.279 μs** | **3.308 μs** | **3.310 μs** | **3.315 μs** |    **1** | **0.1602** |   **2.67 KB** |

benchmark/Microsoft.IdentityModel.Benchmarks/ReadTokenTests.cs
sruke: Added rank column for JWS.

sruke:
> > `ReadJWE_FromString` has less allocations in current master branch
> > You can use profiling to search for more optimization: https://devblogs.microsoft.com/visualstudio/a-look-back-visual-studios-profiling-tool-advancements-in-2023/
> 
> Yes I have noticed and working towards finding a way to reduce allocations using the profiler.



> `ReadJWE_FromString` has less allocations in current master branch
> 
> You can use profiling to search for more optimization: https://devblogs.microsoft.com/visualstudio/a-look-back-visual-studios-profiling-tool-advancements-in-2023/


> > `ReadJWE_FromString` has less allocations in current master branch
> > You can use profiling to search for more optimization: https://devblogs.microsoft.com/visualstudio/a-look-back-visual-studios-profiling-tool-advancements-in-2023/
> 
> Yes I have noticed and working towards finding a way to reduce allocations using the profiler.

@keegan-caruso found that ReadJWETokenTests is not an appropriate benchmark for assessing JWE creation performance. Claims are not populated during JsonWebToken creation since the token is encrypted; they are only populated during validation when the token is decrypted. [ValidateJWEAsyncTests](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/benchmark/Microsoft.IdentityModel.Benchmarks/ValidateJWEAsyncTests.cs) is a more suitable benchmarks that creates and validates a JWE. On running this benchmark, there is no performance concern. There is a .5% reduction in memory allocation.

sruthi/JsonWebTokenSpan
``` ini

BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22631.3007), VM=Hyper-V
AMD EPYC 7763, 1 CPU, 16 logical and 8 physical cores
.NET SDK=8.0.200-preview.23624.5
  [Host]    : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2
  MediumRun : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2

LaunchCount=4  WarmupCount=10  

```
|                                   Method |     Mean |    Error |   StdDev |      P90 |      P95 |     P100 |   Gen0 |   Gen1 | Allocated |
|----------------------------------------- |---------:|---------:|---------:|---------:|---------:|---------:|-------:|-------:|----------:|
|     **JsonWebTokenHandler_ValidateJWEAsync** | **40.20 μs** | **0.113 μs** | **0.252 μs** | **40.56 μs** | **40.63 μs** | **40.90 μs** | **0.9766** | **0.1221** |   **16.2 KB** |

dev
``` ini

BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22631.3007), VM=Hyper-V
AMD EPYC 7763, 1 CPU, 16 logical and 8 physical cores
.NET SDK=8.0.200-preview.23624.5
  [Host]    : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2
  MediumRun : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2

LaunchCount=4  WarmupCount=10  

```
|                                   Method |     Mean |    Error |   StdDev |      P90 |      P95 |     P100 |   Gen0 |   Gen1 | Allocated |
|----------------------------------------- |---------:|---------:|---------:|---------:|---------:|---------:|-------:|-------:|----------:|
|     **JsonWebTokenHandler_ValidateJWEAsync** | **40.33 μs** | **0.133 μs** | **0.286 μs** | **40.61 μs** | **40.72 μs** | **41.39 μs** | **0.9766** | **0.1221** |  **16.29 KB** |

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
```diff
@@ -14,7 +14,12 @@ public partial class JsonWebToken
     {
         internal JsonClaimSet CreatePayloadClaimSet(byte[] bytes, int length)
```
pmaytak, line 15: nit: Should this overload be removed so it's not accidentally used? Looks like it's used in tests only right now. (Same question about CreateHeaderClaimSet

build/releaseBuild.yml
```diff
@@ -202,6 +202,14 @@ jobs:
       BuildDropPath: '$(Build.SourcesDirectory)\src'
       ManifestDirPath: '$(Build.SourcesDirectory)\artifacts'
 
+  - task: NuGetCommand@2
```
jennyf19, line 205: do we need this? @jmprieur

pmaytak, line 205: nit: Should this be reverted? In MSAL we created a separate upload to VSTS pipeline so that we can control specific packages to upload to VSTS, as always uploading packages to the feed may accidentally upload packages that shouldn't be (like some packages for local dev testing or something).

benchmark/Microsoft.IdentityModel.Benchmarks/ReadJWETokenTests.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Benchmarks
+{
+    // dotnet run -c release -f net8.0 --filter Microsoft.IdentityModel.Benchmarks.ReadTokenTests*
+
+    [Config(typeof(BenchmarkConfig))]
+    [HideColumns("Type", "Job", "WarmupCount", "LaunchCount")]
+    [MemoryDiagnoser]
```
pmaytak: nit: these attributes aren't needed, BenchmarkConfig has everything.

benchmark/Microsoft.IdentityModel.Benchmarks/ReadJWSTokenTests.cs
```diff
@@ -0,0 +1,47 @@
...
+            _encodedJWS = jsonWebTokenHandler.CreateToken(jwsTokenDescriptor);
+        }
+
+        [Benchmark]
+        public JsonWebToken ReadJWS_FromString()
+        {
+            return new JsonWebToken(_encodedJWS);
+        }
+
+        [Benchmark]
```
pmaytak, line 39: nit: can also set this as baseline (and JWE tests too)
```suggestion
        [Benchmark(Baseline = true)]
```

benchmark/Microsoft.IdentityModel.Benchmarks/ReadJWSTokenTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        public JsonWebToken ReadJWS_FromMemory()
+        {
+            return new JsonWebToken(_encodedJWS.AsMemory());
```
pmaytak: Extract `_encodedJWS.AsMemory()` outside of the benchmark. Currently these benchmarks don't compare apples to apples.

build/releaseBuild.yml
keegan-caruso: This would be great to put behind a var

sruke: Re-added a condition that was previously removed

jmprieur: > do we need this? @jmprieur

I don't think so if we want to be intentional about the previews.
If we keep it (I'm fine with that for the nightly), then we absolutely need to guard it with a condition:

```
   condition: eq(variables['NugetPackageType'], 'preview')
```

jmprieur: Oh and it's not up to date with the content of dev. It won't even build any longer as the sccript don't exist.

jmprieur: so, let's remove it. you'd have a merge conflict anyways, @sruke

jennyf19: it should be removed. no longer needed

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -57,9 +57,9 @@ public static byte[] Decode(string input, int offset, int length)
         {
             _ = input ?? throw LogHelper.LogArgumentNullException(nameof(input));
 
-            int outputsize = ValidateAndGetOutputSize(input, offset, length);
+            int outputsize = ValidateAndGetOutputSize(input.AsSpan(), offset, length);
             byte[] output = new byte[outputsize];
-            Decode(input, offset, length, output);
+            Decode(input.AsSpan(), offset, length, output);
```
pmaytak: Should `input.AsSpan()` be saved into a variable and reused/passed in? (Same comment for the occurrences below.)

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -117,7 +117,7 @@ public static T Decode<T>(string input, int offset, int length, Func<byte[], int
             byte[] output = ArrayPool<byte>.Shared.Rent(outputsize);
             try
             {
-                Decode(input, offset, length, output);
+                Decode(input.AsSpan(), offset, length, output);
```
pmaytak: Should Line 116 be updated to pass span too?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -335,8 +335,21 @@ public static string Encode(byte[] input, int offset, int length)
         private static int ValidateAndGetOutputSize(string inputString, int offset, int length)
```
pmaytak: nit: can this be removed and overload with Span should be used?

test/Microsoft.IdentityModel.Tokens.Tests/Base64UrlEncodingTests.cs
```diff
@@ -307,5 +307,25 @@ public class Base64UrlEncoderTheoryData : TheoryDataBase
...
+            Assert.Throws<FormatException>(() => Base64UrlEncoding.ValidateAndGetOutputSize("abcde".AsSpan(), 0, 5));
+
+            int result = Base64UrlEncoding.ValidateAndGetOutputSize("abc".AsSpan(), 0, 0);
+            Assert.Equal(0, result);
+
+            result = Base64UrlEncoding.ValidateAndGetOutputSize("abcd".AsSpan(), 0, 4);
+            Assert.Equal(3, result);
+
+            result = Base64UrlEncoding.ValidateAndGetOutputSize("abc=".AsSpan(), 0, 4);
+            Assert.Equal(2, result);
```
pmaytak: nit: makes reading the test slightly more obvious.
```suggestion
            int actualOutputSize = Base64UrlEncoding.ValidateAndGetOutputSize("abc".AsSpan(), 0, 0);
            Assert.Equal(0, actualOutputSize);

            actualOutputSize = Base64UrlEncoding.ValidateAndGetOutputSize("abcd".AsSpan(), 0, 4);
            Assert.Equal(3, actualOutputSize);

            actualOutputSize = Base64UrlEncoding.ValidateAndGetOutputSize("abc=".AsSpan(), 0, 4);
            Assert.Equal(2, actualOutputSize);
```

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -192,10 +223,10 @@ public string EncodedHeader
                 // TODO - need to account for JWE
                 if (_encodedHeader == null)
                 {
-                    if (EncodedToken != null)
-                        _encodedHeader = EncodedToken.Substring(0, Dot1);
+                    if (!_encodedTokenMemory.IsEmpty)
+                        _encodedHeader = _encodedTokenMemory.Span.Slice(0, Dot1).ToString();
```
pmaytak, line 227: My initial thought here was why use `Span` property here and not call slice on memory - `_encodedTokenMemory.Slice(0, Dot1).ToString();`. Looking at the [source](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/ReadOnlyMemory.cs,140) seems like Span is better since if underlying object of Memory is string, Memory.ToString calls .Substring?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -347,56 +401,56 @@ public string InitializationVector
...
-            Dot2 = encodedJson.IndexOf('.', Dot1 + 1);
+            Dot2 = encodedTokenSpan.Slice(Dot1 + 1).IndexOf('.');
             if (Dot2 == -1)
                 throw LogHelper.LogExceptionMessage(new SecurityTokenMalformedException(LogMessages.IDX14120));
 
-            if (Dot2 == encodedJson.Length - 1)
-                Dot3 = -1;
-            else
-                Dot3 = encodedJson.IndexOf('.', Dot2 + 1);
+            Dot2 = Dot1 + Dot2 + 1;
```
pmaytak, line 425: Why is this needed?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -407,115 +461,123 @@ internal void ReadToken(string encodedJson)
                 // empty payload for JWE's {encrypted tokens}.
                 Payload = new JsonClaimSet();
 
-                if (Dot3 == encodedJson.Length)
+                if (Dot3 == encodedTokenSpan.Length) // TODO: Should this be encodedJsonSpan.Length - 1? 
                     throw LogHelper.LogExceptionMessage(new ArgumentException(LogMessages.IDX14121));
 
-                Dot4 = encodedJson.IndexOf('.', Dot3 + 1);
+                Dot3 = Dot2 + Dot3 + 1;
```
pmaytak, line 469: Why is this needed (and also line 473)?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -407,115 +461,123 @@ internal void ReadToken(string encodedJson)
...
 
-                // dir does not have any key bytes
-                ReadOnlyMemory<char> encryptedKeyBytes = encodedJson.AsMemory(Dot1 + 1, Dot2 - Dot1 - 1);
+                // delegating retrieving encrypted Key to the getter on EncryptedKey
+                ReadOnlySpan<char> encryptedKeyBytes = encodedTokenSpan.Slice(Dot1 + 1, Dot2 - Dot1 - 1);
                 if (!encryptedKeyBytes.IsEmpty)
                 {
                     EncryptedKeyBytes = Base64UrlEncoder.UnsafeDecode(encryptedKeyBytes);
-                    _encryptedKey = encodedJson.Substring(Dot1 + 1, Dot2 - Dot1 - 1);
+                    _encryptedKey = encodedTokenSpan.Slice(Dot1 + 1, Dot2 - Dot1 - 1).ToString();
```
pmaytak: Can `encryptedKeyBytes` be reused?
```suggestion
                    _encryptedKey = encryptedKeyBytes.ToString();
```

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
```diff
@@ -1674,6 +1698,26 @@ public void DifferentCultureJsonWebToken()
             Assert.Equal("12.2", numericList[0].Value);
             Assert.Equal("11.1", numericList[1].Value);
         }
+
+        // Test to verify equality between JsonWebTokens created from a string and an equivalent span
+        [Theory, MemberData(nameof(ParseTokenTheoryData))]
+        public void CompareJsonWebToken(JwtTheoryData theoryData)
```
pmaytak: nit: clarifies what the test does and the expected result.
```suggestion
        public void StringAndMemoryConstructors_CreateEquivalentTokens(JwtTheoryData theoryData)
```

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: The variables (Dot1, Dot2 etc.) store the locations of dots in an encoded token. These are used to delay create EncodedPayload, EncodedSignature or other segments from the encoded token. See Ln 265.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: Explained in another comment.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: My understanding was the same. The goal here is to delay converting the Memory/Span into a string until its first accessed. Some properties may never be accessed but we need to keep them as they are public.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
sruke: Good point. Moved the null checkup the call graph and removed this method.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
sruke: Done.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
sruke: AsSpan() is a lightweight operation that does not allocate additional memory. 
Considering its being invoked on the string twice, I agree that we should save it in a variable. 
Updated.

build/releaseBuild.yml
sruke: The condition on the next line gates the packages from being uploaded.

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
sruke: Considering its internal and not private I was hesitant to remove it.

build/releaseBuild.yml
sruke: Had an internal discussion and decided to keep the task. A condition was added as a gate.

benchmark/Microsoft.IdentityModel.Benchmarks/ReadJWETokenTests.cs
sruke: Removed.

benchmark/Microsoft.IdentityModel.Benchmarks/ReadJWSTokenTests.cs
Saibamen: @sruke - not resolved. Please apply this suggestion

benchmark/Microsoft.IdentityModel.Benchmarks/ReadJWETokenTests.cs
```diff
@@ -0,0 +1,44 @@
...
+using System;
+using BenchmarkDotNet.Attributes;
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Benchmarks
+{
+    // dotnet run -c release -f net8.0 --filter Microsoft.IdentityModel.Benchmarks.ReadTokenTests*
+
+    public class ReadJWETokenTests
```
Saibamen, line 13: Why you deleted `[Config(typeof(BenchmarkConfig))]` here and in `ReadJWSTokenTests`?

pmaytak: All the config is defined in one place, instead of duplicated in each class - https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/benchmark/Microsoft.IdentityModel.Benchmarks/BenchmarkConfig.cs

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
pmaytak: I debugged through a test and understand this now. Previous code indexes through the whole token string, while new code indexes only through the segment, so the previous positions have to be added to result in a dot position of a whole string.
Would adding comments like this help?
```csharp
Dot2 = encodedTokenSpan.Slice(Dot1 + 1).IndexOf('.'); // Dot2 index in the second segment
Dot2 = Dot1 + Dot2 + 1; // Dot2 index in the whole token
Dot3 = Dot2 + Dot3 + 1; // Dot3 index in the whole token
```

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
sruke: I understand your earlier question now. Performing a Slice on a span results in the index being reset to 0. Hence, the indices for each of the Dot variables must be recalculated on the whole token. I've included a comment specifically for Dot2, as adding it for every variable could introduce excessive noise.

Saibamen:
Can you paste final benchmark results here?

PR #2459, opened 2024-01-25
Use properName when reporting errors to avoid confusion.

brentschmaltz:
Small cleanup that will report the actual propertyname being processed rather than the a constant.
This comes into play when serializing case-insensitive property names for OpenIdConnectConfiguration and JsonWebKey.


jennyf19 (approved):
:shipit:

PR #2465, opened 2024-01-29
Remove linq from BaseConfigurationComparer

### LINKED ISSUE #2464: [Improvement] Remove linq from BaseConfigurationComparer
**Where is the issue?**
 * [ ] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [ x ] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [ ] S.IM.Tokens.Jwt
 * Other (please describe)
 
 Remove linq in BaseConfigurationComparer

keegan-caruso:
Fixes #2464

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
+++ b/src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 
 using System.Collections.Generic;
-using System.Linq;
 
 namespace Microsoft.IdentityModel.Tokens
 {
@@ -17,11 +16,31 @@ namespace Microsoft.IdentityModel.Tokens
                 return true;
             else if (config1 == null || config2 == null)
                 return false;
-            else if (config1.Issuer == config2.Issuer && config1.SigningKeys.Count == config2.SigningKeys.Count
-                     && !config1.SigningKeys.Select(x => x.InternalId).Except(config2.SigningKeys.Select(x => x.InternalId)).Any())
-                return true;
             else
-                return false;
+            {
+                if (config1.Issuer != config2.Issuer)
+                    return false;
+
+                if (config1.SigningKeys.Count != config2.SigningKeys.Count)
+                    return false;
+
+                foreach (var key in config1.SigningKeys)
+                {
+                    if (!ContainsKeyWithInternalId(config2, key.InternalId))
+                        return false;
+                }
+            }
+
+            return true;
+        }
+
+        private static bool ContainsKeyWithInternalId(BaseConfiguration config, string internalId)
+        {
+            foreach(var key in config.SigningKeys)
+                if (key.InternalId == internalId)
+                    return true;
+
+            return false;
         }
 
         public int GetHashCode(BaseConfiguration config)
@@ -29,10 +48,9 @@ namespace Microsoft.IdentityModel.Tokens
             int defaultHash = string.Empty.GetHashCode();
             int hashCode = defaultHash;
             hashCode ^= string.IsNullOrEmpty(config.Issuer) ? defaultHash : config.Issuer.GetHashCode();
-            foreach(string internalId in config.SigningKeys.Select(x => x.InternalId))
-            {
-                hashCode ^= internalId.GetHashCode();
-            }
+
+            foreach (var key in config.SigningKeys)
+                hashCode ^= key.InternalId.GetHashCode();
 
             return hashCode;
         }
--- a/test/Microsoft.IdentityModel.Tokens.Tests/BaseConfigurationComparerTests.cs
+++ b/test/Microsoft.IdentityModel.Tokens.Tests/BaseConfigurationComparerTests.cs
@@ -1,0 +1,130 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.IdentityModel.TestUtils;
+using Xunit;
+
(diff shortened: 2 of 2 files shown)


MZOLN (approved):

src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
```diff
@@ -17,22 +16,41 @@ public bool Equals(BaseConfiguration config1, BaseConfiguration config2)
...
+                    return false;
+
+                if (config1.SigningKeys.Count != config2.SigningKeys.Count)
+                    return false;
+
+                foreach (var key in config1.SigningKeys)
+                {
+                    if (!ContainsKeyWithInternalId(config2, key.InternalId))
+                        return false;
+                }
```
stephentoub, line 31: Be a little careful here. While I applaud the reduction of LINQ usage on a hot path, Enumerable.Except may or may not be something you want to get rid of like this. Except works by building up a HashSet for one of the inputs and then enumerating the other and checking each item against the HashSet. This makes it generally O(N + M). You've replaced it with a doubly-nested loop, where for every item in config1.SigningKeys you're iterating through config2.SigningKeys... that makes it O(N * M). If these collections are always small, that's probably not a big deal. If they could be bigger, that's a problem.  Also, every time you're iterating over SigningKeys in ContainsKeyWithInternalId, you're allocating an enumerator. So, the "fix" here may not actually be making things better.

westin-m (approved):

test/Microsoft.IdentityModel.Tokens.Tests/BaseConfigurationComparerTests.cs
```diff
@@ -0,0 +1,130 @@
...
+        public BaseConfiguration ConfigurationA { get; set; }
+
+        public BaseConfiguration ConfigurationB { get; set; }
+
+        public bool ShouldBeEqual { get; set; }
+    }
+
+    public class TestConfiguration : BaseConfiguration
+    {
+    }
```
jennyf19, line 129: nice to have a test for this :(

src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
```diff
@@ -17,22 +16,41 @@ public bool Equals(BaseConfiguration config1, BaseConfiguration config2)
...
             int defaultHash = string.Empty.GetHashCode();
             int hashCode = defaultHash;
             hashCode ^= string.IsNullOrEmpty(config.Issuer) ? defaultHash : config.Issuer.GetHashCode();
-            foreach(string internalId in config.SigningKeys.Select(x => x.InternalId))
-            {
-                hashCode ^= internalId.GetHashCode();
-            }
+
+            foreach (var key in config.SigningKeys)
+                hashCode ^= key.InternalId.GetHashCode();
```
stephentoub, line 53: This part of the fix is goodness (and what I called out in our Q&A)

jennyf19 (approved):
:shipit:  Thanks @keegan-caruso

src/Microsoft.IdentityModel.Tokens/BaseConfigurationComparer.cs
keegan-caruso: thanks for taking a look, the typical case is something like 2-5 keys so I wasn't too worried about the nested for loop.

I could also just revert this part.

keegan-caruso: resolving, these are small collections.

stephentoub: Did you measure the impact of this change on both throughput and allocation? It's quite possible it actually made things worse.  With the change in GetHashCode, it's easy to eyeball and it know "yes, this change is definitively better".  With this change here, it's much less obvious, e.g. you're replacing a few allocations for a HashSet with an enumerator allocation per config1 key plus the n-squared-ness I mentioned. You might be right that the result is better, but it's not necessarily obvious. Always good to measure these kinds of changes.

keegan-caruso: You are completely right, better to lead with data. I wrote a microbenchmark just for the equals method using real data:

// * Summary *

BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22631.3007), VM=Hyper-V
Intel Xeon Platinum 8370C CPU 2.80GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK=8.0.101
  [Host]     : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.1 (8.0.123.58001), X64 RyuJIT AVX2


|                            Method |     Mean |   Error |  StdDev |   Gen0 | Allocated |
|---------------------------------- |---------:|--------:|--------:|-------:|----------:|
|         BaseConfigurationComparer | 461.8 ns | 1.13 ns | 0.95 ns | 0.0172 |     440 B |
| BaseConfigurationComparerWithLinq | 926.3 ns | 5.82 ns | 5.44 ns | 0.0410 |    1048 B |

stephentoub: If we're confident this is representative, great.

PR #2470, opened 2024-01-30
Replace propertyName with MetadataName constant.

brentschmaltz:
The propertyName was being compared to itself, so we would have a failure here.

#2471

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Json/OpenIdConnectConfigurationSerializer.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Json/OpenIdConnectConfigurationSerializer.cs
@@ -326,7 +326,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
                             // TODO these two properties are per spec 'boolean', we shipped 6x with them as string, if we change we may break folks.
                             // probably best to mark the property obsolete with the gentle tag, then open up another property and keep them in sync,
                             // remove the obsolete in 8.x
-                            else if (propertyName.Equals(propertyName, StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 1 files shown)


jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Json/OpenIdConnectConfigurationSerializer.cs
```diff
@@ -326,7 +326,7 @@ public static OpenIdConnectConfiguration Read(ref Utf8JsonReader reader, OpenIdC
                             // TODO these two properties are per spec 'boolean', we shipped 6x with them as string, if we change we may break folks.
                             // probably best to mark the property obsolete with the gentle tag, then open up another property and keep them in sync,
                             // remove the obsolete in 8.x
-                            else if (propertyName.Equals(propertyName, StringComparison.OrdinalIgnoreCase))
+                            else if (propertyName.Equals(MetadataName.FrontchannelLogoutSessionSupported, StringComparison.OrdinalIgnoreCase))
```
keegan-caruso, line 329: would be great to modify the existing tests to verify this

keegan-caruso:
@brentschmaltz can you create a bug to track?

jennyf19:
> @brentschmaltz can you create a bug to track?

done: #2471

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Json/OpenIdConnectConfigurationSerializer.cs
brentschmaltz: I agree we need test, will add with the larger serialization PR.

PR #2478, opened 2024-02-02
Link to breaking change announcement in IDX10506

halter73:
When an IDX10506 exception is thrown from JsonWebTokenHandler there's a good chance this is due to a breaking change to ASP.NET Core 8.

This adds a link to the breaking change announcement at https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/8.0/securitytoken-events

A few people have run into this already. See https://stackoverflow.com/questions/77515249/custom-token-validator-not-working-in-net-8 and https://developercommunity.visualstudio.com/t/ASPNET-Core-8:-Delegate-SignatureValida/10525320

@jennyf19


jennyf19 (approved):
:shipit:

jennyf19:
included in [7.3.1 release](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/releases/tag/7.3.1)

PR #2480, opened 2024-02-03
Marcinzo/issuervalidatormetadata

MZOLN:
# PoC injecting metadata configuration managers from outside for single point cache management
This PR adds injection point to AadIssuerValidator for ConfigurationManager

## Description
This PR allows for centralized cache management, including http policies, lifetime policies, etc. and reduces the amount of copies 
of cached data.

This is a conservative change, that still utilizes prev mechanism for a fallback in case the cache cannot be retrieved.

{Detail}

Fixes #{bug number} (in this specific format)

### DIFF
--- a/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
+++ b/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
@@ -55,6 +55,8 @@ namespace Microsoft.IdentityModel.Validators
         private string _aadAuthorityV2;
         private BaseConfigurationManager _configurationManagerV1;
         private BaseConfigurationManager _configurationManagerV2;
+        private IssuerLastKnownGood _issuerLKGV1;
+        private IssuerLastKnownGood _issuerLKGV2;
 
         internal BaseConfigurationManager ConfigurationManagerV1
         {
@@ -196,27 +198,30 @@ namespace Microsoft.IdentityModel.Validators
             {
                 var isV2Issuer = IsV2Issuer(securityToken);
                 var effectiveConfigurationManager = GetEffectiveConfigurationManager(isV2Issuer);
-                var baseConfiguration = await GetBaseConfiguration(effectiveConfigurationManager, validationParameters).ConfigureAwait(false);
 
-                // Decide whether to take the issuer to validate against from LKG or not.
-                var aadIssuer = baseConfiguration.Issuer;
-                if (validationParameters.ValidateWithLKG && effectiveConfigurationManager.LastKnownGoodConfiguration != null &&
-                    (_configurationManagerProvider == null || isV2Issuer == IsV2Authority))
-                    // If asked to use LKG,
-                    // Use the issuer from LKG if using self-managed state.
-                    // When using configurationManagerProvider, use LKG only if authority and token version matches;
-                    // otherwise, keep using the resolved issuer from baseConfiguration.
-                    aadIssuer = effectiveConfigurationManager.LastKnownGoodConfiguration.Issuer;
+                string aadIssuer = null;
+                if (validationParameters.ValidateWithLKG)
+                {
+                    // returns null if LKG issuer expired
+                    aadIssuer = GetEffectiveLKGIssuer(isV2Issuer);
+                }
+                else
+                {
+                    var baseConfiguration = await GetBaseConfiguration(effectiveConfigurationManager, validationParameters).ConfigureAwait(false);
+                    aadIssuer = baseConfiguration.Issuer;
+                }
 
-                var isIssuerValid = IsValidIssuer(aadIssuer, tenantId, issuer);
+                if (aadIssuer != null)
+                {
+                    var isIssuerValid = IsValidIssuer(aadIssuer, tenantId, issuer);
 
-                // The original LKG assignment behavior for previous self-state management.
-                if (_configurationManagerProvider == null && isIssuerValid && !validationParameters.ValidateWithLKG)
-                    effectiveConfigurationManager.LastKnownGoodConfiguration = baseConfiguration;
+                    // The original LKG assignment behavior for previous self-state management.
+                    if (isIssuerValid && !validationParameters.ValidateWithLKG)
+                        SetEffectiveLKGIssuer(aadIssuer, isV2Issuer, effectiveConfigurationManager.LastKnownGoodLifetime);
 
-                if (isIssuerValid)
-                    return issuer;
-
+                    if (isIssuerValid)
+                        return issuer;
+                }
             }
             catch (Exception ex)
             {
@@ -343,6 +348,31 @@ namespace Microsoft.IdentityModel.Validators
             {
                 return validIssuerTemplate == actualIssuer;
             }
+        }
+
+        private void SetEffectiveLKGIssuer(string aadIssuer, bool isV2Issuer, TimeSpan lastKnownGoodLifetime)
+        {
+            var issuerLKG = new IssuerLastKnownGood
+            {
+                Issuer = aadIssuer,
+                LastKnownGoodLifetime = lastKnownGoodLifetime
+            };
+
+            if (isV2Issuer)
+                _issuerLKGV2 = issuerLKG;
+            else
+                _issuerLKGV1 = issuerLKG;
+        }
+
+        private string GetEffectiveLKGIssuer(bool isV2Issuer)
+        {
+            var effectiveLKGIssuer = isV2Issuer ? _issuerLKGV2 : _issuerLKGV1;
+            if (effectiveLKGIssuer != null && effectiveLKGIssuer.IsValid)
+            {
+                return effectiveLKGIssuer.Issuer;
+            }
+
+            return null;
         }
 
         private static bool IsV2Issuer(SecurityToken securityToken)
--- a/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
+++ b/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
@@ -27,7 +27,7 @@ namespace Microsoft.IdentityModel.Validators
         internal const string V2EndpointSuffixWithTrailingSlash = $"{V2EndpointSuffix}/";
         internal const string TenantIdTemplate = "{tenantid}";
 
-        private Func<string, CancellationToken, Task<BaseConfigurationManager>> _configurationManagerProvider;
+        private Func<string, BaseConfigurationManager> _configurationManagerProvider;
 
         internal AadIssuerValidator(
             HttpClient httpClient,
@@ -39,9 +39,10 @@ namespace Microsoft.IdentityModel.Validators
         }
 
         internal AadIssuerValidator(
-            string aadAuthority,
-            Func<string, CancellationToken, Task<BaseConfigurationManager>> configurationManagerProvider)
-            : this(null, aadAuthority)
+            HttpClient httpClient,
+            string aadAuthority,            
+            Func<string, BaseConfigurationManager> configurationManagerProvider)
+            : this(httpClient, aadAuthority)
         {
             _configurationManagerProvider = configurationManagerProvider;
         }
@@ -190,8 +191,6 @@ namespace Microsoft.IdentityModel.Validators
 
             try
             {
-
-                // replace here
                 var effectiveConfigurationManager = GetEffectiveConfigurationManager(securityToken);
                 var baseConfiguration = await GetBaseConfiguration(effectiveConfigurationManager, validationParameters).ConfigureAwait(false);
                 string aadIssuer = baseConfiguration.Issuer;
@@ -270,6 +269,7 @@ namespace Microsoft.IdentityModel.Validators
         /// Gets an <see cref="AadIssuerValidator"/> for an Azure Active Directory (AAD) authority.
         /// </summary>
         /// <param name="aadAuthority">The authority to create the validator for, e.g. https://login.microsoftonline.com/. </param>
+        /// <param name="httpClient">Optional HttpClient to use to retrieve the endpoint metadata (can be null).</param>
         /// <param name="configurationManagerProvider">Configuration manager provider. Injection point for metadata managed outside of the class.</param>
         /// <example><code>
         /// AadIssuerValidator aadIssuerValidator = AadIssuerValidator.GetAadIssuerValidator(authority, configurationManagerProvider);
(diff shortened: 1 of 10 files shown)


src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -180,18 +191,21 @@ internal string AadAuthorityV2
...
                 {
                     if (IsValidIssuer(aadIssuer, tenantId, issuer))
                     {
-                        effectiveConfigurationManager.LastKnownGoodConfiguration = new OpenIdConnectConfiguration() { Issuer = aadIssuer };
+                        // We need to update LKG on the self-managed config manager.
+                        // The LKG of the one coming from provider, should be handled by the provider after successful validation
+                        // unless it is only fetched for the issuer and it's different than the authority.
+                        if (_configurationManagerProvider == null)
+                        {
+                            effectiveConfigurationManager.LastKnownGoodConfiguration = new OpenIdConnectConfiguration() { Issuer = aadIssuer };
```
MZOLN: This is unfinished and needs more considerations. E.g. how should this be set in case of config fetched from provider? Shouldnt provider self manage this?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -306,11 +343,35 @@ private static bool IsValidIssuer(string validIssuerTemplate, string tenantId, s
...
+            if (_configurationManagerProvider != null)
+            {
+                var aadAuthority = isV2Issuer ? AadAuthorityV2 : AadAuthorityV1;
+                var configurationManager = _configurationManagerProvider(aadAuthority);
+
+                if (configurationManager != null)
+                    return configurationManager;
+            }
+
+            // If no provider or provider returned null, fallback to previous strategy
```
MZOLN, line 365: This is where the fallback happens in case configurationManager fetched from provider was null. Is this a desired behavior?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -36,6 +38,15 @@ public class AadIssuerValidator
             IsV2Authority = aadAuthority.Contains(V2EndpointSuffix);
         }
 
+        internal AadIssuerValidator(
+            HttpClient httpClient,
+            string aadAuthority,
```
brentschmaltz: can you remove the whitespace?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -180,18 +191,24 @@ internal string AadAuthorityV2
...
             try
             {
-                BaseConfigurationManager effectiveConfigurationManager = GetEffectiveConfigurationManager(securityToken);
-                if (validationParameters.RefreshBeforeValidation)
-                    effectiveConfigurationManager.RequestRefresh();
+                var isV2Issuer = IsV2Issuer(securityToken);
 
-                BaseConfiguration configuration = await effectiveConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false);
-                string aadIssuer = configuration.Issuer;
+                var effectiveConfigurationManager = GetEffectiveConfigurationManager(isV2Issuer);
```
brentschmaltz, line 200: Could we just pass the token to GetEffectiveConfigurationManager and bypass the extra call to IsV2Issuer?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -180,18 +191,24 @@ internal string AadAuthorityV2
...
 
                 if (!validationParameters.ValidateWithLKG)
                 {
                     if (IsValidIssuer(aadIssuer, tenantId, issuer))
                     {
-                        effectiveConfigurationManager.LastKnownGoodConfiguration = new OpenIdConnectConfiguration() { Issuer = aadIssuer };
+                        // We need to update LKG on the self-managed config manager.
+                        // The LKG of the one coming from provider, should be handled by the provider after successful validation
+                        // unless it is only fetched for the issuer and it's different than the authority.
+                        if (_configurationManagerProvider == null ||
```
brentschmaltz: not understanding the conditions here, why not set LastKnownGood always?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -180,18 +191,24 @@ internal string AadAuthorityV2
...
-                BaseConfigurationManager effectiveConfigurationManager = GetEffectiveConfigurationManager(securityToken);
-                if (validationParameters.RefreshBeforeValidation)
-                    effectiveConfigurationManager.RequestRefresh();
+                var isV2Issuer = IsV2Issuer(securityToken);
 
-                BaseConfiguration configuration = await effectiveConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false);
-                string aadIssuer = configuration.Issuer;
+                var effectiveConfigurationManager = GetEffectiveConfigurationManager(isV2Issuer);
+                var baseConfiguration = await GetBaseConfiguration(effectiveConfigurationManager, validationParameters).ConfigureAwait(false);
+                string aadIssuer = baseConfiguration.Issuer;
```
brentschmaltz: can we just pass baseConfiguration.Issuer rather than create this local variable?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
brentschmaltz: we seem to be missing the logic: 
 if (validationParameters.RefreshBeforeValidation)
                    effectiveConfigurationManager.RequestRefresh();

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -36,6 +38,15 @@ public class AadIssuerValidator
             IsV2Authority = aadAuthority.Contains(V2EndpointSuffix);
         }
 
+        internal AadIssuerValidator(
+            HttpClient httpClient,
+            string aadAuthority,            
+            Func<string, BaseConfigurationManager> configurationManagerProvider)
+            : this(httpClient, aadAuthority)
+        {
+            _configurationManagerProvider = configurationManagerProvider;
```
brentschmaltz, line 50: check for null, passing null is obviously a programming error.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
MZOLN: The scenario tests in SAL cover this.

This is what I was trying to raise during our conversation hinting at a little design weakness in LKG. 

In case the authority is the same as the token, the metadata is already fetched before the AadIssuerValidator is called but the LKG is set by an outside scope only after we verify the whole token on it [link](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/c3487c20fc5431e51c85020205598f40546bf794/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L662C16-L669C18). In this case, if we were to set the LKG here, we could affect the LKG integrity if an error would occur with metadata and the token after the issuer was validated. That is why I have opted to respect that and workaround it with not setting the LKG for the case when the metadata is fetched outside of AadIssuerValidator.

Now, in case of the token version being different than the authority, the issuer validator needs to request a new metadata that is not handled anywhere else (and currently today wont be used for anything else) so nobody would set its LKG anywhere, therefore this needs to be set here.

This seemed like a pragmatic approach and both scenarios are covered in SAL scenario test cases.

brentschmaltz (changes requested):
examine if LKG should be set in AADIssuerValidator.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
MZOLN: GetEffectiveConfigurationManager does not call IsV2Issuer nor uses the token. It only needs information on the issuer version and it's passed as bool.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
MZOLN: I've readjusted this part of the method for readability and ended up using this variable more

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -36,11 +38,25 @@ public class AadIssuerValidator
             IsV2Authority = aadAuthority.Contains(V2EndpointSuffix);
         }
 
+        internal AadIssuerValidator(
+            HttpClient httpClient,
```
keegan-caruso, line 42: does it make sense to accept the http client here? It looks like it will be unused?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/IssuerLastKnownGood.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        /// <summary>
+        /// Gets an indicator whether the value is still within its lifetime and is valid.
+        /// </summary>
+        public bool IsValid
+        {
+            get
+            {
+                return _lastKnownGoodConfigFirstUse + LastKnownGoodLifetime > DateTime.UtcNow;
```
keegan-caruso, line 64: I know we usually don't in this library, but would it make sense to do something like: https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider?view=net-8.0

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -180,25 +196,30 @@ internal string AadAuthorityV2
...
                 {
-                    if (effectiveConfigurationManager.LastKnownGoodConfiguration != null &&
-                        IsValidIssuer(effectiveConfigurationManager.LastKnownGoodConfiguration.Issuer, tenantId, issuer))
+                    var baseConfiguration = await GetBaseConfiguration(effectiveConfigurationManager, validationParameters).ConfigureAwait(false);
+                    aadIssuer = baseConfiguration.Issuer;
+                }
+
+                if (aadIssuer != null)
+                {
+                    var isIssuerValid = IsValidIssuer(aadIssuer, tenantId, issuer);
```
keegan-caruso, line 216: would something like this be more clear? 

```cs
var isIssuerValid = IsValidIssuer(aadIssuer, tenantId, issuer);

// The original LKG assignment behavior for previous self-state management.
if (isIssuerValid)
{
    if (!validationParameters.ValidateWithLKG)
        SetEffectiveLKGIssuer(aadIssuer, isV2Issuer, effectiveConfigurationManager.LastKnownGoodLifetime);

    return issuer;
}
```

keegan-caruso: It is another level of nesting, but I find it a little easier to follow. Up to you in the end.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -258,7 +269,40 @@ public static AadIssuerValidator GetAadIssuerValidator(string aadAuthority, Http
...
+            if (configurationManagerProvider != null)
+                return new AadIssuerValidator(
+                    httpClient,
+                    aadAuthority,
+                    configurationManagerProvider);
+
+            if (s_issuerValidators.TryGetValue(aadAuthority, out AadIssuerValidator aadIssuerValidator))
+                return aadIssuerValidator;
+
+            s_issuerValidators[aadAuthority] = new AadIssuerValidator(
```
keegan-caruso, line 301: should improve docs that the cache won't be used with the provider func.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -306,11 +350,58 @@ private static bool IsValidIssuer(string validIssuerTemplate, string tenantId, s
...
+        }
+
+        private BaseConfigurationManager GetEffectiveConfigurationManager(bool isV2Issuer)
+        {        
+            if (_configurationManagerProvider != null)
+            {
+                var aadAuthority = isV2Issuer ? AadAuthorityV2 : AadAuthorityV1;
+                var configurationManager = _configurationManagerProvider(aadAuthority);
+
+                if (configurationManager != null)
```
keegan-caruso, line 391: This is interesting to me, what is the expected contract here? Should this ever return null?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -180,25 +196,30 @@ internal string AadAuthorityV2
...
-                    effectiveConfigurationManager.RequestRefresh();
+                var isV2Issuer = IsV2Issuer(securityToken);
+                var effectiveConfigurationManager = GetEffectiveConfigurationManager(isV2Issuer);
 
-                BaseConfiguration configuration = await effectiveConfigurationManager.GetBaseConfigurationAsync(CancellationToken.None).ConfigureAwait(false);
-                string aadIssuer = configuration.Issuer;
-
-                if (!validationParameters.ValidateWithLKG)
+                string aadIssuer = null;
+                if (validationParameters.ValidateWithLKG)
```
brentschmaltz, line 203: Love it that you changed the logic from negative to positive.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -258,7 +269,40 @@ public static AadIssuerValidator GetAadIssuerValidator(string aadAuthority, Http
...
+        /// <param name="aadAuthority">The authority to create the validator for, e.g. https://login.microsoftonline.com/. </param>
+        /// <param name="httpClient">Optional HttpClient to use to retrieve the endpoint metadata (can be null).</param>
+        /// <param name="configurationManagerProvider">Configuration manager provider. Injection point for metadata managed outside of the class.</param>
+        /// <example><code>
+        /// AadIssuerValidator aadIssuerValidator = AadIssuerValidator.GetAadIssuerValidator(authority, configurationManagerProvider);
+        /// TokenValidationParameters.IssuerValidator = aadIssuerValidator.Validate;
+        /// </code></example>
+        /// <returns>A <see cref="AadIssuerValidator"/> for the aadAuthority.</returns>
+        /// <exception cref="ArgumentNullException">if <paramref name="aadAuthority"/> is null or empty.</exception>
+        public static AadIssuerValidator GetAadIssuerValidator(string aadAuthority, HttpClient httpClient, Func<string, BaseConfigurationManager> configurationManagerProvider)
```
brentschmaltz: Would we be able to make this internal for now?

brentschmaltz (approved):

keegan-caruso (approved):

PR #2492, opened 2024-02-21
Use user-provided values as Utf8

### LINKED ISSUE #2495: [Feature Request] Convert user-provided values to UTF8 and use these to compare with UTF8 values from JWT
NET stores 'string' type as UTF16, while an incoming JWT is UTF8. Currently the inefficiency is that we convert the JWT values into UTF16, store them, then preform validation.

- Wherever (TokenValidationParameters) user-provided input (ex. audience, issuer) exists, create an associated internal ReadOnlySpan<byte> property of Utf8.
- [When reading values](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e983060a7a610deb791f970954cfec17d9960c18/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs#L31) from an incoming token, validate the values immediately after reading them and fail fast if validation fails.
- When doing the validation use the Utf8-valued property for comparison rather than a string.
- There could be a large number of custom claims in a JWT, meaning we can't create a property for each. Add an extensibility point, like a user-passed in delegate, which would dynamically validate those custom claims.
- Currently on get, the public properties [retrieve the value from the payload](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e983060a7a610deb791f970954cfec17d9960c18/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L933). We should instead ToString() the aforementioned Utf8 properties on demand
- Investigate if using [Utf8JsonReader.ValueSpan](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.valuespan?view=net-8.0) can be used when comparing values.

Benchmark of CompareUtf8 vs CompareUtf16:
| Method       | Toolchain              | IterationCount | RunStrategy | Mean     | Error   | StdDev  | Median   | P90      | P95      | P100     | Gen0   | Allocated |
|------------- |----------------------- |--------------- |------------ |---------:|--------:|--------:|---------:|---------:|---------:|---------:|-------:|----------:|
| CompareUtf16 | Default                | Default        | Throughput  | 223.7 ns | 1.14 ns | 0.95 ns | 223.9 ns | 224.9 ns | 225.0 ns | 225.1 ns | 0.0186 |     312 B |
| CompareUtf8  | Default                | Default        | Throughput  | 167.7 ns | 0.87 ns | 0.68 ns | 167.8 ns | 168.4 ns | 168.5 ns | 168.6 ns | 0.0072 |     120 B |
| CompareUtf16 | InProcessEmitToolchain | 15             | Default     | 217.5 ns | 0.44 ns | 0.64 ns | 217.6 ns | 218.2 ns | 218.8 ns | 219.1 ns | 0.0186 |     312 B |
| CompareUtf8  | InProcessEmitToolchain | 15             | Default     | 174.6 ns | 1.78 ns | 2.61 ns | 173.2 ns | 177.9 ns | 178.2 ns | 179.4 ns | 0.0072 |     120 B |

"What we see is about a 25% gain in performance and a greater than 50% reduction in memory."

Related to #2581, #2583.

pmaytak:
Fixes #2495

- Adds a `ReadOnlyMemory<byte>` property for UTF8 issuer in TokenValidationParameters, JsonWebToken, BaseConfiguration.
   - Uses these properties for comparison validation in Validators.ValidateIssuer
- When creating a JsonWebToken, assign the current reader.ValueSpan to the UTF8 issuer property.


src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -893,11 +894,23 @@ public string RoleClaimType
...
+        {
+            get => _validIssuer;
+            set
+            {
+                _validIssuer = value;
+                ValidIssuerUtf8 = value != null ? Encoding.UTF8.GetBytes(value) : null;
+            }
+        }
+
+        internal ReadOnlySpan<byte> ValidIssuerUtf8 { get; set; }
```
pmaytak: CS8345: Field or auto-implemented property cannot be of ReadOnlySpan<byte> unless it is an instance member of a ref struct.

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
```diff
@@ -73,8 +74,15 @@ internal JsonClaimSet CreatePayloadClaimSet(byte[] bytes, int length)
                     }
                     else if (reader.ValueTextEquals(JwtPayloadUtf8Bytes.Iss))
                     {
-                        _iss = JsonSerializerPrimitives.ReadString(ref reader, JwtRegisteredClaimNames.Iss, ClassName, true);
-                        claims[JwtRegisteredClaimNames.Iss] = _iss;
+                        IssuerUtf8 = Encoding.UTF8.GetBytes(JsonSerializerPrimitives.ReadString(ref reader, JwtRegisteredClaimNames.Iss, ClassName, true));
```
pmaytak: Better way to get bytes from stream?

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -893,11 +894,23 @@ public string RoleClaimType
...
         public IEnumerable<string> ValidAudiences { get; set; }
 
+        private string _validIssuer;
+
         /// <summary>
         /// Gets or sets a <see cref="string"/> that represents a valid issuer that will be used to check against the token's issuer.
         /// The default is <c>null</c>.
         /// </summary>
-        public string ValidIssuer { get; set; }
+        public string ValidIssuer
```
sruke, line 903: Span<T> including ReadOnlySpan<T> can only be allocated on a stack. This prevents them from being a member of a class as that requires a heap allocation. Consider introducing a ReadOnlyMemory<T> to store the UTF8 representation of an issuer. 

ReadOnlyMemort<T>.AsSpan().Equals(..) can then be used during issuer validation.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
sruke, line 913: Span{T} (including ReadOnlySpan{T}) can only be allocated on a stack. This prevents them from being a member of a class as that requires a heap allocation. Consider introducing a ReadOnlyMemory<T> to store the UTF8 representation of an issuer. 

ReadOnlyMemory<T>.AsSpan().Equals(..) can then be used during issuer validation.

sruke: Span{T} (including ReadOnlySpan{T}) can only be allocated on a stack. This prevents them from being a member of a class as that requires a heap allocation. Consider introducing a ReadOnlyMemory<T> to store the UTF8 representation of an issuer. 

ReadOnlyMemory<T>.AsSpan().Equals(..) can then be used during issuer validation.

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
brentschmaltz: We don't want to call ReadString, as you are essentially translating utf8 -> utf16 -> utf8.

Have a look to see if we can use ValueSpan see: https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.valuespan?view=net-8.0

keegan-caruso: this is going to be wasted work, utf8JsonReader -> utf16 string -> utf8 bytes.

It is a larger refactor, but would be curious what the perf would look like if we used: https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.trygetbytesfrombase64?view=net-8.0

keegan-caruso: Do you see a better way to directly get bytes from a utf8JsonReader?

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
brentschmaltz, line 903: If we can save off the utf8 bytes in a ValueSpan, then we can adjust the validation of the issuer to check for the uft8 span.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
pmaytak: Yea, I was thinking using ReadOnlyMemory, but wasn't sure how to convert it to Span and if it's efficient. Thanks for the tip. There's [ReadOnlyMemory<T>.Span](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/ReadOnlyMemory.cs,195) property.

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
pmaytak: Yea, this is just a placeholder code. Was looking for a good way to get bytes out of the reader.
Seems like ValueSpan has some [restrictions](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.valuespan?view=net-8.0#remarks):

> There is no guarantee that this property will always contain well-formed data. If the input JSON passed in to the Utf8JsonReader contains invalid UTF-8 bytes within JSON string tokens, this property will return back those invalid UTF-8 bytes as is. Therefore, if the input is untrusted or not previously validated, call [GetString()](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.getstring?view=net-8.0#system-text-json-utf8jsonreader-getstring) to get the JSON string token.
> 
> Since this property returns the raw bytes, avoid using it for text comparison. Instead call [ValueTextEquals](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.valuetextequals?view=net-8.0) which unescapes the text if necessary.

pmaytak: So if the validation is done [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/051d164e3c025a0d7276f1d6acf38c902a4893fc/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L848), I assume we'd have to refactor that to be a be able to use that method here (since it requires a token and a TVPs). So we'd have to pass TVP into JsonWebToken constructor?

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
pmaytak: I don't fully understand what you mean. Do you mean as internally everywhere we should use `ReadOnlySpan<byte> ValidIssuerUtf8` when we can instead of the string property?

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
```diff
@@ -78,8 +79,10 @@ internal JsonClaimSet CreatePayloadClaimSet(ReadOnlySpan<byte> byteSpan)
                     }
                     else if (reader.ValueTextEquals(JwtPayloadUtf8Bytes.Iss))
                     {
-                        _iss = JsonSerializerPrimitives.ReadString(ref reader, JwtRegisteredClaimNames.Iss, ClassName, true);
-                        claims[JwtRegisteredClaimNames.Iss] = _iss;
+                        
+                        IssuerUtf8 = JsonSerializerPrimitives.ReadStringUtf8(ref reader, JwtRegisteredClaimNames.Jti, ClassName, true).ToArray();
```
pmaytak, line 83: Using the bytes in ValueSpan straight up doesn't work if the characters are escaped. Calling reader.GetString actually unescapes the characters if needed.
There is a reader.ValueIsEscaped property but it's only available on .NET 8+.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -1031,14 +1030,9 @@ public DateTime IssuedAt
...
         /// </remarks>
-        public override string Issuer
-        {
-            get
-            {
-                _iss ??= Payload.GetStringValue(JwtRegisteredClaimNames.Iss);
-                return _iss;
-            }
-        }
+        public override string Issuer => Encoding.UTF8.GetString(IssuerUtf8.ToArray());
```
pmaytak, line 1033: - Encoding.UTF8.GetString with ReadOnlySpan overload is only available on .NET Core 2.1+ (not Net Fx).
   - I assume ToArray allocates a new array.
- Since IssuerUtf8 can change, we can't "cache" the string representation, so have to convert it to string each time.

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -240,17 +241,20 @@ internal static string ValidateIssuer(string issuer, SecurityToken securityToken
...
-            string issuer,
+            ReadOnlyMemory<byte> issuerUtf8,
             SecurityToken securityToken,
             TokenValidationParameters validationParameters,
             BaseConfiguration configuration)
         {
             if (validationParameters == null)
                 throw LogHelper.LogArgumentNullException(nameof(validationParameters));
 
+            string issuer = Encoding.UTF8.GetString(issuerUtf8.ToArray());
```
pmaytak, line 252: We'll still need to convert the Memory to string in places like here.

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
pmaytak: Use https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.copystring?view=net-8.0#system-text-json-utf8jsonreader-copystring(system-span((system-byte)))

Issue #2418, opened 2023-11-24
[Bug] custom SignatureProvider breaks after upgrading .NET 8


superyyrrzz:
**Repro**

Our services met an outage due to a breaking change in .NET 8 (https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/8.0/securitytoken-events) related to [SignatureProvider](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/SignatureProvider.cs) Change.

In our service's existing code in .NET 6, we inherite `SignatureProvider` and override `public abstract bool Verify(byte[] input, byte[] signature)`.

However due to the breaking change in .NET 8, this method is not invoked. Instead, the new `JsonWebTokenHandler` calls the new method `public virtual bool Verify(byte[] input, int inputOffset, int inputLength, byte[] signature, int signatureOffset, int signatureLength)`. This new method throws `NotImplementedException` as we are not aware to implement it.

**Expected behavior**

I propose to update the signature from `virtual` to `abstract`. Usually this is not friendly by forcing all the inheritance to implement a new method. However in this case, a compilation error to devs is more preferred over a runtime exception affecting our customers. If we can notice in advance when upgrading .NET 8, we may possiblely turn on `UseSecurityTokenValidators` to avoid breaking.

```
public abstract bool Verify(byte[] input, int inputOffset, int inputLength, byte[] signature, int signatureOffset, int signatureLength)
```

**Actual behavior**
The new virtual method is invoked by default in .NET 8's JsonWebToken, which throws `NotImplementedException`.

yodasad:
The same issue with
```
public abstract byte[] Sign(byte[] input);
```

This method is no longer called by `JsonWebTokenHandler ` targeting .NET 6 and higher, the method
```
internal virtual bool Sign(ReadOnlySpan<byte> data, Span<byte> destination, out int bytesWritten)
```
is called instead and there is no way one can override it since it is internal.

KelDazan:
Experiencing same issue with `Sign` method.
I am using a custom `CryptoProviderFactory` with custom `ICryptoProvider`, that provides custom implementation of `SignatureProvider`.
Rolling back to `Microsoft.IdentityModel.JsonWebTokens 7.1.2` helps, but i would like to have all security packages up to date, for obvious reasons.

[...]

PR #2497, opened 2024-02-24
Move virtual methods to public in Signature Provider

keegan-caruso:
These methods need to be moved to public
so that the extensibility story works for user provider signature providers.

related to #2418


HarmanDhunna (approved):

jennyf19 (approved):
:shipit:

jmprieur (approved):
Lgtm
Thanks @keegan-caruso

PR #2501, opened 2024-02-27
Add test case. Fix Dot3 in read JWT.

### LINKED ISSUE #2507: [Bug] JsonWebToken.ReadToken doesn't correctly check dot 3 in JWE
7.4.0

**Where is the issue?**
 * [x] M.IM.JsonWebTokens

**Actual behavior**
This check incorrectly checks if dot 3 is at the last index (in a malformed JWE). The code still works since JWE in `a.b.c.` format will be caught when dot 4 is checked.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/fc61f2ca8126fcd1e1de83405a1f9acddc108971/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L466

pmaytak:
Fixes #2507
- Added small missing test case for reading JsonWebToken.
- Fixed an issue with Dot3 index.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -463,11 +463,10 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 // empty payload for JWE's {encrypted tokens}.
                 Payload = new JsonClaimSet();
 
-                if (Dot3 == encodedTokenSpan.Length) // TODO: Should this be encodedJsonSpan.Length - 1? 
-                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogMessages.IDX14121));
-
                 Dot3 = Dot2 + Dot3 + 1;
-
+                if (Dot3 == encodedTokenSpan.Length - 1)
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -463,11 +463,10 @@ internal void ReadToken(ReadOnlyMemory<char> encodedTokenMemory)
                 // empty payload for JWE's {encrypted tokens}.
                 Payload = new JsonClaimSet();
 
-                if (Dot3 == encodedTokenSpan.Length) // TODO: Should this be encodedJsonSpan.Length - 1? 
-                    throw LogHelper.LogExceptionMessage(new ArgumentException(LogMessages.IDX14121));
```
pmaytak, line 467: I think previously this line would never be hit. So changing the exception type to `SecurityTokenMalformedException` shouldn't be a breaking change.

sruke: Agreed.

sruke (approved):

FuPingFranco (approved):

jennyf19 (approved):

PR #2504, opened 2024-02-28
Use Base64.DecodeFromUtf8InPlace

keegan-caruso:
# Use Base64.DecodeFromUtf8InPlace for base64 decode

## Description

Base64.DecodeFromUtf8InPlace is SIMD aware.

CPU: ~ -12% on token read time

before:

| Method             | Mean     | Error     | StdDev    | Median   | P90      | P95      | P100     | Gen0   | Allocated |
|------------------- |---------:|----------:|----------:|---------:|---------:|---------:|---------:|-------:|----------:|
| ReadJWS_FromMemory | 3.487 us | 0.0287 us | 0.0612 us | 3.465 us | 3.601 us | 3.604 us | 3.605 us | 0.1068 |   2.67 KB |
| ReadJWS_FromString | 3.492 us | 0.0101 us | 0.0217 us | 3.490 us | 3.524 us | 3.529 us | 3.545 us | 0.1068 |   2.67 KB |

after:

| Method             | Mean     | Error     | StdDev    | P90      | P95      | P100     | Gen0   | Allocated |
|------------------- |---------:|----------:|----------:|---------:|---------:|---------:|-------:|----------:|
| ReadJWS_FromMemory | 3.068 us | 0.0139 us | 0.0303 us | 3.094 us | 3.117 us | 3.151 us | 0.1068 |   2.67 KB |
| ReadJWS_FromString | 3.076 us | 0.0134 us | 0.0292 us | 3.113 us | 3.121 us | 3.127 us | 0.1068 |   2.67 KB |

Note: this was run with a cpu with the following intrinsics: AVX-512F+CD+BW+DQ+VL+VBMI

### DIFF
--- a/benchmark/Microsoft.IdentityModel.Benchmarks/Microsoft.IdentityModel.Benchmarks.csproj
+++ b/benchmark/Microsoft.IdentityModel.Benchmarks/Microsoft.IdentityModel.Benchmarks.csproj
@@ -23,7 +23,7 @@
   </ItemGroup>-->
   
   <ItemGroup>
-    <PackageReference Include="BenchmarkDotNet" Version="0.13.5" />
+    <PackageReference Include="BenchmarkDotNet" Version="0.13.12" />
   </ItemGroup>
 
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -570,15 +570,16 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         internal JsonClaimSet CreateClaimSet(ReadOnlySpan<char> strSpan, int startIndex, int length, bool createHeaderClaimSet)
         {
             int outputSize = Base64UrlEncoding.ValidateAndGetOutputSize(strSpan, startIndex, length);
+
             byte[] output = ArrayPool<byte>.Shared.Rent(outputSize);
             try
             {
-                Base64UrlEncoding.Decode(strSpan, startIndex, length, output);
+                Base64UrlEncoder.UnsafeDecode(strSpan.Slice(startIndex, length), output);
                 return createHeaderClaimSet ? CreateHeaderClaimSet(output.AsSpan()) : CreatePayloadClaimSet(output.AsSpan());
             }
             finally
             {
-                ArrayPool<byte>.Shared.Return(output);
+                ArrayPool<byte>.Shared.Return(output, true);
             }
         }
 
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -504,7 +504,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                 try
                 {
-                    Header = CreateHeaderClaimSet(Base64UrlEncoder.UnsafeDecode(headerSpan).AsSpan());
+                    Header = CreateHeaderClaimSet(Base64UrlEncoder.Decode(headerSpan).AsSpan());
                 }
                 catch (Exception ex)
                 {
@@ -518,7 +518,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 ReadOnlySpan<char> encryptedKeyBytes = encodedTokenSpan.Slice(Dot1 + 1, Dot2 - Dot1 - 1);
                 if (!encryptedKeyBytes.IsEmpty)
                 {
-                    EncryptedKeyBytes = Base64UrlEncoder.UnsafeDecode(encryptedKeyBytes);
+                    EncryptedKeyBytes = Base64UrlEncoder.Decode(encryptedKeyBytes);
                     _encryptedKey = encryptedKeyBytes.ToString();
                 }
                 else
@@ -532,7 +532,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                 try
                 {
-                    InitializationVectorBytes = Base64UrlEncoder.UnsafeDecode(initializationVectorSpan);
+                    InitializationVectorBytes = Base64UrlEncoder.Decode(initializationVectorSpan);
                 }
                 catch (Exception ex)
                 {
@@ -545,7 +545,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                 try
                 {
-                    AuthenticationTagBytes = Base64UrlEncoder.UnsafeDecode(authTagSpan);
+                    AuthenticationTagBytes = Base64UrlEncoder.Decode(authTagSpan);
                 }
                 catch (Exception ex)
                 {
@@ -558,7 +558,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
                 try
                 {
-                    CipherTextBytes = Base64UrlEncoder.UnsafeDecode(cipherTextSpan);
+                    CipherTextBytes = Base64UrlEncoder.Decode(cipherTextSpan);
                 }
                 catch (Exception ex)
                 {
@@ -574,7 +574,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             byte[] output = ArrayPool<byte>.Shared.Rent(outputSize);
             try
             {
-                Base64UrlEncoder.UnsafeDecode(strSpan.Slice(startIndex, length), output);
+                Base64UrlEncoder.Decode(strSpan.Slice(startIndex, length), output);
                 return createHeaderClaimSet ? CreateHeaderClaimSet(output.AsSpan()) : CreatePayloadClaimSet(output.AsSpan());
             }
             finally
--- a/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
@@ -176,10 +176,7 @@ namespace Microsoft.IdentityModel.Tokens
             return UnsafeDecode(str.AsSpan());
         }
 
-#if NET6_0_OR_GREATER
-        [SkipLocalsInit]
-#endif
-        internal static unsafe byte[] UnsafeDecode(ReadOnlySpan<char> strSpan)
+        internal static byte[] UnsafeDecode(ReadOnlySpan<char> strSpan)
         {
             int mod = strSpan.Length % 4;
             if (mod == 1)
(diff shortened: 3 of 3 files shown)


benchmark/Microsoft.IdentityModel.Benchmarks/Microsoft.IdentityModel.Benchmarks.csproj
```diff
@@ -23,7 +23,7 @@
   </ItemGroup>-->
   
   <ItemGroup>
-    <PackageReference Include="BenchmarkDotNet" Version="0.13.5" />
+    <PackageReference Include="BenchmarkDotNet" Version="0.13.12" />
```
keegan-caruso, line 26: needed to update so that it would report hardware intrinsics

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -253,7 +264,86 @@ internal static unsafe byte[] UnsafeDecode(ReadOnlySpan<char> strSpan)
...
+            if (arrayPoolChars is not null)
+            {
+                charsSpan.Clear();
+                ArrayPool<char>.Shared.Return(arrayPoolChars);
+            }
+        }
+
+        private static unsafe ReadOnlySpan<char> HandlePaddingAndReplace(bool needReplace, Span<char> charsSpan, ReadOnlySpan<char> source)
+        {
+            source.CopyTo(charsSpan);
```
MZOLN, line 278: what happens if charsSpan is smaller than source? Also is there a reason why we wouldnt want to do it in place?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -207,37 +236,19 @@ internal static unsafe byte[] UnsafeDecode(ReadOnlySpan<char> strSpan)
...
-                        charsSpan[source.Length + 1] = base64PadCharacter;
-                    }
-                }
-
-                if (needReplace)
-                {
-                    Span<char> remaining = charsSpan;
-                    int pos;
-                    while ((pos = remaining.IndexOfAny(base64UrlCharacter62, base64UrlCharacter63)) >= 0)
-                    {
```
MZOLN, line 225: This is potentially non-optimal. If I'm understanding it correctly, in an extreme case scenario it could end up being O(n^2) (?) Wouldn't it make sense to simply iterate over it once with a simple for loop and replace in place?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -253,7 +264,86 @@ internal static unsafe byte[] UnsafeDecode(ReadOnlySpan<char> strSpan)
...
+
+            if (needReplace)
+            {
+                Span<char> remaining = charsSpan;
+                int pos;
+                while ((pos = remaining.IndexOfAny(base64UrlCharacter62, base64UrlCharacter63)) >= 0)
+                {
+                    remaining[pos] = (remaining[pos] == base64UrlCharacter62) ? base64Character62 : base64Character63;
+                    remaining = remaining.Slice(pos + 1);
+                }
```
MZOLN, line 296: This does not seem optimal. Wouldnt it end up being a O(N^2) in the edge case? Simple for loop with replacing in place would be more optimal.

E.g.

for (int i = 0; i < remaining.Length; i++)
{
    if (remaining[i] == base64UrlCharacter62)
    {
        remaining[i] = base64Character62;
    }
    else if (remaining[i] == base64UrlCharacter63)
    {
        remaining[i] = base64Character63;
    }
}

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -189,11 +189,40 @@ internal static unsafe byte[] UnsafeDecode(ReadOnlySpan<char> strSpan)
...
+#if NET6_0_OR_GREATER
+            UnsafeNetCoreDecode(strSpan, output, needReplace, decodedLength);
+#else
+            UnsafeNetFWDecode(strSpan, output, needReplace, decodedLength);
+#endif
+        }
+
+#if NET6_0_OR_GREATER
+
+        private static unsafe byte[] UnsafeNetCoreDecode(ReadOnlySpan<char> strSpan, bool needReplace, int decodedLength)
```
sruke: There code duplication between `UnsafeNetCoreDecode(ReadOnlySpan<char> strSpan, bool needReplace, int decodedLength)` and `UnsafeNetCoreDecode(ReadOnlySpan<char> strSpan, Span<byte> output, bool needReplace, int decodedLength)`that can potentially be reduced.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -570,15 +570,17 @@ internal void ReadToken(ReadOnlyMemory<char> encodedTokenMemory)
         internal JsonClaimSet CreateClaimSet(ReadOnlySpan<char> strSpan, int startIndex, int length, bool createHeaderClaimSet)
         {
             int outputSize = Base64UrlEncoding.ValidateAndGetOutputSize(strSpan, startIndex, length);
+
             byte[] output = ArrayPool<byte>.Shared.Rent(outputSize);
             try
             {
-                Base64UrlEncoding.Decode(strSpan, startIndex, length, output);
+                ReadOnlySpan<char> slice = strSpan.Slice(startIndex, length);
+                Base64UrlEncoder.UnsafeDecode(slice, output);
```
sruke: `Base64UrlEncoding.Decode(..)` was throwing ArgumentOutOfRangeException and ArgumentException exceptions. The new code path will not throw these exceptions. We might want to share this in the release notes.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -207,37 +236,19 @@ internal static unsafe byte[] UnsafeDecode(ReadOnlySpan<char> strSpan)
...
             byte[] arrayPoolBytes = null;
             Span<byte> bytesSpan = decodedLength <= StackAllocThreshold ?
                 stackalloc byte[StackAllocThreshold] :
                 arrayPoolBytes = ArrayPool<byte>.Shared.Rent(decodedLength);
 
-            bool converted = Convert.TryFromBase64Chars(source, bytesSpan, out int bytesWritten);
-            Debug.Assert(converted, "Expected TryFromBase64Chars to be successful");
+            int length = Encoding.UTF8.GetBytes(source, bytesSpan);
+            Span<byte> utf8Span = bytesSpan.Slice(0, length);
+            OperationStatus status = System.Buffers.Text.Base64.DecodeFromUtf8InPlace(utf8Span, out int bytesWritten);
```
sruke: If the decode operation was not successful (OperationStatus.InvalidData) should an exception be thrown?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
keegan-caruso: we allocate the buffer right before at the right size, we don't want this to be destructive

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
keegan-caruso: I don't follow, how would it be n^2? We search, we replace, we continue searching from our last position.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
keegan-caruso: I don't really see much. It is a bit limited by needing to keep stackalloc'ed expressions scoped to the same method.

You can make the charsSpan scoped, to allow the stackalloc, but then you can't return it.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
keegan-caruso: yes, good idea.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -570,15 +570,17 @@ internal void ReadToken(ReadOnlyMemory<char> encodedTokenMemory)
         internal JsonClaimSet CreateClaimSet(ReadOnlySpan<char> strSpan, int startIndex, int length, bool createHeaderClaimSet)
         {
             int outputSize = Base64UrlEncoding.ValidateAndGetOutputSize(strSpan, startIndex, length);
+
             byte[] output = ArrayPool<byte>.Shared.Rent(outputSize);
             try
             {
-                Base64UrlEncoding.Decode(strSpan, startIndex, length, output);
+                ReadOnlySpan<char> slice = strSpan.Slice(startIndex, length);
```
brentschmaltz: do we need the local variable slice?
Could we write: Base64UrlEncoder.UnsafeDecode(strSpan.Slice(startIndex, length), output);

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -253,7 +266,88 @@ internal static unsafe byte[] UnsafeDecode(ReadOnlySpan<char> strSpan)
...
+            }
+
+            if (arrayPoolChars is not null)
+            {
+                charsSpan.Clear();
+                ArrayPool<char>.Shared.Return(arrayPoolChars);
+            }
+        }
+
+        private static unsafe ReadOnlySpan<char> HandlePaddingAndReplace(bool needReplace, Span<char> charsSpan, ReadOnlySpan<char> source)
```
brentschmaltz: can we keep the same ordering and name of parameters as other methods: strSpan, output, needReplacement.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -253,7 +266,88 @@ internal static unsafe byte[] UnsafeDecode(ReadOnlySpan<char> strSpan)
...
+            int length = Encoding.UTF8.GetBytes(source, bytesSpan);
+            Span<byte> utf8Span = bytesSpan.Slice(0, length);
+
+            OperationStatus status = System.Buffers.Text.Base64.DecodeFromUtf8InPlace(utf8Span, out int bytesWritten);
+            if (status != OperationStatus.Done)
+                throw LogHelper.LogExceptionMessage(new FormatException(LogHelper.FormatInvariant(LogMessages.IDX10400, strSpan.ToString())));
+
+            utf8Span.Slice(0, bytesWritten).CopyTo(output);
+
+            if (arrayPoolBytes is not null)
```
brentschmaltz: We need to have these in a try / finally block and release the pools in the finally.
see: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/f5f3f7943492b23ff25eaf7829cb4ae629e5856a/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs#L90

sruke (approved):
LGTM!

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -570,15 +570,16 @@ internal void ReadToken(ReadOnlyMemory<char> encodedTokenMemory)
         internal JsonClaimSet CreateClaimSet(ReadOnlySpan<char> strSpan, int startIndex, int length, bool createHeaderClaimSet)
         {
             int outputSize = Base64UrlEncoding.ValidateAndGetOutputSize(strSpan, startIndex, length);
+
             byte[] output = ArrayPool<byte>.Shared.Rent(outputSize);
             try
             {
-                Base64UrlEncoding.Decode(strSpan, startIndex, length, output);
+                Base64UrlEncoder.UnsafeDecode(strSpan.Slice(startIndex, length), output);
```
pmaytak: What's the difference between Base64UrlEncoding and Base64UrlEncoder classes? Very similarly named; difference not apparent.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -207,53 +231,131 @@ internal static unsafe byte[] UnsafeDecode(ReadOnlySpan<char> strSpan)
...
-            if (arrayPoolChars is not null)
+                if (arrayPoolChars is not null)
+                {
+                    charsSpan.Clear();
+                    ArrayPool<char>.Shared.Return(arrayPoolChars);
+                }
+            }
+        }
+
+        private static ReadOnlySpan<char> HandlePaddingAndReplace(Span<char> charsSpan, ReadOnlySpan<char> source, bool needReplace)
```
pmaytak: nit: 
1. Rename `charsSpan` to something like `destination` since we're copying into it.
2. Reorder params so that source comes first, then charsSpan/destination. Easier to read that way source > destination.
3. Since we're updating the contents of `charsSpan`, we don't need to return it? Right now it looks like we're returning a brand new Span.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -189,11 +202,22 @@ internal static unsafe byte[] UnsafeDecode(ReadOnlySpan<char> strSpan)
...
+            Decode(strSpan, output, needReplace, decodedLength);
+#else
+            UnsafeDecode(strSpan, output, needReplace, decodedLength);
+#endif
+        }
+
+#if NET6_0_OR_GREATER
+
+        [SkipLocalsInit]
+        private static byte[] Decode(ReadOnlySpan<char> strSpan, bool needReplace, int decodedLength)
```
pmaytak: Can these two Decode methods be combined? The main difference is when we create a result array or copy into destination on lines 253 and 311. Besides that a bit of duplicate code.
`byte[] Decode(ReadOnlySpan<char> strSpan, bool needReplace, int decodedLength)`
`void Decode(ReadOnlySpan<char> strSpan, Span<byte> output, bool needReplace, int decodedLength)`

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
pmaytak: Ignore point 3, looks like it matches the previous behavior.

pmaytak (approved):
Approved with suggestions. The code looks like it matches the previous one (except using the DecodeFromUtf8InPlace). Main suggestions are about refactoring a bit of the duplicate code. This class was a bit difficult to review since there are similarly named methods, duplicate code, and .NET6+ flags.

kllysng (approved):
LGTM

FuPingFranco (approved):
LGTM

pmaytak (approved):

PR #2508, opened 2024-02-28
Remove reference to Microsoft.IdentityModel.Logging in Microsoft.IdentityModel.Protocols

ghost:
`Microsoft.IdentityModel.Protocols` already has a dependency on `Microsoft.IdentityModel.Tokens` that already depends on `Microsoft.IdentityModel.Logging`.


jennyf19:
thank you @joegoldman2

PR #2520, opened 2024-03-08
MSBuild target to override Project Ref versions

keegan-caruso:
# MSBuild target to override Project Ref versions

## Description

Adds a msbuild target to override project references to use an exact version.


jennyf19 (approved):
I would like to see if this helps resolve the issue.

keegan-caruso:
Might be better to manually write a nuspec...  This is a bit of a hack.

PR #2564, opened 2024-04-18
Make AadIssuerValidator method ValidateAsync public.

victorm-hernandez:
# Make AadIssuerValidator method ValidateAsync public.

## Description

Making the ValidateAsync method of the AadIssuerValidator class public so it can be consumed externally. 

Related to Issue #{2556}

### DIFF
--- a/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
+++ b/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
@@ -212,7 +212,7 @@ namespace Microsoft.IdentityModel.Validators
         /// <exception cref="ArgumentNullException"> if <paramref name="securityToken"/> is null.</exception>
         /// <exception cref="ArgumentNullException"> if <paramref name="validationParameters"/> is null.</exception>
         /// <exception cref="SecurityTokenInvalidIssuerException">if the issuer is invalid or if there is a network issue. </exception>
-        internal async ValueTask<string> ValidateAsync(
+        public async ValueTask<string> ValidateAsync(
             string issuer,
             SecurityToken securityToken,
             TokenValidationParameters validationParameters)


MZOLN (approved):
LGTM

FuPingFranco (approved):
LGMT Thanks @victorm-hernandez !

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -212,7 +212,7 @@ private void SetupAuthorities(string aadAuthority, ProtocolVersion version)
         /// <exception cref="ArgumentNullException"> if <paramref name="securityToken"/> is null.</exception>
         /// <exception cref="ArgumentNullException"> if <paramref name="validationParameters"/> is null.</exception>
         /// <exception cref="SecurityTokenInvalidIssuerException">if the issuer is invalid or if there is a network issue. </exception>
-        internal async ValueTask<string> ValidateAsync(
+        public async ValueTask<string> ValidateAsync(
```
jennyf19, line 215: there are github action warnings, can you please take a look at those.

pmaytak, line 215: Should we still return a ValueTask here? ValidateAsync is mostly sync, except making the async call to get the configuration. 

My understanding ValueTask has limitations like they can only be awaited once unlike Task. Perhaps more general Task<string> return type is more appropriate for a public API?

(Later if we do decide to deprecate this method later, and have one that returns a Task, it be a bit difficult since return type is not part of the method signature).

https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask?view=net-8.0#remarks

victorm-hernandez: That is a good point, taking this into consideration I think we should not haste this part. I will make the internals visible to the IdentityWeb instead. Thanks Peter

victorm-hernandez: About to send an iteration it seems it doesnt recognize the null checks we do at the beginning

victorm-hernandez:
Abandon this PR in favor of making the change internals of Wilson available to IdWeb instead.

Issue #2556, opened 2024-04-10
Thread starvation due to sync over async code in IdentityModel.Tokens.Validators


mirjana1211:
### Microsoft.Identity.Web Library

Microsoft.Identity.Web

### Microsoft.Identity.Web version

2.17.2

### Web app

Not Applicable

### Web API

Protected web APIs (validating tokens)

### Token cache serialization

Not Applicable

### Description

We have been experiencing sporadic issues with our application deployed as an Azure App Service running on Linux (.net8). 
We are using Microsoft.Identity.Web library and Microsoft Entra ID as our identity provider.

From time to time, we get an alert for failed health checks for one of the instances (minimum of 6 instances running). The instance seems to be unresponsive and for a period of time and the issue always resolves itself within a few minutes. There is no relevant info in the logs or in the Diagnose and solve problems section for the app service. 
We can not identify any changes in request rate or resource utilization before the outage. 

We have created a support case with Microsoft and have provided multiple memory dumps for analysis. The information we got from the support engineer is that the app service failure is caused by thread starvation due to sync over async code in IdentityModel.

### Reproduction steps

The outage occurs sporadically and it is not possible to reproduce it on demand.

The problem seems to be related to load. We have several instances of the same application deployed in different Azure regions and the only instance where we have observed the problem is running under a relatively high load (~500-600 req/sec during peak time).

### Relevant code snippets

```csharp
The code in question, according to the dumps analysis, is in the ValidateIssuer method: 

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/Validators.cs.
```

### Expected behavior

No application outages due to blocking calls :)

pmaytak:
Based on the discussions, looks like there are 3 options:

1.  Make the currently internal `ValidateIssuerAsync` API in IdentityModel public and use it in Identity Web instead of the sync overload.
Pro: Quick fix. 
Con: Async delegates API design is not fully complete. This change may make changes to this API more difficult later. Adding only one async delegate for issuer right now makes API less consistent.
2. Make IdentityModel internals visible to Identity Web and use the internal `ValidateIssuerAsync` in Identity Web instead of the sync overload. In IdentityModel 8, the async delegates will be finalized and made public; IdentityModel internals can be hidden for Identity Web again.
Pro: Quick fix and doesn't expose a new public API.
[...]

PR #2565, opened 2024-04-19
Made ValidateAsync and IssuerValidatorAsync public

victorm-hernandez:
# Made ValidateAsync and IssuerValidatorAsync public

- [ x] You've read the [Contributor Guide](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/Contributing.md) and [Code of Conduct](https://opensource.microsoft.com/codeofconduct/).
- [ x] You've included unit or integration tests for your change, where applicable.
- [x ] You've included inline docs for your change, where applicable.
- [ x] If any gains or losses in performance are possible, you've included benchmarks for your changes. [More info](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/Benchmarking-in-Wilson)
- [x ] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

## Description

Exposed the ValidateAsync methods in the AadIssuerValidator class to allow for custom async validation of the issuer.
Exposed the IssuerValidatorAsync delegate in the TokenValidationParameters so the custom async validation can be set.
Updated the priority logic used to decide which custom validator has priority.

Related issue: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2556

Fixes #{2556} (in this specific format)

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validators.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validators.cs
@@ -213,16 +213,29 @@ namespace Microsoft.IdentityModel.Tokens
         /// <returns>The issuer to use when creating the "Claim"(s) in a "ClaimsIdentity".</returns>
         /// <exception cref="ArgumentNullException">If 'validationParameters' is null.</exception>
         /// <exception cref="ArgumentNullException">If 'issuer' is null or whitespace and <see cref="TokenValidationParameters.ValidateIssuer"/> is true.</exception>
-        /// <exception cref="ArgumentNullException">If ' configuration' is null.</exception>
+        /// <exception cref="ArgumentNullException">If 'configuration' is null.</exception>
         /// <exception cref="SecurityTokenInvalidIssuerException">If <see cref="TokenValidationParameters.ValidIssuer"/> is null or whitespace and <see cref="TokenValidationParameters.ValidIssuers"/> is null and <see cref="BaseConfiguration.Issuer"/> is null.</exception>
         /// <exception cref="SecurityTokenInvalidIssuerException">If 'issuer' failed to matched either <see cref="TokenValidationParameters.ValidIssuer"/> or one of <see cref="TokenValidationParameters.ValidIssuers"/> or <see cref="BaseConfiguration.Issuer"/>.</exception>
         /// <remarks>An EXACT match is required.</remarks>
         internal static string ValidateIssuer(string issuer, SecurityToken securityToken, TokenValidationParameters validationParameters, BaseConfiguration configuration)
         {
-            ValueTask<string> vt = ValidateIssuerAsync(issuer, securityToken, validationParameters, configuration);
-            return vt.IsCompletedSuccessfully ?
-                vt.Result :
-                vt.AsTask().ConfigureAwait(false).GetAwaiter().GetResult();
+            if (validationParameters == null)
+                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
+
+            // Given that this is a synchronous method and to avoid consuming excesive threads
+            // we will favor the synchronous delegates over the asynchronous one.
+            if (validationParameters.IssuerValidator == null
+                && validationParameters.IssuerValidatorUsingConfiguration == null
+                && validationParameters.IssuerValidatorAsync != null)
+            {
+                if (LogHelper.IsEnabled(EventLogLevel.Warning))
+                    LogHelper.LogWarning(LogMessages.IDX10266);
+
+                return validationParameters.IssuerValidatorAsync(issuer, securityToken, validationParameters)
+                    .ConfigureAwait(false).GetAwaiter().GetResult();
+            }
+
+            return ValidateIssuerInternal(issuer, securityToken,  validationParameters, configuration);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -213,16 +213,29 @@ public static string ValidateIssuer(string issuer, SecurityToken securityToken,
...
-            ValueTask<string> vt = ValidateIssuerAsync(issuer, securityToken, validationParameters, configuration);
-            return vt.IsCompletedSuccessfully ?
-                vt.Result :
-                vt.AsTask().ConfigureAwait(false).GetAwaiter().GetResult();
+            if (validationParameters == null)
+                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
+
+            // Given that this is a synchronous method and to avoid consuming excesive threads
+            // we will favor the synchronous delegates over the asynchronous one.
+            if (validationParameters.IssuerValidator == null
```
keegan-caruso, line 227: this is a behavioral breaking change?

keegan-caruso: Or am I reading it wrong?

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -213,16 +213,29 @@ public static string ValidateIssuer(string issuer, SecurityToken securityToken,
...
+                && validationParameters.IssuerValidatorUsingConfiguration == null
+                && validationParameters.IssuerValidatorAsync != null)
```
MZOLN, line 229: This is internal, does this limit the use case in any way?

src/Microsoft.IdentityModel.Tokens/Validators.cs
pmaytak: > this is a behavioral breaking change?

👍 Currently IssuerValidatorAsync  is always preferred: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b352bcce8cf2c1676f6ef13558aff29b7766602f/src/Microsoft.IdentityModel.Tokens/Validators.cs#L248-L264

test/Microsoft.IdentityModel.Validators.Tests/MicrosoftIdentityIssuerValidatorTest.cs
```diff
@@ -921,12 +1080,21 @@ public void Validate_UsesLKGWithConfigurationProvider()
...
             // validating cross versions also validates with LKG
             actualIssuer = aadIssuerValidator.Validate(v1TokenIssuer, v1JwtSecurityToken, new TokenValidationParameters { ValidateWithLKG = true });
+            IdentityComparer.AreEqual(v1TokenIssuer, actualIssuer, context);
 
+            actualIssuer = await aadIssuerValidator.ValidateAsync(v1TokenIssuer, v1JwtSecurityToken, new TokenValidationParameters { ValidateWithLKG = true });
             IdentityComparer.AreEqual(v1TokenIssuer, actualIssuer, context);
+
             TestUtilities.AssertFailIfErrors(context);
 
             // if LKG not valid validation fails
```
MZOLN, line 1100: There seem to be a few missing places in these files that didnt switch to async ? Why is that?

src/Microsoft.IdentityModel.Tokens/Validators.cs
victorm-hernandez: Correct this is a change on the behavior. Async custom validators would be preferred when running asynchronously and Sync custom validators otherwise.

src/Microsoft.IdentityModel.Tokens/Validators.cs
victorm-hernandez: With this change this is no longer internal but public.

test/Microsoft.IdentityModel.Validators.Tests/MicrosoftIdentityIssuerValidatorTest.cs
victorm-hernandez: ^_^ good catch, I missed updating this block.

src/Microsoft.IdentityModel.Tokens/Validators.cs
MZOLN: The property on TokenValidationParameters is still internal, the callers cant assign it to the pipeline.
Edit -> you are right, the change in this PR I looked at references...

src/Microsoft.IdentityModel.Tokens/Validators.cs
brentschmaltz: We need a coordinated change across all the delegates that are async and do not rely on throwing, but instead return (too be designed) ValidationResult.

We have to address multiple policies from SDK's that build on top. This is where the first policy fails, throws and logs a bunch of errors, and the second one succeeds.

We have thrown AND logged errors, but the authentication succeeded.
Users should have a way to capture all results, but logging and throwing by default is causing performance and debugging issues as logs have a lot of false negatives.

brentschmaltz (changes requested):
We need to consider all the delegates.

brentschmaltz:
This conflicts with work on reducing exceptions.

PR #2569, opened 2024-04-24
Verify authentication tag length 

### LINKED ISSUE #1641: Extra characters at the end of encrypted JWT tokens don't prevent them from being considered valid
Hey,

Yesterday, the following report was posted on the OpenIddict repository: https://github.com/openiddict/openiddict-core/issues/1254.

I was able to reproduce it on my machine and confirmed this was caused by IdentityModel.
It looks like IdentityModel doesn't reject a JWE payload if it contains extra characters after the authentication tag.

Here's a repro:

```csharp
using System.Diagnostics;
using System.Security.Claims;

namespace ConsoleApp13
{
    class Program
    {
        static void Main(string[] args)
        {
            var identity = new ClaimsIdentity(TokenValidationParameters.DefaultAuthenticationType);
            var signingKey = new RsaSecurityKey(RSA.Create(2048))
            {
                KeyId = "id1"
            };

            var encryptionKey = new RsaSecurityKey(RSA.Create(2048))
            {
                KeyId = "id2"
            };

            var token = handler.CreateToken(new SecurityTokenDescriptor
            {
                Audience = "audience",
                Issuer = "issuer",
                EncryptingCredentials = new EncryptingCredentials(encryptionKey, SecurityAlgorithms.RsaOAEP, SecurityAlgorithms.Aes256CbcHmacSha512),
                SigningCredentials = new SigningCredentials(signingKey, SecurityAlgorithms.RsaSha256Signature),
                Subject = identity
            });

            var parameters = new TokenValidationParameters
            {
                ValidAudience = "audience",
                ValidIssuer = "issuer",
                TokenDecryptionKey = encryptionKey,
                IssuerSigningKey = signingKey
            };

            var result = handler.ValidateToken(token + "A", parameters);
            Debug.Assert(!result.IsValid, "The token shouldn't be considered valid.");
        }
    }
}
```

@brentschmaltz @jennyf19 could you please take a look?

### LINKED ISSUE #2201: [Bug] JWE is still valid after appending data to the Authenticaton Tag
System.IdentityModel.Tokens.Jwt 7.1.0-preview

**Where is the issue?**
 * [x] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [x] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [x] S.IM.Tokens.Jwt
 * Other (please describe)

This is a new app or an experiment.

**Repro**

```csharp

var jwtHandler = new JsonWebTokenHandler();

var claims = new Dictionary<string, object>
{
    { "claim1", "value1" },
    { "claim2", "value2" },
};

var signingKey = RSA.Create();
var validationKey = RSA.Create(signingKey.ExportParameters(false));

var decryptionKey = RSA.Create();
var encryptionKey = RSA.Create(decryptionKey.ExportParameters(false));

var token = jwtHandler.CreateToken(new SecurityTokenDescriptor
{
    Claims = claims,
    SigningCredentials = new SigningCredentials(
        new RsaSecurityKey(signingKey),
        SecurityAlgorithms.RsaSha256),
    EncryptingCredentials = new EncryptingCredentials(
        new RsaSecurityKey(encryptionKey),
        SecurityAlgorithms.RsaOAEP,
        SecurityAlgorithms.Aes256CbcHmacSha512)
});

// altering the JWE by appending some stuff to the Authentication Tag
var invalidToken = token + "the_token_has_been_tampered_with";

var validationResult = jwtHandler.ValidateToken(invalidToken, new()
{
    IssuerSigningKey = new RsaSecurityKey(validationKey),
    TokenDecryptionKey = new RsaSecurityKey(decryptionKey),
    ValidateAudience = false,
    ValidateIssuer = false,
    ValidateLifetime = false
});

if (validationResult.IsValid) // token is valid when it shouldn't be
    throw new Exception("The invalid token has been incorrectly marked as valid!");
```

**Expected behavior**
The validation of the JWE should fail, as the Authentication Tag has been altered, thus no longer valid.

Extracted from [RFC 7516, Section 5.2](https://datatracker.ietf.org/doc/html/rfc7516#section-5.2), regarding JWE decryption:
```
2.   Base64url decode the encoded representations of the JWE
       Protected Header, the JWE Encrypted Key, the JWE Initialization
       Vector, the JWE Ciphertext, the JWE Authentication Tag, and the
       JWE AAD, following the restriction that no line breaks,
       whitespace, or other additional characters have been used.
```

**Actual behavior**
The JWE is successfully validated when the Authentication Tag has been altered.

**Possible solution**
Check the actual length of the Authentication Tag and fail the validation if it does not match the expected value, according to the JWE encryption algorithm used.

kllysng:
[Design linked in Bug](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2201#issuecomment-2072883307)

Fixes #2201 , #1641

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
@@ -33,12 +33,12 @@ namespace Microsoft.IdentityModel.Tokens
         private DecryptionDelegate DecryptFunction;
         private EncryptionDelegate EncryptFunction;
         private const string _className = "Microsoft.IdentityModel.Tokens.AuthenticatedEncryptionProvider";
-        internal const string _skipValidationOfAuthenticationTagLength = "Switch.Microsoft.IdentityModel.SkipValidationOfAuthenticationTagLength"; 
+        internal const string _skipValidationOfAuthenticationTagLength = "Switch.Microsoft.IdentityModel.SkipAuthenticationTagLengthValidation"; 
 
         /// <summary>
         /// Mapping from algorithm to the expected authentication tag length.
         /// </summary>
-        internal static readonly Dictionary<string, int> ExpectedAuthenticationTagBase64UrlLength = new Dictionary<string, int>
+        internal static readonly Dictionary<string, int> ExpectedAuthenticationTagBase64UrlLength = new()
         {
             { SecurityAlgorithms.Aes128Gcm, 24 },
             { SecurityAlgorithms.Aes192Gcm, 24 },
@@ -180,7 +180,7 @@ namespace Microsoft.IdentityModel.Tokens
         private byte[] DecryptWithAesCbc(byte[] ciphertext, byte[] authenticatedData, byte[] iv, byte[] authenticationTag)
         {
             // Verify authentication Tag
-            if (!ExpectedAuthenticationTagBase64UrlLength.TryGetValue(Algorithm, out int expectedTagLength) &&
+            if (ExpectedAuthenticationTagBase64UrlLength.TryGetValue(Algorithm, out int expectedTagLength) &&
                 ShouldValidateAuthenticationTagLength()
                 && expectedTagLength != authenticationTag.Length)
                 throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(LogMessages.IDX10625, Base64UrlEncoder.Encode(authenticationTag))));
--- a/src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
@@ -2,9 +2,10 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
-using System.Security.Cryptography;
 using System.Runtime.InteropServices;
+using System.Security.Cryptography;
 using Microsoft.IdentityModel.Logging;
 
 namespace Microsoft.IdentityModel.Tokens
@@ -32,6 +33,20 @@ namespace Microsoft.IdentityModel.Tokens
         private DecryptionDelegate DecryptFunction;
         private EncryptionDelegate EncryptFunction;
         private const string _className = "Microsoft.IdentityModel.Tokens.AuthenticatedEncryptionProvider";
+        internal const string _skipValidationOfAuthenticationTagLength = "Switch.Microsoft.IdentityModel.SkipValidationOfAuthenticationTagLength"; 
+
+        /// <summary>
+        /// Mapping from algorithm to the expected authentication tag length.
+        /// </summary>
+        internal static readonly Dictionary<string, int> ExpectedAuthenticationTagBase64UrlLength = new Dictionary<string, int>
+        {
+            { SecurityAlgorithms.Aes128Gcm, 24 },
+            { SecurityAlgorithms.Aes192Gcm, 24 },
+            { SecurityAlgorithms.Aes256Gcm, 24 },
+            { SecurityAlgorithms.Aes128CbcHmacSha256, 24 },
+            { SecurityAlgorithms.Aes192CbcHmacSha384, 32 },
+            { SecurityAlgorithms.Aes256CbcHmacSha512, 44 }
+        };
 
         /// <summary>
         /// Initializes a new instance of the <see cref="AuthenticatedEncryptionProvider"/> class used for encryption and decryption.
@@ -165,13 +180,18 @@ namespace Microsoft.IdentityModel.Tokens
         private byte[] DecryptWithAesCbc(byte[] ciphertext, byte[] authenticatedData, byte[] iv, byte[] authenticationTag)
         {
             // Verify authentication Tag
+            if (!ExpectedAuthenticationTagBase64UrlLength.TryGetValue(Algorithm, out int expectedTagLength) &&
+                ShouldValidateAuthenticationTagLength()
+                && expectedTagLength != authenticationTag.Length)
+                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(LogMessages.IDX10625, Base64UrlEncoder.Encode(authenticationTag))));
+
             byte[] al = Utility.ConvertToBigEndian(authenticatedData.Length * 8);
             byte[] macBytes = new byte[authenticatedData.Length + iv.Length + ciphertext.Length + al.Length];
             Array.Copy(authenticatedData, 0, macBytes, 0, authenticatedData.Length);
             Array.Copy(iv, 0, macBytes, authenticatedData.Length, iv.Length);
             Array.Copy(ciphertext, 0, macBytes, authenticatedData.Length + iv.Length, ciphertext.Length);
             Array.Copy(al, 0, macBytes, authenticatedData.Length + iv.Length + ciphertext.Length, al.Length);
-            if (!_symmetricSignatureProvider.Value.Verify(macBytes, 0, macBytes.Length, authenticationTag, 0, _authenticatedkeys.Value.HmacKey.Key.Length, Algorithm))
+            if (!_symmetricSignatureProvider.Value.Verify(macBytes, 0, macBytes.Length, authenticationTag, 0, expectedTagLength, Algorithm))
                 throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(LogMessages.IDX10650, Base64UrlEncoder.Encode(authenticatedData), Base64UrlEncoder.Encode(iv), Base64UrlEncoder.Encode(authenticationTag))));
 
             using Aes aes = Aes.Create();
@@ -187,6 +207,11 @@ namespace Microsoft.IdentityModel.Tokens
             {
                 throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(LogMessages.IDX10654, ex)));
             }
+        }
+
+        private static bool ShouldValidateAuthenticationTagLength()
+        {
+            return !(AppContext.TryGetSwitch(_skipValidationOfAuthenticationTagLength, out bool skipValidation) && skipValidation);
         }
 
         private AuthenticatedKeys CreateAuthenticatedKeys()
--- a/src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
@@ -36,19 +36,6 @@ namespace Microsoft.IdentityModel.Tokens
         internal const string _skipValidationOfAuthenticationTagLength = "Switch.Microsoft.IdentityModel.SkipAuthenticationTagLengthValidation"; 
 
         /// <summary>
-        /// Mapping from algorithm to the expected authentication tag length.
-        /// </summary>
-        internal static readonly Dictionary<string, int> ExpectedAuthenticationTagBase64UrlSizeInBytes = new()
-        {
-            { SecurityAlgorithms.Aes128Gcm, 16 },
-            { SecurityAlgorithms.Aes192Gcm, 16 },
-            { SecurityAlgorithms.Aes256Gcm, 16 },
-            { SecurityAlgorithms.Aes128CbcHmacSha256, 16 },
-            { SecurityAlgorithms.Aes192CbcHmacSha384, 24 },
-            { SecurityAlgorithms.Aes256CbcHmacSha512, 32 }
-        };
-
-        /// <summary>
         /// Initializes a new instance of the <see cref="AuthenticatedEncryptionProvider"/> class used for encryption and decryption.
         /// </summary>
         /// <param name="key">The <see cref="SecurityKey"/> that will be used for crypto operations.</param>
@@ -180,9 +167,8 @@ namespace Microsoft.IdentityModel.Tokens
         private byte[] DecryptWithAesCbc(byte[] ciphertext, byte[] authenticatedData, byte[] iv, byte[] authenticationTag)
         {
             // Verify authentication Tag
-            int expectedTagLength = _authenticatedkeys.Value.HmacKey.Key.Length;
             if (ShouldValidateAuthenticationTagLength()
-                && ExpectedAuthenticationTagBase64UrlSizeInBytes.TryGetValue(Algorithm, out expectedTagLength)
+                && SymmetricSignatureProvider.ExpectedSignatureSizeInBytes.TryGetValue(Algorithm, out int expectedTagLength)
                 && expectedTagLength != authenticationTag.Length)
                 throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(LogMessages.IDX10625, Base64UrlEncoder.Encode(authenticationTag))));
 
@@ -192,7 +178,7 @@ namespace Microsoft.IdentityModel.Tokens
             Array.Copy(iv, 0, macBytes, authenticatedData.Length, iv.Length);
             Array.Copy(ciphertext, 0, macBytes, authenticatedData.Length + iv.Length, ciphertext.Length);
             Array.Copy(al, 0, macBytes, authenticatedData.Length + iv.Length + ciphertext.Length, al.Length);
-            if (!_symmetricSignatureProvider.Value.Verify(macBytes, 0, macBytes.Length, authenticationTag, 0, expectedTagLength, Algorithm))
+            if (!_symmetricSignatureProvider.Value.Verify(macBytes, 0, macBytes.Length, authenticationTag, 0, _authenticatedkeys.Value.HmacKey.Key.Length, Algorithm))
                 throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(LogMessages.IDX10650, Base64UrlEncoder.Encode(authenticatedData), Base64UrlEncoder.Encode(iv), Base64UrlEncoder.Encode(authenticationTag))));
 
             using Aes aes = Aes.Create();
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
```diff
@@ -165,13 +180,19 @@ private AuthenticatedEncryptionResult EncryptWithAesCbc(byte[] plaintext, byte[]
...
+                && expectedTagLength != authenticationTag.Length)
+                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(LogMessages.IDX10625, Base64UrlEncoder.Encode(authenticationTag))));
+
             byte[] al = Utility.ConvertToBigEndian(authenticatedData.Length * 8);
             byte[] macBytes = new byte[authenticatedData.Length + iv.Length + ciphertext.Length + al.Length];
             Array.Copy(authenticatedData, 0, macBytes, 0, authenticatedData.Length);
             Array.Copy(iv, 0, macBytes, authenticatedData.Length, iv.Length);
             Array.Copy(ciphertext, 0, macBytes, authenticatedData.Length + iv.Length, ciphertext.Length);
             Array.Copy(al, 0, macBytes, authenticatedData.Length + iv.Length + ciphertext.Length, al.Length);
-            if (!_symmetricSignatureProvider.Value.Verify(macBytes, 0, macBytes.Length, authenticationTag, 0, _authenticatedkeys.Value.HmacKey.Key.Length, Algorithm))
```
kllysng: @brentschmaltz we discussed possibly replacing authenticatedkeys.Value.HmacKey.Key.Length like this. It would remove the necessity for the check on lines 183-188, but then it wouldn't be behind the app context switch. I propose that we keep it as is (_authenticatedkeys.Value.HmacKey.Key.Length) and update it upon the next major release since we'll remove the context switch at that point.

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
```diff
@@ -32,6 +33,20 @@ private struct AuthenticatedKeys
...
         private EncryptionDelegate EncryptFunction;
         private const string _className = "Microsoft.IdentityModel.Tokens.AuthenticatedEncryptionProvider";
+        internal const string _skipValidationOfAuthenticationTagLength = "Switch.Microsoft.IdentityModel.SkipAuthenticationTagLengthValidation"; 
+
+        /// <summary>
+        /// Mapping from algorithm to the expected authentication tag length.
+        /// </summary>
+        internal static readonly Dictionary<string, int> ExpectedAuthenticationTagBase64UrlSizeInBytes = new()
+        {
+            { SecurityAlgorithms.Aes128Gcm, 16 },
```
kllysng: @brentschmaltz while testing, I discovered that the auth length issue does not repro with the AesGcm algos - do we anticipate needing these values in the future, or shall I remove them?

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
```diff
@@ -32,6 +33,20 @@ private struct AuthenticatedKeys
         private DecryptionDelegate DecryptFunction;
         private EncryptionDelegate EncryptFunction;
         private const string _className = "Microsoft.IdentityModel.Tokens.AuthenticatedEncryptionProvider";
+        internal const string _skipValidationOfAuthenticationTagLength = "Switch.Microsoft.IdentityModel.SkipAuthenticationTagLengthValidation"; 
+
+        /// <summary>
+        /// Mapping from algorithm to the expected authentication tag length.
+        /// </summary>
+        internal static readonly Dictionary<string, int> ExpectedAuthenticationTagBase64UrlSizeInBytes = new()
```
keegan-caruso: The map in SymmetricSignatureProvider has a lot of overlap

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -4189,108 +4188,228 @@ public static TheoryData<CreateTokenTheoryData> IncludeSecurityTokenOnFailureTes
...
+            Assert.Equal(theoryData.IsValid, tokenValidationResult.IsValid);
         }
 
-        public CreateTokenTheoryData(string testId) : base(testId)
+        public static TheoryData<CreateTokenTheoryData> ValidateAuthenticationTagLengthTheoryData()
         {
+            var signingCredentials512 = new SigningCredentials(KeyingMaterial.RsaSecurityKey_2048, SecurityAlgorithms.RsaPKCS1, SecurityAlgorithms.Sha512);
+            return new TheoryData<CreateTokenTheoryData>()
+            {
+                new("A128CBC-HS256_IsNotValidByDefault")
```
keegan-caruso, line 4246: GCM cases?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -4189,108 +4188,228 @@ public static TheoryData<CreateTokenTheoryData> IncludeSecurityTokenOnFailureTes
...
         {
+            // arrange
+            AppContext.SetSwitch(AuthenticatedEncryptionProvider._skipValidationOfAuthenticationTagLength, theoryData.EnableAppContextSwitch);
+            var payload = new JObject()
+            {
+                { JwtRegisteredClaimNames.Email, "Bob@contoso.com" },
+                { JwtRegisteredClaimNames.GivenName, "Bob" },
+                { JwtRegisteredClaimNames.Iss, "http://Default.Issuer.com"},
+                { JwtRegisteredClaimNames.Aud, "http://Default.Audience.com" },
+                { JwtRegisteredClaimNames.Iat, EpochTime.GetIntDate(DateTime.Parse("2017-03-17T18:33:37.095Z")).ToString() },
```
keegan-caruso: it is confusing to have hardcoded dates, something like DateTimeNow makes more sense.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
kllysng: I was not able to repro the bug when testing with AesGcm cases (mentioned [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2569#discussion_r1581717507))

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
kllysng: When I was chatting with Brent, he expressed that we shouldn't conflate auth tag and signature. Using "ExpectedSgnatureSizeInBytes" might be confusing. I'm happy to use it an replace this one altogether, since we may remove the AesGcm values as they don't replicate this bug, if we're okay with it being slightly confusing.

kllysng: I went ahead and used "SymmetricSignatureSizeInBytes" since a customer also mentioned it in the issue.

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
kllysng: I went ahead and removed the mappings as there are similar ones in SymmetricSignatureProvider. See this [comment](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2569/files#r1583822363)

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
kllysng: I went ahead and kept it "_authenticatedkeys.Value.HmacKey.Key.Length" for the same reason why we want an app context switch. We can update this on the next major rel of Wilson

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
kllysng: Happy to make this change - updated to DateTime.Now

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
kllysng: Closing this now - please reopen if needed!

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
```diff
@@ -32,6 +33,7 @@ private struct AuthenticatedKeys
         private DecryptionDelegate DecryptFunction;
         private EncryptionDelegate EncryptFunction;
         private const string _className = "Microsoft.IdentityModel.Tokens.AuthenticatedEncryptionProvider";
+        internal const string _skipValidationOfAuthenticationTagLength = "Switch.Microsoft.IdentityModel.SkipAuthenticationTagLengthValidation";
```
JoshLozensky, line 36: Super nit: can we add "Name" or "Switch" to the end of this variable? If I look at just the variable name "skipValidationOfAuthenticationTagLength" I would expect it to be a bool that tells me whether to skip this or not rather than the name of a switch.

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
```diff
@@ -165,6 +167,11 @@ private AuthenticatedEncryptionResult EncryptWithAesCbc(byte[] plaintext, byte[]
         private byte[] DecryptWithAesCbc(byte[] ciphertext, byte[] authenticatedData, byte[] iv, byte[] authenticationTag)
         {
             // Verify authentication Tag
+            if (ShouldValidateAuthenticationTagLength()
+                && SymmetricSignatureProvider.ExpectedSignatureSizeInBytes.TryGetValue(Algorithm, out int expectedTagLength)
```
JoshLozensky, line 171: Should the other two checks be added to the ShouldValidateAuthenticationTagLength method since they are necessary for making that decision?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -4189,108 +4188,247 @@ public static TheoryData<CreateTokenTheoryData> IncludeSecurityTokenOnFailureTes
...
 
-    public class CreateTokenTheoryData : TheoryDataBase
-    {
-        public CreateTokenTheoryData()
+        [Theory, MemberData(nameof(ValidateAuthenticationTagLengthTheoryData))]
+        public void ValidateTokenAsync_ModifiedAuthNTag(CreateTokenTheoryData theoryData)
         {
+            // arrange
+            AppContext.SetSwitch(AuthenticatedEncryptionProvider._skipValidationOfAuthenticationTagLength, theoryData.EnableAppContextSwitch);
+            var payload = new JObject()
```
JoshLozensky, line 4197: Consider putting the payload, jsonWebTokenHandler, and signingCredentials into the theory data.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -4189,108 +4188,247 @@ public static TheoryData<CreateTokenTheoryData> IncludeSecurityTokenOnFailureTes
...
+            {
+                { JwtRegisteredClaimNames.Email, "Bob@contoso.com" },
+                { JwtRegisteredClaimNames.GivenName, "Bob" },
+                { JwtRegisteredClaimNames.Iss, "http://Default.Issuer.com"},
+                { JwtRegisteredClaimNames.Aud, "http://Default.Audience.com" },
+                { JwtRegisteredClaimNames.Iat, EpochTime.GetIntDate(DateTime.Now).ToString() },
+                { JwtRegisteredClaimNames.Nbf, EpochTime.GetIntDate(DateTime.Now).ToString() },
+                { JwtRegisteredClaimNames.Exp, EpochTime.GetIntDate(DateTime.Now.AddDays(1)).ToString() },
```
JoshLozensky, line 4205: Even if the number is one could you put it into a variable so we can see what it represents, otherwise it's just a magic #... i.e. I can see we are setting the expiry in a day, but why 1 day and if we ever want to change it having a variable to change is easier than reading through the code to find it.

JoshLozensky (approved):

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
kllysng: For consistency with our other context switch, I might keep is as is. https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2421/files

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
kllysng: Thanks for the idea

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -134,6 +134,7 @@ internal static class LogMessages
         // public const string IDX10622 = "IDX10622:";
         // public const string IDX10623 = "IDX10623:";
         // public const string IDX10624 = "IDX10624:";
+        public const string IDX10625 = "IDX10625: Failed to verify the length of the authenticationTag '{0}'.";
```
GeoK: Consider adding an aka.ms link that would point to an article on the [wiki](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki) explaining why one would start seeing this exception, how to set the config switch and what are the risks.
Log message seems misleading too, IdentityModel can verify, but verification fails as actual tag length doesn't correspond to the expected one.

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
kllysng: I appreciate your comment! The purpose of `ShouldValidateAuthenticationTagLength()` is to check if we need to validate the authentication tag length, not validate it, and would prefer to keep single responsibility for this method 😄

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
GeoK: Adding the algorithm and actual/expected tag length to the message can also be helpful.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
kllysng: Appreciate the idea.. All the theory data cases share the same payload, handler, and signing creds and there would be more code duplication to pull them out into the theory datas. 😵 Please let me know if there's another reason, otherwise I will keep as is 🤔

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
GeoK: @kellyyangsong I can't see the comment anymore. Does this PR apply only to AesCbc?

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
```diff
@@ -201,7 +213,6 @@ internal SymmetricSignatureProvider CreateSymmetricSignatureProvider()
             if (!IsSupportedAlgorithm(Key, Algorithm))
                 throw LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(LogMessages.IDX10668, LogHelper.MarkAsNonPII(_className), LogHelper.MarkAsNonPII(Algorithm), Key)));
 
-            ValidateKeySize(Key, Algorithm);
```
GeoK: Why was this line removed?

kllysng: This might have been from a bad merge and should not have been removed. Thanks for the catch.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
kllysng: Added aka.ms link with a wiki page, updated log message, added algorithm, an expected/actual tag lengths.

src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
kllysng: Updated

GeoK (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
JoshLozensky: Defining the vars in the TheoryData method similar to signingCredentials512 and then assigning the variables to the CreateTokenTheoryData members "Payload", "JsonWebTokenHandler", and "SigningCredentials" would make the test cleaner though you are correct that total lines would increase as there would be 3 more lines per theory data. Currently 13 of the 30 lines (43%) in the test are just defining these variables and don't seem to add value to reader comprehension.

Ultimately your choice, imho since theory data is meant to hold the setup while the test is meant to hold the logic it seems to make more conceptual sense to put the vars in the theory data, esp. since it already has members for them but it is just a suggestion.

PR #2573, opened 2024-04-28
Add TimeProvider support to token validation

### LINKED ISSUE #2572: Use TimeProvider instead of DateTime ambient context
Currently, token lifetime validation is performed by getting the current time from a DateTime ambient context, as [seen here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/d353b5a67200933361941550e99d4a15a11de04e/src/Microsoft.IdentityModel.Tokens/ValidatorUtilities.cs#L38).

(I know there's other places where DateTime is used to get the current time, this is just an  example)

This leads to flaky, non-deterministic tests, as there's no control of the current time whatsoever.

With the [introduction of TimeProvider in .NET 8](https://blog.nimblepros.com/blogs/finally-an-abstraction-for-time-in-net/), there's an out-of-the-box solution for abstracting time.

My proposal here is adopt [TimeProvider](https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider?view=net-8.0) instead of relying directly on static DateTime class.

alexmurari:
# Add TimeProvider support to token validation

TimeProvider abstracts time and enables deterministic tests.

## Description

Added a new parameter to the `TokenValidationParameters` class to hold the `TimeProvider` instance.
The default behavior's still the use of `DateTime` class to obtain the current time.

Fixes #2572


alexmurari:

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
```diff
@@ -23,6 +23,10 @@
...
   </PropertyGroup>
 
+  <PropertyGroup>
+    <FrameworksWithoutTimeProvider>|net461|net462|net472|netstandard2.0|net6.0|</FrameworksWithoutTimeProvider>
```
alexmurari, line 27: `TimeProvider` is included by default in .NET 8 onwards. For every other target, we must eat the `Microsoft.Bcl.TimeProvider` nuget.

test/Microsoft.IdentityModel.Tokens.Tests/Microsoft.IdentityModel.Tokens.Tests.csproj
```diff
@@ -24,6 +24,10 @@
     <PackageReference Include="Microsoft.Azure.KeyVault.Cryptography" Version="$(MicrosoftAzureKeyVaultCryptographyVersion)" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' != 'net461'">
```
alexmurari, line 27: Every test target, except .NET FWK 4.6.1 supports the `Microsoft.Extensions.TimeProvider.Testing`.

build/common.props
```diff
@@ -55,6 +55,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
```
alexmurari, line 58: Suppress "_Microsoft.Bcl.TimeProvider doesn't support .NET Framework 4.6.1, consider upgrading [...]_" warnings. (Altough it's _compatible_ with that framework version).

build/commonTest.props
```diff
@@ -25,8 +25,10 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
```
alexmurari, line 28: [Same](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2573/files#r1582438994).

build/commonTest.props
```diff
@@ -25,8 +25,10 @@
...
     <NoWarn>$(NoWarn);SYSLIB0050</NoWarn>
     <NoWarn>$(NoWarn);SYSLIB0051</NoWarn>
+    <NoWarn>$(NoWarn);NU1701</NoWarn>
```
alexmurari, line 31: Suppresses the "_This package may not be fully compatible with your project_" warning. The cause is that `Microsoft.Extensions.TimeProvider.Testing` package is not compatible with  .NET FWK 4.6.1.

pmaytak:
@alexmurari Thanks for the proposal.

After team discussion, there is more design and work needed here. (Our PR process is described in the [contributing guide](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/Contributing.md#before-writing-code), making sure a design is discussed before time is spent making changes.)

- We're careful with adding new dependencies (MS.Bcl.TimeProvider) as it can create conflicts higher in the stack.
  - So we can conditionally use TimeProvider only in .NET 8.
- We need to make sure adding TimeProvider property to TokenValidationParameters is the right thing to do as it's a commonly used class.

Proposal could be:
- Create an internal time provider abstraction, e.g. IInternalTimeProvider.
- On .NET 8 it will use .NET's TimeProvider, on other platforms - the current DateTime class.
- Update all DateTime related code to use this abstraction, including tests.
- This will allow us more easily to add MS.Bcl.TimeProvider package in the future, if it's the right thing to do.

alexmurari:
@pmaytak Thanks for the review and the modified proposal.

I agree to those points.

Client code would still need to provide a `TimeProvider` instance. Since `DateTime` usage is ubiquitous, we have two possibilities:

a. (Constructor/Method injection) Every class/method that needs a `TimeProvider` will have new parameters to receive it. (This kinda defeats the `IInternalTimeProvider` purpose).

**- OR -**

b. (Property-injection) We create an options class (e.g. `TimeProviderOptions`) that has an `TimeProvider` property. It's suitable for property-injection because it has a good local default: `TimeProvider.System`. It's basically an extension point.

``` csharp
public class TimeProviderOptions
{
    private TimeProvider _timeProvider = TimeProvider.System; // Local default

    public TimeProvider TimeProvider
    {
        get
        {
            return _timeProvider;
        }
        set
        {
            if (value == null) 
               throw new ArgumentNullException("value");

            _timeProvider = value;
        }
    }
}
```

We just need somewhere to set an instance of this options class, then the `IInternalTimeProvider` would use it to obtain the TimeProvider instance.

What do you think?

trejjam:
We used a similar path with TimeProvider in OpenIddict:
https://github.com/openiddict/openiddict-core/pull/2071/files#diff-63fa24a3f983e94df2b11646deac927651bd5a5836f02d3d1de6dbd30f61e5baR1092

One of the constraints was also not to use the Bcl package. So there is `#SUPPORTS_TIME_PROVIDER` used instead

trejjam:
This way, the PR becomes simple and minimalistic. I am willing to prepare PR using this approach.

trejjam:
A workaround can look like this:

```cs
private const string TimeProviderName = "TimeProviderName";

/// <summary>
/// Validates the lifetime of a <see cref="SecurityToken"/>.
/// </summary>
internal static bool ValidateLifetime(DateTime? notBefore, DateTime? expires, SecurityToken securityToken, TokenValidationParameters validationParameters)
{
    if (!expires.HasValue && validationParameters.RequireExpirationTime)
    {
        throw LogHelper.LogExceptionMessage(new SecurityTokenNoExpirationException(
            LogHelper.FormatInvariant("IDX10225", LogHelper.MarkAsNonPII(securityToken.GetType().ToString()))
        ));
    }

    if (notBefore.HasValue && expires.HasValue && notBefore.Value > expires.Value)
    {
        throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidLifetimeException(
            LogHelper.FormatInvariant("IDX10224", LogHelper.MarkAsNonPII(notBefore.Value), LogHelper.MarkAsNonPII(expires.Value))
        )
        {
            NotBefore = notBefore,
            Expires = expires,
        });
    }

    var timeProvider = (TimeProvider) validationParameters.InstancePropertyBag[TimeProviderName];

    var utcNow = timeProvider.GetUtcNow().UtcDateTime;
    if (notBefore.HasValue && notBefore.Value > DateTimeUtil.Add(utcNow, validationParameters.ClockSkew))
    {
        throw LogHelper.LogExceptionMessage(new SecurityTokenNotYetValidException(
            LogHelper.FormatInvariant("IDX10222", LogHelper.MarkAsNonPII(notBefore.Value), LogHelper.MarkAsNonPII(utcNow))
        )
        {
            NotBefore = notBefore.Value,
        });
    }

    if (expires.HasValue && expires.Value < DateTimeUtil.Add(utcNow, validationParameters.ClockSkew.Negate()))
    {
        throw LogHelper.LogExceptionMessage(new SecurityTokenExpiredException(
            LogHelper.FormatInvariant("IDX10223", LogHelper.MarkAsNonPII(expires.Value), LogHelper.MarkAsNonPII(utcNow))
        )
        {
            Expires = expires.Value,
        });
    }

    // if it reaches here, that means lifetime of the token is valid
    LogHelper.LogInformation("IDX10239");
    return true;
}

var timeProvider = TimeProvider.System;

var validationParameters = new TokenValidationParameters
{
  LifetimeValidator = ValidateLifetime,
}
validationParameters.InstancePropertyBag.Add(TimeProviderName, timeProvider);
```

alexmurari:
@trejjam That's a great approach! The changes to OpenIddict were minimal.

Fine by me a new PR with this approach. I'm just not sure where client code will configure the desired `TimeProvider` instance.

trejjam:
Hi my PR is here: #2612

alexmurari:
Closing in favor of #2612

PR #2578, opened 2024-05-01
Fix JwtSecurityToken Missing Mapping When Creating a Token. 

### LINKED ISSUE #2089: [Bug] JWE header alg attribute with value http://www.w3.org/2001/04/xmlenc#rsa-oaep instead of RSA-OAEP

Microsoft.IdentityModel 6.30.1

**Where is the issue?**
 * [ X] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [ X] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [X ] S.IM.Tokens.Jwt
 * Other (please describe)

The app is in production and next release requires interoperability with other apps written in different languages (ie: GoLang)

**Repro**

*SigningCredentials and EncryptingCredentials was created using a X509Certificate2 with public and private keys

```csharp
            SecurityTokenDescriptor tokenDescriptor = new SecurityTokenDescriptor
            {
                Issuer = _openIdConfigurationService.OpenIdConfiguration.Issuer,
                IssuedAt = issuedAt,
                Subject = request.Claims,
                NotBefore = issuedAt.AddSeconds(-1),
                Expires = expires,
                SigningCredentials = credentials.SigningCredentials,
                EncryptingCredentials = credentials.EncryptingCredentials,
                TokenType = "JWE"
            };

            JwtSecurityTokenHandler tokenHandler = new JwtSecurityTokenHandler();
            JwtSecurityToken token = tokenHandler.CreateJwtSecurityToken(tokenDescriptor);
```

**Expected behavior**

Algorithm (alg) should be one of the registered at IANA "JSON Web Signature and Encryption Algorithms" (https://www.iana.org/assignments/jose/jose.xhtml) as described on topic 4.1.1 of RFC7516 (https://datatracker.ietf.org/doc/html/rfc7516#page-12) 

```json
{
  "alg": "RSA-OAEP",
  "enc": "A128CBC-HS256",
  "kid": "0758B2A8DF6B980E569421870D7CFE29E65D26EB",
  "typ": "JWE",
  "cty": "JWT"
}
```

**Actual behavior**
```json
{
  "alg": "http://www.w3.org/2001/04/xmlenc#rsa-oaep",
  "enc": "A128CBC-HS256",
  "kid": "0758B2A8DF6B980E569421870D7CFE29E65D26EB",
  "typ": "JWE",
  "cty": "JWT"
}
```

**Possible solution**
Change the KeyWrapper providers to declare header alg complaint to IANA document

FuPingFranco:
## Description

Adds mapping to Algorithm (alg)  so it should be one of the registered at IANA "JSON Web Signature and Encryption Algorithms" (https://www.iana.org/assignments/jose/jose.xhtml) as described on topic 4.1.1 of RFC7516 (https://datatracker.ietf.org/doc/html/rfc7516#page-12).

Fixes #2089


test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
```diff
@@ -55,6 +55,31 @@ public void JwtSecurityTokenHandler_CreateToken_SameTypeMultipleValues()
                 Assert.NotNull(aTypeClaims.SingleOrDefault(c => c.Value == value));
         }
 
+        [Fact]
+        public void JwtSecurityTokenHandler_CreateToken_AddShortFormMappingForRsaOAEP()
+        {
+            AppContext.SetSwitch(JwtSecurityTokenHandler._enableRsaOaepMappingSwitch, true); //Set to false to test the default behavior and adjust the expected values accordingly.
```
westin-m: I think another test case with the default behavior would be better than having to adjust an existing one.

westin-m: nit: Comments should go directly above the line to which they refer

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -35,6 +35,7 @@ public class JwtSecurityTokenHandler : SecurityTokenHandler
         private Dictionary<string, string> _outboundAlgorithmMap = null;
         private static string _shortClaimType = _namespace + "/ShortTypeName";
         private bool _mapInboundClaims = DefaultMapInboundClaims;
+        internal const string _enableRsaOaepMappingSwitch = "Switch.Microsoft.IdentityModel.EnableRsaOaepShortFormMapping";
```
kllysng: Please add `EnableRsaOaepShortFormMapping` to the doc I started yesterday: [App Context Switches in IdentityModel 7x](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/App-Context-Switches-in-IdentityModel-7x) when the PR is ready

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -70,8 +71,16 @@ public class JwtSecurityTokenHandler : SecurityTokenHandler
             { SecurityAlgorithms.RsaSha256Signature, SecurityAlgorithms.RsaSha256 },
             { SecurityAlgorithms.RsaSha384Signature, SecurityAlgorithms.RsaSha384 },
             { SecurityAlgorithms.RsaSha512Signature, SecurityAlgorithms.RsaSha512 },
+            { SecurityAlgorithms.RsaOaepKeyWrap, GetRsaOaepMapping() },
```
brentschmaltz: This issue is also existing in JsonWebTokenHandler.
If we make the switch in the ctor of X509EncryptingCredentials we will modify both cases.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
kllysng: updating the [link](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/App-Context-Switches-in-IdentityModel) now that it's documenting not just 7x app context switches!

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
FuPingFranco: Done, Thanks Westin!

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
FuPingFranco: Moved changes to the X509EncryptingCredentials and did tests for both handlers. Thanks Brent!

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
FuPingFranco: Will do once this PR is in, Thanks Kelly!

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.IdentityModel.Tokens
     /// </summary>
     public class X509EncryptingCredentials : EncryptingCredentials
     {
+        internal const string _useShortNameForRsaOaepKey = "Switch.Microsoft.IdentityModel.UseShortNameForRsaOaepKey";
```
pmaytak, line 14:
```suggestion
        internal const string UseShortNameForRsaOaepKey = "Switch.Microsoft.IdentityModel.UseShortNameForRsaOaepKey";
```

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -48,5 +50,10 @@ public X509Certificate2 Certificate
             get;
             private set;
         }
+
+        private static bool GetRsaOaepMapping()
```
pmaytak:
```suggestion
        private static bool ShouldUseShortNameForRsaOaepKey()
```

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
```diff
@@ -18,6 +18,7 @@
 
 namespace System.IdentityModel.Tokens.Jwt.Tests
 {
+    [CollectionDefinition("JwtSecurityTokenHandlerTests", DisableParallelization = true)]
```
pmaytak: Why is this needed? Do tests fail running in parallel? Unit tests should be independent of each other.

FuPingFranco: Yeap, AppContext is a static dict so, it gives me inconsistent behavior when setting the AppContext when tests run in parallel causing them to fail intermittently.

pmaytak: Hmm, this should be inconsistent too then...

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/28e8784bf62126ea1071cff754e709b763d94cbd/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs#L4196

pmaytak: What about moving the tests that use AppContext into a separate file, like JwtSecurityTokenHandlerTests.WithContextSwitchcs or something, and disable parallelization only there; so that tests in this class will run in parallel.

FuPingFranco: I agree, They do cause a bit of an issue. I'll refactor this. Thanks Peter!

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
```diff
@@ -55,6 +56,108 @@ public void JwtSecurityTokenHandler_CreateToken_SameTypeMultipleValues()
...
+                EncryptingCredentials = encryptingCredentials,
+                TokenType = "JWE"
+            };
+
+            JwtSecurityTokenHandler tokenHandler = new JwtSecurityTokenHandler();
+            JwtSecurityToken token = tokenHandler.CreateJwtSecurityToken(tokenDescriptor);
+
+            Assert.NotNull(token);
+            Assert.Equal(token.Header.Alg, SecurityAlgorithms.RsaOAEP);
+        }
```
pmaytak: Reset the switch so it doesn't affect other tests.
`AppContext.SetSwitch(X509EncryptingCredentials._useShortNameForRsaOaepKey, false);`

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
FuPingFranco: Updated. Thanks Peter!

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
FuPingFranco: This follows the naming convention of other switches used in Wilson. Thanks Peter!

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
FuPingFranco: I moved the test pertinent to this work to a separate file. Will create a task to potentially move the other tests on a separate PR if needed as they do not seem to encounter the same issue.

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
JoshLozensky, line 14: We aren't supposed to be introducing new internals unless absolutely necessary... Does this need to be internal?

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerNonParallelRunTests.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Xunit;
```
JoshLozensky: Why do we have this empty test file?

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,115 @@
...
+using Microsoft.IdentityModel.Tokens;
+using Xunit;
+
+namespace System.IdentityModel.Tokens.Jwt.Tests
+{
+    [Collection(nameof(JwtSecurityTokenHandlerNonParallelRunTests))]
+    public class JwtSecurityTokenHandlerTestsWithContextSwitches
+    {
+        [Fact]
+
```
JoshLozensky: nit blank line after fact

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,115 @@
...
+        public void JwtSecurityTokenHandler_CreateToken_AddShortFormMappingForRsaOAEPEnabled()
```
JoshLozensky: It looks like these tests are all fairly similar, can we split them into TheoryData and Theory to make it easier to read?

JoshLozensky (changes requested):

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
FuPingFranco: It's actually two different handlers so the set-up won't work. It's more work to use theories than what the tests are testing.

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
FuPingFranco: Why are we not supposed to be introducing internals? So far the switches we have for Wilson are set-up this way. Refer to https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/909669f4053f485ad94223fa51e2b8be2f61a76f/src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs#L36

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerNonParallelRunTests.cs
FuPingFranco: This leverages CollectionDefinition, we can use this class as the collection name to disable other test that we might not want to be ran parallel in the future.

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
pmaytak: Internals are okay, they're needed for tests. But we should be careful using internals between the source libraries.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -35,7 +35,7 @@ public class JwtSecurityTokenHandler : SecurityTokenHandler
         private Dictionary<string, string> _outboundAlgorithmMap = null;
         private static string _shortClaimType = _namespace + "/ShortTypeName";
         private bool _mapInboundClaims = DefaultMapInboundClaims;
-
+
```
kllysng, line 38: If this is the only change in this file, please remove diff 😊

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerNonParallelRunTests.cs
kllysng:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

﻿using Xunit;
```
We'll have analyzers for stuff like this soon 😊

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,115 @@
+﻿using System.Security.Claims;
```
kllysng:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

﻿using System.Security.Claims;
```

FuPingFranco: Thanks Kelly!

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
FuPingFranco: This makes it easier to read since it's the begining of the public properties, will probably leave this in. Thanks Kelly

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerNonParallelRunTests.cs
pmaytak: Do you mean is someone already working on adding them or in general?

pmaytak: @FuPingFranco You can add this rule (in a separate PR) to the main .editorconfig like we do in MSAL.NET.

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/c9f26e81c114fc9c4bfa4175b7702592e6cb6a90/.editorconfig#L423-L425

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
JoshLozensky: Building on Peter's answer, in the past we have used internals between source libraries but this has been in large part what causes our libraries to break when there are mismatching versions and we are now trying to remove existing internals between libraries and avoid adding more. If this will indeed only ever be used for testing then afaik it is fine but if it is between different libraries even in the same repo my understanding is we are trying to avoid that.

Feel free to resolve if its only purpose is testing.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerNonParallelRunTests.cs
JoshLozensky: Ah makes sense

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,115 @@
...
+    [Collection(nameof(JwtSecurityTokenHandlerNonParallelRunTests))]
+    public class JwtSecurityTokenHandlerTestsWithContextSwitches
+    {
+        [Fact]
+
+        public void JwtSecurityTokenHandler_CreateToken_AddShortFormMappingForRsaOAEPEnabled()
+        {
+            AppContext.SetSwitch(X509EncryptingCredentials._useShortNameForRsaOaepKey, true);
+            var encryptingCredentials = new X509EncryptingCredentials(Default.Certificate);
+            SecurityTokenDescriptor tokenDescriptor = new SecurityTokenDescriptor
```
JoshLozensky: All of the token descriptors are the same, can we make this once and use it for all? Alternatively put it in a private helper method if you want a new one each time, it is very time consuming for a reader to have to compare them all to find out there is no difference. 

Additionally, for the timespan magic numbers, could you put them into descriptive variables like oneMin, tenSec, etc... or even just make TimeSpan variables with descriptive names? Otherwise a reader will need to look up the TimeSpan implementation to understand what is happening.

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
FuPingFranco: @JoshLozensky I believe the intent of the team is to not use InternalsVisibleTo for development purposes but, using "internal" keyword it's fine. In this case we would never want anyone to be able to access the name of the switch and try to do funny stuff with it.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerNonParallelRunTests.cs
FuPingFranco: @pmaytak I meant in general. Since there's other switches that might run into this issue.

FuPingFranco: I'll take Kelly suggestion for this one but it's good to know I can also do it on the  .editorconfig. Thanks!

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
JoshLozensky: @FuPingFranco here is the comment

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,117 @@
...
+                EncryptingCredentials = encryptingCredentials,
+                TokenType = "JWE"
+            };
+
+            JsonWebTokenHandler tokenHandler = new JsonWebTokenHandler();
+            var token = tokenHandler.CreateToken(tokenDescriptor);
+            JsonWebToken jsonToken = tokenHandler.ReadToken(token) as JsonWebToken;
+
+            Assert.NotNull(jsonToken);
+            Assert.Equal(jsonToken._alg, SecurityAlgorithms.RsaOaepKeyWrap);
```
brentschmaltz: Why not use jsonWebToken.Alg?

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,117 @@
...
+                NotBefore = DateTime.UtcNow.Subtract(new TimeSpan(1, 0, 0)),
+                Expires = DateTime.UtcNow.Subtract(new TimeSpan(0, 10, 0)),
+                SigningCredentials = Default.AsymmetricSigningCredentials,
+                EncryptingCredentials = encryptingCredentials,
+                TokenType = "JWE"
+            };
+
+            JsonWebTokenHandler tokenHandler = new JsonWebTokenHandler();
+            var token = tokenHandler.CreateToken(tokenDescriptor);
+            JsonWebToken jsonToken = tokenHandler.ReadToken(token) as JsonWebToken;
```
brentschmaltz: You could write: JsonWebToken jsonToken = new JsonWebToken(token);

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,117 @@
...
+            Assert.NotEqual(jsonToken._alg, SecurityAlgorithms.RsaOaepKeyWrap);
+            Assert.Equal(jsonToken._alg, SecurityAlgorithms.RsaOAEP);
+        }
+
+        [Fact]
+        public void JsonWebTokenHandler_CreateToken_AddShortFormMappingForRsaOAEPDisabled()
+        {
+            AppContext.SetSwitch(X509EncryptingCredentials._useShortNameForRsaOaepKey, false);
+            var encryptingCredentials = new X509EncryptingCredentials(Default.Certificate);
+            SecurityTokenDescriptor tokenDescriptor = new SecurityTokenDescriptor
```
brentschmaltz: As a note, since you are just testing the 'alg' value in the header, you could skip a lot of the values.
'Issuer' probably would be enough.

brentschmaltz: you could write
```
var token = tokenHandler.CreateToken(new SecurityTokenDescriptor{
    {
        Issuer = Default.Issuer,
         SigningCredentials = Default.AsymmetricSigningCredentials,
         EncryptingCredentials =  new X509EncryptingCredentials(Default.Certificate),
    });

JsonWebToken jsonToken = new JsonWebToken(token);
'''

File: test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs [hunk 17]
```diff
@@ -0,0 +1,117 @@
...
+                SigningCredentials = Default.AsymmetricSigningCredentials,
+                EncryptingCredentials = encryptingCredentials,
+                TokenType = "JWE"
+            };
+
+            JsonWebTokenHandler tokenHandler = new JsonWebTokenHandler();
+            var token = tokenHandler.CreateToken(tokenDescriptor);
+            JsonWebToken jsonToken = tokenHandler.ReadToken(token) as JsonWebToken;
+
+            Assert.NotNull(jsonToken);
```
brentschmaltz: no need for this assert.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
brentschmaltz: Creating the token could be in a sub-rountine, 
CreateToken().

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
pmaytak: I'd argue they are incorrect (I think there's only one other switch here?) but we are inconsistent with naming throughout...

Usually (per Microsoft guidelines) constants are in PascalCase.

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -48,5 +50,10 @@ public X509Certificate2 Certificate
             get;
             private set;
         }
+
+        private static bool ShouldUseShortNameForRsaOaepKey()
+        {
+            return AppContext.TryGetSwitch("Switch.Microsoft.IdentityModel.UseShortNameForRsaOaepKey", out var useKeyWrap) && useKeyWrap;
```
pmaytak: Use the const here.

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
```diff
@@ -21,7 +23,7 @@ public class X509EncryptingCredentials : EncryptingCredentials
         /// </remarks>
         /// <exception cref="ArgumentNullException">if 'certificate' is null.</exception>
         public X509EncryptingCredentials(X509Certificate2 certificate)
-            : this(certificate, SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm, SecurityAlgorithms.DefaultSymmetricEncryptionAlgorithm)
+            : this(certificate, ShouldUseShortNameForRsaOaepKey() ? SecurityAlgorithms.RsaOAEP : SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm, SecurityAlgorithms.DefaultSymmetricEncryptionAlgorithm)
```
pmaytak: You can also extract this `ShouldUseShortNameForRsaOaepKey() ? SecurityAlgorithms.RsaOAEP : SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm,` into a separate method to make this constructor call a bit simpler to read.

```csharp
private static string GetRsaOeapKeyName()
{
return AppContext.TryGetSwitch("Switch.Microsoft.IdentityModel.UseShortNameForRsaOaepKey", out var useKeyWrap) && useKeyWrap ?
SecurityAlgorithms.RsaOAEP :
SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm;
}
```

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerNonParallelRunTests.cs
pmaytak: > This leverages CollectionDefinition, we can use this class as the collection name to disable other test that we might not want to be ran parallel in the future.

Add this as a comment to the class.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
pmaytak: The underscore variables should really be private...

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,117 @@
...
+
+﻿using System.Security.Claims;
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.TestUtils;
+using Microsoft.IdentityModel.Tokens;
+using Xunit;
+
+namespace System.IdentityModel.Tokens.Jwt.Tests
+{
+    [Collection(nameof(JwtSecurityTokenHandlerNonParallelRunTests))]
```
pmaytak: I think you can just pass the string name like you had before for simplicity (so we don't have to worry about a separate class).
```suggestion
    [CollectionDefinition("JwtSecurityTokenHandlerNonParallelTests", DisableParallelization = true)]
```

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,117 @@
...
+                TokenType = "JWE"
+            };
+
+            JwtSecurityTokenHandler tokenHandler = new JwtSecurityTokenHandler();
+            JwtSecurityToken token = tokenHandler.CreateJwtSecurityToken(tokenDescriptor);
+
+            Assert.NotNull(token);
+            Assert.NotEqual(token.Header.Alg, SecurityAlgorithms.RsaOaepKeyWrap);
+            Assert.Equal(token.Header.Alg, SecurityAlgorithms.RsaOAEP);
+        }
```
pmaytak, line 28: Disable the context switch at the end of each test so it doesn't affect other tests.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,117 @@
...
+                EncryptingCredentials = encryptingCredentials,
+                TokenType = "JWE"
+            };
+
+            JwtSecurityTokenHandler tokenHandler = new JwtSecurityTokenHandler();
+            JwtSecurityToken token = tokenHandler.CreateJwtSecurityToken(tokenDescriptor);
+
+            Assert.NotNull(token);
+            Assert.NotEqual(token.Header.Alg, SecurityAlgorithms.RsaOaepKeyWrap);
+            Assert.Equal(token.Header.Alg, SecurityAlgorithms.RsaOAEP);
```
pmaytak: These are mutually exclusive. You can just assert the equal.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,117 @@
...
+                EncryptingCredentials = encryptingCredentials,
+                TokenType = "JWE"
+            };
+
+            JwtSecurityTokenHandler tokenHandler = new JwtSecurityTokenHandler();
+            JwtSecurityToken token = tokenHandler.CreateJwtSecurityToken(tokenDescriptor);
+
+            Assert.NotNull(token);
+            Assert.NotEqual(token.Header.Alg, SecurityAlgorithms.RsaOAEP);
+            Assert.Equal(token.Header.Alg, SecurityAlgorithms.RsaOaepKeyWrap);
```
pmaytak, line 62: For long name, assert `SecurityAlgorithms.DefaultAsymmetricKeyWrapAlgorithm` since that's what we pass into the constructor based on the flag.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -0,0 +1,117 @@
...
+using Microsoft.IdentityModel.Tokens;
+using Xunit;
+
+namespace System.IdentityModel.Tokens.Jwt.Tests
+{
+    [Collection(nameof(JwtSecurityTokenHandlerNonParallelRunTests))]
+    public class JwtSecurityTokenHandlerTestsWithContextSwitches
+    {
+        [Fact]
+        public void JwtSecurityTokenHandler_CreateToken_AddShortFormMappingForRsaOAEPEnabled()
```
pmaytak: You can combine these tests using `InlineData` - (bool enableSwitch, string expectedAlgorithm)

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/909669f4053f485ad94223fa51e2b8be2f61a76f/test/Microsoft.IdentityModel.Tokens.Tests/CryptoExtensibilityTests.cs#L151

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
FuPingFranco: I was curious why were they not private when working on them? I changed it to use 'Alg' same as in JwtSecurityTokenHandler. Thanks Brent & Peter!

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
FuPingFranco: Done. Thanks Brent!

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
FuPingFranco: This is good to know, I've made the changes. Thanks Brent!

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
FuPingFranco: Removed

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
FuPingFranco: Done

src/Microsoft.IdentityModel.Tokens/X509EncryptingCredentials.cs
FuPingFranco: Done. Thanks Peter!

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
FuPingFranco: Makes sense. Thanks Peter!

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
FuPingFranco: Done

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
FuPingFranco: Removed additional assert

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
FuPingFranco: You're right, forgot about inlinedata. Thanks Peter!

JoshLozensky (approved):

pmaytak (approved):

PR #2582, opened 2024-05-04
Update JsonWebToken to enable extensibility

### LINKED ISSUE #2581: [Feature Request] Update JsonWebToken to enable extensibility
Update `JsonWebToken` - extract and expose the method that reads the header/payload property values from the reader so it can be overridden in child classes to add any extra own logic.

Related to #2583, #2495.

pmaytak:
Fixes #2581.

Extract token reading code into a separate `ReadPayloadValue` so it can be overridden with custom logic in child classes.
Make `ReadPayloadValue` private protected (accessible by types derived from the containing class, but only within its containing assembly or friend assemblies).

Example derived class:
```csharp
    public class CustomJsonWebToken : JsonWebToken
    {
        public CustomJsonWebToken (string jwtEncodedString) : base(jwtEncodedString) { }

        /// <summary>
        /// Tries to read custom claims first.
        /// If nothing matched, reads common claims and the rest of unnamed claims.
        /// </summary>
        private protected override void ReadPayloadValue(ref Utf8JsonReader reader, Dictionary<string, object> claims)
        {
            if (reader.ValueTextEquals("CustomClaim"))
            {
                _customClaim = JsonSerializerPrimitives.ReadString(ref reader, "CustomClaim", ClassName, true);
                claims["CustomClaim"] = _customClaim ;
            }
            else
            {
                base.ReadPropertyValue(ref reader, claims);
            }
        }

        private string _customClaim;

        public string CustomClaim
        {
            get
            {
                _customClaim ??= Payload.GetStringValue("CustomClaim");
                return _customClaim;
            }
        }
    }
```

There was no change in performance.


src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
```diff
@@ -112,5 +48,74 @@ internal JsonClaimSet CreatePayloadClaimSet(ReadOnlySpan<byte> byteSpan)
 
             return new JsonClaimSet(claims);
         }
+
+        private protected virtual void ReadPropertyValue(ref Utf8JsonReader reader, Dictionary<string, object> claims)
```
brentschmaltz: Passing the Dictionary will make sure TryGetPayload still works for custom claims.
A complete solution for users would include overriding TryGetPayload.

Should we change the Dictionary to IDictionary?

brentschmaltz: For now let's make the method internal protected.

src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonWebToken.PayloadClaimSet.cs
```diff
@@ -112,5 +48,74 @@ internal JsonClaimSet CreatePayloadClaimSet(ReadOnlySpan<byte> byteSpan)
 
             return new JsonClaimSet(claims);
         }
+
+        private protected virtual void ReadPayloadValue(ref Utf8JsonReader reader, IDictionary<string, object> claims)
```
keegan-caruso, line 52: small test showing the extensibility?

pmaytak: Added.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/CustomJsonWebToken.cs
kllysng, line 1: This file makes me smile 😄

keegan-caruso (approved):

kllysng (approved):

PR #2584, opened 2024-05-06
Remove Task allocation from AadIssuerValidator

eerhardt:
There's no need to await this task just to return its result. We can just return the underlying Task from configurationManager.GetBaseConfigurationAsync.

cc @keegan-caruso @jennyf19 @jmprieur


keegan-caruso (approved):

pmaytak (approved):

jmprieur (approved):
LGTM
Thanks @eerhardt

PR #2586, opened 2024-05-06
Reduce Allocations in ValidateSignature

eerhardt:
When we need to create an enumerable with a single element in it, we are allocating a `new List<SecurityKey> { key }`. This allocates both the list, and then allocates a 4 element array backing the list (because this syntax is C# sugar for just calling Add on the collection).

Instead, use a collection expression which optimizes the single element case.

There was one method in Saml that had this pattern in it, but was unused. I opted to remove it. Please let me know if it should be preserved for some reason.

Using an internal benchmark with 1,000 iterations, these allocations will be cut in half:

image

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
@@ -1311,7 +1311,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 }
                     
                 if (key != null)
-                    keys = new List<SecurityKey> { key };
+                    keys = [key];
             }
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
```diff
@@ -1311,7 +1311,7 @@ internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken
                 }
                     
                 if (key != null)
-                    keys = new List<SecurityKey> { key };
+                    keys = [key];
```
keegan-caruso, line 1314: what will the type end up as in this case?

keegan-caruso (approved):

eerhardt:
See https://github.com/dotnet/roslyn/pull/71147 for what Roslyn does in these cases. It generates a `<>z__ReadOnlySingleElementList` class that is used for the single element case.

pmaytak (approved):

PR #2589, opened 2024-05-09
Remove byte[] allocation in VerifyRsa / VerifyECDsa

eerhardt:
Calling HashAlgorithm.ComputeHash will always allocate and return a new byte array. On .NET 6+ we can instead use a temporary buffer (either on the stack for normal hashing, or use the array pool for hash algorithms that need more than 2k bits) to reduce the intermediate allocation.

cc @bartonjs - in case you want to take a look.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
@@ -409,19 +409,22 @@ namespace Microsoft.IdentityModel.Tokens
             Span<byte> hash = hashByteLength <= 256 ? stackalloc byte[256] : array = ArrayPool<byte>.Shared.Rent(hashByteLength);
             hash = hash.Slice(0, hashByteLength);
 
-            bool hashResult = HashAlgorithm.TryComputeHash(bytes, hash, out int bytesWritten);
-            Debug.Assert(hashResult && bytesWritten == hashByteLength, "HashAlgorithm.TryComputeHash failed");
-
-            bool result = isRSA ?
-                RSA.VerifyHash(hash, signature, HashAlgorithmName, RSASignaturePadding) :
-                ECDsa.VerifyHash(hash, signature);
-
-            if (array is not null)
+            try
             {
-                ArrayPool<byte>.Shared.Return(array, clearArray: true);
-            }
+                bool hashResult = HashAlgorithm.TryComputeHash(bytes, hash, out int bytesWritten);
+                Debug.Assert(hashResult && bytesWritten == hashByteLength, "HashAlgorithm.TryComputeHash failed");
 
-            return result;
+                return isRSA ?
+                    RSA.VerifyHash(hash, signature, HashAlgorithmName, RSASignaturePadding) :
+                    ECDsa.VerifyHash(hash, signature);
+            }
+            finally
+            {
+                if (array is not null)
+                {
+                    ArrayPool<byte>.Shared.Return(array, clearArray: true);
+                }
+            }
         }
 #endif
 
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
@@ -2,6 +2,10 @@
 // Licensed under the MIT License.
 
 using System;
+#if NET6_0_OR_GREATER
+using System.Buffers;
+using System.Diagnostics;
+#endif
 using System.Security.Cryptography;
 using Microsoft.IdentityModel.Logging;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -373,13 +385,45 @@ private byte[] SignUsingOffsetRsa(byte[] bytes, int offset, int count)
...
             return RSA.VerifyHash(HashAlgorithm.ComputeHash(bytes), signature, HashAlgorithmName, RSASignaturePadding);
+#endif
         }
 
         private bool VerifyUsingOffsetRsa(byte[] bytes, int offset, int count, byte[] signature)
         {
+#if NET6_0_OR_GREATER
+            return VerifyUsingSpan(isRSA: true, bytes.AsSpan(offset, count), signature);
+#else
             return RSA.VerifyHash(HashAlgorithm.ComputeHash(bytes, offset, count), signature, HashAlgorithmName, RSASignaturePadding);
```
bartonjs, line 400: Since you already have the HashAlgorithmName, why not just call VerifyData and let the framework do the hashing for you (which won't create a temporary array)?

eerhardt: > and let the framework do the hashing for you (which won't create a temporary array)?

This doesn't appear to be the case. When I step into it, I get:

https://github.com/dotnet/runtime/blob/9926d609dab5929941b7924a04521ea2c2df27f3/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/RSA.cs#L661-L662

Which allocates the `byte[]`.

Here is my test code (on windows):

```C#

RSA rSA = RSA.Create();

byte[] data = [0x1, 0x2, 0x3, 0x4];

byte[] signature = rSA.SignData(data, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);

for (int i = 0; i < 1_000; i++)
{
    rSA.VerifyData(data, signature, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
}
```

image

eerhardt: Should I submit a similar PR to RSA and ECDsa in dotnet/runtime?

bartonjs: It looks like VerifyData(ReadOnlySpan) uses span-hashing, but VerifyData(byte[]) uses array-hashing; because we never override `public virtual bool VerifyData(byte[], int, int, byte[], HashAlgorithmName, RSASignaturePadding)`.  So the change in dotnet/runtime to make everyone happier is probably to override that method on all N of our RSA types.  (RSACryptoServiceProvider, RSACng, RSAOpenSsl, RSASecurityTransforms, RSAAndroid, and maybe up to double that from the "RSA.Create()" types).  Which is going to involve a lot of copy/paste of the argument validation.  Bleh.

Here, though, it means
```C#
return rsa.VerifyData(
#if NET5_OR_GREATER
    bytes.AsSpan(offset, count),
#else
    bytes, offset, count,
#endif
    signature,
    HashAlgorithmName,
    RSASignaturePadding);
```

should take your numbers down to 0 (for .NET5+)

eerhardt: RSA is still allocating here:

https://github.com/dotnet/runtime/blob/9926d609dab5929941b7924a04521ea2c2df27f3/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/RSA.cs#L322

image

`HashData` returns a `new byte[]`.

Test Code:

```C#

RSA algo = RSA.Create();

byte[] data = [0x1, 0x2, 0x3, 0x4];

byte[] signature = algo.SignData(data, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);

for (int i = 0; i < 1_000; i++)
{
    algo.VerifyData(data.AsSpan(), signature, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
}
```

bartonjs: Weird, since it should have hit the `if (this is IRuntimeAlgorithm)` above and bypassed the allocation routine.  Must've missed a marker interface somewhere, too.

eerhardt: In my case, `this` is an `RSABCrypt`, which doesn't implement that interface.

https://github.com/dotnet/runtime/blob/acd31754892ab0431ac2c40038f541ffa7168be7/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/RSABCrypt.cs#L11

On Linux it looks like we get an `RSAWrapper`, which also doesn't implement it:

https://github.com/dotnet/runtime/blob/9926d609dab5929941b7924a04521ea2c2df27f3/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/RSAWrapper.cs#L10

MZOLN (approved):

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -373,13 +385,45 @@ private byte[] SignUsingOffsetRsa(byte[] bytes, int offset, int count)
...
+        {
+            int hashByteLength = HashAlgorithm.HashSize / 8;
+            byte[] array = null;
+            Span<byte> hash = hashByteLength <= 256 ? stackalloc byte[256] : array = ArrayPool<byte>.Shared.Rent(hashByteLength);
+            hash = hash.Slice(0, hashByteLength);
+
+            bool hashResult = HashAlgorithm.TryComputeHash(bytes, hash, out int bytesWritten);
+            Debug.Assert(hashResult && bytesWritten == hashByteLength, "HashAlgorithm.TryComputeHash failed");
+
+            bool result = isRSA ?
```
keegan-caruso: there are a few exceptions that could come from either of these methods, should we move the Shared.Return call to a try-finally?

eerhardt: Done.

keegan-caruso (approved):
LGTM

PR #2597, opened 2024-05-14
Reduce string allocations by comparing issuers in-place

### LINKED ISSUE #2595: [Bug] Reduce allocations in Aad validation by not using string.Replace when it isn't necessary
latest

**Where is the issue?**
 * [ ] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [ ] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [x] M.IM.Validators
 * [ ] M.IM.Xml
 * [ ] S.IM.Tokens.Jwt
 * Other (please describe)

**Repro**

In doing a allocation profile of validating a token 1,000 times, one area that stuck out was using string.Replace in a few places:

image

One case that can be fixed is:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/a1530a0cebd47793979ad6e316622013acb3d9ef/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs#L386-L393

We should be able to do that comparison allocation free.

Similarly this check:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/a1530a0cebd47793979ad6e316622013acb3d9ef/src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs#L76-L97

Should be able to be done without the 2 string allocations that is happening.

kllysng:
Fixes #2595 

### Benchmarks for changes for AadIssuerValidator 
BenchmarkDotNet v0.13.12, Windows 11 (10.0.22631.3593/23H2/2023Update/SunValley3) (Hyper-V)
AMD EPYC 7763, 1 CPU, 16 logical and 8 physical cores
.NET SDK 9.0.100-preview.3.24204.13
  [Host]    : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2
  MediumRun : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2

LaunchCount=4  WarmupCount=10

| Method           | Mean     | Error    | StdDev   | Median   | P90      | P95      | P100     | Gen0   | Allocated |
|----------------- |---------:|---------:|---------:|---------:|---------:|---------:|---------:|-------:|----------:|
| Before | 53.03 ns | 0.482 ns | 1.048 ns | 52.65 ns | 54.76 ns | 55.20 ns | 56.80 ns | 0.0086 |     144 B |
| After  | 20.88 ns | 0.026 ns | 0.059 ns | 20.87 ns | 20.94 ns | 20.97 ns | 21.09 ns |      - |         - |

### Benchmarks for changes for AadTokenValidationParametersExtension 
BenchmarkDotNet v0.13.12, Windows 11 (10.0.22631.3593/23H2/2023Update/SunValley3) (Hyper-V)
AMD EPYC 7763, 1 CPU, 16 logical and 8 physical cores
.NET SDK 9.0.100-preview.3.24204.13
  [Host]    : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2
  MediumRun : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2

LaunchCount=4  WarmupCount=10

| Method           | Mean     | Error    | StdDev   | P90      | P95      | P100     | Gen0   | Allocated |
|----------------- |---------:|---------:|---------:|---------:|---------:|---------:|-------:|----------:|
| Before | 86.88 ns | 0.542 ns | 1.202 ns | 88.61 ns | 89.01 ns | 90.18 ns | 0.0162 |     272 B |
| After  | 36.93 ns | 0.345 ns | 0.751 ns | 37.80 ns | 38.10 ns | 40.37 ns | 0.0081 |     136 B |


src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -390,14 +389,62 @@
 
             if (validIssuerTemplate.Contains(TenantIdTemplate))
             {
-                return validIssuerTemplate.Replace(TenantIdTemplate, tenantId) == actualIssuer;
+                return IssuersWithTemplatesAreEqual(validIssuerTemplate, TenantIdTemplate, actualIssuer, tenantId);
             }
             else
             {
```
victorm-hernandez: >IssuersWithTemplatesAreEqual [](http://example.com/codeflow?start=29&length=28)

It would be useful to have an example of the expected strings in a comment.

victorm-hernandez: >issuer1 [](http://example.com/codeflow?start=65&length=7)

These are the values passed: signingKeyIssuer, tenantIdFromToken, tokenIssuer, tenantIdFromToken

The term template here is a bit confusing, what you call issuer1Template here really is the value that will be substituted on the template, why not call it what it is, the tenant Id, the same goes for the last parameter.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -390,14 +389,62 @@
...
                 return validIssuerTemplate == actualIssuer;
             }
         }
 
+        /// <summary>
```
victorm-hernandez: >issuers [](http://example.com/codeflow?start=45&length=7)

both tenant ids

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -77,25 +77,20 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
...
                 // 3. signing key issuers will never match sts.windows.net as v1 endpoint doesn't have issuers attached to keys
                 // v2TokenIssuer is the representation of Token.Issuer (if it was a v2 issuer)
-                if (effectiveSigningKeyIssuer != tokenIssuer && effectiveSigningKeyIssuer != v2TokenIssuer)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidIssuerException(LogHelper.FormatInvariant(LogMessages.IDX40005, LogHelper.MarkAsNonPII(tokenIssuer), LogHelper.MarkAsNonPII(effectiveSigningKeyIssuer))));
+                if (!AadIssuerValidator.IssuersWithTemplatesAreEqual(signingKeyIssuer, tenantIdFromToken, tokenIssuer, tenantIdFromToken)
+                        && !AadIssuerValidator.IssuersWithTemplatesAreEqual(signingKeyIssuer, tenantIdFromToken, openIdConnectConfiguration.Issuer ?? string.Empty, tenantIdFromToken))
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidIssuerException(LogHelper.FormatInvariant(LogMessages.IDX40005, LogHelper.MarkAsNonPII(tokenIssuer), LogHelper.MarkAsNonPII(openIdConnectConfiguration.Issuer))));
```
kllysng: Here, `effectiveSigningKeyIssuer` is replaced by `openIdConnectConfigration.Issuer` because `effectiveSigningKeyIssuer` was originally created using `string.Replace`, and we want to move away from that. The side effect of using `openIdConnectConfiguration.Issuer` is that it _could_ look like 'https://login.microsoftonline.com/{tenantid}/v2.0', instead of 'https://login.microsoftonline.com/f645ad92-e38d-4d1a-b510-d1b09a74a8ca/v2.0'. 

If we don't like that, I can use string interpolation instead of string.Replace for a better allocation profile that `string.Replace` so that we can keep 'https://login.microsoftonline.com/f645ad92-e38d-4d1a-b510-d1b09a74a8ca/v2.0'.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -390,14 +389,62 @@
...
+            // ensure indices exist for both issuers
+            if (firstHalfIssuer1EndIndex == -1  || firstHalfIssuer2EndIndex == -1)
+                return false;
+
+            // if the indices aren't equal, then the issuers aren't equal
+            if (firstHalfIssuer1EndIndex != firstHalfIssuer2EndIndex)
+                return false;
+
+            // ensure the first part of the issuer1 matches the first part of issuer2
+            if (!StringSegmentsAreEqual(issuer1, 0, firstHalfIssuer1EndIndex, issuer2, 0, firstHalfIssuer2EndIndex))
```
victorm-hernandez: You can save the allocation above (line 405-406, 420-421) by checking that the issuer1Template and issuer2Template length is the same at the very beginning of the function.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -390,14 +389,62 @@
...
             if (validIssuerTemplate.Contains(TenantIdTemplate))
             {
-                return validIssuerTemplate.Replace(TenantIdTemplate, tenantId) == actualIssuer;
+                return IssuersWithTemplatesAreEqual(validIssuerTemplate, TenantIdTemplate, actualIssuer, tenantId);
             }
             else
             {
                 return validIssuerTemplate == actualIssuer;
             }
         }
```
victorm-hernandez: >firstHalfIssuer2EndIndex [](http://example.com/codeflow?start=16&length=24)

There is no need for this variable since it is only used when compared with the firstHalfIssuer1EndIndex. You may directly include the IndexOf as part of the if statement below and just keep a single variable. 

After this check both firstHalfIssuer1EndIndex and firstHalfIssuer2EndIndex will have the same value.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -390,14 +389,62 @@
...
+        {
+            int firstHalfIssuer1EndIndex = issuer1.IndexOf(issuer1Template, StringComparison.Ordinal);
+            int firstHalfIssuer2EndIndex = issuer2.IndexOf(issuer2Template, StringComparison.Ordinal);
+
+            // ensure indices exist for both issuers
+            if (firstHalfIssuer1EndIndex == -1  || firstHalfIssuer2EndIndex == -1)
+                return false;
+
+            // if the indices aren't equal, then the issuers aren't equal
+            if (firstHalfIssuer1EndIndex != firstHalfIssuer2EndIndex)
```
victorm-hernandez: >secondHalfIssuer2StartIndex [](http://example.com/codeflow?start=16&length=27)

ditto: Similar to the first check these 2 variables should hold the same value, there is no need to instantiate the second variable, you can use the if statement to compare the operation result with the first variable.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -390,14 +389,62 @@ private static bool IsValidIssuer(string validIssuerTemplate, string tenantId, s
...
+            if (!StringSegmentsAreEqual(issuer1, secondHalfIssuer1StartIndex, issuer1.Length, issuer2, secondHalfIssuer2StartIndex, issuer2.Length))
+                return false;
+
+            return true;
+        }
+
+        /// <summary>
+        /// Compare two string segments for equality without string allocations.
+        /// </summary>
+        private static bool StringSegmentsAreEqual(string str1, int str1Start, int str1End, string str2, int str2Start, int str2End)
```
keegan-caruso: on netcore we should be able to use span based comparisons

https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.compareto?view=net-8.0

and we can keep the ordinal comparisons

keegan-caruso:
I'd really like to see a measurement that shows allocations

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -390,14 +389,62 @@ private static bool IsValidIssuer(string validIssuerTemplate, string tenantId, s
 
             if (validIssuerTemplate.Contains(TenantIdTemplate))
             {
-                return validIssuerTemplate.Replace(TenantIdTemplate, tenantId) == actualIssuer;
+                return IssuersWithTemplatesAreEqual(validIssuerTemplate, TenantIdTemplate, actualIssuer, tenantId);
```
eerhardt: Instead of `Contains`, if we use `IndexOf`, we can find the place where `TenantIdTemplate` exists in the string in the first pass and don't need to search for it again in `IssuersWithTemplatesAreEqual`.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
kllysng: Offline discussion - plan to retain same value

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
eerhardt: We can use the Span based comparisons on non-netcore as well.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
kllysng: Great, done

update - sorry I misread your comment. The template lengths are not expected to be the same.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
kllysng: Great, done

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
kllysng: Thanks, done

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
kllysng: Thanks!

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
kllysng: Thanks, using spans now :)

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
kllysng: Great point, done

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -388,16 +387,46 @@ private static bool IsValidIssuer(string validIssuerTemplate, string tenantId, s
...
+        ///     tenantId = "12345678"
+        /// </summary>
+        internal static bool IssuersWithTemplatesAreEqual(ReadOnlySpan<char> issuer1, ReadOnlySpan<char> issuer1Template, int templateStartIndex, ReadOnlySpan<char> issuer2, ReadOnlySpan<char> tenantId)
+        {
+            var issuer2TemplateStartIndex = issuer2.IndexOf(tenantId);
+            if (templateStartIndex == -1 || issuer2TemplateStartIndex == -1 || templateStartIndex != issuer2TemplateStartIndex)
+                return false;
+
+            // ensure the first part of the issuer1 matches the first part of issuer2
+            if (issuer1.Slice(0, templateStartIndex).SequenceCompareTo(issuer2.Slice(0, templateStartIndex)) != 0)
```
keegan-caruso: Why not SequenceEquals with a StringComparer? (Honest question, it might not matter much)

kllysng: Ah yes, we don't need to know their relative order, so SequenceEquals is better. 😄 Changing to SequenceEquals

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -388,16 +387,46 @@ private static bool IsValidIssuer(string validIssuerTemplate, string tenantId, s
...
+        /// This function replaces: issuer1.Replace(issuer1Template, tenantId) == issuer2
+        /// Example:
+        ///     issuer1 = "https://login.microsoftonline.com/{tenantid}/v2.0"
+        ///     issuer1Template = "{tenantid}"
+        ///     issuer2 = "https://login.microsoftonline.com/12345678/v2.0"
+        ///     tenantId = "12345678"
+        /// </summary>
+        internal static bool IssuersWithTemplatesAreEqual(ReadOnlySpan<char> issuer1, ReadOnlySpan<char> issuer1Template, int templateStartIndex, ReadOnlySpan<char> issuer2, ReadOnlySpan<char> tenantId)
+        {
+            var issuer2TemplateStartIndex = issuer2.IndexOf(tenantId);
```
eerhardt: You shouldn't need to IndexOf here. You already know where the tenantId should start. The general algorithm could be:

1. Check the first part matches on both (from `0` to `templateStartIndex`): e.g. `https://login.microsoftonline.com/`
2. After the first part, check that `issuer2` contains the `tenantId` (from `templateStartIndex` to `templateStartIndex + tenantId.Length`): e.g. `12345678`.
3. Check the last part matches on both: e.g. `/v2.0`

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -76,31 +76,37 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
...
                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidIssuerException(LogHelper.FormatInvariant(LogMessages.IDX40005, LogHelper.MarkAsNonPII(tokenIssuer), LogHelper.MarkAsNonPII(effectiveSigningKeyIssuer))));
             }
 
             return true;
         }
 
+        private static string CreateIssuer(string issuer, string tenantIdTemplate, string tenantId, int templateStartIndex)
+        {
+#if NET6_0_OR_GREATER
+            return string.Concat(issuer.AsSpan(0, templateStartIndex), tenantId, issuer.AsSpan(templateStartIndex + tenantIdTemplate.Length, issuer.Length - tenantIdTemplate.Length - templateStartIndex));
```
eerhardt, line 103: Do we have to create a string here just do do the comparison above?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
kllysng: This makes sense, will update now 🤝

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
kllysng: We have to create the string above (checked with team offline) - at least we are creating only one of them now, rather than both.

But now that you bring this up, I think we can delay the creation a bit until we need it, which should help.

jennyf19 (approved):
:shipit: There were no tests to update?

kllysng:
> :shipit: There were no tests to update?

There were lots of E2E style tests that covered this.. I could write some additional unit tests just for added comfort. Could never hurt!

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -388,16 +387,49 @@ private static bool IsValidIssuer(string validIssuerTemplate, string tenantId, s
...
+        /// <summary>
+        /// Compare two Issuers with templates without string allocations.
+        /// This function replaces: issuer1.Replace(issuer1Template, tenantId) == issuer2
+        /// Example:
+        ///     issuer1 = "https://login.microsoftonline.com/{tenantid}/v2.0"
+        ///     issuer1Template = "{tenantid}"
+        ///     issuer2 = "https://login.microsoftonline.com/12345678/v2.0"
+        ///     tenantId = "12345678"
+        /// </summary>
+        internal static bool IssuersWithTemplatesAreEqual(ReadOnlySpan<char> issuer1, ReadOnlySpan<char> issuer1Template, int templateStartIndex, ReadOnlySpan<char> issuer2, ReadOnlySpan<char> tenantId)
```
brentschmaltz: When this is called from IsValidIssuer (above), a check is made if the Issuer value specified by the user is in templated form. If not, a comparison is made between the issuer1 and issuer2 and can return true.

This method returns false when the values are the equal.
Since the method is internal, it can be called from other locations.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -388,16 +387,49 @@ private static bool IsValidIssuer(string validIssuerTemplate, string tenantId, s
...
             {
                 return validIssuerTemplate == actualIssuer;
             }
         }
 
+        /// <summary>
+        /// Compare two Issuers with templates without string allocations.
+        /// This function replaces: issuer1.Replace(issuer1Template, tenantId) == issuer2
+        /// Example:
+        ///     issuer1 = "https://login.microsoftonline.com/{tenantid}/v2.0"
```
brentschmaltz: The naming throws me off a bit.
issuer1 is the value from metadata or TokenValidationParameters, it is actually the expected value, that may contain a value that may be substituted: {tenantid}. Maybe 'expectedIssuer' or 'issuerToMatch' are suggestions.

 issuer2 is the 'iss' value of the JWT. Could just be named 'issuer' or 'jwtIssuer'.

kllysng: Updated the naming, thank you for providing suggestions!

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
kllysng: @brentschmaltz would it be preferred to change `IssuersWithTemplatesAreEqual` to a private method?

kllysng: Updated per offline discussion :) Which includes, removing `IssuersWithTemplatesAreEqual` and moving logic to `IsValidIssuer`, making `IsValidIssuer` internal, and calling it from `AadTokenValidationParametersExtension.cs` (instead of calling `IssuersWithTemplatesAreEqual`. Also added tests for `IsValidIssuer`.

jennyf19 (approved):
:shipit:

FuPingFranco (approved):
LGTM!

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -383,18 +382,31 @@ private static string CreateV1Authority(string aadV2Authority, string suffixToRe
...
+                if (!actualIssuerSpan.Slice(indexOfTenantIdTemplate, tenantId.Length).SequenceEqual(tenantId.AsSpan()))
+                    return false;
+
+                // ensure the second halves are equal
+                return validIssuerTemplateSpan.Slice(indexOfTenantIdTemplate + TenantIdTemplate.Length).SequenceEqual(actualIssuerSpan.Slice(indexOfTenantIdTemplate + tenantId.Length));
             }
             else
             {
-                return validIssuerTemplate == actualIssuer;
+                return validIssuerTemplateSpan.SequenceEqual(actualIssuerSpan);
```
keegan-caruso, line 409: By default, string == is an ordinal comparison, can you confirm that SequenceEqual is the same behavior?

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -76,31 +76,36 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
...
-
-                // comparing effectiveSigningKeyIssuer with v2TokenIssuer is required as well because of the following scenario:
+                // comparing effectiveSigningKeyIssuer with v2TokenIssuer is required because of the following scenario:
                 // 1. service trusts /common/v2.0 endpoint 
                 // 2. service receieves a v1 token that has issuer like sts.windows.net
                 // 3. signing key issuers will never match sts.windows.net as v1 endpoint doesn't have issuers attached to keys
                 // v2TokenIssuer is the representation of Token.Issuer (if it was a v2 issuer)
-                if (effectiveSigningKeyIssuer != tokenIssuer && effectiveSigningKeyIssuer != v2TokenIssuer)
+                if (!AadIssuerValidator.IsValidIssuer(signingKeyIssuer, tenantIdFromToken, tokenIssuer)
+                    && !AadIssuerValidator.IsValidIssuer(signingKeyIssuer, tenantIdFromToken, openIdConnectConfiguration.Issuer == null ? string.Empty : openIdConnectConfiguration.Issuer))
```
keegan-caruso: This will read better if IsValidIssuer handles null as well as string.Empty. We can avoid the ternary then.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -383,18 +382,31 @@ private static string CreateV1Authority(string aadV2Authority, string suffixToRe
...
+            if (indexOfTenantIdTemplate >= 0 && actualIssuer.Length > indexOfTenantIdTemplate)
             {
-                return validIssuerTemplate.Replace(TenantIdTemplate, tenantId) == actualIssuer;
+                // ensure the first part of the validIssuerTemplate matches the first part of actualIssuer
+                if (!validIssuerTemplateSpan.Slice(0, indexOfTenantIdTemplate).SequenceEqual(actualIssuerSpan.Slice(0, indexOfTenantIdTemplate)))
+                    return false;
+
+                // ensure that actualIssuer contains the tenantId from indexOfTenantIdTemplate for the length of tenantId.Length
+                if (!actualIssuerSpan.Slice(indexOfTenantIdTemplate, tenantId.Length).SequenceEqual(tenantId.AsSpan()))
+                    return false;
```
keegan-caruso, line 402: Looks like we are missing a test case for this path

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -76,31 +76,36 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
...
 #endif
-
-                // comparing effectiveSigningKeyIssuer with v2TokenIssuer is required as well because of the following scenario:
+                // comparing effectiveSigningKeyIssuer with v2TokenIssuer is required because of the following scenario:
                 // 1. service trusts /common/v2.0 endpoint 
                 // 2. service receieves a v1 token that has issuer like sts.windows.net
                 // 3. signing key issuers will never match sts.windows.net as v1 endpoint doesn't have issuers attached to keys
                 // v2TokenIssuer is the representation of Token.Issuer (if it was a v2 issuer)
-                if (effectiveSigningKeyIssuer != tokenIssuer && effectiveSigningKeyIssuer != v2TokenIssuer)
+                if (!AadIssuerValidator.IsValidIssuer(signingKeyIssuer, tenantIdFromToken, tokenIssuer)
```
keegan-caruso, line 88: Missing test branch coverage, can you take a look?

keegan-caruso (changes requested):
Lets at least get the missing test case. Pls evaluate the rest of the comments.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
kllysng: According to the documentation [here](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/MemoryExtensions.cs#L2361C13-L2361C126), yes looks like the comparison is ordinal.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
kllysng: Good point - would make it cleaner. Done

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
kllysng: Thank you, done 🙏

kllysng: image

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
kllysng: image

These lines are covered by [these test cases](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/12cd3dcdea353a45ae05568974dab2a17b53e88a/test/Microsoft.IdentityModel.Validators.Tests/AadSigningKeyIssuerValidatorTests.cs#L256C17-L278C20)

keegan-caruso (approved):

PR #2598, opened 2024-05-15
Perform AOT check on repository

SaurabhMSFT:
# Perform AOT check on repository

Summary of the changes
Added a script and yml file to perform AOT check on repository for every pull request. This script will notify us if a non AOT compatible change is added to the repo.

### DIFF
--- a/.github/workflows/aot-check.yml
+++ b/.github/workflows/aot-check.yml
@@ -1,0 +1,32 @@
+name: "AOT Check"
+
+on:
+  push:
+    branches: [ "dev", "dev6x" ]
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+    branches: [ "dev", "dev6x" ]
+
+jobs:
+  analyze:
+    runs-on: windows-latest
+    name: Wilson GitHub AOT check
+
+    strategy:
+      matrix:
+        language: [ 'csharp' ]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 2
+
+    - name: Runs powershell script
+      id: aot-powershell
+      run: build\test-aot.ps1
+
--- a/.github/workflows/aot-check.yml
+++ b/.github/workflows/aot-check.yml
@@ -2,29 +2,25 @@ name: "AOT Check"
 
 on:
   push:
-    branches: [ "dev", "dev6x" ]
+    branches: [ "dev" ]
   pull_request:
     types:
       - opened
       - synchronize
       - reopened
       - ready_for_review
-    branches: [ "dev", "dev6x" ]
+    branches: [ "dev" ]
 
 jobs:
   analyze:
     runs-on: windows-latest
     name: Wilson GitHub AOT check
 
-    strategy:
-      matrix:
-        language: [ 'csharp' ]
-
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
       with:
-        fetch-depth: 2
+        fetch-depth: 1
 
     - name: Runs powershell script
       id: aot-powershell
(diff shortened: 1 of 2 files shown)


.github/workflows/aot-check.yml
```diff
@@ -0,0 +1,32 @@
...
+    branches: [ "dev", "dev6x" ]
+
+jobs:
+  analyze:
+    runs-on: windows-latest
+    name: Wilson GitHub AOT check
+
+    strategy:
+      matrix:
+        language: [ 'csharp' ]
```
eerhardt: what's this for? Is it used?

.github/workflows/aot-check.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 2
```
eerhardt: Why 2?

.github/workflows/aot-check.yml
```diff
@@ -0,0 +1,32 @@
+name: "AOT Check"
+
+on:
+  push:
+    branches: [ "dev", "dev6x" ]
```
eerhardt: Is the script going to be checked into `dev6x`? I didn't think that version was AOT compatible. I think we did all the work in v7.

build/test-aot.ps1
```diff
@@ -0,0 +1,57 @@
...
+    Write-Host "Actual warning count:", actualWarningCount, "is not as expected. Expected warning count is:", $expectedWarningCount
+}
+
+Exit $testPassed
+
+
+
+
+
+
```
eerhardt:
```suggestion
```

build/test-aot.ps1
```diff
@@ -0,0 +1,57 @@
...
+Write-Host "Actual warning count is: ", $actualWarningCount
+$expectedWarningCount = 0
+
+if ($LastExitCode -ne 0)
+{
+    Write-Host "There was an error while publishing AotCompatibility Test App. LastExitCode is:", $LastExitCode
+    Write-Host $publishOutput
+}
+
+#$runtime = if ($IsWindows) {  "win-x64" } elseif ($IsMacOS) { "macos-x64"} else {"linux-x64"}
```
eerhardt: Can we keep the script able to run on all 3 OSes? That way devs can run the script locally without having to move to a specific OS.

.github/workflows/aot-check.yml
jennyf19: Yes, that's correct, good point @eerhardt

build/test-aot.ps1
SaurabhMSFT: I have restored the checks.

build/test-aot.ps1
SaurabhMSFT: Done

.github/workflows/aot-check.yml
SaurabhMSFT: I have removed the dev6x branch.

.github/workflows/aot-check.yml
SaurabhMSFT: I have changed the fetch depth to 1 so it fetches the top commit only.

.github/workflows/aot-check.yml
SaurabhMSFT: This was copied from another yml file and was unused. I have removed it.

eerhardt (approved):
LGTM. Thanks!

We can also delete https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/tree/dev/test/Microsoft.IdentityModel.AotCompatibility.Tests now since this replaces those tests. We can do that in this PR or in a follow up.

dualtagh (approved):

PR #2621, opened 2024-05-31
Update to latest net9 preview

### LINKED ISSUE #2526: [Feature Request] M.IM conditionally targets .NET9-preview
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2343

Make sure the .NET9-preview is added to the GitHub action and the right builds. We should also have a build variable for .NET 9

start with 9.0.100-preview.2.24157.14
https://dotnet.microsoft.com/en-us/download/dotnet/9.0 and we'll update it every month. Using a * does not work well in this case.

An even better condition to add the new target framework is to modify `<TargetFrameworks>` conditionally (adding the line in green below)

```diff
<TargetFrameworks>net6.0; net7.0; net8.0; net462; net472; netstandard2.0</TargetFrameworks>
+ <TargetFrameworks Condition="'$(TargetNet9)'== 'True'">$(TargetFrameworks); net9.0</TargetFrameworks>
```

FuPingFranco:
## Description

Target latest NET9 preview-4.

Test Passing Builds:

- TargetNet9 = True (https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1309327)
- TargetNet9 = False (https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1309326)

Fixes #2526

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
@@ -198,7 +198,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// <inheritdoc/>
         public override bool Sign(ReadOnlySpan<byte> input, Span<byte> signature, out int bytesWritten)
         {
-            if (input == null || input.Length == 0)
+            if (input.Length == 0)
                 throw LogHelper.LogArgumentNullException(nameof(input));
 
             if (_disposed)
--- a/src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
@@ -207,7 +207,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// <inheritdoc/>
         public override bool Sign(ReadOnlySpan<byte> input, Span<byte> signature, out int bytesWritten)
         {
-            if (input == null || input.Length == 0)
+            if (input.Length == 0)
                 throw LogHelper.LogArgumentNullException(nameof(input));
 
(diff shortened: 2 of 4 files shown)


src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -198,7 +198,7 @@ internal bool ValidKeySize()
         /// <inheritdoc/>
         public override bool Sign(ReadOnlySpan<byte> input, Span<byte> signature, out int bytesWritten)
         {
-            if (input == null || input.Length == 0)
```
FuPingFranco, line 201: Comparing a span to 'null' might be redundant, 'null' literal will be implicitly converted to a 'Span<T>.Empty' (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2265

src/Microsoft.IdentityModel.Tokens/SymmetricSignatureProvider.cs
```diff
@@ -207,7 +207,7 @@ public override byte[] Sign(byte[] input)
         /// <inheritdoc/>
         public override bool Sign(ReadOnlySpan<byte> input, Span<byte> signature, out int bytesWritten)
         {
-            if (input == null || input.Length == 0)
```
FuPingFranco, line 210: Same as above

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
jennyf19: interesting. thanks.

jennyf19 (approved):
:shipit:

PR #2622, opened 2024-06-02
Don't always test string claims for Date Time

### LINKED ISSUE #2615: [Bug] Evaluate usage of DateTime.TryParse for performance
latest version.

**Where is the issue?**
 * [x] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [x] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [ ] S.IM.Tokens.Jwt
 * Other (please describe)

**Repro**

Looking at a CPU profile of validating JWT tokens, between 5-8% of the time is spent in 2 calls to `DateTime.TryParse`:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/753959a8af318497ecf60c466a23f7c972f310a0/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs#L713

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/753959a8af318497ecf60c466a23f7c972f310a0/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs#L602-L604

For every string claim in the JWT, we are calling `DateTime.TryParse` on each value, whether it is expected to be a DateTime or not. This is expensive. We should consider if can eliminate these calls, or only call them if necessary.

image

image

keegan-caruso:
# Don't always test string claims for Date Time

## Description

Cannot remove APIs in in these internal methods as they are friends

There are serveral claim types that should never
be a datetime value. This changes introduces
a list of claim types that will not be tested if they are a Date Time format. 
This list of claim types is not exhaustive.

Added test data for a test token with more claims
and a benchmark using this data.

Benchmark results:

BenchmarkDotNet v0.13.12, Windows 11 (10.0.22631.3593/23H2/2023Update/SunValley3) (Hyper-V)
Intel Xeon Platinum 8370C CPU 2.80GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK 8.0.301
  [Host]    : .NET 8.0.6 (8.0.624.26715), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  MediumRun : .NET 8.0.6 (8.0.624.26715), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI

LaunchCount=4  WarmupCount=10

| Method                                                                | Mean     | Error    | StdDev   | P90      | P95      | P100     | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|---------------------------------------------------------------------- |---------:|---------:|---------:|---------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
| JsonWebTokenHandler_ValidateTokenAsync_CreateClaims                   | 45.71 us | 0.565 us | 1.227 us | 47.23 us | 47.35 us | 47.46 us |  1.10 |    0.03 | 0.6104 |  17.63 KB |        1.04 |
| JsonWebTokenHandler_ValidateTokenAsync_CreateClaims_WithOptimizations | 41.50 us | 0.195 us | 0.416 us | 42.06 us | 42.14 us | 42.72 us |  1.00 |    0.00 | 0.6104 |  16.88 KB |        1.00 |

Fixes #2615

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -604,10 +604,10 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         // breaking compatibility.
         internal static string GetStringClaimValueType(string str)
         {
-            return GetStringClaimValueType(string.Empty, str);
+            return GetStringClaimValueType(str, string.Empty);
         }
 
-        internal static string GetStringClaimValueType(string claimType, string str)
+        internal static string GetStringClaimValueType(string str, string claimType)
         {
             if (!string.IsNullOrEmpty(claimType) && !JsonSerializerPrimitives.TryAllStringClaimsAsDateTime() && JsonSerializerPrimitives.IsKnownToNotBeDateTime(claimType))
                 return ClaimValueTypes.String;
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -12,6 +12,7 @@ using System.Text.RegularExpressions;
 using Microsoft.IdentityModel.Abstractions;
 using Microsoft.IdentityModel.Logging;
 using Microsoft.IdentityModel.Tokens;
+using Microsoft.IdentityModel.Tokens.Json;
 
 using TokenLogMessages = Microsoft.IdentityModel.Tokens.LogMessages;
 
@@ -599,8 +600,18 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         }
 
         // If a string is in IS8061 format, assume a DateTime is in UTC
+        // Because this is a friend class, we can't remove this method without
+        // breaking compatibility.
         internal static string GetStringClaimValueType(string str)
         {
+            return GetStringClaimValueType(string.Empty, str);
+        }
+
+        internal static string GetStringClaimValueType(string claimType, string str)
+        {
+            if (!string.IsNullOrEmpty(claimType) && !JsonSerializerPrimitives.TryAllStringClaimsAsDateTime() && JsonSerializerPrimitives.KnownNonDateTimesClaimTypes.Contains(claimType))
+                return ClaimValueTypes.String;
+
             if (DateTime.TryParse(str, out DateTime dateTimeValue))
             {
                 string dtUniversal = dateTimeValue.ToUniversalTime().ToString("o", CultureInfo.InvariantCulture);
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -609,7 +609,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
         internal static string GetStringClaimValueType(string claimType, string str)
         {
-            if (!string.IsNullOrEmpty(claimType) && !JsonSerializerPrimitives.TryAllStringClaimsAsDateTime() && JsonSerializerPrimitives.KnownNonDateTimesClaimTypes.Contains(claimType))
+            if (!string.IsNullOrEmpty(claimType) && !JsonSerializerPrimitives.TryAllStringClaimsAsDateTime() && JsonSerializerPrimitives.IsKnownToNotBeDateTime(claimType))
                 return ClaimValueTypes.String;
 
             if (DateTime.TryParse(str, out DateTime dateTimeValue))
--- a/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
@@ -146,7 +146,7 @@ namespace Microsoft.IdentityModel.Tokens.Json
 
             if (jsonElement.ValueKind == JsonValueKind.String)
             {
-                if (!string.IsNullOrEmpty(claimType) && !TryAllStringClaimsAsDateTime() && KnownNonDateTimesClaimTypes.Contains(claimType))
+                if (!string.IsNullOrEmpty(claimType) && !TryAllStringClaimsAsDateTime() && IsKnownToNotBeDateTime(claimType))
                     return jsonElement.GetString();
 
                 if (DateTime.TryParse(jsonElement.GetString(), CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out DateTime dateTime))
@@ -718,7 +718,7 @@ namespace Microsoft.IdentityModel.Tokens.Json
         /// sourced from expected Entra V1 and V2 claims, OpenID Connect claims, and a selection of
         /// restricted claim names.
         /// </summary>
-        public static HashSet<string> KnownNonDateTimesClaimTypes = new(StringComparer.Ordinal)
+        private static HashSet<string> KnownNonDateTimesClaimTypes = new(StringComparer.Ordinal)
         {
             "acr",
             "acrs",
@@ -798,6 +798,17 @@ namespace Microsoft.IdentityModel.Tokens.Json
             "ztdid"
         };
 
+        internal static bool IsKnownToNotBeDateTime(string claimType)
+        {
+            if (string.IsNullOrEmpty(claimType))
+                return true;
+
+            if (KnownNonDateTimesClaimTypes.Contains(claimType))
+                return true;
+
+            return false;
+        }
+
         internal static object ReadStringAsObject(ref Utf8JsonReader reader, string propertyName, string className, bool read = false)
         {
             // returning null keeps the same logic as JsonSerialization.ReadObject
@@ -810,7 +821,7 @@ namespace Microsoft.IdentityModel.Tokens.Json
 
             string originalString = reader.GetString();
 
-            if (!TryAllStringClaimsAsDateTime() && KnownNonDateTimesClaimTypes.Contains(propertyName))
+            if (!TryAllStringClaimsAsDateTime() && IsKnownToNotBeDateTime(propertyName))
             {
                 reader.Read();
                 return originalString;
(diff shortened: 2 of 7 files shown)


src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -695,6 +706,98 @@ internal static string ReadStringOrNumberAsString(ref Utf8JsonReader reader, str
...
+        {
+            return (AppContext.TryGetSwitch(TryToCreateDateTimeClaimsSwitch, out bool tryAsDateTime) && tryAsDateTime);
+        }
+
+        /// <summary>
+        /// This is a non-exhaustive list of claim types that are not expected to be DateTime values
+        /// sourced from expected Entra V1 and V2 claims, OpenID Connect claims, and a selection of
+        /// restricted claim names.
+        /// </summary>
+        public static HashSet<string> KnownNonDateTimesClaimTypes = new(StringComparer.Ordinal)
```
keegan-caruso: There may be a better type to put this on, open to suggestions

brentschmaltz: I think this can be internal.
The switch will handle backcompat.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -695,6 +706,98 @@ internal static string ReadStringOrNumberAsString(ref Utf8JsonReader reader, str
             return retVal;
         }
 
+        internal const string TryToCreateDateTimeClaimsSwitch = "Switch.Microsoft.IdentityModel.TryAllStringClaimsAsDateTime";
+
+        public static bool TryAllStringClaimsAsDateTime()
+        {
+            return (AppContext.TryGetSwitch(TryToCreateDateTimeClaimsSwitch, out bool tryAsDateTime) && tryAsDateTime);
```
brentschmaltz, line 713: what is the default?

keegan-caruso: Default is the new behavior

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
keegan-caruso: I'll put this behind an internal method, no need for the Collection to be exposed

JoshLozensky: are these values not already available as constants somewhere? do we really need to make all of these strings?

JoshLozensky (approved):

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
brentschmaltz: I take it that means that by default we will not attempt to convert and of the claims in KnownNonDateClaimTypes.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -695,6 +706,109 @@ internal static string ReadStringOrNumberAsString(ref Utf8JsonReader reader, str
...
+        }
+
+        /// <summary>
+        /// This is a non-exhaustive list of claim types that are not expected to be DateTime values
+        /// sourced from expected Entra V1 and V2 claims, OpenID Connect claims, and a selection of
+        /// restricted claim names.
+        /// </summary>
+        private static HashSet<string> KnownNonDateTimesClaimTypes = new(StringComparer.Ordinal)
```
brentschmaltz: We may want to think about header claims.
[JWT:](https://datatracker.ietf.org/doc/html/rfc7519#section-5) , cty, typ, [JWS](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1) alg, crit, jku, jwk, kid, x5u, x5c, x5t, x5t#256, typ  [JWE](https://datatracker.ietf.org/doc/html/rfc7516#section-4.1) enc, zip

Question is do we lump payload and header in the same switch?
I think it's probably OK to have one list, we can always expose this later and give users complete control.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
keegan-caruso: Yes exactly.

brentschmaltz (changes requested):
Asking to add the additional header claim values.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
keegan-caruso: added

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
keegan-caruso: These are all from external sources, I think it is more important to be able to read the values exactly in this case vs having them in constants elsewhere. I'd agree if they got uses elsewhere.

brentschmaltz (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -599,8 +600,18 @@ internal static IEnumerable<SecurityKey> ConcatSigningKeys(TokenValidationParame
...
 
         // If a string is in IS8061 format, assume a DateTime is in UTC
+        // Because this is a friend class, we can't remove this method without
+        // breaking compatibility.
         internal static string GetStringClaimValueType(string str)
         {
+            return GetStringClaimValueType(string.Empty, str);
+        }
+
+        internal static string GetStringClaimValueType(string claimType, string str)
```
pmaytak: Revert the order to match the above overload.
```suggestion
        internal static string GetStringClaimValueType(string str, string claimType)
```

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -129,6 +129,14 @@ public static JsonElement CreateJsonElement(string json)
...
 
         internal static object CreateObjectFromJsonElement(JsonElement jsonElement, int currentDepth)
+        {
+            return CreateObjectFromJsonElement(string.Empty, jsonElement, currentDepth);
+        }
+
+        /// <remarks>
+        /// <paramref name="claimType"/> is not considered on recursive calls.
+        /// </remarks>
+        internal static object CreateObjectFromJsonElement(string claimType, JsonElement jsonElement, int currentDepth)
```
pmaytak: Same comment. Change the param order to match above.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -695,6 +706,122 @@ internal static string ReadStringOrNumberAsString(ref Utf8JsonReader reader, str
...
+        {
+            return (AppContext.TryGetSwitch(TryToCreateDateTimeClaimsSwitch, out bool tryAsDateTime) && tryAsDateTime);
+        }
+
+        /// <summary>
+        /// This is a non-exhaustive list of claim types that are not expected to be DateTime values
+        /// sourced from expected Entra V1 and V2 claims, OpenID Connect claims, and a selection of
+        /// restricted claim names.
+        /// </summary>
+        private static HashSet<string> KnownNonDateTimesClaimTypes = new(StringComparer.Ordinal)
```
pmaytak: nit:
```suggestion
        private static HashSet<string> KnownNonDateTimeClaimTypes = new(StringComparer.Ordinal)
```

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
```diff
@@ -664,15 +664,15 @@ internal void AddDictionaryClaims(IDictionary<string, object> claimsCollection)
                 this[kvp.Key] = kvp.Value;
         }
 
-        internal static string GetClaimValueType(object value)
+        private static string GetClaimValueType(string claimType, object value)
```
pmaytak, line 667: Safe to make private?

pmaytak, line 667: Is there a reason for adding new params to the begging? I think per guidelines, they're usually added to the end.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -695,6 +706,122 @@ internal static string ReadStringOrNumberAsString(ref Utf8JsonReader reader, str
             return retVal;
         }
 
+        internal const string TryToCreateDateTimeClaimsSwitch = "Switch.Microsoft.IdentityModel.TryAllStringClaimsAsDateTime";
```
pmaytak, line 709: Do we really need a switch for this? Or is this more of a safety measure to disable quickly if there's some unintended behaivor?

keegan-caruso: exactly. It is unlikely, but we wanted an opt out.

src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs
keegan-caruso: This one yes, it wasn't used outside of this method.

keegan-caruso: Closer to the signature of the other methods.

pmaytak (approved):

PR #2625, opened 2024-06-10
Use RSASignaturePadding.Pss by default

### LINKED ISSUE #2626: [CodeQL] Wilson should no longer use RSASignaturePadding.Pkcs1 for RSA signing by default
RSASignaturePadding.Pkcs1 is no longer an approved padding scheme for RSA signing.

Possible Solution:
Introduced an AppContext switch where RSASignaturePadding.Pkcs1 is not used by default unless the switch is enabled.

sruke:
**What?**
- Introduced an AppContext switch: `Switch.Microsoft.IdentityModel.EnablePkcs1SignaturePaddingSwitch`
- Change in behavior: `RSASignaturePadding.Pkcs1` is not used by default unless the above switch is enabled.

**How to Enable the AppContext Switch?**
Set the following in your code. .NET runtime will pick it up and Wilson will use it:
```csharp
AppContext.SetSwitch("Switch.Microsoft.IdentityModel.EnablePkcs1SignaturePaddingSwitch", true);
```

**Why?**
`RSASignaturePadding.Pkcs1` is no longer an approved padding scheme for RSA signing.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
@@ -41,6 +41,8 @@ namespace Microsoft.IdentityModel.Tokens
         private VerifyDelegate _verifyFunction = VerifyNotFound;
         private VerifyUsingOffsetDelegate _verifyUsingOffsetFunction = VerifyUsingOffsetNotFound;
 
+        internal const string _enablePkcs1SignaturePaddingSwitch = "Switch.Microsoft.IdentityModel.EnablePkcs1SignaturePaddingSwitch";
+
         // Encryption algorithms do not need a HashAlgorithm, this is called by RSAKeyWrap
         internal AsymmetricAdapter(SecurityKey key, string algorithm, bool requirePrivateKey)
             : this(key, algorithm, null, requirePrivateKey)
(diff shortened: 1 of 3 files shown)


westin-m (approved):
Looks good. Thanks, @sruke!

test/Microsoft.IdentityModel.Tokens.Tests/AsymmetricSignatureTests.cs
```diff
@@ -7,12 +7,20 @@
...
 #pragma warning disable SYSLIB0027 // Type or member is obsolete
 
 namespace Microsoft.IdentityModel.Tokens.Tests
 {
+    /// <summary>
+    /// This class defines the test collection for all tests that set an AppContext switch.
+    /// All tests in this collection will run sequentially to prevent unexpected behavior
+    /// due to interference between tests as multiple tests set the switch.
+    /// </summary>
+    [CollectionDefinition(nameof(AppContextTestCollection), DisableParallelization = true)]
```
FuPingFranco, line 20: Unit tests are failing on PR build and GitHub actions could be related to this. Can consider writing tests on a separate class instead.

test/Microsoft.IdentityModel.Tokens.Tests/AsymmetricSignatureTests.cs
```diff
@@ -51,6 +59,49 @@ public void UnsupportedRSATypes()
...
+            try
+            {
+                AppContext.SetSwitch("Switch.Microsoft.IdentityModel.EnablePkcs1SignaturePaddingSwitch", theoryData.EnablePkcs1SignaturePaddingSwitch);
+                var asymmetricAdapter = new AsymmetricAdapter(new RsaSecurityKey(new DerivedRsa(2048)), theoryData.Algorithm, false);
+            }
+            catch(Exception ex)
+            {
+                theoryData.ExpectedException.ProcessException(ex, context);
+            }
+
```
FuPingFranco, line 77: Could consider re-setting the state back to default before asserting, funny things happen when using context switches

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -195,8 +197,14 @@ private void InitializeUsingRsa(RSA rsa, string algorithm)
             }
             else
             {
-                // default RSASignaturePadding for other supported RSA algorithms is Pkcs1
-                RSASignaturePadding = RSASignaturePadding.Pkcs1;
+                if (AppContext.TryGetSwitch(_enablePkcs1SignaturePaddingSwitch, out bool enablePaddingSwitch) && enablePaddingSwitch)
```
pmaytak, line 200: I think it would be good to extract this into a separate helper method to simplify since it's always these two steps and since we always call this line when checking switches. Something like `bool SwitchEnabled(string switchName) => return bool`

jennyf19:
closing for now

PR #2628, opened 2024-06-11
Add framework for reducing exceptions and logging.

brentschmaltz:
This PR introduces an internal set of building blocks that will be used for capturing SecurityToken validation results in a consistent and deterministic manner.

We introduced the TokenValidationResult to avoid throwing back to the calling application on errors.
This work is an extension where each step captures and attaches the result to the TokenValidationResult.
All logs and exceptions are handled by upper layers.

This will be used to reduce throwing exceptions and logging.

When multiple security policies are associated with a single endpoint, for a successful validation, error logs will not be automatically written.

Latency is reduced as Exceptions are not thrown when one policy succeeds.


src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -38,7 +38,7 @@ public class TokenValidationParameters
         /// Default for the maximum token size.
         /// </summary>
         /// <remarks>250 KB (kilobytes).</remarks>
-        public const Int32 DefaultMaximumTokenSizeInBytes = 1024 * 250;
+        public const int DefaultMaximumTokenSizeInBytes = 1024 * 250;
```
jmprieur, line 41: Is it a breaking change?

jmprieur (commented):
This is really cool, @brentschmaltz

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
brentschmaltz: @jmprieur Good question.
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/built-in-types

I believe they are the same, similar to String and string.

jennyf19 (approved):
Looks really well laid out. Thanks @brentschmaltz

src/Microsoft.IdentityModel.Tokens/Validation/MessageDetail.cs
```diff
@@ -0,0 +1,46 @@
...
+            get
+            {
+                _message ??= LogHelper.FormatInvariant(FormatString.ToString(), Parameters);
+                return _message;
+            }
+        }
+
+        private ReadOnlyMemory<char> FormatString { get; }
+
+        private object[] Parameters { get; }
```
keegan-caruso, line 44: This will always box value type params, no?

src/Microsoft.IdentityModel.Tokens/Validation/IssuerValidationResult.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using System;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Contains the result of validating a <see cref="SecurityToken"/> issuer.
+    /// The <see cref="TokenValidationResult"/> contains a collection of <see cref="ValidationResult"/> for each step in the token validation.
+    /// </summary>
+    internal class IssuerValidationResult : ValidationResult
```
keegan-caruso, line 12: Consider sealing the internal types. It is a slight perf improvement.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationResult.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                _isValid = value;
+            }
+        }
+
+        // TODO - HasValidOrExceptionWasRead, IsValid, Exception are temporary and will be removed when TokenValidationResult derives from ValidationResult.
+        /// <summary>
+        /// Gets or sets a boolean recording if IsValid or Exception was called.
+        /// </summary>
+        protected bool HasValidOrExceptionWasRead { get; set; }
```
keegan-caruso, line 85: I find the name to be a bit confusing. It reads as "Has Valid" or "Exception read". But every result has "IsValid". The name doesn't make it clear that it is if either has been read.

src/Microsoft.IdentityModel.Tokens/Validation/MessageDetail.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    /// <summary>
+    /// Contains information about a message that is used to generate a message for logging or exceptions.
+    /// </summary>
+    internal class MessageDetail
+    {
+        private string _message;
+
+        // TODO - remove the need to create NonPII objects, we could use tuples <bool, object> where bool == true => object is PII.
+        // TODO - does this need to be ReadOnlyMemory<char>?
```
keegan-caruso, line 17: I believe this can just be a string.

src/Microsoft.IdentityModel.Tokens/Validation/LogDetail.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Contains information so that logs can be written when needed.
+    /// </summary>
+    internal class LogDetail
```
keegan-caruso, line 11: Did we consider LogDetail and MessageDetail as structs? Could avoid the allocations

src/Microsoft.IdentityModel.Tokens/Validation/ValidationResult.cs
iNinja: If we want to be overly clear of the details we could go with something along the lines of `WasEitherIsValidOrExeceptionRead`, but could we simply say something like `ResultHasBeenValidated` or `ResultWasChecked`?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.cs
```diff
@@ -0,0 +1,771 @@
...
+                return new TokenValidationResult
+                {
+                    Exception = ex,
+                    IsValid = false
+                };
+            }
+        }
+
+        /// <summary>
+        ///  Private method for token validation, responsible for:
```
FuPingFranco: nit: "Internal method for token validation..."

FuPingFranco (approved):

PR #2630, opened 2024-06-12
add base methods to tokenhandler

### LINKED ISSUE #1259: Consider adding more base methods to TokenHandler
`TokenHandler` seems pretty useless as a base class, as it doesn't expose any useful method for token generation or validation.

Please consider adding abstract or virtual properties/methods for at least the following APIs:
  - `Type TokenType { get; }`
  - `bool CanReadToken(string token)`
  - `string CreateToken(SecurityTokenDescriptor tokenDescriptor)`
  - `TokenValidationResult ValidateToken(string token, TokenValidationParameters validationParameters)`

westin-m:
Introduce additional base methods to TokenHandler.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
@@ -90,7 +90,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// <param name="tokenDescriptor">The <see cref="SecurityTokenDescriptor"/> used to create the token.</param>
         /// <returns>A <see cref="string"/>.</returns>
         /// <exception cref="NotImplementedException"></exception>
-        public virtual string CreateSecuritTokenAsString(SecurityTokenDescriptor tokenDescriptor) => throw new NotImplementedException();
+        public virtual string CreateSecurityTokenAsString(SecurityTokenDescriptor tokenDescriptor) => throw new NotImplementedException();
 
         /// <summary>
         /// Creates a <see cref="SecurityToken"/> as described by the provided <see cref="SecurityTokenDescriptor"/>.
--- a/src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
@@ -5,6 +5,7 @@ using Microsoft.IdentityModel.Logging;
 using System;
 using System.ComponentModel;
 using System.Security.Claims;
+using System.Text.Json;
 using System.Threading.Tasks;
 using static Microsoft.IdentityModel.Logging.LogHelper;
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
```diff
@@ -75,6 +91,21 @@ public int TokenLifetimeInMinutes
...
+        /// <returns>A <see cref="bool"/>.</returns>
+        public virtual bool CanReadToken(string token) => throw new NotImplementedException();
+
+        /// <summary>
+        /// Creates a token as described by the provided <see cref="SecurityTokenDescriptor"/>.
+        /// </summary>
+        /// <param name="tokenDescriptor">The <see cref="SecurityTokenDescriptor"/> used to create the token.</param>
+        /// <returns>A <see cref="string"/>.</returns>
+        /// <exception cref="NotImplementedException"></exception>
+        public virtual string CreateToken(SecurityTokenDescriptor tokenDescriptor) => throw new NotImplementedException();
```
FuPingFranco: I believe this method already exists on the JsonWebTokenHandler and/or SecurityTokenHandler you'll need to override them as this is causing test to fail.

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
```diff
@@ -53,8 +54,23 @@ public int TokenLifetimeInMinutes
             set => _defaultTokenLifetimeInMinutes = (value < 1) ? throw LogExceptionMessage(new ArgumentOutOfRangeException(nameof(value), FormatInvariant(LogMessages.IDX10104, LogHelper.MarkAsNonPII(value)))) : value;
         }
 
+        /// <summary>
+        /// Gets the <see cref="JsonTokenType"/> of the token.
+        /// </summary>
+        public JsonTokenType TokenType { get; }
```
keegan-caruso: I don't follow this one.

brentschmaltz: A TokenHandler is the base class for Saml, Saml2 and JWT tokens.
To call out JsonTokenType, doesn't fit here.

brentschmaltz (changes requested):
TokenType should be abstract.

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
brentschmaltz: We can remove the method from SecurityTokenHander.

brentschmaltz: We should also wait until we create the 7.x branch, then we can remove the '=> throw new NotImplementedException()" as @keegan-caruso suggested.

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
```diff
@@ -75,6 +76,28 @@ public int TokenLifetimeInMinutes
...
+        /// <returns>'True' if the ReadToken method can parse the token string.</returns>
+        public virtual bool CanReadToken(string tokenString) => false;
+
+        /// <summary>
+        /// Creates a <see cref="SecurityToken"/> as a <see cref="string"/> as described by the provided <see cref="SecurityTokenDescriptor"/>.
+        /// </summary>
+        /// <param name="tokenDescriptor">The <see cref="SecurityTokenDescriptor"/> used to create the token.</param>
+        /// <returns>A <see cref="string"/>.</returns>
+        /// <exception cref="NotImplementedException"></exception>
+        public virtual string CreateSecuritTokenAsString(SecurityTokenDescriptor tokenDescriptor) => throw new NotImplementedException();
```
keegan-caruso: typo: CreateSecurityTokenAsString

keegan-caruso:
Why did we remove `TokenType`?

westin-m:
> Why did we remove `TokenType`?

`TokenType` isn't moving from `SecurityTokenHandler`. That would need to happen in 8.x

keegan-caruso:
Pending API decision.

PR #2631, opened 2024-06-13
L2 Cache Config Manager + Interface Changes

kllysng:
Introducing ConfigurationManager and Interfaces to support L2 Cache implementation

[Test Plan]()


src/Microsoft.IdentityModel.Tokens/IConfigurationRetrievalTime.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    // L2 TODO: internal until L2 cache is implemented S2S.
+    internal interface IConfigurationRetrievalTime
```
keegan-caruso: Prefer verbs with a noun suffix when it makes sense, something like: IConfigurationTimeRetriever

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -22,6 +21,7 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
         private DateTimeOffset _syncAfter = DateTimeOffset.MinValue;
         private DateTimeOffset _lastRefresh = DateTimeOffset.MinValue;
         private bool _isFirstRefreshRequest = true;
+        private bool _skipDistributedConfigurationManager = false;
```
brentschmaltz, line 24: What is the logic in skipping?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -30,6 +30,13 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
         private T _currentConfiguration;
         private Exception _fetchMetadataFailure;
         private TimeSpan _bootstrapRefreshInterval = TimeSpan.FromSeconds(1);
+        private static readonly DistributedConfigurationOptions s_distributedConfigurationOptions = new();
+
+        // L2 TODO: internal until L2 cache is implemented S2S.
+        /// <summary>
+        /// TODO inject via ctors
+        /// </summary>
+        internal IDistributedConfigurationManager<T> DistributedConfigurationManager { get; set; }
```
brentschmaltz, line 40: How will the user set this property (after it is public)?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -156,21 +163,54 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
-                        if (_configValidator != null)
+                        T configuration = null;
+                        if (DistributedConfigurationManager != null && !_skipDistributedConfigurationManager)
                         {
-                            ConfigurationValidationResult result = _configValidator.Validate(configuration);
-                            if (!result.Succeeded)
-                                throw LogHelper.LogExceptionMessage(new InvalidConfigurationException(LogHelper.FormatInvariant(LogMessages.IDX20810, result.ErrorMessage)));
+                            // TODO handle try/catch
+                            configuration = await DistributedConfigurationManager.GetConfigurationAsync(MetadataAddress, s_distributedConfigurationOptions, CancellationToken.None).ConfigureAwait(false);
+                            if (_configValidator != null)
```
brentschmaltz: This can be written so that _configValidator logic is in one place.

src/Microsoft.IdentityModel.Tokens/BaseConfiguration.cs
```diff
@@ -10,7 +11,7 @@ namespace Microsoft.IdentityModel.Tokens
     /// <summary>
     ///  Represents a generic metadata configuration which is applicable for both XML and JSON based configurations.
     /// </summary>
-    public abstract class BaseConfiguration
+    public abstract class BaseConfiguration /*: IConfigurationRetrievalTime*/ // L2 TODO: internal until L2 cache is implemented S2S.
```
brentschmaltz: If we release this in 8.x, do we need the interface IConfigurationRetievalTime?

src/Microsoft.IdentityModel.Tokens/IConfigurationRetrievalTime.cs
brentschmaltz: Do we need this interface?

GeoK: It's needed to help figure out the time-to-live (TTL) for the config object since it might come from an L2 cache. ConfigurationManager works with configuration type T (not the base configuration), so config objects made outside the code can also use the interface without inheriting from BaseConfiguration.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
GeoK: This was TBD, also not sure if this is the right way to solve it due to multithreading. The challenge is when the application code or the library calls RequestRefresh. If an L2 cache is set up, we'll refresh the value but retrieve it from the L2 cache. Instead, we should probably get a fresh value from the IdP and update the L2 cache.

jennyf19: @pmaytak  you might want to take a look at this PR as well.

src/Microsoft.IdentityModel.Tokens/BaseConfiguration.cs
kllysng: It just needs to be commented out right now because the interface is internal right now

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
kllysng: I think that this property will be set via the ctors, does that sound right @GeoK?

src/Microsoft.IdentityModel.Tokens/IConfigurationRetrievalTime.cs
kllysng: Updated to `IConfigurationTimeRetriever`, thanks!

src/Microsoft.IdentityModel.Tokens/BaseConfiguration.cs
kellyyangsong: THis is commented out since we're keeping the interface internal right now.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -156,21 +164,66 @@ public async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+                            try
+                            {
+                                configuration = await DistributedConfigurationManager.GetConfigurationAsync(MetadataAddress, s_distributedConfigurationOptions, CancellationToken.None).ConfigureAwait(false);
+                                if (_configValidator != null)
+                                {
+                                    ConfigurationValidationResult result = _configValidator.Validate(configuration);
+                                    if (!result.Succeeded)
+                                    {
+                                        configuration = null;
+                                        LogHelper.LogExceptionMessage(new InvalidConfigurationException(LogHelper.FormatInvariant(LogMessages.IDX20812, result.ErrorMessage)));
```
kllysng, line 179: I updated the log message so that it is different from that of the l1 cache. Is `InvalidConfigurationException` okay or do we want a new custom exception?

src/Microsoft.IdentityModel.Protocols/Configuration/IDistributedConfigurationManager.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="metadataAddress"></param>
+        /// <param name="configuration"></param>
+        /// <param name="distributedConfigurationOptions"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        Task SetConfigurationAsync(string metadataAddress, T configuration, DistributedConfigurationOptions distributedConfigurationOptions, CancellationToken cancellationToken = default);
```
kllysng, line 32: Do we want to implement a timeout for this call in ConfigurationManager?

src/Microsoft.IdentityModel.Protocols/Configuration/IDistributedConfigurationManager.cs
```diff
@@ -0,0 +1,34 @@
...
+    internal interface IDistributedConfigurationManager<T> where T : class
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="metadataAddress"></param>
+        /// <param name="distributedConfigurationOptions"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        Task<T> GetConfigurationAsync(string metadataAddress, DistributedConfigurationOptions distributedConfigurationOptions, CancellationToken cancellationToken = default);
```
kllysng, line 22: Do we want a timeout for this call in ConfigurationManager?

brentschmaltz:
Moving to a different project.

PR #2643, opened 2024-06-18
Remove exceptions thrown during issuer validation. Add validation source.

iNinja:
# Issuer validation: Remove exceptions and add validation source

Removed ArgumentNullException thrown as part of issuer validation, and added source.

## Description
- Replaced instances that threw ArgumentNullException with a return of IssuerValidationResult containing the exception information.
- Addressed TODO around identifying the source of the issuer validation (configuration, validation parameters: valid issuer or among list of valid issuers).

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Tokens/LogMessages.cs
@@ -13,7 +13,7 @@ namespace Microsoft.IdentityModel.Tokens
     {
 #pragma warning disable 1591
         // general
-        // public const string IDX10000 = "IDX10000:";
+        public const string IDX10000 = "IDX10000: The parameter '{0}' cannot be a 'null' or an empty object. ";
 
         // properties, configuration 
         public const string IDX10101 = "IDX10101: MaximumTokenSizeInBytes must be greater than zero. value: '{0}'";
--- a/src/Microsoft.IdentityModel.Tokens/Validation/IssuerValidationResult.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/IssuerValidationResult.cs
@@ -25,7 +25,8 @@ namespace Microsoft.IdentityModel.Tokens
         /// Creates an instance of <see cref="IssuerValidationResult"/>
         /// </summary>
         /// <paramref name="issuer"/> is the issuer that was validated successfully.
-        public IssuerValidationResult(string issuer, ValidationSource source)
+        /// <paramref name="source"/> is the <see cref="ValidationSource"/> indicating how this issuer was validated.
+        public IssuerValidationResult(string issuer, ValidationSource source = ValidationSource.NotValidated)
             : base(ValidationFailureType.ValidationSucceeded)
         {
             Issuer = issuer;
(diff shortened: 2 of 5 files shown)


src/Microsoft.IdentityModel.Tokens/Validation/IssuerValidationResult.cs
```diff
@@ -11,17 +11,26 @@ namespace Microsoft.IdentityModel.Tokens
...
+        }
+
         private Exception _exception;
 
         /// <summary>
         /// Creates an instance of <see cref="IssuerValidationResult"/>
         /// </summary>
         /// <paramref name="issuer"/> is the issuer that was validated successfully.
-        public IssuerValidationResult(string issuer)
+        public IssuerValidationResult(string issuer, ValidationSource source)
```
JoshLozensky: Is this intended to be a breaking change? Since this is a public method we usually overload if we want to alter the args to avoid breaking existing users.

iNinja: I ignored for a moment that this is public and can be implemented by others as a delegate.
Thanks for pointing this out.

iNinja: Added a default value and updated the documentation. Not sure however if we want to add extra values for others implementing their own delegates or we just use this when relying on the internal implementation.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -13,7 +13,7 @@ internal static class LogMessages
     {
 #pragma warning disable 1591
         // general
-        // public const string IDX10000 = "IDX10000:";
+        public const string IDX10000 = "IDX10000: The parameter '{0}' cannot be a 'null' or an empty object. ";
```
jmprieur, line 16: We shouldn't reuse a string that got obsoleted. you'd want to create a new constant (incrementing the numbers)

iNinja: Currently there is a method, `LogArgumentNullException`, that has the following code:

`return LogArgumentException<ArgumentNullException>(EventLevel.Error, argument, "IDX10000: The parameter '{0}' cannot be a 'null' or an empty object. ", argument);`

I figured we are using IDX10000 for this very message, so instead of repeating this pattern, I would add it to `LogMessages.cs` and consume the constant.

brentschmaltz: We took the string out and embedded the code in LogHelper, i think it is OK to bring it back.
As we need to reference a const string.

src/Microsoft.IdentityModel.Tokens/Validation/IssuerValidationResult.cs
```diff
@@ -11,17 +11,27 @@ namespace Microsoft.IdentityModel.Tokens
     /// </summary>
     internal class IssuerValidationResult : ValidationResult
     {
+        internal enum ValidationSource
```
brentschmaltz, line 14: @iNinja  I think this is ok to move forward while everything is internal, but we will need a more general solution as users will need to set their own ValidationSource.

ValidationFailureType has an extensibility model that may shed light on a solution.

brentschmaltz (approved):
We will have to generalize ValidationSource.
We should also consider other validators before we make a choice.

src/Microsoft.IdentityModel.Tokens/Validation/IssuerValidationResult.cs
iNinja: Thanks. I will make a note to adopt a model based on `ValidationFailureType` as we progress.

PR #2644, opened 2024-06-18
Rm SkipAuthenticationTagLengthValidation and UseShortNameForRsaOaepKey App Context Switches

### LINKED ISSUE #2629: Remove app context switches for IdentityModel 8x
Remove all app context switches for IdentityModel 8x. 

The current app context switches _should_ be documented [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/App-Context-Switches-in-IdentityModel). (If you find a switch that isn't documented for 7x or even 6x, please add it to the documentation.)

kllysng:
[Documented App Context Switches](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/App-Context-Switches-in-IdentityModel#identity-model-7x)

[PR](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2569) introducing SkipAuthenticationTagLengthValidation switch

[PR](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2578) introducing UseShortNameForRsaOaepKey switch

Fixes #2629


test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
```diff
@@ -1,66 +0,0 @@
...
-        {
-            AppContext.SetSwitch(X509EncryptingCredentials._useShortNameForRsaOaepKey, useShortNameForRsaOaepKey);
-            var encryptingCredentials = new X509EncryptingCredentials(Default.Certificate);
-            JwtSecurityTokenHandler tokenHandler = new JwtSecurityTokenHandler();
-
-            JwtSecurityToken token = CreateJwtSecurityToken(tokenHandler, encryptingCredentials);
-
-            Assert.Equal(token.Header.Alg, algorithm);
-
-            AppContext.SetSwitch(X509EncryptingCredentials._useShortNameForRsaOaepKey, false);
```
jennyf19, line 27: we don't need the tests? I would think removing the AppContext.SetSwitch would be enough?

kllysng: What about the tests in X509EncryptingCredentialsTests.cs?

JoshLozensky (approved):

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
kllysng: I think it would make sense to change all the occurrences of `SecurityAlgorithms.RsaOaepKeyWrap` in tests to `RsaOAEP`, since the short name is default now.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -136,7 +136,7 @@ internal static class LogMessages
         // public const string IDX10622 = "IDX10622:";
         // public const string IDX10623 = "IDX10623:";
         // public const string IDX10624 = "IDX10624:";
-        public const string IDX10625 = "IDX10625: Failed to verify the authenticationTag length, the actual tag length '{0}' does not match the expected tag length '{1}'. authenticationTag: '{2}', algorithm: '{3}' See: https://aka.ms/IdentityModel/SkipAuthenticationTagLengthValidation";
```
pmaytak, line 139: I think there maybe a benefit of leaving the doc link? Since this is a breaking change, and people may run into this when upgrading to 8.x. In the doc we can add a comment like for the other switch that it was a breaking change and link to migration docs?

FuPingFranco: +1

pmaytak: I guess people can just look at 8.x migration docs too if they run into issues. Up to you :).

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
FuPingFranco: I agree with @kellyyangsong we should change the current tests. I think we can removed these tests but can you write a quick negative test to validate our default is now the short form to check if the long form causes issues as expected. Thanks!

pmaytak: +1 Add a test that asserts the behavior when the long form of the name is used. It would probably work to add just an extra test case / theory data to X509EncryptingCredentialsTests.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
kllysng: Removing the link is what we did for removing the hmac unsafe relax context switch - will take your feedback and add link when we document the breaking change 💯

pmaytak (approved):
Approved with suggestions.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
kllysng: Done :)

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.WithContextSwitches.cs
kllysng: We have multiple test cases here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/a2486af8f2eb5d6e470d0fb6fd1d3c1bf606e49a/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs#L2764C24-L2784C185

jennyf19: As long as we have test coverage and not losing any, there is no concern.

FuPingFranco (approved):
LGTM. Thanks Kelly!

PR #2648, opened 2024-06-18
Fixes index out-of-range error

sruke:
Why?
The issuer signing key validation logic was updated to use AadIssuerValidator.IsValidIssuer(..). In this process, the existing logic was modified to compare the signing key issuer template (https://login.microsoftonline.com/{tenantid}/v2.0) with the templated issuer in the OpenIdConnect configuration (https://login.microsoftonline.com/{tenantid}/v2.0). This led to an index out-of-range error because IsValidIssuer expects the actualIssuer from the token to contain a GUID for tid rather than a {tenantid} template.

How?
Restored logic to generate the effective token issuer by replacing the {tenantid} template with the actual tenant ID from the token. Added an additional check to ensure the actual issuer is long enough to match the validIssuerTemplate with the tenant ID replaced, preventing out-of-range errors.

Added tests to cover both issuer validation and issuer signing key validation.

### DIFF
--- a/src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
+++ b/src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
@@ -76,9 +76,13 @@ namespace Microsoft.IdentityModel.Validators
 #if NET6_0_OR_GREATER
                 if (!string.IsNullOrEmpty(tokenIssuer) && !tokenIssuer.Contains(tenantIdFromToken, StringComparison.Ordinal))
                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidIssuerException(LogHelper.FormatInvariant(LogMessages.IDX40004, LogHelper.MarkAsNonPII(tokenIssuer), LogHelper.MarkAsNonPII(tenantIdFromToken))));
+
+                var v2TokenIssuer = openIdConnectConfiguration.Issuer?.Replace(AadIssuerValidator.TenantIdTemplate, tenantIdFromToken, StringComparison.Ordinal);
 #else
                 if (!string.IsNullOrEmpty(tokenIssuer) && !tokenIssuer.Contains(tenantIdFromToken))
                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidIssuerException(LogHelper.FormatInvariant(LogMessages.IDX40004, LogHelper.MarkAsNonPII(tokenIssuer), LogHelper.MarkAsNonPII(tenantIdFromToken))));
+
+                var v2TokenIssuer = openIdConnectConfiguration.Issuer?.Replace(AadIssuerValidator.TenantIdTemplate, tenantIdFromToken);
 #endif
                 // comparing effectiveSigningKeyIssuer with v2TokenIssuer is required because of the following scenario:
                 // 1. service trusts /common/v2.0 endpoint 
@@ -86,7 +90,7 @@ namespace Microsoft.IdentityModel.Validators
                 // 3. signing key issuers will never match sts.windows.net as v1 endpoint doesn't have issuers attached to keys
                 // v2TokenIssuer is the representation of Token.Issuer (if it was a v2 issuer)
                 if (!AadIssuerValidator.IsValidIssuer(signingKeyIssuer, tenantIdFromToken, tokenIssuer)
-                    && !AadIssuerValidator.IsValidIssuer(signingKeyIssuer, tenantIdFromToken, openIdConnectConfiguration.Issuer))
+                    && !AadIssuerValidator.IsValidIssuer(signingKeyIssuer, tenantIdFromToken, v2TokenIssuer))
                 {
                     int templateStartIndex = signingKeyIssuer.IndexOf(AadIssuerValidator.TenantIdTemplate, StringComparison.Ordinal);
                     string effectiveSigningKeyIssuer = templateStartIndex > -1 ? CreateIssuer(signingKeyIssuer, AadIssuerValidator.TenantIdTemplate, tenantIdFromToken, templateStartIndex) : signingKeyIssuer;
--- a/src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
+++ b/src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
@@ -77,12 +77,10 @@ namespace Microsoft.IdentityModel.Validators
                 if (!string.IsNullOrEmpty(tokenIssuer) && !tokenIssuer.Contains(tenantIdFromToken, StringComparison.Ordinal))
                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidIssuerException(LogHelper.FormatInvariant(LogMessages.IDX40004, LogHelper.MarkAsNonPII(tokenIssuer), LogHelper.MarkAsNonPII(tenantIdFromToken))));
 
-                var v2TokenIssuer = openIdConnectConfiguration.Issuer?.Replace(AadIssuerValidator.TenantIdTemplate, tenantIdFromToken, StringComparison.Ordinal);
 #else
                 if (!string.IsNullOrEmpty(tokenIssuer) && !tokenIssuer.Contains(tenantIdFromToken))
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -76,17 +76,21 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
 #if NET6_0_OR_GREATER
                 if (!string.IsNullOrEmpty(tokenIssuer) && !tokenIssuer.Contains(tenantIdFromToken, StringComparison.Ordinal))
                     throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidIssuerException(LogHelper.FormatInvariant(LogMessages.IDX40004, LogHelper.MarkAsNonPII(tokenIssuer), LogHelper.MarkAsNonPII(tenantIdFromToken))));
+
+                var v2TokenIssuer = openIdConnectConfiguration.Issuer?.Replace(AadIssuerValidator.TenantIdTemplate, tenantIdFromToken, StringComparison.Ordinal);
```
keegan-caruso: Both of these replace calls allocate. Not sure if there is a way to avoid it yet.

test/Microsoft.IdentityModel.Validators.Tests/AadIssuerValidatorTests.cs
```diff
@@ -1,24 +1,127 @@
...
+                    ExpectedResult = false,
+                },
+                new AadIssuerValidatorTheoryData("ValidateIssuer_PPETemplateWithV1Issuer_Failure")
+                {
+                    TemplatedIssuer = ValidatorConstants.AadInstancePPE + "/" + AadIssuerValidator.TenantIdTemplate,
+                    TokenIssuer =  ValidatorConstants.AadInstance + "/" + ValidatorConstants.TenantIdAsGuid,
+                    TenantIdClaim = ValidatorConstants.TenantIdAsGuid,
+                    ExpectedResult = false,
+                },
+                new AadIssuerValidatorTheoryData("ValidateIssuerSigningKey_V2SigningKeyIssuer_V1TokenIssuer_Failure")
```
keegan-caruso, line 102: We should add a test case for 1.1 as well

https://login.microsoftonline.com/common/v1.1/.well-known/openid-configuration

keegan-caruso: (I don't expect any issues)

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
sruke: I don't see a way to avoid it today. Avoiding the allocation can even be an incremental change.

test/Microsoft.IdentityModel.Validators.Tests/AadIssuerValidatorTests.cs
sruke: Good idea!

sruke: Added tests in [this commit](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2648/commits/b2ff02ab4471d64cef6bd1548dcf28c00f5283ac)

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
kllysng: What about: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2648/files#diff-0bf3e7afdfb06195ce65362ac431a7b2c8baadb1cab0745231b78f1a5c6500b2R104-R110?

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -76,17 +76,21 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
...
+                var v2TokenIssuer = openIdConnectConfiguration.Issuer?.Replace(AadIssuerValidator.TenantIdTemplate, tenantIdFromToken);
 #endif
                 // comparing effectiveSigningKeyIssuer with v2TokenIssuer is required because of the following scenario:
                 // 1. service trusts /common/v2.0 endpoint 
                 // 2. service receieves a v1 token that has issuer like sts.windows.net
                 // 3. signing key issuers will never match sts.windows.net as v1 endpoint doesn't have issuers attached to keys
                 // v2TokenIssuer is the representation of Token.Issuer (if it was a v2 issuer)
                 if (!AadIssuerValidator.IsValidIssuer(signingKeyIssuer, tenantIdFromToken, tokenIssuer)
-                    && !AadIssuerValidator.IsValidIssuer(signingKeyIssuer, tenantIdFromToken, openIdConnectConfiguration.Issuer))
+                    && !AadIssuerValidator.IsValidIssuer(signingKeyIssuer, tenantIdFromToken, v2TokenIssuer))
```
keegan-caruso: I think the right fix is to check if both issuer strings contain a template, if they both contain a template, you should be able to compare the full templated string, without allocations.

test/Microsoft.IdentityModel.Validators.Tests/AadIssuerValidatorTests.cs
```diff
@@ -1,24 +1,154 @@
...
             // act
-            var result = AadIssuerValidator.IsValidIssuer(templatedIssuer, ValidatorConstants.TenantIdAsGuid, issuer);
+            var result = AadIssuerValidator.IsValidIssuer(theoryData.TemplatedIssuer, theoryData.TenantIdClaim, theoryData.TokenIssuer);
 
             // assert
-            Assert.Equal(expectedResult, result);
+            Assert.Equal(theoryData.ExpectedResult, result);
+        }
+
+        public static TheoryData<AadIssuerValidatorTheoryData> AadIssuerValidationTestCases()
```
brentschmaltz, line 21: thanks for using theoryData.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
sruke: Added a test in [commit](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2648/commits/17b6cd5ed8452a0408370abbbb9ea856b11f4c1b) to show why I believe the additional allocation is necessary.

Simply comparing the template between signingKeyIssuer and actualIssuer IMO is insufficient, as it does not account for the tid from the token, unlike the original behavior.

sruke: I might be wrong, I'm trying to convince myself that it is indeed okay to remove the allocation.

test/Microsoft.IdentityModel.Validators.Tests/AadIssuerValidatorTests.cs
sruke: Debugging using inline data without a test id to set conditional breakpoints was difficult.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
keegan-caruso: I wrote the logic to compare templated strings and the latest tests pass on the latest commit.

let me try to explain:

signing key issuer: https://login.microsoftonline.com/{tenantid}/v2.0
oidc issuer: https://login.microsoftonline.com/{tenantid}/v2.0

If we replace {tenantid} in oidc issuer:
realOidcIssuer: https://login.microsoftonline.com/f645ad92-e38d-4d1a-b510-d1b09a74a8ca/v2.0 

Then in IsValidIssuer we compare 

https://login.microsoftonline.com/f645ad92-e38d-4d1a-b510-d1b09a74a8ca/v2.0 
against
https://login.microsoftonline.com/{tenantid}/v2.0
f645ad92-e38d-4d1a-b510-d1b09a74a8ca

And verifying what is before the template, after template, and the tid matches.

Aren't we just testing the replace logic to substitute the tid from the token ?

sruke:
Addressed in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2680/files#diff-0bf3e7afdfb06195ce65362ac431a7b2c8baadb1cab0745231b78f1a5c6500b2

PR #2655, opened 2024-06-19
Audience validation: remove exceptions

iNinja:
# Audience validation: Remove exceptions

Removed exceptions thrown as part of audience validation.

## Description
- Implemented new (internal momentarily) method for audience validation which replaces exception throwing with a return of the new AudienceValidationResult containing the exception information.
- Added new delegate signature for audience validation
- Updated ambiguous references in documentation to specify the old version of ValidateAudience.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
@@ -63,6 +63,6 @@ namespace Microsoft.IdentityModel.Tokens
         /// <summary>
         /// Gets the audience that was validated or intended to be validated.
         /// </summary>
-        public string Audience { get; } = "null";
+        public string Audience { get; };
     }
 }
--- a/src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
@@ -1,0 +1,68 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Contains the result of validating the audiences from a <see cref="SecurityToken"/>.
+    /// The <see cref="TokenValidationResult"/> contains a collection of <see cref="ValidationResult"/> for each step in the token validation.
+    /// </summary>
+    internal class AudienceValidationResult : ValidationResult
+    {
+        private Exception _exception;
+
+        /// <summary>
+        /// Creates an instance of <see cref="AudienceValidationResult"/>.
+        /// </summary>
+        /// <paramref name="audience"/> is the audience that was validated successfully.
+        public AudienceValidationResult(string audience) : base(ValidationFailureType.ValidationSucceeded)
+        {
+            IsValid = true;
+            Audience = audience;
+        }
+
+        /// <summary>
+        /// Creates an instance of <see cref="IssuerValidationResult"/>
+        /// </summary>
+        /// <paramref name="audience"/> is the audience that was intended to be validated.
+        /// <paramref name="validationFailure"/> is the <see cref="ValidationFailureType"/> that occurred during validation.
+        /// <paramref name="exceptionDetail"/> is the <see cref="ExceptionDetail"/> that occurred during validation.
+        public AudienceValidationResult(string audience, ValidationFailureType validationFailure, ExceptionDetail exceptionDetail)
+            : base(validationFailure, exceptionDetail)
+        {
+            IsValid = false;
+            Audience = audience;
+        }
+
+        /// <summary>
+        /// Gets the <see cref="Exception"/> that occurred during validation.
+        /// </summary>
+        public override Exception Exception
+        {
+            get
+            {
+                if (_exception != null || ExceptionDetail == null)
+                    return _exception;
+
+                HasValidOrExceptionWasRead = true;
+                _exception = ExceptionDetail.GetException();
+                SecurityTokenInvalidAudienceException securityTokenInvalidAudienceException = _exception as SecurityTokenInvalidAudienceException;
+                if (securityTokenInvalidAudienceException != null)
+                {
+                    securityTokenInvalidAudienceException.InvalidAudience = Audience;
+                    securityTokenInvalidAudienceException.ExceptionDetail = ExceptionDetail;
+                    securityTokenInvalidAudienceException.Source = "Microsoft.IdentityModel.Tokens";
+                }
+
+                return _exception;
+            }
+        }
+
+        /// <summary>
+        /// Gets the audience that was validated or intended to be validated.
+        /// </summary>
+        public string Audience { get; } = "null";
+    }
+}
--- a/src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
@@ -48,8 +48,7 @@ namespace Microsoft.IdentityModel.Tokens
 
                 HasValidOrExceptionWasRead = true;
                 _exception = ExceptionDetail.GetException();
-                SecurityTokenInvalidAudienceException securityTokenInvalidAudienceException = _exception as SecurityTokenInvalidAudienceException;
-                if (securityTokenInvalidAudienceException != null)
+                if (_exception is SecurityTokenInvalidAudienceException securityTokenInvalidAudienceException)
                 {
                     securityTokenInvalidAudienceException.InvalidAudience = Audience;
                     securityTokenInvalidAudienceException.ExceptionDetail = ExceptionDetail;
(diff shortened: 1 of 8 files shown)


src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
```diff
@@ -0,0 +1,68 @@
...
+                }
+
+                return _exception;
+            }
+        }
+
+        /// <summary>
+        /// Gets the audience that was validated or intended to be validated.
+        /// </summary>
+        public string Audience { get; } = "null";
```
keegan-caruso: Generally, you would enforce that the audience is not null  in the constructor

src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        /// <summary>
+        /// Gets the <see cref="Exception"/> that occurred during validation.
+        /// </summary>
+        public override Exception Exception
+        {
+            get
+            {
+                if (_exception != null || ExceptionDetail == null)
```
westin-m, line 46: This will return null if  `_exception` and `ExceptionDetail` are both null. Is that intentional?

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Audience.cs
```diff
@@ -88,7 +106,121 @@ public static void ValidateAudience(IEnumerable<string> audiences, SecurityToken
...
+                            LogMessages.IDX10206,
+                            null),
+                        typeof(SecurityTokenInvalidAudienceException),
+                        new StackFrame(true)));
+
+            // create enumeration of all valid audiences from validationParameters
+            IEnumerable<string> validationParametersAudiences;
+
+            if (validationParameters.ValidAudiences == null)
+                validationParametersAudiences = new[] { validationParameters.ValidAudience };
```
keegan-caruso: We shouldn't have to allocate an array for the single value case

keegan-caruso: We should be able to avoid this and the concat if we just pass ValidAudience and ValidAudiences to AudienceIsValid

src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
iNinja: Yes. We return an exception if either an exception is stored, or the details to build one are present.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Audience.cs
iNinja: This was copy pasted from the original variant of the ValidateAudience method, but I agree that we should refactor this. Thank you for the input.

src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
iNinja: This was a slip, as initially the success case was not receiving the Audience. But right now all constructors pass it, so this is superfluous and being removed.

src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
westin-m: Does the predicate `ExceptionDetail == null` mean the details to build an exception are present?

iNinja: No. We have 3 situations here, in order of evaluation:

- _exception is not null => we return _exception with the concrete exception stored.
- _exception is null, ExceptionDetail is null => we return null. There is no exception information.
- _exception is null, ExceptionDetail is **not null**, we build an exception from ExceptionDetail and return it.

brentschmaltz: @westin-m  ExceptionDetail != null means we can build the exception.

src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
```diff
@@ -0,0 +1,68 @@
...
+                    return _exception;
+
+                HasValidOrExceptionWasRead = true;
+                _exception = ExceptionDetail.GetException();
+                SecurityTokenInvalidAudienceException securityTokenInvalidAudienceException = _exception as SecurityTokenInvalidAudienceException;
+                if (securityTokenInvalidAudienceException != null)
```
brentschmaltz: If this is null, we have a programming error.

iNinja: I was not sure if the cast from `Exception` to `SecurityTokenInvalidAudienceException` would turn this into an optional, so I left it as it was when duplicating from `IssuerValidationResult`.

If you think we don't think this check, I can remove it in both.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Audience.cs
iNinja: @keegan-caruso the current method is being used by the existing ValidateAudience. This is an internal copy of it as it is a work in progress.

I will make an item to look into this at a further date. I would like to run some benchmarks before changing how this works and not risk affecting production services.

iNinja: Updated the method.

src/Microsoft.IdentityModel.Tokens/Validation/AudienceValidationResult.cs
iNinja: Changed to pattern matching.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Audience.cs
```diff
@@ -3,12 +3,29 @@
...
+    /// Definition for delegate that will validate the audiences value in a token.
+    /// </summary>
+    /// <param name="audiences">The audiences to validate.</param>
+    /// <param name="securityToken">The <see cref="SecurityToken"/> that is being validated.</param>
+    /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+    /// <param name="callContext"></param>
+    /// <returns>A <see cref="IssuerValidationResult"/>that contains the results of validating the issuer.</returns>
+    /// <remarks>This delegate is not expected to throw.</remarks>
+    internal delegate AudienceValidationResult ValidateAudience(
+        IEnumerable<string> audiences,
```
brentschmaltz, line 24: How much work would it be to change IEnumberable<string> to IList<string>?

brentschmaltz: Remember this call will arrive from 
```cs
TokenHandler.ValidateTokenAsync(
    string token,
    CallContext callContext, 
    CancellationToken cancellationToken)
{
    // ...
}
```

iNinja: After speaking with Keegan about this, I generally agree that we should try to move to List.
However, right now we have 2 methods co-existing, one of which is public and receives an `IEnumerable`.

I am tempted to do something like:

```csharp
List<string> internalAudiences;  
if (audiences is List<string>)  
    internalAudiences = audiences;  
else  
    internalAudiences = audiences.ToList()
```

inside of the audience matching method and iterate over the `List<string>`, without breaking our public interface or  introducing the variance from `IEnumerable` to `IList` across the two methods.

iNinja: Updated to use IList on the new ValidateAudience method.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Audience.cs
```diff
@@ -62,17 +79,10 @@ public static void ValidateAudience(IEnumerable<string> audiences, SecurityToken
...
-                validationParametersAudiences = new[] { validationParameters.ValidAudience };
-            else if (string.IsNullOrWhiteSpace(validationParameters.ValidAudience))
-                validationParametersAudiences = validationParameters.ValidAudiences;
-            else
-                validationParametersAudiences = validationParameters.ValidAudiences.Concat(new[] { validationParameters.ValidAudience });
+            if (audiences is not List<string> audiencesAsList)
+                audiencesAsList = audiences.ToList();
 
-            if (AudienceIsValid(audiences, validationParameters, validationParametersAudiences))
+            if (AudienceIsValid(audiencesAsList, validationParameters))
```
keegan-caruso, line 85: If we are taking this change, then AudienceIsValid should also be updated to be List<string> instead of IEnumerable<string>

iNinja: Addressed as discussed.

FuPingFranco (approved):
LGTM

keegan-caruso (approved):

PR #2658, opened 2024-06-20
Add a constant for the create prompt in OpenIdConnectPrompt class

ghost:
The class `OpenIdConnectPrompt` is missing the `create` prompt from [Initiating User Registration via OpenID Connect 1.0
](https://openid.net/specs/openid-connect-prompt-create-1_0.html). This PR adds it.


JoshLozensky (approved):

JoshLozensky:
@joegoldman2 thanks for the contribution!!

PR #2672, opened 2024-06-27
Issuer signing key validation: Remove exceptions

iNinja:
# Issuer signing key validation: Remove exceptions

Removed exceptions as part of signing key validation.

## Description
- Implemented new (internal momentarily) method for signing key validation which replaces exception throwing with a return of the new SigningKeyValidationResult containing the exception information.
- Added new delegate signature for signing key validation


src/Microsoft.IdentityModel.Tokens/Validation/SigningKeyValidationResult.cs
```diff
@@ -0,0 +1,69 @@
...
+        private Exception? _exception;
+
+        /// <summary>
+        /// Creates an instance of <see cref="SigningKeyValidationResult"/>
+        /// </summary>
+        /// <paramref name="signingKey"/> is the security key that was validated successfully.
+        public SigningKeyValidationResult(SecurityKey? signingKey)
+            : base(ValidationFailureType.ValidationSucceeded)
+        {
+            SigningKey = signingKey;
```
keegan-caruso, line 23: Can this actually be null in the success case?

src/Microsoft.IdentityModel.Tokens/Validation/Validators.IssuerSecurityKey.cs
```diff
@@ -2,12 +2,37 @@
...
+#nullable enable
 namespace Microsoft.IdentityModel.Tokens
 {
+    /// <summary>
+    /// Definition for delegate that will validate the <see cref="SecurityKey"/> that signed a <see cref="SecurityToken"/>.
+    /// </summary>
+    /// <param name="signingKey">The security key to validate.</param>
+    /// <param name="securityToken">The <see cref="SecurityToken"/> that is being validated.</param>
+    /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+    /// <param name="callContext"></param>
```
keegan-caruso, line 21: a few of these are empty

src/Microsoft.IdentityModel.Tokens/Validation/SigningKeyValidationResult.cs
iNinja: There is a case where the TVP have `ValidateIssuerSigningKey` set to `false` and we return in a successful manner while logging a warning about the fact that no validation has been performed. The security key can be `null` in this scenario without affecting the outcome.
This is a common pattern across our validators that we will need to decide if we want to differentiate as `NotEvaluated` or something similar. Right now the existing public methods return in the exact same manner if either the validation succeeds or no validation is performed (by not throwing an exception).

iNinja: Also in the case where `TokenValidationParameters` have `RequireSignedTokens` set to `false`, a similar logic follows if the key is null.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.IssuerSecurityKey.cs
iNinja: Following up on our chat regarding all `CallContext` references.

src/Microsoft.IdentityModel.Tokens/Validation/SigningKeyValidationResult.cs
keegan-caruso: thanks

GeoK (approved):

keegan-caruso (approved):

FuPingFranco (approved):

PR #2680, opened 2024-07-02
Remove dependency on AadIssuerValidator.GetTenantIdFromToken in ValidateIssuerSigningKey

keegan-caruso:
# Remove dependency on AadIssuerValidator.GetTenantIdFromToken in ValidateIssuerSigningKey

## Description

Removes the AadIssuerValidator.GetTenantIdFromToken dependency with an AppContext switch to fallback to previous behavior


src/Microsoft.IdentityModel.Validators/LogMessages.cs
```diff
@@ -24,5 +24,8 @@ internal static class LogMessages
         public const string IDX40005 = "IDX40005: Token issuer: '{0}', does not match the signing key issuer: '{1}'.";
         public const string IDX40007 = "IDX40007: RequireSignedTokens property on ValidationParameters is set to true, but the issuer signing key is null.";
         public const string IDX40008 = "IDX40008: When setting LastKnownGoodLifetime, the value must be greater than or equal to zero. value: '{0}'.";
+
+        public const string IDX40009 = "IDX40009: 'tid' claim didn't have a value.";
```
GeoK: Perhaps "'tid' claim was not found in the token"

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -67,9 +76,14 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
                 if (string.IsNullOrWhiteSpace(signingKeyIssuer))
                     return true;
 
-                var tenantIdFromToken = AadIssuerValidator.GetTenantIdFromToken(securityToken);
+                var tenantIdFromToken = GetTid(securityToken);
                 if (string.IsNullOrEmpty(tenantIdFromToken))
-                    return true;
+                {
+                    if (DontFailOnMissingTid())
+                        return true;
```
brentschmaltz, line 83: Why do we return 'true' here?

test/Microsoft.IdentityModel.Validators.Tests/AadSigningKeyIssuerValidatorTests.cs
```diff
@@ -167,13 +170,15 @@ public void ValidateIssuerSigningKeyTests(AadSigningKeyIssuerTheoryData theoryDa
...
             {
+                theoryData.SetupAction?.Invoke();
                 var result = AadTokenValidationParametersExtension.ValidateIssuerSigningKey(theoryData.SecurityKey, theoryData.SecurityToken, theoryData.OpenIdConnectConfiguration);
                 theoryData.ExpectedException.ProcessNoException(context);
                 Assert.True(result);
             }
             catch (Exception ex)
             {
                 theoryData.ExpectedException.ProcessException(ex, context);
+                theoryData.TearDownAction?.Invoke();
```
pmaytak, line 184: nit: put this in `finally`?

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -101,6 +115,26 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
...
+        private static string GetTid(SecurityToken securityToken)
+        {
+            if (securityToken is JsonWebToken jsonWebToken)
+            {
+                if (jsonWebToken.TryGetPayloadValue<string>("tid", out string tid))
+                    return tid;
+
+                return string.Empty;
+            }
+            else if (securityToken is JwtSecurityToken jwtSecurityToken)
```
msbw2: Unnecessary `else` since previous block returns.
```suggestion
            if (securityToken is JwtSecurityToken jwtSecurityToken)
```

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -101,6 +115,26 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
...
+            {
+                if ((jwtSecurityToken.Payload.TryGetValue("tid", out object tidObject) && tidObject is string tid))
+                    return tid;
+
+                return string.Empty;
+            }
+            else
```
msbw2: Another unnecessary `else` since all previous blocks returns.
```suggestion
```

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -41,6 +43,13 @@ public static void EnableAadSigningKeyIssuerValidation(this TokenValidationParam
             };
         }
 
+        internal const string DontFailOnMissingTidSwitch = "Switch.Microsoft.IdentityModel.DontFailOnMissingTidValidateIssuerSigning";
+
+        private static bool DontFailOnMissingTid()
```
brentschmaltz, line 50: If this extension represents the correct way to validate a token from AAD, i don't think we should have an opt out.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -625,6 +625,11 @@ public Claim GetClaim(string key)
             return Payload.GetClaim(key, Issuer ?? ClaimsIdentity.DefaultIssuer);
         }
 
+        /// <summary>
+        /// Gets the keys on the JsonWebToken.
+        /// </summary>
+        public IReadOnlyCollection<string> Keys => Payload._jsonClaims.Keys;
```
brentschmaltz: Do we need to make this public as it is only used internally.
What can a user do with this value?

brentschmaltz: Name might be a bit misleading as we normally think of keys as a securitykey.
Also, these keys are only from the payload.

keegan-caruso: JsonWebToken isn't internals visible to Validators. I'd really like to avoid it.

The utility is the same as we are using it for, if you want to look at the types of the claims without creating the claims.

Happy to rename if we can come up with a better name

GeoK: I agree on the naming being a bit misleading. Perhaps PayloadClaimNames?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -625,6 +625,11 @@ public Claim GetClaim(string key)
             return Payload.GetClaim(key, Issuer ?? ClaimsIdentity.DefaultIssuer);
         }
 
+        /// <summary>
+        /// Gets the names of the payload claims on the JsonWebToken.
+        /// </summary>
+        public IReadOnlyCollection<string> PayloadClaimNames => Payload._jsonClaims.Keys;
```
brentschmaltz: I think this should be internal.
There is not much a user can do with it.

brentschmaltz (approved):

test/Microsoft.IdentityModel.Validators.Tests/AadSigningKeyIssuerValidatorTests.cs
```diff
@@ -321,6 +339,70 @@ public static TheoryData<AadSigningKeyIssuerTheoryData> ValidateIssuerSigningKey
...
+                });
+
+                theoryData.Add(new AadSigningKeyIssuerTheoryData
+                {
+                    TestId = "Fails_With_Multiple_tids",
+                    SecurityKey = KeyingMaterial.JsonWebKeyP256,
+                    SecurityToken = new JsonWebToken(
+                        Default.Jwt(Default.SecurityTokenDescriptor(
+                            Default.SymmetricSigningCredentials,
+                            [tidClaim, issClaim, new Claim("TID", Guid.NewGuid().ToString())]))),
```
pmaytak, line 401: Worth it to add another test case with different order?

`[issClaim, new Claim("TID", Guid.NewGuid().ToString()), tidClaim]))),`

pmaytak: And maybe this too:
`[issClaim, new Claim("TID", Guid.NewGuid().ToString())]))),`

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -101,6 +117,48 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
             return true;
         }
 
+        private static string GetTid(SecurityToken securityToken)
+        {
+            switch (securityToken)
+            {
+                case JsonWebToken jsonWebToken:
+                    if (jsonWebToken.TryGetPayloadValue<string>("tid", out string tid))
```
pmaytak: `"tid"`
Should use existing constants for this?

pmaytak (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -625,6 +625,11 @@ public Claim GetClaim(string key)
             return Payload.GetClaim(key, Issuer ?? ClaimsIdentity.DefaultIssuer);
         }
 
+        /// <summary>
+        /// Gets the names of the payload claims on the JsonWebToken.
+        /// </summary>
+        internal IReadOnlyCollection<string> PayloadClaimNames => Payload._jsonClaims.Keys;
```
msbw2, line 631: The `Keys` property on a dictionary allocates a `KeyCollection` object.

keegan-caruso: Suggestions for how we can do better?

msbw2: I can't really think of anything. Dictionaries don't expose enough (to protect internal state). I just wanted to call this out in case it was overlooked and maybe someone else might have an idea of how to get the same info via an existing collection.

keegan-caruso: I don't think we can do much better with our current data structures. It is internal now so we can change it in the future if needed

pmaytak: It creates a [KeyCollection](https://github.com/dotnet/runtime/blob/284aeaf17c2a1c9eb2cbfc5d3d82f312edbb850a/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L216) and passes dictionary reference, which is what's enumerated over. I think this works?

Another option would be iterating over _jsonClaims directly. Add something like `internal HasClaimType(string type)` to JsonWebToken.

PR #2681, opened 2024-07-03
Token Type validation:  Remove exceptions

FuPingFranco:
# Token Type validation:  Remove exceptions

## Description

- Implemented new (internal momentarily) method for token type validation which replaces exception throwing with a return of the new TokenTypeValidationResult containing the exception information.


src/Microsoft.IdentityModel.Tokens/Validation/Validators.TokenType.cs
```diff
@@ -57,5 +59,117 @@ public static string ValidateTokenType(string type, SecurityToken securityToken,
...
+#pragma warning disable CA1031 // Do not catch general exception types
+            catch (Exception ex)
+#pragma warning restore CA1031 // Do not catch general exception types
+            {
+                return new TokenTypeValidationResult(
+                    type,
+                    ValidationFailureType.TokenTypeValidationFailed,
+                    new ExceptionDetail(
+                        new MessageDetail(
+                            LogMessages.IDX10259,
```
iNinja, line 169: `LogMessages.IDX10259` does not take parameters.
Given we are passing the exception as the `innerException`, I am not sure we need to include more information on this message.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationFailureType.cs
```diff
@@ -39,6 +39,12 @@ private class IssuerValidationFailure : ValidationFailureType { internal IssuerV
         public static readonly ValidationFailureType AudienceValidationFailed = new AudienceValidationFailure("AudienceValidationFailed");
         private class AudienceValidationFailure : ValidationFailureType { internal AudienceValidationFailure(string name) : base(name) { } }
 
+        /// <summary>
+        /// Defines a type that represents that token type validation failed.
+        /// </summary>
+        public static readonly ValidationFailureType TokenTypeValidationFailed = new TokenTypeValidationFailure("TokenTypeValidationFailure");
```
iNinja: nit: All ValidationFailure names use the verb form `TokenTypeValidationFailed`.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.TokenType.cs
```diff
@@ -57,5 +59,117 @@ public static string ValidateTokenType(string type, SecurityToken securityToken,
...
+                    type,
+                    ValidationFailureType.NullArgument,
+                    new ExceptionDetail(
+                        new MessageDetail(
+                            LogMessages.IDX10000,
+                            LogHelper.MarkAsNonPII(nameof(validationParameters))),
+                        typeof(ArgumentNullException),
+                        new StackFrame(true)));
+
+            if (validationParameters.TypeValidator == null && (validationParameters.ValidTypes == null || !validationParameters.ValidTypes.Any()))
```
iNinja, line 105: From previous PRs, I remember we are trying to avoid using the `IEnumerable` type, particulrarly with `Any()` and `foreach` as they incur an allocation for the enumerator.
Would it make sense to extract `validationParameters.ValidTypes` as a `List<T>` and use that instead?
This point could use input from someone more familiar with .NET.

This would also apply to further uses of `ValidTypes` below.

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -73,6 +73,7 @@ internal static class LogMessages
         public const string IDX10256 = "IDX10256: Unable to validate the token type. TokenValidationParameters.ValidTypes is set, but the 'typ' header claim is null or empty.";
         public const string IDX10257 = "IDX10257: Token type validation failed. Type: '{0}'. Did not match: validationParameters.TokenTypes: '{1}'.";
         public const string IDX10258 = "IDX10258: Token type validated. Type: '{0}'.";
+        public const string IDX10259 = "IDX10259: Unable to validate the token type, delegate threw an exception";
```
SaurabhMSFT: Missing full stop.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.TokenType.cs
```diff
@@ -57,5 +59,117 @@ public static string ValidateTokenType(string type, SecurityToken securityToken,
 
             return type;
         }
+
+        /// <summary>
+        /// Validates the type of the token.
+        /// </summary>
+        /// <param name="type">The token type or <c>null</c> if it couldn't be resolved (e.g from the 'typ' header for a JWT).</param>
```
SaurabhMSFT, line 66: 'typ' -> 'type'

src/Microsoft.IdentityModel.Tokens/Validation/Validators.TokenType.cs
```diff
@@ -57,5 +59,117 @@ public static string ValidateTokenType(string type, SecurityToken securityToken,
...
+        /// <exception cref="ArgumentNullException">If <paramref name="securityToken"/> is null.</exception>
+        /// <exception cref="SecurityTokenInvalidTypeException">If <paramref name="type"/> is null or whitespace and <see cref="TokenValidationParameters.ValidTypes"/> is not null.</exception>
+        /// <exception cref="SecurityTokenInvalidTypeException">If <paramref name="type"/> failed to match <see cref="TokenValidationParameters.ValidTypes"/>.</exception>
+        /// <remarks>An EXACT match is required. <see cref="StringComparison.Ordinal"/> (case sensitive) is used for comparing <paramref name="type"/> against <see cref="TokenValidationParameters.ValidTypes"/>.</remarks>
+#pragma warning disable CA1801 // TODO: remove pragma disable once callContext is used for logging
+        internal static TokenTypeValidationResult ValidateTokenType(string? type, SecurityToken? securityToken, TokenValidationParameters validationParameters, CallContext callContext)
+#pragma warning restore CA1801 // TODO: remove pragma disable once callContext is used for logging
+        {
```
SaurabhMSFT, line 78: This method uses a mix of bracket style on if blocks. Some single line statements are enclosed in a bracket and some are not. It would be preferable to choose one style.

src/Microsoft.IdentityModel.Tokens/Validation/TokenTypeValidationResult.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// Creates an instance of <see cref="TokenTypeValidationResult"/>
+        /// </summary>
+        /// <paramref name="type"/> is the type against which the token was validated.
+        /// <paramref name="validationFailure"/> is the <see cref="ValidationFailureType"/> that occurred during validation.
+        /// <paramref name="exceptionDetail"/> is the <see cref="ExceptionDetail"/> that occurred during validation.
+        public TokenTypeValidationResult(string? type, ValidationFailureType validationFailure, ExceptionDetail exceptionDetail)
+            : base(validationFailure, exceptionDetail)
+        {
+            Type = type;
+            IsValid = false;
```
SaurabhMSFT, line 38: This is not necessary and should be removed.

src/Microsoft.IdentityModel.Tokens/Validation/TokenTypeValidationResult.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                if (_exception != null || ExceptionDetail == null)
+                    return _exception;
+
+                HasValidOrExceptionWasRead = true;
+                _exception = ExceptionDetail.GetException();
+                if (_exception is SecurityTokenInvalidTypeException securityTokenInvalidTypeException)
+                {
+                    securityTokenInvalidTypeException.InvalidType = Type;
+                    securityTokenInvalidTypeException.Source = "Microsoft.IdentityModel.Tokens";
```
SaurabhMSFT: Can you put this string in a constant ?

src/Microsoft.IdentityModel.Tokens/Validation/TokenTypeValidationResult.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        /// <summary>
+        /// Creates an instance of <see cref="TokenTypeValidationResult"/>.
+        /// </summary>
+        /// <paramref name="type"/> is the type against which the token was validated.
+        public TokenTypeValidationResult(string? type)
+            : base(ValidationFailureType.ValidationSucceeded)
+        {
+            Type = type;
+            IsValid = true;
```
SaurabhMSFT, line 25: If 'ValidationResult' is 'ValidationFailureType.ValidationSucceeded' then is it  valid in all conditions? If yes, then this set logic should be in `ValidationResult` constructor.

If not, then a constructor of `ValidationResult` should be updated to include the `IsValid` parameter. 

Two step constructions should be avoided wherever possible.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.TokenType.cs
FuPingFranco: I agree it's not necessary but, I think it's worthwhile to have it clear that this was caused by the customers delegate validation instead of our default validation as the innerexception might not contain enough details of the overall validation process.  I can remove it if you think it's too much. Do you have a preference?

iNinja: The new `LogMessages.IDX10259` gives enough information to note that the exception was thrown by the delegate. I believe that, paired with the inner exception, provides enough information to the user as to what happened.

Hopefully if we move forward with the discussion we are currently having, this whole use case about us calling the delegate will disappear altogether.

FuPingFranco: I agree, will leave this as is for now to keep the scope of this PR in line with our current work until we start moving forward with the new approach being discussed now. Thanks for the feedback!

src/Microsoft.IdentityModel.Tokens/Validation/TokenTypeValidationResult.cs
FuPingFranco: The `ValidationResult` is still a work in progress so, we will make additional changes on a separate PR also, since we are returning IsValid as true when no validation has been performed this won't affect the current state of `ValidationResult` and the Type parameter still needs to be assigned even if we remove this constructor, so I do not see how we would avoid the two-step construction of the object. However, it could be simplified to only declaring type and setting the IsValid on `ValidationResult` base. Thanks for the feedback @SaurabhMSFT, I'll definitely keep it mind for when we have a finalized `ValidationResult`.

src/Microsoft.IdentityModel.Tokens/Validation/TokenTypeValidationResult.cs
FuPingFranco: There's more work in progress to finalize our `ValidationResult` object. Will keep for now. Thanks @SaurabhMSFT.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.TokenType.cs
FuPingFranco: I think this can be left with Any() for now as IEnumerable since, that is the type that comes from the `TokenValidationParameters` however, we should consider this for the new `ValidationParameters` as changing it into IList doesn't seem like a good fit to me for the small gain in allocation as it stands right now. ValidTypes in this case will only check if the ValidTypes come as empty and the other instance where we use it, is to check whether or not a type is a part of the collection, not sure if the additional overhead of transforming to IList is worth it?

FuPingFranco: I checked with @keegan-caruso on this for an additional opinion and we think it can be left as Any() for now.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationFailureType.cs
FuPingFranco: Updated along with LifetimeValidationFailed inconsistency. Thanks!

PR #2682, opened 2024-07-03
Dispose SignatureProviders after moving to staging cache.

brentschmaltz:
This change solves a couple of issues.

1. A race condition where a SignatureProvider was disposed but was still able to be obtained from the cache.
2. SignatureProviders were not disposed when compacting, which results in a number of objects on the finalizer Q.

SignatureProviders are ref counted, but not removed and disposed when the ref count goes to zero. The logic is most likely the SignatureProvider will be used again in a short period.

The significant change is that on compaction SignatureProviders are moved to a 'compactCache'. Periodically, the 'compactCache' is examined for SignatureProviders that can be disposed.

The CryptoProviderFactory will never return a SignatureProvider from the 'compactCache' when asked to create a SignatureProvider.

SignatureProviders that were not cached, will be disposed when released.

It is helpful when reviewing, all tasks are run in serial on a background task.

### DIFF
--- a/buildpackLocal.bat
+++ b/buildpackLocal.bat
@@ -1,0 +1,3 @@
+dotnet clean Product.proj > clean.log
+dotnet build /r Product.proj
+dotnet pack --no-restore -o c:\localpackages --no-build Product.proj
--- a/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
+++ b/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
@@ -18,6 +18,8 @@ namespace Microsoft.IdentityModel.Tokens
         private static readonly ConcurrentDictionary<string, string> _typeToAlgorithmMap = new ConcurrentDictionary<string, string>();
         private static readonly object _cacheLock = new object();
         private static int _defaultSignatureProviderObjectPoolCacheSize = Environment.ProcessorCount * 4;
+        private static string _typeofAsymmetricSignatureProvider = typeof(AsymmetricSignatureProvider).ToString();
+        private static string _typeofSymmetricSignatureProvider = typeof(SymmetricSignatureProvider).ToString();
         private int _signatureProviderObjectPoolCacheSize = _defaultSignatureProviderObjectPoolCacheSize;
 
         /// <summary>
@@ -513,7 +515,13 @@ namespace Microsoft.IdentityModel.Tokens
             {
                 signatureProvider = CustomCryptoProvider.Create(algorithm, key, willCreateSignatures) as SignatureProvider;
                 if (signatureProvider == null)
-                    throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(LogMessages.IDX10646, LogHelper.MarkAsNonPII(algorithm), key, LogHelper.MarkAsNonPII(typeof(SignatureProvider)))));
+                    throw LogHelper.LogExceptionMessage(
+                        new InvalidOperationException(
+                            LogHelper.FormatInvariant(
+                                LogMessages.IDX10646,
+                                LogHelper.MarkAsNonPII(algorithm),
+                                key,
+                                LogHelper.MarkAsNonPII(typeof(SignatureProvider)))));
 
                 return signatureProvider;
             }
@@ -523,7 +531,7 @@ namespace Microsoft.IdentityModel.Tokens
             bool createAsymmetric = true;
             if (key is AsymmetricSecurityKey)
             {
-                typeofSignatureProvider = typeof(AsymmetricSignatureProvider).ToString();
+                typeofSignatureProvider = _typeofAsymmetricSignatureProvider;
             }
             else if (key is JsonWebKey jsonWebKey)
             {
@@ -533,11 +541,11 @@ namespace Microsoft.IdentityModel.Tokens
                     {
                         if (convertedSecurityKey is AsymmetricSecurityKey)
                         {
-                            typeofSignatureProvider = typeof(AsymmetricSignatureProvider).ToString();
+                            typeofSignatureProvider = _typeofAsymmetricSignatureProvider;
                         }
                         else if (convertedSecurityKey is SymmetricSecurityKey)
                         {
-                            typeofSignatureProvider = typeof(SymmetricSignatureProvider).ToString();
+                            typeofSignatureProvider = _typeofSymmetricSignatureProvider;
                             createAsymmetric = false;
                         }
                     }
@@ -545,10 +553,10 @@ namespace Microsoft.IdentityModel.Tokens
                     else
                     {
                         if (jsonWebKey.Kty == JsonWebAlgorithmsKeyTypes.RSA || jsonWebKey.Kty == JsonWebAlgorithmsKeyTypes.EllipticCurve)
-                            typeofSignatureProvider = typeof(AsymmetricSignatureProvider).ToString();
+                            typeofSignatureProvider = _typeofAsymmetricSignatureProvider;
                         else if (jsonWebKey.Kty == JsonWebAlgorithmsKeyTypes.Octet)
                         {
-                            typeofSignatureProvider = typeof(SymmetricSignatureProvider).ToString();
+                            typeofSignatureProvider = _typeofSymmetricSignatureProvider;
                             createAsymmetric = false;
                         }
                     }
@@ -560,12 +568,20 @@ namespace Microsoft.IdentityModel.Tokens
             }
             else if (key is SymmetricSecurityKey)
             {
-                typeofSignatureProvider = typeof(SymmetricSignatureProvider).ToString();
+                typeofSignatureProvider = _typeofSymmetricSignatureProvider;
                 createAsymmetric = false;
             }
 
             if (typeofSignatureProvider == null)
-                throw LogHelper.LogExceptionMessage(new NotSupportedException(LogHelper.FormatInvariant(LogMessages.IDX10621, LogHelper.MarkAsNonPII(typeof(SymmetricSignatureProvider)), LogHelper.MarkAsNonPII(typeof(SecurityKey)), LogHelper.MarkAsNonPII(typeof(AsymmetricSecurityKey)), LogHelper.MarkAsNonPII(typeof(SymmetricSecurityKey)), LogHelper.MarkAsNonPII(key.GetType()))));
+                throw LogHelper.LogExceptionMessage(
+                    new NotSupportedException(
+                        LogHelper.FormatInvariant(
+                            LogMessages.IDX10621,
+                            LogHelper.MarkAsNonPII(typeof(SymmetricSignatureProvider)),
+                            LogHelper.MarkAsNonPII(typeof(SecurityKey)),
+                            LogHelper.MarkAsNonPII(typeof(AsymmetricSecurityKey)),
+                            LogHelper.MarkAsNonPII(typeof(SymmetricSecurityKey)),
+                            LogHelper.MarkAsNonPII(key.GetType()))));
 
             if (CacheSignatureProviders && cacheProvider)
             {
@@ -592,7 +608,7 @@ namespace Microsoft.IdentityModel.Tokens
                         signatureProvider = new SymmetricSignatureProvider(key, algorithm, willCreateSignatures);
 
                     if (ShouldCacheSignatureProvider(signatureProvider))
-                        CryptoProviderCache.TryAdd(signatureProvider);
+                        signatureProvider.IsCached = CryptoProviderCache.TryAdd(signatureProvider);
                 }
             }
             else
@@ -737,7 +753,7 @@ namespace Microsoft.IdentityModel.Tokens
             signatureProvider.Release();
             if (CustomCryptoProvider != null && CustomCryptoProvider.IsSupportedAlgorithm(signatureProvider.Algorithm))
                 CustomCryptoProvider.Release(signatureProvider);
-            else if (signatureProvider.CryptoProviderCache == null && signatureProvider.RefCount == 0)
+            else if (signatureProvider.CryptoProviderCache == null && signatureProvider.RefCount == 0 && !signatureProvider.IsCached)
                 signatureProvider.Dispose();
         }
     }
--- a/src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
+++ b/src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.IdentityModel.Abstractions;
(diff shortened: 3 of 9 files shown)


src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -717,7 +733,7 @@ public virtual void ReleaseRsaKeyWrapProvider(RsaKeyWrapProvider provider)
         }
 
         /// <summary>
-        /// When finished with a <see cref="SignatureProvider"/> call this method for cleanup. The default behavior is to call <see cref="SignatureProvider.Dispose()"/>
+        /// When finished with a <see cref="SignatureProvider"/> call this method for cleanup./>
```
ghost:
```suggestion
        /// When finished with a <see cref="SignatureProvider"/> call this method for cleanup.
```

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -30,61 +29,59 @@ namespace Microsoft.IdentityModel.Tokens
     internal class EventBasedLRUCache<TKey, TValue>
     {
         internal delegate void ItemRemoved(TValue Value);
+        internal delegate bool ShouldRemove(TValue Value);
 
         private readonly int _capacity;
-
         // The percentage of the cache to be removed when _maxCapacityPercentage is reached.
```
msbw2: This comment belongs with the `_compactionPercentage` field (which could be made a `const`).

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -30,61 +29,59 @@ namespace Microsoft.IdentityModel.Tokens
     internal class EventBasedLRUCache<TKey, TValue>
     {
         internal delegate void ItemRemoved(TValue Value);
+        internal delegate bool ShouldRemove(TValue Value);
 
         private readonly int _capacity;
-
```
msbw2, line 35: NIT: Keep this line for spacing.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: a rebase changed the comments

src/Microsoft.IdentityModel.Tokens/SignatureProvider.cs
```diff
@@ -156,5 +156,7 @@ public virtual bool Verify(byte[] input, int inputOffset, int inputLength, byte[
         /// Gets or sets a bool indicating if this <see cref="SignatureProvider"/> is expected to create signatures.
         /// </summary>
         public bool WillCreateSignatures { get; protected set; }
+
+        internal bool WasCached { get; set; }
```
pmaytak: nit: `IsCached`? I think implies the current state a bit clearer.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -523,7 +531,7 @@ private SignatureProvider CreateSignatureProvider(SecurityKey key, string algori
             bool createAsymmetric = true;
             if (key is AsymmetricSecurityKey)
             {
-                typeofSignatureProvider = typeof(AsymmetricSignatureProvider).ToString();
+                typeofSignatureProvider = _typeofAsymmetricSignatureProvider;
```
pmaytak, line 534: This is a nice change.

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -47,13 +47,13 @@ public void CreateAndReleaseSignatureProviders(SignatureProviderTheoryData theor
                 {
                     var disposeCalled = GetSignatureProviderIsDisposedByReflect(signatureProvider);
```
pmaytak, line 48: Should we add a similar check here that `_keyedHashObjectPool` was cleared too?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -30,61 +29,59 @@ namespace Microsoft.IdentityModel.Tokens
...
         private bool _shouldStopImmediately = false;
 
-        internal ItemRemoved OnItemRemoved
-        {
-            get;
-            set;
-        }
+        internal ItemRemoved OnItemRemoved { get; set; }
+
+        internal ItemRemoved OnItemExpired { get; set; }
```
keegan-caruso: Consider using a different delegate per callback, It reads odd to have ItemRemoved Delegate for for OnItemExpired,

or just use Func<TValue, bool> since these are all internal.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -134,9 +117,10 @@ private int EventQueuePollingInterval
             _map = new ConcurrentDictionary<TKey, LRUCacheItem<TKey, TValue>>(comparer ?? EqualityComparer<TKey>.Default);
             _removeExpiredValuesIntervalInSeconds = removeExpiredValuesIntervalInSeconds;
             _removeExpiredValues = removeExpiredValues;
+            _timeForNextExpiredValuesRemoval = DateTime.UtcNow.AddSeconds(_removeExpiredValuesIntervalInSeconds);
+            _timeForNextCompaction = DateTime.UtcNow.AddSeconds(20);
```
keegan-caruso: can we use a named const instead of 20?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -28,39 +26,57 @@ public InMemoryCryptoProviderCache() : this(new CryptoProviderCacheOptions())
...
+        }
+
+        private void OnSignatureProviderRemoved(SignatureProvider signatureProvider)
+        {
+            signatureProvider.CryptoProviderCache = null;
+        }
+
+        private static bool ShouldRemoveSignatureProvider(SignatureProvider signatureProvider)
+        {
+            return signatureProvider.CryptoProviderCache == null && signatureProvider.RefCount == 0;
```
pmaytak, line 78: Should check cache state? Or is this already implied - ShouldRemoveSignatureProvider is called only when provider is cached?

```suggestion
            return signatureProvider.CryptoProviderCache == null && signatureProvider.RefCount == 0 && !signatureProvider.WasCached;
```

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -131,7 +147,7 @@ public override bool TryAdd(SignatureProvider signatureProvider)
             // The cache does NOT already have a crypto provider associated with this key.
             if (!signatureProviderCache.Contains(cacheKey))
             {
-                signatureProviderCache.SetValue(cacheKey, signatureProvider);
+                signatureProviderCache.SetValue(cacheKey, signatureProvider, DateTime.MaxValue);
```
keegan-caruso: nit: revert, the two param call already uses DateTime.MaxValue

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -186,118 +171,157 @@ public bool Contains(TKey key)
...
                     if (node.Value.ExpirationTime < DateTime.UtcNow)
                     {
                         if (_map.TryRemove(node.Value.Key, out var cacheItem))
-                            OnItemRemoved?.Invoke(cacheItem.Value);
+                            OnItemExpired?.Invoke(cacheItem.Value);
+                    }
+                }
+            }
+            catch
+            {
```
keegan-caruso, line 303: why not log here?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -186,118 +171,157 @@ public bool Contains(TKey key)
...
+                        _compactedItems.RemoveAt(i);
                     }
                 }
             }
-            catch (ObjectDisposedException ex)
+            catch
             {
-                if (LogHelper.IsEnabled(EventLogLevel.Warning))
-                    LogHelper.LogWarning(LogHelper.FormatInvariant(LogMessages.IDX10902, LogHelper.MarkAsNonPII(nameof(RemoveExpiredValues)), ex));
+            }
```
keegan-caruso, line 338: why not log here?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -514,6 +557,19 @@ public bool TryGetValue(TKey key, out TValue value)
...
+        /// Removes a particular key from the cache.
+        public bool TryRemove(TKey key)
+        {
+            if (key == null)
+                throw LogHelper.LogArgumentNullException(nameof(key));
+
+            if (!_map.TryRemove(key, out var cacheItem))
+                return false;
+
+            OnItemCompacted?.Invoke(cacheItem.Value);
```
keegan-caruso: It is confusing to call OnItemCompacted in TryRemove when we have a OnItemRemoved.

Maybe we can do better with naming?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -326,21 +355,28 @@ private void CompactLRU()
...
-                        OnItemRemoved?.Invoke(cacheItem.Value);
+                    // Since all items could have been removed by the public TryRemove() method, leaving the map empty, we need to check if a default value is returned.
+                    // Remove the item from the map only if the returned item is NOT default value.
+                    KeyValuePair<TKey, LRUCacheItem<TKey, TValue>> item = _map.FirstOrDefault();
+                    if (!item.Equals(default))
+                    {
+                        if (_map.TryRemove(item.Key, out LRUCacheItem<TKey, TValue> cacheItem))
+                        {
+                            OnItemCompacted?.Invoke(cacheItem.Value);
+                            _compactedItems.Add(cacheItem);
```
pmaytak, line 392: I don't quite understand, why use the compacted cache here. Why not just dispose/release the signature provider immediately? Is it to provide that extra time after the provider was removed from the cache, but don't dispose of it immediately in case some other code calls for it (the race condition)?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
pmaytak: Same. This makes it seem like we're doing something to the items, when it's the whole cache that's being compacted. Thinking about this more, I begin to understand it, but initially it's not super clear.

Maybe use terms like "pending removal" or "staging removal"?

Removal is also overloaded here - removal from the LRU cache or removal from _compactedItems...

I suggest add a comment to `_compactedItems` about what it's used for.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: @keegan-caruso I see, on the catch?
I think it is very unlikely that TryRemove will throw an exception, this is just in case there is a bug in .net.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: @keegan-caruso I will add it back.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: @keegan-caruso Nice catch, will fix.

src/Microsoft.IdentityModel.Tokens/SignatureProvider.cs
brentschmaltz: @pmaytak I agree, will change.

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
brentschmaltz: @pmaytak @keegan-caruso did you mean for the Symmetric case?
KeyedHash is only for Symmetric signatures.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
brentschmaltz: @keegan-caruso good catch, that was from early work where i was thinking of removing SignatureProviders that had been around a while.

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
pmaytak: Actually I meant for both - `_keyedHashObjectPool` in SymmetricSignatureProvider and `_asymmetricAdapterObjectPool` in AsymmetricSignatureProvider..

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -534,7 +590,7 @@ public bool TryRemove(TKey key, out TValue value)
             }
 
             value = cacheItem.Value;
-            OnItemRemoved?.Invoke(cacheItem.Value);
+            OnItemCompacted?.Invoke(cacheItem.Value);
```
pmaytak: In both of these TryRemove methods, shouldn't we add the cacheItem to the `_compactedItems` like we do in Compact methods?

Because currently, if TryRemove is called, OnItemCompacted will just set the `signatureProvider.CryptoProviderCache = null` but since the provider is never added to `_compactedItems`, OnItemRemoved will never be called for it, which disposes of it.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: I added it, we may be able to use it.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: @keegan-caruso @pmaytak i renamed delegates to: OnItemMovedToCompactedList, OnItemRemovedFromCompactedList, OnShouldRemoveFromCompactedList.
Then in InMemoryCryptoProviderCache i renamed the implementation to: 

OnItemMovedToCompactedList = SetCryptoProviderCacheToNull
OnItemRemovedFromCompactedList = DisposeSignatureProvider
OnShouldRemoveFromCompactedList = IsCacheNullAndRefCountZero

I think this is easier to follow.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: In our call graph the SignatureProvider calls this when the SignatureProvider was disposed or Signing / Verifying faulted. 
In the case where the finally clause is executed, dispose is called on the cached objects.

A better design would be to have TryRemove move the SignatureProvider to the compacted list.
This would need a new action in LRUCache.

For now i think it is OK but a bit fragile.

buildpackLocal.bat
```diff
@@ -0,0 +1,3 @@
+dotnet clean Product.proj > clean.log
```
pmaytak, line 1: Is this test code? Remove?

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
```diff
@@ -28,39 +26,56 @@ public InMemoryCryptoProviderCache() : this(new CryptoProviderCacheOptions())
...
+        internal CryptoProviderFactory CryptoProviderFactory { get; set; }
+
+        private static void DisposeSignatureProvider(SignatureProvider signatureProvider)
+        {
+            signatureProvider.Dispose();
+        }
+
+        private void SetCryptoProviderCacheToNull(SignatureProvider signatureProvider)
+        {
+            signatureProvider.CryptoProviderCache = null;
```
pmaytak, line 73: Shouldn't we also set `signatureProvider.IsCached = false` here? Otherwise, we set it to true when provider is cached and never set it to false. So this `if` will never execute:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/07a4b923e2448cdfb6c01841c87ede91cf8b6a98/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs#L756

src/Microsoft.IdentityModel.Tokens/SignatureProvider.cs
```diff
@@ -68,6 +68,8 @@ public void Dispose()
         /// <param name="disposing">true, if called from Dispose(), false, if invoked inside a finalizer</param>
         protected abstract void Dispose(bool disposing);
 
+        internal bool IsCached { get; set; }
```
pmaytak, line 71: Is this property needed? Can we use `SignatureProvider.CryptoProviderCache` for the same purpose? We set `IsCached` and `CryptoProviderCache` around the same place where we add the provider to the cache and we check these properties at the same place in the factory also.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -121,22 +117,26 @@ private int EventQueuePollingInterval
...
+        /// <param name="compactIntervalInSeconds">The period to wait to compact items, in seconds.</param>
         internal EventBasedLRUCache(
             int capacity,
             TaskCreationOptions options = TaskCreationOptions.None,
             IEqualityComparer<TKey> comparer = null,
             bool removeExpiredValues = false,
             int removeExpiredValuesIntervalInSeconds = 300,
-            bool maintainLRU = false)
+            bool maintainLRU = false,
+            int compactIntervalInSeconds = 20)
```
pmaytak, line 128: Does this need to be this frequent? For comparison, the expired item removal is every 5 minutes.

brentschmaltz: @pmaytak This is not how often we compact or try and compact, but the amount of time that must pass before we will compact again.
We compact when it is detected the capacity has been exceeded.

src/Microsoft.IdentityModel.Tokens/SignatureProvider.cs
brentschmaltz: @pmaytak you are correct, the only reason i put this is for insurance. If we have a bug somewhere and the CryptoCacheProvider is somehow null then we can dispose.

src/Microsoft.IdentityModel.Tokens/InMemoryCryptoProviderCache.cs
brentschmaltz: @pmaytak SetCryptoProviderCacheToNull will only involve SignatureProviders that have been cached, they get disposed when the cache is compacted.

The code in CryptoProviderFactory will only be executed when a SignatureProvider is released AND it was never cached.

Because of the api on InMemoryCryptoProviderCache.TryRemove that will trigger SetCrytpriverCacheToNull when removed, i wanted to remember that the SignatureProvider was cached.

Now i remember why i had used the variable name 'wasChached' as i wanted to distinguish that the SignatureProvider was cached originally.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -592,7 +608,7 @@ private SignatureProvider CreateSignatureProvider(SecurityKey key, string algori
                         signatureProvider = new SymmetricSignatureProvider(key, algorithm, willCreateSignatures);
 
                     if (ShouldCacheSignatureProvider(signatureProvider))
-                        CryptoProviderCache.TryAdd(signatureProvider);
+                        signatureProvider.IsCached = CryptoProviderCache.TryAdd(signatureProvider);
```
GeoK, line 611: When is this state changed? Is it needed just to immediately dispose signature providers that are not cached?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -534,7 +615,7 @@ public bool TryRemove(TKey key, out TValue value)
             }
 
             value = cacheItem.Value;
-            OnItemRemoved?.Invoke(cacheItem.Value);
+            OnItemMovedToCompactedList?.Invoke(cacheItem.Value);
```
GeoK, line 618: Do we need to add the cacheItem to _compactedItems?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -408,12 +465,20 @@ public bool SetValue(TKey key, TValue value, DateTime expirationTime)
...
                         if (_maintainLRU)
                             AddActionToEventQueue(CompactLRU);
                         else
                             AddActionToEventQueue(Compact);
+
+                        if (DateTime.UtcNow >= _timeForNextCompaction)
+                        {
+                            if (Interlocked.CompareExchange(ref _processCompactedValuesState, ActionQueuedOrRunning, ActionNotQueued) == ActionNotQueued)
+                            {
+                                _eventQueue.Enqueue(ProcessCompactedValues);
```
pmaytak, line 479:
```suggestion
                                AddActionToEventQueue(ProcessCompactedValues);
```

GeoK (approved):

pmaytak (approved):

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
brentschmaltz: TryAdd could fail.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: @GeoK If we move the cacheItem to the _compactedItems list, then eventually Dispose will be called.
We chose not to because this is a public API and could be called by the user, who may or may not want to dispose.

So, we just remove the item and set the signatureProvider.CryptoProviderCache to null.
Since EventBasedLRUCache doesn't know the type, we had to call a delegate defined in InMemoryCryptoProviderCache.

PR #2688, opened 2024-07-08
Token Type validation: Remove exceptions

FuPingFranco:
# Token Type validation:  Remove exceptions

## Description

- Implemented new (internal momentarily) method for token type validation which replaces exception throwing with a return of the new TokenTypeValidationResult containing the exception information.


iNinja (approved):
LGTM

PR #2690, opened 2024-07-09
Serialize byte arrays as base64 strings in Json Tokens

### LINKED ISSUE #2524: [Bug] byte[] values in JwtPayload et al. are serialized as arrays instead of base64.
Microsoft.IdentityModel 7.x.x

**Where is the issue?**
 * [ ] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [x] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [ ] S.IM.Tokens.Jwt
 * Other (please describe)

The app is in production and I have upgraded to a new version of Microsoft.IdentityModel.*

**Repro**

```csharp
Console.WriteLine(new JwtPayload { { "Fail", new byte[] { 0x21, 0x62, 0x36, 0x34 } } }.SerializeToJson());
//6.x.x - {"Fail":"IWI2NA=="}
//7.x.x - {"Fail":["33","98","54","52"]}
```

**Expected behavior**
byte[] arrays are serialized as base64.

**Actual behavior**
byte[] arrays are serialized as arrays.

**Possible solution**

Prior to 7.0.0, byte[] arrays were serialized as base64. This is both the default behavior of System.Text.Json and Json.NET. However since 7.0.0 this changed to serialize them as arrays. If this was intentional, it is an undocumented breaking change affecting backwards compatibility and interoperability with other systems expecting the older format!

keegan-caruso:
# Serialize byte arrays as base64 strings in Json Tokens

## Description

Serialize byte arrays as base64 strings in Json tokens. This was the behavior in 6.x releases.

Fixes #2524

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
@@ -1129,6 +1129,10 @@ namespace Microsoft.IdentityModel.Tokens.Json
 
                 writer.WriteEndObject();
             }
+            else if ((typeof(byte[])).IsAssignableFrom(objType))
+            {
+                writer.WriteBase64String(key, (byte[])obj);
+            }
             else if (typeof(IList).IsAssignableFrom(objType))
             {
                 IList list = (IList)obj;
--- a/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
@@ -1118,6 +1118,8 @@ namespace Microsoft.IdentityModel.Tokens.Json
                 writer.WriteBoolean(key, b);
             else if (obj is DateTime dt)
                 writer.WriteString(key, dt.ToUniversalTime().ToString("O", CultureInfo.InvariantCulture));
+            else if (obj is byte[] byteArray)
+                writer.WriteBase64String(key, byteArray);
             else if (typeof(IDictionary).IsAssignableFrom(objType))
             {
                 IDictionary dictionary = (IDictionary)obj;
@@ -1128,10 +1130,6 @@ namespace Microsoft.IdentityModel.Tokens.Json
                     WriteObject(ref writer, k.ToString(), dictionary[k]);
 
                 writer.WriteEndObject();
-            }
-            else if ((typeof(byte[])).IsAssignableFrom(objType))
-            {
-                writer.WriteBase64String(key, (byte[])obj);
             }
             else if (typeof(IList).IsAssignableFrom(objType))
             {
--- a/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
+++ b/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenTests.cs
@@ -44,6 +44,33 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
         };
 
         [Fact]
+        public void ByteArrayClaimsEncodedAsExpected()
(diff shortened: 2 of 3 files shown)


src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -1129,6 +1129,10 @@ public static void WriteObject(ref Utf8JsonWriter writer, string key, object obj
 
                 writer.WriteEndObject();
             }
+            else if ((typeof(byte[])).IsAssignableFrom(objType))
```
msbw2: Shouldn't this be `is byte[] bytes`? Is it even possible to inherit from `byte[]`?
```suggestion
            else if (obj is byte[] bytes)
```

westin-m (approved):

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
westin-m: > Shouldn't this be `is byte[] bytes`? Is it even possible to inherit from `byte[]`?

What's there makes sense to me because `obj` is being cast and `IsAssignableFrom` checks if that can be done, whereas `is` will check if the type inherits/implements. Either way of doing it will yield the same result.

msbw2: The main thing is `IsAssignableFrom` involves reflection whereas a simple `is` check doesn't. The reflection makes sense I think for other types but not for primitives.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -1129,6 +1129,10 @@ public static void WriteObject(ref Utf8JsonWriter writer, string key, object obj
```
msbw2: Not for this PR but this entire `if`-`else if`... chain I think should just be a giant `switch`. I'm also curious if the `IDictionary` handler works since the generic form doesn't inherit from the non-generic form.

pmaytak: > I'm also curious if the IDictionary handler works since the generic form doesn't inherit from the non-generic form.

It does. `is IDictionary` similarly returns true for generic one.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
pmaytak: +1 to use `is`
`is` seems to be more versatile as you can use it with expressions. It does seem to be equivalent to `IsAssignableFrom`. 

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#is-operator
https://learn.microsoft.com/en-us/dotnet/api/system.type.isassignablefrom?view=net-8.0#returns
https://stackoverflow.com/a/15853224

pmaytak (approved):
Approved with suggestions.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
keegan-caruso: Either way worked, but yes, is is preferable here.

msbw2 (approved):

keegan-caruso:
Re-ran the PR build and it succeeded.

PR #2692, opened 2024-07-09
Improve bypassing token signature validation

MZOLN:
# Improve bypassing token signature validation

Signature key will be processed while bypassing token signature validation. This code change ensures that the whole validation of signature block is skipped when the bypass flag is set.

### DIFF
--- a/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
+++ b/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
@@ -1343,12 +1343,12 @@ namespace System.IdentityModel.Tokens.Jwt
             bool kidMatched = false;
             IEnumerable<SecurityKey> keys = null;
 
+            if (validationParameters.RequireSignedTokens)
+                return jwtToken;
+
             if (string.IsNullOrEmpty(jwtToken.RawSignature))
             {
-                if (validationParameters.RequireSignedTokens)
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10504, jwtToken)));
-                else
-                    return jwtToken;
+                throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSignatureException(LogHelper.FormatInvariant(TokenLogMessages.IDX10504, jwtToken)));
             }
 
             if (validationParameters.IssuerSigningKeyResolverUsingConfiguration != null)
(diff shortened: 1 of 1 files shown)


src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1343,12 +1343,12 @@ private JwtSecurityToken ValidateSignature(string token, JwtSecurityToken jwtTok
             bool kidMatched = false;
             IEnumerable<SecurityKey> keys = null;
 
+            if (validationParameters.RequireSignedTokens == false)
```
brentschmaltz, line 1346: The logic was if there was a signature, we checked it if the flag was true (default).
With this change, we will not check it.

jmprieur: Does it mean it's not suitable?

MZOLN: I'm not sure that I understand the comment, did I make a mistake? 
What I was going after was to not check signature at all if the flag is set to false and otherwise check it fully.  does that assumption make sense?

brentschmaltz: @jmprieur @MZOLN i am calling out a behavior change where we always checked the signature if it existed.  Customers that had set the flag to false by mistake, would still have the signature validated if it existed.

MZOLN: OH I understand now. Thank you for the call out.

The reason for the change is because customers who set this flag purposefully, e.g. for testing, have a hard time figuring out how to use it. Currently you have to remove the signature part of the token and leave the trailing "." behind (for the split to work). Their expectation is that all signature validation is bypassed. Maybe another flag would be a  better solution?

brentschmaltz: @MZOLN that is why the flag says - Require...., rather than Skip...

I wouldn't make this change for that testing reason.

brentschmaltz: @MZOLN i don't think we should make this change.

jennyf19:
@MZOLN or @brentschmaltz are we keeping this open? I thought this was going to be closed. thanks!

brentschmaltz:
@jennyf19 closed.

PR #2693, opened 2024-07-10
Only use fxcop in netfw

keegan-caruso:
# Only use fxcop in netfw

## Description

Fx Cop should only be used in net framework targets


jennyf19 (approved):
:shipit:

build/common.props
```diff
@@ -65,4 +65,11 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="$(MicrosoftSourceLinkGitHubVersion)" PrivateAssets="All"/>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.1">
```
pmaytak, line 69: Why use FxCop at all? Thought it was deprecated. Normal Roslyn analyzers should work on .NET Fx?

pmaytak (approved):

PR #2697, opened 2024-07-10
Algorithm validation: Remove exceptions

FuPingFranco:
# Algorithm validation:  Remove exceptions

## Description

- Implemented new (internal momentarily) method for algorithm validation which replaces exception throwing with a return of the new AlgorithmValidationResult containing the exception information.


src/Microsoft.IdentityModel.Tokens/Validation/Validators.Algorithm.cs
```diff
@@ -42,5 +43,65 @@ public static void ValidateAlgorithm(string algorithm, SecurityKey securityKey,
                 });
             }
         }
+
+        /// <summary>
+        /// Validates if a given algorithm for a <see cref="SecurityKey"/> is valid.
```
westin-m:
```suggestion
        /// Validates a given algorithm for a <see cref="SecurityKey"/>.
```
"validates if it is valid" is redundant

test/Microsoft.IdentityModel.Tokens.Tests/Validation/AlgorithmValidationResultTests.cs
```diff
@@ -0,0 +1,151 @@
...
+            else
+                theoryData.ExpectedException.ProcessNoException();
+
+            IdentityComparer.AreAlgorithmValidationResultsEqual(
+                algorithmValidationResult,
+                theoryData.AlgorithmValidationResult,
+                context);
+
+            TestUtilities.AssertFailIfErrors(context);
+
```
westin-m:
```suggestion
```
extra line

westin-m (approved):

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Algorithm.cs
```diff
@@ -42,5 +43,65 @@ public static void ValidateAlgorithm(string algorithm, SecurityKey securityKey,
...
+                    ValidationFailureType.NullArgument,
+                    new ExceptionDetail(
+                        new MessageDetail(
+                            LogMessages.IDX10000,
+                            LogHelper.MarkAsNonPII(nameof(validationParameters))),
+                        typeof(ArgumentNullException),
+                        new StackFrame(true)));
+            }
+
+            if (validationParameters.AlgorithmValidator != null)
```
iNinja, line 78: Can we remove this? We are starting the process to remove all calls to delegates, as they will be implemented in the new ValidationParameters object and called from there.

iNinja: On the same note, we need to add the definition for the new delegate at the beginning of this file. I believe the name should be `AlgorithmValidatorDelegate` and match the signature to the new method returning the result type.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Algorithm.cs
```diff
@@ -42,5 +43,65 @@ public static void ValidateAlgorithm(string algorithm, SecurityKey securityKey,
...
+        /// <summary>
+        /// Validates a given algorithm for a <see cref="SecurityKey"/>.
+        /// </summary>
+        /// <param name="algorithm">The algorithm to be validated.</param>
+        /// <param name="securityKey">The <see cref="SecurityKey"/> that signed the <see cref="SecurityToken"/>.</param>
+        /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
+        /// <param name="validationParameters"><see cref="TokenValidationParameters"/> required for validation.</param>
+        /// <param name="callContext"></param>
+#pragma warning disable CA1801 // TODO: remove pragma disable once callContext is used for logging
+        internal static AlgorithmValidationResult ValidateAlgorithm(string algorithm, SecurityKey securityKey, SecurityToken securityToken, TokenValidationParameters validationParameters, CallContext callContext)
```
iNinja: Please align the parameters one per line:

```csharp
internal static AlgorithmValidationResult ValidateAlgorithm(  
    string algorithm,  
    SecurityKey securityKey,   
    SecurityToken securityToken,   
    CallContext callContext)
```

iNinja: Also we should wrap this in a #nullable block and mark `CallContext` as nullable to make it optional.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Algorithm.cs
iNinja, line 1: I believe we have enough clarity now that we could move the new implementation of `ValidateAlgorithm` to use `ValidAlgorithms` from the new `ValidationParameters`, and likely as a `List<string>`.
Let's discuss in our chat.

test/Microsoft.IdentityModel.Tokens.Tests/Validation/AlgorithmValidationResultTests.cs
```diff
@@ -0,0 +1,150 @@
...
+                            ValidationFailureType.AlgorithmValidationFailed,
+                            new ExceptionDetail(
+                                new MessageDetail(
+                                    LogMessages.IDX10696,
+                                    LogHelper.MarkAsNonPII(SecurityAlgorithms.Sha256),
+                                    securityKey),
+                                typeof(SecurityTokenInvalidAlgorithmException),
+                                new StackFrame(true)))
+                    },
+                    new AlgorithmTheoryData
```
iNinja, line 121: Can we add a test case for the Valid scenario where there's no `ValidAlgorithms` assigned to our `ValidationParameters`?
Just to highlight all the success/failure scenarios.

iNinja (changes requested):

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Algorithm.cs
FuPingFranco: Discussed offline, I'll make additional changes on a separate PR once, we have alignment on the overall design of the project.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Algorithm.cs
FuPingFranco: Discussed offline, I'll make additional changes on a separate PR once, we have alignment on the overall design of the project.

test/Microsoft.IdentityModel.Tokens.Tests/Validation/AlgorithmValidationResultTests.cs
FuPingFranco: Not sure I follow, if we have no `ValidAlgorithms` assigned to our `ValidationParameters` then, it would always be an invalid scenario and should fail so, the scenario above in line 99 should cover that right?

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Algorithm.cs
FuPingFranco: Updated, Thanks @iNinja !

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Algorithm.cs
```diff
@@ -42,5 +44,71 @@ public static void ValidateAlgorithm(string algorithm, SecurityKey securityKey,
...
+                        typeof(ArgumentNullException),
+                        new StackFrame(true)));
+            }
+
+            if (validationParameters.AlgorithmValidator != null)
+            {
+                if (!validationParameters.AlgorithmValidator(algorithm, securityKey, securityToken, validationParameters))
+                {
+                    return new AlgorithmValidationResult(
+                    algorithm,
```
keegan-caruso: nit: missing indent

src/Microsoft.IdentityModel.Tokens/Validation/AlgorithmValidationResult.cs
```diff
@@ -0,0 +1,70 @@
...
+                    securityTokenInvalidAlgorithmException.InvalidAlgorithm = Algorithm;
+                    securityTokenInvalidAlgorithmException.Source = TokenSource;
+                }
+
+                return _exception;
+            }
+        }
+
+        /// <summary>
+        /// Gets the security token algorithm used to sign the token.
```
brentschmaltz: The comment might read better as: Gets the algorithm used to sign the token.
Point is algorithms are applied to a number of objects.

test/Microsoft.IdentityModel.Tokens.Tests/Validation/AlgorithmValidationResultTests.cs
FuPingFranco: Added test case. Thanks Ignacio!

PR #2698, opened 2024-07-10
Additional Write API take a Stream

westin-m:
Allow users to provide a Stream to Write in OIDCConfigurationSerializer

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
@@ -82,7 +82,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         /// </summary>
         /// <param name="configuration"><see cref="OpenIdConnectConfiguration"/> object to serialize.</param>
         /// <returns>json string representing the configuration object.</returns>
-        /// <exception cref="ArgumentNullException">Thrown if <paramref name="configuration"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="configuration"/> is <see langword="null"/>.</exception>
         public static string Write(OpenIdConnectConfiguration configuration)
         {
             if (configuration == null)
@@ -95,12 +95,12 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
         }
 
         /// <summary>
-        /// Writes an <see cref="OpenIdConnectConfiguration"/> to a json string using the provided <see cref="Stream"/>.
+        /// Writes an <see cref="OpenIdConnectConfiguration"/> as JSON to the <paramref name="stream"/>.
         /// </summary>
         /// <param name="configuration">The <see cref="OpenIdConnectConfiguration"/> to serialize.</param>
         /// <param name="stream">The <see cref="Stream"/> to write to.</param>
         /// <remarks>Because a <see cref="Stream"/> is provided, this method does not return a value.</remarks>
-        /// <exception cref="ArgumentNullException">Thrown if <paramref name="configuration"/> or <paramref name="stream"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="configuration"/> or <paramref name="stream"/> is <see langword="null"/>.</exception>
         public static void Write(OpenIdConnectConfiguration configuration, Stream stream)
         {
             if (configuration == null)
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Json/OpenIdConnectConfigurationSerializer.cs
```diff
@@ -602,21 +602,44 @@ public static string Write(OpenIdConnectConfiguration OpenIdConnectConfiguration
...
-                }
-                finally
-                {
-                    writer?.Dispose();
-                }
+                return Write(OpenIdConnectConfiguration, memoryStream);
             }
         }
 
+        public static string Write(OpenIdConnectConfiguration OpenIdConnectConfiguration, Stream stream)
```
brentschmaltz: I would suggest writing this a bit different.
Instead of returning a string, return void.

The caller knows the stream type and will know how to get the buffer.
This will avoid the case when the Stream is not a MemoryStream and you have to copy the bytes.
It is most likely the caller wants the bytes, otherwise why pass the Stream?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Json/OpenIdConnectConfigurationSerializer.cs
```diff
@@ -617,6 +622,27 @@ public static string Write(OpenIdConnectConfiguration OpenIdConnectConfiguration
             }
         }
 
+        /// <summary>
+        /// Writes an <see cref="OpenIdConnectConfiguration"/> to a UTF8-encoded JSON string using the provided <see cref="Stream"/>.
+        /// </summary>
+        /// <param name="OpenIdConnectConfiguration">The <see cref="OpenIdConnectConfiguration"/> to be written.</param>
+        /// <param name="stream">The <see cref="Stream"/> to write to.</param>
+        /// <remarks>Because a <see cref="Stream"/> is provided, this method does not return a value.</remarks>
+        public static void Write(OpenIdConnectConfiguration OpenIdConnectConfiguration, Stream stream)
```
brentschmaltz, line 620: To prevent regressions, we should have a test for this API.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectConfigurationTests.cs
```diff
@@ -345,5 +364,25 @@ public void NonemptyCollectionSerialization()
             }
             TestUtilities.AssertFailIfErrors(context);
         }
+
+        [Fact]
+        public void NonemptyCollectionSerializationWithStream()
+        {
+            using (MemoryStream stream = new MemoryStream())
```
keegan-caruso: nit, you can do something like

using MemoryStream stream = new();

And avoid indenting the rest of the method here.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectConfigurationTests.cs
```diff
@@ -345,5 +364,25 @@ public void NonemptyCollectionSerialization()
             }
             TestUtilities.AssertFailIfErrors(context);
         }
+
+        [Fact]
+        public void NonemptyCollectionSerializationWithStream()
```
keegan-caruso, line 395: I think a "round trip" test is useful here. Write to a stream, read it, create the object, compare the properties.

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
```diff
@@ -93,6 +94,27 @@ public static string Write(OpenIdConnectConfiguration configuration)
             return OpenIdConnectConfigurationSerializer.Write(configuration);
         }
 
+        /// <summary>
+        /// Writes an <see cref="OpenIdConnectConfiguration"/> to a json string using the provided <see cref="Stream"/>.
```
keegan-caruso: This wording confuses me.

```cs
/// Writes <see cref="OpenIdConnectConfiguration"/> as JSON to the <paramref "name"="stream">
```

maybe?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfiguration.cs
```diff
@@ -93,6 +94,27 @@ public static string Write(OpenIdConnectConfiguration configuration)
...
+        /// </summary>
+        /// <param name="configuration">The <see cref="OpenIdConnectConfiguration"/> to serialize.</param>
+        /// <param name="stream">The <see cref="Stream"/> to write to.</param>
+        /// <remarks>Because a <see cref="Stream"/> is provided, this method does not return a value.</remarks>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="configuration"/> or <paramref name="stream"/> is null.</exception>
```
keegan-caruso: nit: use a langref to refer to null

keegan-caruso (approved):

PR #2700, opened 2024-07-11
Add CaseSensitiveClaimsIdentity type.

pmaytak:
A derived `ClaimsIdentity` where claim retrieval is case-sensitive. The current `ClaimsIdentity` retrieves claims in a case-insensitive manner which is different than querying the underlying `SecurityToken`. The `CaseSensitiveClaimsIdentity` provides consistent retrieval logic between the `SecurityToken` and `ClaimsIdentity`.


src/Microsoft.IdentityModel.Tokens/CaseSensitiveClaimsIdentity.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <summary>
+        /// Retrieves a <see cref="IEnumerable{Claim}"/> where each Claim.Type equals <paramref name="type"/>.
+        /// </summary>
+        /// <param name="type">The type of the claim to match.</param>
+        /// <returns>A <see cref="IEnumerable{Claim}"/> of matched claims.</returns>
+        /// <remarks>Comparison is: StringComparison.Ordinal.</remarks>
+        /// <exception cref="ArgumentNullException">if <paramref name="type"/> is null.</exception>
+        public override IEnumerable<Claim> FindAll(string type)
+        {
+            return base.FindAll(claim => claim?.Type.Equals(type, StringComparison.Ordinal) == true);
```
jmprieur, line 92: do we need the `== true` ?

pmaytak: Yes. Claim can be null, so expression results in `bool?`.
`CS0266	Cannot implicitly convert type 'bool?' to 'bool'. An explicit conversion exists (are you missing a cast?)`

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// AppContext switches for Microsoft.IdentityModel.Tokens and referencing packages.
+    /// </summary>
+    public static class AppContextSwitches
```
keegan-caruso: Either the class should also be internal or the consts should also be public

src/Microsoft.IdentityModel.Tokens/CaseSensitiveClaimsIdentity.cs
```diff
@@ -0,0 +1,109 @@
...
+using System;
+using System.Collections.Generic;
+using System.Security.Claims;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// A derived <see cref="ClaimsIdentity"/> where claim retrieval is case-sensitive. The current <see cref="ClaimsIdentity"/> retrieves claims in a case-insensitive manner which is different than querying the underlying <see cref="SecurityToken"/>. The <see cref="CaseSensitiveClaimsIdentity"/> provides consistent retrieval logic between the <see cref="SecurityToken"/> and <see cref="ClaimsIdentity"/>.
+    /// </summary>
+    public class CaseSensitiveClaimsIdentity : ClaimsIdentity
```
GeoK, line 13: Are we planning to add a SecurityToken property?

test/Microsoft.IdentityModel.Tokens.Tests/CaseSensitiveClaimsIdentityTests.cs
```diff
@@ -0,0 +1,242 @@
...
+        private static string CreateUnsignedToken(JObject payload)
+        {
+            // Add default claims to the beginning of the payload
+            var jObject = new JObject(_defaultPayload);
+            jObject.Merge(payload);
+            return string.Concat(Base64UrlEncoder.Encode("{}"), ".", Base64UrlEncoder.Encode(jObject.ToString()), ".");
+        }
+    }
+}
+}
```
msbw2:
```suggestion
```

src/Microsoft.IdentityModel.Tokens/Validation/TokenValidationResult.cs
```diff
@@ -136,6 +136,8 @@ internal ClaimsIdentity ClaimsIdentityNoLocking
                     if (_validationParameters != null && SecurityToken != null && _tokenHandler != null && Issuer != null)
                     {
                         _claimsIdentity = _tokenHandler.CreateClaimsIdentityInternal(SecurityToken, _validationParameters, Issuer);
+                        if (_claimsIdentity is not CaseSensitiveClaimsIdentity && !AppContextSwitches.UseClaimsIdentityType())
```
msbw2: Why do you want to keep the above line that creates the potentially case-insensitive version? Is there a side effect you'd like to retain? Everywhere else you've just replaced the relevant call.

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
```diff
@@ -125,6 +124,18 @@ internal virtual ClaimsIdentity CreateClaimsIdentityInternal(SecurityToken secur
                         MarkAsNonPII("internal virtual ClaimsIdentity CreateClaimsIdentityInternal(SecurityToken securityToken, TokenValidationParameters tokenValidationParameters, string issuer)"),
                         MarkAsNonPII(GetType().FullName))));
         }
+
+        internal static ClaimsIdentity CreateCaseSensitiveClaimsIdentityFromTokenValidationParameters(SecurityToken securityToken, TokenValidationParameters validationParameters, string issuer)
+        {
+            ClaimsIdentity identity = validationParameters.CreateClaimsIdentity(securityToken, issuer);
```
msbw2: Why do you want to attempt to create from `validationParameters` and possibly end up creating a case-insensitive version? Just side effects?

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
```diff
@@ -125,6 +124,18 @@ internal virtual ClaimsIdentity CreateClaimsIdentityInternal(SecurityToken secur
                         MarkAsNonPII("internal virtual ClaimsIdentity CreateClaimsIdentityInternal(SecurityToken securityToken, TokenValidationParameters tokenValidationParameters, string issuer)"),
                         MarkAsNonPII(GetType().FullName))));
         }
+
+        internal static ClaimsIdentity CreateCaseSensitiveClaimsIdentityFromTokenValidationParameters(SecurityToken securityToken, TokenValidationParameters validationParameters, string issuer)
```
msbw2: This method name is slightly misleading since it won't do what it indicates if the switch is on.

src/Microsoft.IdentityModel.Tokens/CaseSensitiveClaimsIdentity.cs
pmaytak: Added https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2700/commits/c4bbcf6642c1c32c93aa5e692897630b1f38c82b

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
pmaytak: CreateClaimsIdentity can be overriden in derived TVP by devs (extra claims could be added, for ex.) We're honoring that here but also changing the type to CsClaimsIdentity if needed.

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
pmaytak: Indeed, will update.

src/Microsoft.IdentityModel.Tokens/Validation/TokenValidationResult.cs
pmaytak: Same behavior as with TVP.CreateClaimsIdentity; `_tokenHandler` can be derived and return normal ClaimsIdentity, which we'd honor. Although, we may not need this check here, since `_claimsIdentity` here will always probably be CsClaimsIdentity when switch is set.

src/Microsoft.IdentityModel.Tokens/TokenHandler.cs
msbw2: Move this into the factory.

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -240,7 +240,13 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
             if (LogHelper.IsEnabled(EventLogLevel.Informational))
                 LogHelper.LogInformation(LogMessages.IDX10245, securityToken);
 
-            return new ClaimsIdentity(authenticationType: AuthenticationType ?? DefaultAuthenticationType, nameType: nameClaimType ?? ClaimsIdentity.DefaultNameClaimType, roleType: roleClaimType ?? ClaimsIdentity.DefaultRoleClaimType);
+            if (AppContextSwitches.UseClaimsIdentityType())
```
msbw2: I think all this can be moved to the factory. Basically this switch check should only have one usage in prod code: in the factory.

src/Microsoft.IdentityModel.Tokens/ClaimsIdentityFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+            };
+        }
+
+        internal static ClaimsIdentity Create(SecurityToken securityToken, TokenValidationParameters validationParameters, string issuer)
+        {
+            ClaimsIdentity claimsIdentity = validationParameters.CreateClaimsIdentity(securityToken, issuer);
+
+            if (claimsIdentity is not CaseSensitiveClaimsIdentity && !AppContextSwitches.UseClaimsIdentityType())
+            {
+                claimsIdentity = new CaseSensitiveClaimsIdentity(claimsIdentity);
```
msbw2: Since `ClaimsIdentity` doesn't have the `SecurityToken` field it isn't copied over here. You need to set it explicitly here and in the method below.

pmaytak: Done. Will add more tests.

jennyf19 (approved):
:shipit:

test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlSecurityTokenHandlerTests.cs
```diff
@@ -120,7 +120,11 @@ public static TheoryData<SamlTheoryData> CanReadTokenTheoryData
         public void CreateClaimsIdentities(SamlTheoryData theoryData)
         {
             TestUtilities.WriteHeader($"{this}.CreateClaimsIdentities", theoryData);
-            var context = new CompareContext($"{this}.CreateClaimsIdentities, {theoryData.TestId}") { IgnoreType = true };
+            var context = new CompareContext($"{this}.CreateClaimsIdentities, {theoryData.TestId}")
+            {
+                IgnoreType = true,
```
keegan-caruso: nit: is this still needed?

keegan-caruso (approved):

src/Microsoft.IdentityModel.Tokens/CaseSensitiveClaimsIdentity.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        /// <summary>
+        /// Gets the <see cref="SecurityToken"/> that was used to create this claims identity.
```
pmaytak:
```suggestion
        /// Gets the <see cref="SecurityToken"/> associated with this claims identity.
```

test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlSecurityTokenHandlerTests.cs
pmaytak: Nope. removed.

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
sergey-v9: Is there no option to use this switch in the user code since it is internal now?
The only remaining way to use it is to pass the value explicitly, like this:
```csharp
AppContext.SetSwitch("Microsoft.IdentityModel.Tokens.UseClaimsIdentityType", true);
```

pmaytak: Yes, that was the current behavior that we didn't change. They're internal to avoid new dependencies on them so that eventually the switches can be removed in later versions. This class is meant to only be used internally for ease of use as long as the switch is needed, and users can reference this file for what switches are supported.

PR #2705, opened 2024-07-12
Create DistributedMetadataCacheOptions

westin-m:
Create new Options class and new interface for transforming byte arrays before/after caching


src/Microsoft.IdentityModel.Protocols/Configuration/DistributedMetadataCacheOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.Extensions.Caching.Distributed;
+
+namespace Microsoft.IdentityModel.Protocols.Configuration
+{
+    /// <summary>
+    /// Specifies the DistributedMetadataCacheOptions which can be used to configure the distributed metadata cache.
+    /// </summary>
+    public class DistributedMetadataCacheOptions
```
keegan-caruso, line 11: should this be public at this point?

PR #2716, opened 2024-07-15
Call TVP.CreateClaimsIdentity to support users that have overloaded.

brentschmaltz:
Call TVP.CreateClaimsIdentity to maintain back compat for users that have overloaded TVP.CreateClaimsIdentity.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -212,7 +212,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         {
             _ = validationParameters ?? throw LogHelper.LogArgumentNullException(nameof(validationParameters));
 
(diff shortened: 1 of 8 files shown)


keegan-caruso:
We should remove these lines as well to not change type:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/ClaimsIdentityFactory.cs#L50-L56

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/ClaimsIdentityFactory.cs#L70-L76

brentschmaltz:
@keegan-caruso I haven't analyzed that, it is good that you called it out.

brentschmaltz:
@keegan-caruso those methods are only used by tests, i changed the tests and removed the methods.

pmaytak (approved):

keegan-caruso (approved):

PR #2719, opened 2024-07-16
Algorithm validation: Remove exceptions

FuPingFranco:
# Algorithm validation:  Remove exceptions

## Description

- Implemented new (internal momentarily) method for algorithm validation which replaces exception throwing with a return of the new AlgorithmValidationResult containing the exception information.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Validators.Algorithm.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Validators.Algorithm.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Diagnostics;
 using System.Linq;
 using Microsoft.IdentityModel.Logging;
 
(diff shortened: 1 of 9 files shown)


src/Microsoft.IdentityModel.Tokens/Validation/Validators.Algorithm.cs
```diff
@@ -2,12 +2,60 @@
...
+                    ValidationFailureType.NullArgument,
+                    new ExceptionDetail(
+                        new MessageDetail(
+                            LogMessages.IDX10000,
+                            LogHelper.MarkAsNonPII(nameof(validationParameters))),
+                        typeof(ArgumentNullException),
+                        new StackFrame(true)));
+            }
+
+            if (validationParameters.ValidAlgorithms != null && validationParameters.ValidAlgorithms.Any() && !validationParameters.ValidAlgorithms.Contains(algorithm, StringComparer.Ordinal))
```
iNinja: Since we changed `ValidAlgorithms` to be an `IList`, we can now check `validationParameters.ValidAlgorithms.Count > 0` and avoid dealing with enumerators.

FuPingFranco: Good catch, updated. Thanks @iNinja !

iNinja (approved):
LGTM

PR #2724, opened 2024-07-16
Cache context switches

### LINKED ISSUE #2722: [Bug] Cache values of AppContext switches
`AppContext.TryGetSwitch` [statically caches internally](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/AppContext.cs,0a1f341850c88646) but takes out a lock. 
.NET almost always [caches these values](https://github.com/dotnet/aspnetcore/blob/79f745dfd906db54916bf3da2430720eaeda6254/src/Servers/Kestrel/Core/src/KestrelServerOptions.cs#L34-L38). They're not expected to change while the process is running unlike normal config.

Currently we don't cache values. 

To-do:
- Update all `AppContext.TryGetSwitch` to statically cache the value.
- See if moving all context switches into [one class](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs) for easier management would be helpful.
- Run perf tests to compare difference.

pmaytak:
Fixes #2722.

- Caches context switch values.
- Moves all context switches into one file.
- Adds a method reset all context switches for testing.

Perf results for validation and claims creation which ends up hitting two switches: UseClaimsIdentityTypeSwitch, TryAllStringClaimsAsDateTimeSwitch.

Without caching:

| Method                                              | Mean     | Error    | StdDev   | P90      | P95      | P100     | Gen0   | Allocated |
|---------------------------------------------------- |---------:|---------:|---------:|---------:|---------:|---------:|-------:|----------:|
| JsonWebTokenHandler_ValidateTokenAsync_CreateClaims | 36.41 μs | 0.136 μs | 0.299 μs | 36.84 μs | 36.93 μs | 37.06 μs | 0.9766 |  16.52 KB |

With caching:

| Method                                              | Mean     | Error    | StdDev   | P90      | P95      | P100     | Gen0   | Allocated |
|---------------------------------------------------- |---------:|---------:|---------:|---------:|---------:|---------:|-------:|----------:|
| JsonWebTokenHandler_ValidateTokenAsync_CreateClaims | 35.13 μs | 0.047 μs | 0.100 μs | 35.26 μs | 35.28 μs | 35.34 μs | 0.9766 |  16.52 KB |

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
@@ -21,15 +21,6 @@ namespace Microsoft.IdentityModel.Tokens
         internal static bool UseClaimsIdentityType => _useClaimsIdentity ??= (AppContext.TryGetSwitch(UseClaimsIdentityTypeSwitch, out bool useClaimsIdentityType) && useClaimsIdentityType);
 
         /// <summary>
-        /// Used for testing to reset the <see cref="UseClaimsIdentityType"/> switch to its default value.
-        /// </summary>
-        internal static void ResetUseClaimsIdentityTypeSwitch()
-        {
-            _useClaimsIdentity = null;
-            AppContext.SetSwitch(UseClaimsIdentityTypeSwitch, false);
-        }
-
-        /// <summary>
         /// When validating the issuer signing key, specifies whether to fail if the 'tid' claim is missing.
         /// </summary>
         internal const string DoNotFailOnMissingTidSwitch = "Switch.Microsoft.IdentityModel.DontFailOnMissingTidValidateIssuerSigning";
@@ -37,15 +28,6 @@ namespace Microsoft.IdentityModel.Tokens
         private static bool? _doNotFailOnMissingTid;
 
         internal static bool DontFailOnMissingTid => _doNotFailOnMissingTid ??= (AppContext.TryGetSwitch(DoNotFailOnMissingTidSwitch, out bool doNotFailOnMissingTid) && doNotFailOnMissingTid);
-
-        /// <summary>
-        /// Used for testing to reset the <see cref="UseClaimsIdentityType"/> switch to its default value.
-        /// </summary>
-        internal static void ResetDoNotFailOnMissingTidSwitch()
-        {
-            _doNotFailOnMissingTid = null;
-            AppContext.SetSwitch(DoNotFailOnMissingTidSwitch, false);
-        }
 
         /// <summary>
         /// When reading claims from the token, specifies whether to try to convert all string claims to DateTime.
@@ -58,10 +40,16 @@ namespace Microsoft.IdentityModel.Tokens
         internal static bool TryAllStringClaimsAsDateTime => _tryAllStringClaimsAsDateTime ??= (AppContext.TryGetSwitch(TryAllStringClaimsAsDateTimeSwitch, out bool tryAsDateTime) && tryAsDateTime);
 
         /// <summary>
-        /// Used for testing to reset the <see cref="UseClaimsIdentityType"/> switch to its default value.
+        /// Used for testing to reset all switches to its default value.
         /// </summary>
-        internal static void ResetTryAllStringClaimsAsDateTimeSwitch()
+        internal static void ResetAllSwitches()
         {
+            _useClaimsIdentity = null;
+            AppContext.SetSwitch(UseClaimsIdentityTypeSwitch, false);
+
+            _doNotFailOnMissingTid = null;
+            AppContext.SetSwitch(DoNotFailOnMissingTidSwitch, false);
+
             _tryAllStringClaimsAsDateTime = null;
             AppContext.SetSwitch(TryAllStringClaimsAsDateTimeSwitch, false);
         }
--- a/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
@@ -14,41 +14,56 @@ namespace Microsoft.IdentityModel.Tokens
         /// <summary>
         /// Enables a fallback to the previous behavior of using <see cref="ClaimsIdentity"/> instead of <see cref="CaseSensitiveClaimsIdentity"/> globally.
         /// </summary>
-        private const string UseClaimsIdentityTypeSwitch = "Microsoft.IdentityModel.Tokens.UseClaimsIdentityType";
+        internal const string UseClaimsIdentityTypeSwitch = "Microsoft.IdentityModel.Tokens.UseClaimsIdentityType";
 
         private static bool? _useClaimsIdentity;
 
-        internal static bool UseClaimsIdentityType
+        internal static bool UseClaimsIdentityType => _useClaimsIdentity ??= (AppContext.TryGetSwitch(UseClaimsIdentityTypeSwitch, out bool useClaimsIdentityType) && useClaimsIdentityType);
+
+        /// <summary>
+        /// Used for testing to reset the <see cref="UseClaimsIdentityType"/> switch to its default value.
+        /// </summary>
+        internal static void ResetUseClaimsIdentityTypeSwitch()
         {
-            get => _useClaimsIdentity ??= (AppContext.TryGetSwitch(UseClaimsIdentityTypeSwitch, out bool useClaimsIdentityType) && useClaimsIdentityType);
-            set => _useClaimsIdentity = value;
+            _useClaimsIdentity = null;
+            AppContext.SetSwitch(UseClaimsIdentityTypeSwitch, false);
         }
 
         /// <summary>
         /// When validating the issuer signing key, specifies whether to fail if the 'tid' claim is missing.
         /// </summary>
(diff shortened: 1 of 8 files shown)


keegan-caruso (approved):

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -14,8 +14,41 @@ internal static class AppContextSwitches
...
-        internal const string UseClaimsIdentityTypeSwitch = "Microsoft.IdentityModel.Tokens.UseClaimsIdentityType";
+        private const string UseClaimsIdentityTypeSwitch = "Microsoft.IdentityModel.Tokens.UseClaimsIdentityType";
 
-        internal static bool UseClaimsIdentityType() => (AppContext.TryGetSwitch(UseClaimsIdentityTypeSwitch, out bool useClaimsIdentityType) && useClaimsIdentityType);
+        private static bool? _useClaimsIdentity;
+
+        internal static bool UseClaimsIdentityType
+        {
+            get => _useClaimsIdentity ??= (AppContext.TryGetSwitch(UseClaimsIdentityTypeSwitch, out bool useClaimsIdentityType) && useClaimsIdentityType);
+            set => _useClaimsIdentity = value;
```
brentschmaltz: Is the set used just for test?

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -14,8 +14,41 @@ internal static class AppContextSwitches
...
+        }
+
+        /// <summary>
+        /// When validating the issuer signing key, specifies whether to fail if the 'tid' claim is missing.
+        /// </summary>
+        private const string DontFailOnMissingTidSwitch = "Switch.Microsoft.IdentityModel.DontFailOnMissingTidValidateIssuerSigning";
```
brentschmaltz: May want to name this 'DoNotFail...' rather than 'DontFail...'.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerClaimsIdentityTests.cs
```diff
@@ -41,7 +41,7 @@ public void CreateClaimsIdentity_ReturnsCaseSensitveClaimsIdentity_ByDefault()
         [Fact]
         public void CreateClaimsIdentity_ReturnsClaimsIdentity_WithAppContextSwitch()
         {
-            AppContext.SetSwitch(AppContextSwitches.UseClaimsIdentityTypeSwitch, true);
+            AppContextSwitches.UseClaimsIdentityType = true;
```
brentschmaltz: Are we going to have a race condition with public void Create_FromTokenValidationParameters_ReturnsCorrectClaimsIdentity(bool useClaimsIdentity)?

Or some other place?

keegan-caruso: it is part of a test collection so it is serialized.

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
pmaytak: It was; removed it; using a different way to reset.

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -16,6 +16,54 @@ internal static class AppContextSwitches
...
+        internal static void ResetUseClaimsIdentityTypeSwitch()
+        {
+            _useClaimsIdentity = null;
+            AppContext.SetSwitch(UseClaimsIdentityTypeSwitch, false);
+        }
+
+        /// <summary>
+        /// When validating the issuer signing key, specifies whether to fail if the 'tid' claim is missing.
+        /// </summary>
+        internal const string DontFailOnMissingTidSwitch = "Switch.Microsoft.IdentityModel.DontFailOnMissingTidValidateIssuerSigning";
```
jennyf19: thought Brent had a comment to use "DoNot" instead of "Dont" somewhere?

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -14,41 +14,56 @@ internal static class AppContextSwitches
...
 
         private static bool? _useClaimsIdentity;
 
-        internal static bool UseClaimsIdentityType
+        internal static bool UseClaimsIdentityType => _useClaimsIdentity ??= (AppContext.TryGetSwitch(UseClaimsIdentityTypeSwitch, out bool useClaimsIdentityType) && useClaimsIdentityType);
+
+        /// <summary>
+        /// Used for testing to reset the <see cref="UseClaimsIdentityType"/> switch to its default value.
+        /// </summary>
+        internal static void ResetUseClaimsIdentityTypeSwitch()
```
keegan-caruso: nit: we could just null them all out in one method.

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
pmaytak: Done.

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -14,41 +14,56 @@ internal static class AppContextSwitches
...
         {
-            get => _useClaimsIdentity ??= (AppContext.TryGetSwitch(UseClaimsIdentityTypeSwitch, out bool useClaimsIdentityType) && useClaimsIdentityType);
-            set => _useClaimsIdentity = value;
+            _useClaimsIdentity = null;
+            AppContext.SetSwitch(UseClaimsIdentityTypeSwitch, false);
```
keegan-caruso: why do we set a value in the reset?

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
pmaytak: Yep, done, variable names changed. The string is left the same, though.

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
pmaytak: Refactored: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2724/commits/036811e9ba4ee1c67108e573f381a658a357aab9

keegan-caruso (approved):

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -16,6 +16,42 @@ internal static class AppContextSwitches
         /// </summary>
         internal const string UseClaimsIdentityTypeSwitch = "Microsoft.IdentityModel.Tokens.UseClaimsIdentityType";
 
-        internal static bool UseClaimsIdentityType() => (AppContext.TryGetSwitch(UseClaimsIdentityTypeSwitch, out bool useClaimsIdentityType) && useClaimsIdentityType);
+        private static bool? _useClaimsIdentity;
+
+        internal static bool UseClaimsIdentityType => _useClaimsIdentity ??= (AppContext.TryGetSwitch(UseClaimsIdentityTypeSwitch, out bool useClaimsIdentityType) && useClaimsIdentityType);
```
eerhardt, line 21: I wonder if we should use a simpler pattern here to ensure that tiered JIT can recognize and elide the branches. In general, if the tiered JIT can see that a static bool will always be `true` or `false`, it can take out branches resulting in faster code. I'm not sure if the JIT can recognize this pattern with a nullable boolean as immutable, since the field isn't readonly.

One pattern looks like:

```csharp
 internal static bool UseClaimsIdentityType { get; } = AppContext.TryGetSwitch(UseClaimsIdentityTypeSwitch, out bool useClaimsIdentityType) && useClaimsIdentityType;
 ```

This makes a readonly field and a readonly property that simply returns the field value. The initialization happens in a static constructor, and after that the value will never change. So if the code gets re-JIT'd, the JIT will recognize the value as always `true` or `false`.

See https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-8/#tiering-and-dynamic-pgo for a good description of how tiering works.

pmaytak: How can this approach be used in tests? Currently, we reset/null `_useClaimsIdentity`, so it is refreshes from the switch value.

eerhardt: In dotnet/runtime we use something called the [RemoteExecutor](https://github.com/dotnet/arcade/tree/main/src/Microsoft.DotNet.RemoteExecutor) which spins up a new process with the desired AppContext switches, and then tests inside that process.

Example: https://github.com/dotnet/runtime/blob/4f38f921ddeb6773d1e8316298215d84dc643cdd/src/libraries/System.Threading.ThreadPool/tests/ThreadPoolTests.cs#L198-L211

PR #2725, opened 2024-07-16
Remove Delegate Checks in Multiple Validators and Prevents Null Setting of Delegates

FuPingFranco:
# Remove Delegate Checks in Multiple Validators and Prevents Null Setting of Delegates.

## Description

- Removed delegate checks from our default token type, lifetime and token replay validation.
- Updated unit tests and added new tests for ValidationParameters class.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Security.Claims;
+using System.Threading;
 using Microsoft.IdentityModel.Abstractions;
 using Microsoft.IdentityModel.Logging;
 
@@ -20,8 +21,13 @@ namespace Microsoft.IdentityModel.Tokens
         private string _nameClaimType = ClaimsIdentity.DefaultNameClaimType;
         private string _roleClaimType = ClaimsIdentity.DefaultRoleClaimType;
         private Dictionary<string, object> _instancePropertyBag;
+        private IList<string> _validTokenTypes = [];
 
-        private IssuerValidationDelegateAsync _issuerValidationDelegate = Validators.ValidateIssuerAsync;
+        private AudienceValidatorDelegate _audienceValidator = Validators.ValidateAudience;
+        private IssuerValidationDelegateAsync _issuerValidatorAsync = Validators.ValidateIssuerAsync;
+        private LifetimeValidatorDelegate _lifetimeValidator = Validators.ValidateLifetime;
+        private TypeValidatorDelegate _typeValidator = Validators.ValidateTokenType;
+        private TokenReplayValidatorDelegate _tokenReplayValidator = Validators.ValidateTokenReplay;
 
         /// <summary>
         /// This is the default value of <see cref="ClaimsIdentity.AuthenticationType"/> when creating a <see cref="ClaimsIdentity"/>.
@@ -112,14 +118,25 @@ namespace Microsoft.IdentityModel.Tokens
         public AlgorithmValidator AlgorithmValidator { get; set; }
 
         /// <summary>
-        /// Gets or sets a delegate that will be used to validate the audience.
+        /// Allows overriding the delegate that will be used to validate the audience.
         /// </summary>
         /// <remarks>
-        /// If set, this delegate will be called to validate the 'audience', instead of default processing.
+        /// If set, this delegate will be responsible for validating the 'audience', instead of default processing.
         /// This means that no default 'audience' validation will occur.
-        /// Even if <see cref="ValidateAudience"/> is false, this delegate will still be called.
         /// </remarks>
-        public AudienceValidator AudienceValidator { get; set; }
+        /// <exception cref="ArgumentNullException">Thrown when the value is set as null.</exception>
+        /// <returns>The <see cref="AudienceValidatorDelegate"/> used to validate the issuer of a token</returns>
+        public AudienceValidatorDelegate AudienceValidator
+        {
+            get
+            {
+                return _audienceValidator;
+            }
+            set
+            {
+                _audienceValidator = value ?? throw new ArgumentNullException(nameof(value), "AudienceValidator cannot be set as null.");
+            }
+        }
 
         /// <summary>
         /// Gets or sets the AuthenticationType when creating a <see cref="ClaimsIdentity"/>.
@@ -278,17 +295,19 @@ namespace Microsoft.IdentityModel.Tokens
         public IList<SecurityKey> IssuerSigningKeys { get; }
 
         /// <summary>
-        /// Gets or sets a delegate that will be used to validate the issuer of the token.
+        /// Allows overriding the delegate that will be used to validate the issuer of the token.
         /// </summary>
+        /// <exception cref="ArgumentNullException">Thrown when the value is set as null.</exception>
+        /// <returns>The <see cref="IssuerValidationDelegateAsync"/> used to validate the issuer of a token</returns>
         public IssuerValidationDelegateAsync IssuerValidatorAsync
         {
             get
             {
-                return _issuerValidationDelegate;
+                return _issuerValidatorAsync;
             }
             set
             {
-                _issuerValidationDelegate = value ?? throw LogHelper.LogArgumentNullException(nameof(value));
+                _issuerValidatorAsync = value ?? throw new ArgumentNullException(nameof(value), "IssuerValidatorAsync cannot be set as null.");
             }
         }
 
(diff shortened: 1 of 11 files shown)


src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -440,16 +440,16 @@ public string RoleClaimType
...
         /// Implementations that need to resolve it from a different location can use the 'token' parameter.
         /// </summary>
         /// <remarks>
-        /// If set, this delegate will be called to validate the 'type' of the token, instead of default processing.
-        /// This means that no default 'type' validation will occur.
+        /// If no delegate is set, the default implementation will be used. The default checks the type
+        /// against the <see cref="ValidTypes"/> property, if it's present then, it will succeed.
         /// </remarks>
-        public TypeValidator TypeValidator { get; set; }
+        public TypeValidatorDelegate TypeValidator { get; set; } = Validators.ValidateTokenType;
```
brentschmaltz: We should not allow this to be set to null.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -491,12 +491,12 @@ public string RoleClaimType
...
 
         /// <summary>
-        /// Gets the <see cref="IList{String}"/> that contains valid types that will be used to check against the JWT header's 'typ' claim.
+        /// Gets or sets the <see cref="IList{String}"/> that contains valid types that will be used to check against the JWT header's 'typ' claim.
         /// If this property is not set, the 'typ' header claim will not be validated and all types will be accepted.
         /// In the case of a JWE, this property will ONLY apply to the inner token header.
         /// The default is <c>null</c>.
         /// </summary>
-        public IList<string> ValidTypes { get; }
+        public IList<string> ValidTypes { get; set; }
```
brentschmaltz: We should not allow this to be set to null.
Use this pattern: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/ecc2b0813a57da798d361c2edd5e2ae38cfd71f7/src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs#L28

src/Microsoft.IdentityModel.Tokens/Validation/Validators.TokenType.cs
```diff
@@ -54,19 +70,12 @@ internal static TokenTypeValidationResult ValidateTokenType(string? type, Securi
                         new StackFrame(true)));
             }
 
-            if (validationParameters.TypeValidator == null && (validationParameters.ValidTypes == null || !validationParameters.ValidTypes.Any()))
+            if (validationParameters.ValidTypes == null || validationParameters.ValidTypes.Count == 0)
```
brentschmaltz: If ValidationParameters.ValidTypes cannot be null, then the null check is not needed.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
FuPingFranco: Is thread safety a concern when validating Token types? I wondering if this is enough to prevent a null setting

private IList<string> _validTypes = [];

public IList<string> ValidTypes
{
    get => _validTypes;
    set => _validTypes = value ?? [];
}

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
FuPingFranco: How's the approach below? although it is somewhat of an anti-pattern to throw an exception. Let's discuss this offline. 

private TypeValidatorDelegate _typeValidator = Validators.ValidateTokenType;

public TypeValidatorDelegate TypeValidator
{
    get { return _typeValidator; }
    set
    {
        if (value == null)
        {
            throw new ArgumentNullException(nameof(value), "TypeValidator cannot be set as null.");
        }
        _typeValidator = value;
    }
}

brentschmaltz: yes, we sometimes write this as
```cs
_typeValidator = value ?? throw new ArgumentNullException(nameof(value), "TypeValidator cannot be set as null.");
```

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
brentschmaltz: We use Interlocked.CompareExchange so that in multiple threads will have the same object.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
brentschmaltz: We should have a single test showing that setting null faults.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.TokenType.cs
FuPingFranco: Removed, thanks @brentschmaltz!

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
FuPingFranco: Updated, thanks!

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
FuPingFranco: Added test on separate file and updated code to prevent null setting of delegates. Thanks!

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -112,14 +118,25 @@ public ValidationParameters()
         public AlgorithmValidator AlgorithmValidator { get; set; }
 
         /// <summary>
-        /// Gets or sets a delegate that will be used to validate the audience.
+        /// Allows overriding the delegate that will be used to validate the audience.
         /// </summary>
         /// <remarks>
-        /// If set, this delegate will be called to validate the 'audience', instead of default processing.
+        /// If set, this delegate will be responsible for validating the 'audience', instead of default processing.
```
iNinja, line 124: We need to update the wording around `if set`, since it will always be set at least to our default.
I am okay doing this after this PR, though, as I need this merged for JWS/JWE validation.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -112,14 +118,25 @@ public ValidationParameters()
...
         /// This means that no default 'audience' validation will occur.
-        /// Even if <see cref="ValidateAudience"/> is false, this delegate will still be called.
         /// </remarks>
-        public AudienceValidator AudienceValidator { get; set; }
+        /// <exception cref="ArgumentNullException">Thrown when the value is set as null.</exception>
```
iNinja, line 127: Same with the wording around exceptions. They are not thrown but returned as part of the result.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -494,9 +550,16 @@ public string RoleClaimType
...
         /// In the case of a JWE, this property will ONLY apply to the inner token header.
-        /// The default is <c>null</c>.
+        /// The default is an empty collection.
         /// </summary>
-        public IList<string> ValidTypes { get; }
+        /// <exception cref="ArgumentNullException">Thrown when the value is set as null.</exception>
+        /// <returns>The <see cref="IList{String}"/> that contains valid token types that will be used to check against the token's 'typ' claim.</returns>
+        public IList<string> ValidTypes
+        {
+            get => _validTokenTypes ?? Interlocked.CompareExchange(ref _validTokenTypes, new List<string>(), null);
```
iNinja: `_validTokenTypes` is never null, as you are assigning `[]` at the beginning of the file and throwing if someone attempts to assign null. This could be simplified.

iNinja (approved):
Minor comments, other than that LGTM.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
FuPingFranco: Will update on the next pass for Audience validator then.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
FuPingFranco: Will update as part of the validator.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
FuPingFranco: Updated. Thank you.

MZOLN (approved):

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -112,14 +118,25 @@ public ValidationParameters()
...
+        /// <returns>The <see cref="AudienceValidatorDelegate"/> used to validate the issuer of a token</returns>
+        public AudienceValidatorDelegate AudienceValidator
+        {
+            get
+            {
+                return _audienceValidator;
+            }
+            set
+            {
+                _audienceValidator = value ?? throw new ArgumentNullException(nameof(value), "AudienceValidator cannot be set as null.");
```
alexholub113, line 137: consider to use nameof in exception message
`nameof(ValidationParameters.AudienceValidator)}`

Issue #2711, opened 2024-07-15
IdentityModel Validation model change


jennyf19:
IdentityModel is responsible for validating SecurityTokens. Validating a SecurityToken requires validating multiple parts and reporting the results. Common parts to validate are the issuer, audience, expiration. Default validation is included for important parts of the SecurityToken. The current model provides extensibility using delegates for validation. 

## Issues we want to address:

- Simplify the extensibility model by reducing options.
- Remove exceptions as the model for control.
- Current validation and delegates throw exceptions on errors. 
- Current validation logs errors by default.
- Callers will receive a complete description of how the validation occurred.

## Proposal 

- Define a set of objects that return the results of a validation step.
- Define a new set of delegates that are async and return a result. 

Each validation step will return a specialized ValidationResult type that contain details that will provide upper layers to the examine errors with contain exception details, log details with a stacktrace that can be thrown or logged. 

Related PRs: #2709, #2688, #2679, #2672, #2671, #2669, #2655

jennyf19:
# New token validation model
The existing token validation model on `Microsoft.IdentityModel` relies on the exception based control flow dominant in C#. In some high-performance scenarios where latency is key, this can cause unintended increases when validations fail and throw.  

As part of an effort to modernize aspects of the library that have become bloated over the years, we are introducing an alternative token validation model that breaks away from some of the previous behaviors and intends to offer a slimmer and more performant API.  

## Key points on the new model
- Performance increase by removing unnecessary exception throwing within the validation itself.
- Return a result where it is clear whether the validation was successful, including relevant information depending on the case.
  - In success scenarios, provide the validated information for audit purposes.
  - In failure scenarios, provide enough information to identify what part of the token was invalid or could not be validated without the need for further steps, and for diagnostic purposes provide the ability to create an exception that can be thrown and observed. 
- Do not automatically print logs as part of the validation. These can be printed from the validation result using C#'s [high-performance logging](https://learn.microsoft.com/en-us/dotnet/core/extensions/high-performance-logging) based on `ILogger`.
- Simpler validation parameters object, as the current one has grown in complexity over the years by including multiple parameters for things like `ValidIssuer` and `ValidIssuers`, or multiple delegates for the same validation. `ValidationParameters` is being introduced as an alternative to `TokenValidationParameters`, to offer a clearer configuration object that enables simplifying the default validation code as well.
[...]

PR #2729, opened 2024-07-17
Decrypt token: Remove exceptions + use new ValidationParameters

iNinja:
# Decrypt token: Remove exceptions + use new ValidationParameters

## Description
- Added new TokenDecryptingResult
- Added missing delegate for ValidateAlgorithm
- Updated ValidationParameters to provide the delegates required for token decryption
- Refactored DecryptToken method and dependencies to not throw exceptions and return result type.

Raising this as a draft PR to discuss the approach and validate the plan.
I will add tests before making it official.

Part of #2711.


src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -343,16 +344,65 @@ internal static string DecompressToken(byte[] tokenBytes, string algorithm, int
...
-                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10603, keysAttempted, (object)exceptionStrings ?? "", LogHelper.MarkAsSecurityArtifact(decryptionParameters.EncodedToken, SafeLogJwtToken))));
+            ExceptionDetail exceptionDetail = ValidateDecryption(
+                decryptionParameters,
+                decryptionSucceeded,
+                algorithmNotSupportedByCryptoProvider,
+                exceptionStrings,
+                keysAttempted,
+                null);
+
+            if (exceptionDetail is not null)
```
iNinja: This method uses an approach I am trying to avoid having entire duplicates.
The original method throws in 3 situations. I have implemented the new method that will return the `ExceptionDetail` to be included in the Result type, and kept the current behaviour by generating the exception and throwing it.

There are 2 extra allocations with this approach for the existing code path, to create the MessageDetail and ExceptionDetail before throwing.
I am open to opinions on whether this is worth to avoid having 2 separate methods with separate implementations.

src/Microsoft.IdentityModel.JsonWebTokens/TokenDecryptingResult.cs
```diff
@@ -0,0 +1,84 @@
...
+            JWT = jwtToken;
+            IsValid = false;
+        }
+
+        /// <summary>
+        /// Gets the decoded contents of the JWE.
+        /// </summary>
+        /// <exception cref="InvalidOperationException"/> if the result is not valid, and the decrypted token is not available.
+        /// <remarks>It is expected that this method will only be called if <see cref="ValidationResult.IsValid"/> returns true.</remarks>
+        public string DecryptionToken()
```
iNinja: Changed this based on my conversation with Keegan about throwing an exception from a property.
Making the result here non optional simplifies the calling points, but it requires that the caller checks `IsValid` beforehand.

src/Microsoft.IdentityModel.Tokens/Delegates.cs
```diff
@@ -169,4 +169,17 @@ namespace Microsoft.IdentityModel.Tokens
...
+    /// <summary>
+    /// Resolves the decryption key for the security token.
+    /// </summary>
+    /// <param name="token">The string representation of the token to be decrypted.</param>
+    /// <param name="securityToken">The <see cref="SecurityToken"/> to be decrypted, which is null by default.</param>
+    /// <param name="kid">The key identifier, which may be null.</param>
+    /// <param name="validationParameters">The <see cref="ValidationParameters"/> to be used for validating the token.</param>
+    /// <param name="callContext">The <see cref="CallContext"/> to be used for logging.</param>
+    /// <returns>The <see cref="SecurityKey"/> used to decrypt the token.</returns>
+    internal delegate IList<SecurityKey> ResolveTokenDecryptionKeyDelegate(string token, SecurityToken securityToken, string kid, ValidationParameters validationParameters, CallContext? callContext);
```
iNinja, line 183: I would have preferred to add this at the top of the file doing the decryption like we did with the validators, but we cannot have Microsoft.IdentityModel.Tokens depend on Microsoft.IdentityModel.JsonWebTokens.
There may be a better place for this. I am open to suggestions.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -103,13 +103,13 @@ public ValidationParameters()
...
+        /// Allows overriding the delegate used to validate the cryptographic algorithm used.
         /// </summary>
         /// <remarks>
-        /// If set, this delegate will validate the cryptographic algorithm used and
-        /// the algorithm will not be checked against <see cref="ValidAlgorithms"/>.
+        /// If no delegate is set, the default implementation will be used. The default checks the algorithm
+        /// against the <see cref="ValidAlgorithms"/> property, if present. If not, it will succeed.
         /// </remarks>
-        public AlgorithmValidator AlgorithmValidator { get; set; }
+        public AlgorithmValidationDelegate AlgorithmValidator { get; set; } = Validators.ValidateAlgorithm;
```
brentschmaltz: Should we allow this to be set to null?

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -228,6 +228,11 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
         /// </summary>
         public string DebugId { get; set; }
 
+        /// <summary>
+        /// Gets the <see cref="SecurityKey"/> representing the ephemeral public key used for ECDH-ES decryption.
+        /// </summary>
+        public SecurityKey EcdhEsDecryptionKey { get; set; }
```
brentschmaltz: This is sensitive information, could be the private key, do we need this?

iNinja: I was reading to familiarise myself more with ECDH-ES as we currently support it and it appears to me that this is a public key that is also ephemeral, so I would not imagine it is too sensitive.

On the current implementation, the key is set in TVP.DecryptionKey or very similar. Since we are doing away with that field in favour of the List one, I think it is safest to allow the user to set this in a field specific for this use rather than iterating the list of keys and seeing if any is of this type.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
iNinja: I am still bouncing this idea in my head. Whether they should set a mostly empty validator to skip, or be allowed to simply set to null to skip the call altogether.

FuPingFranco: I thought we were keeping the naming consistent with our current methods by just adding 'Delegate' as a postfix. i.e: AlgorithmValidatorDelegate?

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
iNinja: This is not printed anywhere.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
iNinja: You are right. It should be `AlgorithmValidatorDelegate`.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -103,13 +103,13 @@ public ValidationParameters()
...
+        /// Allows overriding the delegate used to validate the cryptographic algorithm used.
         /// </summary>
         /// <remarks>
-        /// If set, this delegate will validate the cryptographic algorithm used and
-        /// the algorithm will not be checked against <see cref="ValidAlgorithms"/>.
+        /// If no delegate is set, the default implementation will be used. The default checks the algorithm
+        /// against the <see cref="ValidAlgorithms"/> property, if present. If not, it will succeed.
         /// </remarks>
-        public AlgorithmValidator AlgorithmValidator { get; set; }
+        public AlgorithmValidatorDelegate AlgorithmValidator { get; set; } = Validators.ValidateAlgorithm;
```
FuPingFranco: Will you be making the change to throw a nre if set to null or do you want me to do it on my PR?

iNinja: I think it may be best if you do it on yours to keep everything together.

src/Microsoft.IdentityModel.JsonWebTokens/Results/TokenDecryptingResult.cs
```diff
@@ -0,0 +1,87 @@
...
+using Microsoft.IdentityModel.Tokens;
+
+#nullable enable
+namespace Microsoft.IdentityModel.JsonWebTokens
+{
+    /// <summary>
+    /// Contains the result of decrypting a JWT in clear text.
+    /// The <see cref="TokenValidationResult"/> contains a collection of <see cref="ValidationResult"/> for each step in the token validation.
+    /// </summary>
+    internal class TokenDecryptingResult : ValidationResult
```
brentschmaltz: I think the name TokenDecryptionResult would better match the other ValidationResults.

src/Microsoft.IdentityModel.JsonWebTokens/Results/TokenDecryptingResult.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        private Exception? _exception;
+        private string? _decryptedToken;
+
+        /// <summary>
+        /// Creates an instance of <see cref="TokenDecryptingResult"/> containing the clear text result of decrypting a JWE.
+        /// </summary>
+        /// <paramref name="decryptedToken"/>The clear text result of decrypting the JWE.
+        /// <paramref name="jwtToken"/>The JWE that contains the cypher text.
+        public TokenDecryptingResult(string decryptedToken, JsonWebToken jwtToken)
```
brentschmaltz: parameter should be SecurityToken rather than JsonWebToken.

src/Microsoft.IdentityModel.JsonWebTokens/Results/TokenDecryptingResult.cs
```diff
@@ -0,0 +1,87 @@
...
+            JWT = jwtToken;
+        }
+
+        /// <summary>
+        /// Creates an instance of <see cref="TokenDecryptingResult"/>
+        /// </summary>
+        /// <paramref name="jwtToken"/> is the JWT that could not be decrypted.
+        /// <paramref name="validationFailure"/> is the <see cref="ValidationFailureType"/> that occurred during reading.
+        /// <paramref name="exceptionDetail"/> is the <see cref="ExceptionDetail"/> that occurred during reading.
+        public TokenDecryptingResult(JsonWebToken? jwtToken, ValidationFailureType validationFailure, ExceptionDetail exceptionDetail)
```
brentschmaltz, line 38: In what scenario would JsonWebToken be null?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// Decrypts a JWE and returns the clear text.
+        /// </summary>
+        /// <param name="jwtToken">The JWE that contains the cypher text.</param>
+        /// <param name="validationParameters">The <see cref="TokenValidationParameters"/> to be used for validating the token.</param>
+        /// <param name="configuration">The <see cref="BaseConfiguration"/> to be used for validating the token.</param>
+        /// <param name="callContext"></param>
+        /// <returns>The decoded / cleartext contents of the JWE.</returns>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="jwtToken"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="validationParameters"/> is null.</exception>
+        /// <exception cref="SecurityTokenException">Thrown if <see cref="JsonWebToken.Enc"/> is null or empty.</exception>
```
brentschmaltz: Since we do not actually throw, the comment on 'exception' should not indicate the runtime throws.
Instead, the comments should explain what to expect in the TokenDecryptionResult (i suggest a rename of the ValidationResult).

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            if (keys == null)
+                return new TokenDecryptingResult(
+                    jwtToken,
+                    ValidationFailureType.TokenDecryptingFailed,
+                    new ExceptionDetail(
+                        new MessageDetail(
+                            TokenLogMessages.IDX10609,
+                            LogHelper.MarkAsSecurityArtifact(jwtToken, JwtTokenUtilities.SafeLogJwtToken)),
+                        typeof(SecurityTokenException),
```
brentschmaltz: Consider setting the exception type to: SecurityTokenDecryptionFailedException

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -0,0 +1,238 @@
...
+                new JwtTokenDecryptionParameters
+                {
+                    DecompressionFunction = JwtTokenUtilities.DecompressToken,
+                    Keys = keys,
+                    MaximumDeflateSize = MaximumTokenSizeInBytes
+                },
+                callContext);
+        }
+
+        internal IEnumerable<SecurityKey>? GetContentEncryptionKeys(JsonWebToken jwtToken, ValidationParameters validationParameters, BaseConfiguration configuration, CallContext? callContext)
```
brentschmaltz: This should return an 
```cs
IList<SecurityKey>
```

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -0,0 +1,238 @@
...
+                foreach (var key in validationParameters.TokenDecryptionKeys)
+                {
+                    if (key != null && string.Equals(key.KeyId, jwtToken.Kid, GetStringComparisonRuleIf509OrECDsa(key)))
+                        return key;
+                }
+            }
+
+            if (!string.IsNullOrEmpty(jwtToken.X5t) && validationParameters.TokenDecryptionKeys != null)
+            {
+                foreach (var key in validationParameters.TokenDecryptionKeys)
```
brentschmaltz: TokenDecryptionKeys is an IList, quicker to use for(int i =0; i < validationParameters.TokenDecryptionKeys.Count; i++)

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.DecryptTokenResult.cs
```diff
@@ -0,0 +1,151 @@
...
+                    if (!cryptoProviderFactory.IsSupportedAlgorithm(jsonWebToken.Enc, key))
+                    {
+                        if (LogHelper.IsEnabled(EventLogLevel.Warning))
+                            LogHelper.LogWarning(TokenLogMessages.IDX10611, LogHelper.MarkAsNonPII(decryptionParameters.Enc), key);
+
+                        algorithmNotSupportedByCryptoProvider = true;
+                        continue;
+                    }
+
+                    AlgorithmValidationResult result = validationParameters.AlgorithmValidator(zipAlgorithm, key, jsonWebToken, validationParameters, callContext);
```
brentschmaltz, line 82: Consider adding a List of AlgorithmValidationResult and attaching to the TokenDecryptionResult (renamed), then no need to have the exceptionStrings.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.DecryptTokenResult.cs
```diff
@@ -0,0 +1,151 @@
...
+
+            bool decryptionSucceeded = false;
+            bool algorithmNotSupportedByCryptoProvider = false;
+            byte[] decryptedTokenBytes = null;
+
+            // keep track of exceptions thrown, keys that were tried
+            StringBuilder exceptionStrings = null;
+            StringBuilder keysAttempted = null;
+            string zipAlgorithm = null;
+            foreach (SecurityKey key in decryptionParameters.Keys)
```
brentschmaltz, line 60: Consider refactoring and adding a method that takes one key.
This will simplify the logic.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.DecryptTokenResult.cs
```diff
@@ -0,0 +1,151 @@
...
+#pragma warning restore CA1031 // Do not catch general exception types
+                {
+                    (exceptionStrings ??= new StringBuilder()).AppendLine(ex.ToString());
+                }
+
+                if (key != null)
+                    (keysAttempted ??= new StringBuilder()).AppendLine(key.ToString());
+            }
+
+            ExceptionDetail exceptionDetail = ValidateDecryption(
```
brentschmaltz: I would recommend consider removing this call as after refactoring into a method that has one key would simplify the logic where this is not needed.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
FuPingFranco: Re-opened as the handling of null AlgrithmValidatorDelegate setting will be done in this PR.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.DecryptTokenResult.cs
iNinja: Adding a work item to look at this after we have our `ValidateTokenAsync` implementation working.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.DecryptTokenResult.cs
iNinja: Adding a work item to look at this after we have our `ValidateTokenAsync` implementation working.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.DecryptTokenResult.cs
iNinja: Adding a work item to look at this after we have our `ValidateTokenAsync` implementation working.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
iNinja: Not adding the test to ValidationParameters on this PR to avoid conflicts. I'll update the test once Franco's PR is merged.

FuPingFranco (approved):
LGTM, you'll need to rebase as my changes went it on Friday

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -0,0 +1,270 @@
...
+                {
+                    DecompressionFunction = JwtTokenUtilities.DecompressToken,
+                    Keys = keys,
+                    MaximumDeflateSize = MaximumTokenSizeInBytes
+                },
+                callContext);
+        }
+
+        internal (IList<SecurityKey>?, ExceptionDetail?) GetContentEncryptionKeys(JsonWebToken jwtToken, ValidationParameters validationParameters, BaseConfiguration configuration, CallContext? callContext)
+        {
```
jennyf19, line 85: this method is quite large...can it be broken into smaller helper methods, which are also easier to test?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -0,0 +1,270 @@
...
+        /// <exception cref="SecurityTokenException">Returned inside <see cref="TokenDecryptionResult.Exception"/> if <see cref="JsonWebToken.Enc"/> is null or empty.</exception>
+        /// <exception cref="SecurityTokenDecompressionFailedException">Returned inside <see cref="TokenDecryptionResult.Exception"/> if the decompression failed.</exception>
+        /// <exception cref="SecurityTokenEncryptionKeyNotFoundException">Returned inside <see cref="TokenDecryptionResult.Exception"/> if <see cref="JsonWebToken.Kid"/> is not null AND the decryption fails.</exception>
+        /// <exception cref="SecurityTokenDecryptionFailedException">Returned inside <see cref="TokenDecryptionResult.Exception"/> if the JWE was not able to be decrypted.</exception>
+        internal TokenDecryptionResult DecryptToken(
+            JsonWebToken jwtToken,
+            ValidationParameters validationParameters,
+            BaseConfiguration configuration,
+            CallContext? callContext)
+        {
```
jennyf19, line 37: could you have a helper method to just validate the inputs?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.DecryptTokenResult.cs
```diff
@@ -0,0 +1,151 @@
+﻿using System;
```
jennyf19: remember to add license/header info to all new files

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.DecryptTokenResult.cs
```diff
@@ -0,0 +1,151 @@
...
+    {
+        /// <summary>
+        /// Decrypts a JWT token.
+        /// </summary>
+        /// <param name="jsonWebToken">The JWT token to decrypt.</param>
+        /// <param name="validationParameters">The <see cref="ValidationParameters"/> to be used for validating the token.</param>
+        /// <param name="decryptionParameters">The decryption parameters container.</param>
+        /// <param name="callContext">The call context used for logging.</param>
+        /// <returns>The decrypted, and if the 'zip' claim is set, decompressed string representation of the token.</returns>
+        internal static TokenDecryptionResult DecryptJwtToken(
```
jennyf19, line 24: nit: any possibility of splitting this into smaller helper methods?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.DecryptTokenResult.cs
```diff
@@ -0,0 +1,151 @@
...
+            bool algorithmNotSupportedByCryptoProvider = false;
+            byte[] decryptedTokenBytes = null;
+
+            // keep track of exceptions thrown, keys that were tried
+            StringBuilder exceptionStrings = null;
+            StringBuilder keysAttempted = null;
+            string zipAlgorithm = null;
+            foreach (SecurityKey key in decryptionParameters.Keys)
+            {
+                var cryptoProviderFactory = validationParameters.CryptoProviderFactory ?? key.CryptoProviderFactory;
```
jennyf19, line 62: can the decryption logic be extracted to a helper method?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.DecryptTokenResult.cs
```diff
@@ -0,0 +1,151 @@
...
+                    break;
+                }
+#pragma warning disable CA1031 // Do not catch general exception types
+                catch (Exception ex)
+#pragma warning restore CA1031 // Do not catch general exception types
+                {
+                    (exceptionStrings ??= new StringBuilder()).AppendLine(ex.ToString());
+                }
+
+                if (key != null)
```
jennyf19, line 109: do we need to implement any key management or caching (not related to the purpose of `keysAttempted` but more around overall impact on performance.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -343,16 +344,65 @@ internal static string DecompressToken(byte[] tokenBytes, string algorithm, int
...
 
-            if (!decryptionSucceeded && algorithmNotSupportedByCryptoProvider)
-                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10619, LogHelper.MarkAsNonPII(decryptionParameters.Alg), LogHelper.MarkAsNonPII(decryptionParameters.Enc))));
+        private static ExceptionDetail ValidateDecryption(
+            JwtTokenDecryptionParameters decryptionParameters,
+            bool decryptionSucceeded,
+            bool algorithmNotSupportedByCryptoProvider,
+            StringBuilder exceptionStrings,
+            StringBuilder keysAttempted,
+#pragma warning disable CA1801 // Review unused parameters
```
jennyf19, line 359: can callContext be removed?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -343,16 +344,65 @@ internal static string DecompressToken(byte[] tokenBytes, string algorithm, int
...
+            CallContext callContext)
+#pragma warning restore CA1801 // Review unused parameters
+        {
```
jennyf19, line 362: if `decryptionSucceeded` can we just return, as no further action is needed?

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -343,16 +344,65 @@ internal static string DecompressToken(byte[] tokenBytes, string algorithm, int
...
+            ExceptionDetail exceptionDetail = null;
 
-            if (!decryptionSucceeded)
-                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(TokenLogMessages.IDX10609, LogHelper.MarkAsSecurityArtifact(decryptionParameters.EncodedToken, SafeLogJwtToken))));
+            if (!decryptionSucceeded && keysAttempted is not null)
```
jennyf19: if you return if decryptionSucceeded, then you can remove all subsequent checks

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -343,16 +344,65 @@ internal static string DecompressToken(byte[] tokenBytes, string algorithm, int
...
+            else if (!decryptionSucceeded && algorithmNotSupportedByCryptoProvider)
+                return new ExceptionDetail(
+                    new MessageDetail(
+                        TokenLogMessages.IDX10619,
+                        LogHelper.MarkAsNonPII(decryptionParameters.Alg),
+                        LogHelper.MarkAsNonPII(decryptionParameters.Enc)),
+                    typeof(SecurityTokenDecryptionFailedException),
+                    new StackFrame(true),
+                    null);
+            else if (!decryptionSucceeded)
```
jennyf19: this could become a more general decryption failure case w/just an `else`

src/Microsoft.IdentityModel.Tokens/Validation/TokenDecryptionResult.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary>
+        /// Gets the decoded contents of the SecurityToken.
+        /// </summary>
+        /// <exception cref="InvalidOperationException"/> if the result is not valid, and the decrypted token is not available.
+        /// <remarks>It is expected that this method will only be called if <see cref="ValidationResult.IsValid"/> returns true.</remarks>
+        public string DecryptedToken()
+        {
+            if (_decryptedToken is null)
+                throw new InvalidOperationException("Attempted to retrieve the DecryptedToken from a failed TokenDecrypting result.");
```
jennyf19, line 63: is there any action the developer can take to resolve this type of exception? if so, might be worth starting a wiki page w/an aka.ms link we can include.

test/Microsoft.IdentityModel.TestUtils/IdentityComparer.cs
```diff
@@ -1078,6 +1078,81 @@ public static bool AreTokenReadingResultsEqual(object object1, object object2, C
             return context.Merge(localContext);
         }
 
+        public static bool AreTokenDecryptingResultsEqual(object object1, object object2, CompareContext context)
```
jennyf19, line 1081: great to see the additional test coverage. 

is the performance roughly the same, improving or worse w/these changes?

jennyf19 (approved):
A few nits, nothing blocking.
:shipit:

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
brentschmaltz: @jennyf19 we will be using this to hook up logging in the future.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
iNinja: Added some helpers to make our null checks minimal that can be reused across all methods that return results.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
iNinja: We have an issue created to look into this refactor.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
iNinja: Refactored this piece to only use a method if decryption failed and return the exception to be returned.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
iNinja: Refactored.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
iNinja: Refactored.

src/Microsoft.IdentityModel.Tokens/Validation/TokenDecryptionResult.cs
iNinja: There is a remark on this method's documentation explaining how to ensure this exception is never fired.

test/Microsoft.IdentityModel.TestUtils/IdentityComparer.cs
iNinja: We will be getting to the benchmarks and optimisations shortly.

PR #2733, opened 2024-07-18
remove prefix for release builds

brentschmaltz:
Set nugetSuffix in buildConfiguration.xml to empty so preview will not show up in release build.


jennyf19 (approved):
:shipit:

PR #2742, opened 2024-07-23
Remove old 6x tests used that are not needed anymore

brentschmaltz:
These tests were used during the transition from newtonsoft to System.Text.Json and Utf8JsonReader / Writer.
The are no longer needed and get in the way.


jennyf19 (approved):
:shipit:

sruke (approved):

PR #2745, opened 2024-07-24
Allow Jwt payload to be the empty string.

### LINKED ISSUE #2656: [Bug / Feature Request]  Create JsonWebToken with empty Payload
I need to create [Detached JWS signature](https://tools.ietf.org/html/rfc7797) for all requests to a webservice. Also for `GET` requests (which do not have a body/payload). This library creates JWTs of thich I remove the payload to get the JWS variant. But for GET-requests there is no payload. The `CreateToken()`-method fails when the payload is `null` or `string.Empty`.

I want to be able to call `JsonWebTokenHandler.CreateToken()` with `string.Empty` as payload. 
The result should be an empty payload in the created token:
```
{header}..{signature}
```

I've concidered using `JsonWebTokenHandler.CreateToken(SecurityTokenDescriptor tokenDescriptor)`. The issue is that then an (empty) json-object is automatically created as payload.  

I'm now calling the `internal static CreateToken()` on `JsonWebTokenHandler` using reflection with an `string.Empty` payload. This actually works perfectly and results in the expected behaviour as described above. Only reflection is not a future proof solution.

**Additional context**
[iDeal API desciption with requirements of Detached JWS for all requests](https://currencenl.atlassian.net/wiki/spaces/IPD/pages/3417604322/Security+for+Direct+Connection#For-requests-and-responses-to-iDEAL)

brentschmaltz:
Nothing in the jwt spec requires the payload to have any values.
This PR allows the payload to be empty.

Fixes #2656

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -114,7 +114,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         /// Initializes a new instance of the <see cref="JsonWebToken"/> class where the header contains the crypto algorithms applied to the encoded header and payload.
         /// </summary>
         /// <param name="header">A string containing JSON which represents the cryptographic operations applied to the JWT and optionally any additional properties of the JWT.</param>
-        /// <param name="payload">A string containing JSON which represents the claims contained in the JWT. Each claim is a JSON object of the form { Name, Value }.</param>
+        /// <param name="payload">A string containing JSON which represents the claims contained in the JWT. Each claim is a JSON object of the form { Name, Value }. Can be the empty.</param>
         /// <remarks>
         /// See: <see href="https://datatracker.ietf.org/doc/html/rfc7519"/> (JWT).
         /// See: <see href="https://datatracker.ietf.org/doc/html/rfc7515"/> (JWS).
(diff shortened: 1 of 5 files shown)


jmprieur (approved):
Lgtm
I have not seen a. Round-trip test?
Empty payload - > token with header.. Signature - > parse to get 0 claims??

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -130,7 +130,7 @@ public JsonWebToken(string header, string payload)
             if (string.IsNullOrEmpty(header))
                 throw LogHelper.LogArgumentNullException(nameof(header));
 
-            if (string.IsNullOrEmpty(payload))
+            if (payload == null)
```
remyblok: Docs on Line 117 and 127 should be updated now.

brentschmaltz: @remyblok updated.

FuPingFranco (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -114,7 +114,7 @@ public JsonWebToken(ReadOnlyMemory<char> encodedTokenMemory)
         /// Initializes a new instance of the <see cref="JsonWebToken"/> class where the header contains the crypto algorithms applied to the encoded header and payload.
         /// </summary>
         /// <param name="header">A string containing JSON which represents the cryptographic operations applied to the JWT and optionally any additional properties of the JWT.</param>
-        /// <param name="payload">A string containing JSON which represents the claims contained in the JWT. Each claim is a JSON object of the form { Name, Value }.</param>
+        /// <param name="payload">A string containing JSON which represents the claims contained in the JWT. Each claim is a JSON object of the form { Name, Value }. Can be the empty.</param>
```
sruke, line 117: Consider rephrasing to "Can be empty".

PR #2751, opened 2024-07-26
Remove SlimLock when updating metadata.

brentschmaltz:
Metadata is now updated on a background Task.
Previously updating was behind a SlimLock which resulted in a call to RequestRefresh would block all token validations while obtaining metadata from the discovery endpoint.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationValidator.cs
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationValidator.cs
@@ -39,11 +39,20 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect.Configuration
                     Succeeded = false
                 };
             }
-            var numberOfValidKeys = openIdConnectConfiguration.JsonWebKeySet.Keys.Where(key => key.ConvertedSecurityKey != null).Count();
+
+            int numberOfValidKeys = 0;
+            for( int i = 0;  i < openIdConnectConfiguration.JsonWebKeySet.Keys.Count; i++)
+                if (openIdConnectConfiguration.JsonWebKeySet.Keys[i].ConvertedSecurityKey != null)
+                    numberOfValidKeys++;
 
             if (numberOfValidKeys < MinimumNumberOfKeys)
             {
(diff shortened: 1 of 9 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -145,83 +148,149 @@ public async Task<T> GetConfigurationAsync()
...
+        /// </summary>
+        private void UpdateCurrentConfiguration()
+        {
+#pragma warning disable CA1031 // Do not catch general exception types
+            try
+            {
+                T configuration = _configRetriever.GetConfigurationAsync(
+                    MetadataAddress,
+                    _docRetriever,
+                    CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
```
GeoK, line 258: Generally speaking, we should make this method async to free up the thread while waiting for the HTTP response. Having that logic allows only for one thread to be in this state at a time, at most once in a 5-minute timeframe, I'm not sure if there are any benefits.

brentschmaltz: @GeoK this is called from _Task.Run so the task pool is being used is the thread blocked?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationValidator.cs
```diff
@@ -39,11 +39,20 @@ public ConfigurationValidationResult Validate(OpenIdConnectConfiguration openIdC
                     Succeeded = false
                 };
             }
-            var numberOfValidKeys = openIdConnectConfiguration.JsonWebKeySet.Keys.Where(key => key.ConvertedSecurityKey != null).Count();
+
+            int numberOfValidKeys = 0;
+            for( int i = 0;  i < openIdConnectConfiguration.JsonWebKeySet.Keys.Count; i++)
```
jennyf19: nit: spacing? `for( ` as opposed to `for ( `?

westin-m: +1 in favor of `for (`

jennyf19 (approved):
:shipit:

GeoK (approved):

PR #2754, opened 2024-07-30
Fix OIDC Test

westin-m:
Change the URL in the config manager test to prevent an html response from coming back, which causes the test to fail

### DIFF
--- a/test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
+++ b/test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
@@ -327,7 +327,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests
             }
 
             // get configuration from https address, should throw
-            configManager = new ConfigurationManager<OpenIdConnectConfiguration>("https://someaddress.com", new OpenIdConnectConfigurationRetriever());
+            configManager = new ConfigurationManager<OpenIdConnectConfiguration>("https://someadress.com", new OpenIdConnectConfigurationRetriever());
             ee = new ExpectedException(typeof(InvalidOperationException), "IDX20803:", typeof(IOException));
             try
             {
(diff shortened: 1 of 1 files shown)


jennyf19 (approved):

JoshLozensky:
Is sending traffic to a random uri potentially insecure? Rather than misspelling it could we use something msft owns consistent with other tests in the repo like contoso.com? If that runs into similar issues, why can't we handle the "<" character?

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -327,7 +327,7 @@ public void GetConfiguration()
             }
 
             // get configuration from https address, should throw
-            configManager = new ConfigurationManager<OpenIdConnectConfiguration>("https://someaddress.com", new OpenIdConnectConfigurationRetriever());
+            configManager = new ConfigurationManager<OpenIdConnectConfiguration>("https://someadress.com", new OpenIdConnectConfigurationRetriever());
```
jennyf19:
```suggestion
            configManager = new ConfigurationManager<OpenIdConnectConfiguration>("http://127.0.0.1", new OpenIdConnectConfigurationRetriever());
```

jennyf19: matches this [PR](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/ec39ddb318b045ea1b0a2bd993356c6726e5a3bd) and resolves Josh's comment too.

westin-m: I'll push something that applies this to both parts of the test that refer to someaddress.com

westin-m:
> Is sending traffic to a random uri potentially insecure? Rather than misspelling it could we use something msft owns consistent with other tests in the repo like contoso.com?

We'll use the localhost address instead.
> If that runs into similar issues, why can't we handle the "<" character?

We're only deserializing JSON, not HMTL.

JoshLozensky (approved):

PR #2755, opened 2024-07-30
JsonWebKeySet stores the String it was created with

westin-m:
Store the original string the JsonWebKeySet was created with


test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectConfigurationRetrieverTests.cs
```diff
@@ -28,7 +30,13 @@ public async Task FromNetwork()
         [Fact]
         public async Task FromFile()
         {
-            var context = new CompareContext();
+            var context = new CompareContext
+            {
+                PropertiesToIgnoreWhenComparing = new Dictionary<Type, List<string>>
```
GeoK, line 35: Would be great to add comments explaining why this property should be ignored when comparing.

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -97,6 +99,21 @@ public JsonWebKeySet(string json)
         [JsonIgnore]
         public bool SkipUnresolvedJsonWebKeys { get; set; } = DefaultSkipUnresolvedJsonWebKeys;
 
+        /// <summary>
+        /// The original <see cref="string"/> used to create this instance.
+        /// </summary>
+        public string JsonWebKeySetString
```
brentschmaltz: Add attribute [JsonIgnore].

brentschmaltz: How about change the property name to: JsonData

pmaytak:
Just for my understanding, what's the reasoning for keeping the original string? When would it ever be used for, if it's already parsed into properties?

jennyf19 (approved):
:shipit:

westin-m:
> Just for my understanding, what's the reasoning for keeping the original string? When would it ever be used for, if it's already parsed into properties?

@pmaytak The original string can be used by the OIDC config manager to, on L2 cache reads, deserialize and store the keys alongside the metadata document without needing the JWKs to serialize the security keys

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -97,6 +99,22 @@ public JsonWebKeySet(string json)
         [JsonIgnore]
         public bool SkipUnresolvedJsonWebKeys { get; set; } = DefaultSkipUnresolvedJsonWebKeys;
 
+        /// <summary>
+        /// The original <see cref="string"/> used to create this instance.
+        /// </summary>
+        [JsonIgnore]
+        public string JsonData
```
pmaytak: `internal`? If it's only used for our config manager?

test/Microsoft.IdentityModel.Tokens.Tests/Json/DataSets.cs
```diff
@@ -732,6 +732,7 @@ public static JsonWebKeySet JsonWebKeySet1
                 JsonWebKeySet jsonWebKeySet = new JsonWebKeySet();
                 jsonWebKeySet.Keys.Add(JsonWebKey1);
                 jsonWebKeySet.Keys.Add(JsonWebKey2);
+                jsonWebKeySet.JsonData = JsonWebKeySetString1;
```
pmaytak: Can we use the JsonWebKeySet constructor that accepts JSON instead of setting the JsonData and adding keys here? That way we'll know JSON data and keys are equivalent; plus an extra check that `JsonWebKeySetSerializer` reads the string properly.

pmaytak:
> > Just for my understanding, what's the reasoning for keeping the original string? When would it ever be used for, if it's already parsed into properties?
> 
> @pmaytak The original string can be used by the OIDC config manager to, on L2 cache reads, deserialize and store the keys alongside the metadata document without needing the JWKs to serialize the security keys

@westin-m But we can add additional keys and remove keys from `Keys` collection. So is there a benefit then to keeping the original JSON string if it can get out of sync? If we're serializing the key set, we'd want to serialize the current `Keys` always anyway?

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectConfigurationRetrieverTests.cs
```diff
@@ -28,7 +30,15 @@ public async Task FromNetwork()
...
+                {
+                    // If the objects being compared are created from the same string and they are equal, the string itself can be ignored.
```
pmaytak, line 37: I think we should still compare the `JsonData` to verify this assumption against regressions - is the original string passed into the constructor actually assigned to the property.

test/Microsoft.IdentityModel.Tokens.Tests/JsonWebKeySetTests.cs
```diff
@@ -23,7 +23,14 @@ public void Constructors(JsonWebKeySetTheoryData theoryData)
...
                 var jsonWebKeys = new JsonWebKeySet(theoryData.Json);
                 var keys = jsonWebKeys.GetSigningKeys();
+                var originalString = jsonWebKeys.JsonData;
                 theoryData.ExpectedException.ProcessNoException(context);
+
+                if (theoryData.Json != null)
+                    IdentityComparer.AreStringsEqual(originalString, theoryData.Json, context);
+                else
+                    IdentityComparer.AreStringsEqual(originalString, string.Empty, context);
+
```
pmaytak: I think `JsonWebKeySetTheoryData.JsonData` should be set to `string.Empty` by default, since that's the default test case we're testing for, and then this code wouldn't be needed.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectConfigurationRetrieverTests.cs
westin-m: > is the original string passed into the constructor actually assigned to the property

This is verified by `JsonWebKeySetTests`

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -53,6 +54,7 @@ public JsonWebKeySet(string json)
         {
             if (string.IsNullOrEmpty(json))
                 throw LogHelper.LogArgumentNullException(nameof(json));
+            _jsonData = json;
```
keegan-caruso, line 58:
```suggestion

            _jsonData = json;
```

keegan-caruso: should have a newline

src/Microsoft.IdentityModel.Tokens/JsonWebKeySet.cs
```diff
@@ -97,6 +99,22 @@ public JsonWebKeySet(string json)
         [JsonIgnore]
         public bool SkipUnresolvedJsonWebKeys { get; set; } = DefaultSkipUnresolvedJsonWebKeys;
 
+        /// <summary>
+        /// The original <see cref="string"/> used to create this instance.
```
keegan-caruso: I don't think a cref string is useful here. Maybe something like this?

```suggestion
        /// The original string used to create this instance if a string was provided. Otherwise, an empty string.
```

test/Microsoft.IdentityModel.Tokens.Tests/Json/JsonWebKeySetTheoryData.cs
```diff
@@ -14,7 +14,7 @@ public class JsonWebKeySetTheoryData : TheoryDataBase
 
         public IList<SecurityKey> ExpectedSigningKeys { get; set; }
 
-        public string Json { get; set; }
+        public string Json { get; set; } = string.Empty;
```
keegan-caruso:
```suggestion
        public string Json { get; set; }
```

We don't need to initialize properties to the default values.

GeoK (approved):

PR #2757, opened 2024-07-31
Signature Validation: Remove exceptions

iNinja:
# Signature Validation: Remove exceptions

- [ ] You've read the [Contributor Guide](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/Contributing.md) and [Code of Conduct](https://opensource.microsoft.com/codeofconduct/).
- [ ] You've included unit or integration tests for your change, where applicable.

## Description
Part of the ValidateJWS/JWEAsync, removed exception throwing from signature validation. Refactored code and simplified some paths along the way.

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
@@ -1,0 +1,300 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using Microsoft.IdentityModel.Abstractions;
+using Microsoft.IdentityModel.JsonWebTokens.Results;
+using Microsoft.IdentityModel.Logging;
+using Microsoft.IdentityModel.Tokens;
+using TokenLogMessages = Microsoft.IdentityModel.Tokens.LogMessages;
+
+namespace Microsoft.IdentityModel.JsonWebTokens
+{
+#nullable enable
+    /// <remarks>This partial class contains methods and logic related to the validation of tokens' signatures.</remarks>
+    public partial class JsonWebTokenHandler : TokenHandler
+    {
+        /// <summary>
+        /// Validates the JWT signature.
+        /// </summary>
+        private static SignatureValidationResult ValidateSignature(
+            JsonWebToken jwtToken,
+            ValidationParameters validationParameters,
+            BaseConfiguration configuration,
+            CallContext callContext)
+        {
+            // Delegate is set by the user, we call it and return the result.
+            if (validationParameters.SignatureValidator is not null)
+                return validationParameters.SignatureValidator(jwtToken, validationParameters, configuration, callContext);
+
+            // If the user wants to accept unsigned tokens, they must implement the delegate.
+            if (!jwtToken.IsSigned)
+                return new SignatureValidationResult(
+                    ValidationFailureType.SignatureValidationFailed,
+                    new ExceptionDetail(
+                        new MessageDetail(
+                            TokenLogMessages.IDX10504,
+                            LogHelper.MarkAsSecurityArtifact(
+                                jwtToken.EncodedToken,
+                                JwtTokenUtilities.SafeLogJwtToken)
+                            ),
+                        typeof(SecurityTokenInvalidSignatureException),
+                        new StackFrame()));
+
+            SecurityKey? key = null;
+            if (validationParameters.IssuerSigningKeyResolver is not null)
+            {
+                key = validationParameters.IssuerSigningKeyResolver(
+                    jwtToken.EncodedToken,
+                    jwtToken,
+                    jwtToken.Kid,
+                    validationParameters,
+                    configuration,
+                    callContext);
+            }
+            else
+            {
+                // Resolve the key using the token's 'kid' and 'x5t' headers.
+                // Fall back to the validation parameters' keys if configuration keys are not set.
+                key = JwtTokenUtilities.ResolveTokenSigningKey(jwtToken.Kid, jwtToken.X5t, configuration.SigningKeys)
+                    ?? JwtTokenUtilities.ResolveTokenSigningKey(jwtToken.Kid, jwtToken.X5t, validationParameters.IssuerSigningKeys);
+            }
+
+            if (key is not null)
+            {
+                jwtToken.SigningKey = key;
+
+                // If the key is found, validate the signature.
+                return ValidateSignatureWithKey(jwtToken, key, validationParameters, callContext);
+            }
+
+            // Key could not be resolved. Depending on the configuration, try all keys or return an error.
+            if (validationParameters.TryAllIssuerSigningKeys)
+                return ValidateSignatureUsingAllKeys(jwtToken, validationParameters, configuration, callContext);
+            else
+                return new SignatureValidationResult(
+                    ValidationFailureType.SignatureValidationFailed,
+                    new ExceptionDetail(
+                        new MessageDetail(TokenLogMessages.IDX10500),
+                        typeof(SecurityTokenSignatureKeyNotFoundException),
+                        new StackFrame()));
+        }
+
+        private static SignatureValidationResult ValidateSignatureUsingAllKeys(
+            JsonWebToken jwtToken,
+            ValidationParameters
+            validationParameters, BaseConfiguration configuration,
+            CallContext callContext)
+        {
+            // control gets here if:
+            // 1. User specified delegate: IssuerSigningKeyResolver returned null
+            // 2. ResolveIssuerSigningKey returned null
+            // Try all the keys. This is the degenerate case, not concerned about perf.
+            ICollection<SecurityKey> keys = configuration?.SigningKeys ?? new List<SecurityKey>();
+            if (!(keys is List<SecurityKey> keysList))
+                keysList = keys.ToList();
+
+            if (!validationParameters.IssuerSigningKeys.IsNullOrEmpty())
+                keysList.AddRange(validationParameters.IssuerSigningKeys);
+
+            // keep track of exceptions thrown, keys that were tried
+            StringBuilder? exceptionStrings = null;
+            StringBuilder? keysAttempted = null;
+            var kidExists = !string.IsNullOrEmpty(jwtToken.Kid);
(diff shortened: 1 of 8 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,314 @@
...
+
+            if (kidExists && kidMatched)
+            {
+                JsonWebToken localJwtToken = jwtToken; // avoid closure on non-exceptional path
+                var isKidInTVP = keysInTokenValidationParameters.Any(x => x.KeyId.Equals(localJwtToken.Kid));
+                var keyLocation = isKidInTVP ? "TokenValidationParameters" : "Configuration";
+                return new ExceptionDetail(
+                    new MessageDetail(
+                        TokenLogMessages.IDX10511,
+                        LogHelper.MarkAsNonPII(keysAttempted?.ToString() ?? ""),
```
jennyf19: do we need a check if logging is enabled first?

iNinja: This is basically the message attached to the exception when generated. This will not log anything.
I thought whether we should have this information given it was information before, but now that we are returning an error, we need to tell something back to the user and that seemed to fit best.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,314 @@
...
+        {
+            // control gets here if:
+            // 1. User specified delegate: IssuerSigningKeyResolver returned null
+            // 2. ResolveIssuerSigningKey returned null
+            // Try all the keys. This is the degenerate case, not concerned about perf.
+            ICollection<SecurityKey> keys = configuration?.SigningKeys ?? new List<SecurityKey>();
+            if (!(keys is List<SecurityKey> keysList))
+                keysList = keys.ToList();
+
+            if (!validationParameters.IssuerSigningKeys.IsNullOrEmpty())
```
brentschmaltz: We don't want to call 'IsNullOrEmpty' as this uses IEnumerable.Any().

ValidationParameters.IssuerSigningKeys is never null and is an IList of string.
You can check for Count and use for( int i =0; ...)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,314 @@
...
+            ValidationParameters
+            validationParameters, BaseConfiguration? configuration,
+            CallContext callContext)
+        {
+            // control gets here if:
+            // 1. User specified delegate: IssuerSigningKeyResolver returned null
+            // 2. ResolveIssuerSigningKey returned null
+            // Try all the keys. This is the degenerate case, not concerned about perf.
+            ICollection<SecurityKey> keys = configuration?.SigningKeys ?? new List<SecurityKey>();
+            if (!(keys is List<SecurityKey> keysList))
```
brentschmaltz: Beter to check for IList[SecurityKey] than List[SecurityKey]

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,314 @@
...
+                keysList = keys.ToList();
+
+            if (!validationParameters.IssuerSigningKeys.IsNullOrEmpty())
+                keysList.AddRange(validationParameters.IssuerSigningKeys);
```
brentschmaltz: Instead of using AddRange, consider refactoring this method into a second method that takes an IList[SecurityKey]

src/Microsoft.IdentityModel.Tokens/Validation/SignatureValidationResult.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// The <see cref="TokenValidationResult"/> contains a collection of <see cref="ValidationResult"/> for each step in the token validation.
+    /// </summary>
+    internal class SignatureValidationResult: ValidationResult
+    {
+        private Exception? _exception;
+
+        /// <summary>
+        /// Creates an instance of <see cref="SignatureValidationResult"/> representing the successful result of validating a signature.
+        /// </summary>
+        public SignatureValidationResult() : base(ValidationFailureType.ValidationSucceeded)
```
brentschmaltz: By default, i think the ValidationFailureType sb ValidationNotEvaluated.

src/Microsoft.IdentityModel.Tokens/Validation/SignatureValidationResult.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            IsValid = true;
```
brentschmaltz: By default IsValid == false?

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -290,12 +291,12 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
...
         /// priority.
         /// </remarks>
-        public IssuerSigningKeyResolver IssuerSigningKeyResolver { get; set; }
+        public IssuerSigningKeyResolverDelegate IssuerSigningKeyResolver { get; set; }
 
         /// <summary>
         /// Gets or sets an <see cref="IList{SecurityKey}"/> used for signature validation.
         /// </summary>
-        public IList<SecurityKey> IssuerSigningKeys { get; }
+        public IList<SecurityKey> IssuerSigningKeys { get; set; }
```
brentschmaltz: We shouldn't have a setter here, and initialize lazy using Interlocked.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -430,7 +431,10 @@ public string RoleClaimType
         /// <remarks>
         /// If set, this delegate will be called to validate the signature of the token, instead of default processing.
         /// </remarks>
-        public SignatureValidator SignatureValidator { get; set; }
+        public SignatureValidatorDelegate SignatureValidator {
+            get { return _signatureValidator; }
+            set { _signatureValidator = value; }
```
brentschmaltz, line 442: We don't want users to set this to null, but for now let's leave it as is.

We know we have some work to do to figure this out, but the PR has enough to get us to an EndToEnd.

src/Microsoft.IdentityModel.Tokens/Validation/SignatureValidationResult.cs
iNinja: Yes, all of our results are initialised to not valid and not evaluated if nothing is changed.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
iNinja: It was done in order to use `AddRange`, but if I remove that method call I can deal with just `IList`.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
iNinja: Created issue #2760 to keep track of this.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
iNinja: Refactored to use Interlocked.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
iNinja: Refactored to avoid merging lists.

src/Microsoft.IdentityModel.Tokens/Validation/SignatureValidationResult.cs
iNinja: Added a separate method to create a successful result and left the default as it was.

jennyf19 (approved):
:shipit:

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,368 @@
...
+        /// <param name="callContext">The context in which the method is called.</param>
+        /// <exception cref="ArgumentNullException">Returned if <paramref name="jwtToken"/> or <paramref name="validationParameters"/> is null.</exception>"
+        /// <exception cref="SecurityTokenInvalidSignatureException">Returned by the default implementation if the token is not signed, or if the validation fails.</exception>
+        /// <exception cref="SecurityTokenInvalidAlgorithmException">Returned if the algorithm is not supported by the key.</exception>
+        /// <exception cref="SecurityTokenSignatureKeyNotFoundException">Returned if the key cannot be resolved.</exception>
+        internal static SignatureValidationResult ValidateSignature(
+            JsonWebToken jwtToken,
+            ValidationParameters validationParameters,
+            BaseConfiguration? configuration,
+            CallContext callContext)
```
kllysng, line 36: 🤔Should this be:
```suggestion
            CallContext? callContext)
```

FuPingFranco (approved):
LGTM

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,368 @@
...
+                    jwtToken,
+                    validationParameters,
+                    configuration,
+                    exceptionStrings,
+                    keysAttempted,
+                    kidExists,
+                    kidMatched));
+        }
+
+        private static (SignatureValidationResult? validResult, bool KidMatched, KeyMatchFailedResult? failedResult) ValidateUsingKeys(
```
kllysng, line 159: nit:
```suggestion
        private static (SignatureValidationResult? validResult, bool kidMatched, KeyMatchFailedResult? failedResult) ValidateUsingKeys(
```

PR #2759, opened 2024-08-01
Re-factor Issuer Validator to Follow New Validation Model

FuPingFranco:
# Re-factor Issuer Validator to Follow New Validation Model

## Description

- Removed checks for `ValidIssuer` from Issuer validation.
- Added new tests in ValidationParameters class.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Security.Claims;
+using System.Threading;
 using Microsoft.IdentityModel.Abstractions;
 using Microsoft.IdentityModel.Logging;
 
@@ -20,7 +21,7 @@ namespace Microsoft.IdentityModel.Tokens
         private string _nameClaimType = ClaimsIdentity.DefaultNameClaimType;
         private string _roleClaimType = ClaimsIdentity.DefaultRoleClaimType;
         private Dictionary<string, object> _instancePropertyBag;
-        private IList<string> _validIssuers = [];
+        private IList<string> _validIssuers;
         private IList<string> _validTokenTypes = [];
 
         private AlgorithmValidatorDelegate _algorithmValidator = Validators.ValidateAlgorithm;
@@ -90,7 +91,7 @@ namespace Microsoft.IdentityModel.Tokens
             ValidateWithLKG = other.ValidateWithLKG;
             ValidAlgorithms = other.ValidAlgorithms;
             ValidAudiences = other.ValidAudiences;
-            ValidIssuers = other.ValidIssuers;
+            _validIssuers = other.ValidIssuers;
             ValidTypes = other.ValidTypes;
         }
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.IdentityModel.Tokens/Validation/Validators.Issuer.cs
```diff
@@ -95,10 +91,8 @@ public static partial class Validators
             if (validationParameters.ConfigurationManager != null)
                 configuration = await validationParameters.ConfigurationManager.GetBaseConfigurationAsync(cancellationToken).ConfigureAwait(false);
 
-            // Throw if all possible places to validate against are null or empty
-            if (string.IsNullOrWhiteSpace(validationParameters.ValidIssuer)
-                && validationParameters.ValidIssuers.IsNullOrEmpty()
-                && string.IsNullOrWhiteSpace(configuration?.Issuer))
+            // Return failed IssuerValidationResult if all possible places to validate against are null or empty.
+            if (validationParameters.ValidIssuers.IsNullOrEmpty() && string.IsNullOrWhiteSpace(configuration?.Issuer))
```
iNinja: Similar to Brent's comment on the other PR, now that `ValidIssuers` is an `IList`, we should check `.Count == 0`.

iNinja (dismissed):
Minor change and we're good.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -21,6 +20,7 @@ internal class ValidationParameters
         private string _nameClaimType = ClaimsIdentity.DefaultNameClaimType;
         private string _roleClaimType = ClaimsIdentity.DefaultRoleClaimType;
         private Dictionary<string, object> _instancePropertyBag;
+        private IList<string> _validIssuers = [];
```
brentschmaltz: Leave these _validIssuers and _validTokenTypes null.
Use the pattern shown in the Get below.

In general, we should never have setters on properties that are ICollection<T> or IList<T> so we never have to check for null and can just use count.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -518,10 +518,16 @@ public TypeValidatorDelegate TypeValidator
...
-        /// The default is <c>null</c>.
+        /// Gets or sets the <see cref="IList{String}"/> that contains valid issuers that will be used to check against the token's issuer.
+        /// The default is an empty collection.
         /// </summary>
-        public IList<string> ValidIssuers { get; }
+        /// <exception cref="ArgumentNullException">Thrown when the value is set as null.</exception>
+        /// <returns>The <see cref="IList{String}"/> that contains valid issuers that will be used to check against the token's 'iss' claim.</returns>
+        public IList<string> ValidIssuers
+        {
+            get { return _validIssuers; }
```
brentschmaltz: We should remove the set and use this patern: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/149148109446c4aa6a341cb4d766ec27b5213bdd/src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs#L28

brentschmaltz (approved):

brentschmaltz (dismissed):
Properties that return IList<T> should not be settable.
If the property can be null, for a successful validation, use a delayed creation.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Issuer.cs
iNinja: Looks like we are still using `IsNullOrEmpty()` on an `IList`.

The comment above needs to be addressed.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
FuPingFranco: Updated

iNinja (approved):

SaurabhMSFT (approved):

PR #2767, opened 2024-08-06
try to fix codeQL

jennyf19:


keegan-caruso:
Fixed by #2768

.github/workflows/codeql-analysis.yml
```diff
@@ -1,25 +1,15 @@
...
-      - 'test/Microsoft.IdentityModel.ManagedKeyVaultSecurityKey.Tests/**'
-      - '/src/Microsoft.IdentityModel.KeyVaultExtensions/**'
-      - '/src/Microsoft.IdentityModel.ManagedKeyVaultSecurityKey/**'
     types:
       - opened
       - synchronize
       - reopened
       - ready_for_review
-    branches: [ "dev", "dev6x", "dev7x"]
+    branches: [ "dev" ]
```
keegan-caruso, line 12: I think we might want at least dev7 and maybe dev6, they are still active, no?

jennyf19: they have codeql on their own respective branches so it will run on PRs into those branches

PR #2771, opened 2024-08-07
ValidateTokenAsync: New code path

iNinja:
# ValidateTokenAsync: New code path

- [ ] You've read the [Contributor Guide](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/Contributing.md) and [Code of Conduct](https://opensource.microsoft.com/codeofconduct/).
- [ ] You've included unit or integration tests for your change, where applicable.

Join all the refactored validators into a new `ValidateTokenAsync` entry point.

## Description
- Implementation of ValidateTokenAsync, ValidateJWSAsync, ValidateJWEAsync
- Added InternalTokenValidationResult to hold the intermediate validation results. Moved result types to their own folder.
- Removed unused delegate from TokenValidationParameters

Tests to be added before marking ready for review.
Creation of `ClaimsIdentity` still needs a resolution, as the current implementation in `TokenValidationResult` uses `TokenValidationParameters`.

Part of #2711


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
```diff
@@ -0,0 +1,309 @@
...
+        /// <exception cref="ArgumentNullException">Returned if <paramref name="token"/> is null or empty.</exception>
+        /// <exception cref="ArgumentNullException">Returned if <paramref name="validationParameters"/> is null.</exception>
+        /// <exception cref="ArgumentException">Returned if 'token.Length' is greater than <see cref="TokenHandler.MaximumTokenSizeInBytes"/>.</exception>
+        /// <exception cref="SecurityTokenMalformedException">Returned if <paramref name="token"/> is not a valid <see cref="JsonWebToken"/>, <see cref="ReadToken(string, CallContext)"/></exception>
+        /// <exception cref="SecurityTokenMalformedException">Returned if the validationParameters.TokenReader delegate is not able to parse/read the token as a valid <see cref="JsonWebToken"/>, <see cref="ReadToken(string, CallContext)"/></exception>
+        internal async Task<TokenValidationResult> ValidateTokenAsync(
+            string token,
+            ValidationParameters validationParameters,
+            CallContext callContext,
+            CancellationToken? cancellationToken)
```
brentschmaltz, line 40: should we allow this to be null?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
```diff
@@ -0,0 +1,309 @@
...
+        {
+            // These exceptions will be removed once we add ExceptionDetails to TokenValidationResult.
+            if (string.IsNullOrEmpty(token))
+                return new TokenValidationResult { Exception = LogHelper.LogArgumentNullException(nameof(token)), IsValid = false };
+
+            if (validationParameters is null)
+                return new TokenValidationResult { Exception = LogHelper.LogArgumentNullException(nameof(validationParameters)), IsValid = false };
+
+            if (token.Length > MaximumTokenSizeInBytes)
+                return new TokenValidationResult { Exception = LogHelper.LogExceptionMessage(new ArgumentException(LogHelper.FormatInvariant(TokenLogMessages.IDX10209, LogHelper.MarkAsNonPII(token.Length), LogHelper.MarkAsNonPII(MaximumTokenSizeInBytes)))), IsValid = false };
```
brentschmaltz, line 50: please stack these parameters.

src/Microsoft.IdentityModel.Tokens/Validation/Results/InternalTokenValidationResult.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        /// <summary>
+        /// Creates a new instance of <see cref="InternalTokenValidationResult"/> to aggregate validation results.
+        /// </summary>
+        /// <param name="securityToken">The <see cref="SecurityToken"/> being validated.</param>
+        /// <param name="tokenHandler">The <see cref="TokenHandler"/> performing the validation.</param>
+        /// <exception cref="ArgumentNullException"></exception>
+        public InternalTokenValidationResult(SecurityToken? securityToken, TokenHandler tokenHandler)
+        {
+            _securityToken = securityToken;
```
brentschmaltz, line 29: Why should securityToken be allowed to be null?

brentschmaltz (approved):
Let's get this committed and then start writing tests and E2E developer experiences.

src/Microsoft.IdentityModel.Tokens/Validation/Results/InternalTokenValidationResult.cs
iNinja: Because there are cases such as validating a token from a string where the reading / decrypting can fail and we end up without a security token, and yet we need to return a result to indicate what went wrong.

FuPingFranco (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
iNinja: For now I have kept it optional and assigned `CancellationToken.None` if it is not present.
We can make this mandatory over the next weeks if we think it is worth enforcing.

PR #2777, opened 2024-08-12
updates for one build

jennyf19:
Ensures that the same build variable name (`MicrosoftIdentityModelVersion`) is used both for the version of the library itself, and in the libraries that take a dependency on Wilson. This name was already used in other libraries. This becomes the way to set the version in Release builds.


build/version.props
```diff
@@ -1,17 +1,22 @@
 <!-- This file may be overwritten by automation. Only values allowed here are VersionPrefix and VersionSuffix.  -->
 <Project>
```
keegan-caruso, line 2: Can we delete this file?

---

Or do we want it just for the one prop?

jennyf19: good catch, I had meant to delete it. it's deleted now.

Wilson.sln
```diff
@@ -105,7 +105,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{BC99A01F
 		build\dependenciesTest.props = build\dependenciesTest.props
 		build\targets.props = build\targets.props
 		build\targetsTest.props = build\targetsTest.props
-		build\version.props = build\version.props
```
keegan-caruso: why remove from the sln?

westin-m (approved):

Wilson.sln
jennyf19: It's deleted

kllysng: I'm sorry, I don't see version.props as deleted

build/version.props
```diff
@@ -1,17 +1,17 @@
...
-    <PropertyGroup Condition=" '$(VersionSuffix)' == '' ">
-      <Version>$(VersionPrefix)</Version>
-    </PropertyGroup>
-
-    <PropertyGroup Condition=" '$(VersionSuffix)' != '' ">
-      <Version>$(VersionPrefix)-$(VersionSuffix)</Version>
-    </PropertyGroup>
-
-</Project>
+    <PreviewVersionSuffix Condition="'$(PreviewVersionSuffix)' == '' and '$(BuildingInsideVisualStudio)' != 'true'">preview-$([System.DateTime]::Now.AddYears(-2019).Year)$([System.DateTime]::Now.ToString("MMddHHmmss"))</PreviewVersionSuffix>
```
JoshLozensky, line 7: Why did we want to add seconds here?

JoshLozensky (approved):

Wilson.sln
jmprieur: @jennyf19, I had readded during the WE it as it's also used by testscommons. But it's now different.
I've rea-added it to the solution too (right now)

jmprieur (approved):

PR #2778, opened 2024-08-13
Add BannedApiAnalyzers to prevent use of ClaimsIdentity constructors

pmaytak:
Add [BannedApiAnalyzers](https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/BannedApiAnalyzers.Help.md) to prevent use of ClaimsIdentity constructors and enforce use of CaseSensitiveClaimsIdentity, which makes claims retrieval consistent. 

The disallowed definitions are contained in BannedSymbols.txt. There is no wildcard to include all constructors for a type, so each one has to be added.

If a banned member is used, the following error is thrown:
`RS0030	The symbol has been marked as banned in this project, and an alternate should be used instead.`


keegan-caruso (approved):

jennyf19 (approved):
:shipit:

PR #2779, opened 2024-08-13
Adding benchmark for new ValidateTokenAsync model vs old

FuPingFranco:
# Adding benchmark for new ValidateTokenAsync model vs old.

## Description

Adding new benchmarks comparing the new ValidateTokenAysnc flow vs old.


benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncWithVPTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            var claimsIdentity = result.ClaimsIdentity;
+            var claims = claimsIdentity.Claims;
+            return claims.ToList();
+        }
+
+        [Benchmark]
+        public async Task<TokenValidationResult> JsonWebTokenHandler_ValidateTokenAsyncWithVP() => await _jsonWebTokenHandler.ValidateTokenAsync(_jwsExtendedClaims, _validationParameters, _callContext, null).ConfigureAwait(false);
+
+        [Benchmark]
+        public async Task<TokenValidationResult> JsonWebTokenHandler_ValidateTokenAsync() => await _jsonWebTokenHandler.ValidateTokenAsync(_jws, _tokenValidationParameters).ConfigureAwait(false);
```
keegan-caruso:
```suggestion
        public async Task<TokenValidationResult> JsonWebTokenHandler_ValidateTokenAsync() => await _jsonWebTokenHandler.ValidateTokenAsync(_jwsExtendedClaims, _tokenValidationParameters).ConfigureAwait(false);
```

FuPingFranco: Nice catch, I was testing the diff and forgot to put it back. Thanks Keegan!

jennyf19 (approved):
This is great, thanks @FuPingFranco 
:shipit:

benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncWithVPTests.cs
pmaytak, line 1: Why not put these tests into the existing ValidateTokenAsync class? Seems like a lot of code is the same and it would be easier to see the usage comparison between old and new APIs.

benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncWithVPTests.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Threading.Tasks;
+using BenchmarkDotNet.Attributes;
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Benchmarks
+{
+    // dotnet run -c release -f net8.0 --filter Microsoft.IdentityModel.Benchmarks.ValidateTokenAsyncWithVPTests*
+
+    public class ValidateTokenAsyncWithVPTests
```
pmaytak: nit: Per [naming guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions#using-abbreviations-and-acronyms) abbreviations shouldn't be used often unless they are well known. In this case, it wasn't obvious what `VP` stood for. So I would maybe spell out `ValidationParameters` or describe in some other way the difference between new vs old behaviors. It does lengthen the names but I think it makes it clearer.

benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncWithVPTests.cs
FuPingFranco: That's a good point but, mainly I created a separate file to delineate that these set of benchmarks are mainly focused on testing the performance of the new model also, there will be more benchmarks to come and it's easier to keep them separate, similar to how we are implementing our work right now and having them in the old file will potentially be automatically ran on the PR pipeline (need to check) which I don't think we want that yet. However, I am open to merge them into one file later on if you think it's cleaner?

benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncWithVPTests.cs
FuPingFranco: This was merged into one file so longer needed. Thanks Peter

benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncWithVPTests.cs
keegan-caruso: If this is the goal it might make more sense to restructure this a bit so that we have one to one comparisons with baselines?

keegan-caruso: For example: https://benchmarkdotnet.org/articles/samples/IntroCategoryBaseline.html#sample-introcategorybaseline

pmaytak: I suggest to combine them - it'll help reduce the code duplication and help to see the full picture of our public validate API. We also already combine some other benchmarks like JwtAuthenticationHandler and ProtocolHandler.

FuPingFranco: Updated PR with suggestions. Thanks @pmaytak & @keegan-caruso!

iNinja (approved):
LGTM

PR #2762, opened 2024-08-05
Revert "Remove SlimLock when updating metadata. (#2751)"

keegan-caruso:
This reverts commit bbc09a4811ba2fd4eb58edb693114c3f8a0fd0a1.

#2751 introduces a bug where on startup multiple calls can concurrently fetch from the document retriever

PR #2780, opened 2024-08-13
Add lock when configuration is null

brentschmaltz:
Metadata is now updated on a background Task.
Previously updating was behind a SlimLock on every update to the metadata endpoint which resulted in a blocking call when RequestRefresh or AutomaticRefresh were called.

The lock is applied only when configuration has not been obtained, this stops the case when a number of threads startup and each thread would hit the metadata endpoint.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -19,16 +19,22 @@ namespace Microsoft.IdentityModel.Protocols
     public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class
     {
         private DateTimeOffset _syncAfter = DateTimeOffset.MinValue;
-        private DateTimeOffset _lastRefresh = DateTimeOffset.MinValue;
+        private DateTimeOffset _lastRequestRefresh = DateTimeOffset.MinValue;
         private bool _isFirstRefreshRequest = true;
+        private readonly SemaphoreSlim _configurationNullLock = new SemaphoreSlim(1);
 
-        private readonly SemaphoreSlim _refreshLock;
         private readonly IDocumentRetriever _docRetriever;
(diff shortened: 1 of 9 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -144,83 +149,174 @@ public async Task<T> GetConfigurationAsync()
...
+                                LogHelper.FormatInvariant(
+                                    LogMessages.IDX20806,
+                                    LogHelper.MarkAsNonPII(MetadataAddress ?? "null"),
+                                    LogHelper.MarkAsNonPII(ex)),
+                                ex));
+                    }
+                }
+                finally
+                {
+                    _configurationNullLock.Release();
```
keegan-caruso, line 246: this could be in a much narrower try finally block so on the typical path after the first fetch the try finally is never entered.

brentschmaltz: @keegan-caruso 
Not following, the finally is associated with GetConfigurationAsync and the ConfigValidator.
If either of those throws, we need to enter the finally.

The try will only get entered if configuration has never been obtained.

keegan-caruso: Ah, I'm sorry, I read the diff wrong. I follow now.

keegan-caruso (approved):

westin-m (approved):

shaheer-k:
@brentschmaltz 
The SimLock was reverted in #2762 by @keegan-caruso 
But it seems like its added again in #2780 
This change is creating issues in our application as the configuration not seems to be refreshd incase of changes in the JWKS metadata. We are not able to upgrade to 8.0.2 due to this issue. Our tests are failing when upgrading to 8.0.2 due to this particualr issue. Please suggest what changes we should be applying to get it work again in version 8.0.2 onwards

PR #2784, opened 2024-08-16
Move the version to version.projs a it's also included by tests

jmprieur:
Move the version processing to version.props a it's also included by tests


jennyf19 (approved):
ty
:shipit:

PR #2788, opened 2024-08-20
Remove lock when creating a SignatureProvider

brentschmaltz:
Remove the lock on creating a SignatureProvider an allow a rare case where a SignatureProvider is created and not cached.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
+++ b/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
@@ -16,7 +16,6 @@ namespace Microsoft.IdentityModel.Tokens
     {
         private static CryptoProviderFactory _default;
         private static readonly ConcurrentDictionary<string, string> _typeToAlgorithmMap = new ConcurrentDictionary<string, string>();
-        private static readonly object _cacheLock = new object();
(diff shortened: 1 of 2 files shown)


src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -591,25 +590,23 @@ private SignatureProvider CreateSignatureProvider(SecurityKey key, string algori
                     return signatureProvider;
                 }
 
-                lock (_cacheLock)
-                {
-                    if (CryptoProviderCache.TryGetSignatureProvider(key, algorithm, typeofSignatureProvider, willCreateSignatures, out signatureProvider))
-                    {
-                        signatureProvider.AddRef();
```
keegan-caruso, line 598: Could you elaborate on why we no longer need the AddRef call?

brentschmaltz: We do AddRef above on line 589.

We don't AddRef below as when created the ref count is 1.
Then if we fail to add to the cache, then we leave the count as one.

keegan-caruso (approved):

jennyf19 (approved):
:shipit:

PR #2791, opened 2024-08-22
Update version.props to 8.0.2

jennyf19:


build/version.props
```diff
@@ -2,7 +2,7 @@
 <Project>
 <!-- MicrosoftIdentityModelVersion  -->
   <PropertyGroup>
-    <MicrosoftIdentityModelCurrentVersion>8.0.1</MicrosoftIdentityModelCurrentVersion>
+    <MicrosoftIdentityModelCurrentVersion>8.0.2</MicrosoftIdentityModelCurrentVersion>
```
jennyf19:
```suggestion
    <MicrosoftIdentityModelCurrentVersion>8.0.3</MicrosoftIdentityModelCurrentVersion>
```

build/version.props
```diff
@@ -2,7 +2,7 @@
 <Project>
 <!-- MicrosoftIdentityModelVersion  -->
   <PropertyGroup>
-    <MicrosoftIdentityModelCurrentVersion>8.0.1</MicrosoftIdentityModelCurrentVersion>
+    <MicrosoftIdentityModelCurrentVersion>8.0.3</MicrosoftIdentityModelCurrentVersion>
```
jennyf19, line 5: should be the next version, not the current version, so the previews are correct.

PR #2799, opened 2024-08-29
Exception refinement: Adding additional information.

iNinja:
# Exception refinement: Adding additional information.

Refining the agreed model to simplify where possible and match the original validation path's output.

## Description
- Removed generic error from Result type. All uses share the same constant error type. Removing it from the generic simplifies our interfaces and makes the code easier to read.
- Restored use of `Activator` to instantiate the exceptions in an AOT compatible manner.
- Added initial proposal of extending ExceptionDetail to assign additional information to exceptions

## Notes:
For the additional data, 3 approaches were evaluated:

- Creating specialised result types: this results in a lot of code duplication if we want to keep our value type result, or an extra allocation if we turn it into a class to use inheritance.
- Passing a dictionary to ExceptionDetail that can then be used to extract the information in each exception results in an extra allocation of the dictionary.
- Extending ExceptionDetail to allow the provision of specialised classes that receive the extra data and assign it only when the exception is instantiated. **This was the chosen path forward for this PR** since it avoids any new allocations and allows extensibility between ExceptionDetail subclasses and Exception subclasses. It makes no assumptions about the types.

This PR is raised as a draft to discuss the approach, and if we agree, I will extend the approach to the remaining exceptions before removing the draft status.

Part of #2711


iNinja:
Cherry picking commits to clean history and raising again

PR #2801, opened 2024-08-29
Added sample tests for the new validation path

iNinja:
# Added sample tests for the new validation path

## Description
Added sample tests for new validation path and aligned errors returned.

Part of #2711


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -261,10 +261,7 @@ private static Result<SecurityKey, ExceptionDetail> ValidateSignatureWithKey(
                 else
                     return new ExceptionDetail(
                         new MessageDetail(
-                            TokenLogMessages.IDX10504,
-                            LogHelper.MarkAsSecurityArtifact(
-                                jsonWebToken.EncodedToken,
-                                JwtTokenUtilities.SafeLogJwtToken)),
+                            TokenLogMessages.IDX10500),
```
brentschmaltz: Why did we change this to not return the token?

iNinja: Because 10500 is: "IDX10500: Signature validation failed. No security keys were provided to validate the signature.";

I can keep it as 10504, but I aligned it to return the exact same error that the current path does.

test/Microsoft.IdentityModel.SampleTests/SampleTokenValidationClass.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.IdentityModel.Tokens.Jwt;
 using System.Security.Claims;
+using System.Threading;
```
brentschmaltz, line 7: Why do we need a new assembly?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
brentschmaltz: But we lost the EncodedToken, please change it back.

test/Microsoft.IdentityModel.SampleTests/SampleTokenValidationClassTests.cs
```diff
@@ -221,6 +222,218 @@ public void TokenWithMissingSecurityCredentials()
...
+                testTokenCreator.CreateExpiredToken,
+                typeof(SecurityTokenExpiredException),
+                "IDX10223");
+        }
+
+        /// <summary>
+        /// Tests how the class under test handles a token which is not yet valid
+        /// </summary>
+        [Fact]
+        public void NetYetValidToken_NewPath()
```
brentschmaltz, line 288: This is too fractured.
We need sets of tests using TheoryData for Lifetime, Audience, etc.
Why not part of JsonWebTokenHandler.ValidateLifetimeTests

test/Microsoft.IdentityModel.SampleTests/SampleTokenValidationClass.cs
iNinja: CancellationToken requires it.

iNinja:
These tests are being removed.

PR #2803, opened 2024-08-30
Removed enum for ExceptionType

brentschmaltz:
The enums that map to an Exception to create have been removed. This creates a dependency in the code, benchmarks show very little change.

Reverted change related to LastKnownGood in that modified the release call graph.
Introduced a model for a derived ExceptionDetail coordinated with an additional ctor for Exceptions.


FuPingFranco (approved):
LGTM

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
```diff
@@ -150,7 +150,7 @@ await ValidateJWEAsync(jsonWebToken, validationParameters, currentConfiguration,
                 return result;
             }
 
-            if (TokenUtilities.IsRecoverableErrorType(result.UnwrapError().Type))
+            if (TokenUtilities.IsRecoverableException(result.UnwrapError().GetException()))
```
iNinja, line 153: Why are we allocating the exception when we could simply check the type and save the allocation?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
```diff
@@ -182,13 +182,13 @@ await ValidateJWEAsync(jsonWebToken, validationParameters, currentConfiguration,
                 {
                     validationParameters.RefreshBeforeValidation = false;
                     validationParameters.ValidateWithLKG = true;
-                    ExceptionType recoverableExceptionType = result.UnwrapError().Type;
+                    Exception recoverableExceptionType = result.UnwrapError().GetException();
```
iNinja, line 185: Same question around allocating the exception.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ExceptionDetail.cs
```diff
@@ -53,19 +57,60 @@ public ExceptionDetail(
         /// Creates an instance of an <see cref="Exception"/> using <see cref="ExceptionDetail"/>
         /// </summary>
         /// <returns>An instantance of an Exception.</returns>
-        public Exception GetException()
+        public virtual Exception GetException()
```
iNinja, line 60: The Activator used in this PR: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2800, is more extensible than this approach. Why are we resorting to still using a bunch of ifs checking for the different exceptions?

One more thing to consider is that the `ArgumentNullException` not inheriting from `SecurityTokenException` and therefore doesn't have a stack trace.

iNinja:
This PR is re-doing the work done in a previous PR, https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2800.

I am not seeing the point on merging these changes over the linked PR. I am happy to take the approach to initialise exceptions with the ExceptionDetail presented here onto that one.

brentschmaltz:
Added changes to https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2801.

Will pull in additional work around adding ExceptionDetails to SecurityTokenException.

PR #2804, opened 2024-08-30
Add cloud instance name validation

### LINKED ISSUE #2832: Support cloud instance name validation
* Add a new extension method EnableEntraIdSigningKeyCloudInstanceValidation that adds additional validation rule to the IssuerSigningKeyValidatorUsingConfiguration delegate call chain.
* New validation rule checks that the cloud instance name of the openid-configuration matches the cloud instance name of the signing key.
* SecurityTokenInvalidCloudInstanceException (new exception) should be thrown in case check is failed.

alexholub113:
# Added optional check to prevent using keys that are shared across multiple clouds.

- Added a new extension method EnableEntraIdSigningKeyCloudInstanceNameValidation that adds additional validation to the IssuerSigningKeyValidatorUsingConfiguration delegate call chain.
- New validation checks that the cloud instance name of the openid-configuration matches the cloud instance name of the signing key.
- New exception SecurityTokenInvalidCloudInstanceException is thrown in case check is failed.

Resolves #2832


src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenInvalidCloudInstanceException.cs
```diff
@@ -0,0 +1,92 @@
...
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// This exception is thrown when specific cloud instance was not matched with cloud instance of signing key.
+    /// </summary>
+    [Serializable]
+    public class SecurityTokenInvalidCloudInstanceException : SecurityTokenValidationException
```
GeoK: Perhaps extend from SecurityTokenInvalidSigningKeyException

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -17,21 +17,51 @@ namespace Microsoft.IdentityModel.Validators
     /// </summary>
     public static class AadTokenValidationParametersExtension
     {
+        /// <summary>
+        /// Enables the validation of the signing keys used by the Microsoft identity platform (AAD) against the token.
+        /// </summary>
+        /// <param name="tokenValidationParameters">The <see cref="TokenValidationParameters"/> that are used to validate the token.</param>
+        /// <param name="cloudInstanceName">The optional cloud instance name to validate against.</param>
+        public static void EnableAadSigningKeyValidation(this TokenValidationParameters tokenValidationParameters, string cloudInstanceName = null)
```
GeoK: Let's rename "Aad" to "EntraId"

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -109,6 +138,45 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
...
+                if (string.IsNullOrWhiteSpace(signingKeyCloudInstance))
+                    return true;
+
+                // Store the cloud instance name in the security key's property bag.
+                securityKey.PropertyBag[OpenIdProviderMetadataNames.CloudInstanceName] = signingKeyCloudInstance;
+
+                if (cloudInstanceName == null)
+                    return true;
+
+                if (signingKeyCloudInstance != cloudInstanceName)
```
GeoK: Let's be explicit on the string comparison here (ordinal).

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -109,6 +138,45 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
...
+                    throw LogHelper.LogExceptionMessage(
+                        new SecurityTokenInvalidCloudInstanceException(LogHelper.FormatInvariant(LogMessages.IDX40012, LogHelper.MarkAsNonPII(cloudInstanceName), LogHelper.MarkAsNonPII(signingKeyCloudInstance)))
```
GeoK: If you decide on extending from SecurityTokenInvalidSigningKeyException, don't forget to set the signing key on the exception.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -109,6 +138,45 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
...
         }
 
+        /// <summary>
+        /// Stores the cloud instance name of the signing key in a property bag and validates it.
+        /// </summary>
+        /// <param name="securityKey">The <see cref="SecurityKey"/> that signed the <see cref="SecurityToken"/>.</param>
+        /// <param name="configuration">The <see cref="BaseConfiguration"/> provided.</param>
+        /// <param name="cloudInstanceName">The cloud instance name to validate against.</param>
+        /// <returns><c>true</c> if the cloud instance name of the signing key is valid; otherwise, <c>false</c>.</returns>
+        internal static bool StoreValidateSigningKeyCloudInstanceName(SecurityKey securityKey, BaseConfiguration configuration, string cloudInstanceName)
```
GeoK: No need for this method to return a bool if it is not inspected.

GeoK: nit: Consider dropping the "Store" prefix as putting the instance in a property bag is an implementation detail.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -17,21 +17,51 @@ namespace Microsoft.IdentityModel.Validators
     /// </summary>
     public static class AadTokenValidationParametersExtension
     {
+        /// <summary>
+        /// Enables the validation of the signing keys used by the Microsoft identity platform (AAD) against the token.
```
GeoK: Perhaps something like: Enables validation of Microsoft Entra ID token signing keys.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: hm... right now the SecurityTokenInvalidCloudInstanceException doesn't extend SecurityTokenInvalidSigningKeyException and I don't see an reason why it should

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: fixed

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: fixed

alexholub113: fixed

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: fixed

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: fixed

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: fixed

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenInvalidCloudInstanceException.cs
alexholub113: fixed

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -17,21 +17,51 @@ namespace Microsoft.IdentityModel.Validators
...
+
+                return ValidateIssuerSigningKeyCertificate(securityKey, tvp);
+            };
+        }
+
         /// <summary>
         /// Enables the validation of the issuer of the signing keys used by the Microsoft identity platform (AAD) against the issuer of the token.
         /// </summary>
         /// <param name="tokenValidationParameters">The <see cref="TokenValidationParameters"/> that are used to validate the token.</param>
+        [Obsolete("Use EnableAadSigningKeyValidation(TokenValidationParameters, string) instead.")]
```
GeoK: This got changed to EnableEntraIdSigningKeyValidation.

test/Microsoft.IdentityModel.Validators.Tests/AadTokenValidationParametersExtensionTests.cs
```diff
@@ -19,11 +19,187 @@
...
+                ValidateAudience = false,
+                ValidateLifetime = false
+            };
+
+            return theoryData;
+        }
+
         [Theory, MemberData(nameof(EnableAadSigningKeyIssuerValidationTestCases))]
-        public async Task EnableAadSigningKeyIssuerValidationTests(AadSigningKeyIssuerTheoryData theoryData)
+        [Obsolete("The 'EnableAadSigningKeyIssuerValidation' extension method is obsolete.")]
```
GeoK: Can we suppress the obsolete warning here (CS0618) for consistence?

src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
```diff
@@ -15,6 +17,7 @@ public abstract class SecurityKey
         private CryptoProviderFactory _cryptoProviderFactory;
         private object _internalIdLock = new object();
         private string _internalId;
+        private IDictionary<string, object> _propertyBag;
```
keegan-caruso: This should be the concrete type instead of the interface for a private field.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -17,21 +17,51 @@ namespace Microsoft.IdentityModel.Validators
     /// </summary>
     public static class AadTokenValidationParametersExtension
     {
+        /// <summary>
+        /// Enables validation of Microsoft Entra ID token signing keys.
+        /// </summary>
+        /// <param name="tokenValidationParameters">The <see cref="TokenValidationParameters"/> that are used to validate the token.</param>
+        /// <param name="cloudInstanceName">The optional cloud instance name to validate against.</param>
+        public static void EnableEntraIdSigningKeyValidation(this TokenValidationParameters tokenValidationParameters, string cloudInstanceName = null)
```
brentschmaltz: Default parameters on a public method can lead to problems.
If we want to add a new method with one more parameter, this leads to a runtime or compile time break.

When would the cloudInstanceName be null?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdProviderMetadataNames.cs
```diff
@@ -46,6 +46,7 @@ public static class OpenIdProviderMetadataNames
         public const string IntrospectionEndpointAuthSigningAlgValuesSupported = "introspection_endpoint_auth_signing_alg_values_supported";
         public const string JwksUri = "jwks_uri";
         public const string Issuer = "issuer";
+        public const string CloudInstanceName = "cloud_instance_name";
```
brentschmaltz: Is "CloudInstanceName" in a discovery spec?

Probably belongs in the extension file.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -17,21 +17,51 @@ namespace Microsoft.IdentityModel.Validators
...
+
+                return ValidateIssuerSigningKeyCertificate(securityKey, tvp);
+            };
+        }
+
         /// <summary>
         /// Enables the validation of the issuer of the signing keys used by the Microsoft identity platform (AAD) against the issuer of the token.
         /// </summary>
         /// <param name="tokenValidationParameters">The <see cref="TokenValidationParameters"/> that are used to validate the token.</param>
+        [Obsolete("Use EnableEntraIdSigningKeyValidation(TokenValidationParameters, string) instead.")]
```
brentschmaltz: Do we need to Obsolete this?

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -49,28 +79,27 @@ public static void EnableAadSigningKeyIssuerValidation(this TokenValidationParam
...
             if (securityToken == null)
                 throw LogHelper.LogArgumentNullException(nameof(securityToken));
 
-            var openIdConnectConfiguration = configuration as OpenIdConnectConfiguration;
-            if (openIdConnectConfiguration == null)
+            if (configuration is not OpenIdConnectConfiguration openIdConnectConfiguration)
                 return true;
 
-            var matchedKeyFromConfig = openIdConnectConfiguration.JsonWebKeySet?.Keys.FirstOrDefault(x => x != null && x.Kid == securityKey.KeyId);
+            JsonWebKey matchedKeyFromConfig = openIdConnectConfiguration.JsonWebKeySet?.Keys.FirstOrDefault(x => x != null && x.Kid == securityKey.KeyId);
```
brentschmaltz: We duplicate this linq code to find the securitykey in ValidateSigningKeyIssuer and ValidateSigningKeyCloudInstanceName. Since this code is on the hot path, we should only find the securitykey once.

We should replace the linq code with simple loops and compares.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -109,6 +138,43 @@ internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityT
...
+
+            JsonWebKey matchedKeyFromConfig = openIdConnectConfiguration.JsonWebKeySet?.Keys.FirstOrDefault(x => x != null && x.Kid == securityKey.KeyId);
+            if (matchedKeyFromConfig != null && matchedKeyFromConfig.AdditionalData.TryGetValue(OpenIdProviderMetadataNames.CloudInstanceName, out object value))
+            {
+                string signingKeyCloudInstance = value as string;
+                if (string.IsNullOrWhiteSpace(signingKeyCloudInstance))
+                    return;
+
+                // Store the cloud instance name in the security key's property bag.
+                securityKey.PropertyBag[OpenIdProviderMetadataNames.CloudInstanceName] = signingKeyCloudInstance;
```
keegan-caruso: Let's think about thread safety of this collection.

test/Microsoft.IdentityModel.Validators.Tests/AadTokenValidationParametersExtensionTests.cs
alexholub113: Fixed

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: Fixed

src/Microsoft.IdentityModel.Tokens/SecurityKey.cs
alexholub113: removed this code

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: removed this param

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdProviderMetadataNames.cs
alexholub113: Thanks, fixed.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: removed this

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: thanks, this code was reverted

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: I changed it a bit, now we don't have single extension method that calls ValidateSigningKeyIssuer and ValidateSigningKeyCloudInstanceName. So, there is nothing to share.

about linq
since the JsonWebKeySet or Keys may be null we have to check for null this properties, so the 'loop' version would not differ from FirstOrDefault.

so, I left FirstOrDefault.
Please have a look and let me know if we can improve it

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -17,6 +17,32 @@ namespace Microsoft.IdentityModel.Validators
...
+            tokenValidationParameters.IssuerSigningKeyValidatorUsingConfiguration = (securityKey, securityToken, tvp, config) =>
+            {
+                // preserve and run provided logic
+                if (userProvidedIssuerSigningKeyValidatorUsingConfiguration != null)
+                    return userProvidedIssuerSigningKeyValidatorUsingConfiguration(securityKey, securityToken, tvp, config);
+
+                if (userProvidedIssuerSigningKeyValidator != null)
+                    return userProvidedIssuerSigningKeyValidator(securityKey, securityToken, tvp);
+
+                ValidateSigningKeyCloudInstanceName(securityKey, config);
```
GeoK: This check must run before user-provided delegates, otherwise it might not run at all.

alexholub113: good catch, I'll add test for that

GeoK (approved):

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenInvalidCloudInstanceNameException.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Runtime.Serialization;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// This exception is thrown when the cloud instance name of the signing key was not matched with the cloud instance name from configuration.
+    /// </summary>
+    [Serializable]
+    public class SecurityTokenInvalidCloudInstanceNameException : SecurityTokenInvalidSigningKeyException
+    {
```
brentschmaltz: Suggest the name: SecurityTokenInvalidCloudInstanceException  (drop the 'Name') as there may be other reasons in the future.

@GeoK not sure about the need to inherit from SecurityTokenInvalidSigningKeyException, perhaps SecurityTokenException

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -17,6 +17,33 @@ namespace Microsoft.IdentityModel.Validators
     /// </summary>
     public static class AadTokenValidationParametersExtension
     {
+        /// <summary>
+        /// Enables validation of the cloud instance name of the Microsoft Entra ID token signing keys.
+        /// </summary>
+        /// <param name="tokenValidationParameters">The <see cref="TokenValidationParameters"/> that are used to validate the token.</param>
+        public static void EnableEntraIdSigningKeyCloudInstanceNameValidation(this TokenValidationParameters tokenValidationParameters)
```
brentschmaltz: Perhaps shorten the name to: EnableEntraIdCoudInstanceValidation OR EnableEntraCloudInstanceValidation.
Not sure we need to be specific about Id or specify name as may expand critera.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
brentschmaltz: @alexholub113 it has been shown that linq is much slower that simple loops.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -17,6 +17,33 @@ namespace Microsoft.IdentityModel.Validators
...
+        {
+            if (tokenValidationParameters == null)
+                throw LogHelper.LogArgumentNullException(nameof(tokenValidationParameters));
+
+            IssuerSigningKeyValidatorUsingConfiguration userProvidedIssuerSigningKeyValidatorUsingConfiguration = tokenValidationParameters.IssuerSigningKeyValidatorUsingConfiguration;
```
brentschmaltz, line 30: I think this can be written without these local variables.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
```diff
@@ -17,6 +17,33 @@ namespace Microsoft.IdentityModel.Validators
...
+
+            tokenValidationParameters.IssuerSigningKeyValidatorUsingConfiguration = (securityKey, securityToken, tvp, config) =>
+            {
+                ValidateSigningKeyCloudInstanceName(securityKey, config);
+
+                // preserve and run provided logic
+                if (userProvidedIssuerSigningKeyValidatorUsingConfiguration != null)
+                    return userProvidedIssuerSigningKeyValidatorUsingConfiguration(securityKey, securityToken, tvp, config);
+
+                if (userProvidedIssuerSigningKeyValidator != null)
```
brentschmaltz, line 40: I don't think we should call this delegate (without config), just userProvidedIssuerSigningKeyValidatorUsingConfiguration.

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: makes sense, I added new method GetJsonWebKeyBySecurityKey that using loop

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: renamed to EnableEntraIdCoudInstanceValidation

src/Microsoft.IdentityModel.Validators/AadTokenValidationParametersExtension.cs
alexholub113: These variables are necessary to avoid infinite recursion

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenInvalidCloudInstanceNameException.cs
alexholub113: Renamed class.
in my opinion SecurityTokenInvalidCloudInstanceNameException is still SecurityTokenInvalidSigningKeyException but more specific

GeoK: Discussed this with Brent offline - it is ok as-is.

brentschmaltz (approved):

PR #2807, opened 2024-09-04
[Performance] Avoid lock contention for getting count of inner dictionary size

ms-ruochenyu:
# [Performance] Avoid lock contention for getting count of inner dictionary size

Avoid lock contention for getting count of inner dictionary size in EventBasedLRUCache.

## Description
We have seen bottlenecks in some of internal cloud services where `EventBasedLRUCache<TKey, TValue>` calls Count() for its inner `_map`. This property calculation brings with heavy lock contention, espeicial in `SetValue()` and `Compact()`

This change adds a new private field `int this._mapSize` to be updated during entries in `_map` being updated. `_mapSize` will be updated as atomic and only used to new cache size calculation so, there should not be beheavior change from observation outside.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
+++ b/src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
@@ -43,6 +43,8 @@ namespace Microsoft.IdentityModel.Tokens
         // if true, then items will be maintained in a LRU fashion, moving to front of list when accessed in the cache.
         private readonly bool _maintainLRU;
         private ConcurrentDictionary<TKey, LRUCacheItem<TKey, TValue>> _map;
+        // ConcurrentDictionary<TK,TV>.get_Count() takes a lock where performance is a concern, so we maintain a separate count updated by atomic for size calculation.
+        private int _mapSize;
         // When the current cache size gets to this percentage of _capacity, _compactionPercentage% of the cache will be removed.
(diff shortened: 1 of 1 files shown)


ms-ruochenyu:

keegan-caruso:
@ms-ruochenyu Can you please open and link an issue to this?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -498,6 +511,7 @@ public bool SetValue(TKey key, TValue value, DateTime expirationTime)
                 }
 
                 _map[key] = newCacheItem;
+                Interlocked.Increment(ref _mapSize);
```
brentschmaltz, line 514: _map[key] = newCacheItem, does not imply the count has increased.

brentschmaltz:
@ms-ruochenyu what is the performance gain?

ms-ruochenyu:
> @ms-ruochenyu what is the performance gain?

@brentschmaltz, below is a sample we collected in 1 of internal workload. The CPU cost was clustered to taking bucket write lock inside `ConcurrentDictionary.Count()`. So does `Compact()` call. image

brentschmaltz:
@ms-ruochenyu i am unable to understand your CPU diagram.

keegan-caruso:
@ms-ruochenyu - Please open a tracking issue for this.

PR #2810, opened 2024-09-04
Add initial regression tests for the new validation path

iNinja:
# Add initial regression tests for the new validation path
## Description
Added theory based test for comparing the ValidationParameters based new token validation path against JsonWebTokenHandler's original path with TokenValidationParameters.

More tests are being added as a follow up.

Part of #2711


test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerValidationParametersTests.cs
```diff
@@ -0,0 +1,240 @@
...
+                    EncryptingCredentials = theoryData.EncryptingCredentials,
+                    AdditionalHeaderClaims = theoryData.AdditionalHeaderParams,
+                    Audience = theoryData.Audience,
+                    Issuer = theoryData.Issuer,
+                };
+
+                jwtString = jsonWebTokenHandler.CreateToken(securityTokenDescriptor);
+            }
+
+            TokenValidationResult tokenValidationResult = await jsonWebTokenHandler.ValidateTokenAsync(jwtString, theoryData.TokenValidationParameters);
```
keegan-caruso: nit: we could use better names, perhaps

tokenValidationParameterResult and validationParameterResult

It makes reading the rest of the method easier.

keegan-caruso (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerValidationParametersTests.cs
iNinja: Updated.

brentschmaltz (approved):

PR #2811, opened 2024-09-05
Regression tests: Added inner exception detail and invalid signature failure due to invalid algorithm used

iNinja:
# Regression tests: Added inner exception detail and invalid signature failure due to invalid algorithm used

- Added inner exception detail to provide inner exceptions without allocating them ahead of time
- Wrap invalid algorithm validation error inside an invalid signature validation error to match the existing path
- Added test case comparing the invalid algorithm on the signature validation error

Part of #2711

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ExceptionDetail.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ExceptionDetail.cs
@@ -28,7 +28,7 @@ namespace Microsoft.IdentityModel.Tokens
             ValidationFailureType failureType,
             [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type exceptionType,
             StackFrame stackFrame)
-            : this(MessageDetail, failureType, exceptionType, stackFrame, null)
+            : this(MessageDetail, failureType, exceptionType, stackFrame, innerException: null)
         {
         }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ExceptionDetail.cs
```diff
@@ -57,17 +57,37 @@ public ExceptionDetail(
...
+            ValidationFailureType failureType,
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type exceptionType,
+            StackFrame stackFrame,
+            ExceptionDetail innerExceptionDetail)
+        {
+            InnerExceptionDetail = innerExceptionDetail;
+            MessageDetail = messageDetail;
+            _exceptionType = exceptionType;
+            FailureType = failureType;
+            StackFrames = new List<StackFrame>(4)
```
keegan-caruso, line 71: Can you just do?

```suggestion
            StackFrames = [stackFrame];
```

If not, why default to a list of size 4?

FuPingFranco (approved):
LGTM

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ExceptionDetail.cs
iNinja: I was checking the allocations involved into using lists, and it appears that they start with a size of 0 and then double each time they reach their limit.
So doing `[stackFrame]`  would allocate it with a size of 1, then the next time a stack frame is added it would reallocate to 2, then 4, then 8, and so on.
I put the number 4 in a very precarious way while experimenting with it, and I believe we should consider what the minimum size of a stack trace would be for us to avoid re-allocating too many times as we add the different frames.

keegan-caruso: That is fine.

It is interesting that you ended up at 4: https://github.com/dotnet/runtime/blob/66e45ef152002a2e1cbea939ded6643dfa48307c/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L23

It is the default if you don't specify an initial size. 

Edit for clarity:

It does start at 0, but grows to 4 on the first add

keegan-caruso (approved):

Issue #2695, opened 2024-07-10
[Bug] Unknow reasons for Microsoft.IdentityModel.Tokens.SecurityTokenKeyWrapException: IDX10618: Key unwrap failed using decryption Keys


maomaomqiu:
Hi, team

We use WebAuth, but we found exception from Microsoft.identity.ServiceEssentials.Core.
This case is pretty rare. Also, we found other team encounter such exceptions. (In their cases, they also encounter such issue, and the probability also rare)

Once the little probability exception happened, it will bring catastrophic result.

# Package reference chain:

WebAuth -> MISE -> SAL -> Wilson (Exception happened here)

# Error Msg
Message: 'MISE12018: MiseHost (1.22.2.0): MISE12014: The request failed with exception: Microsoft.Identity.ServiceEssentials.Exceptions.MiseAuthenticationTicketProviderException: (layer1)

Component: AuthenticationTicketProvider:1.22.2.0

CorrelationId:8225b67f-b9a6-44b9-aaa6-b7503f1a13aa

Microsoft.Identity.ServiceEssentials.Exceptions.MiseAuthenticationTicketProviderException: MISE12034: AuthenticationTicketProvider (layer1)Name:AuthenticationTicketProvider, GetVersion:1.22.2.0.

 ---> System.AggregateException: S2S12096: Microsoft.IdentityModel.S2S.S2SAuthenticationManager caught exceptions when validating the token. See AuthenticationResult.InboundPolicyEvaluationResults for additional details. (S2S12086: An exception has been caught while validating the request applying the policy with id : 'c3a6fb3d-2f0a-4e6b-858a-406bbb4c6fdc'. Exception: Microsoft.IdentityModel.Tokens.SecurityTokenKeyWrapException: IDX10618: Key unwrap failed using decryption Keys: '[PII of type 'System.Text.StringBuilder' is hidden. For more details, see [https://aka.ms/IdentityModel/PII.]'](https://aka.ms/IdentityModel/PII.%5d').

Exceptions caught:

 '[PII of type 'System.Text.StringBuilder' is hidden. For more details, see [https://aka.ms/IdentityModel/PII.]'](https://aka.ms/IdentityModel/PII.%5d').

[...]

PR #2812, opened 2024-09-05
Return unwrapped keys if able

keegan-caruso:
## Description

JsonWebTokenHandler would only return unwrapped keys if there was no errors. This change is to align with the behavior in JwtSecurityTokenHandler, that is it returns the keys that were able to be unwrapped, and only throw if no keys were able to be unwrapped.

Relates to #2695

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
@@ -1422,7 +1422,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 (keysAttempted ??= new StringBuilder()).AppendLine(key.ToString());
             }
 
-            if (unwrappedKeys.Count > 0 && exceptionStrings is null)
+            if (unwrappedKeys.Count > 0 || exceptionStrings is null)
                 return unwrappedKeys;
             else
                 throw LogHelper.LogExceptionMessage(new SecurityTokenKeyWrapException(LogHelper.FormatInvariant(TokenLogMessages.IDX10618, (object)keysAttempted ?? "", (object)exceptionStrings ?? "", jwtToken)));
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
@@ -202,7 +202,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 (keysAttempted ??= new StringBuilder()).AppendLine(key.ToString());
             }
 
-            if (unwrappedKeys.Count > 0 && exceptionStrings is null)
+            if (unwrappedKeys.Count > 0 || exceptionStrings is null)
                 return (unwrappedKeys, null);
             else
             {
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
@@ -64,7 +64,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 return result.exceptionDetail.AddStackFrame(decryptionGetKeysStackFrame);
             }
 
-            if (result.contentEncryptionKeys == null)
+            if (result.contentEncryptionKeys == null || result.contentEncryptionKeys.Count == 0)
             {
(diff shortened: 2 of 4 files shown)


test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -710,7 +712,6 @@ public static TheoryData<CreateTokenTheoryData> SecurityTokenDecryptionTheoryDat
                             EncryptingCredentials = KeyingMaterial.DefaultSymmetricEncryptingCreds_Aes128_Sha2,
                             Claims = Default.PayloadDictionary
                         },
-                        JsonWebTokenHandler = new JsonWebTokenHandler(),
```
keegan-caruso, line 713: I removed the properties that aren't being used

sruke:
Should the logic [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/0198cb0255bd9cd06ad1fb0d4a502192f2526d38/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs#L205) in JsonWebTokenHandler.DecryptToken be updated as well?

keegan-caruso:
> Should the logic [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/0198cb0255bd9cd06ad1fb0d4a502192f2526d38/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs#L205) in JsonWebTokenHandler.DecryptToken be updated as well?

Yeah, good idea. done.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.DecryptTokenTests.cs
```diff
@@ -193,10 +194,42 @@ public static TheoryData<TokenDecryptingTheoryData> JsonWebTokenHandlerDecryptTo
                             ValidationFailureType.TokenDecryptionFailed,
                             typeof(SecurityTokenDecryptionFailedException),
                             null),
+                    },
+                    new TokenDecryptingTheoryData
```
pmaytak, line 199: This needs another test case where `unwrappedKeys` is 0 similar to AlgorithmMatch in the other test class.

pmaytak (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.DecryptTokenTests.cs
```diff
@@ -193,10 +194,42 @@ public static TheoryData<TokenDecryptingTheoryData> JsonWebTokenHandlerDecryptTo
...
+                    {
+                        TestId = "OneKeyThrowsOnUnwrap_DecryptionSucceeds",
```
iNinja: Can we rename this to follow the "Valid_Conditions" and "Invalid_Conditions" pattern?
I assume something like "Valid_OneKeyThrowsOnUnwrap_DecryptionSucceeds" would work to make clear what we are testing for.

iNinja (approved):
Small comment, otherwise looks good.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
```diff
@@ -1422,7 +1422,7 @@ internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken
                 (keysAttempted ??= new StringBuilder()).AppendLine(key.ToString());
             }
 
-            if (unwrappedKeys.Count > 0 && exceptionStrings is null)
+            if (unwrappedKeys.Count > 0 || exceptionStrings is null)
```
brentschmaltz, line 1425: Why do we care about exceptionStrings if we have found any possible keys?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -202,7 +202,7 @@ internal Result<string> DecryptToken(
                 (keysAttempted ??= new StringBuilder()).AppendLine(key.ToString());
             }
 
-            if (unwrappedKeys.Count > 0 && exceptionStrings is null)
+            if (unwrappedKeys.Count > 0 || exceptionStrings is null)
```
brentschmaltz, line 205: Why do we care about exceptionStrings if we have some keys?

PR #2813, opened 2024-09-05
Enable TimeProvider on New Model

FuPingFranco:
# Enable TimeProvider on New Model

## Description

Enable TimeProvider on new Wilson model.


iNinja:
This looks good so far. If you could add the [FakeTimeProvider](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.time.testing.faketimeprovider?view=net-8.0) to the tests to prevent any flakiness, it would be perfect.

FuPingFranco:
> This looks good so far. If you could add the [FakeTimeProvider](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.time.testing.faketimeprovider?view=net-8.0) to the tests to prevent any flakiness, it would be perfect.

I updated our new validation unit tests that required `DateTime.UtcNow` to use `TimeProvider` but decided to implement a `MockTimeProvider` instead as the `FakeTimeProvider` raises a new warning due to it being untested for net462.

keegan-caruso:
> > This looks good so far. If you could add the [FakeTimeProvider](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.time.testing.faketimeprovider?view=net-8.0) to the tests to prevent any flakiness, it would be perfect.
> 
> I updated our new validation unit tests that required `DateTime.UtcNow` to use `TimeProvider` but decided to implement a `MockTimeProvider` instead as the `FakeTimeProvider` raises a new warning due to it being untested for net462.

I don't understand, https://www.nuget.org/packages/Microsoft.Extensions.TimeProvider.Testing/8.9.1#dependencies-body-tab

Has a net462 target.

build/dependenciesTest.props
```diff
@@ -3,6 +3,7 @@
     <BannedApiAnalyzersVersion>3.3.4</BannedApiAnalyzersVersion>
     <DotNetCoreAppRuntimeVersion>2.1.30</DotNetCoreAppRuntimeVersion>
     <MicrosoftAzureKeyVaultCryptographyVersion>2.0.5</MicrosoftAzureKeyVaultCryptographyVersion>
+    <MicrosoftExtensionsTimeProviderTestingVersion>8.8.0</MicrosoftExtensionsTimeProviderTestingVersion>
```
keegan-caruso: Can we use latest?

keegan-caruso: Also this is unused, pending the other discussion about using the testing package.

iNinja: We should consider removing this until we need it.

iNinja (approved):
Looks good overall. Minor comment.

build/dependenciesTest.props
FuPingFranco: Removed, I will follow-up with .NET team to understand the issue with this odd warning. Thanks Ignacio & Keegan!

keegan-caruso (approved):

PR #2815, opened 2024-09-06
Regression testing: Add JWE use cases

iNinja:
# Regression testing: Add JWE use cases

- Added regression tests for encrypted tokens
- Ported context switch introduced for obtaining the EcdSa key from the main path to the new one

Note: Need to evaluate whether to keep the context switch in the new path or make it the default.

Part of #2711


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -170,18 +170,38 @@ internal Result<string> DecryptToken(
...
-                        SecurityKey kdf = ecdhKeyExchangeProvider.GenerateKdf(apu, apv);
-                        var kwp = key.CryptoProviderFactory.CreateKeyWrapProviderForUnwrap(kdf, ecdhKeyExchangeProvider.GetEncryptionAlgorithm());
-                        var unwrappedKey = kwp.UnwrapKey(Base64UrlEncoder.DecodeBytes(jwtToken.EncryptedKey));
-                        unwrappedKeys.Add(new SymmetricSecurityKey(unwrappedKey));
+                        ECDsaSecurityKey? publicKey;
+
+                        // Since developers may have already worked around this issue, implicitly taking a dependency on the
+                        // old behavior, we guard the new behavior behind an AppContext switch. The new/RFC-conforming behavior
+                        // is treated as opt-in. When the library is at the point where it is able to make breaking changes
+                        // (such as the next major version update) we should consider whether or not this app-compat switch
```
keegan-caruso: Do we want to drop this in this new path? Or would we wait for IM 9.x? I think I'm okay with dropping this at this point. Curious to hear other opinions.

iNinja: I would like to, but don't feel like I have the right context to know if this is something we can do right now. Looking for more opinions.

brentschmaltz: I think we should follow the RFC and drop this for the new code path.
We can call this out in release notes as users will need to modify code to get here.

iNinja: Removed the workaround and, with it, the need for the `ephemeralDecryptionKey` in `ValidationParameters`.

keegan-caruso (approved):

FuPingFranco (approved):

PR #2817, opened 2024-09-06
Use new Base64Url API

### LINKED ISSUE #2534: M.IM8 leverages the .NET 9 Base64Url implementation
https://github.com/dotnet/runtime/issues/1658

This type should be included in preview6 of the .NET runtime available second week of July.

Replace usage of the Identity Model base64 url implementation with the implementation from the runtime.

Evaluate deprecation of the Identity Model base64 url code.

msbw2:
Use the new `Base64Url` API.


keegan-caruso:
@msbw2 Can you post a before / after of this benchmark as well here: JsonWebTokenHandler_ValidateTokenAsyncWithTVP ?

Would be good to see the difference at that level.

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
```diff
@@ -39,6 +39,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.IdentityModel.Logging\Microsoft.IdentityModel.Logging.csproj" />
+    <PackageReference Include="Microsoft.Bcl.Memory" Version="9.0.0-preview.7.24405.7" />
```
keegan-caruso: Let's move to rc1 and move the version to a var in this file:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/build/dependencies.props

pmaytak:
> @msbw2 Can you post a before / after of this benchmark as well here: JsonWebTokenHandler_ValidateTokenAsyncWithTVP ?
> 
> Would be good to see the difference at that level.

Benchmarks are here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2828#issuecomment-2360345219

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
pmaytak, line 1: There are some API breaking changes, we can't remove these internal methods.

pmaytak (changes requested):

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
pmaytak: This threw a `MissingMethodException` for `Base64UrlEncoding.Encode(Byte[], Int32, Int32]` when I referenced this from SAL. 

We should add all the internal methods and call the new / updated implementation from them.

NuGet.Config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="dotnet9" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet9/nuget/v3/index.json" />
```
jennyf19: we should be able to use public nuget, like in id web:

[microsoft-identity-web/NuGet.Config at master · AzureAD/microsoft-identity-web (github.com)](https://github.com/AzureAD/microsoft-identity-web/blob/master/NuGet.Config#L5)

benchmark/Microsoft.IdentityModel.Benchmarks/identitymodel.benchmarks.yml
```diff
@@ -57,3 +57,8 @@ scenarios:
       variables:
         filterArg: "*ValidateTokenAsyncTests*"
 
+  Base64UrlEncoderTests:
+    application:
+      job: benchmarks
+      variables:
+        filterArg: "*Base64UrlEncoderTests*"
```
pmaytak: Is there a benefit on running these benchmarks in builds? I think these are good to run right now to verify the work. Any perf regressions to encoding/decoding should be caught by other validation benchmarks.

jennyf19: I don't think they are today, correct? If not, let's open an issue figure out prioritization.

jennyf19 (approved):
:shipit:

benchmark/Microsoft.IdentityModel.Benchmarks/identitymodel.benchmarks.yml
pmaytak: No, these are new tests.

jennyf19: Yes, but no benchmarks are run today as part of the PR builds, correct?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -567,17 +568,25 @@ internal void ReadToken(ReadOnlyMemory<char> encodedTokenMemory)
...
-            int outputSize = Base64UrlEncoding.ValidateAndGetOutputSize(strSpan, startIndex, length);
+            ReadOnlySpan<char> slice = Base64UrlEncoding.Validate(strSpan, startIndex, length);
+            int outputsize = Base64Url.GetMaxDecodedLength(length);
+
+            byte[] rented = null;
+
+            const int MaxStackallocThreshold = 256;
+            Span<byte> destination = outputsize <= MaxStackallocThreshold
+                ? stackalloc byte[outputsize]
+                : (rented = ArrayPool<byte>.Shared.Rent(outputsize));
```
pmaytak: Are there unit tests that cover both of these branches?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -177,47 +120,68 @@ public static byte[] DecodeBytes(string str)
...
+                if (rented is not null)
+                    ArrayPool<byte>.Shared.Return(rented, true);
+            }
+        }
 
-            Span<byte> output = new byte[decodedLength];
+        private static bool IsOnlyValidBase64Chars(ReadOnlySpan<char> strSpan)
+        {
+            foreach (char c in strSpan)
+                if (!char.IsDigit(c) && !char.IsLetter(c) && c != Base64Character62 && c != Base64Character63 && c != Base64PadCharacter)
```
pmaytak, line 152: nit: Would it be more performant to have a set with values? Base64IsValid ends up using SearchValues.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -177,47 +120,68 @@ public static byte[] DecodeBytes(string str)
             return Decode(str.AsSpan());
         }
```
pmaytak, line 121: Do we have tests for this `DecodeBytes(string str)`? We use it everywhere but maybe good to have explicit tests like in Base64UrlEncoderTests?

Same questions for `Decode(string arg)` below.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -3,48 +3,19 @@
...
         /// </summary>
         /// <param name="inputString">The base64url encoded string to decode.</param>
         /// <returns>The decoded bytes.</returns>
-        public static byte[] Decode(string inputString)
-        {
-            _ = inputString ?? throw LogHelper.LogArgumentNullException(nameof(inputString));
-
-            return Decode(inputString, 0, inputString.Length);
-        }
+        public static byte[] Decode(string inputString) => Decode(inputString, 0, inputString.Length);
```
pmaytak: Are there unit tests for these:
`Decode(string inputString)`
`Decode(string input, int offset, int length)`

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -78,23 +51,26 @@ public static byte[] Decode(string input, int offset, int length)
...
+        /// The result of <see cref="Validate(ReadOnlySpan{char}, int, int)"/> is passed to the action.
         /// </remarks>
         public static T Decode<T, TX>(string input, int offset, int length, TX argx, Func<byte[], int, TX, T> action)
         {
             _ = action ?? throw new ArgumentNullException(nameof(action));
 
-            ReadOnlySpan<char> inputSpan = input.AsSpan();
-            int outputsize = ValidateAndGetOutputSize(inputSpan, offset, length);
+            ReadOnlySpan<char> slice = Validate(input.AsSpan(), offset, length);
+            int outputsize = Base64Url.GetMaxDecodedLength(length);
```
pmaytak: nit: `outputSize`

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -162,106 +141,20 @@ public static T Decode<T, TX, TY, TZ>(
...
-        /// <param name="offset">The index of the character in <paramref name="input"/> to start decoding from.</param>
-        /// <param name="length">The number of characters beginning from <paramref name="offset"/> to decode.</param>
-        /// <param name="output">The byte array to place the decoded results into.</param>
-        /// <remarks>
-        /// Changes from Base64UrlEncoder implementation:
-        /// 1. Padding is optional.
-        /// 2. '+' and '-' are treated the same.
-        /// 3. '/' and '_' are treated the same.
-        /// </remarks>
-        internal static void Decode(ReadOnlySpan<char> input, int offset, int length, byte[] output)
```
pmaytak, line 192: I think this should also be added back, unless it was verified that it it isn't used in other SDKs?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -337,15 +222,12 @@ public static string Encode(byte[] input, int offset, int length)
         /// <param name="strSpan">The string represented by a span to validate.</param>
         /// <param name="offset">The index of the character in <paramref name="strSpan"/> to start the decode operation.</param>
         /// <param name="length">The number of characters in <paramref name="strSpan"/> to decode, starting from <paramref name="offset"/>.</param>
-        /// <returns>The size of the decoded bytes array.</returns>
-        internal static int ValidateAndGetOutputSize(ReadOnlySpan<char> strSpan, int offset, int length)
```
pmaytak: This is used in IM.JsonWebTokens. So needs to be reverted otherwise we'll run into the package version mismatch error. You can probably just keep this as-is and add a comment that it's obsolete and the new Validate method should be used.

test/Microsoft.IdentityModel.Tokens.Tests/Base64UrlEncodingTests.cs
```diff
@@ -309,23 +309,23 @@ public Base64UrlEncoderTheoryData(string testId) : base(testId) { }
...
+            Assert.Throws<ArgumentException>(() => Base64UrlEncoding.Validate("abc".AsSpan(), 0, -1));
+            Assert.Throws<ArgumentException>(() => Base64UrlEncoding.Validate("abc".AsSpan(), 0, 4));
+            Assert.Throws<FormatException>(() => Base64UrlEncoding.Validate("abcde".AsSpan(), 0, 5));
 
-            int actualOutputSize = Base64UrlEncoding.ValidateAndGetOutputSize("abc".AsSpan(), 0, 0);
+            int actualOutputSize = Base64UrlEncoding.Validate("abc".AsSpan(), 0, 0).Length;
             Assert.Equal(0, actualOutputSize);
 
-            actualOutputSize = Base64UrlEncoding.ValidateAndGetOutputSize("abcd".AsSpan(), 0, 4);
-            Assert.Equal(3, actualOutputSize);
```
pmaytak: Hmm. Is this behavior wrong? Why would it return 3?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
msbw2: Most likely not. This was copied from the internal implementation of the `Base64Url` API, and since this is all existing code being updated no new tests were added.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
msbw2: This is `internal` so it should be fine, right?

test/Microsoft.IdentityModel.Tokens.Tests/Base64UrlEncodingTests.cs
msbw2: The previous implementation had a different way of calculating the output size.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
msbw2: `SearchValues` isn't available where this would be needed.

build/dependencies.props
```diff
@@ -11,6 +11,7 @@
     <SystemMemoryVersion>4.5.5</SystemMemoryVersion>
     <SystemSecurityCryptographyCngVersion>4.5.0</SystemSecurityCryptographyCngVersion>
     <SystemTextJson>8.0.4</SystemTextJson>
+    <MicrosoftBclMemory>9.0.0-rc.1.24431.7</MicrosoftBclMemory>
```
keegan-caruso: We can't take an rc package in targets other than net9.

jennyf19: We'd have to conditionally add it. @msbw2 let's park this to the .NET 9 release in Nov.

keegan-caruso:
Coverage is good 

Base64UrlEncoder at 89%
Base64UrlEncoding at 82%

https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1366079&view=codecoverage-tab

benchmark/Microsoft.IdentityModel.Benchmarks/identitymodel.benchmarks.yml
pmaytak: Yes, I think so.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
pmaytak: I think this is ok. This will get tested by all other tests that create a JsonWebToken.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
pmaytak: Internals can still be used between the projects here or from our other SDKs, so we treat them as public.

test/Microsoft.IdentityModel.Tokens.Tests/Base64UrlEncodingTests.cs
pmaytak: What is the difference? Does the new implementation allocate more bytes then? Shouldn't the encoding algorithm be the same?

msbw2: The `Base64Url` API calculates the max possible size whereas the custom implementation here appears to be able to find the exact size of the decoded value.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -17,11 +18,9 @@ namespace Microsoft.IdentityModel.Tokens
     /// </summary>
     public static class Base64UrlEncoder
```
ghost, line 19:
```suggestion
    [Obsolete("This class is obsolete and will be removed in a future release. Use System.Buffers.Text.Base64Url instead")]
    public static class Base64UrlEncoder
```

Should this class (and all methods) be marked as obsolete?

msbw2: That would be nice, but unfortunately there's some quirks about this class and the various methods that don't match exactly with the `Base64Url` API, so I'm not entirely sure that's feasible. I've seen quite a few usages of this downstream that probably wouldn't be able to migrate easily.

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
```diff
@@ -39,6 +39,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.IdentityModel.Logging\Microsoft.IdentityModel.Logging.csproj" />
+    <PackageReference Include="Microsoft.Bcl.Memory" Version="$(MicrosoftBclMemory)" />
```
pmaytak: Should this be included in a conditional - if TFM != net9.0?

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -256,10 +256,6 @@ internal static class LogMessages
         public const string IDX10815 = "IDX10815: Depth of JSON: '{0}' exceeds max depth of '{1}'.";
         public const string IDX10816 = "IDX10816: Decompressing would result in a token with a size greater than allowed. Maximum size allowed: '{0}'.";
 
-        // Base64UrlEncoding
-        public const string IDX10820 = "IDX10820: Invalid character found in Base64UrlEncoding. Character: '{0}', Encoding: '{1}'.";
-        public const string IDX10821 = "IDX10821: Incorrect padding detected in Base64UrlEncoding. Encoding: '{0}'.";
```
pmaytak, line 268: Did the analyzer throw an exception that this is removing an internal API?

pmaytak:
> Coverage is good
> 
> Base64UrlEncoder at 89% Base64UrlEncoding at 82%
> 
> https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1366079&view=codecoverage-tab

Should the test coverage be 100%? Curious what's not covered.

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
keegan-caruso: yes

pmaytak (approved):

keegan-caruso (approved):

PR #2818, opened 2024-09-09
Added ArgumentNullException type with ValidationError

iNinja:
# Added ArgumentNullException type with ValidationError

- Added ISecurityTokenException interface to align system exceptions and custom exceptions under a single entry point to inject the validation error used to return the stack traces.
- Added SecurityTokenArgumentNullException implementing the interface.
- Updated usage of ArgumentNullException in code and tests to consume the new type.

Part of #2711

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenException.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenException.cs
@@ -53,9 +53,6 @@ namespace Microsoft.IdentityModel.Tokens
         public SecurityTokenException(string message, Exception innerException)
             : base(message, innerException)
         {
-            Nullable<string> something;
-            something.HasValue();
-            something.Value;
         }
 
         /// <summary>
--- a/src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenException.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenException.cs
@@ -20,10 +20,12 @@ namespace Microsoft.IdentityModel.Tokens
     /// Represents a security token exception.
     /// </summary>
     [Serializable]
-    public class SecurityTokenException : Exception
+    public class SecurityTokenException : Exception, ISecurityTokenException
     {
         [NonSerialized]
         private string _stackTrace;
+
+        private ValidationError _validationError;
 
         /// <summary>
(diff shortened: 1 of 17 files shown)


src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenException.cs
```diff
@@ -66,6 +68,15 @@ protected SecurityTokenException(SerializationInfo info, StreamingContext contex
         {
         }
 
+        /// <summary>
```
keegan-caruso, line 71: Can we fill this out?

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
```diff
@@ -7,10 +7,15 @@
...
+    internal interface ISecurityTokenException
+    {
+        void SetValidationError(ValidationError validationError);
+    }
+
     /// <summary>
     /// Contains information so that Exceptions can be logged or thrown written as required.
     /// </summary>
-    internal class ValidationError
+    public class ValidationError
```
keegan-caruso, line 18: Do we want to move this to public at this point?

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenException.cs
iNinja: I'll update it tomorrow, thanks for pointing it out.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
iNinja: We will need to make this public for the time when we make the new validation path public.
Right now, in order to implement the interface I needed to make the type public, but I kept the constructors and methods internal to prevent it from being used at all.
With the current changes the type "exists" but is not usable outside of the library.

keegan-caruso (approved):
Just need to fill out the one comment, otherwise looks good.

src/Microsoft.IdentityModel.Tokens/Exceptions/SecurityTokenException.cs
iNinja: Added.

FuPingFranco (approved):

PR #2821, opened 2024-09-10
Update to 9.0.100-rc.1

### LINKED ISSUE #2789: Update to 9.0.100-rc1
Breaking changes due to https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0057

Several places need to be fixed, especially in the tests. right now it's set to warn in the common.props
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/build/common.props#L48

and here:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/build/commonTest.props#L32

Todo:
remove the two <warn> in the props and address all the issues.
Will probably need
`#if NET9_0_OR_GREATER`
in all the places

kllysng:
Fixes #2789


build/template-Build-run-tests-sign.yml
```diff
@@ -28,7 +28,7 @@ steps:
 - task: UseDotNet@2
   displayName: 'Use .Net Core SDK 9.x'
   inputs:
-    version: 9.0.100-preview.4.24267.66
+    version: 9.0.100-preview.7.24407.12
```
jennyf19: we should move to RC1 https://dotnet.microsoft.com/en-us/download/dotnet/9.0

pmaytak:
```suggestion
    version: 9.0.0-rc.1.24431.7
```

jennyf19: would like @kellyyangsong to take this locally and make sure the tests work locally instead of taking the suggestion via the UI.

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
```diff
@@ -147,7 +147,11 @@ public ConfigurationValidationResult Validate(WsFederationConfiguration configur
 
                     try
                     {
+#if NET9_0_OR_GREATER
+                        cert = X509CertificateLoader.LoadCertificate(Convert.FromBase64String(signatureCertData.Current));
+#else
                         cert = new X509Certificate2(Convert.FromBase64String(signatureCertData.Current));
+#endif
```
pmaytak: Since these lines repeat throughout, one code duplication reduction to consider is extract this into a helper method, so that `#if#` flag is only in one place.

```csharp
cert = CertificateHelper.Load(Convert.FromBase64String(signatureCertData.Current));
```
```csharp
internal static X509Certificate2 Load(byte[] data)
{
#if NET9_0_OR_GREATER
    return X509CertificateLoader.LoadCertificate(data);
#else
    return new X509Certificate2(data);
#endif
}
```

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
```diff
@@ -71,9 +85,15 @@ static public class KeyingMaterial
         public static SigningCredentials X509SigningCreds_SelfSigned2048_SHA512 = new SigningCredentials(X509SecurityKeySelfSigned2048_SHA512, SecurityAlgorithms.RsaSha512);
 
         // all asymmetric material has private key unless public is included in variable name
-        public static SecureString CertPassword = ConvertToSecureString("abcd");
+        public static string CertPasswordString = "abcd";
+        public static SecureString CertPassword = ConvertToSecureString(CertPasswordString);
```
pmaytak: Why was this change and the ones above made?

pmaytak (commented):
Main suggestion is to refactor the certificate creation into a separate helper. This will also help, for example, if in the future we decide to use `Microsoft.Bcl.Cryptography` package, then the code change will only have to be in one place.

build/template-Build-run-tests-sign.yml
pmaytak: Whatever works best.

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
kllysng: Ah they're vestigial to before we decided to pragma warning disable since the constructor didn't have an overload. X509CertificateLoader.LoadPcks12() has similar args but with a string instead of secure string password arg. But we won't be using that - so reverting these changes. Thank you!

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationValidator.cs
kllysng: Great point, added a `X509CertificateHelper`.

build/template-Build-run-tests-sign.yml
kllysng: Updated to rc1.

kllysng: Updated to rc1.

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
kllysng: Reverted these changes.

test/Microsoft.IdentityModel.Xml.Tests/EnvelopedSignatureReaderTests.cs
```diff
@@ -213,8 +212,8 @@ public static TheoryData<EnvelopedSignatureTheoryData> ReadXmlElementsTheoryData
...
...tk2bXMYq8dO3xSRASmdVmgmoyoQGJrnh1E0SihxCFNgiHqyGQ1JAgMBAAGjggE9MIIBOTAJBgNVHRMEAjAAMBEGCWCGSAGG+EIBAQQEAwIEsDAdBgNVHQ4EFgQU0HAvIqGfYUW/akxHd9SNK2O4GwEwgbsGA1UdIwSBszCBsIAUzxJQXpMfXwleF48WV0F6n3lyhLihgYykgYkwgYYxCzAJBgNVBAYTAkNaMQ8wDQYDVQQHEwZQcmFndWUxGTAXBgNVBAoTEENaLk5JQywgei5zLnAuby4xMTAvBgNVBAMTKENaLk5JQyBTSEEyIFJvb3QgQ2VydGlmaWNhdGlvbiBBdXRob3JpdHkxGDAWBgkqhkiG9w0BCQEWCWNhQG5pYy5jeoIJAJxtRGsvNinfMBQGA1UdEgQNMAuBCWNhQG5pYy5jejAWBgNVHREEDzANgQtyb290QG5pYy5jejAOBgNVHQ8BAf8EBAMCBaAwDQYJKoZIhvcNAQELBQADggIBAIytUqXh7AU/OunkSpUTaEY5Ze2sdV76JPwYVzNn2O6hCyzGvPXARP0IchUv8Vy+eCFLctUxvCbUS6aC+ObKgGXq4MxoSV8lijMLEW9crCpFDoLLd3LQw0GjVk2mCE7XTuaT0choPYlZmjv+wF2ZKm5/B+Qjek2j7SkY1yn7hxgJdd5ljHE6wmDXLJ8gHuVBNwvc5iHDjHYh7jL5c5jCDBcr1fFCsIARU05RAkfpWurl8GKY8t6IPm7iopOLjru3Gl45ZBdVrAPMQ8Fz1M9VElUJ6ngeKkXHkSwGhCBG3X0MYsltND6mZSkJqN0nOs+cJ6HHO8IZW1f2pH3aCTUSDYWoaZbK74NC6d61sr5Rth4foLQnMzCS5RaXuANMvyZW3Ol5ScvLl/KRZM4f4CB6rmYinyHfXIoPF+uCjavyOYnW1RDBASg0Ld/WUJlWb75m5GNkRELIc4c5FU54ysMW9o5wnGpQvtXNdCBJK8tAyZO9Wf2hjZeOZgJ6r1IngfSeSFu7EOFqWnwVOF+3juwWOLCwxrKcURAEngwH01ydwU3oG/rN+7JtdS3IwfaBt9sfDiLQ60qec/6PQc643UztE6oToHLRXsidwrObwyAKLSFJoh/uxWT85JgAoekq5zBen94HfELfMEc9tex6Qlf1tLDs7OWD6Mlw6j9aAcw/4Nfh")));
```
kllysng: Had to add a `X509CertificateHelper` to this package as adding an internals visible to for `Microsoft.IdentityModel.Tokens.Xml.Tests` to the Wilson.Tokens package assembly info didn't work like it did for Wilson.Protocols.WsFederation, for example. The only thing I can think of is because the namespace of this file doesn't match the package/.csproj, which is `Microsoft.IdentityModel.Xml.Tests` and not `Microsoft.IdentityModel.Tokens.Xml.Tests`.

kllysng: image
Despite adding to Wilson.Tokens assembly info, adding Wilson.Tokens package ref to .csproj, using Wilson.Tokens at the top of this file..

src/Microsoft.IdentityModel.Tokens/X509CertificateHelper.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Helper class to load X509Certificate2 from byte array.
+    /// </summary>
+    internal class X509CertificateHelper
+    {
+        public static X509Certificate2 Load(byte[] data)
```
pmaytak: nit: What about naming class more generic like `CertificateHelper` and method `LoadX509Certificate`? That way we can add more methods here to load different cert types in the future, if needed (otherwise, we'd have to create a separate class for each cert type). And since this is an internal class that's used in multiple projects, we should make sure the API is right, since we won't be able to change it.

pmaytak: @keegan-caruso Thoughts on adding an internal class to be used in multiple projects? I suppose since we'll add the public API analyzers, this is okay. You mentioned previously an approach of create a separate duplicate helper class in each project to avoid the issues with internal friendlies.

test/Microsoft.IdentityModel.Xml.Tests/EnvelopedSignatureReaderTests.cs
jennyf19: Not sure I'm following, so the internalsVisibleTo did not work?

src/Microsoft.IdentityModel.Tokens/X509CertificateHelper.cs
kllysng: I was considering making a separate helper class per project too.. happy to do that. Also happy to name it a bit more generic + .LoadX509()

test/Microsoft.IdentityModel.Xml.Tests/EnvelopedSignatureReaderTests.cs
kllysng: Did not work, some of the tests in this project have a different namespace than the project namespace.

src/Microsoft.IdentityModel.Tokens/X509CertificateHelper.cs
brentschmaltz: I agree with @pmaytak a more general name may be an advantage.
Is there a reason why this is internal?

src/Microsoft.IdentityModel.Tokens/X509CertificateHelper.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+#if NET9_0_OR_GREATER
```
brentschmaltz: This API is available in net8 are we choosing net9 deliberately?

src/Microsoft.IdentityModel.Tokens/X509CertificateHelper.cs
kllysng: Spoke with @keegan-caruso - we'll use [compile include ](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/f0d09d408948c6920f171299bee9167d0850a51b/src/Microsoft.IdentityModel.Logging/Microsoft.IdentityModel.Logging.csproj#L20C1-L22C15) instead of internals visible to. This way we avoid the issues with internals but we keep the helper class in one spot for maintainability.

@brentschmaltz we want to keep this internal. This is a helper class that we don't want to ship.

src/Microsoft.IdentityModel.Tokens/X509CertificateHelper.cs
kllysng: The breaking change was introduced in net9

test/Microsoft.IdentityModel.Xml.Tests/EnvelopedSignatureReaderTests.cs
kllysng: I'll be adding a public helper to the tests class to avoid sharing internals and this issue will be fixed

src/Microsoft.IdentityModel.Tokens/X509CertificateHelper.cs
kllysng: Done

test/Microsoft.IdentityModel.Protocols.WsFederation.Tests/WsFederationMessageTests.cs
```diff
@@ -139,11 +139,12 @@ public static TheoryData<WsFederationSigninMessageTheoryData> WaSignInTheoryData
...
+                X509Certificate2 cert = TestUtils.CertificateHelper.LoadX509Certificate(Convert.FromBase64String(_x509DataADFS));
```
msbw2: NIT: Why was this moved up here?

test/Microsoft.IdentityModel.Protocols.WsFederation.Tests/WsFederationMessageTests.cs
```diff
@@ -139,11 +139,12 @@ public static TheoryData<WsFederationSigninMessageTheoryData> WaSignInTheoryData
 
                 // customer data
                 string _x509DataADFS = "MIIDCDCCAfCgAwIBAgIQNz4YVbYAIJVFCc47HFD3RzANBgkqhkiG9w0BAQsFADBAMT4wPAYDVQQDEzVBREZTIFNpZ25pbmcgLSBzdHMuc3ViMi5mcmFjYXMzNjUubXNmdG9ubGluZXJlcHJvLmNvbTAeFw0xNTAzMzExMDQyMTNaFw0zNDA1MzAxMDQyMTNaMEAxPjA8BgNVBAMTNUFERlMgU2lnbmluZyAtIHN0cy5zdWIyLmZyYWNhczM2NS5tc2Z0b25saW5lcmVwcm8uY29tMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAxdjzB+wGV6hYekOvWwKoL/DFNBiLQsLx6w02FzcFnpGwR38gVTn/glg9CNSsOT0riRM3/MwU8o2fwseQyVtv9Kee/yvia8cB6GD0CARlYizb6GkJJzMvWkPSas1zpn10Bs3SBBgn0pvAKZCWWir5WJ7DRY32X2yo2do8mQftsoLGYsEU8+jj9AMYQWaR3A86AEWjXoQY3AodfMMzdVFX+O/kjsvKcBfPqGRT6jUSGBOOaqzMOJBT39SueD8zePDW7SejBl7fRi4TLx5H6xuMldOAAH6oD70yIrobqosGG9X/LdijHajMSoaYzZIlG7fl4PCVvAjh1Dytw/y8K70flQIDAQABMA0GCSqGSIb3DQEBCwUAA4IBAQBy08dAsSTKd2YlNE8bM4O5C2bFKR1YakR8L/zLEy8g+RNsKN5V/cIll0b/tf9iQ5464nc+nM///U+UVxqT8ipeR7ThIPwuWX98cFZFQNNGkha4PaYap/osquEpRAJOcTqZf2K95ipeQ+5Hhw00mK0hcV1QT/7maTUqCHDfBCaD+uYAFvaNBXOYpdoIGM9cMk7Qjc/yowLDm+DpmJek54MWmN+iZ0YtDEhMSh//QPFMLPT5Ucat+qRTen1HZNGdxfZ7NIIDL3dNKVDN+vDUbW7rjvPyxA8Rtj4JplI9ENqpzRq4m1sDWUTk2hJYw9Ec1kGo7AFKRmOS6DRbwUn5Ptdc";
```
msbw2: Does this cert expire?

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
```diff
@@ -30,6 +33,7 @@ static public class KeyingMaterial
...
         public static string SelfSigned1024_SHA256_Public = "MIICWTCCAcKgAwIBAgIQPq8RWqoOL51G/qUxq7tBizANBgkqhkiG9w0BAQsFADA1MTMwMQYDVQQDHioAUwBlAGwAZgBTAGkAZwBuAGUAZAAxADAAMgA0AF8AUwBIAEEAMgA1ADYwHhcNMTQxMjI2MTUyNzAwWhcNMzkxMjMxMjM1OTU5WjA1MTMwMQYDVQQDHioAUwBlAGwAZgBTAGkAZwBuAGUAZAAxADAAMgA0AF8AUwBIAEEAMgA1ADYwgZ8wDQYJKoZIhvcNAQEBBQADgY0AMIGJAoGBAMl0ubbO3e3FpeMN2/gjz0YElurF0UCT25777CmNfzpvxtuvziUxa/ALMxmz+sVLzKb6kMuLNagU12cYZS3ksgntqxz8R/VyZO33hysNAMeMcdeOMhMm6ubTXEqqdHAb4TUPhme6o7Bvpcx3yPlxECoy6XE5A0VlYBgWzqMJN6F5AgMBAAGjajBoMGYGA1UdAQRfMF2AEKiIeaB2ShCC/Mrhyxg2Lf+hNzA1MTMwMQYDVQQDHioAUwBlAGwAZgBTAGkAZwBuAGUAZAAxADAAMgA0AF8AUwBIAEEAMgA1ADaCED6vEVqqDi+dRv6lMau7QYswDQYJKoZIhvcNAQELBQADgYEAaqChtfN/l6xTcMItwFG9jhDPuWeLDXAplM0vSwbia1fIaAXdcFRSaH+5QwqoQSDROcfiWRbPNWhFXfzOj7FEBmtbGifiqDvHislRHYrqnz9FRKiay0KYn0tJ2RUsTlKxZNz0WVu9M05wJjYH4TB04ad5FhgxJZ2h/y1X+An4a/o=";
+#pragma warning disable SYSLIB0057 // X509CertificateLoader does not have the correct overloads for this constructor
```
msbw2: Is there a separate issue to track this instance?

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
```diff
@@ -97,11 +100,11 @@ static public class KeyingMaterial
         public static X509SecurityKey DefaultX509Key_2048_With_KeyId = new X509SecurityKey(DefaultCert_2048) { KeyId = DefaultX509Key_2048_KeyId };
         public static SigningCredentials DefaultX509SigningCreds_2048_RsaSha2_Sha2 = new SigningCredentials(DefaultX509Key_2048, SecurityAlgorithms.RsaSha256Signature, SecurityAlgorithms.Sha256Digest);
         public static X509Certificate2 DefaultAsymmetricCert_2048 = new X509Certificate2(Convert.FromBase64String(DefaultX509Data_2048), CertPassword, X509KeyStorageFlags.MachineKeySet);
-
+#pragma warning restore SYSLIB0057
```
msbw2: This section with `SYSLIB0057` seems to cover too much. Can you move this to only encompass the necessary parts and add more pragmas as needed with some tracking items for them if none already exist?

test/Microsoft.IdentityModel.TestUtils/CertificateHelper.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT License.
+
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.IdentityModel.TestUtils
+{
+    /// <summary>
+    /// Helper class to load X509Certificate2 from byte array.
+    /// </summary>
+    public class CertificateHelper
```
pmaytak, line 12: I think it's okay to share internals with our test libraries, since tests reference the projects as source, so this class may not be needed.

pmaytak (approved):

test/Microsoft.IdentityModel.Protocols.WsFederation.Tests/WsFederationMessageTests.cs
kllysng: I don't know but looking at the blame, these tests have been here for years. I am guessing no.

test/Microsoft.IdentityModel.Protocols.WsFederation.Tests/WsFederationMessageTests.cs
kllysng: There were some intermediate changes with huge diffs, I must have missed this. Will fix, thank you!

test/Microsoft.IdentityModel.TestUtils/CertificateHelper.cs
kllysng: I'm going to keep the test public helper class as there were issues with sharing internals with one of the test projects.

src/Microsoft.IdentityModel.Tokens/X509CertificateHelper.cs
kllysng: Talked offline and looks like we prefer to use it in net8 - updating.

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
kllysng: Do you mean an issue opened on dotnet or an internal one for us to monitor whether they've added the correct overloads?

src/Microsoft.IdentityModel.Tokens/X509CertificateHelper.cs
kllysng: Update: image

kllysng: @pmaytak found that: 

For .NET 9, in the Definition, the assembly is SystemSecurity.Cryptography.dll.
[X509CertificateLoader.LoadCertificate Method (System.Security.Cryptography.X509Certificates) | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificateloader.loadcertificate?view=net-9.0#definition)
 
For .NET 8, it's the backwards compat package: Microsoft.Bcl.Cryptography.dll
[X509CertificateLoader.LoadCertificate Method (System.Security.Cryptography.X509Certificates) | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificateloader.loadcertificate?view=net-8.0#definition)

I would rather not introduce a new dependency to also use this api in .net 8. Thoughts?

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
kllysng: Introducing a helper with the warning disable encompassing only the X509Certificate2 constructor

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
kllysng: I opened an [issue ](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2833)and linked it.

src/Microsoft.IdentityModel.Tokens/X509CertificateHelper.cs
kllysng: Discussed offline and we don't want to add another dependency to use this in .net 8.

src/Microsoft.IdentityModel.Tokens/CertificateHelper.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Helper class to load X509Certificate2 from byte array.
+    /// </summary>
+    internal class CertificateHelper
+    {
+        public static X509Certificate2 LoadX509Certificate(byte[] data)
```
brentschmaltz, line 13: If you wanted to, you could add another overload that takes a string in base64format, then convert in the method.

Reducing code for the caller.

brentschmaltz (approved):
Small suggestion for an additional API (optional)

src/Microsoft.IdentityModel.Tokens/CertificateHelper.cs
kllysng: I promise I will do this in a follow up PR, I don't want to reset any approvals at this stage 🙏 🙇

msbw2 (approved):

src/Microsoft.IdentityModel.Tokens/CertificateHelper.cs
kllysng: As promised: #2834

PR #2826, opened 2024-09-16
Uses spans for issuer comparison and fixes prior index out of range error

### LINKED ISSUE #2595: [Bug] Reduce allocations in Aad validation by not using string.Replace when it isn't necessary

sruke:
Why?
Improves performance during issuer validation by replacing string comparison with span comparison.

What?
The performance improvements were originally introduced in [this PR](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/dc15328457df098e157363713fc5d32724458c72) that had to be reverted due an index out of range error. This PR reintroduces the changes and addresses the error.

### DIFF
--- a/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
+++ b/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
@@ -382,18 +382,38 @@ namespace Microsoft.IdentityModel.Validators
             }
         }
 
-        private static bool IsValidIssuer(string validIssuerTemplate, string tenantId, string actualIssuer)
+        internal static bool IsValidIssuer(string issuerTemplate, string tenantId, string tokenIssuer)
         {
-            if (string.IsNullOrEmpty(validIssuerTemplate))
+            if (string.IsNullOrEmpty(issuerTemplate) || string.IsNullOrEmpty(tenantId) || string.IsNullOrEmpty(tokenIssuer))
                 return false;
 
-            if (validIssuerTemplate.Contains(TenantIdTemplate))
+            ReadOnlySpan<char> issuerTemplateSpan = issuerTemplate.AsSpan();
+            ReadOnlySpan<char> tokenIssuerSpan = tokenIssuer.AsSpan();
+            int templateTenantIdPosition = issuerTemplate.IndexOf(TenantIdTemplate, StringComparison.Ordinal);
+
+            // If the template contains the tenantIdTemplate, ensure the actual issuer matches the template with the tenantId replaced.
+            if (templateTenantIdPosition >= 0 && tokenIssuer.Length > templateTenantIdPosition)
             {
-                return validIssuerTemplate.Replace(TenantIdTemplate, tenantId) == actualIssuer;
+                // Ensure the prefix of the issuer template matches the token issuer's prefix
+                if (!issuerTemplateSpan.Slice(0, templateTenantIdPosition).SequenceEqual(tokenIssuerSpan.Slice(0, templateTenantIdPosition)))
+                    return false;
+
+                // Ensure tokenIssuer is atleast as long as issuerTemplate with tenantIdTemplate replaced
+                if (tokenIssuer.Length <= templateTenantIdPosition + tenantId.Length)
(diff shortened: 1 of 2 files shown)


test/Microsoft.IdentityModel.Validators.Tests/AadIssuerValidatorTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            var validationResult = AadIssuerValidator.IsValidIssuer(
+                theoryData.TemplatedIssuer,
+                theoryData.TenantIdClaim,
+                theoryData.TokenIssuer);
+
+            // Assert
+            Assert.Equal(theoryData.ExpectedResult, validationResult);
+        }
+
+        public static TheoryData<AadIssuerValidatorTheoryData> AadIssuerValidationTestCases()
```
keegan-caruso, line 24: Overall looks good!

We should add a test case for no template, comparing issuer template directly against token issuer

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -382,18 +382,38 @@ private ConfigurationManager<OpenIdConnectConfiguration> CreateConfigManager(
...
+            // If the template contains the tenantIdTemplate, ensure the actual issuer matches the template with the tenantId replaced.
+            if (templateTenantIdPosition >= 0 && tokenIssuer.Length > templateTenantIdPosition)
             {
-                return validIssuerTemplate.Replace(TenantIdTemplate, tenantId) == actualIssuer;
+                // Ensure the prefix of the issuer template matches the token issuer's prefix
+                if (!issuerTemplateSpan.Slice(0, templateTenantIdPosition).SequenceEqual(tokenIssuerSpan.Slice(0, templateTenantIdPosition)))
+                    return false;
+
+                // Ensure tokenIssuer is atleast as long as issuerTemplate with tenantIdTemplate replaced
+                if (tokenIssuer.Length <= templateTenantIdPosition + tenantId.Length)
```
SaurabhMSFT, line 402: Is there a test case for this `if` block?

sruke: `Malformed_V2_TokenIssuer_Failure` is the test for the if block where the tokenIssuer is malformed and its length is less than the issuerTemplate with tenantId replaced.

test/Microsoft.IdentityModel.Validators.Tests/AadIssuerValidatorTests.cs
sruke: Good point! Added a success and a failure test where the issuer template contains a GUID instead of a tenant id template.

keegan-caruso (approved):
LGTM!

kllysng (approved):

PR #2838, opened 2024-09-23
Regression tests: Audience

iNinja:
# Regression tests: Audience
- Extracted regression tests related to audience to a separate file.
- Added extra scenarios for audience related validations.
- Fixed audience validation unit tests and renamed variables to improve clarity.

Note: Upcoming PRs will continue to extract the different aspects of regression tests into separate files.

Part of #2711

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
@@ -10,7 +10,7 @@ namespace Microsoft.IdentityModel.Tokens
 {
     internal class AudienceValidationError : ValidationError
     {
-        private string? _invalidAudience;
+        private IList<string>? _invalidAudiences;
 
         public AudienceValidationError(
             MessageDetail messageDetail,
@@ -19,13 +19,13 @@ namespace Microsoft.IdentityModel.Tokens
             IList<string>? invalidAudiences)
             : base(messageDetail, ValidationFailureType.AudienceValidationFailed, exceptionType, stackFrame)
         {
-            _invalidAudience = Utility.SerializeAsSingleCommaDelimitedString(invalidAudiences);
+            _invalidAudiences = invalidAudiences;
         }
 
         internal override void AddAdditionalInformation(ISecurityTokenException exception)
         {
             if (exception is SecurityTokenInvalidAudienceException invalidAudienceException)
-                invalidAudienceException.InvalidAudience = _invalidAudience;
+                invalidAudienceException.InvalidAudience = Utility.SerializeAsSingleCommaDelimitedString(_invalidAudiences);
         }
     }
 }
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
@@ -1,0 +1,32 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+#nullable enable
+namespace Microsoft.IdentityModel.Tokens
+{
+    internal class AudienceValidationError : ValidationError
+    {
+        private string? _invalidAudience;
+
+        public AudienceValidationError(
+            MessageDetail messageDetail,
+            Type exceptionType,
+            StackFrame stackFrame,
+            IList<string>? invalidAudiences)
+            : base(messageDetail, ValidationFailureType.AudienceValidationFailed, exceptionType, stackFrame)
+        {
+            _invalidAudience = Utility.SerializeAsSingleCommaDelimitedString(invalidAudiences);
+        }
+
+        internal override void AddAdditionalInformation(ISecurityTokenException exception)
+        {
+            if (exception is SecurityTokenInvalidAudienceException invalidAudienceException)
+                invalidAudienceException.InvalidAudience = _invalidAudience;
+        }
+    }
+}
+#nullable restore
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Validators.SkipValidations.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Validators.SkipValidations.cs
@@ -1,0 +1,98 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+#nullable enable
+namespace Microsoft.IdentityModel.Tokens
+{
+    public static partial class Validators
+    {
+        internal static AlgorithmValidatorDelegate SkipAlgorithmValidation = delegate (
+            string algorithm,
+            SecurityKey securityKey,
+            SecurityToken securityToken,
+            ValidationParameters
+            validationParameters,
+            CallContext callContext)
+        {
+            return algorithm;
+        };
+
+        internal static AudienceValidatorDelegate SkipAudienceValidation = delegate (
+            IList<string> audiences,
+            SecurityToken? securityToken,
+            ValidationParameters validationParameters,
+            CallContext callContext)
+        {
+            return "skipped"; // The audience that was validated.
+        };
+
+        internal static IssuerValidationDelegateAsync SkipIssuerValidation = delegate (
+            string issuer,
+            SecurityToken securityToken,
+            ValidationParameters validationParameters,
+            CallContext callContext,
+            CancellationToken cancellationToken)
+        {
+            return Task.FromResult(new ValidationResult<ValidatedIssuer>(
+                new ValidatedIssuer(issuer, IssuerValidationSource.NotValidated)));
+        };
+
+        internal static IssuerSigningKeyValidatorDelegate SkipIssuerSigningKeyValidation = delegate (
+            SecurityKey signingKey,
+            SecurityToken securityToken,
+            ValidationParameters validationParameters,
+            BaseConfiguration? configuration,
+            CallContext? callContext)
+        {
+            return new ValidatedSigningKeyLifetime(
+                null, // ValidFrom
+                null, // ValidTo
+                null);// ValidationTime
+        };
+
+        internal static LifetimeValidatorDelegate SkipLifetimeValidation = delegate (
+            DateTime? notBefore,
+            DateTime? expires,
+            SecurityToken? securityToken,
+            ValidationParameters validationParameters,
+            CallContext callContext)
+        {
+            return new ValidatedLifetime(notBefore, expires);
+        };
+
+        internal static SignatureValidatorDelegate SkipSignatureValidation = delegate (
+            SecurityToken securityToken,
+            ValidationParameters validationParameters,
+            BaseConfiguration? configuration,
+            CallContext? callContext)
+        {
+            // This key is not used during the validation process. It is only used to satisfy the delegate signature.
+            // Follow up PR will change this to remove the SecurityKey return value.
+            return new(result: new JsonWebKey());
+        };
+
+        internal static TokenReplayValidatorDelegate SkipTokenReplayValidation = delegate (
+            DateTime? expirationTime,
+            string securityToken,
+            ValidationParameters validationParameters,
+            CallContext callContext)
+        {
+            return expirationTime;
+        };
+
+        internal static TypeValidatorDelegate SkipTypeValidation = delegate (
+            string? type,
+            SecurityToken? securityToken,
+            ValidationParameters validationParameters,
+            CallContext callContext)
+        {
+            return new ValidatedTokenType("skipped", 0);
+        };
+    }
+}
+#nullable restore
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Validators.SkipValidations.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Validators.SkipValidations.cs
@@ -11,7 +11,7 @@ namespace Microsoft.IdentityModel.Tokens
 {
     public static partial class Validators
     {
-        internal static AlgorithmValidatorDelegate SkipAlgorithmValidation = delegate (
+        internal static AlgorithmValidationDelegate SkipAlgorithmValidation = delegate (
             string algorithm,
             SecurityKey securityKey,
             SecurityToken securityToken,
@@ -22,7 +22,7 @@ namespace Microsoft.IdentityModel.Tokens
             return algorithm;
         };
 
-        internal static AudienceValidatorDelegate SkipAudienceValidation = delegate (
+        internal static AudienceValidationDelegate SkipAudienceValidation = delegate (
             IList<string> audiences,
             SecurityToken? securityToken,
             ValidationParameters validationParameters,
@@ -42,7 +42,7 @@ namespace Microsoft.IdentityModel.Tokens
                 new ValidatedIssuer(issuer, IssuerValidationSource.NotValidated)));
         };
 
-        internal static IssuerSigningKeyValidatorDelegate SkipIssuerSigningKeyValidation = delegate (
+        internal static IssuerSigningKeyValidationDelegate SkipIssuerSigningKeyValidation = delegate (
             SecurityKey signingKey,
             SecurityToken securityToken,
             ValidationParameters validationParameters,
@@ -55,7 +55,7 @@ namespace Microsoft.IdentityModel.Tokens
                 null);// ValidationTime
         };
 
-        internal static LifetimeValidatorDelegate SkipLifetimeValidation = delegate (
+        internal static LifetimeValidationDelegate SkipLifetimeValidation = delegate (
             DateTime? notBefore,
             DateTime? expires,
             SecurityToken? securityToken,
@@ -65,7 +65,7 @@ namespace Microsoft.IdentityModel.Tokens
             return new ValidatedLifetime(notBefore, expires);
         };
 
-        internal static SignatureValidatorDelegate SkipSignatureValidation = delegate (
+        internal static SignatureValidationDelegate SkipSignatureValidation = delegate (
             SecurityToken securityToken,
             ValidationParameters validationParameters,
             BaseConfiguration? configuration,
@@ -76,7 +76,7 @@ namespace Microsoft.IdentityModel.Tokens
             return new(result: new JsonWebKey());
         };
 
-        internal static TokenReplayValidatorDelegate SkipTokenReplayValidation = delegate (
+        internal static TokenReplayValidationDelegate SkipTokenReplayValidation = delegate (
             DateTime? expirationTime,
             string securityToken,
             ValidationParameters validationParameters,
@@ -85,7 +85,7 @@ namespace Microsoft.IdentityModel.Tokens
             return expirationTime;
         };
 
-        internal static TypeValidatorDelegate SkipTypeValidation = delegate (
+        internal static TokenTypeValidationDelegate SkipTokenTypeValidation = delegate (
             string? type,
             SecurityToken? securityToken,
             ValidationParameters validationParameters,
--- a/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
+++ b/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
@@ -16,7 +16,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
         [Theory, MemberData(nameof(ValidateTokenAsync_AudienceTestCases))]
         public async Task ValidateTokenAsync_Audience(ValidateTokenAsyncAudienceTheoryData theoryData)
         {
-            var context = TestUtilities.WriteHeader($"{this}.ValidateTokenAsync", theoryData);
+            var context = TestUtilities.WriteHeader($"{this}.ValidateTokenAsync_Audience", theoryData);
 
             string jwtString = CreateToken(theoryData.Audience);
 
--- a/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
+++ b/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
@@ -12,7 +12,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
 {
     public partial class JsonWebTokenHandlerValidateTokenAsyncTests
     {
-        [Theory, MemberData(nameof(ValidateTokenAsync_AudienceTestCases))]
+        [Theory, MemberData(nameof(ValidateTokenAsync_AudienceTestCases), DisableDiscoveryEnumeration = true)]
         public async Task ValidateTokenAsync_Audience(ValidateTokenAsyncAudienceTheoryData theoryData)
         {
             var context = TestUtilities.WriteHeader($"{this}.ValidateTokenAsync_Audience", theoryData);
--- a/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
+++ b/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.IdentityModel.TestUtils;
 using Microsoft.IdentityModel.Tokens;
-using TokenValidators = Microsoft.IdentityModel.Tokens.Validators;
 using Xunit;
 
 namespace Microsoft.IdentityModel.JsonWebTokens.Tests
@@ -138,11 +137,11 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
                     validationParameters.IgnoreTrailingSlashWhenValidatingAudience = ignoreTrailingSlashWhenValidatingAudience;
 
                     // Skip all validations except audience
-                    validationParameters.AlgorithmValidator = TokenValidators.SkipAlgorithmValidation;
-                    validationParameters.IssuerValidatorAsync = TokenValidators.SkipIssuerValidation;
-                    validationParameters.IssuerSigningKeyValidator = TokenValidators.SkipIssuerSigningKeyValidation;
-                    validationParameters.LifetimeValidator = TokenValidators.SkipLifetimeValidation;
-                    validationParameters.SignatureValidator = TokenValidators.SkipSignatureValidation;
+                    validationParameters.AlgorithmValidator = SkipValidationDelegates.SkipAlgorithmValidation;
+                    validationParameters.IssuerValidatorAsync = SkipValidationDelegates.SkipIssuerValidation;
+                    validationParameters.IssuerSigningKeyValidator = SkipValidationDelegates.SkipIssuerSigningKeyValidation;
+                    validationParameters.LifetimeValidator = SkipValidationDelegates.SkipLifetimeValidation;
+                    validationParameters.SignatureValidator = SkipValidationDelegates.SkipSignatureValidation;
 
                     return validationParameters;
                 }
--- a/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
+++ b/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.IdentityModel.TestUtils;
 using Microsoft.IdentityModel.Tokens;
+using TokenValidators = Microsoft.IdentityModel.Tokens.Validators;
 using Xunit;
 
 namespace Microsoft.IdentityModel.JsonWebTokens.Tests
@@ -115,16 +116,16 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
                     List<string>? audiences,
                     bool ignoreTrailingSlashWhenValidatingAudience = false) =>
 
+                    // Only validate the audience.
                     new TokenValidationParameters
                     {
                         ValidateAudience = true,
-                        ValidateIssuer = true,
-                        ValidateLifetime = true,
-                        ValidateTokenReplay = true,
-                        ValidateIssuerSigningKey = true,
-                        IssuerSigningKey = Default.AsymmetricSigningKey,
+                        ValidateIssuer = false,
+                        ValidateLifetime = false,
+                        ValidateTokenReplay = false,
+                        ValidateIssuerSigningKey = false,
+                        RequireSignedTokens = false,
                         ValidAudiences = audiences,
-                        ValidIssuer = Default.Issuer,
                         IgnoreTrailingSlashWhenValidatingAudience = ignoreTrailingSlashWhenValidatingAudience,
                     };
 
@@ -133,10 +134,15 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
                     bool ignoreTrailingSlashWhenValidatingAudience = false)
                 {
                     ValidationParameters validationParameters = new ValidationParameters();
-                    validationParameters.ValidIssuers.Add(Default.Issuer);
                     audiences?.ForEach(audience => validationParameters.ValidAudiences.Add(audience));
-                    validationParameters.IssuerSigningKeys.Add(Default.AsymmetricSigningKey);
                     validationParameters.IgnoreTrailingSlashWhenValidatingAudience = ignoreTrailingSlashWhenValidatingAudience;
+
+                    // Skip all validations except audience
+                    validationParameters.AlgorithmValidator = TokenValidators.SkipAlgorithmValidation;
+                    validationParameters.IssuerValidatorAsync = TokenValidators.SkipIssuerValidation;
+                    validationParameters.IssuerSigningKeyValidator = TokenValidators.SkipIssuerSigningKeyValidation;
+                    validationParameters.LifetimeValidator = TokenValidators.SkipLifetimeValidation;
+                    validationParameters.SignatureValidator = TokenValidators.SkipSignatureValidation;
 
                     return validationParameters;
                 }
@@ -157,9 +163,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
             SecurityTokenDescriptor securityTokenDescriptor = new SecurityTokenDescriptor
             {
                 Subject = Default.ClaimsIdentity,
-                SigningCredentials = Default.AsymmetricSigningCredentials,
                 Audience = audience,
-                Issuer = Default.Issuer,
             };
 
             return jsonWebTokenHandler.CreateToken(securityTokenDescriptor);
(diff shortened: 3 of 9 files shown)


test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+                Audience = theoryData.Audience,
+                Issuer = Default.Issuer,
+            };
+
+            string jwtString = jsonWebTokenHandler.CreateToken(securityTokenDescriptor);
+
+            TokenValidationResult tokenValidationParametersResult =
+                    await jsonWebTokenHandler.ValidateTokenAsync(jwtString, theoryData.TokenValidationParameters);
+            ValidationResult<ValidatedToken> validationParametersResult =
+                await jsonWebTokenHandler.ValidateTokenAsync(
```
jennyf19: can we name the first as legacy to provide more clarity?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+                    },
+                    new ValidateTokenAsyncAudienceTheoryData("Invalid_ValidAudiencesIsNull")
+                    {
+                        TokenValidationParameters = CreateTokenValidationParameters(null),
+                        ValidationParameters = CreateValidationParameters(null),
+                        Audience = string.Empty,
+                        ExpectedIsValid = false,
+                        // TVP path has a special case when ValidAudience is null or empty and ValidAudiences is null.
+                        ExpectedException = ExpectedException.SecurityTokenInvalidAudienceException("IDX10208:"),
+                        // VP path has a default empty List for ValidAudiences, so it will always return IDX10206 if no audiences are provided.
```
jennyf19, line 109:
```suggestion
                        // TVP path has a default empty List for ValidAudiences, so it will always return IDX10206 if no audiences are provided.
```

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.TestUtils;
+using Microsoft.IdentityModel.Tokens;
+using Xunit;
+
+namespace Microsoft.IdentityModel.JsonWebTokens.Tests
+{
+    public partial class JsonWebTokenHandlerValidateTokenAsyncTests
```
jennyf19, line 13: It's not clear in the VS test runner which test this is for, not sure how this will look w/the one for lifetime
image

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+
+            if (validationParametersResult.IsSuccess != theoryData.ExpectedIsValid)
+                context.AddDiff($"validationParametersResult.IsSuccess != theoryData.ExpectedIsValid");
+
+            if (theoryData.ExpectedIsValid &&
+                tokenValidationParametersResult.IsValid &&
+                validationParametersResult.IsSuccess)
+            {
+                IdentityComparer.AreEqual(
+                    tokenValidationParametersResult.ClaimsIdentity,
```
jennyf19: same comment as other PR

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+                        ValidationParameters = CreateValidationParameters([Default.Audience]),
+                        Audience = "InvalidAudience",
+                        ExpectedIsValid = false,
+                        ExpectedException = ExpectedException.SecurityTokenInvalidAudienceException("IDX10214:"),
+                        // ValidateTokenAsync with ValidationParameters returns a different error message to account for the
+                        // removal of the ValidAudience property from the ValidationParameters class.
+                        ExpectedExceptionValidationParameters = ExpectedException.SecurityTokenInvalidAudienceException("IDX10215:"),
+                    },
+                    new ValidateTokenAsyncAudienceTheoryData("Valid_AudienceWithinValidAudiences")
+                    {
```
jennyf19, line 53: there is a lot of duplicate code, which makes it harder to review and understand, imho. Not sure of a better solution though.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+                    new ValidateTokenAsyncAudienceTheoryData("Invalid_ValidAudiencesIsNull")
+                    {
+                        TokenValidationParameters = CreateTokenValidationParameters(null),
+                        ValidationParameters = CreateValidationParameters(null),
+                        Audience = string.Empty,
+                        ExpectedIsValid = false,
+                        // TVP path has a special case when ValidAudience is null or empty and ValidAudiences is null.
+                        ExpectedException = ExpectedException.SecurityTokenInvalidAudienceException("IDX10208:"),
+                        // VP path has a default empty List for ValidAudiences, so it will always return IDX10206 if no audiences are provided.
+                        ExpectedExceptionValidationParameters = ExpectedException.SecurityTokenInvalidAudienceException("IDX10206:"),
```
jennyf19, line 110: are we okay with the changes in behavior?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+                        ValidationParameters = CreateValidationParameters([Default.Audience], true),
+                    },
+                    new ValidateTokenAsyncAudienceTheoryData("Invalid_AudienceWithSlash_IgnoreTrailingSlashFalse")
+                    {
+                        // Audience has a trailing slash and IgnoreTrailingSlashWhenValidatingAudience is false.
+                        Audience = Default.Audience + "/",
+                        TokenValidationParameters = CreateTokenValidationParameters([Default.Audience], false),
+                        ValidationParameters = CreateValidationParameters([Default.Audience], false),
+                        ExpectedIsValid = false,
+                        ExpectedException = ExpectedException.SecurityTokenInvalidAudienceException("IDX10214:"),
```
jennyf19, line 72: are we okay with the change in behavior?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+using Xunit;
+
+namespace Microsoft.IdentityModel.JsonWebTokens.Tests
+{
+    public partial class JsonWebTokenHandlerValidateTokenAsyncTests
+    {
+        [Theory, MemberData(nameof(ValidateTokenAsync_AudienceTestCases))]
+        public async Task ValidateTokenAsync_Audience(ValidateTokenAsyncAudienceTheoryData theoryData)
+        {
+            var context = TestUtilities.WriteHeader($"{this}.ValidateTokenAsync", theoryData);
```
FuPingFranco: Might want to be more specific on this test header name similar to the lifetime regression test consider `ValidateTokenAsync_Audience`

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
FuPingFranco: +1

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+                Subject = Default.ClaimsIdentity,
+                SigningCredentials = Default.AsymmetricSigningCredentials,
+                Audience = theoryData.Audience,
+                Issuer = Default.Issuer,
+            };
+
+            string jwtString = jsonWebTokenHandler.CreateToken(securityTokenDescriptor);
+
+            TokenValidationResult tokenValidationParametersResult =
+                    await jsonWebTokenHandler.ValidateTokenAsync(jwtString, theoryData.TokenValidationParameters);
```
FuPingFranco: nit: fix indentation

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+                        ExpectedException = ExpectedException.SecurityTokenInvalidAudienceException("IDX10206:"),
+                    },
+                    new ValidateTokenAsyncAudienceTheoryData("Invalid_ValidAudiencesIsNull")
+                    {
+                        TokenValidationParameters = CreateTokenValidationParameters(null),
+                        ValidationParameters = CreateValidationParameters(null),
+                        Audience = string.Empty,
+                        ExpectedIsValid = false,
+                        // TVP path has a special case when ValidAudience is null or empty and ValidAudiences is null.
+                        ExpectedException = ExpectedException.SecurityTokenInvalidAudienceException("IDX10208:"),
```
FuPingFranco, line 108: nit: can we name this var `TVPExpectedException` and the one below `VPExpectedException` since you provided a comment already.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: TVP, TokenValidationParameters, is the "legacy" path. 
VP, or ValidationParameters, is the new one. 

They are the two paths that we are comparing and in this scenario there is a difference in behaviour due to the fact that TVP had separate `ValidAudience` and `ValidAudiences`.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: You are looking at the previous regression tests, which I am decoupling into separate files.
The current one shows as:

Screenshot 2024-09-24 at 12 29 28

As part of the next PRs, I am removing `JsonWebTokenHandlerValidationParametersTests`, and using partial classes to keep all the different aspects inside the same class in the text explorer.
We are looking at having `ValidateTokenAsync_Audience`, `ValidateTokenAsync_Lifetime`, and so on, all under `JsonWebTokenHandlerValidateTokenAsyncTests`.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Originally we were using a single test class and have all the theory data alongside, but it was requested by Brent that we decouple the different aspects to reduce the size of the files and make it easier to follow what aspects of the validation are being tested.

In order to separate this, we have to duplicate some of the test scenario code. At the end of the day, the tests are doing the same steps:
- Create the token based on some configuration
- Validate the created token using the legacy validation path
- Validate the created token using the new path
- Compare the result from one against the other
- Compare the claims and claims identify from both so that they are identical.
- Ensure the expected validity and exception match

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: This was defined as a group that all List fields in `ValidationParameters` would default to an empty list if not provided.

I am doing my best in the tests to highlight the defined behaviour, especially where there are any differences to the "legacy" one.
From previous conversations, I was told that customers should not rely on the contents of the error messages, and the type of the exceptions match. Based on that, it looks okay to me.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: `ExpectedException` is part of `TheoryDataBase`. I added the `ExpectedExceptionValidationParameters` to differentiate the few cases where there is a difference.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: This is the difference between IDX10214 and IDX10215:

```
public const string IDX10214 = "IDX10214: Audience validation failed. Audiences: '{0}'. Did not match: validationParameters.ValidAudience: '{1}' or validationParameters.ValidAudiences: '{2}'.";
        
public const string IDX10215 = "IDX10215: Audience validation failed. Audiences: '{0}'. Did not match: validationParameters.ValidAudiences: '{1}'.";
```

This was introduced when refactoring the audience validation. I don't think I would consider this a behaviour change as much as a change in the information being logged.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+                    tokenValidationParametersResult.ClaimsIdentity,
+                    validationParametersResult.UnwrapResult().ClaimsIdentity,
+                    context);
+                IdentityComparer.AreEqual(
+                    tokenValidationParametersResult.Claims,
+                    validationParametersResult.UnwrapResult().Claims,
+                    context);
+            }
+            else
+            {
```
sruke: Should the exception message or type be compared between tokenValidationParametersResult and validationParametersResult to ensure they match, except in one scenario to account for the removal if ValidAudience property from the ValidationParameters class?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
sruke: Consider documenting the change in behavior in the release notes to ensure customers are informed when they onboard to the new path. I see three scenarios in this file where the log message IDs differ.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,215 @@
...
+                        ValidationParameters = CreateValidationParameters([Default.Audience]),
+                        Audience = null,
+                        ExpectedIsValid = false,
+                        ExpectedException = ExpectedException.SecurityTokenInvalidAudienceException("IDX10206:"),
+                    },
+                    new ValidateTokenAsyncAudienceTheoryData("Invalid_ValidAudiencesIsNull")
+                    {
+                        TokenValidationParameters = CreateTokenValidationParameters(null),
+                        ValidationParameters = CreateValidationParameters(null),
+                        Audience = string.Empty,
```
sruke, line 105: nit: Should this be a valid Audience while the Audiences property on TVP or VP is null?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Addressing this as part of the extracted common code on the lifetime PR.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Replied on the other PR.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: The current validation path considers the case where no valid audiences are provided to be invalid and throws.
The new validation path took that as a base and kept the same behaviour.

If you believe that having no valid audiences should be considered the same as skipping the validation, we can look into it and find out what the correct behaviour would be.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Currently we have two scenarios:
- if no custom exception is specified for the new path, both paths are compared against the same expected exception, so the types are validated to be the same
- if a custom exception is specified for the new path, each exception is compared against its own expected exception. In this scenario, we could add an extra check to ensure that the "legacy" and the "new" exception types still match even if the messages are different.

sruke: Adding an additional check to ensure the exception types remain the same would be beneficial as any change to the expected exception type is a breaking change.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
brentschmaltz: The name passed to TestUtilities.WriteHeader should be the same as the name of the test method.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,213 @@
...
+            {
+                return new TheoryData<ValidateTokenAsyncAudienceTheoryData>
+                {
+                    new ValidateTokenAsyncAudienceTheoryData("Valid_AudiencesMatch")
+                    {
+                        Audience = Default.Audience,
+                        TokenValidationParameters = CreateTokenValidationParameters([Default.Audience]),
+                        ValidationParameters = CreateValidationParameters([Default.Audience]),
+                    },
+                    new ValidateTokenAsyncAudienceTheoryData("Invalid_AudiencesDontMatch")
```
brentschmaltz, line 39: Suggest grouping all Valid and Invalid together.
Could add a #region Valid.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
brentschmaltz: It is OK to have a different error message, IDX10214 vs. IDX10215

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,213 @@
...
+                        // VP path has a default empty List for ValidAudiences, so it will always return IDX10206 if no audiences are provided.
+                        ExpectedExceptionValidationParameters = ExpectedException.SecurityTokenInvalidAudienceException("IDX10206:"),
+                    },
+                };
+
+                static TokenValidationParameters CreateTokenValidationParameters(
+                    List<string>? audiences,
+                    bool ignoreTrailingSlashWhenValidatingAudience = false) =>
+
+                    new TokenValidationParameters
```
brentschmaltz, line 119: We should only be validating the Audience, all other checks should be off.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
brentschmaltz: Yes we need to make sure the Exception is the same.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,213 @@
...
+
+            SecurityTokenDescriptor securityTokenDescriptor = new SecurityTokenDescriptor
+            {
+                Subject = Default.ClaimsIdentity,
+                SigningCredentials = Default.AsymmetricSigningCredentials,
+                Audience = theoryData.Audience,
+                Issuer = Default.Issuer,
+            };
+
+            string jwtString = jsonWebTokenHandler.CreateToken(securityTokenDescriptor);
```
brentschmaltz: We should not need to create signed tokens or validate the signature.
This increases complexity of these tests that are specific to Audience validation.

see:  JsonUtilities.CreateUnsignedToken
Which is a quick way to create a JWT with a specific {claim, claimvalue} pair.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
```diff
@@ -0,0 +1,213 @@
...
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.TestUtils;
+using Microsoft.IdentityModel.Tokens;
+using Xunit;
+
+namespace Microsoft.IdentityModel.JsonWebTokens.Tests
+{
+    public partial class JsonWebTokenHandlerValidateTokenAsyncTests
+    {
+        [Theory, MemberData(nameof(ValidateTokenAsync_AudienceTestCases))]
```
brentschmaltz: Please make sure 'DisableDiscoveryEnumeration = true' is added to all tests.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
brentschmaltz: If the goal is to find the Audience tests, then a different ns such as 'Microsoft.IdentityModel.JsonWebTokens.Audience.Tests' would create a hierarchy that would have one drop down for each validation step.

Not sure why a partial class is helpful.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
jennyf19: okay. right. It might be helpful to mark TVP as legacy or put "legacy TVP path"

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
jennyf19: It moves them under the same type of class. I can see why this would be helpful, but it's also not the style currently used in the library, so that could cause some confusion. Might be easier to just update the namespace and keep the consistency. Sometimes, having to context switch, even for something small like that, can add up.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
jennyf19: Customers "should not" but they do. Would like to keep the error messages and types of exceptions returned the same. If it's not possible, let's discuss as group, I would consider those as breaking changes.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
jennyf19: I don't think SEAL/R9 team would agree on this. They would consider it a breaking change. Let's discuss.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
jennyf19: We could add that as a follow-up to all the tests right? I think it's something we should fix all-up and maybe do in one large PR like @kellyyangsong did in SAL today?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Removed signature from the token and skipped all validations but audience.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Removed signature from the token and skipped all validations but audience.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: I tried to group them by scenario:
i.e. the audience has a slash at the end, it is valid if the "IgnoreTrailingSlashes..." flag is set, but invalid if it is false.

Looking to make it easier for someone new to the project looking at the tests to understand the use cases.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Exception type is checked to be the same now. We need to discuss about the new messages for some cases.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Decided to stick to `ExpectedException` as it is the general rule. In all cases where the two exceptions are 100% the same, having `TVPExpectedException` would be misleading.
Cases where the error message is different in the VP path are a minority and we are to discuss whether we can keep them.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Updated.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Leaving this open until we can have a conversation around the topic.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: The types of exceptions are being kept and validated. Let's discuss about the message as some messages are very specific to things that have changed, like fields that exist in TokenValidationParameters but no longer in ValidationParameters.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Extracted the duplicated code to a common file.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
kllysng: I'm working on it :)

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
```diff
@@ -0,0 +1,32 @@
...
+        private string? _invalidAudience;
+
+        public AudienceValidationError(
+            MessageDetail messageDetail,
+            Type exceptionType,
+            StackFrame stackFrame,
+            IList<string>? invalidAudiences)
+            : base(messageDetail, ValidationFailureType.AudienceValidationFailed, exceptionType, stackFrame)
+        {
+            _invalidAudience = Utility.SerializeAsSingleCommaDelimitedString(invalidAudiences);
```
brentschmaltz: Do we need to do this in the ctor?
There is a possibility that this string will never be used.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.SkipValidations.cs
```diff
@@ -0,0 +1,98 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+#nullable enable
+namespace Microsoft.IdentityModel.Tokens
+{
+    public static partial class Validators
```
brentschmaltz: Should these be in test area?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
brentschmaltz: @jennyf19 that PR was committed, we should follow the pattern for all new work.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.SkipValidations.cs
iNinja: Moved to TestUtils.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
iNinja: Moved out of the ctor.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: Added this to the new tests.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
jennyf19: I think we aligned that we cannot change the `IDX???` value but we can change the underlying error message content.

jennyf19 (approved):
can we open an issue to track the changes to the `IDX???` values so we can revisit before releasing?
:shipit:

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Audience.cs
iNinja: We should always keep the exception type, such as `SecurityTokenInvalidAudienceException`.
We need to finish defining the need to change the `IDX` number on limited circumstances such as the data being passed as part of the message changing.

Otherwise, we need to look into how to modify the base message so that it can fit the two scenarios.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
```diff
@@ -0,0 +1,32 @@
...
+            MessageDetail messageDetail,
+            Type exceptionType,
+            StackFrame stackFrame,
+            IList<string>? invalidAudiences)
+            : base(messageDetail, ValidationFailureType.AudienceValidationFailed, exceptionType, stackFrame)
+        {
+            _invalidAudiences = invalidAudiences;
+        }
+
+        internal override void AddAdditionalInformation(ISecurityTokenException exception)
```
brentschmaltz, line 25: I suspect that since we are creating derived ValidationError types, we do not need the method AddAdditionalInformation or the interface ISecurityTokenException.

For example, when we create the AudienceValidationError, we call the ctor that takes the list of available audiences.

iNinja: Let’s keep the conversation going on that to find the best solution that can support any exception our users may decide to return.
I’d rather not expand the scope of this PR to that as it’s already a week in review.

FuPingFranco (approved):

PR #2839, opened 2024-09-23
Regression tests: Lifetime

iNinja:
# Regression tests: Lifetime
- Added scenarios for lifetime related validations.

Note: Upcoming PRs will continue to extract the different aspects of regression tests into separate files.

Part of #2711

### DIFF
--- a/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
+++ b/test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
@@ -3,7 +3,6 @@
 
 #nullable enable
 using System;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.IdentityModel.TestUtils;
 using Microsoft.IdentityModel.Tokens;
@@ -18,62 +17,9 @@ namespace Microsoft.IdentityModel.JsonWebTokens.Tests
         {
             var context = TestUtilities.WriteHeader($"{this}.ValidateTokenAsync_Lifetime", theoryData);
 
-            JsonWebTokenHandler jsonWebTokenHandler = new JsonWebTokenHandler();
-            jsonWebTokenHandler.SetDefaultTimesOnTokenCreation = false; // Allow for null values to be passed in to validate.
+            string jwtString = CreateToken(theoryData.IssuedAt, theoryData.NotBefore, theoryData.Expires);
 
-            SecurityTokenDescriptor securityTokenDescriptor = new SecurityTokenDescriptor
-            {
-                Subject = Default.ClaimsIdentity,
-                SigningCredentials = Default.AsymmetricSigningCredentials,
-                Audience = Default.Audience,
-                Issuer = Default.Issuer,
-                IssuedAt = theoryData.IssuedAt,
-                NotBefore = theoryData.NotBefore,
-                Expires = theoryData.Expires,
-            };
-
-            string jwtString = jsonWebTokenHandler.CreateToken(securityTokenDescriptor);
-
-            TokenValidationResult tokenValidationParametersResult =
-                    await jsonWebTokenHandler.ValidateTokenAsync(jwtString, theoryData.TokenValidationParameters);
-            ValidationResult<ValidatedToken> validationParametersResult =
-                await jsonWebTokenHandler.ValidateTokenAsync(
-                    jwtString, theoryData.ValidationParameters!, theoryData.CallContext, CancellationToken.None);
-
-            if (tokenValidationParametersResult.IsValid != theoryData.ExpectedIsValid)
-                context.AddDiff($"tokenValidationParametersResult.IsValid != theoryData.ExpectedIsValid");
-
-            if (validationParametersResult.IsSuccess != theoryData.ExpectedIsValid)
-                context.AddDiff($"validationParametersResult.IsSuccess != theoryData.ExpectedIsValid");
-
-            if (theoryData.ExpectedIsValid &&
-                tokenValidationParametersResult.IsValid &&
-                validationParametersResult.IsSuccess)
-            {
-                IdentityComparer.AreEqual(
-                    tokenValidationParametersResult.ClaimsIdentity,
-                    validationParametersResult.UnwrapResult().ClaimsIdentity,
-                    context);
-                IdentityComparer.AreEqual(
-                    tokenValidationParametersResult.Claims,
-                    validationParametersResult.UnwrapResult().Claims,
-                    context);
-            }
-            else
-            {
-                theoryData.ExpectedException.ProcessException(tokenValidationParametersResult.Exception, context);
-
-                if (!validationParametersResult.IsSuccess)
-                {
-                    // If there is a special case for the ValidationParameters path, use that.
-                    if (theoryData.ExpectedExceptionValidationParameters != null)
-                        theoryData.ExpectedExceptionValidationParameters
-                            .ProcessException(validationParametersResult.UnwrapError().GetException(), context);
-                    else
-                        theoryData.ExpectedException
-                            .ProcessException(validationParametersResult.UnwrapError().GetException(), context);
-                }
-            }
+            await ValidateAndCompareResults(jwtString, theoryData, context);
 
             TestUtilities.AssertFailIfErrors(context);
         }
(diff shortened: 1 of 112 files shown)


test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
```diff
@@ -0,0 +1,219 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.TestUtils;
+using Microsoft.IdentityModel.Tokens;
+using Xunit;
+
+namespace Microsoft.IdentityModel.JsonWebTokens.Tests
+{
+    public partial class JsonWebTokenHandlerValidateTokenAsyncTests
```
jennyf19, line 13: why partial? also I can't get these tests to show up in the VS test runner.

jennyf19: got it work after re-starting

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
```diff
@@ -0,0 +1,219 @@
...
+                Audience = Default.Audience,
+                Issuer = Default.Issuer,
+                IssuedAt = theoryData.IssuedAt,
+                NotBefore = theoryData.NotBefore,
+                Expires = theoryData.Expires,
+            };
+
+            string jwtString = jsonWebTokenHandler.CreateToken(securityTokenDescriptor);
+
+            TokenValidationResult tokenValidationParametersResult =
```
jennyf19: nit: improve the variable name to indicate `legacy`

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
```diff
@@ -0,0 +1,219 @@
...
+                context.AddDiff($"tokenValidationParametersResult.IsValid != theoryData.ExpectedIsValid");
+
+            if (validationParametersResult.IsSuccess != theoryData.ExpectedIsValid)
+                context.AddDiff($"validationParametersResult.IsSuccess != theoryData.ExpectedIsValid");
+
+            if (theoryData.ExpectedIsValid &&
+                tokenValidationParametersResult.IsValid &&
+                validationParametersResult.IsSuccess)
+            {
+                IdentityComparer.AreEqual(
```
jennyf19: do we want to ensure that each specific claim and its value are correct, not just that the claims exist?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
```diff
@@ -0,0 +1,219 @@
...
+                    validationParametersResult.UnwrapResult().ClaimsIdentity,
+                    context);
+                IdentityComparer.AreEqual(
+                    tokenValidationParametersResult.Claims,
+                    validationParametersResult.UnwrapResult().Claims,
+                    context);
+            }
+            else
+            {
+                theoryData.ExpectedException.ProcessException(tokenValidationParametersResult.Exception, context);
```
jennyf19: Do we need to do more here?  random thoughts: check the type, message, and other properties of the exceptions and ensure that the specific error messages generated are as expected, not just the presence of an exception? or do we think this is enough?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
```diff
@@ -0,0 +1,219 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#nullable enable
```
FuPingFranco, line 4: nit: add a new line after directive to be consistent with Audience regression test PR or vice versa.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
FuPingFranco: +1

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
iNinja: Explained on the other PR, but the main idea is that we split all the regression tests across files for readability and ease to follow. We are separating them by aspect of the validation: lifetime, audience, issuer, replay, actor, signature, token type, encryption, and by using the partial class we can have all the test methods show under the single class in the text explorer.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
iNinja: We are comparing one set of claims against the other, not just that the claims exist.
If they differ, it gets marked in the context and the test fails.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
iNinja: At this point we check that:
- The type of the exception is the one we expect
- (if set) the inner exception matches the type we expect
- The `IDX` identifier matches the expectation

What we do not check here:
- The exact contents after the `IDX123123:`
- Custom exception properties

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
iNinja: Renamed to `legacyTokenValidationParametersResult`. Please let me know if you think of something better.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
iNinja: Added checks for the 2 points mentioned.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.TestUtils;
+using Microsoft.IdentityModel.Tokens;
+using Xunit;
+
+namespace Microsoft.IdentityModel.JsonWebTokens.Tests
+{
+    public partial class JsonWebTokenHandlerValidateTokenAsyncTests
+    {
+        [Theory, MemberData(nameof(ValidateTokenAsync_LifetimeTestCases))]
```
kllysng:
```suggestion
        [Theory, MemberData(nameof(ValidateTokenAsync_LifetimeTestCases), DisableDiscoveryEnumeration = true)]
```

We want to add this to all of our Theory tests. This enables us to load and run tests faster.

jennyf19 (approved):
Same for the audiences PR, if we can track all changes to `IDX???` error codes in one GitHub issue, that would be great.

FuPingFranco (approved):
LGTM

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.cs
kllysng: @iNinja why was this comment resolved?

Issue #2842, opened 2024-09-24
[Bug] Suppress / fix policheck failure


keegan-caruso:
See: https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1366079&view=sariftools.scans.build-tab

PR #2843, opened 2024-09-24
fix policheck issue

jennyf19:
#2842


keegan-caruso (approved):

westin-m (approved):
Definitely not profanity!

PR #2852, opened 2024-09-27
Set custom BenchmarkDotNetconfig as default

pmaytak:
Currently, if you run the tests from command line and specify runtimes (`--runtimes net8.0 net9.0`), BDN will use a default config for the runtimes in addition to running tests for the custom config.

This change makes sure our custom BDN config is used as default.

See [BDN doc](https://benchmarkdotnet.org/articles/guides/console-args.html#specifying-custom-default-settings-for-console-argument-parser).

### DIFF
--- a/benchmark/Microsoft.IdentityModel.Benchmarks/BenchmarkConfig.cs
+++ b/benchmark/Microsoft.IdentityModel.Benchmarks/BenchmarkConfig.cs
@@ -16,14 +16,15 @@ namespace Microsoft.IdentityModel.Benchmarks
         {
             AddJob(Job.MediumRun
                 .WithLaunchCount(4)
-                .WithMaxAbsoluteError(TimeInterval.FromMilliseconds(10)).AsDefault())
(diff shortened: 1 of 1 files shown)


benchmark/Microsoft.IdentityModel.Benchmarks/BenchmarkConfig.cs
```diff
@@ -16,7 +16,7 @@ public BenchmarkConfig()
         {
             AddJob(Job.MediumRun
                 .WithLaunchCount(4)
-                .WithMaxAbsoluteError(TimeInterval.FromMilliseconds(10)))
+                .WithMaxAbsoluteError(TimeInterval.FromMilliseconds(10)).AsDefault())
```
msbw2:
```suggestion
                .WithMaxAbsoluteError(TimeInterval.FromMilliseconds(10))
                .AsDefault())
```

jennyf19 (approved):
:shipit:

msbw2 (approved):

PR #2853, opened 2024-09-27
Implement New Audience Validator Method in SAML2TokenHandler

FuPingFranco:
# Implement New Audience Validator Method in SAML2TokenHandler

## Description

Adding new internal method to validate audience in SAML2TokenHandler

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/LogMessages.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/LogMessages.cs
@@ -12,6 +12,15 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
     internal static class LogMessages
     {
 #pragma warning disable 1591
+
+        // general
+        internal const string IDX10000 = "IDX10000: The parameter '{0}' cannot be a 'null' or an empty object. "; //Q: Should we just use one set of validation errors?
+
+        // Audience validation errors
+        internal const string IDX10206 = "IDX10206: Unable to validate audience. The 'audiences' parameter is empty.";
+        internal const string IDX10207 = "IDX10207: Unable to validate audience. The 'audiences' parameter is null.";
+        internal const string IDX10215 = "IDX10215: Audience validation failed. Audiences: '{0}'. Did not match: validationParameters.ValidAudiences: '{1}'.";
+
         // token validation
         internal const string IDX13400 = "IDX13400: The '{0}', can only process SecurityTokens of type: '{1}'. The SecurityToken received is of type: '{2}'.";
         internal const string IDX13001 = "IDX13001: A SAML2 assertion that specifies an AuthenticationContext DeclarationReference is not supported.To handle DeclarationReference, extend the Saml2SecurityTokenHandler and override ProcessAuthenticationStatement.";
--- a/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateToken.Internal.cs
+++ b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateToken.Internal.cs
@@ -1,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+
+#nullable enable
+namespace Microsoft.IdentityModel.Tokens.Saml2
+{
+    /// <summary>
+    /// A <see cref="SecurityTokenHandler"/> designed for creating and validating Saml2 Tokens. See: http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf
+    /// </summary>
+    public partial class Saml2SecurityTokenHandler : SecurityTokenHandler
+    {
+        /// <summary>
+        /// Determines if the audience found in a <see cref="Saml2SecurityToken"/> is valid.
+        /// </summary>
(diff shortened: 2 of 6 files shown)


test/Microsoft.IdentityModel.Tokens.Saml.Tests/ValidateTheoryData.cs
```diff
@@ -16,14 +15,14 @@ public static void AddValidateAudienceTheoryData(List<TokenTheoryData> theoryDat
                 Audiences = new List<string>(),
                 First = true,
                 ExpectedException = ExpectedException.ArgumentNullException("IDX10000:"),
-                TestId = "TokenValidationParameters null",
+                TestId = "TokenValidationParameters_Null",
```
kllysng, line 18: Instead of setting TestId this way - could you do it like so `new TokenTheoryData("TokenValidationParameters_Null")`? TokenTheoryData already has a ctor that takes a TestId and doing so this way allows `theoryData.CallContext.DebugId` to be set.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/LogMessages.cs
```diff
@@ -12,6 +12,15 @@ namespace Microsoft.IdentityModel.Tokens.Saml2
     internal static class LogMessages
     {
 #pragma warning disable 1591
+
+        // general
+        internal const string IDX10000 = "IDX10000: The parameter '{0}' cannot be a 'null' or an empty object. "; //Q: Should we just use one set of validation errors?
```
jennyf19, line 17: why do we need new `IDX????` codes?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateToken.Internal.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System.Collections.Generic;
+
+#nullable enable
+namespace Microsoft.IdentityModel.Tokens.Saml2
+{
+    /// <summary>
+    /// A <see cref="SecurityTokenHandler"/> designed for creating and validating Saml2 Tokens. See: http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf
+    /// </summary>
+    public partial class Saml2SecurityTokenHandler : SecurityTokenHandler
```
jennyf19, line 12: should this be internal for now?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateToken.Internal.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// Determines if the audience found in a <see cref="Saml2SecurityToken"/> is valid.
+        /// </summary>
+        /// <param name="audiences">The audiences found in the <see cref="Saml2SecurityToken"/></param>
+        /// <param name="securityToken">The <see cref="Saml2SecurityToken"/> that is being validated.</param>
+        /// <param name="validationParameters">The <see cref="ValidationParameters"/> to be used for validating the token.</param>
+        /// <param name="callContext"></param>
+        /// <remarks><see cref="Validators.ValidateAudience(IList{string}, SecurityToken, ValidationParameters, CallContext)"/> for additional details.</remarks>
+        internal static ValidationResult<string> ValidateAudience(IList<string> audiences, SecurityToken securityToken, ValidationParameters validationParameters, CallContext callContext)
+        {
+            return Validators.ValidateAudience(audiences, securityToken, validationParameters, callContext);
```
iNinja, line 24: I do not think that adding methods simply to call the delegates is a good idea. We can refactor the methods that call this to use the delegates directly.

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2SecurityTokenHandlerTests.Validations.cs
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            TestUtilities.AssertFailIfErrors(context);
+        }
+
+        public static TheoryData<AudienceValidationTheoryData> ValidateAudienceParameterTestCases
+        {
+            get
+            {
+                return new TheoryData<AudienceValidationTheoryData>
```
iNinja, line 59: Are all these tests re-testing the audience validation delegate that already has these tests?
See: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/test/Microsoft.IdentityModel.Tokens.Tests/Validation/AudienceValidationResultTests.cs

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/LogMessages.cs
iNinja: If this is just to use in the tests, we could simply import the `LogMessages` from the outer package, Tokens.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateToken.Internal.cs
iNinja: Please see this draft: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2855

FuPingFranco:
Closing this PR in favor of a different agreed approach.

PR #2857, opened 2024-09-30
Use FixedTimeEquals in NETCore targets 

### LINKED ISSUE #2825: [Feature Request] Use fixedtimeequals on Net Core targets
This code:  https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/411d0841bf30db8bf1a8448e255cacf7b77534c4/src/Microsoft.IdentityModel.Tokens/Utility.cs#L155-L158

Should be replaced with [cryptographicoperations.fixedtimeequals](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.cryptographicoperations.fixedtimeequals?view=net-8.0)

On Net Core targets

westin-m:
Closes #2825

There is another class in wilson named CryptographicOperations, so the fully qualified name is necessary.

| without FixedTimeEquals | Mean     | Error    | StdDev   |
|-------------------- |---------:|---------:|---------:|
| AreEqual_ByteArrays | 31.59 ns | 0.185 ns | 0.155 ns |
| AreEqual_ByteSpans  | 30.66 ns | 0.169 ns | 0.141 ns |

| with FixedTimeEquals | Mean     | Error   | StdDev  |
|-------------------- |---------:|--------:|--------:|
| AreEqual_ByteArrays | 150.2 ns | 0.47 ns | 0.44 ns |
| AreEqual_ByteSpans  | 160.1 ns | 0.56 ns | 0.52 ns |

The following results are after applying NoOptimization
| without FixedTimeEquals              | Mean     | Error   | StdDev  |
|-------------------- |---------:|--------:|--------:|
| AreEqual_ByteArrays | 173.8 ns | 0.36 ns | 0.33 ns |
| AreEqual_ByteSpans  | 154.4 ns | 0.21 ns | 0.19 ns |

| with FixedTimeEquals              | Mean     | Error   | StdDev  |
|-------------------- |---------:|--------:|--------:|
| AreEqual_ByteArrays | 157.8 ns | 1.96 ns | 1.83 ns |
| AreEqual_ByteSpans  | 155.4 ns | 0.20 ns | 0.19 ns |


kllysng (dismissed):

msbw2 (approved):

pmaytak:
Could you also run a small benchmark?

keegan-caruso:
A similar change is needed for the ReadOnlySpan overload.

armandRobled:
Patche-2_Robledo6

kllysng:
> Could you also run a small benchmark?

Are we okay with these benchmarks? Not sure if the results are as expected 🤔

westin-m:
Reopening this because our current code is getting optimized, which we don't want.

pmaytak (approved):
Both of these methods are covered by tests?

westin-m:
> Both of these methods are covered by tests?

Yes, there is already coverage.

JoshLozensky (approved):

PR #2865, opened 2024-10-02
Set internal _syncAfter using only AutomaticRefreshInterval.

brentschmaltz:
The internal _syncAfter field controls if metadata is refreshed (http call).
The value was set to the minimum of AutomaticRefreshInterval and RefreshInterval.

RefreshInterval controls the interval that must pass for RequestRefresh() to make an http call.
RefreshInterval is usually a much smaller timespan, so in effect many more calls would be made to obtain metadata.

This fix only uses the AutomaticRefreshInterval to set _syncAfter.
This has the side effect of extending the time for the next AutomaticRequest, however, that makes sense.

#2866

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -313,7 +313,7 @@ namespace Microsoft.IdentityModel.Protocols
             }
             finally
             {
-                _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
+                _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval);
                 Interlocked.Exchange(ref _configurationRetrieverState, ConfigurationRetrieverIdle);
             }
 #pragma warning restore CA1031 // Do not catch general exception types
--- a/test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
+++ b/test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
@@ -511,6 +511,48 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests
         }
 
         [Fact]
+        public async Task CheckSyncAfter()
+        {
+            // This test checks that the _syncAfter field is set correctly after a refresh.
+            var context = new CompareContext($"{this}.CheckSyncAfter");
+
+            var docRetriever = new FileDocumentRetriever();
+            var configManager = new ConfigurationManager<OpenIdConnectConfiguration>("OpenIdConnectMetadata.json", new OpenIdConnectConfigurationRetriever(), docRetriever);
+
+            // This is the minimum time that should pass before an automatic refresh occurs
+            // stored in advance to avoid any time drift issues.
+            DateTimeOffset minimumRefreshInterval = DateTimeOffset.UtcNow + configManager.AutomaticRefreshInterval;
+
+            // get the first configuration, internal _syncAfter should be set to a time greater than UtcNow + AutomaticRefreshInterval.
+            var configuration = await configManager.GetConfigurationAsync(CancellationToken.None);
+
+            // force a refresh by setting internal field
+            TestUtilities.SetField(configManager, "_syncAfter", DateTimeOffset.UtcNow - TimeSpan.FromHours(1));
+            configuration = await configManager.GetConfigurationAsync(CancellationToken.None);
+            // wait 100ms here because update of config is run as a new task.
+            Thread.Sleep(100);
+
+            // check that _syncAfter is greater than DateTimeOffset.UtcNow + AutomaticRefreshInterval
+            DateTimeOffset syncAfter = (DateTimeOffset)TestUtilities.GetField(configManager, "_syncAfter");
+            if (syncAfter < minimumRefreshInterval)
+                context.Diffs.Add($"(AutomaticRefreshInterval) syncAfter '{syncAfter}' < DateTimeOffset.UtcNow + configManager.AutomaticRefreshInterval: '{minimumRefreshInterval}'.");
+
+            // make same check for RequestRefresh
+            // force a refresh by setting internal field
+            TestUtilities.SetField(configManager, "_lastRequestRefresh", DateTimeOffset.UtcNow - TimeSpan.FromHours(1));
+            configManager.RequestRefresh();
+            // wait 100ms here because update of config is run as a new task.
+            Thread.Sleep(100);
+
+            // check that _syncAfter is greater than DateTimeOffset.UtcNow + AutomaticRefreshInterval
+            syncAfter = (DateTimeOffset)TestUtilities.GetField(configManager, "_syncAfter");
+            if (syncAfter < minimumRefreshInterval)
+                context.Diffs.Add($"(RequestRefresh) syncAfter '{syncAfter}' < DateTimeOffset.UtcNow + configManager.AutomaticRefreshInterval: '{minimumRefreshInterval}'.");
+
+            TestUtilities.AssertFailIfErrors(context);
+        }
+
+        [Fact]
         public async Task GetConfigurationAsync()
         {
             var docRetriever = new FileDocumentRetriever();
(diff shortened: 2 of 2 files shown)


test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -510,6 +510,48 @@ public static TheoryData<ConfigurationManagerTheoryData<OpenIdConnectConfigurati
...
+            // stored in advance to avoid any time drift issues.
+            DateTimeOffset minimumRefreshInterval = DateTimeOffset.UtcNow + configManager.AutomaticRefreshInterval;
+
+            // get the first configuration, internal _syncAfter should be set to a time greater than UtcNow + AutomaticRefreshInterval.
+            var configuration = await configManager.GetConfigurationAsync(CancellationToken.None);
+
+            // force a refresh by setting internal field
+            TestUtilities.SetField(configManager, "_syncAfter", DateTimeOffset.UtcNow - TimeSpan.FromHours(1));
+            configuration = await configManager.GetConfigurationAsync(CancellationToken.None);
+            // wait 100ms here because update of config is run as a new task.
```
msbw2: This sounds flaky. Is there a way to store a reference to the fire-and-forget `Task` and expose it only for tests?

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -510,6 +510,48 @@ public static TheoryData<ConfigurationManagerTheoryData<OpenIdConnectConfigurati
...
+            Thread.Sleep(100);
+
+            // check that _syncAfter is greater than DateTimeOffset.UtcNow + AutomaticRefreshInterval
```
msbw2, line 535: Can we introduce a `TimeProvider` to make this an exact comparison? Then you wouldn't need to access internal state to "advance" time.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -510,6 +510,48 @@ public static TheoryData<ConfigurationManagerTheoryData<OpenIdConnectConfigurati
...
+            DateTimeOffset syncAfter = (DateTimeOffset)TestUtilities.GetField(configManager, "_syncAfter");
+            if (syncAfter < minimumRefreshInterval)
```
msbw2, line 537: To fix the current test failure:
```suggestion
            if (syncAfter <= minimumRefreshInterval)
```

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -510,6 +510,48 @@ public static TheoryData<ConfigurationManagerTheoryData<OpenIdConnectConfigurati
...
+            // make same check for RequestRefresh
+            // force a refresh by setting internal field
+            TestUtilities.SetField(configManager, "_lastRequestRefresh", DateTimeOffset.UtcNow - TimeSpan.FromHours(1));
+            configManager.RequestRefresh();
+            // wait 100ms here because update of config is run as a new task.
+            Thread.Sleep(100);
+
+            // check that _syncAfter is greater than DateTimeOffset.UtcNow + AutomaticRefreshInterval
+            syncAfter = (DateTimeOffset)TestUtilities.GetField(configManager, "_syncAfter");
+            if (syncAfter < minimumRefreshInterval)
```
msbw2, line 549: Same here:
```suggestion
            if (syncAfter <= minimumRefreshInterval)
```

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -510,6 +510,48 @@ public static TheoryData<ConfigurationManagerTheoryData<OpenIdConnectConfigurati
...
+            var configuration = await configManager.GetConfigurationAsync(CancellationToken.None);
+
+            // force a refresh by setting internal field
+            TestUtilities.SetField(configManager, "_syncAfter", DateTimeOffset.UtcNow - TimeSpan.FromHours(1));
+            configuration = await configManager.GetConfigurationAsync(CancellationToken.None);
+            // wait 100ms here because update of config is run as a new task.
+            Thread.Sleep(100);
+
+            // check that _syncAfter is greater than DateTimeOffset.UtcNow + AutomaticRefreshInterval
+            DateTimeOffset syncAfter = (DateTimeOffset)TestUtilities.GetField(configManager, "_syncAfter");
```
msbw2, line 536: To make this look cleaner, could you change `TestUtilities.GetField` to be generic?
```suggestion
            var syncAfter = TestUtilities.GetField<DateTimeOffset>(configManager, "_syncAfter");
```

GeoK (approved):

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
brentschmaltz: @msbw2 this seems to work.
Can you provide an example of what you would like to see?

msbw2: I'm thinking of an `internal` property `UpdateCurrentConfigurationTask` which is only used in tests to `await` the task to avoid sleeping and hoping the task isn't delayed for some reason. I would consider this a non-blocker, though.

msbw2: I think this might be why this test is failing. You set `_syncAfter` above to (now - 1h), and the failure message indicates that that's the value when you retrieve it below. It seems 100ms isn't long enough for the task to be reliably run, but I'd advise against increasing it experimentally.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -313,7 +313,7 @@ private void UpdateCurrentConfiguration()
             }
             finally
             {
-                _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval < RefreshInterval ? AutomaticRefreshInterval : RefreshInterval);
+                _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval);
                 Interlocked.Exchange(ref _configurationRetrieverState, ConfigurationRetrieverIdle);
```
pmaytak, line 317: We don't seem to set `_configurationRetrieverState` to Running anywhere?

msbw2: It looks like it is set to running just before this method gets started in a new task. Or does the fact that it's in a new task change the call context?

kllysng: On ln 251? It's being set to ` ConfigurationRetrieverIdle`. Regardless, I think we should address it outside of this PR

kllysng (approved):

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
msbw2: Oh, so the arguments are flipped in the call to `Interlocked.CompareExchange`.

jennyf19 (approved):

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
pmaytak: Yes, I think so. (The CompareExchange is not super intuitive since it compares the first and third parameters...)
```csharp
if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) != ConfigurationRetrieverRunning)
                {
                }
```
Logic:
If the state is idle (is not running),
   then set state to running and call the update
If the state is running
   then don't do anything.

at the end of Update method, we correctly reset the state to idle https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/e8e9ecfe074b272e4c101b8865ee63409f59b400/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs#L317

PR #2870, opened 2024-10-03
Adjust for RefreshInterval not influencing AutomaticRefreshInterval.

### LINKED ISSUE #2774: [Bug] Starvation: AAD Token Validation
Note that to get help, you need to run the latest version. 
```csharp
    <PackageVersion Include="IdentityModel" Version="7.0.0" />    
    <PackageVersion Include="Microsoft.Identity.Web" Version="2.20.0" />
```

**Where is the issue?**
 * [x] M.IM.Validators

existing

**Repro**
If you send 200 simultaneous requests to an API when the applicaiton first starts then the application will be unresponsive for up to a minute while we believe it makes requests to AAD to get the metadata required to validate tokens on the request.  After "warming up" the issue goes away.

I think I found the cause. And I think I found a workaround. But I reckon it's something worthy of a fix.

## Cause

The server is configured to use AAD here:

```csharp
builder.Services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
    .AddMicrosoftIdentityWebApi(_ => { }, identity =>
    {
        identity.TenantId = config.Authentication.TenantId;
        identity.ClientId = config.Authentication.ClientId;
        identity.Instance = config.Authentication.Instance;
    });
```

And when the 200 requests arrive at once, the [code in AadIssuerValidator](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/d2632a77aa627707d3e2f04248daea62ab45fd47/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs#L214) calls [GetBaseConfigurationAsync](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/d2632a77aa627707d3e2f04248daea62ab45fd47/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs#L256) which in turn [calls GetConfigurationAsync](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/d2632a77aa627707d3e2f04248daea62ab45fd47/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs#L234) which in turn fails this condition [the following check](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/d2632a77aa627707d3e2f04248daea62ab45fd47/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs#L146) the first 200 (or however many concurrent requests came in) times:

```csharp
{
    if (_currentConfiguration != null && _syncAfter > DateTimeOffset.UtcNow)
    {
        return _currentConfiguration;
    }

    await _refreshLock.WaitAsync(cancel).ConfigureAwait(false);
    try
    {
        if (_syncAfter <= DateTimeOffset.UtcNow)
        {
```

and then the remaining 199 requests all queue up waiting for that lock...

## workaround
we created a startup probe, that runs before the container accepts traffic, that runs the following code and basically hydrates the metadata using the config options

```csharp
        var options = _optionsMonitor.Get(JwtBearerDefaults.AuthenticationScheme);
        var metadata = await options.ConfigurationManager!.GetConfigurationAsync(CancellationToken.None);

        var validator = _issuerValidatorFactory.GetAadIssuerValidator(options.Authority!);
        var configManagerPropertyV1 = GetConfigurationManagerPropertyInfo("ConfigurationManagerV1");
        var configManagerPropertyV2 = GetConfigurationManagerPropertyInfo("ConfigurationManagerV2");

        try
        {
            var v1Metadata = await GetConfigurationManagerConfig(configManagerPropertyV1, validator);
            var v2Metadata = await GetConfigurationManagerConfig(configManagerPropertyV2, validator);

            if (metadata.SigningKeys.Any() && v1Metadata.SigningKeys.Any() && v2Metadata.SigningKeys.Any())
            {
                return HealthCheckResult.Healthy();
            }
        }
        catch (Exception ex)
        {
            return HealthCheckResult.Unhealthy(exception: ex);
        }
        

        return HealthCheckResult.Unhealthy();
}
    /// <summary>
    /// Get oauth metadata by using reflection to access the internal <see cref="ConfigurationManager{OpenIdConnectConfiguration}"/>
    /// in <see cref="AadIssuerValidator"/>.
    /// </summary>
    /// <param name="configManagerProperty"></param>
    /// <param name="validator"></param>
    /// <returns></returns>
    private static async Task<OpenIdConnectConfiguration> GetConfigurationManagerConfig(PropertyInfo? configManagerProperty,
        AadIssuerValidator validator)
    {
        var getter = configManagerProperty!.GetGetMethod(nonPublic: true);
        var configManager = getter!.Invoke(validator, null) as ConfigurationManager<OpenIdConnectConfiguration>;
        return await configManager!.GetConfigurationAsync(CancellationToken.None);
    }

    private static PropertyInfo GetConfigurationManagerPropertyInfo(string propertyName)
    {
        return typeof(AadIssuerValidator).GetProperty(propertyName, BindingFlags.NonPublic | BindingFlags.Instance)!;
    }

```
**Expected behavior**
The metadata is already available before the application can receive the requests or it re-uses the recently fetched metadata once subsequent requests get the lock.

**Actual behavior**
All requests waiting for the lock re-retrieve the token's metadata/signing-keys regardless if they were fetched recently enough

**Possible solution**
i haven't given a lot of thought to how this system works, so i reckon my workaround isn't ideal, but i think it's enough to inspire the right kind of change?
 
Here you can see the cache size growing for requests when it should just be a single one:
image
image
image

brentschmaltz:
Removed last instance of RefreshInterval influencing AutomaticRefreshInterval.

Remove dead code in ConfigurationManager.
There was code inside a null check that would not get executed, it was removed.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -212,7 +212,7 @@ namespace Microsoft.IdentityModel.Protocols
             }
             else
             {
-                if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverIdle, ConfigurationRetrieverRunning) != ConfigurationRetrieverRunning)
+                if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
                 {
                     _ = Task.Run(UpdateCurrentConfiguration, CancellationToken.None);
                 }
@@ -302,8 +302,9 @@ namespace Microsoft.IdentityModel.Protocols
         }
 
         /// <summary>
-        /// Requests that then next call to <see cref="GetConfigurationAsync()"/> obtain new configuration.
-        /// <para>If it is a first force refresh or the last refresh was greater than <see cref="BaseConfigurationManager.RefreshInterval"/> then the next call to <see cref="GetConfigurationAsync()"/> will retrieve new configuration.</para>
+        /// Triggers updating metadata when:
+        /// <para>1. Called the first time.</para>
+        /// <para>2. The time between when this method was called and DateTimeOffset.Now is greater than <see cref="BaseConfigurationManager.RefreshInterval"/>.</para>
         /// <para>If <see cref="BaseConfigurationManager.RefreshInterval"/> == <see cref="TimeSpan.MaxValue"/> then this method does nothing.</para>
         /// </summary>
         public override void RequestRefresh()
@@ -313,10 +314,10 @@ namespace Microsoft.IdentityModel.Protocols
             if (now >= DateTimeUtil.Add(_lastRequestRefresh.UtcDateTime, RefreshInterval) || _isFirstRefreshRequest)
             {
                 _isFirstRefreshRequest = false;
-                _lastRequestRefresh = now;
-                if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverIdle, ConfigurationRetrieverRunning) != ConfigurationRetrieverRunning)
+                if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
                 {
                     _ = Task.Run(UpdateCurrentConfiguration, CancellationToken.None);
+                    _lastRequestRefresh = now;
                 }
             }
         }
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -190,7 +190,7 @@ namespace Microsoft.IdentityModel.Protocols
                                         result.ErrorMessage)));
                     }
 
-                    ValidConfigurationObtained(configuration);
+                    UpdateConfiguration(configuration);
                 }
                 catch (Exception ex)
                 {
@@ -249,7 +249,7 @@ namespace Microsoft.IdentityModel.Protocols
 
                 if (_configValidator == null)
                 {
-                    ValidConfigurationObtained(configuration);
+                    UpdateConfiguration(configuration);
                 }
                 else
                 {
@@ -262,7 +262,7 @@ namespace Microsoft.IdentityModel.Protocols
                                     LogMessages.IDX20810,
                                     result.ErrorMessage)));
                     else
-                        ValidConfigurationObtained(configuration);
+                        UpdateConfiguration(configuration);
                 }
             }
             catch (Exception ex)
@@ -282,7 +282,7 @@ namespace Microsoft.IdentityModel.Protocols
 #pragma warning restore CA1031 // Do not catch general exception types
         }
 
-        private void ValidConfigurationObtained(T configuration)
+        private void UpdateConfiguration(T configuration)
         {
             _currentConfiguration = configuration;
             _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval +
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -27,7 +27,6 @@ namespace Microsoft.IdentityModel.Protocols
         private readonly IConfigurationRetriever<T> _configRetriever;
         private readonly IConfigurationValidator<T> _configValidator;
         private T _currentConfiguration;
-        private TimeSpan _bootstrapRefreshInterval = TimeSpan.FromSeconds(1);
 
         // task states are used to ensure the call to 'update config' (UpdateCurrentConfiguration) is a singleton. Uses Interlocked.CompareExchange.
         // metadata is not being obtained
@@ -169,6 +168,7 @@ namespace Microsoft.IdentityModel.Protocols
                     return _currentConfiguration;
                 }
 
+#pragma warning disable CA1031 // Do not catch general exception types
                 try
                 {
                     // Don't use the individual CT here, this is a shared operation that shouldn't be affected by an individual's cancellation.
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -200,51 +200,19 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
                 {
                     fetchMetadataFailure = ex;
 
-                    // In this case configuration was never obtained.
```
pmaytak, line 203: As I understand, the logic in that catch is that
- if we failed to retrieve the initial config, then we set _syncAfter to boostrapRefreshInterval.
  - then we throw an exception that we failed to retrieve config.
  - _bootstrapRefreshInterval is used because is much shorter (1 second) than our normal refresh (5 min).
- If we retrieved the initial config but failed validating it
   - then we set _syncAfter to our normal refresh interval
   - and throw an exception IDX20806 that we failed to update the config.

I do agree that it seems like that code, and bootstrapRefreshInterval, is not needed because if we fail to retrieve the initial config, _currentConfiguration will still be null. So that the following call to GetConfigurationAsync will try to retrieve the config again, ignoring the _syncAfter value.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -200,51 +200,19 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
-                                LogHelper.FormatInvariant(
-                                    LogMessages.IDX20806,
-                                    LogHelper.MarkAsNonPII(MetadataAddress ?? "null"),
-                                    LogHelper.MarkAsNonPII(ex)),
-                                ex));
-                    }
+                    LogHelper.LogExceptionMessage(
+                        new InvalidOperationException(
+                             LogHelper.FormatInvariant(
+                                LogMessages.IDX20806,
```
pmaytak, line 202: We used to throw IDX20803 and IDX20806, wonder if this is behavioral breaking change?

Maybe we should throw IDX20803? It's more general.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/7f90e3475e483b7fc1d1d19c658138b48d05de66/src/Microsoft.IdentityModel.Protocols/LogMessages.cs#L24-L27

pmaytak:
The test is failing
```text
  Failed Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests.ConfigurationManagerTests.BootstrapRefreshIntervalTest [16 ms]
  Error Message:
   Microsoft.IdentityModel.TestUtils.TestException : Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests.ConfigurationManagerTests.BootstrapRefreshIntervalTest

ConfigurationManager._syncAfter: '10/4/2024 9:15:42 AM +00:00' should equal '1/1/0001 12:00:00 AM +00:00'.

  Stack Trace:
     at Microsoft.IdentityModel.TestUtils.TestUtilities.AssertFailIfErrors(String testId, List`1 errors) in D:\a\_work\1\s\test\Microsoft.IdentityModel.TestUtils\TestUtilities.cs:line 371
   at Microsoft.IdentityModel.TestUtils.TestUtilities.AssertFailIfErrors(CompareContext context) in D:\a\_work\1\s\test\Microsoft.IdentityModel.TestUtils\TestUtilities.cs:line 352
   at Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests.ConfigurationManagerTests.<BootstrapRefreshIntervalTest>d__7.MoveNext() in D:\a\_work\1\s\test\Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests\ConfigurationManagerTests.cs:line 262
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
brentschmaltz: @pmaytak if we failed to obtain configuration, we will still throw an exception with IDX20803, right at the end.
What i am thinking is that if we never received a valid configuration, why should we change any values?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
brentschmaltz: @pmaytak  i know we logged 20806, but i don't see where we threw it.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -313,12 +277,18 @@ private void UpdateCurrentConfiguration()
             }
             finally
             {
-                _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval);
                 Interlocked.Exchange(ref _configurationRetrieverState, ConfigurationRetrieverIdle);
             }
 #pragma warning restore CA1031 // Do not catch general exception types
         }
 
+        private void ValidConfigurationObtained(T configuration)
```
msbw2: This name seems off. Maybe something like `StoreValidConfiguration`?
```suggestion
        private void StoreValidConfiguration(T configuration)
```

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
pmaytak: You are right.

nit: Do we need to log IDX20806 here? IDX20803 is thrown at the end of this method and logs the exact same exception (`fetchMetadataFailure`).

pmaytak (approved):

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
pmaytak, line 1: nit: The `RequestRefresh` method comment is inaccurate. It inaccurately says that it makes the next call to GetConfigAsync retrieve config, but it actually can trigger the config update immidiately.

pmaytak, line 1: nit: Left a comment in the previous PR to fix CompareExchange logic to make sure multiple config update requests don't trigger if an update is already running.

I was able to repro this just by calling `configManager.RequestRefresh` twice in a row. The second call starts another update, even though the first one is running. Switching the parameters in the CompareExchange fixes this - only the first call triggers the update.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2865#discussion_r1786800851

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
brentschmaltz: @pmaytak We log because in case some edge case _currentConfiguration was set before we return..

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
pmaytak: This also applies to calling GetConfigurationAsync multiple times when config exists.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -313,12 +277,18 @@ private void UpdateCurrentConfiguration()
...
                 Interlocked.Exchange(ref _configurationRetrieverState, ConfigurationRetrieverIdle);
             }
 #pragma warning restore CA1031 // Do not catch general exception types
         }
 
+        private void UpdateConfiguration(T configuration)
+        {
+            _currentConfiguration = configuration;
+            _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval +
+                TimeSpan.FromSeconds(new Random().Next((int)AutomaticRefreshInterval.TotalSeconds / 20)));
```
jennyf19, line 289: curious why `/ 20` ?

jennyf19 (approved):

PR #2873, opened 2024-10-03
Fix Flaky Tests: NameAndRoleClaimDelegates and RoleClaims

### LINKED ISSUE #2841: [Bug] Fix flaky tests: NameAndRoleClaimDelegates and RoleClaims in System.IdentityModel.Tokens.Jwt.Tests.CreateAndValidateTokens
More flaky tests, run with all the tests and they fail, run individually and they pass:

`System.IdentityModel.Tokens.Jwt.Tests.CreateAndValidateTokens.NameAndRoleClaimDelegates`

AND:
`System.IdentityModel.Tokens.Jwt.Tests.CreateAndValidateTokens.RoleClaims`
   Source: CreateAndValidateTokens.cs line 1081
   Duration: 5 ms

  Message: 
Assert.True() Failure
Expected: True
Actual:   False

  Stack Trace: 
CreateAndValidateTokens.CheckForRoles(IEnumerable`1 expectedRoles, IEnumerable`1 unexpectedRoles, ClaimsPrincipal principal, String expectedRoleClaimType) line 1250
CreateAndValidateTokens.RoleClaims() line 1098

kllysng:
See explanation [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2841#issuecomment-2392462499).

Fixes #2841


msbw2 (approved):

westin-m (approved):

jennyf19 (commented):
Thanks for picking this up!

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -112,6 +112,10 @@ public bool MapInboundClaims
                 if (!_mapInboundClaims && value && _inboundClaimTypeMap.Count == 0)
                     _inboundClaimTypeMap = new Dictionary<string, string>(DefaultInboundClaimTypeMap);
 
+                // If the inbound claim type mapping is on and being turned off, make sure that the _inboundClaimTypeMap is empty.
+                if (_mapInboundClaims && !value && _inboundClaimTypeMap.Count > 0)
+                    _inboundClaimTypeMap.Clear();
```
kllysng: This is logically what I believe should happen - but is there a reason why we didn't have this before?

pmaytak: Quick thought: `_mapInboundClaims` flag should be checked if it's true before map is used. If it's false, the map should never be used. So this should not be needed.

pmaytak: Do we have the same behavior in JsonWebTokenHandler?

pmaytak (dismissed):
This may be a breaking change.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
pmaytak: The test tests that the constructor of JwtSecurityTokenHandler sets up the map or not depending if the static flag DefaultMapInboundClaims was set or not; then it also checks if setting MapInboundClaims property initializes the map list.

Both MapInboundClaims and InboundClaimTypeMap are public. In our code we check the flag first before using the map. There are a couple of usage scenarios.

1. MapInboundClaims is false and InboundClaimTypeMap is empty. Set flag to true. This will initialize the map to the default map values, which makes sense since we don't know if the map will be explicitly initialized, so we set it to default. The map can still be overwritten to non-default values.
2. InboundClaimTypeMap is set to custom values. MapInboundClaims is set to true. This will not reset the map to default values because it already as values.
3. InboundClaimTypeMap is set to custom values. MapInboundClaims is set to true, then set to false. With the previous code, the map will not be cleared and when the flag is set to true, custom values will remain. With updated code, the custom values will be cleared, and when the flag is set to true, map will be set to default values. This is the possible breaking change. Probably unlikely but still.

So I would only update the test to reset DefaultMapInboundClaims to true (default) and move it and add a collection attribute to it.

kllysng: These are good points - I appreciate the thoughtful write up and reflection. I've reverted the src code changes and reset the property in the test so that the other tests aren't affected - running this test and the other two previously flaky tests "to failure" for 5 minutes didn't incur any failures whereas previously they would immediately fail.

kllysng:
> This may be a breaking change.

Fixed!

msbw2 (approved):

jennyf19 (approved):

PR #2879, opened 2024-10-07
return CaseSensitiveClaimsIdentity in clone

jennyf19:
will create an issue shortly.
not getting test failures locally:
image


jmprieur (approved):
LGTM
But this does not build (probably the extra methods in the public API)

jennyf19:
> LGTM But this does not build (probably the extra methods in the public API)

yes, waiting for Brent's PR

pmaytak (approved):

PR #2894, opened 2024-10-10
Fix typo

### LINKED ISSUE #1873: [Bug] Microsoft.IdentityModel.Xml.XmlUtil.GetXsiTypeAsQualifiedName searches 'Type' attribute instead of 'type'
Microsoft.IdentityModel.Tokens.Saml 6.18

**Where is the issue?**
 * [x] M.IM.Tokens.Saml

New app

**Repro**
Saml2.xml
```xml
<saml:Assertion xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" ID="_e9c223c5-1792-4a85-9a44-938e0dcb017d" IssueInstant="2022-06-10T09:59:33.679Z" Version="2.0">
	<saml:Issuer>FooBar</saml:Issuer>
	<saml:Subject>
		<saml:NameID>UserKey</saml:NameID>
		<saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer"></saml:SubjectConfirmation>
	</saml:Subject>
	<saml:Conditions NotBefore="2022-06-10T09:59:33.665Z" NotOnOrAfter="2022-06-10T10:59:33.669Z">
		<saml:Condition xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="saml:OneTimeUseType" />
	</saml:Conditions>
</saml:Assertion>
```

```csharp
var tokenString = File.ReadAllText(@"Saml2.xml");
var handler = new Microsoft.IdentityModel.Tokens.Saml2.Saml2SecurityTokenHandler();
var token = handler.ReadSaml2Token(tokenString);
```

**Expected behavior**
Token should be read successfully.

**Actual behavior**
Fails with following exception:
Microsoft.IdentityModel.Tokens.Saml2.Saml2SecurityTokenReadException
  HResult=0x80131500
  Message=IDX13119: An abstract element was encountered which does not specify its concrete type. Element name: '[PII of type 'System.String' is hidden. For more details, see https://aka.ms/IdentityModel/PII.]' Namespace: '[PII of type 'System.String' is hidden. For more details, see https://aka.ms/IdentityModel/PII.]'
  Source=Microsoft.IdentityModel.Tokens.Saml
  StackTrace:
   at Microsoft.IdentityModel.Tokens.Saml2.Saml2Serializer.ReadConditions(XmlDictionaryReader reader) in /_/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs:line 879

**Possible solution**
Fix typo in XmlSignatureConstants.Attributes.Type.

westin-m:
Closes #1873 

Tested with repro in bug


msbw2 (approved):

jennyf19 (approved):
:shipit:

jennyf19:
wonder if this needs to be in a major version - thoughts @brentschmaltz ?

pmaytak:
> wonder if this needs to be in a major version - thoughts @brentschmaltz ?

Does it work currently at all? Like `xsi:Type`? If yes, I suppose this could be a breaking change, since folks rely on this behavior.

westin-m:
> Does it work currently at all? Like `xsi:Type`? If yes, I suppose this could be a breaking change, since folks rely on this behavior.

This works because `xsi:type` is what's expected. It's a runtime change, which I think should be fine to bring in.

PR #2902, opened 2024-10-14
Suggested changes to AudienceValidation

brentschmaltz:
Changes to AudienceValidation to help with consistency:
1. Cache stack frames in AudienceValidationError to remove possibilitiy of use by other validators.
2. Pass tokenAudience and validateAudiences to AudienceValidationError for increased fidelity.
3. Attempt to resolve Exception in derived ValidationError first before calling base.
4. Add ValidationFailureTypes to describe errors with increased fidelity.
5. Use MessageDetail.NullParameter consistently for null parameter errors.

Some smaller items:
JsonWebTokenHandler set error to ValidationFailureType.SignatureAlgorithmValidationFailed, when algorithm is not valid.
LifetimeValidationError remove ctor that has inner validation error
Renamed IssuerValidationSource values to be more readable

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
@@ -245,10 +245,9 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     new MessageDetail(
                         TokenLogMessages.IDX10518,
                         result.UnwrapError().MessageDetail.Message),
-                    ValidationFailureType.SignatureValidationFailed,
-                    typeof(SecurityTokenInvalidSignatureException),
-                    new StackFrame(true),
-                    result.UnwrapError());
+                    ValidationFailureType.SignatureAlgorithmValidationFailed,
+                    typeof(SecurityTokenInvalidAlgorithmException),
+                    new StackFrame(true));
 
             SignatureProvider signatureProvider = cryptoProviderFactory.CreateForVerifying(key, jsonWebToken.Alg);
             try
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
@@ -39,11 +39,7 @@ namespace Microsoft.IdentityModel.Tokens
         /// <returns>An instance of an Exception.</returns>
         internal override Exception GetException()
         {
-            if (ExceptionType == typeof(SecurityTokenArgumentNullException))
-            {
-                return new SecurityTokenArgumentNullException(MessageDetail.Message);
-            }
-            else if (ExceptionType == typeof(SecurityTokenInvalidAudienceException))
+            if (ExceptionType == typeof(SecurityTokenInvalidAudienceException))
                 return new SecurityTokenInvalidAudienceException(MessageDetail.Message) { InvalidAudience = Utility.SerializeAsSingleCommaDelimitedString(_tokenAudiences) };
 
             return base.GetException(ExceptionType, null);
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
@@ -10,31 +10,43 @@ namespace Microsoft.IdentityModel.Tokens
 {
     internal class AudienceValidationError : ValidationError
     {
-        private IList<string>? _invalidAudiences;
+        private IList<string>? _tokenAudiences;
+        private IList<string>? _validAudiences;
+
+        // stack frames associated with AudienceValidationErrors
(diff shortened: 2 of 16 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -245,10 +245,9 @@ private static ValidationResult<SecurityKey> ValidateSignatureWithKey(
                     new MessageDetail(
                         TokenLogMessages.IDX10518,
                         result.UnwrapError().MessageDetail.Message),
-                    ValidationFailureType.SignatureValidationFailed,
-                    typeof(SecurityTokenInvalidSignatureException),
-                    new StackFrame(true),
-                    result.UnwrapError());
+                    ValidationFailureType.SignatureAlgorithmValidationFailed,
```
iNinja, line 248: I am not comfortable with this change. We are modifying the behaviour of the method here by changing the exception type being returned.
The original path returns a `SecurityTokenInvalidSignatureException` with the `SecurityTokenInvalidAlgorithmException` as the `innerException`, and that is what this code was doing.

I believe we agreed that we should respect the existing exception types being returned and that we would only modify the error messages being returned.

`ValidateJWSAsync` has a special behaviour for "recoverable exception types" which as of now are `SecurityTokenInvalidSignatureException`, `SecurityTokenInvalidIssuerException`, and `SecurityTokenSignatureKeyNotFoundException`.

Could you please elaborate on the reason for this change?

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
```diff
@@ -10,31 +10,43 @@ namespace Microsoft.IdentityModel.Tokens
...
         /// Creates an instance of an <see cref="Exception"/> using <see cref="ValidationError"/>
         /// </summary>
         /// <returns>An instance of an Exception.</returns>
-        public override Exception GetException()
+        internal override Exception GetException()
         {
-            return new SecurityTokenInvalidAudienceException(MessageDetail.Message) { InvalidAudience = Utility.SerializeAsSingleCommaDelimitedString(_invalidAudiences) };
+            if (ExceptionType == typeof(SecurityTokenArgumentNullException))
+            {
+                return new SecurityTokenArgumentNullException(MessageDetail.Message);
```
iNinja: This is a common case, would be best handled in the base method instead of repeating it in every `ValidationError` subclass.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerValidationParametersTests.cs
```diff
@@ -191,9 +192,7 @@ public static TheoryData<JsonWebTokenHandlerValidationParametersTheoryData> Json
                         SigningCredentials = KeyingMaterial.DefaultSymmetricSigningCreds_256_Sha2,
                         ExpectedIsValid = false,
                         ExpectedException = ExpectedException.SecurityTokenInvalidSignatureException("IDX10511:"),
-                        ExpectedExceptionValidationParameters = ExpectedException.SecurityTokenInvalidSignatureException(
-                            "IDX10518:",
-                            innerTypeExpected: typeof(SecurityTokenInvalidAlgorithmException))
+                        ExpectedExceptionValidationParameters = ExpectedException.SecurityTokenInvalidAlgorithmException("IDX10518:")
```
iNinja, line 195: Mentioned in another comment. We are returning a different exception type here than the original path.

iNinja (changes requested):
For the most part this is great. There is one point that I can't agree with without having a proper discussion with @jmprieur and @jennyf19, mentioned in the review.

iNinja (approved):
LGTM

FuPingFranco (approved):
LGTM, Thanks Brent!

jennyf19:
> For the most part this is great. There is one point that I can't agree with without having a proper discussion with @jmprieur and @jennyf19, mentioned in the review.

which is that?

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Issuer.cs
```diff
@@ -9,11 +9,12 @@
 #nullable enable
 namespace Microsoft.IdentityModel.Tokens
 {
+    // TODO how do we extend this?
```
FuPingFranco, line 12: Can we create a follow up issue to track this?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
iNinja: This is the comment, @jennyf19.

We discussed it, and accepted the explanation that the intention here is to return more specific errors, and not consider the invalid algorithm in the signature as a recoverable exception/error.

So now the case where the signature is done with an unsupported algorithm will return a `SecurityTokenInvalidAlgorithmException` instead of the original `SecurityTokenInvalidSignatureException` wrapping a `SecurityTokenInvalidAlgorithmException`, which was considered recoverable.

PR #2903, opened 2024-10-14
Mark Wilson APIs as Shipped

westin-m:
Edit `mark-shipped.ps1`  to affect both internal and public APIs. 

Keep some unready APIs as unshipped.


Tools/mark-shipped.ps1
```diff
@@ -4,16 +4,18 @@ param ()
 Set-StrictMode -version 2.0
 $ErrorActionPreference = "Stop"
 
-function MarkShipped([string]$dir) {
-    $shippedFilePath = Join-Path $dir "PublicAPI.Shipped.txt"
+function MarkShipped([string]$dir, [string]$access) {
```
jmprieur: would we want to have default values?

jmprieur (approved):

Tools/mark-shipped.ps1
iNinja: or fail if one is missing, maybe.

jennyf19:
Looks like there are merge conflicts

Tools/mark-shipped.ps1
westin-m: The parameters are now mandatory and function calls will fail without both

FuPingFranco (approved):
I.M.JsonWebToken, I.M.Tokens, I.M.Tokens.Saml and I.M.Validators LGTM.

PR #2904, opened 2024-10-15
Remove SecurityKey from signature validation delegate's result

iNinja:
# Remove SecurityKey from signature validation delegate's result

- Introduced `TokenValidationUnit` type to represent void return types from the generic `ValidationResult` 
- Removed `SecurityKey` from the return value of the signature validation delegate

Notes: The `Unit` type is required in order to return "void" from a generic, since we cannot write `ValidationResult<void>` in C#. This change simplifies the signature validation delegate by removing the `SecurityKey` return type, which is no longer needed.

This work is part of #2911.


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -27,7 +27,7 @@ public partial class JsonWebTokenHandler : TokenHandler
         /// <exception cref="SecurityTokenInvalidSignatureException">Returned by the default implementation if the token is not signed, or if the validation fails.</exception>
         /// <exception cref="SecurityTokenInvalidAlgorithmException">Returned if the algorithm is not supported by the key.</exception>
```
FuPingFranco, line 28: nit: Will you be updating this documentation in this iteration?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -115,7 +115,7 @@ internal static ValidationResult<SecurityKey> ValidateSignature(
             }
         }
 
-        private static ValidationResult<SecurityKey> ValidateSignatureUsingAllKeys(
+        private static ValidationResult<TokenValidationUnit> ValidateSignatureUsingAllKeys(
             JsonWebToken jwtToken,
             ValidationParameters
             validationParameters, BaseConfiguration? configuration,
```
FuPingFranco: nit: can you stack the configuration parameter?

src/Microsoft.IdentityModel.Tokens/Validation/Results/TokenValidationUnit.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Unit type used to represent an empty value in ValidationResults.
+    /// </summary>
+    /// <remarks>All copies of TokenValidationUnit are considered equal.</remarks>
+    internal record struct TokenValidationUnit
```
FuPingFranco, line 10: Not sure if using the term `TokenValidationUnit` is very descriptive for the intent? Is there other places in Jwt handling we'll be using this record struct?

FuPingFranco (approved):
Nothing blocking but would like to get a second opinion on the naming of `TokenValidationUnit`.

src/Microsoft.IdentityModel.Tokens/Validation/Results/TokenValidationUnit.cs
iNinja: Given the C# language does not implement a Unit type (https://en.wikipedia.org/wiki/Unit_type), we need to resort to creating our own. But at the same time I am trying to be wary and future-proof of simply calling this type `Unit` in case other libraries or even the runtime decide to implement it as well.
If we want to rename it to something else, my only requirement is that we keep the `Unit` in the name to make it clear to anyone familiar with the concept of what it is.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
iNinja: Outside of the scope of this work. 

We need to align on how we want to communicate the exceptions being "returned" as opposed to thrown, since they can be retrieved from the validation error.

src/Microsoft.IdentityModel.Tokens/Validation/Results/TokenValidationUnit.cs
FuPingFranco: Makes sense, I ok with this name for now. We can discuss a name change if others find it too general. Thanks for the explanation.

iNinja:
Dropping this to avoid adding a new type.

PR #2923, opened 2024-10-17
Added properties to ValidationResult without throwing

### LINKED ISSUE #2924: Add properties to retrieve Result and Error from ValidationResult without throwing
Per our conversation with Brent and Franco, we are adding properties to ValidationResult that do not throw, unlike the internal `UnwrapResult` and `UnwrapError`, but return a nullable / default value if called in the wrong situation.

- If `ValidationResult.IsValid` is `true`:
  - The internal `UnwrapResult()` returns a non optional `TResult`.
  - The internal `UnwrapError()` throws an exception
  - `Result` returns an optional `TResult?`.
  - `Error` returns `null`.
- If `ValidationResult.IsValid` is `false`:
  - The internal `UnwrapResult()` throws an exception.
  - The internal `UnwrapError()` returns a non optional `ValidationError`.
  - `Result` returns `null` or `default` if it is a non nullable type (value type)
  - `Error` returns an optional `ValidationError?`

iNinja:
# Added properties to ValidationResult without throwing

- Added new nullable properties to `ValidationResult` without throwing exceptions.
- Renamed IsSuccess to IsValid

Part of #2711

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidationResult.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidationResult.cs
@@ -2,12 +2,13 @@
 // Licensed under the MIT License.
 
 using System;
+using Microsoft.IdentityModel.Logging;
 
 #nullable enable
 namespace Microsoft.IdentityModel.Tokens
 {
     /// <summary>
-    /// Represents a validation result that can be either successful or unsuccessful.
+    /// Represents a validation result that can be either valid or invalid.
     /// </summary>
     /// <typeparam name="TResult"></typeparam>
     internal readonly struct ValidationResult<TResult> : IEquatable<ValidationResult<TResult>>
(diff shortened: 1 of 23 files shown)


jennyf19 (approved):

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidationResult.cs
```diff
@@ -58,25 +59,60 @@ public ValidationResult(ValidationError error)
...
+        /// Gets the error associated with the validation result.
+        /// </summary>
+        /// <returns>The error associated with the validation result.</returns>
+        /// <remarks>This property is only valid if the result type is not valid.</remarks>
+        public ValidationError? Error
+        {
+            get
+            {
+                if (IsValid)
+                    LogHelper.LogWarning("Warning: Accessing the Error property in a valid result is invalid.");
```
FuPingFranco, line 92: nit: I wonder if we should have this as an IDX message?

FuPingFranco (approved):
LGTM

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidationResult.cs
iNinja: I'll raise an issue to review the messages and look into that as well.

PR #2927, opened 2024-10-18
Regression tests: Issuer signing key

### LINKED ISSUE #2911: Add comparison tests for IssuerSigningKey validation in new validation model

- Add tests comparing `ValidateTokenAsync` in the new path against the existing one for IssuerSigningKey scenarios.

Part of #2711

iNinja:
# Regression tests: Issuer signing key
- Added internal TimeProvider to `TokenValidationParameters` to simplify testing of expired signing key scenarios
- Added regression/comparison tests for issuer signing key scenarios

Part of #2711

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
@@ -341,7 +341,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
 
             ValidationResult<ValidatedSigningKeyLifetime> issuerSigningKeyValidationResult =
                 validationParameters.IssuerSigningKeyValidator(
-                    signatureValidationResult.UnwrapResult(), jsonWebToken, validationParameters, configuration, callContext);
+                    jsonWebToken.SigningKey, jsonWebToken, validationParameters, configuration, callContext);
             if (!issuerSigningKeyValidationResult.IsSuccess)
             {
                 StackFrame issuerSigningKeyValidationFailureStackFrame = StackFrames.IssuerSigningKeyValidationFailed ??= new StackFrame(true);
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
@@ -541,6 +541,11 @@ namespace Microsoft.IdentityModel.Tokens
         public SignatureValidatorUsingConfiguration SignatureValidatorUsingConfiguration { get; set; }
 
         /// <summary>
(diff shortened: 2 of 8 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
```diff
@@ -341,7 +341,7 @@ await ValidateJWSAsync(actorToken, actorParameters, configuration, callContext,
 
             ValidationResult<ValidatedSigningKeyLifetime> issuerSigningKeyValidationResult =
                 validationParameters.IssuerSigningKeyValidator(
-                    signatureValidationResult.UnwrapResult(), jsonWebToken, validationParameters, configuration, callContext);
+                    jsonWebToken.SigningKey, jsonWebToken, validationParameters, configuration, callContext);
```
jennyf19, line 344: is this the right return type?

src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
```diff
@@ -540,6 +540,11 @@ public string RoleClaimType
         /// </remarks>
         public SignatureValidatorUsingConfiguration SignatureValidatorUsingConfiguration { get; set; }
 
+        /// <summary>
+        /// Gets or sets the time provider.
+        /// </summary>
+        internal TimeProvider TimeProvider { get; set; } = TimeProvider.System;
```
jennyf19, line 546: do we need to do this in more places for consistency? just curious how it could relate to this issue: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2572

iNinja: I have updated the places where we used it with ValidationParameters.

I believe that if we make this change public, it will address #2572.
For now, I am only using it in my tests and all other scenarios behave the same as without it.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
iNinja: The `ValidationResult<ValidatedSigningKeyLifetime>` is unchanged, and it is what we are currently returning.

The change here is to take the `SecurityKey` from the `JsonWebToken` instead of from the signature validation result, as we did in the initial implementation.

iNinja: Using the key from the `JsonWebToken` is in line with the original behaviour in the now legacy path.

FuPingFranco (approved):
LGTM

kllysng (approved):

PR #2931, opened 2024-10-23
Regression tests: Token Replay

### LINKED ISSUE #2915: Add comparison tests for Token Replay validation in new validation model

- Add tests comparing `ValidateTokenAsync` in the new path against the existing one for TokenReplay scenarios.

Part of #2711

iNinja:
# Regression tests: Token Replay
- Removed calls to `MarkAsUnsafeSecurityArtifact` in the new path as they were printing the tokens in clear text and we cannot use `JwtTokenUtilities.SafeLogJwt` due to the fact that this handles other `SecurityToken`s as well. This way the logger still masks the token as a `SecurityArtifact`.
- Aligned returned exception when token has no expiration and token cache is set to match legacy path.
- Added regression/comparison tests for token replay scenarios

Resolves #2915 

Part of #2711


src/Microsoft.IdentityModel.Tokens/Validation/Validators.TokenReplay.cs
```diff
@@ -60,16 +59,16 @@ public static partial class Validators
                     return new ValidationError(
                         new MessageDetail(
                             LogMessages.IDX10227,
-                            LogHelper.MarkAsUnsafeSecurityArtifact(securityToken, t => t.ToString())),
+                            securityToken),
                         ValidationFailureType.TokenReplayValidationFailed,
-                        typeof(SecurityTokenReplayDetectedException),
+                        typeof(SecurityTokenNoExpirationException),
```
kllysng, line 64: Is changing the exception type a breaking change?

iNinja: This code path is not yet public. The change simply aligns the new path with the existing one which returns a `SecurityTokenNoExpirationException` in this scenario.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.Extensibility.cs
```diff
@@ -19,7 +19,7 @@ public async Task ValidateTokenAsync_Lifetime_Extensibility(ValidateTokenAsyncLi
         {
             var context = TestUtilities.WriteHeader($"{this}.{nameof(ValidateTokenAsync_Lifetime_Extensibility)}", theoryData);
 
-            string jwtString = CreateToken(theoryData.IssuedAt, theoryData.NotBefore, theoryData.Expires);
+            string jwtString = CreateTokenWithLifetime(theoryData.IssuedAt, theoryData.NotBefore, theoryData.Expires);
```
kllysng, line 22: May I ask why we removed CreateToken and replaced with this one instead? I'm trying to think about keeping our test classes localized and as simple as possible. 🤔

src/Microsoft.IdentityModel.Tokens/Validation/Validators.TokenReplay.cs
kllysng: Thanks for clarifying! Resolving this comment

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.Extensibility.cs
iNinja: This may be a bit tricky when looking at one PR in isolation, but `JsonWebTokenHandler.ValidateTokenAsyncTests` is a class spread out over many partials in different files.
Because the private methods are still part of the overall class, the names of the multiple `CreateToken` methods we have across files collide. The one that you see removed in this file is because I accidentally created two methods that did the exact same thing in the Lifetime comparison and Lifetime extensibility tests, so I removed one of the copies and called the other, for simplicity.

westin-m (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.Extensibility.cs
jennyf19: great question @kellyyangsong, thanks for the response @iNinja

jennyf19 (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Lifetime.Extensibility.cs
kllysng: @iNinja thank you for taking the time to explain!

PR #2932, opened 2024-10-23
Regression tests: Token Type

### LINKED ISSUE #2914: Add comparison tests for Token Type validation in new validation model

- Add tests comparing `ValidateTokenAsync` in the new path against the existing one for Token Type scenarios.

Part of #2711

iNinja:
# Regression tests: Token Type
- Added TokenTypeValidationError to create the exception and assign the invalid token type.
- Renamed ValidationParameters.TypeValidator to ValidationParameters.TokenTypeValidator
- Adjusted the creation of SecurityTokenInvalidTypeException to return the correct invalid type when it is empty or null
- Added regression/comparison tests for token type validation scenarios

Resolves #2914 

Part of #2711


src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -499,7 +499,7 @@ public TokenReplayValidationDelegate TokenReplayValidator
         /// </remarks>
         /// <exception cref="ArgumentNullException">Thrown when the value is set as null.</exception>
         /// <returns>The <see cref="TokenTypeValidationDelegate"/> used to validate the token type of a token</returns>
-        public TokenTypeValidationDelegate TypeValidator
+        public TokenTypeValidationDelegate TokenTypeValidator
```
kllysng, line 502: Is this a breaking change?

kllysng (changes requested):
Please see my comment regarding the breaking change in ValidationParameters

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
iNinja: The type `ValidationParameters` is internal and not used anywhere outside of our new `ValidateTokenAsync`.
So I would say not a public interface change, even if the property is marked as public.

kllysng: That's right, thanks Ignacio. Resolving this thread!

kllysng (approved):
Thank you @iNinja

westin-m (approved):

PR #2934, opened 2024-10-23
Regression tests: Algorithm

### LINKED ISSUE #2913: Add comparison tests for Algorithm validation in new validation model

- Add tests comparing `ValidateTokenAsync` in the new path against the existing one for Algorithm validation scenarios.

Part of #2711

iNinja:
# Regression tests: Algorithm
- Added AlgorithmValidationError
- Refactored error creation in ValidateSignature to surface the invalid algorithm as part of the error
- Added regression/comparison tests for algorithm validation scenarios

Resolves #2913 

Part of #2711


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -241,13 +241,29 @@ private static ValidationResult<SecurityKey> ValidateSignatureWithKey(
...
+                        new MessageDetail(
+                            TokenLogMessages.IDX10518,
+                            algorithmValidationError.MessageDetail.Message),
+                        typeof(SecurityTokenInvalidAlgorithmException),
+                        new StackFrame(true),
+                        algorithmValidationError.InvalidAlgorithm);
+                }
+                else
+                {
+                    // overridden delegate did not return an AlgorithmValidationError
```
kllysng, line 257: Tests: do you plan on adding test coverage for this else block?

iNinja: We would have look at this in our extensibility scenarios. We are adding specific tests for the cases where the delegates are implemented by users.

kllysng: Thank you, resolving this!

kllysng (approved):
Thanks, @iNinja

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Algorithm.cs
```diff
@@ -0,0 +1,134 @@
...
+                theoryData.Add(new ValidateTokenAsyncAlgorithmTheoryData("Valid_ValidAlgorithmsIsEmptyList")
+                {
+                    SigningCredentials = KeyingMaterial.DefaultX509SigningCreds_2048_RsaSha2_Sha2,
+                    TokenValidationParameters = CreateTokenValidationParameters(
+                        KeyingMaterial.DefaultX509SigningCreds_2048_RsaSha2_Sha2.Key, validAlgorithms: []),
+                    ValidationParameters = CreateValidationParameters(
+                        KeyingMaterial.DefaultX509SigningCreds_2048_RsaSha2_Sha2.Key, validAlgorithms: []),
+                });
+
+                theoryData.Add(new ValidateTokenAsyncAlgorithmTheoryData("Invalid_TokenIsSignedWithAnInvalidAlgorithm")
```
westin-m, line 64: **Question**: Is this the only negative test case that's needed?

jennyf19 (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Algorithm.cs
iNinja: Yes. The algorithm validation fails **only** when **all** the following conditions are met:
- ValidAlgorithms is not null
- ValidAlgorithms is not empty
- The algorithm used is **not** present in the ValidAlgorithms list

Any other scenario is considered valid. I intentionally added tests highlighting those to help explain this.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -241,13 +241,29 @@ private static ValidationResult<SecurityKey> ValidateSignatureWithKey(
...
             if (!result.IsValid)
-                return new ValidationError(
-                    new MessageDetail(
-                        TokenLogMessages.IDX10518,
-                        result.UnwrapError().MessageDetail.Message),
-                    ValidationFailureType.SignatureAlgorithmValidationFailed,
-                    typeof(SecurityTokenInvalidAlgorithmException),
-                    new StackFrame(true));
+            {
+                if (result.UnwrapError() is AlgorithmValidationError algorithmValidationError)
```
brentschmaltz, line 245: Curious as to why we need to recreate the AlgorithmValidationError?

PR #2948, opened 2024-10-29
SAML new model validation: Issuer

### LINKED ISSUE #2942: Implement and test issuer validation in SAMLTokenHandler using the new token validation model
Following the implementation in the SAML2TokenHandler, we need to add it to the other SAML token handler and add tests for it.

Part of #2711

iNinja:
# SAML new model validation: Issuer
- Implemented issuer validation for SAML tokens using the new validation model

Resolves #2942 

Part of #2711


src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ValidateToken.Internal.cs
```diff
@@ -48,6 +48,19 @@ internal async Task<ValidationResult<ValidatedToken>> ValidateTokenAsync(
                 return conditionsResult.UnwrapError().AddStackFrame(StackFrames.AssertionConditionsValidationFailed);
             }
 
+            var issuerValidationResult = await validationParameters.IssuerValidatorAsync(
```
kllysng, line 51: This is a not yet released class right? So this *isn't* a behavorial breaking change?

iNinja: That is correct. This is all internal and work in progress. There's a few more PRs coming adding functionality to this method.

kllysng: Thank you, makes sense

kllysng (approved):

jennyf19 (approved):
:shipit:

PR #2951, opened 2024-10-29
Restore EnsureAotCompatibility test

kllysng:
The test is passing locally without issue. Running an ADO PR build to ensure it passes there too.

Update - test passed in ADO build with no issues, running again

Update again - passed again

### DIFF
--- a/test/Microsoft.IdentityModel.AotCompatibility.Tests/AotCompatibilityTests.cs
+++ b/test/Microsoft.IdentityModel.AotCompatibility.Tests/AotCompatibilityTests.cs
@@ -32,7 +32,7 @@ namespace Microsoft.IdentityModel.AotCompatibility.Tests
         ///
         /// You can also 'dotnet publish' the 'Microsoft.IdentityModel.AotCompatibility.TestApp.csproj' as well to get the errors.
         /// </summary>
-        [Fact(Skip = "need to adjust timeout")]
+        [Fact]
         public void EnsureAotCompatibility()
         {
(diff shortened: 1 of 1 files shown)


jennyf19 (approved):

msbw2:
I think this might be a build step. At least it would make sense to be one to me instead of a unit test.

keegan-caruso:
> I think this might be a build step. At least it would make sense to be one to me instead of a unit test.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/90dfd314cf39f40aae8f23a51155762d91c3416c/build/common.props#L78

jmprieur (approved):
LGTM
Thanks @kellyyangsong 

Do we understand why it was skipped in the first place?

kllysng:
> LGTM Thanks @kellyyangsong
> 
> Do we understand why it was skipped in the first place?

@jmprieur yes, thanks for asking, I should have explained in a comment. It was skipped because it would timeout which for this test is taking longer than 3 min in the ADO build. I re-ran this many times and never saw it take longer than 50s to complete.

kllysng:
> I think this might be a build step. At least it would make sense to be one to me instead of a unit test.


> > I think this might be a build step. At least it would make sense to be one to me instead of a unit test.
> 
> https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/90dfd314cf39f40aae8f23a51155762d91c3416c/build/common.props#L78

We have an analyzer that will cause warnings - but I still think this test is useful until we start treating warnings as errors.

PR #2957, opened 2024-10-31
Add additional locks for reading and writing config

msbw2:
Add additional locks for reading and writing config.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -22,6 +22,7 @@ namespace Microsoft.IdentityModel.Protocols
         private DateTimeOffset _lastRequestRefresh = DateTimeOffset.MinValue;
         private bool _isFirstRefreshRequest = true;
         private readonly SemaphoreSlim _configurationNullLock = new SemaphoreSlim(1);
+        private readonly ReaderWriterLockSlim _syncLock = new();
 
         private readonly IDocumentRetriever _docRetriever;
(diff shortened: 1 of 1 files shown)


msbw2:
`DateTimeOffset` is not atomic on read/write, so `_syncAfter` needs a read/write lock. `_currentConfiguration` is a reference type and therefore does not need the lock since reads and writes for reference types are atomic.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -147,8 +148,12 @@ public async Task<T> GetConfigurationAsync()
         /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
         public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
         {
-            if (_currentConfiguration != null && _syncAfter > DateTimeOffset.UtcNow)
-                return _currentConfiguration;
+            _syncLock.EnterReadLock();
```
brentschmaltz, line 151: We don't want to lock here as that means every token validation will lock.

I am working on a solution that is in this branch: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/tree/brentsch/SyncAfter

msbw2: FYI `double` is not guaranteed to be atomic for reads and writes.

PR #2961, opened 2024-11-01
SAML2 new model validation: Signature

### LINKED ISSUE #2950: Implement and test signature validation in SAML2TokenHandler using the new token validation model
As part of the new validation model implementation for SAML2 tokens, we need to validate the signature using ValidationParameters.

Part of https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2711

iNinja:
# SAML2 new model validation: Signature
- Implemented signature validation for SAML2 tokens using the new validation model

Note: This re-uses the work introduced for SAML in a previous PR and should not be merged before that one.

Resolves #2950 

Part of #2711


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+                return ValidationError.NullParameter(
+                    nameof(samlToken),
+                    new StackFrame(true));
+            }
+
+            if (validationParameters is null)
+            {
+                return ValidationError.NullParameter(
+                    nameof(validationParameters),
+                    new StackFrame(true));
```
brentschmaltz: Make sure we use the new model for stack frames here afterwards.

iNinja: I will raise a separate PR to update all the stack frames to use the new model once it is merged.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            // Delegate is set by the user, we call it and return the result.
+            if (validationParameters.SignatureValidator is not null)
+                return validationParameters.SignatureValidator(samlToken, validationParameters, null, callContext);
+
+            // If the user wants to accept unsigned tokens, they must implement the delegate
```
msbw2: I think this could be more clear:
```suggestion
            // If the user wants to accept unsigned tokens, they must set validationParameters.SignatureValidator
```

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+            if (samlToken.Assertion.Signature is null)
```
msbw2, line 39: I think you need to check that `samlToken.Assertion` is not `null`. Otherwise, this results in a null ref.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            if (validationParameters is null)
+            {
+                return ValidationError.NullParameter(
+                    nameof(validationParameters),
+                    new StackFrame(true));
+            }
+
+            // Delegate is set by the user, we call it and return the result.
+            if (validationParameters.SignatureValidator is not null)
```
msbw2, line 35: Do you also need to consider `validationParameters.SignatureValidatorUsingConfiguration`?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                resolvedKey = validationParameters.IssuerSigningKeyResolver(
+                    samlToken.Assertion.CanonicalString,
+                    samlToken,
+                    samlToken.Assertion.Signature.KeyInfo?.Id,
+                    validationParameters,
+                    null,
+                    callContext);
+            }
+            else
```
msbw2, line 61: `validationParameters.IssuerSigningKeyResolverUsingConfiguation`? There are also a few other places where the signing key could be provided and/or validated via `validationParameters`.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            if (resolvedKey is null)
+            {
+                if (validationParameters.TryAllIssuerSigningKeys)
+                    keys = validationParameters.IssuerSigningKeys;
+            }
+            else
+            {
+                keys = [resolvedKey];
```
msbw2: Can we just add a block to handle the single key `resolvedKey` instead of allocating a new list with one item?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+                        validationParameters,
+                        callContext);
+
+                    if (!algorithmValidationResult.IsValid)
+                    {
+                        errors.Add(algorithmValidationResult.UnwrapError());
+                    }
+                    else
+                    {
+                        var validationError = samlToken.Assertion.Signature.Verify(
```
msbw2: Is `Verify` guaranteed to not be `null`?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+
+                        if (validationError is null)
+                        {
+                            samlToken.SigningKey = key;
+
+                            return key;
+                        }
+                        else
+                        {
+                            errors.Add(validationError.AddStackFrame(new StackFrame()));
```
msbw2: Should `true` be passed to the constructor for `StackFrame` here?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+                    keys = validationParameters.IssuerSigningKeys;
+            }
+            else
+            {
+                keys = [resolvedKey];
+                keyMatched = true;
+            }
+
+            bool canMatchKey = samlToken.Assertion.Signature.KeyInfo != null;
+            List<ValidationError> errors = new();
```
msbw2: Can you create this only when necessary? If you need to add to it, `(errors ??= new()).Add(...);` would work.

msbw2: Also, what is the expected number of `ValidationError`s? Besides 0, would 1 be the next most common? If so, we can just store the first and then only when we hit 2 create the list.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+                        }
+                    }
+
+                    keysAttempted.Append(key.ToString());
+                    if (canMatchKey && !keyMatched && key.KeyId is not null && samlToken.Assertion.Signature.KeyInfo is not null)
+                        keyMatched = samlToken.Assertion.Signature.KeyInfo.MatchesKey(key);
+                }
+            }
+
+            if (canMatchKey && keyMatched)
```
msbw2, line 104: Is this right? Shouldn't it be if `!keyMatched`?
```suggestion
            if (canMatchKey && !keyMatched)
```

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+                        TokenLogMessages.IDX10514,
+                        keysAttempted.ToString(),
+                        samlToken.Assertion.Signature.KeyInfo,
+                        GetErrorStrings(errors),
+                        samlToken),
+                    ValidationFailureType.SignatureValidationFailed,
+                    typeof(SecurityTokenInvalidSignatureException),
+                    new StackFrame(true));
+
+            if (keysAttempted.Length > 0)
```
msbw2: Why is it an error if multiple keys are tried?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
msbw2: I think I see now. In the valid case this method should have already returned and thus this wouldn't be reached. Therefore, everything at this point is an error.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
msbw2: Same here.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+            return new XmlValidationError(
+                new MessageDetail(TokenLogMessages.IDX10500),
+                ValidationFailureType.SignatureValidationFailed,
+                typeof(SecurityTokenSignatureKeyNotFoundException),
+                new StackFrame(true));
+        }
+
+        private static string GetErrorStrings(List<ValidationError> errors)
+        {
+            StringBuilder sb = new();
```
msbw2, line 186:
```cs
if (errors.Length == 1)
    return errors[0].ToString();
```

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+            else
+            {
+                keys = [resolvedKey];
+                keyMatched = true;
+            }
+
+            bool canMatchKey = samlToken.Assertion.Signature.KeyInfo != null;
+            List<ValidationError> errors = new();
+            StringBuilder keysAttempted = new();
```
msbw2: Is there a `StringBuilder` pool available here?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateToken.Internal.cs
```diff
@@ -63,6 +63,13 @@ internal async Task<ValidationResult<ValidatedToken>> ValidateTokenAsync(
                 return validatedIssuerResult.UnwrapError().AddStackFrame(StackFrames.IssuerValidationFailed);
             }
 
+            var signatureValidationResult = ValidateSignature(samlToken, validationParameters, callContext);
+            if (!signatureValidationResult.IsValid)
+            {
+                StackFrames.SignatureValidationFailed ??= new StackFrame(true);
```
msbw2, line 88: When would this already be set such that we'd end up here?

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2SecurityTokenHandlerTests.ValidateTokenAsyncTests.Signature.cs
```diff
@@ -0,0 +1,211 @@
...
+                    ValidatedToken validatedToken = validationResult.UnwrapResult();
+                    IdentityComparer.AreEqual(validatedToken.SecurityToken, tokenValidationResult.SecurityToken, context);
+                }
+                else
+                {
+                    // Verify the exception provided by both paths match.
+                    var tokenValidationResultException = tokenValidationResult.Exception;
+                    var validationResultException = validationResult.UnwrapError().GetException();
+
+                    if (theoryData.TestId == "Invalid_TokenSignedWithDifferentKey_KeyIdPresent_TryAllKeysFalse")
```
msbw2: Why is this needed? Just left over from debugging?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: `SignatureValidatorUsingConfiguration` only exists in `TokenValidationParameters`. This path uses the new `ValidationParameters` object that has a single `SignatureValidator`

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2SecurityTokenHandlerTests.ValidateTokenAsyncTests.Signature.cs
iNinja: Yes. This was for debug purposes.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: Neither `SamlSecurityToken` nor `Saml2SecurityToken` can be instantiated without the assertion. It is always not null.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: This path uses `ValidationParameters` which unified the issuer signing keys into a single list and the delegates into the one `SignatureValidator`.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: `Verify` is a method on `Signature`, as long as `Signature` exists, `Verify` will always be there.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: Yes, but we are replacing this in another open PR.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateToken.Internal.cs
iNinja: I am not sure I follow the question. If the signature validation fails, we ensure the stack frame is created, then we return the error including it.
We are moving the caching logic away in a separate PR.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: Updated to avoid creating the list in a successful scenario.
I'll keep further optimisations for the future, as we have an issue created to refactor the signature validation process further.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: I don't think so. I changed this to avoid creating the string builder in the successful scenario.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: Added an optimisation for the case where the key is resolved, separating the degenerate scenario where we iterate over every available key.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ValidateSignature.cs
```diff
@@ -66,57 +65,40 @@ internal static ValidationResult<SecurityKey> ValidateSignature(
...
             }
 
-            bool canMatchKey = samlToken.Assertion.Signature.KeyInfo != null;
-            List<ValidationError> errors = new();
-            StringBuilder keysAttempted = new();
-
             if (keys is not null)
             {
+                // Control reaches here only if the key could not be resolved and TryAllIssuerSigningKeys is set to true.
+                // We try all the keys in the list and return the first valid key. This is the degenerate case.
```
westin-m, line 86: > This is the degenerate case.

**Question (non-blocking)** What does "degenerate case" mean in this context? Are there other cases that can be called out in comments to clarify this?

iNinja: I originally borrowed the term from our JWT handler. I read it as a bit of an edge case where we are not focused on obtaining the best performance, as if that were the case, we would have gone with the alternative/trivial case.

In this context, it means that the application was configured to say "if you can't resolve the key, here's potentially a bunch of others you can try, even though that may take time", by enabling `TryAllIssuerKeys` and receiving multiple keys to try.

The expected most common scenario is that the token has a key ID identifying the key it was signed with, we either received from the user or from the configuration a key with that same ID, we match them, verify the signature and either succeed or fail in a single attempt.

westin-m (approved):

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ValidateSignature.cs
```diff
@@ -126,38 +108,88 @@ internal static ValidationResult<SecurityKey> ValidateSignature(
...
+                        callContext);
+
+            if (!algorithmValidationResult.IsValid)
+            {
+                return algorithmValidationResult.UnwrapError().AddCurrentStackFrame();
+            }
+            else
+            {
+                var validationError = samlToken.Assertion.Signature.Verify(
+                key,
```
brentschmaltz: tab the parameters over one tab.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ValidateSignature.cs
```diff
@@ -66,57 +65,40 @@ internal static ValidationResult<SecurityKey> ValidateSignature(
                 resolvedKey = SamlTokenUtilities.ResolveTokenSigningKey(samlToken.Assertion.Signature.KeyInfo, validationParameters);
             }
 
-            if (resolvedKey is null)
+            bool canMatchKey = samlToken.Assertion.Signature.KeyInfo != null;
+            List<ValidationError>? errors = null;
+            ValidationError? error = null;
+            StringBuilder? keysAttempted = null;
```
brentschmaltz: Can we move these local variables closer to usage.
For example, keysAttempted many never be used if we return on line 78.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ValidateSignature.cs
```diff
@@ -126,38 +108,88 @@ internal static ValidationResult<SecurityKey> ValidateSignature(
...
                 new MessageDetail(TokenLogMessages.IDX10500),
                 ValidationFailureType.SignatureValidationFailed,
                 typeof(SecurityTokenSignatureKeyNotFoundException),
-                new StackFrame(true));
+                ValidationError.GetCurrentStackFrame());
+        }
+
+        private static ValidationResult<SecurityKey> ValidateSignatureUsingKey(SecurityKey key, SamlSecurityToken samlToken, ValidationParameters validationParameters, CallContext callContext)
+        {
+            ValidationResult<string> algorithmValidationResult = validationParameters.AlgorithmValidator(
```
SaurabhMSFT, line 143: `var` here

iNinja: It is clearer to people reading the code if we use the specific return type instead of leaving var and people guessing / inspecting the return type.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ValidateSignature.cs
```diff
@@ -126,38 +108,88 @@ internal static ValidationResult<SecurityKey> ValidateSignature(
...
+                }
+                else
+                {
+                    return validationError.AddCurrentStackFrame();
+                }
+            }
         }
 
-        private static string GetErrorStrings(List<ValidationError> errors)
+        private static string GetErrorStrings(ValidationError? error, List<ValidationError>? errors)
```
SaurabhMSFT: could you use errorList instead of error? The variable names are too to close to differentiate on casual reading.

SaurabhMSFT: `GetErrorStrings` -> `GetErrorString` as we are returning a single string.

brentschmaltz: Several spots in our runtime we build strings, I opened https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/2986 to address.

iNinja: Thanks. I'm trying to think of a way to handle this. I'll share if I come up with any ideas.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,200 @@
...
+            if (validationParameters.SignatureValidator is not null)
+                return validationParameters.SignatureValidator(samlToken, validationParameters, null, callContext);
+
+            // If the user wants to accept unsigned tokens, they must set validationParameters.SignatureValidator
+            if (samlToken.Assertion.Signature is null)
+                return new XmlValidationError(
+                    new MessageDetail(
+                        TokenLogMessages.IDX10504,
+                        samlToken.Assertion.CanonicalString),
+                    ValidationFailureType.SignatureValidationFailed,
```
brentschmaltz: Could we return a new ValidationFailureType?
MissingSignatureFailureType.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,200 @@
...
+                    typeof(SecurityTokenValidationException),
+                    ValidationError.GetCurrentStackFrame());
+
+            IList<SecurityKey>? keys = null;
```
brentschmaltz: As in SamlSecurityTokenHandler, it is easier to review if local variables are declared just before they are used.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,200 @@
...
+
+                error = result.UnwrapError();
+            }
+            else
+            {
+                if (validationParameters.TryAllIssuerSigningKeys)
+                    keys = validationParameters.IssuerSigningKeys;
+            }
+
+            if (keys is not null)
```
brentschmaltz: We may not need the 'keys' local variable.
Could we write
```cs
if (validationParameters.TryAllIssuerSigningKeys && resolvedKey == null)
    for (int i=0; i < validationParameters.IssuerSigningKeys.Count; i++)
    {

    }
```

May be able to remove the else clause above.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,200 @@
...
+            {
+                // Control reaches here only if the key could not be resolved and TryAllIssuerSigningKeys is set to true.
+                // We try all the keys in the list and return the first valid key. This is the degenerate case.
+                for (int i = 0; i < keys.Count; i++)
+                {
+                    SecurityKey key = keys[i];
```
brentschmaltz: A key could be null, we should check.
Could we write
```cs
if (keys[i] != null)
{
    var result = ValidateSignatureUsingKey(keys[i], samlToken, validationParameters, callContext);
    ...
}
```
Also skipping the local variable 'key'.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,200 @@
...
+                        callContext);
+
+            if (!algorithmValidationResult.IsValid)
+            {
+                return algorithmValidationResult.UnwrapError().AddCurrentStackFrame();
+            }
+            else
+            {
+                var validationError = samlToken.Assertion.Signature.Verify(
+                key,
```
brentschmaltz: tab over parameters.

brentschmaltz (changes requested):
Some small requests, looks good.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,200 @@
...
+                    return key;
+                }
+                else
+                {
+                    return validationError.AddCurrentStackFrame();
+                }
+            }
+        }
+
+        private static string GetErrorStrings(ValidationError? error, List<ValidationError>? errors)
```
SaurabhMSFT, line 173: This seems like a copy of method in the previous file. Is there any reason we are duplicating the code.

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2SecurityTokenHandlerTests.ValidateTokenAsyncTests.Signature.cs
```diff
@@ -0,0 +1,207 @@
...
+
+                    theoryData.ExpectedException.ProcessException(tokenValidationResultException, context);
+                    theoryData.ExpectedExceptionValidationParameters!.ProcessException(validationResultException, context);
+                }
+
+                TestUtilities.AssertFailIfErrors(context);
+            }
+        }
+
+        public static TheoryData<ValidateTokenAsyncSignatureTheoryData> ValidateTokenAsync_Signature_TestCases
```
SaurabhMSFT, line 65: this should be private static method instead of a property

SaurabhMSFT (dismissed):
.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
brentschmaltz: Agreed this could be common, nothing token specific about it.

iNinja: Yes, because the SAML2 code is almost a copy of the SAML one, but we are not in a position to refactor them into anything different at this time.

test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2SecurityTokenHandlerTests.ValidateTokenAsyncTests.Signature.cs
iNinja: As I replied in a different PR where you suggested the same about making the TheoryData private, XUnit requires this to be public.

Can you elaborate on on the reason why this should be a method instead of a property? We have all of our tests written using this as a property.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: @brentschmaltz unless we do some refactor to share code between SAML and SAML2, I don't see the point in introducing code sharing for one helper method.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: Added a type, but following the naming you used in the extensibility `IssuerValidationDelegateThrew`, I think that is a good naming convention making this one `TokenIsNotSigned`. I don't think we need `FailureType` in the name.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: We use this variable in 4 places in this method. I think I'd rather keep it.

Adding the null check.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: Removed the `keys` local variable and simplified the code a bit. We do need to check for `keyMatched` to avoid re-using the key if it was resolved.

brentschmaltz (approved):

PR #2976, opened 2024-11-04
New token validation model: Simplify stack frame caching

iNinja:
# New token validation model: Simplify stack frame caching
As part of the new validation model, we are manually capturing stack frames when returning an error in order to build the stack frame once the exception is created. In order to avoid the performance hit incurred when capturing the stack trace at runtime, we cache each stack frame and avoid creating it if it already exists.
Currently, we have around 100 stack frames between all of the possible code branches that need to be captured, named, stored, and checked. This PR attempts to simplify this process so that we memoise the stack frames in a concurrent dictionary and retrieve them when needed.

- Added a `ConcurrentDictionary` to the `ValidationError` class to store stack frames for each error message
- Added methods `AddCurrentStackFrame` and `GetCurrentStackFrame` to the `ValidationError`
- Added a test to validate the stack frames being stored and retrieved correctly

Part of #2711

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
@@ -2,8 +2,10 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 using Microsoft.IdentityModel.Logging;
 
 namespace Microsoft.IdentityModel.Tokens
@@ -73,7 +75,7 @@ namespace Microsoft.IdentityModel.Tokens
             if (innerException == null && InnerValidationError == null)
             {
                 if (exceptionType == typeof(SecurityTokenArgumentNullException))
-                    return new SecurityTokenArgumentNullException(MessageDetail.Message);
+                    exception = new SecurityTokenArgumentNullException(MessageDetail.Message);
                 else if (exceptionType == typeof(SecurityTokenInvalidAudienceException))
                     exception = new SecurityTokenInvalidAudienceException(MessageDetail.Message);
                 else if (exceptionType == typeof(SecurityTokenInvalidIssuerException))
@@ -187,6 +189,11 @@ namespace Microsoft.IdentityModel.Tokens
                 }
             }
 
+            if (exception is SecurityTokenException securityTokenException)
+                securityTokenException.SetValidationError(this);
+            else if (exception is SecurityTokenArgumentNullException securityTokenArgumentNullException)
+                securityTokenArgumentNullException.SetValidationError(this);
+
             return exception;
         }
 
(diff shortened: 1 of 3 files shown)


brentschmaltz (approved):

kllysng (approved):

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
```diff
@@ -236,5 +243,41 @@ public ValidationError AddStackFrame(StackFrame stackFrame)
...
+        /// Adds the current stack frame to the list of stack frames and returns the updated object.
+        /// If there is no cache entry for the given file path and line number, a new stack frame is created and added to the cache.
+        /// </summary>
+        /// <param name="filePath">The path to the file from which this method is called. Captured automatically by default.</param>
+        /// <param name="lineNumber">The line number from which this method is called. CAptured automatically by default.</param>
+        /// <param name="skipFrames">The number of stack frames to skip when capturing. Used to avoid capturing this method and get the caller instead.</param>
+        /// <returns>The updated object.</returns>
+        internal ValidationError AddCurrentStackFrame([CallerFilePath] string filePath = "", [CallerLineNumber] int lineNumber = 0, int skipFrames = 1)
+        {
+            // We add 1 to the skipped frames to skip the current method
```
msbw2, line 257: If `skipFrames` defaults to `1` and you add `1` to it, is this comment accurate?

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
```diff
@@ -236,5 +243,41 @@ public ValidationError AddStackFrame(StackFrame stackFrame)
...
+            [CallerFilePath] string filePath = "", [CallerLineNumber] int lineNumber = 0, int skipFrames = 1)
+        {
+            // String is allocated, but it goes out of scope immediately after the call
+            string key = filePath + lineNumber;
+            StackFrame frame = CachedStackFrames.GetOrAdd(key, new StackFrame(skipFrames, true));
+            return frame;
+        }
+
+        // ConcurrentDictionary is thread-safe and only locks when adding a new item.
+        private static ConcurrentDictionary<string, StackFrame> CachedStackFrames { get; } = new();
```
msbw2, line 281: Will this cause issues in tests? Multiple sources can call into the same method, so you might return the wrong stack trace.

msbw2: Okay. Since this is a single frame from the frames provided, it should be fine.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
iNinja: Say we call:
DoSomething(), which calls
AddCurrentStackFrame(), which calls
GetCurrentStackFrame(), which instantiates
new StackFrame(true).

`GetCurrentStackFrame()` passes the 1 skip frame to skip itself from the stack and get the method that called it, which is the normal behaviour for any cases where we get the stack frame and use it, say to instantiate a new `ValidationError`.
But when using `AddCurrentStackFrame(()`, we need to skip one more layer to account for both `GetCurrentStackFrame()` and `AddCurrentStackFrame()`, and only capture the caller to this method.

Currently, we use the two methods for different situations:
- `GetCurrentStackFrame()` is used when we are creating a new `ValidationError` and we are pointing at where that occurs.
- `AddCurrentStackFrame()` is used when we have an existing `ValidationError` with some stack frames in it, to add the next stack frame and create the right chain for the stack trace.

You can see an example of this in the test class used in the test.

PR #2979, opened 2024-11-05
Add ability to exclude default header claims when creating token

msbw2:
Add ability to exclude default header claims when creating token.


keegan-caruso:
Is there an issue tracking this?

msbw2:
> Is there an issue tracking this?

Not at the moment. This was just something discussed since we wanted to remove `kid` before but found it to be impossible without changes.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
```diff
@@ -957,14 +961,20 @@ internal static void AddSubjectClaims(
             nbfSet |= nbfReset;
         }
 
+        internal static bool IncludeDefaultHeaderClaim(string claim, ISet<string> excludedDefaultHeaderClaims) =>
+            excludedDefaultHeaderClaims is null ||
+            excludedDefaultHeaderClaims.Count == 0 ||
+            !excludedDefaultHeaderClaims.Contains(claim);
+
         internal static void WriteJwsHeader(
```
keegan-caruso, line 886: The number of parameters is getting excessive.

Suggestion: always use the SecurityTokenDescriptor, we would allocate on the code path that doesn't pass in an object, but if the user wants to avoid the allocation, they can use the overload that does accept a SecurityTokenDescriptor

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
```diff
@@ -94,6 +94,16 @@ public class SecurityTokenDescriptor
...
         public IDictionary<string, object> AdditionalInnerHeaderClaims { get; set; }
 
+        /// <summary>
+        /// Gets or sets the <see cref="HashSet{T}"/> which contains the default header claims that should be excluded from the JWT header.
+        /// </summary>
+        /// <remarks>
+        /// The default header claims are 'alg', 'kid', 'x5t', 'enc', and 'zip'.
+        /// </remarks>
+#pragma warning disable CA2227 // Collection properties should be read only
+        public ISet<string> ExcludedDefaultHeaderClaims { get; set; }
```
keegan-caruso, line 107: What is the relationship between if kid is included here and "IncludeKeyIdInHeader"?

Can that be included in the doc string here and likely in "IncludeKeyIdInHeader"?

keegan-caruso, line 107: In general, Exclude APIs are not common in Identity Model. What are the alternatives that we considered?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
msbw2: I think this should be a separate PR, but I agree.

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
msbw2: Ideally we'd remove `IncludeKeyIdInHeader` and let this supersede it.

msbw2: I don't think any others were considered. I proposed this since only excluding `kid` and `x5t` was very targeted and iffy, and the only other option that comes to mind is allocating a set with all the default header claims in every `SecurityTokenDescriptor` which can be modified as desired, but then you end up with those allocations.

msbw2: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2991

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
msbw2: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2991

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
```diff
@@ -94,6 +94,16 @@ public class SecurityTokenDescriptor
         /// </summary>
         public IDictionary<string, object> AdditionalInnerHeaderClaims { get; set; }
 
+        /// <summary>
+        /// Gets or sets the <see cref="HashSet{T}"/> which contains the default header claims that should be excluded from the JWT header.
+        /// </summary>
+        /// <remarks>
+        /// The default header claims are 'alg', 'kid', 'x5t', 'enc', and 'zip'.
```
jmprieur: and by default, they are not excluded.?

jmprieur (approved):
LGTM
Thanks @msbw2

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
keegan-caruso: Let's at least update the doc string for both properties for now.

src/Microsoft.IdentityModel.Tokens/SecurityTokenDescriptor.cs
msbw2: By default this set is empty so the default header claims that apply are included.

keegan-caruso:
If still needed, someone from the team will pick this up.

PR #2980, opened 2024-11-05
SAML and SAML2 new model validation: Read Token

iNinja:
# SAML and SAML2 new model validation: Read Token
- Added ReadToken method to both SAML and SAML2 security token handlers returning ValidationResult
- Added ValidationError subclasses to return custom exceptions when reading fails
- Added tests


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Exceptions/Saml2ValidationError.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.IdentityModel.Tokens.Saml2
+{
+    internal class Saml2ValidationError : ValidationError
```
JoshLozensky, line 9: Do we really need separate classes for SAML validation errors?

iNinja: This is a decision we made as part of the new validation model, to move the responsibility of creating specific exceptions to subclasses of ValidationError.

We are still defining the final shape, but it is likely that we will enforce the delegates to return a specialised kind of ValidationError to ensure that any custom implementation accounts for the specifics that these exceptions can present to the user.

iNinja: This case may not have the same weight as those that provide custom data to the exception, but I tried to align the approaches.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -804,6 +804,49 @@ public override SecurityToken ReadToken(XmlReader reader, TokenValidationParamet
             throw LogHelper.LogExceptionMessage(new NotSupportedException(LogMessages.IDX11950));
         }
 
+        /// <summary>
```
brentschmaltz: Should this be in SamlSecurityTokenHandler.Validation.Internal.cs?

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -804,6 +804,49 @@ public override SecurityToken ReadToken(XmlReader reader, TokenValidationParamet
...
+            if (string.IsNullOrEmpty(token))
+                return ValidationError.NullParameter(nameof(token), ValidationError.GetCurrentStackFrame());
+
+            if (token.Length > MaximumTokenSizeInBytes)
+                return new ValidationError(
+                        new MessageDetail(
+                            TokenLogMessages.IDX10209,
+                            LogHelper.MarkAsNonPII(token.Length),
+                            LogHelper.MarkAsNonPII(MaximumTokenSizeInBytes)),
+                        ValidationFailureType.TokenReadingFailed,
```
brentschmaltz: We should have ValidationFailureType.TokenExceedsMaximumSize.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -804,6 +804,49 @@ public override SecurityToken ReadToken(XmlReader reader, TokenValidationParamet
...
+                        typeof(ArgumentException),
```
brentschmaltz: ArgumentOutOfRangeException would be more correct. Since ArgumentOutOfRangeException derives from ArgumentException this would not be breaking for users.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Exceptions/Saml2ValidationError.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        internal Saml2ValidationError(MessageDetail MessageDetail, ValidationFailureType failureType, Type exceptionType, StackFrame stackFrame) : base(MessageDetail, failureType, exceptionType, stackFrame)
```
brentschmaltz: We don't seem to call this ctor, do we need it?

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Exceptions/SamlValidationError.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.IdentityModel.Tokens.Saml
+{
+    internal class SamlValidationError : ValidationError
+    {
+        internal SamlValidationError(MessageDetail MessageDetail, ValidationFailureType failureType, Type exceptionType, StackFrame stackFrame) : base(MessageDetail, failureType, exceptionType, stackFrame)
```
brentschmaltz: We don't use this ctor, do we need it?

brentschmaltz (changes requested):
Looks good, some small comments on removing ctors and moving code to .internal.cs.

SaurabhMSFT (approved):

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
iNinja: Moved to .ReadToken, like we did with JWT.

brentschmaltz (approved):

PR #2984, opened 2024-11-06
SAML and SAML2 new model validation: Algorithm

iNinja:
# SAML and SAML2 new model validation: Algorithm
- Added tests for SAML and SAML2 algorithm validation using the new validation model

Note: This PR is branched from the signature one because it depends on changes made there. Once that PR is merged, only the algorithm tests should remain.

Part of #2711


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
```diff
@@ -0,0 +1,200 @@
...
+            bool keyMatched = false;
+
+            if (validationParameters.IssuerSigningKeyResolver is not null)
+            {
+                resolvedKey = validationParameters.IssuerSigningKeyResolver(
+                    samlToken.Assertion.CanonicalString,
+                    samlToken,
+                    samlToken.Assertion.Signature.KeyInfo?.Id,
+                    validationParameters,
+                    null,
```
brentschmaltz: Should IssuerSigningKeyResolverDelegate take BaseConfiguration as a parameter?

brentschmaltz (approved):
There are four PR's somewhat related, will review again once all PR's are in.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateSignature.cs
iNinja: The SAML and SAML2 token handlers at this point do not deal with configuration and refreshing it, and simply copy any relevant parameters to ValidationParameters at the beginning of the execution of ValidateTokenAsync.

Passing the BaseConfiguration here would potentially present duplicated keys.

brentschmaltz (approved):

jennyf19 (approved):

PR #2987, opened 2024-11-07
JsonWebTokenHandler IssuerExtensibility

brentschmaltz:
Added tests for IssuerExtensibility to understand the user experience.
Some improvements to the call graph, but those were small.


iNinja (approved):
LGTM

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/IssuerValidationError.cs
```diff
@@ -9,25 +9,36 @@ namespace Microsoft.IdentityModel.Tokens
...
             Type exceptionType,
             StackFrame stackFrame,
             string? invalidIssuer)
-            : base(messageDetail, ValidationFailureType.IssuerValidationFailed, exceptionType, stackFrame)
+            : this(messageDetail, ValidationFailureType.IssuerValidationFailed, exceptionType, stackFrame, invalidIssuer, null)
         {
-            _invalidIssuer = invalidIssuer;
         }
 
+        internal IssuerValidationError(
```
keegan-caruso, line 21: Multiple ctors with different numbers of optional parameters gets confusing fast.

This is internal, is it safe to change the above ctor? That is, because of the friend assemblies, is this already used in another assembly?

If it is not, can we just have the one ctor with both params optional?

See RS0026 as well

https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md

Edit:

Ignore me, I misread these as optional, these are nullable.

test/Microsoft.IdentityModel.TestUtils/IdentityComparer.cs
```diff
@@ -1406,6 +1406,128 @@ internal static bool AreValidatedTokenTypesEqual(ValidatedTokenType validatedTok
...
+                {
+                    localContext.Diffs.Add($"(validationError1.StackFrames[0].GetFileName(): " +
+                        $"'{validationError1.StackFrames[0].GetFileName()}', " +
+                        $"does not contain validationError2.StackFrames[0].GetFileName():" +
+                        $"'{validationError1.StackFrames[0].GetFileName()}'.");
+                }
+            }
+
+            AreStringsEqual(
+                validationError1.GetType().ToString(),
```
keegan-caruso: Should we use GetType().FullName instead? Idea being you validate the FQ name of the type.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Common.cs
```diff
@@ -101,7 +101,7 @@ public ValidateTokenAsyncBaseTheoryData(string testId) : base(testId) { }
 
         internal TokenValidationParameters? TokenValidationParameters { get; set; }
 
-        internal ValidationParameters? ValidationParameters { get; set; }
+        internal virtual ValidationParameters? ValidationParameters { get; set; }
```
keegan-caruso: Where is this overridden at? Why can't it just be set? using a virtual makes it a little more complicated.

test/Microsoft.IdentityModel.TestUtils/CustomValidationDelegates.cs
```diff
@@ -0,0 +1,215 @@
...
+using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Tokens;
+
+#nullable enable
+namespace Microsoft.IdentityModel.TestUtils
+{
+    internal class CustomSecurityTokenInvalidIssuerException : SecurityTokenInvalidIssuerException
```
keegan-caruso: **nit**: It might read a little easier to separate out exceptions and delegates

test/Microsoft.IdentityModel.TestUtils/CustomValidationDelegates.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+            return await Task.FromResult(new ValidationResult<ValidatedIssuer>(
+                new IssuerValidationError(
+                    new MessageDetail(nameof(IssuerValidatorDelegate), null),
+                    typeof(SecurityTokenInvalidIssuerException),
+                    ValidationError.GetCurrentStackFrame(),
+                    issuer)));
+        }
+
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
```
keegan-caruso: This shouldn't be async should it?

test/Microsoft.IdentityModel.TestUtils/CustomValidationDelegates.cs
```diff
@@ -0,0 +1,215 @@
...
+            StackFrame stackFrame,
+            string? invalidIssuer) :
+            base(messageDetail, exceptionType, stackFrame, invalidIssuer)
+        {
+        }
+    }
+
+    internal class CustomIssuerValidatorDelegates
+    {
+        internal async static Task<ValidationResult<ValidatedIssuer>> CustomIssuerValidatorDelegate(
```
keegan-caruso: All of these should end with the "async" suffix I think. Even for test code, it communicates intent better.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.Issuer.Extensibility.cs
```diff
@@ -0,0 +1,283 @@
...
+        public static TheoryData<IssuerExtensibilityTheoryData> Issuer_ExtensibilityTestCases
+        {
+            get
+            {
+                var theoryData = new TheoryData<IssuerExtensibilityTheoryData>();
+                CallContext callContext = new CallContext();
+                string issuerGuid = Guid.NewGuid().ToString();
+
+                #region return CustomIssuerValidationError
+                // Test cases where delegate is overridden and return an CustomIssuerValidationError
```
keegan-caruso, line 66: having the explanations next to each is awesome.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.Issuer.Extensibility.cs
```diff
@@ -0,0 +1,283 @@
...
+                        new StackFrame("CustomValidationDelegates.cs", 88),
+                        issuerGuid)
+                });
+
+                // CustomIssuerValidationError : IssuerValidationError, ExceptionType: CustomSecurityTokenInvalidIssuerException : SecurityTokenInvalidIssuerException
+                theoryData.Add(new IssuerExtensibilityTheoryData(
+                    "CustomIssuerValidatorCustomExceptionDelegate",
+                    issuerGuid,
+                    CustomIssuerValidatorDelegates.CustomIssuerValidatorCustomExceptionDelegate,
+                    new List<StackFrame> {
```
keegan-caruso: nit: could do `[]` instead of new List throughout, it reads a little cleaner.

keegan-caruso (approved):
Just some small things, but nothing functionally blocking

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests.Common.cs
iNinja: It was originally overridden, but after some reviews that was dropped. I guess we can remove the virtual keyword.

PR #2988, opened 2024-11-08
Use int to guard when metadata update is due

### LINKED ISSUE #2900: [Bug] Race condition on ConfigurationManager._syncAfter
Microsoft.IdentityModel 8.1.2

**Where is the issue?**
 * [ ] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [X] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [ ] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [ ] S.IM.Tokens.Jwt
 * Other (please describe)

**Description**

Every calling thread to `ConfigurationManager.GetConfigurationAsync` will result in a read to `_syncAfter` to check if a sync should be started; later, a background thread will update `_syncAfter` with the new sync time. `syncAfter` is a DateTimeOffset, a struct, so writes to it aren't atomic, so there's a possibility that the calling thread will read corrupt data.

**Possible solution**

* Rewrite `_syncAfter` so that it can be atomically read/written - we could refactor it to be total seconds (`int32`) or any other type that is atomic.
* Rearchitect the metadata refresh to happen completely in the background, to avoid every calling thread having to access the syncAfter variable.

- [_syncAfter read](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/2347f09f25206d4546d66eae974c6d5e55f81e42/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs#L150)

brentschmaltz:
The previous guard was a DateTimeOffset which could be in an indeterminate state when being read.
This PR records the DateTimeOffset when ConfigurationManager is instantiated StartupTime.
The next refresh (automatic or requestRefresh) is an offset in seconds from the StartupTime.

An int is used to store the offset of the last refresh.
If a service runs over 68 years, int will overflow, we could have some complicated logic to avoid this, it seems ok to leave it as is.

We also could have performed a simple calculation of:
NumberOfTimesAutomaticRefreshRequested  * _automaticRefreshIntervalInSeconds > (StartupTime - UtcNow).TotalSeconds
We add a small amount %5 in a random manner when calculating the next interval, adds complexity.

We added internals to aid in testing.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -18,8 +18,6 @@ namespace Microsoft.IdentityModel.Protocols
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable")]
     public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class
     {
-        private DateTimeOffset _syncAfter = DateTimeOffset.MinValue;
-        private DateTimeOffset _lastRequestRefresh = DateTimeOffset.MinValue;
         private bool _isFirstRefreshRequest = true;
         private readonly SemaphoreSlim _configurationNullLock = new SemaphoreSlim(1);
 
@@ -133,7 +131,8 @@ namespace Microsoft.IdentityModel.Protocols
         /// Obtains an updated version of Configuration.
         /// </summary>
         /// <returns>Configuration of type T.</returns>
-        /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
+        /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then
+        /// <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
         public async Task<T> GetConfigurationAsync()
         {
             return await GetConfigurationAsync(CancellationToken.None).ConfigureAwait(false);
@@ -144,11 +143,17 @@ namespace Microsoft.IdentityModel.Protocols
         /// </summary>
         /// <param name="cancel">CancellationToken</param>
         /// <returns>Configuration of type T.</returns>
-        /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
+        /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then
+        /// <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
         public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
         {
-            if (_currentConfiguration != null && _syncAfter > DateTimeOffset.UtcNow)
-                return _currentConfiguration;
+            if (_currentConfiguration != null)
+            {
+                // StartupTime is the time when ConfigurationManager was instantiated.
+                double nextRefresh = _automaticRefreshIntervalInSeconds + _timeInSecondsWhenLastRefreshOccurred;
+                if (nextRefresh > GetSecondsSinceInstanceWasCreated)
+                    return _currentConfiguration;
+            }
 
             Exception fetchMetadataFailure = null;
 
@@ -157,7 +162,7 @@ namespace Microsoft.IdentityModel.Protocols
             //   reach out to the metadata endpoint. Since multiple threads could be calling this method
             //   we need to ensure that only one thread is actually fetching the metadata.
             // else
-            //   if task is running, return the current configuration
+            //   if update task is running, return the current configuration
             //   else kick off task to update current configuration
             if (_currentConfiguration == null)
             {
@@ -168,9 +173,11 @@ namespace Microsoft.IdentityModel.Protocols
                     return _currentConfiguration;
                 }
 
-#pragma warning disable CA1031 // Do not catch general exception types
                 try
                 {
+                    _configurationRetrieverState = ConfigurationRetrieverRunning;
+                    Interlocked.Increment(ref _numberOfTimesAutomaticRefreshRequested);
+
                     // Don't use the individual CT here, this is a shared operation that shouldn't be affected by an individual's cancellation.
                     // The transport should have it's own timeouts, etc.
                     T configuration = await _configRetriever.GetConfigurationAsync(
@@ -192,7 +199,9 @@ namespace Microsoft.IdentityModel.Protocols
 
                     UpdateConfiguration(configuration);
                 }
+#pragma warning disable CA1031 // Do not catch general exception types
                 catch (Exception ex)
+#pragma warning restore CA1031 // Do not catch general exception types
                 {
                     fetchMetadataFailure = ex;
 
@@ -206,15 +215,16 @@ namespace Microsoft.IdentityModel.Protocols
                 }
                 finally
                 {
+                    _configurationRetrieverState = ConfigurationRetrieverIdle;
                     _configurationNullLock.Release();
                 }
-#pragma warning restore CA1031 // Do not catch general exception types
             }
             else
             {
                 if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
                 {
-                    _ = Task.Run(UpdateCurrentConfiguration, CancellationToken.None);
+                    Interlocked.Increment(ref _numberOfTimesAutomaticRefreshRequested);
+                    _ = Task.Run(RetrieveAndUpdateConfiguration, CancellationToken.None);
                 }
             }
 
@@ -227,7 +237,7 @@ namespace Microsoft.IdentityModel.Protocols
                     LogHelper.FormatInvariant(
                         LogMessages.IDX20803,
                         LogHelper.MarkAsNonPII(MetadataAddress ?? "null"),
-                        LogHelper.MarkAsNonPII(_syncAfter),
+                        LogHelper.MarkAsNonPII(_timeInSecondsWhenLastRefreshOccurred),
                         LogHelper.MarkAsNonPII(fetchMetadataFailure)),
                     fetchMetadataFailure));
         }
@@ -235,11 +245,10 @@ namespace Microsoft.IdentityModel.Protocols
         /// <summary>
         /// This should be called when the configuration needs to be updated either from RequestRefresh or AutomaticRefresh
         /// The Caller should first check the state checking state using:
-        ///   if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverIdle, ConfigurationRetrieverRunning) != ConfigurationRetrieverRunning).
+        ///   if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle).
         /// </summary>
-        private void UpdateCurrentConfiguration()
+        private void RetrieveAndUpdateConfiguration()
         {
-#pragma warning disable CA1031 // Do not catch general exception types
             try
             {
                 T configuration = _configRetriever.GetConfigurationAsync(
@@ -265,7 +274,9 @@ namespace Microsoft.IdentityModel.Protocols
                         UpdateConfiguration(configuration);
                 }
             }
+#pragma warning disable CA1031 // Do not catch general exception types
             catch (Exception ex)
+#pragma warning restore CA1031 // Do not catch general exception types
             {
                 LogHelper.LogExceptionMessage(
                     new InvalidOperationException(
@@ -279,14 +290,24 @@ namespace Microsoft.IdentityModel.Protocols
             {
                 Interlocked.Exchange(ref _configurationRetrieverState, ConfigurationRetrieverIdle);
             }
-#pragma warning restore CA1031 // Do not catch general exception types
         }
 
+        /// <summary>
+        /// Called only when configuration is successfully obtained.
+        /// </summary>
+        /// <param name="configuration">Set <see cref="_currentConfiguration" /> to this value.</param>
         private void UpdateConfiguration(T configuration)
         {
             _currentConfiguration = configuration;
-            _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval +
-                TimeSpan.FromSeconds(new Random().Next((int)AutomaticRefreshInterval.TotalSeconds / 20)));
+            // StartupTime is the time when ConfigurationManager was instantiated.
+            // SecondsSinceInstanceWasCreated is the number of seconds since ConfigurationManager was instantiated.
+            // For automatic refresh, we add a 5% jitter.
+            // Record in seconds when the last time configuration was obtained.
+            double timeInSecondsWhenLastAutomaticRefreshOccurred = GetSecondsSinceInstanceWasCreated +
+                                 ((_automaticRefreshIntervalInSeconds >= int.MaxValue) ? 0 : (_random.Next((int)_maxJitter)));
+
+            // transfer to int in single operation.
+            _timeInSecondsWhenLastRefreshOccurred = (int)((timeInSecondsWhenLastAutomaticRefreshOccurred <= int.MaxValue) ? (int)timeInSecondsWhenLastAutomaticRefreshOccurred : int.MaxValue);
         }
 
         /// <summary>
@@ -294,7 +315,7 @@ namespace Microsoft.IdentityModel.Protocols
         /// </summary>
         /// <param name="cancel">CancellationToken</param>
         /// <returns>Configuration of type BaseConfiguration    .</returns>
-        /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
+        /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager._automaticRefreshIntervalInSeconds"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
         public override async Task<BaseConfiguration> GetBaseConfigurationAsync(CancellationToken cancel)
         {
             T obj = await GetConfigurationAsync(cancel).ConfigureAwait(false);
@@ -309,18 +330,29 @@ namespace Microsoft.IdentityModel.Protocols
         /// </summary>
         public override void RequestRefresh()
         {
-            DateTimeOffset now = DateTimeOffset.UtcNow;
+            if (_configurationRetrieverState == ConfigurationRetrieverRunning)
+                return;
 
-            if (now >= DateTimeUtil.Add(_lastRequestRefresh.UtcDateTime, RefreshInterval) || _isFirstRefreshRequest)
+            double nextRefresh = _requestRefreshIntervalInSeconds + _timeInSecondsWhenLastRequestRefreshWasRequested;
+            if (nextRefresh < GetSecondsSinceInstanceWasCreated || _isFirstRefreshRequest)
             {
                 _isFirstRefreshRequest = false;
                 if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
                 {
-                    _ = Task.Run(UpdateCurrentConfiguration, CancellationToken.None);
-                    _lastRequestRefresh = now;
+                    Interlocked.Increment(ref _numberOfTimesRequestRefreshRequested);
+                    double recordWhenRefreshOccurred = GetSecondsSinceInstanceWasCreated;
+                    // transfer to int in single operation.
+                    _timeInSecondsWhenLastRequestRefreshWasRequested = (int)((recordWhenRefreshOccurred <= int.MaxValue) ? (int)recordWhenRefreshOccurred : int.MaxValue);
+                    _ = Task.Run(RetrieveAndUpdateConfiguration, CancellationToken.None);
                 }
             }
         }
+
+        /// <summary>
+        /// SecondsSinceInstanceWasCreated is the number of seconds since ConfigurationManager was instantiated.
+        /// </summary>
+        /// <returns>double</returns>
+        private double GetSecondsSinceInstanceWasCreated => (DateTimeOffset.UtcNow - StartupTime).TotalSeconds;
 
         /// <summary>
         /// 12 hours is the default time interval that afterwards, <see cref="GetBaseConfigurationAsync(CancellationToken)"/> will obtain new configuration.
--- a/src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
@@ -16,12 +16,21 @@ namespace Microsoft.IdentityModel.Tokens
     public abstract class BaseConfigurationManager
     {
         private TimeSpan _automaticRefreshInterval = DefaultAutomaticRefreshInterval;
+        internal double _automaticRefreshIntervalInSeconds = DefaultAutomaticRefreshInterval.TotalSeconds;
+        internal double _maxJitter = DefaultAutomaticRefreshInterval.TotalSeconds / 20;
         private TimeSpan _refreshInterval = DefaultRefreshInterval;
+        internal double _requestRefreshIntervalInSeconds = DefaultRefreshInterval.TotalSeconds;
         private TimeSpan _lastKnownGoodLifetime = DefaultLastKnownGoodConfigurationLifetime;
         private BaseConfiguration _lastKnownGoodConfiguration;
         private DateTime? _lastKnownGoodConfigFirstUse;
+        internal Random _random = new();
+
+        // Seconds since the BaseConfigurationManager was created when the last refresh occurred with a %5 random jitter.
+        internal int _timeInSecondsWhenLastRefreshOccurred;
+        internal int _timeInSecondsWhenLastRequestRefreshWasRequested;
 
         internal EventBasedLRUCache<BaseConfiguration, DateTime> _lastKnownGoodConfigurationCache;
+
 
         /// <summary>
         /// Gets or sets the <see cref="TimeSpan"/> that controls how often an automatic metadata refresh should occur.
@@ -32,9 +41,59 @@ namespace Microsoft.IdentityModel.Tokens
             set
             {
                 if (value < MinimumAutomaticRefreshInterval)
-                    throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(value), LogHelper.FormatInvariant(LogMessages.IDX10108, LogHelper.MarkAsNonPII(MinimumAutomaticRefreshInterval), LogHelper.MarkAsNonPII(value))));
+                    throw LogHelper.LogExceptionMessage(
+                        new ArgumentOutOfRangeException(
+                            nameof(value),
+                            LogHelper.FormatInvariant(
+                                LogMessages.IDX10108,
+                                LogHelper.MarkAsNonPII(MinimumAutomaticRefreshInterval),
+                                LogHelper.MarkAsNonPII(value))));
 
                 _automaticRefreshInterval = value;
+                Interlocked.Exchange(ref _automaticRefreshIntervalInSeconds, value.TotalSeconds);
+                Interlocked.Exchange(ref _maxJitter, value.TotalSeconds / 20);
+            }
+        }
+
+        /// <summary>
+        /// Records the time this instance was created.
+        /// Used to determine if the automatic refresh or request refresh interval has passed.
+        /// The logic is to remember the number of seconds since startup that the last refresh occurred.
+        /// Store that value in _timeInSecondsWhenLastAutomaticRefreshOccurred or _timeInSecondsWhenLastRequestRefreshOccurred.
+        /// Then compare to (UtcNow - Startup).TotalSeconds.
+        /// The set is used for testing purposes.
+        /// </summary>
+        internal DateTimeOffset StartupTime { get; set; } = DateTimeOffset.UtcNow;
+
+        /// <summary>
+        ///  Incremented each time <see cref="GetBaseConfigurationAsync(CancellationToken)"/> results in a http request.
+        /// </summary>
+        internal long _numberOfTimesAutomaticRefreshRequested;
+
+        /// <summary>
+        /// Thread safe getter for <see cref="_numberOfTimesAutomaticRefreshRequested"/>.
+        /// </summary>
+        internal long NumberOfTimesAutomaticRefreshRequested
+        {
+            get
+            {
+                return Interlocked.Read(ref _numberOfTimesAutomaticRefreshRequested);
+            }
+        }
+
+        /// <summary>
+        /// Incremented each time <see cref="RequestRefresh"/> results in a http request.
+        /// </summary>
+        internal long _numberOfTimesRequestRefreshRequested;
+
+        /// <summary>
+        /// Thread safe getter for <see cref="_numberOfTimesRequestRefreshRequested"/>.
+        /// </summary>
+        internal long NumberOfTimesRequestRefreshRequested
+        {
+            get
+            {
+                return Interlocked.Read(ref _numberOfTimesRequestRefreshRequested);
             }
         }
 
@@ -165,6 +224,7 @@ namespace Microsoft.IdentityModel.Tokens
                     throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(value), LogHelper.FormatInvariant(LogMessages.IDX10107, LogHelper.MarkAsNonPII(MinimumRefreshInterval), LogHelper.MarkAsNonPII(value))));
 
                 _refreshInterval = value;
+                Interlocked.Exchange(ref _requestRefreshIntervalInSeconds, value.TotalSeconds);
             }
         }
 
--- a/test/Microsoft.IdentityModel.TestUtils/EventDrivenConfigurationManager.cs
+++ b/test/Microsoft.IdentityModel.TestUtils/EventDrivenConfigurationManager.cs
@@ -1,0 +1,116 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.TestUtils
+{
+    public class EventDrivenConfigurationRetriever<T> : IConfigurationRetriever<T>
+    {
+        private ManualResetEvent _signalEvent;
+        private ManualResetEvent _waitEvent;
+        private T _configuration;
+
+        /// <summary>
+        /// Initializes an new instance of <see cref="EventDrivenConfigurationRetriever{T}"/> with a configuration instance.
+        /// </summary>
+        /// <param name="configuration">The Configuration that will be returned</param>
+        /// <param name="signalEvent">A <see cref="ManualResetEvent"/>that is be signaled when inside GetConfigurationAsync.</param>
+        /// <param name="waitEvent">A <see cref="ManualResetEvent"/>that waits inside GetConfigurationAsync.</param>
+        public EventDrivenConfigurationRetriever(
+            T configuration,
+            ManualResetEvent signalEvent,
+            ManualResetEvent waitEvent)
+        {
+            _configuration = configuration;
+            _signalEvent = signalEvent;
+            _waitEvent = waitEvent;
+        }
+
+        public Task<T> GetConfigurationAsync(string address, IDocumentRetriever retriever, CancellationToken cancel)
+        {
+            _waitEvent.WaitOne();
+            _signalEvent.Set();
+            return Task.FromResult(_configuration);
+        }
+    }
+
+    /// <summary>
+    /// This type is used for testing the functionality of using a last known good configuration, as well
+    /// as a refreshed configuration.
+    /// </summary>
+    /// <typeparam name="T">must be a class inherit from <see cref="BaseConfiguration"/>.</typeparam>
+    public class EventControlledConfigurationManger<T> : ConfigurationManager<T>, IConfigurationManager<T> where T : class
+    {
+        private ManualResetEvent _configSignalEvent;
+        private ManualResetEvent _configWaitEvent;
+        private ManualResetEvent _refreshSignalEvent;
+        private ManualResetEvent _refreshWaitEvent;
+
+        /// <summary>
+        /// Initializes an new instance of <see cref="EventControlledConfigurationManger{T}"/> with a Configuration instance.
+        /// </summary>
+        /// <param name="metadataAddress">The metadata address to obtain configuration.</param>
+        /// <param name="configurationRetriever">The <see cref="IConfigurationRetriever{T}"/>that reads the metadata.</param>
+        /// <param name="documentRetriever">The <see cref="IDocumentRetriever"/>that obtains the metadata.</param>
+        /// <param name="configSignalEvent">A <see cref="ManualResetEvent"/>that is signaled when GetConfigurationAsync is exiting.</param>
+        /// <param name="configWaitEvent">A <see cref="ManualResetEvent"/>that waits in GetConfigurationAsync after calling base.GetConfigurationAsync. </param>
+        /// <param name="refreshSignalEvent">A <see cref="ManualResetEvent"/> that is signaled when RequestRefresh is exiting.</param>
+        /// <param name="refreshWaitEvent">A <see cref="ManualResetEvent"/>that waits after base.RequestRefresh is called.</param>
+        public EventControlledConfigurationManger(
+            string metadataAddress,
+            IConfigurationRetriever<T> configurationRetriever,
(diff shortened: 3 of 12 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -157,20 +160,19 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
-
-#pragma warning disable CA1031 // Do not catch general exception types
                 try
                 {
+                    await _configurationNullLock.WaitAsync(cancel).ConfigureAwait(false);
+                    if (_currentConfiguration != null)
+                        return _currentConfiguration;
+
+                    Interlocked.Exchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning);
+                    NumberOfTimesAutomaticRefreshRequested++;
```
keegan-caruso: Confirming my understanding:

This doesn't need to be interlocked.increment because of the null lock above?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -157,20 +160,19 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
-                }
-
-#pragma warning disable CA1031 // Do not catch general exception types
                 try
                 {
+                    await _configurationNullLock.WaitAsync(cancel).ConfigureAwait(false);
+                    if (_currentConfiguration != null)
+                        return _currentConfiguration;
+
+                    Interlocked.Exchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning);
```
keegan-caruso: Similar: since this in in a lock effectively, does this actually need an Interlocked.Exchange?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -279,22 +285,32 @@ private void UpdateCurrentConfiguration()
...
         private void UpdateConfiguration(T configuration)
         {
             _currentConfiguration = configuration;
-            _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval +
-                TimeSpan.FromSeconds(new Random().Next((int)AutomaticRefreshInterval.TotalSeconds / 20)));
+            // StartupTime is the time when ConfigurationManager was instantiated.
+            // (DateTimeOffset.UtcNow - StartupTime).TotalSeconds is the number of seconds since ConfigurationManager was instantiated.
+            // For automatic refresh, we add a 5% jit.
+            // Record in seconds when the last time configuration was obtained.
+            double secondsWhenRefreshOccurred = (DateTimeOffset.UtcNow - StartupTime).TotalSeconds +
```
keegan-caruso: (DateTimeOffset.UtcNow - StartupTime).TotalSeconds could be good to pull into a helper to give it a good name.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -279,22 +285,32 @@ private void UpdateCurrentConfiguration()
...
+        /// </summary>
+        /// <param name="configuration"></param>
         private void UpdateConfiguration(T configuration)
         {
             _currentConfiguration = configuration;
-            _syncAfter = DateTimeUtil.Add(DateTime.UtcNow, AutomaticRefreshInterval +
-                TimeSpan.FromSeconds(new Random().Next((int)AutomaticRefreshInterval.TotalSeconds / 20)));
+            // StartupTime is the time when ConfigurationManager was instantiated.
+            // (DateTimeOffset.UtcNow - StartupTime).TotalSeconds is the number of seconds since ConfigurationManager was instantiated.
+            // For automatic refresh, we add a 5% jit.
```
keegan-caruso:
```suggestion
            // For automatic refresh, we add a 5% jitter.
```

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: The name is a little confusing here. It feels odd to call it "when occurred" when we add the jitter here.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -279,22 +285,32 @@ private void UpdateCurrentConfiguration()
...
+            // Record in seconds when the last time configuration was obtained.
+            double secondsWhenRefreshOccurred = (DateTimeOffset.UtcNow - StartupTime).TotalSeconds +
+                                 ((_automaticRefreshIntervalInSeconds >= int.MaxValue) ? 0 : (new Random().Next((int)_automaticRefreshIntervalInSeconds / 20)));
```
keegan-caruso: We need to use a shared random and not create a new one each time here.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -279,22 +285,32 @@ private void UpdateCurrentConfiguration()
...
+            _timeInSecondsWhenLastAutomaticRefreshOccurred = (int)((secondsWhenRefreshOccurred <= int.MaxValue) ? (int)secondsWhenRefreshOccurred : int.MaxValue);
         }
 
         /// <summary>
         /// Obtains an updated version of Configuration.
         /// </summary>
         /// <param name="cancel">CancellationToken</param>
         /// <returns>Configuration of type BaseConfiguration    .</returns>
-        /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
+        /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager._automaticRefreshIntervalInSeconds"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
```
keegan-caruso, line 318: Use @msbw2 suggestion instead.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -309,15 +325,20 @@ public override async Task<BaseConfiguration> GetBaseConfigurationAsync(Cancella
...
-            if (now >= DateTimeUtil.Add(_lastRequestRefresh.UtcDateTime, RefreshInterval) || _isFirstRefreshRequest)
+            double nextRefresh = _requestRefreshIntervalInSeconds + _timeInSecondsWhenLastRequestRefreshOccurred;
+            if (nextRefresh < (DateTimeOffset.UtcNow - StartupTime).TotalSeconds || _isFirstRefreshRequest)
             {
                 _isFirstRefreshRequest = false;
                 if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
                 {
+                    NumberOfTimesRequestRefreshRequested++;
+                    double recordWhenRefreshOccurred = (DateTimeOffset.UtcNow - StartupTime).TotalSeconds;
+                    // transfer to int in single operation.
```
keegan-caruso, line 344: Another good one to pull into a helper

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -279,22 +285,32 @@ private void UpdateCurrentConfiguration()
             {
                 Interlocked.Exchange(ref _configurationRetrieverState, ConfigurationRetrieverIdle);
             }
-#pragma warning restore CA1031 // Do not catch general exception types
         }
 
+        /// <summary>
+        /// Called only when configuration is successfully obtained.
+        /// </summary>
+        /// <param name="configuration"></param>
```
keegan-caruso: empty tag

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -165,6 +189,7 @@ public TimeSpan RefreshInterval
                     throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(value), LogHelper.FormatInvariant(LogMessages.IDX10107, LogHelper.MarkAsNonPII(MinimumRefreshInterval), LogHelper.MarkAsNonPII(value))));
 
                 _refreshInterval = value;
```
keegan-caruso, line 226: Does this also need an Interlocked.Exchange?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -147,8 +145,13 @@ public async Task<T> GetConfigurationAsync()
         /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
         public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
         {
-            if (_currentConfiguration != null && _syncAfter > DateTimeOffset.UtcNow)
-                return _currentConfiguration;
+            if (_currentConfiguration != null)
+            {
+                // StartupTime is the time when ConfigurationManager was instantiated.
+                double nextRefresh = _automaticRefreshIntervalInSeconds + _timeInSecondsWhenLastAutomaticRefreshOccurred;
+                if (nextRefresh > (DateTimeOffset.UtcNow - StartupTime).TotalSeconds)
```
keegan-caruso: For all of the uses of UtcNow, can we use a TimeProvider?

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -35,9 +41,27 @@ public TimeSpan AutomaticRefreshInterval
...
+                Interlocked.Exchange(ref _automaticRefreshIntervalInSeconds, value.TotalSeconds);
             }
         }
 
+        /// <summary>
+        /// Records the time this instance was created.
+        /// Used to determine if the automatic refresh or request refresh interval has passed.
+        /// The set is used for testing purposes.
+        /// </summary>
+        internal DateTimeOffset StartupTime { get; set; } = DateTimeOffset.UtcNow;
```
keegan-caruso, line 66: I think it would help to be more explicit in spelling out the algorithm and how things end up relative to this.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -207,13 +211,14 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
                 finally
                 {
                     _configurationNullLock.Release();
+                    Interlocked.Exchange(ref _configurationRetrieverState, ConfigurationRetrieverIdle);
```
keegan-caruso: I'm not convinced this needs to be an Interlocked.Exchange either. A write to an int is atomic, and we don't need to write and also know the original value in one operation.

test/Microsoft.IdentityModel.TestUtils/EventDrivenConfigurationManager.cs
```diff
@@ -0,0 +1,116 @@
...
+    public class EventDrivenConfigurationRetriever<T> : IConfigurationRetriever<T>
+    {
+        private ManualResetEvent _signalEvent;
+        private ManualResetEvent _waitEvent;
+        private T _configuration;
+
+        /// <summary>
+        /// Initializes an new instance of <see cref="MockConfigurationManager{T}"/> with a Configuration instance.
+        /// </summary>
+        /// <param name="configuration"></param>
```
keegan-caruso: fill out or remove these three

test/Microsoft.IdentityModel.TestUtils/EventDrivenConfigurationManager.cs
```diff
@@ -0,0 +1,116 @@
...
+    {
+        private ManualResetEvent _configSignalEvent;
+        private ManualResetEvent _configWaitEvent;
+        private ManualResetEvent _refreshSignalEvent;
+        private ManualResetEvent _refreshWaitEvent;
+
+        /// <summary>
+        /// Initializes an new instance of <see cref="EventControlledConfigurationManger{T}"/> with a Configuration instance.
+        /// </summary>
+        /// <param name="metadataAddress"></param>
```
keegan-caruso: fill out or remove the param tags

keegan-caruso (commented):
I need a little more time with this, but this is my first pass.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
msbw2: Fixed the suggestion:
```suggestion
        /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager._automaticRefreshIntervalInSeconds"/> then
        /// <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
```

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -157,20 +160,19 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
-                    _configurationNullLock.Release();
-                    return _currentConfiguration;
-                }
-
-#pragma warning disable CA1031 // Do not catch general exception types
                 try
                 {
+                    await _configurationNullLock.WaitAsync(cancel).ConfigureAwait(false);
+                    if (_currentConfiguration != null)
+                        return _currentConfiguration;
```
msbw2: Do you need to release the lock as before?

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -35,9 +41,27 @@ public TimeSpan AutomaticRefreshInterval
...
+
+        /// <summary>
+        /// Each time GetConfigurationAsync() results in a http request, this value is incremented.
+        /// </summary>
+        internal long NumberOfTimesAutomaticRefreshRequested { get; set; }
+
+        /// <summary>
+        /// Each time RequestRefresh() results in a http request, this value is incremented.
```
msbw2: However Visual Studio generates it:
```suggestion
        /// Each time <see cref="RequestRefresh" /> results in a http request, this value is incremented.
```

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -35,9 +41,27 @@ public TimeSpan AutomaticRefreshInterval
...
 
+        /// <summary>
+        /// Records the time this instance was created.
+        /// Used to determine if the automatic refresh or request refresh interval has passed.
+        /// The set is used for testing purposes.
+        /// </summary>
+        internal DateTimeOffset StartupTime { get; set; } = DateTimeOffset.UtcNow;
+
+        /// <summary>
+        /// Each time GetConfigurationAsync() results in a http request, this value is incremented.
```
msbw2: There is no `GetConfigurationAsync` in this class. It's only in a subclass.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
msbw2, line 66: If you use a `TimeProvider` you can get rid of the setter.
```suggestion
        internal DateTimeOffset StartupTime { get; } = DateTimeOffset.UtcNow;
```

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
msbw2: Never mind. I see it in the `finally` block.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
msbw2: `int.MaxValue` seconds is nearly 600000 hours. This doesn't seem like a reasonable value for `_automaticRefreshIntervalInSeconds`, so I think when it's set there could be a check like this that rejects unrealistically large values. Then you don't need this check.

Also, you can precompute `_automaticRefreshIntervalInSeconds / 20` in the setter and store it in a value called something like `_maxJitter`.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -279,22 +285,32 @@ private void UpdateCurrentConfiguration()
...
-                TimeSpan.FromSeconds(new Random().Next((int)AutomaticRefreshInterval.TotalSeconds / 20)));
+            // StartupTime is the time when ConfigurationManager was instantiated.
+            // (DateTimeOffset.UtcNow - StartupTime).TotalSeconds is the number of seconds since ConfigurationManager was instantiated.
+            // For automatic refresh, we add a 5% jit.
+            // Record in seconds when the last time configuration was obtained.
+            double secondsWhenRefreshOccurred = (DateTimeOffset.UtcNow - StartupTime).TotalSeconds +
+                                 ((_automaticRefreshIntervalInSeconds >= int.MaxValue) ? 0 : (new Random().Next((int)_automaticRefreshIntervalInSeconds / 20)));
+
+            // transfer to int in single operation.
+            _timeInSecondsWhenLastAutomaticRefreshOccurred = (int)((secondsWhenRefreshOccurred <= int.MaxValue) ? (int)secondsWhenRefreshOccurred : int.MaxValue);
```
msbw2: If it manages to get this high (almost 600000 hours continuous), I almost think it should be an error condition. This is currently hidden and will be difficult to debug if somehow it's ever hit.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -309,15 +325,20 @@ public override async Task<BaseConfiguration> GetBaseConfigurationAsync(Cancella
         /// </summary>
```
msbw2, line 330: The doc above needs to be updated.

Also, can we add a check `RefreshInterval == TimeSpan.MaxValue` to do an early return?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
msbw2: That's also my understanding. At a single point in time there can only be one writer, but since this is exposed externally for telemetry purposes there could be a reader that reads this value while it's being written. Since this is a `long` the write is not guaranteed to be atomic so it could report incorrect values.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -35,9 +41,27 @@ public TimeSpan AutomaticRefreshInterval
...
+        /// </summary>
+        internal long NumberOfTimesAutomaticRefreshRequested { get; set; }
```
msbw2: `long` is not guaranteed to be atomic during reads and writes, so if this is being exposed to be consumed externally there may be a threading issue.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
msbw2: This is analogous to the seconds since the Unix epoch, so I think mentioning that would help (`BaseConfigurationManager` epoch).

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: Good call out. For the calling code it should do a Interlocked.Read, which we get for int64/uint64

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
keegan-caruso: Telemetry code can use Interlocked.Read

brentschmaltz: Good catch.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: It is okay to not do it right now, but let's at least have a tracking issue linked here.

keegan-caruso (approved):
I want to make sure @pmaytak gets a chance to look at this, but this is looking good.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
msbw2: > Telemetry code can use Interlocked.Read

Sure, but it needs to be documented. Right now, there's no indication of this other than it referencing a method with the `Async` suffix which requires the user to infer that a lock is required on read.

keegan-caruso: good point @msbw2 , unresolving, specifically asking for an update to the doc string.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -35,9 +44,31 @@ public TimeSpan AutomaticRefreshInterval
                     throw LogHelper.LogExceptionMessage(new ArgumentOutOfRangeException(nameof(value), LogHelper.FormatInvariant(LogMessages.IDX10108, LogHelper.MarkAsNonPII(MinimumAutomaticRefreshInterval), LogHelper.MarkAsNonPII(value))));
 
                 _automaticRefreshInterval = value;
+                Interlocked.Exchange(ref _automaticRefreshIntervalInSeconds, value.TotalSeconds);
+                Interlocked.Exchange(ref _maxJitter, value.TotalSeconds / 20);
             }
         }
 
+        /// <summary>
+        /// Records the time this instance was created.
```
pmaytak, line 59:
```suggestion
        /// Records the time when this ConfigurationManager instance was created.
```

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -35,9 +44,31 @@ public TimeSpan AutomaticRefreshInterval
...
+        /// </summary>
+        internal DateTimeOffset StartupTime { get; set; } = DateTimeOffset.UtcNow;
+
+        /// <summary>
+        /// Each call to <see cref="GetBaseConfigurationAsync(CancellationToken)"/> results in a http request, this value is incremented.
+        /// </summary>
+        internal long NumberOfTimesAutomaticRefreshRequested { get; set; }
+
+        /// <summary>
+        /// Each call to <see cref="RequestRefresh"/> results in a http request, this value is incremented.
```
pmaytak:
```suggestion
        /// Incremented each time <see cref="RequestRefresh"/> tries to refresh configuration. Does not necessarily equal to the count of successful refresh requests.
```

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -35,9 +44,31 @@ public TimeSpan AutomaticRefreshInterval
...
+        /// Used to determine if the automatic refresh or request refresh interval has passed.
+        /// The logic is to remember the number of seconds since startup that the last refresh occurred.
+        /// Store that value in _timeInSecondsWhenLastAutomaticRefreshOccurred or _timeInSecondsWhenLastRequestRefreshOccurred.
+        /// Then compare to (UtcNow - Startup).TotalSeconds.
+        /// The set is used for testing purposes.
+        /// </summary>
+        internal DateTimeOffset StartupTime { get; set; } = DateTimeOffset.UtcNow;
+
+        /// <summary>
+        /// Each call to <see cref="GetBaseConfigurationAsync(CancellationToken)"/> results in a http request, this value is incremented.
```
pmaytak:
```suggestion
        /// Incremented each time <see cref="GetBaseConfigurationAsync(CancellationToken)"/> tries to refresh configuration. Does not necessarily equal to the count of successful refresh requests.
```

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
msbw2: Also, I think the writer (`x++`) needs to lock this on write so that the reader doesn't lock in between write operations.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -16,13 +16,22 @@ namespace Microsoft.IdentityModel.Tokens
...
         private TimeSpan _refreshInterval = DefaultRefreshInterval;
+        internal double _requestRefreshIntervalInSeconds = DefaultRefreshInterval.TotalSeconds;
         private TimeSpan _lastKnownGoodLifetime = DefaultLastKnownGoodConfigurationLifetime;
         private BaseConfiguration _lastKnownGoodConfiguration;
         private DateTime? _lastKnownGoodConfigFirstUse;
+        internal Random _random = new();
+
+        // Seconds since the BaseConfigurationManager was created when the last refresh occurred.
+        internal int _timeInSecondsWhenLastAutomaticRefreshOccurred;
+        internal int _timeInSecondsWhenLastRequestRefreshOccurred;
```
pmaytak: nit: I think `_timeInSecondsWhenLastRequestRefreshOccurred` needs to be renamed to something like `_timeInSecondsWhenLastRequestRefreshWasRequested` because it is set in RequestRefresh before the task is run, which may end up failing to refresh the config.

`_timeInSecondsWhenLastAutomaticRefreshOccurred` should be renamed to something more general like `_timeInSecondsWhenLastRefreshOccurred` because it is called from UpdateConfiguration (where _currentConfiguration variable is updated) which is called from both, GetConfigurationAsync and UpdateCurrentConfiguration, which is also called from RequestRefresh.

So `_timeInSecondsWhenLastAutomaticRefreshOccurred` currently represents when the config was updated from both, RequestRefresh and GetConfigurationAsync.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -157,20 +162,19 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
-                if (_currentConfiguration != null)
-                {
-                    _configurationNullLock.Release();
-                    return _currentConfiguration;
-                }
-
-#pragma warning disable CA1031 // Do not catch general exception types
                 try
                 {
+                    await _configurationNullLock.WaitAsync(cancel).ConfigureAwait(false);
```
pmaytak: This is a bug. WaitAsync needs to be outside of try.

See example that happened in MSAL: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4472

> The semaphore.Wait is called inside a try. When the cancellation token is cancelled, a TaskCancelledException is thrown without the semaphore being acquired, and finally is called trying to release the semaphore which was not acquired causing a SemaphoreFullException.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -157,20 +162,19 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+                    if (_currentConfiguration != null)
+                        return _currentConfiguration;
+
+                    _configurationRetrieverState = ConfigurationRetrieverRunning;
```
pmaytak, line 178: Good catch to set the state.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -206,14 +212,15 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
                     _configurationNullLock.Release();
                 }
-#pragma warning restore CA1031 // Do not catch general exception types
             }
             else
             {
                 if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
                 {
+                    NumberOfTimesAutomaticRefreshRequested++;
                     _ = Task.Run(UpdateCurrentConfiguration, CancellationToken.None);
```
pmaytak: Can we please rename `UpdateCurrentConfiguration` to `RetrieveAndUpdateConfiguration`? So that it's more distinct from `UpdateConfiguration` which updates the config instance variable.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -16,13 +16,22 @@ namespace Microsoft.IdentityModel.Tokens
...
+        internal double _maxJitter = DefaultAutomaticRefreshInterval.TotalSeconds / 20;
         private TimeSpan _refreshInterval = DefaultRefreshInterval;
+        internal double _requestRefreshIntervalInSeconds = DefaultRefreshInterval.TotalSeconds;
         private TimeSpan _lastKnownGoodLifetime = DefaultLastKnownGoodConfigurationLifetime;
         private BaseConfiguration _lastKnownGoodConfiguration;
         private DateTime? _lastKnownGoodConfigFirstUse;
+        internal Random _random = new();
+
+        // Seconds since the BaseConfigurationManager was created when the last refresh occurred.
+        internal int _timeInSecondsWhenLastAutomaticRefreshOccurred;
```
pmaytak: Add a comment that this value isn't accurate, it may include 5% jitter.

pmaytak (changes requested):

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
msbw2: This is in `BaseConfigurationManager` so `s/ConfigurationManager/BaseConfigurationManager/`.

pmaytak:
Just dropping links for related reading:
https://stackoverflow.com/questions/9666/is-accessing-a-variable-in-c-sharp-an-atomic-operation
https://www.ecma-international.org/wp-content/uploads/ECMA-335_6th_edition_june_2012.pdf

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
pmaytak: I suppose it's not an issue now because we don't specify max semaphore count, but if we do specify the max count later, this will result in a bug.

msbw2: Nonetheless it shouldn't release something it doesn't hold.

pmaytak: Thinking about it more, I think we need to specify a max count of 1 - `SemaphoreSlim(1, 1)`?

Thinking of this scenario:
- Initial semaphore count is 1.
- Thread 1 comes in, takes out a semaphore (semaphore count is 0).
- Thread 2 comes in, waits because semaphore count is 0.
- Thread 1 completes succesfully.
- Thread 1 calls Release in finally (semaphore count is 1).
- Thread 2 moves forward, its cancellation token is expired, so TaskCancelledException is thrown.
  - Semaphore is not taken (semaphore count is still 1).
- Thread 2 goes to catch, then calls release in finally (semaphore count is 2).
- Thread 3 comes in, calls Wait (semaphore count is 1).
- Thread 4 comes in, calls Wait (semaphore count is 0).

brentschmaltz: @pmaytak are you suggesting a race condition where the cancellation token expires while waiting, then the semaphore is not taken?

pmaytak: Yes.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -18,8 +18,6 @@ namespace Microsoft.IdentityModel.Protocols
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable")]
     public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class
     {
-        private DateTimeOffset _syncAfter = DateTimeOffset.MinValue;
-        private DateTimeOffset _lastRequestRefresh = DateTimeOffset.MinValue;
         private bool _isFirstRefreshRequest = true;
         private readonly SemaphoreSlim _configurationNullLock = new SemaphoreSlim(1);
```
keegan-caruso, line 22:
```suggestion
        private readonly SemaphoreSlim _configurationNullLock = new SemaphoreSlim(1, 1);
```

Echoing @pmaytak feedback

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: @pmaytak could you do something like:

```cs
catch (Exception ex) when (ex is not TaskCanceledException)
```

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
keegan-caruso: You can do an interlocked increment on the long as well I think to avoid the lock.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -157,20 +162,19 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
             // else
-            //   if task is running, return the current configuration
+            //   if update task is running, return the current configuration
             //   else kick off task to update current configuration
             if (_currentConfiguration == null)
             {
                 await _configurationNullLock.WaitAsync(cancel).ConfigureAwait(false);
                 if (_currentConfiguration != null)
-                {
-                    _configurationNullLock.Release();
```
msbw2: This still needs to release now that it's outside of the `try`-`finally`.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -309,19 +327,33 @@ public override async Task<BaseConfiguration> GetBaseConfigurationAsync(Cancella
...
-                    _lastRequestRefresh = now;
                 }
             }
         }
 
+        /// <summary>
+        /// SecondsSinceInstanceWasCreated is the number of seconds since ConfigurationManager was instantiated.
+        /// </summary>
+        /// <returns>double</returns>
+        private double SecondsSinceInstanceWasCreated()
```
msbw2: Can this be renamed by adding the prefix `Get` or converted to a property?
```suggestion
        private double GetSecondsSinceInstanceWasCreated()
```
```suggestion
        private double SecondsSinceInstanceWasCreated => (DateTimeOffset.UtcNow - StartupTime).TotalSeconds;
```

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -16,13 +16,22 @@ namespace Microsoft.IdentityModel.Tokens
...
     {
         private TimeSpan _automaticRefreshInterval = DefaultAutomaticRefreshInterval;
+        internal double _automaticRefreshIntervalInSeconds = DefaultAutomaticRefreshInterval.TotalSeconds;
+        internal double _maxJitter = DefaultAutomaticRefreshInterval.TotalSeconds / 20;
         private TimeSpan _refreshInterval = DefaultRefreshInterval;
+        internal double _requestRefreshIntervalInSeconds = DefaultRefreshInterval.TotalSeconds;
         private TimeSpan _lastKnownGoodLifetime = DefaultLastKnownGoodConfigurationLifetime;
         private BaseConfiguration _lastKnownGoodConfiguration;
         private DateTime? _lastKnownGoodConfigFirstUse;
+        internal Random _random = new();
```
pmaytak, line 26: Can be `private` and `static`?

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -16,13 +16,22 @@ namespace Microsoft.IdentityModel.Tokens
...
+
+        // Seconds since the BaseConfigurationManager was created when the last refresh occurred with a %5 random jitter.
+        internal int _timeInSecondsWhenLastRefreshOccurred;
+        internal int _timeInSecondsWhenLastRequestRefreshWasRequested;
```
pmaytak, line 30: Internal variables should be cased as `TimeInSecondsWhenLastRefreshOccurred`, `TimeInSecondsWhenLastRequestRefreshWasRequested`.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -16,13 +16,22 @@ namespace Microsoft.IdentityModel.Tokens
     public abstract class BaseConfigurationManager
     {
         private TimeSpan _automaticRefreshInterval = DefaultAutomaticRefreshInterval;
+        internal double _automaticRefreshIntervalInSeconds = DefaultAutomaticRefreshInterval.TotalSeconds;
+        internal double _maxJitter = DefaultAutomaticRefreshInterval.TotalSeconds / 20;
         private TimeSpan _refreshInterval = DefaultRefreshInterval;
+        internal double _requestRefreshIntervalInSeconds = DefaultRefreshInterval.TotalSeconds;
```
pmaytak, line 22: Internal variables should be cased as `AutomaticRefreshIntervalInSeconds`, `MaxJitter`, `RequestRefreshIntervalInSeconds`.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -16,13 +16,22 @@ namespace Microsoft.IdentityModel.Tokens
...
         private TimeSpan _lastKnownGoodLifetime = DefaultLastKnownGoodConfigurationLifetime;
         private BaseConfiguration _lastKnownGoodConfiguration;
         private DateTime? _lastKnownGoodConfigFirstUse;
+        internal Random _random = new();
+
+        // Seconds since the BaseConfigurationManager was created when the last refresh occurred with a %5 random jitter.
```
pmaytak, line 28:
```suggestion
        // Seconds since the BaseConfigurationManager was created when the last refresh occurred with a 5% random jitter.
```

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -32,9 +41,59 @@ public TimeSpan AutomaticRefreshInterval
...
+        internal long NumberOfTimesAutomaticRefreshRequested
+        {
+            get
+            {
+                return Interlocked.Read(ref _numberOfTimesAutomaticRefreshRequested);
+            }
+        }
+
+        /// <summary>
+        /// Incremented each time <see cref="RequestRefresh"/> results in a http request.
```
pmaytak, line 85:
```suggestion
        /// Incremented each time <see cref="RequestRefresh"/> results in a HTTP request.
```

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -32,9 +41,59 @@ public TimeSpan AutomaticRefreshInterval
...
+        /// Used to determine if the automatic refresh or request refresh interval has passed.
+        /// The logic is to remember the number of seconds since startup that the last refresh occurred.
+        /// Store that value in _timeInSecondsWhenLastAutomaticRefreshOccurred or _timeInSecondsWhenLastRequestRefreshOccurred.
+        /// Then compare to (UtcNow - Startup).TotalSeconds.
+        /// The set is used for testing purposes.
+        /// </summary>
+        internal DateTimeOffset StartupTime { get; set; } = DateTimeOffset.UtcNow;
+
+        /// <summary>
+        ///  Incremented each time <see cref="GetBaseConfigurationAsync(CancellationToken)"/> results in a http request.
```
pmaytak, line 69:
```suggestion
        ///  Incremented each time <see cref="GetBaseConfigurationAsync(CancellationToken)"/> results in a HTTP request.
```

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
```diff
@@ -32,9 +41,59 @@ public TimeSpan AutomaticRefreshInterval
...
+        /// </summary>
+        internal long _numberOfTimesAutomaticRefreshRequested;
```
pmaytak, line 71: Currently we have to remember to use Interlocked when incrementing the variable. Can this be made `private` and another method be added to increment this? Something like 
```csharp
internal void IncrementNumberOfTimesAutomaticRefreshRequested() => Interlocked.Increment(ref NumberOfTimesAutomaticRefreshRequested);`
```
This way we for sure will use Interlocked when incrementing and it won't be possible to set this variable to some other value. 

Same comment for `_numberOfTimesRequestRefreshRequested`.

msbw2: Likewise, can we expose methods/properties for accessing these values externally? Something like this (variables are probably wrong but this general idea):
```cs
public long NumberOfTimesAutomaticRefreshRequested => Interlocked.Read(ref _numberOfTimesAutomaticRefreshRequested);
```

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
keegan-caruso: private but not static, older frameworks won't be thread safe, so a shared static isn't thread safe.

src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs
keegan-caruso: we would need to update the editor config: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/.editorconfig#L71

keegan-caruso:
closing in favor of #3090

PR #2991, opened 2024-11-09
Use SecurityTokenDescriptor when creating tokens

### LINKED ISSUE #2993: Adding new parameters to APIs for creating tokens results in parameter explosion
Any time a new parameter needs to be added to configure token creation various APIs need to be updated. These APIs have a lot of parameters already, and the `SecurityTokenDescriptor` class already exists to avoid this. Update all APIs so that internally they're using this class so that eventually the APIs with lots of parameters can be deprecated and removed.

msbw2:
Use `SecurityTokenDescriptor` when creating tokens.


keegan-caruso:
Let's make sure there is an issue so we can track and include in release notes easier.

msbw2:
> Let's make sure there is an issue so we can track and include in release notes easier.

@keegan-caruso Updated.

armandRobled:
"Suscribe"

Luis Armando Robledo salazar

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
```diff
@@ -35,14 +35,7 @@ public virtual string CreateToken(string payload)
...
 
-            return CreateToken(
-                    payload,
-                    null,
-                    null,
-                    null,
-                    null,
-                    null,
-                    null);
+            return CreateToken(payload, new SecurityTokenDescriptor());
```
keegan-caruso: SecurityTokenDescriptor can be a static instance in this case to save the allocation.

keegan-caruso (approved):
LGTM, just that one alloc we can avoid.

westin-m (approved):

PR #2994, opened 2024-11-11
SAML and SAML2 new model validation: Token Replay

iNinja:
# SAML and SAML2 new model validation: Token Replay
- Adjusted token replay validation in the new validation model to match existing behaviour.

Part of #2711


jennyf19:
Looks like the tests are failing? is that expected?

iNinja:
> Looks like the tests are failing? is that expected?

Probably not 😄. Apparently when I merged the signature validation code, I mistakenly left the signature validation **after** the issuer signing key one.

This failure was then hidden by the issue we had with the Abstractions assemblies signature that was failing all builds.
It is fixed now.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateToken.Internal.cs
```diff
@@ -144,33 +154,21 @@ internal virtual ValidationResult<ValidatedConditions> ValidateConditions(
...
-                    callContext);
+                var oneTimeUseValidationError = ValidateOneTimeUseCondition(samlToken, validationParameters, callContext);
 
-                if (!oneTimeUseValidationResult.IsValid)
-                {
-                    StackFrames.OneTimeUseValidationFailed ??= new StackFrame(true);
-                    return oneTimeUseValidationResult.UnwrapError().AddStackFrame(StackFrames.OneTimeUseValidationFailed);
-                }
+                if (oneTimeUseValidationError is not null)
+                    return oneTimeUseValidationError.AddCurrentStackFrame();
```
JoshLozensky, line 160: Why do we return the error instead of throw it?

JoshLozensky (approved):

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ValidateToken.Internal.cs
iNinja: This is the core of the work we are doing. Enhancing performance by removing exception throwing and returning validation errors that can be logged, thrown, or just checked on the above layers.

I think we have a deep dive coming on the topic, but basically by not throwing exceptions in some cases were are seeing over 30% performance improvement in our validation flows.

westin-m (approved):

PR #2997, opened 2024-11-12
SAML and SAML2 Extensibility Testing: Issuer

iNinja:
# SAML and SAML2 Extensibility Testing: Issuer
- Adjusted JWT's issuer extensibility work to simplify the creation of the stack frames for comparison, and to assert that the stack traces are present in the generated exceptions, and they match the expectations.
- Added exception handling for the issuer validation delegates on SAML and SAML2, in case an override were to throw.
- Added extensibility tests for issuer validation to SAML and SAML2.

Note: This work is based on https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2987 with some adjustments made to the work agreed on there.

Part of #2711


iNinja:
Re-issuing extensibility PRs with better alignment.

PR #3022, opened 2024-11-20
Add request count and duration telemetry

### LINKED ISSUE #3023: Add metadata request time to client telemetry
See [Product Backlog Item 3071909](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3071909): Add metadata request time to client telemetry

westin-m:
Addresses client telemetry repair item

Add counter for get configuration requests and histogram for total duration of the same requests.


src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -23,6 +24,27 @@ public static class IdentityModelTelemetryUtil
...
+        internal const string OperationStatusTag = "OperationStatus";
+        internal const string ExceptionTypeTag = "ExceptionType";
+
+        // Validation result values
+        internal const string Success = "Success";
+        internal const string Failure = "Failure";
+
+        // Configuration manager refresh reasons
+        internal const string Requested = "Requested";
+        internal const string Scheduled = "Scheduled";
```
westin-m: This was a suggested value for one of the dimensions, but I'm not yet sure how to discern the discern reason for the refresh

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -147,8 +147,14 @@ public async Task<T> GetConfigurationAsync()
         /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
         public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
         {
+            var methodStartTicks = WatchUtility.Watch.Elapsed.Ticks;
+
             if (_currentConfiguration != null && _syncAfter > DateTimeOffset.UtcNow)
+            {
+                TrackGetConfiguration(
```
brentschmaltz: What are we measuring here?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -164,6 +170,8 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
                 await _configurationNullLock.WaitAsync(cancel).ConfigureAwait(false);
                 if (_currentConfiguration != null)
                 {
+                    TrackGetConfiguration(
```
brentschmaltz: What is being measured here?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -220,8 +232,14 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
 
             // If metadata exists return it.
             if (_currentConfiguration != null)
+            {
+                TrackGetConfiguration(
+                    methodStartTicks, IdentityModelTelemetryUtil.Requested, IdentityModelTelemetryUtil.Success, string.Empty);
```
brentschmaltz: The logic that got us here does not necessarily imply that valid metadata was received.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -322,6 +340,19 @@ public override void RequestRefresh()
             }
         }
 
+        internal void TrackGetConfiguration(long methodStartTicks, string refreshReason, string operationStatus, string exceptionType)
+        {
+            IdentityModelTelemetryUtil.IncrementConfigurationManagerCounter(
+                    MetadataAddress, refreshReason, operationStatus, exceptionType);
+
+            IdentityModelTelemetryUtil.RecordTotalDuration(
+                (WatchUtility.Watch.Elapsed.Ticks - methodStartTicks) / TimeSpan.TicksPerMillisecond,
```
brentschmaltz: Instead of performing a division each time, could we define the telemetry to send details in ticks and let the receiver perform any necessary calculations.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -322,6 +340,19 @@ public override void RequestRefresh()
             }
         }
 
+        internal void TrackGetConfiguration(long methodStartTicks, string refreshReason, string operationStatus, string exceptionType)
```
brentschmaltz: We should be tracking that a RequestRefresh was requested.

build/dependenciesTest.props
```diff
@@ -7,6 +7,7 @@
     <MicrosoftNETTestSdkVersion>17.11.1</MicrosoftNETTestSdkVersion>
     <NetStandardVersion>2.0.3</NetStandardVersion>
     <NewtonsoftVersion>13.0.3</NewtonsoftVersion>
+    <OpenTelemetryVersion>1.10.0</OpenTelemetryVersion>
```
GeoK: I'd suggest starting with 1.6.0 that is used in other SDKs.
1.10.0 is still very new.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -135,5 +157,42 @@ internal static bool UpdateDefaultTelemetryData(string key, string value)
...
+            string operationStatus,
+            string exceptionType)
+        {
+            var tagList = new TagList()
+            {
+                { WilsonVersionTag, ClientVer },
+                { MetadataAddressTag, metadataAddress },
+                { RefreshReasonTag, refreshReason},
+                { OperationStatusTag, operationStatus },
+                { ExceptionTypeTag, exceptionType }
```
GeoK: Should we have two tagLists (one without an exception if empty/null)?

build/dependencies.props
```diff
@@ -8,6 +8,7 @@
     <MicrosoftSourceLinkGitHubVersion>1.0.0</MicrosoftSourceLinkGitHubVersion>
     <NetStandardVersion>2.0.3</NetStandardVersion>
     <NewtonsoftVersion>13.0.3</NewtonsoftVersion>
+    <SystemDiagnosticSourceVersion>7.0.2</SystemDiagnosticSourceVersion>
```
keegan-caruso: 7.0.2 because it is the lowest version with a 462 target?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetry.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <summary>
+        /// Meter name for MicrosoftIdentityModel.
+        /// </summary>
+        public const string MeterName = "MicrosoftIdentityModel_Meter";
+        public const string ServiceName = "MicrosoftIdentityModel";
+
+        /// <summary>
+        /// The meter responsible for creating instruments.
+        /// </summary>
+        public static readonly Meter WilsonMeter = new(MeterName, "1.0.0");
```
keegan-caruso: We should use "IdentityModel" over "Wilson" throughout all of this. "Wilson" confuses people.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -6,6 +6,7 @@
 using System.Collections.Concurrent;
 using System.Reflection;
 using System.Net.Http;
+using System.Diagnostics;
```
keegan-caruso: sort

src/Microsoft.IdentityModel.Logging/WatchUtility.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Diagnostics;
+
+namespace Microsoft.IdentityModel.Logging
+{
+    /// <summary>
+    /// A utility class to measure time.
+    /// </summary>
+    internal class WatchUtility
+    {
+        internal static readonly Stopwatch Watch = Stopwatch.StartNew();
```
keegan-caruso: This isn't the way we want to do this, we shouldn't have a static StopWatch always running.

Use a TimeProvider to get the time at the start and end of operations, ~you can also measure in milliseconds directly and not ticks.~ The time provider has an API to get a TimeSpan.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -183,11 +191,15 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
                         ConfigurationValidationResult result = _configValidator.Validate(configuration);
                         // in this case we have never had a valid configuration, so we will throw an exception if the validation fails
                         if (!result.Succeeded)
+                        {
+                            TrackGetConfiguration(
```
keegan-caruso: Stack parameters for readability.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTelemetryTests.cs
```diff
@@ -0,0 +1,200 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+//using System;
```
keegan-caruso:
```suggestion
```

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTelemetryTests.cs
```diff
@@ -0,0 +1,200 @@
...
+
+        [Theory, MemberData(nameof(GetConfiguration_ExpectedTagList_TheoryData), DisableDiscoveryEnumeration = true)]
+        public async Task GetConfigurationAsync_ExpectedTagList(ConfigurationManagerTelemetryTheoryData<OpenIdConnectConfiguration> theoryData)
+        {
+            var configurationManager = new ConfigurationManager<OpenIdConnectConfiguration>(
+                theoryData.MetadataAddress,
+                new OpenIdConnectConfigurationRetriever(),
+                theoryData.DocumentRetriever,
+                theoryData.ConfigurationValidator);
+            try
```
keegan-caruso, line 77:
```suggestion

            try
```

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: I'd prefer this to have two methods, one with exceptionType, one without. It is confusing above to pass string.empty most of the time.

src/Microsoft.IdentityModel.Logging/WatchUtility.cs
westin-m: Would modifying the use so that it's only running during the interval being measured still be too costly?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: The TimeProvider class has methods to directly get a TimeSpan which would clean this up

build/dependencies.props
westin-m: I meant to have the latest here. fixed

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetryUtil.cs
```diff
@@ -135,5 +135,52 @@ internal static bool UpdateDefaultTelemetryData(string key, string value)
             telemetryData[key] = value;
             return true;
         }
+
+        //internal static void IncrementConfigurationManagerCounter(
```
brentschmaltz: Why are these commented out?

westin-m: I moved them elsewhere and didn't clean up that file yet

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.cs
```diff
@@ -511,6 +513,10 @@ await ValidateJWEAsync(jsonWebToken, validationParameters, currentConfiguration)
                     // where a new valid configuration was somehow published during validation time.
                     if (currentConfiguration != null)
                     {
+                        _telemetryClient.IncrementOperationCounter(
```
brentschmaltz: Consider hiding the implementation inside Microsoft.IdentityModel.Logging.Logger.
Instead of each class having a telemetryClient, expose a method on Logger something like: Logger.IncrementOperationCounter(TelemetryConstants.LKG).

In this way when we expand telemetry into other classes, the adoption pattern is simpler.

src/Microsoft.IdentityModel.Logging/ConfigurationManagerTelemetryInstrumentation.cs
```diff
@@ -0,0 +1,54 @@
...
+                { TelemetryConstants.OperationStatusTag, operationStatus },
+                { TelemetryConstants.ExceptionTypeTag, exceptionType }
+            };
+
+            IdentityModelTelemetry.IncrementConfigurationManagerCounter(tagList);
+        }
+
+        public void LogOperationDuration(string clientVer, long durationInMilliseconds)
+        {
+            var tagList = new TagList()
```
keegan-caruso: This one can be static

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -322,6 +351,36 @@ public override void RequestRefresh()
...
         }
 
+        internal static void TrackRetrievalDuration(TimeSpan duration)
+        {
+            _telemetryClient.LogOperationDuration(
+                IdentityModelTelemetryUtil.ClientVer,
+                (long)duration.TotalMilliseconds);
+        }
+
+        internal static void TrackRetrievalDuration(TimeSpan duration, string exception)
```
keegan-caruso: should this be exception => exceptionType, same as a few lines down?

src/Microsoft.IdentityModel.Logging/TelemetryConstants.cs
```diff
@@ -0,0 +1,24 @@
...
+    internal static class TelemetryConstants
+    {
+        // Static attribute tags
+        public const string IdentityModelVersionTag = "IdentityModelVersion";
+        public const string MetadataAddressTag = "MetadataAddress";
+        public const string OperationStatusTag = "OperationStatus";
+        public const string ExceptionTypeTag = "ExceptionType";
+
+        // Configuration manager refresh statuses
+        public const string Automatic = "Automatic";
```
brentschmaltz: Does it matter how long these strings are?
The reason I ask is because this in a sense breaks layering.

M.IM.Logging, which sits below M.IM.Protocols knows details about M.IM.Protocols.
Ideally these constants should be declared in the assembly that uses them.
However, then we run into the possibility of a collision, which are usually resolved by using the fully qualified name, in this case would be "Microsoft.IdentityModel.Protocols.Automatic".

As we add additional telemetry, we will always need to modify Logging + the other assembly, would like to avoid that.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
westin-m: Yes, I will fix that. Thanks!

src/Microsoft.IdentityModel.Logging/TelemetryConstants.cs
westin-m: I'm not sure. My thought is that shorter may be better for telemetry

src/Microsoft.IdentityModel.Logging/ConfigurationManagerTelemetryInstrumentation.cs
westin-m: My understanding is that it cannot implement the interface member if it's static

brentschmaltz (changes requested):
Asking for layering changes.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -36,6 +36,8 @@ public class JwtSecurityTokenHandler : SecurityTokenHandler
         private static string _shortClaimType = _namespace + "/ShortTypeName";
         private bool _mapInboundClaims = DefaultMapInboundClaims;
 
+        internal ITelemetryInstrumentation _telemetryClient = new ConfigurationManagerTelemetryInstrumentation();
```
GeoK: This looks strange. Can we put all telemetry methods under a common interface / default implementation that is not class-dependent?

westin-m: Good call. Renamed to `TelemetryInstrumentation`

build/dependencies.props
```diff
@@ -8,6 +8,7 @@
     <MicrosoftSourceLinkGitHubVersion>1.0.0</MicrosoftSourceLinkGitHubVersion>
     <NetStandardVersion>2.0.3</NetStandardVersion>
     <NewtonsoftVersion>13.0.3</NewtonsoftVersion>
+    <SystemDiagnosticSourceVersion>9.0.0</SystemDiagnosticSourceVersion>
```
keegan-caruso: We should think about if we want to use 9.x versions or not. 9.x libraries are going to be standard term support, vs 8.x libraries are going to be lts.

src/Microsoft.IdentityModel.Logging/TelemetryInstrumentation.cs
```diff
@@ -0,0 +1,54 @@
...
+                { TelemetryConstants.OperationStatusTag, operationStatus },
+                { TelemetryConstants.ExceptionTypeTag, exceptionType }
+            };
+
+            IdentityModelTelemetry.IncrementOperationCounter(tagList);
+        }
+
+        public void LogOperationDuration(long durationInMilliseconds)
+        {
+            var tagList = new TagList()
```
keegan-caruso: Same comment, this doesn't need to be created every time. It doesn't change.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -240,13 +260,18 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+
             try
             {
                 T configuration = _configRetriever.GetConfigurationAsync(
                     MetadataAddress,
                     _docRetriever,
                     CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
 
+                long elapsedTime = (long)_timeProvider.GetElapsedTime(startTimestamp).TotalMilliseconds;
+                _telemetryClient.LogOperationDuration(elapsedTime);
```
keegan-caruso: Can the telemetry client take a timespan instead? It makes more sense for the telemetry client to be concerned with this.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -267,6 +292,11 @@ private void UpdateCurrentConfiguration()
             }
             catch (Exception ex)
             {
+                long elapsedTime = (long)_timeProvider.GetElapsedTime(startTimestamp).TotalMilliseconds;
+                _telemetryClient.LogOperationDuration(
```
keegan-caruso: The interface would be better if the telemetry client took the exception instead.

build/dependencies.props
westin-m: it's close, but it looks like 8.x versions will have a few more months of support and I bet we'll want them. 8.x it is! good point

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTelemetryTests.cs
```diff
@@ -0,0 +1,181 @@
...
+    public class ConfigurationManagerTelemetryTests
+    {
+        readonly List<Metric> ExportedItems;
+        readonly MeterProvider MeterProvider;
+
+        public ConfigurationManagerTelemetryTests()
+        {
+            ExportedItems = new List<Metric>();
+
+            MeterProvider = Sdk.CreateMeterProviderBuilder()
```
GeoK: Seems like MeterProvider is not needed anymore and we can remove references to OpenTelemetry.Exporter.InMemory/Console

src/Microsoft.IdentityModel.Logging/ITelemetryInstrumentation.cs
```diff
@@ -0,0 +1,24 @@
...
+    internal interface ITelemetryInstrumentation
+    {
+        internal void LogOperationDuration(
+            TimeSpan operationDuration);
+
+        internal void LogOperationDuration(
+            TimeSpan operationDuration,
+            Exception exception);
+
+        internal void IncrementOperationCounter(
```
GeoK: This API will not work anymore as soon as we start tracking something else. I'd advise renaming the methods to indicate what exactly is being tracked or perhaps adding a parameter that describes the operation.

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetry.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// Histogram to capture total duration of configuration manager operations in milliseconds.
+        /// </summary>
+        internal static readonly Histogram<long> TotalDurationHistogram = IdentityModelMeter.CreateHistogram<long>(
+            TotalDurationHistogramName,
+            unit: "ms",
+            description: "Configuration retrieval latency during configuration manager operations.");
+
+        internal static void RecordTotalDurationHistogram(long requestDurationInMs, in TagList tagList)
```
GeoK: Name of the method doesn't directly correspond to the operation being tracked. Consider a scenario where a new, unrelated histogram is introduced—what other changes would you need to make?

src/Microsoft.IdentityModel.Logging/InternalsVisibleTo.cs
```diff
@@ -1,8 +1,13 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("Microsoft.IdentityModel.JsonWebTokens, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
GeoK: Can we compile include default telemetry implementation to avoid friend assemblies?

src/Microsoft.IdentityModel.Logging/TelemetryConstants.cs
```diff
@@ -0,0 +1,24 @@
...
+    internal static class TelemetryConstants
+    {
+        // Static attribute tags
+        public const string IdentityModelVersionTag = "IdentityModelVersion";
+        public const string MetadataAddressTag = "MetadataAddress";
+        public const string OperationStatusTag = "OperationStatus";
+        public const string ExceptionTypeTag = "ExceptionType";
+
+        // Configuration manager refresh statuses
+        public const string Automatic = "Microsoft.IdentityModel.Protocols.Automatic";
```
GeoK: Why prefixing the values?

src/Microsoft.IdentityModel.Logging/TelemetryInstrumentation.cs
```diff
@@ -0,0 +1,57 @@
...
+            IdentityModelTelemetry.IncrementOperationCounter(tagList);
+        }
+
+        public void IncrementOperationCounter(string operationStatus, Exception exception)
+        {
+            var tagList = new TagList()
+            {
+                { TelemetryConstants.IdentityModelVersionTag, IdentityModelTelemetryUtil.ClientVer },
+                { TelemetryConstants.OperationStatusTag, operationStatus },
+                { TelemetryConstants.ExceptionTypeTag, exception.GetType().ToString() }
```
GeoK: Can any of these be null?

src/Microsoft.IdentityModel.Logging/IdentityModelTelemetry.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// The meter responsible for creating instruments.
+        /// </summary>
+        public static readonly Meter IdentityModelMeter = new(MeterName, "1.0.0");
+
+        internal const string TotalDurationHistogramName = "IdentityModelConfigurationRequestTotalDurationInMS";
+
+        /// <summary>
+        /// Counter to capture requests to configuration manager.
+        /// </summary>
+        internal const string IdentityModelConfigurationManagerCounterName = "IdentityModelConfigurationManagerCounter";
```
GeoK: "Counter" suffix is reduntant and can be removed.

src/Microsoft.IdentityModel.Logging/TelemetryConstants.cs
westin-m: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3022#discussion_r1871794159

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTelemetryTests.cs
westin-m: Good catch

src/Microsoft.IdentityModel.Logging/TelemetryConstants.cs
GeoK: Can we rather group these by namespace, e.g Automatic, Direct, etc. to be under TelemetryConstants.Protocols?
I don't see value in sending the long string that includes the namespace (Microsoft.IdentityModel.Protocols.Automatic).

src/Microsoft.IdentityModel.Logging/TelemetryInstrumentation.cs
westin-m: No, GetType() will at the very least return type `Object` (Except in very specific cases that do not apply here)

src/Microsoft.IdentityModel.Logging/InternalsVisibleTo.cs
westin-m: It looks like doing this would mean including the interface, the implementation, the constants, IdentityModelTelemetry, and adding them all to that assembly's api file.

GeoK: Adding new friend assemblies is definitely not a good ideal direction. Can we compile in these 3? They seem lightweight.

skblr2022:
Table identitymodelconfigurationmanager_counter
1> Could the field net_IdentityModelVersion be named as "net_WilsonVersion" to maintain parity with other tables?
2> Any wiki doc explaining the different values of net_OperationStatus? What does Direct / Automatic / FirstRefresh / LastKnownGood indicate?
3> Does this table capture both data for both successful and failed metadata requests? Ex: In the case of validation failure with msg "SignatureKeyNotFound" there would be an eSTS metadata request. So, wondering how to identify failure requests?
4> Why is the net_MetadataAddress add empty or null for some records?
 
Table identitymodelconfigurationrequesttotaldurationinms_histogram
1> I am inferring from the table name, as it has inms_ instead of inus_, the net_bounds data in this table is recorded in milliseconds. Why is the data not captured in microseconds?
2> Here again, could the field net_IdentityModelVersion be named as "net_WilsonVersion" to maintain parity with other tables?
3> Similar to #3 above are failed metadata requests captured in this table?

westin-m:
> 2> Any wiki doc explaining the different values of net_OperationStatus? What does Direct / Automatic / FirstRefresh / LastKnownGood indicate? 

I can make a doc, good idea

> 3> Does this table capture both data for both successful and failed metadata requests? Ex: In the case of validation failure with msg "SignatureKeyNotFound" there would be an eSTS metadata request. So, wondering how to identify failure requests? 

failed requests will have an exception message attached, they just aren't in the table right now because I was going for success cases to show all the operations. Unit tests prove the existence of this in the taglist right now

> 4> Why is the net_MetadataAddress add empty or null for some records?

I added metadata address partway through testing. We weren't sure at first if we wanted it. It is null in some of the LKG entries because I used some unit test-like code to contrive the error which precedes that tag being used. Normally, it will look like the LKG entry that has a metadata address

> Table identitymodelconfigurationrequesttotaldurationinms_histogram 1> I am inferring from the table name, as it has inms_ instead of inus_, the net_bounds data in this table is recorded in milliseconds. Why is the data not captured in microseconds?

Milliseconds was recommended to me and it matched other, similar code. I can change it if microseconds is preferred. 

> 3> Similar to #3 above are failed metadata requests captured in this table?

same as above, a different taglist is created that contains an exception. right now, unit tests prove that this portion is populated

GeoK (approved):

build/dependencies.props
```diff
@@ -8,6 +8,7 @@
     <MicrosoftSourceLinkGitHubVersion>1.0.0</MicrosoftSourceLinkGitHubVersion>
     <NetStandardVersion>2.0.3</NetStandardVersion>
     <NewtonsoftVersion>13.0.3</NewtonsoftVersion>
+    <SystemDiagnosticSourceVersion>8.0.1</SystemDiagnosticSourceVersion>
```
pmaytak: Is this the minimum version we can use? I remember we had issues with this when we added it to MSAL because Azure Functions didn't have this dependency. (I think it was in .NET 6 Azure Functions, so we downgraded to 6.0.1.)

brentschmaltz: Yep, we have to lower this.

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryInstrumentation.cs
```diff
@@ -0,0 +1,64 @@
...
+
+using System;
+using System.Diagnostics;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens.Telemetry
+{
+    internal class TelemetryInstrumentation : ITelemetryInstrumentation
+    {
+
```
brentschmaltz: could remove this extra line.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -36,6 +37,8 @@ public class JwtSecurityTokenHandler : SecurityTokenHandler
         private static string _shortClaimType = _namespace + "/ShortTypeName";
         private bool _mapInboundClaims = DefaultMapInboundClaims;
 
+        internal ITelemetryInstrumentation _telemetryClient = new TelemetryInstrumentation();
```
brentschmaltz: There is a naming oddity here.
Should ITelemetryInstrumentation be ITelemeteryClient?
Then TelemetryInstrumentation be TelemetryClient?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.cs
```diff
@@ -511,6 +514,10 @@ await ValidateJWEAsync(jsonWebToken, validationParameters, currentConfiguration)
                     // where a new valid configuration was somehow published during validation time.
                     if (currentConfiguration != null)
                     {
+                        _telemetryClient.IncrementConfigurationRefreshRequestCounter(
```
brentschmaltz, line 517: We should not count here as there will be a count in RequestRefresh and GetBaseConfigurationAsync.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -183,18 +189,34 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
-                                    LogHelper.FormatInvariant(
-                                        LogMessages.IDX20810,
-                                        result.ErrorMessage)));
+                        {
+                            var ex = new InvalidConfigurationException(
+                                LogHelper.FormatInvariant(
+                                    LogMessages.IDX20810,
+                                    result.ErrorMessage));
+
+                            _telemetryClient.IncrementConfigurationRefreshRequestCounter(
```
brentschmaltz: This will be double counted as we report the same error in the catch.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -311,6 +352,10 @@ public override void RequestRefresh()
         {
             DateTimeOffset now = DateTimeOffset.UtcNow;
 
+            _telemetryClient.IncrementConfigurationRefreshRequestCounter(
+                MetadataAddress,
+                TelemetryConstants.Protocols.Direct);
```
brentschmaltz: Using the name Protocols.Direct in a method called RequestRefresh is kind of confusing.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -183,18 +189,34 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+                        TelemetryConstants.Protocols.FirstRefresh);
+
                     UpdateConfiguration(configuration);
                 }
                 catch (Exception ex)
                 {
                     fetchMetadataFailure = ex;
+                    _telemetryClient.IncrementConfigurationRefreshRequestCounter(
+                        MetadataAddress,
+                        TelemetryConstants.Protocols.FirstRefresh,
```
brentschmaltz, line 228: Perhaps this counter should be FirstAutomatic.

src/Microsoft.IdentityModel.Logging/Microsoft.IdentityModel.Logging.csproj
```diff
@@ -32,6 +32,10 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' or '$(TargetFramework)' == 'netstandard2.0' ">
```
brentschmaltz: This reference should be in the assembly where the types are used.
Currently this is M.IM.Tokens, which includes this reference, which allows the project to build.

I think the telemetry stuff should be in M.IM.Logging.

src/Microsoft.IdentityModel.Tokens/Telemetry/IdentityModelTelemetry.cs
```diff
@@ -0,0 +1,49 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Diagnostics;
+using System.Diagnostics.Metrics;
+
+namespace Microsoft.IdentityModel.Tokens.Telemetry
```
brentschmaltz: Curious why this we didn't put this in M.IM.Logging?

src/Microsoft.IdentityModel.Tokens/Telemetry/ITelemetryInstrumentation.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.IdentityModel.Tokens.Telemetry
+{
+    internal interface ITelemetryInstrumentation
```
brentschmaltz: Since this is all internal, not seeing a strong reason for this interface.
How does it help?

brentschmaltz (changes requested):
Some naming.
Code should live in assembly where dependencies are included.
Why do we need an interface?

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryConstants.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            /// <summary>
+            /// Telemetry tag indicating configuration retrieval when there is no previously cached configuration.
+            /// </summary>
+            public const string FirstRefresh = "FirstRefresh";
+
+            /// <summary>
+            /// Telemetry tag indicating configuration retrieval when the last known good configuration is needed.
+            /// </summary>
+            public const string LKG = "LastKnownGood";
```
pmaytak:
```suggestion
            public const string Lkg = "LastKnownGood";
```

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers:~:text=including%20acronyms%20over%20two%20letters%20in%20length

src/Microsoft.IdentityModel.Tokens/Telemetry/IdentityModelTelemetry.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    internal class IdentityModelTelemetry
```
pmaytak: nit: I would rename TelemetryInstrumentation to TelemetryClient and merge this class into it to simplify the logic.

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryInstrumentation.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void IncrementConfigurationRefreshRequestCounter(string metadataAddress, string operationStatus)
+        {
+            var tagList = new TagList()
+            {
+                { TelemetryConstants.IdentityModelVersionTag, IdentityModelTelemetryUtil.ClientVer },
```
pmaytak: nit: Should the value of `IdentityModelTelemetryUtil.ClientVer` be cached? The assembly name won't change and currently we're getting for each telemetry call.

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryConstants.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            /// <summary>
+            /// Telemetry tag indicating configuration retrieval after the refresh interval has expired.
+            /// </summary>
+            public const string Automatic = "Automatic";
+
+            /// <summary>
+            /// Telemetry tag indicating configuration retrieval per a call to RequestRefresh.
+            /// </summary>
+            public const string Direct = "Direct";
```
pmaytak: nit: Is there a more appropriate antonym for Automatic? For ex, Manual, Explicit?

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryConstants.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// Telemetry tag describing the operation being performed.
+        /// </summary>
+        public const string OperationStatusTag = "OperationStatus";
+
+        /// <summary>
+        /// Telemetry tag indicating the type of exception that occurred.
+        /// </summary>
+        public const string ExceptionTypeTag = "ExceptionType";
+
+        public static class Protocols
```
pmaytak, line 30: nit: I understand the intention but `Protocols` seems a bit more general here. Maybe use something more specific like TagValues

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -35,6 +36,9 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
         private const int ConfigurationRetrieverRunning = 1;
         private int _configurationRetrieverState = ConfigurationRetrieverIdle;
 
+        internal TimeProvider _timeProvider = TimeProvider.System;
+        internal ITelemetryInstrumentation _telemetryClient = new TelemetryInstrumentation();
```
pmaytak: nit: Internal variables should be in PascalCase.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -148,7 +152,9 @@ public async Task<T> GetConfigurationAsync()
         public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
         {
             if (_currentConfiguration != null && _syncAfter > DateTimeOffset.UtcNow)
+            {
```
pmaytak: nit: undo to conform to the current project style.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -214,13 +236,19 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+                        MetadataAddress,
+                        TelemetryConstants.Protocols.Automatic);
+
                     _ = Task.Run(UpdateCurrentConfiguration, CancellationToken.None);
                 }
             }
 
             // If metadata exists return it.
             if (_currentConfiguration != null)
+            {
```
pmaytak: nit: undo to conform to the current project style.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTelemetryTests.cs
```diff
@@ -0,0 +1,69 @@
...
+using Microsoft.IdentityModel.Tokens.Telemetry.Tests;
+using Microsoft.IdentityModel.Validators;
+using Xunit;
+
+namespace Microsoft.IdentityModel.JsonWebTokens.Tests
+{
+    public class JsonWebTokenHandlerTelemetryTests
+    {
+        [Fact]
+        public async Task ValidateJWSWithConfigAsync_ExpectedTagsExist()
```
pmaytak:
```suggestion
        public async Task ValidateJwsWithConfigAsync_ExpectedTagsExist()
```

src/Microsoft.IdentityModel.Tokens/Telemetry/IdentityModelTelemetry.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        /// <summary>
+        /// Meter name for MicrosoftIdentityModel.
+        /// </summary>
+        public const string MeterName = "MicrosoftIdentityModel_Meter";
```
pmaytak: I think all variables here can be private,

src/Microsoft.IdentityModel.Tokens/Telemetry/ITelemetryInstrumentation.cs
keegan-caruso: It should be able to help with testing.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
westin-m: This could potentially be confusing alongside the existing "automatic" tag. The tags all have doc strings, so I'll leave this for now.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
westin-m: I'll change it to "Manual", per Peter's suggestion

src/Microsoft.IdentityModel.Tokens/Telemetry/ITelemetryInstrumentation.cs
westin-m: Like Keegan said, the intention is to improve the testing experience

src/Microsoft.IdentityModel.Tokens/Telemetry/IdentityModelTelemetry.cs
westin-m: There is less visibility to this if it's put in Logging and we were trying to avoid more friend assemblies

src/Microsoft.IdentityModel.Tokens/Telemetry/IdentityModelTelemetry.cs
westin-m: As more telemetry gets added, this class and TelemetryClient will grow a lot and I would rather keep them separate and focused in anticipation of that.

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryConstants.cs
westin-m: "protocols" is being used to match the specific package that uses these values, not to describe the purpose of the values.

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryConstants.cs
westin-m: I like Manual, thank you

src/Microsoft.IdentityModel.Tokens/Telemetry/IdentityModelTelemetry.cs
pmaytak: I see. My underlying feeling behind the comment was that it's a bit unclear at first glance what the purpose and the difference between the two classes is. Looks like TelemetryInstrumentation prepares the data and IdentityModelTelemetry actually pushes the data. At the very least, I suggest adding class comments to clarify the purpose. And also maybe renaming the class to make it clearer (like TelemetryRecorder, or Uploader or something?)

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryConstants.cs
pmaytak: What if the same constant values will used for multiple projects or like a feature that spans multiple projects? I think it's okay to just put all telemetry constants in one class and maybe add regions. I think it may end up be too cumbersome to keep constants organized here. Or at least renaming the class to match a feature, so like Configuration in this case or something.

src/Microsoft.IdentityModel.Tokens/Telemetry/IdentityModelTelemetry.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Microsoft.IdentityModel.Tokens.Telemetry
+{
+    internal class IdentityModelTelemetry
+    {
+        /// <summary>
+        /// Meter name for MicrosoftIdentityModel.
+        /// </summary>
+        private const string MeterName = "MicrosoftIdentityModel_Meter";
+        //public const string ServiceName = "MicrosoftIdentityModel";
```
keegan-caruso:
```suggestion
```

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryClient.cs
```diff
@@ -0,0 +1,65 @@
...
+                { TelemetryConstants.OperationStatusTag, operationStatus },
+                { TelemetryConstants.ExceptionTypeTag, exception.GetType().ToString() }
+            };
+
+            IdentityModelTelemetry.IncrementConfigurationRefreshRequestCounter(tagList);
+        }
+
+        public void LogConfigurationRetrievalDuration(string metadataAddress, TimeSpan operationDuration)
+        {
+
```
keegan-caruso:
```suggestion
```

src/Microsoft.IdentityModel.Tokens/Telemetry/IdentityModelTelemetry.cs
westin-m: Understood. I'll add some comments to clarify and think about the name a bit more.

brentschmaltz (changes requested):
Suggest putting telemetry code in M.IM.Logging assembly.

brentschmaltz (approved):

PR #3027, opened 2024-11-21
Extensibility tests: Audience - JWT, SAML and SAML2

iNinja:
# Extensibility tests: Audience - JWT, SAML and SAML2
- Handle potential exceptions thrown by the audience validation delegate on JWT, SAML, and SAML2
- Added custom audience validation delegates for testing
- Added audience extensibility tests for JWT, SAML and SAML2

Part of #2711


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
```diff
@@ -267,13 +267,30 @@ private async ValueTask<ValidationResult<ValidatedToken>> ValidateJWSAsync(
...
+            try
+            {
+                audienceValidationResult = validationParameters.AudienceValidator(
+                    tokenAudiences, jsonWebToken, validationParameters, callContext);
 
-            if (!audienceValidationResult.IsValid)
+                if (!audienceValidationResult.IsValid)
+                {
+                    StackFrame audienceValidationFailureStackFrame = StackFrames.AudienceValidationFailed ??= new StackFrame(true);
+                    return audienceValidationResult.UnwrapError().AddStackFrame(audienceValidationFailureStackFrame);
```
pmaytak, line 279: Why not `.AddCurrentStackFrame()` like in other cases?

pmaytak (approved):

msbw2:
See my comments on #3028 as this PR is very similar to that one and thus the same comments apply.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
iNinja: This is the old implementation. I missed updating it :)
Thanks for noticing.

westin-m (approved):

PR #3029, opened 2024-11-21
Extensibility tests: Issuer signing key - JWT, SAML and SAML2

iNinja:
# Extensibility tests: Issuer signing key - JWT, SAML and SAML2
- Handle potential exception thrown by the  issuer signing key validation delegate in JWT, SAML, and SAML2
- Added custom issuer signing key validation delegates for testing
- Added lifetime extensibility tests for JWT, SAML, and SAML2

Part of #2711


src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/IssuerSigningKeyValidationError.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+
```
jennyf19:
```suggestion
```

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/IssuerSigningKeyValidationError.cs
```diff
@@ -0,0 +1,52 @@
...
+                    SigningKey = InvalidSigningKey
+                };
+                exception.SetValidationError(this);
+
+                return exception;
+            }
+
+            return base.GetException();
+        }
+
```
jennyf19, line 39:
```suggestion
```

src/Microsoft.IdentityModel.Tokens/Validation/ValidationFailureType.cs
```diff
@@ -134,5 +134,10 @@ private class XmlValidationFailure : ValidationFailureType { internal XmlValidat
         /// </summary>
         public static readonly ValidationFailureType IssuerValidatorThrew = new IssuerValidatorFailure("IssuerValidatorThrew");
         private class IssuerValidatorFailure : ValidationFailureType { internal IssuerValidatorFailure(string name) : base(name) { } }
+
+        /// <summary>
+        /// Defines a type that represents that the issuer signing key validation delegate threw an exception.
+        /// </summary>
+        public static readonly ValidationFailureType IssuerSigningKeyValidatorThrew = new IssuerSigningKeyValidationFailure("IssuerSigningKeyValidatorThrew");
```
jennyf19, line 146: `IssuerSigningKeyValidatorThrew` -> `IssuerSigningKeyValidatorFailure` ?

jennyf19 (approved):

src/Microsoft.IdentityModel.Tokens/Validation/ValidationFailureType.cs
iNinja: I am not sure I follow the question here. Is it about the inheritance, the names?

pmaytak: +1 to naming. It wasn't immediately clear to me what this var represented. `Failure` instead of `Threw` makes it more clear, I think. This naming pattern is used in other places, so maybe rename all in a separate PR, if you agree.

iNinja: We will be reviewing all the `ValidationFailureType` static instances to make sure the names correctly reflect the reason the validation has failed.
We initially had `AudienceValidationFailed`, `LifetimeValidationFailed`, but we are looking into being more specific.

I think this one may need a review on the name to say something like "the user provided issuer signing key validation delegate threw an exception".
Happy to take any ideas for names, otherwise I'll make sure to run the list of names by JM and the larger group when we're happy.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.Extensibility.IssuerSigningKey.cs
```diff
@@ -0,0 +1,284 @@
...
+                {
+                    ExpectedException = new ExpectedException(
+                        typeof(CustomSecurityTokenInvalidSigningKeyException),
+                        nameof(CustomIssuerSigningKeyValidationDelegates.CustomIssuerSigningKeyValidatorCustomExceptionDelegate)),
+                    IssuerSigningKeyValidationError = new CustomIssuerSigningKeyValidationError(
+                        new MessageDetail(
+                            nameof(CustomIssuerSigningKeyValidationDelegates.CustomIssuerSigningKeyValidatorCustomExceptionDelegate), null),
+                        ValidationFailureType.SigningKeyValidationFailed,
+                        typeof(CustomSecurityTokenInvalidSigningKeyException),
+                        new StackFrame("CustomIssuerSigningKeyValidationDelegates.cs", 175),
```
pmaytak, line 99: Do we assert or use this line number anywhere? My initial thought is that this can cause flaky tests. Plus every time an internal refactor is done, these test cases can be affected. Maybe something to discuss.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.Extensibility.IssuerSigningKey.cs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        public static TheoryData<IssuerSigningKeyExtensibilityTheoryData> IssuerSigningKey_ExtensibilityTestCases
+        {
+            get
+            {
+                var theoryData = new TheoryData<IssuerSigningKeyExtensibilityTheoryData>();
+                CallContext callContext = new CallContext();
+                var utcNow = DateTime.UtcNow;
+                var utcPlusOneHour = utcNow + TimeSpan.FromHours(1);
```
pmaytak, line 61: Can TimeProvider be used for this?

test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlSecurityTokenHandler.Extensibility.IssuerSigningKey.cs
```diff
@@ -0,0 +1,283 @@
...
+            }
+            catch (Exception ex)
+            {
+                theoryData.ExpectedException.ProcessException(ex, context);
+            }
+
+            TestUtilities.AssertFailIfErrors(context);
+        }
+
+        public static TheoryData<IssuerSigningKeyExtensibilityTheoryData> IssuerSigningKey_ExtensibilityTestCases
```
pmaytak, line 53: Is there a way to reduce code duplication for these test cases? Looks like the SAML and SAML2 only differ by the last test case, and JWT and SAML ones also differ just by the extra stack trace count. Maybe move common cases to a different file (maybe this applies to other tests too, so can be done as a follow up).

pmaytak (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.Extensibility.IssuerSigningKey.cs
iNinja: The number is included because the constructor for `StackFrame` requires file name and line number.
Because of the reason you mentioned regarding flaky tests, we are not asserting this number anywhere.

I am considering simply setting the number to 0 to avoid confusion.

test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlSecurityTokenHandler.Extensibility.IssuerSigningKey.cs
iNinja: I have a task to look exactly into this. Appreciate the input.

PR #3030, opened 2024-11-21
Extensibility tests: Token Type - JWT

iNinja:
# Extensibility tests: Token Type - JWT
- Updated `TokenTypeValidationError` and updated the default token type validation delegate to use it
- Added log message and validation failure type for the case where the token type validation delegate throws
- Added custom validation error and validation delegates for token type validation
- Handle the case where the token type validation delegate throws in `JsonWebTokenHandler`
- Added extensibility tests for token type validation on JWT

Part of #2711


src/Microsoft.IdentityModel.Tokens/Validation/ValidationFailureType.cs
```diff
@@ -134,5 +134,10 @@ private class XmlValidationFailure : ValidationFailureType { internal XmlValidat
         /// </summary>
         public static readonly ValidationFailureType IssuerValidatorThrew = new IssuerValidatorFailure("IssuerValidatorThrew");
         private class IssuerValidatorFailure : ValidationFailureType { internal IssuerValidatorFailure(string name) : base(name) { } }
+
+        /// <summary>
+        /// Defines a type that represents that the token type validation delegate threw and exception.
```
westin-m:
```suggestion
        /// Defines a type that represents the fact that the token type validation delegate threw an exception.
```
Small suggestion for more clarity

westin-m (approved):

src/Microsoft.IdentityModel.Tokens/Validation/ValidationFailureType.cs
iNinja: Thanks. Sounds better this way.

jennyf19 (approved):

PR #3033, opened 2024-11-25
Fix powershell script for nuget update

SaurabhMSFT:
## Description

The powershell script that update nuget config file should be run in the correct directory.


build/template-Build-run-tests-sign.yml
iNinja, line 1: Lines 51, 52, and 54 appear to be debug information.

Could we clean that and just keep the `workingDirectory` change?

SaurabhMSFT: That is temporary changes to debug the build.

NuGet.Config
```diff
@@ -2,6 +2,6 @@
 <configuration>
   <packageSources>
     <clear />
-	<add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+  <add key="IDDP" value="https://identitydivision.pkgs.visualstudio.com/_packaging/IDDP/nuget/v3/index.json" />
```
brentschmaltz: Our external customers will not be able to reach these packages and their local builds will fail.

build/template-Build-run-tests-sign.yml
```diff
@@ -47,12 +47,14 @@ steps:
   displayName: 'Strong Name Bypass'
 
 - powershell: |
+    $q = (dotnet nuget list source --format Short)
```
iNinja: This line is introduced, but on line 51 we still call the command again to verify whether it contains NuGet's index.

SaurabhMSFT: this is to print the first run message. let me see if i can avoid it using an env variable.

jennyf19:
did you run a release build off of this branch to ensure that still works?

SaurabhMSFT:
> did you run a release build off of this branch to ensure that still works?

Yes, it works https://dev.azure.com/IdentityDivision/IDDP/_build/results?buildId=1395135&view=results

PR #3042, opened 2024-11-28
Fix bug with AadIssuerValidator

ssmelov:
# Fix bug with AadIssuerValidator

#3047

## Description
The current implementation of AadIssuerValidator always expects that token issuer length should be greater than templateTenantIdPosition + tenantId's length.
It's not true for the case when Issuer in token and Issuer template in OIDC configuration match, but don't have a trailing slash.

### Example:
Issuer template in OIDC configuration: `https://sts.windows.net/{tenantid}`
Issuer in JWT token: `https://sts.windows.net/f645ad92-e38d-4d1a-b510-d1b09a74a8ca`

Condition `tokenIssuer.Length <= templateTenantIdPosition + tenantId.Length` of `IsValidIssuer` method results to true because `tokenIssuer.Length` equals to `templateTenantIdPosition` + `tenantId.Length`. As a result, the issuer is considered as invalid even when it matches to the issuer template from OIDC config.

### DIFF
--- a/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
+++ b/src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
@@ -399,7 +399,7 @@ namespace Microsoft.IdentityModel.Validators
                     return false;
 
                 // Ensure tokenIssuer is atleast as long as issuerTemplate with tenantIdTemplate replaced
-                if (tokenIssuer.Length <= templateTenantIdPosition + tenantId.Length)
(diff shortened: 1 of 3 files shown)


sruke (approved):

JoshLozensky (approved):

jennyf19:
@ssmelov do you mind opening an issue for this and linking back to the PR? Thank you.

brentschmaltz (approved):

jennyf19:
> @ssmelov do you mind opening an issue for this and linking back to the PR? Thank you.

nevermind, opened one so we can track it more easily: #3047

Issue #2983, opened 2024-11-06
Update the code coverage action to only run on code files in src


pmaytak:
We should update the code coverage action to only run on \src folder and only code files (.cs).
 
The Include section in the settings does include .dlls, so interesting that the analysis runs on everything else too. Maybe source files need to be specified instead of DLLs?
 
For example, here it runs on workflow file:
[Add release tactics template by kellyyangsong · Pull Request #2941 · AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2941)

PR #3048, opened 2024-12-03
Disable Coverage PR comments

westin-m:
Related to #2983 

Until code coverage report is configured ideally, do not post a comment with the report each time a push is made. 

The report is still visible in the summary section of the action that runs unit tests.


JoshLozensky (approved):

jennyf19 (approved):
ideally we could block a PR based on code coverage results.

PR #3051, opened 2024-12-04
Implement lazy ClaimsIdentity creation from ValidatedToken on SAML and SAML2 on the new validation model

iNinja:
# Implement lazy ClaimsIdentity creation from ValidatedToken on SAML and SAML2
- Implemented the `ClaimsIdentity` creation methods to allow `ValidatedToken` to lazily create the claims when accessed in SAML and SAML2 token handlers.
- Updated new validation model on SAML and SAML2 to return the validated fields as part of `ValidatedToken`
- Added tests to compare the `ClaimsIdentity` returned in the legacy path against the new one

Part of #2711


src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2SecurityTokenHandler.ClaimsIdentity.cs
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the MIT License.
+
+using System.Security.Claims;
+using Microsoft.IdentityModel.Logging;
+
+#nullable enable
+namespace Microsoft.IdentityModel.Tokens.Saml2
+{
+    /// <summary>
+    /// A <see cref="SecurityTokenHandler"/> designed for creating and validating Saml Tokens. See: http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf
```
westin-m:
```suggestion
    /// A <see cref="SecurityTokenHandler"/> designed for creating and validating Saml2 Tokens. See: http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf
```

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ClaimsIdentity.cs
```diff
@@ -0,0 +1,85 @@
...
+    public partial class SamlSecurityTokenHandler : SecurityTokenHandler
+    {
+        internal override ClaimsIdentity CreateClaimsIdentityInternal(SecurityToken securityToken, ValidationParameters validationParameters, string issuer)
+        {
+            return CreateClaimsIdentity(securityToken as SamlSecurityToken, validationParameters, issuer);
+        }
+
+        internal ClaimsIdentity CreateClaimsIdentity(SamlSecurityToken? samlToken, ValidationParameters validationParameters, string issuer)
+        {
+            if (samlToken == null)
```
westin-m, line 24: Clarifying question: why is `samlToken` nullable in the parameter list if this method throws when it is `null`?

iNinja: In this situation it is because the `CreateClaimsIdentityInternal` method receives a `SecurityToken` that needs to be cast to `SamlSecurityToken` and that returns a nullable value. In order to keep things simple, I decided to leave the parameter here as nullable, because otherwise I would need to push that check to the other method, and this one is already performing checks.

In practice, this path is reached only from a `ValidatedToken` issued by a `SamlSecurityTokenHandler`, so the `SecurityToken` object would always be a `SamlSecurityToken`, the cast would succeed, and the `samlToken` would never be null.

kllysng (approved):

westin-m (approved):

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ClaimsIdentity.cs
westin-m: Then is there a reason you don't directly cast it, but rather use `as`?
This is not a blocking question.

iNinja: Thanks for bringing this up.

After thinking about it for a minute, I think you're right. We will end up throwing the exception if the `securityToken` were to be some other class, so may as well do it at the casting point to make the error more noticeable.

PR #3054, opened 2024-12-06
Add logging to the new validation model

iNinja:
# Add logging to the new validation model
This PR introduces .NET's [high performance logging](https://learn.microsoft.com/en-us/dotnet/core/extensions/high-performance-logging) to the new validation model.
The aim is to provide a way to log the results after the fact and on demand, instead of logging as we perform the validation.

- Added Microsoft.Extensions.Logging.Abstractions to Microsoft.IdentityModel.Tokens to access ILogger (version 9.0.0 does not explicitly support dotnet6, so I am using 8.0.2)
- Added LoggingEventIds to store event ids used by the logging methods in M.IM.Tokens
- Added logging to ValidationError and ValidatedToken to log the results obtained from ValidateTokenAsync

Part of #2711


src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
```diff
@@ -72,5 +72,9 @@
     <AdditionalFiles Include="PublicAPI/$(TargetFramework)/InternalAPI.Shipped.txt" />
     <AdditionalFiles Include="PublicAPI/$(TargetFramework)/InternalAPI.Unshipped.txt" />
   </ItemGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.2" />
```
keegan-caruso: We should move the version to a var here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/build/dependencies.props

iNinja: Will do, thank you.

build/dependencies.props
```diff
@@ -11,6 +11,7 @@
     <SystemMemoryVersion>4.5.5</SystemMemoryVersion>
     <SystemSecurityCryptographyCngVersion>4.5.0</SystemSecurityCryptographyCngVersion>
     <SystemTextJson>8.0.5</SystemTextJson>
+    <MicrosoftExtensionsLoggingAbstractionsVersion>8.0.2</MicrosoftExtensionsLoggingAbstractionsVersion>
```
kllysng: Very nit - could this be inserted to preserve alphabetical order?

kllysng (approved):
One small comment

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
```diff
@@ -168,6 +175,69 @@ private object ClaimsIdentitySyncObj
...
+                bool actorWasValidated) => s_tokenValidationSucceeded(
+                    logger,
+                    validatedAudience,
+                    validatedLifetime,
+                    validatedIssuer,
+                    validatedTokenType,
+                    validatedSigningKeyId,
+                    actorWasValidated,
+                    null);
+
```
westin-m:
```suggestion
```
extra line

westin-m (approved):

build/dependencies.props
iNinja: Definitely!

PR #3056, opened 2024-12-10
Updated documentation for the new validation model and restructured internals

iNinja:
# Updated documentation for the new validation model and restructured internals
Clean up work and documentation updates for the new validation model.

- Removed static stack frames and replaced with the simplified approach of `GetCurrentStackFrame()` and `AddCurrentStackFrame()`
- Updated `IssuerValidationSource` to be extensible.
- Extracted validated fields onto their own files and made the structures read-only.
- Updated documentation

Part of #2711


src/Microsoft.IdentityModel.Tokens/Validation/Validators.IssuerSigningKey.cs
```diff
@@ -40,18 +37,15 @@ public static partial class Validators
...
+        /// <param name="callContext">The <see cref="CallContext"/> that contains call information.</param>
         internal static ValidationResult<ValidatedSigningKeyLifetime> ValidateIssuerSigningKey(
             SecurityKey securityKey,
             SecurityToken securityToken,
             ValidationParameters validationParameters,
 #pragma warning disable CA1801 // Review unused parameters
             BaseConfiguration? configuration,
 #pragma warning restore CA1801 // Review unused parameters
-            CallContext? callContext)
+            CallContext callContext)
```
kllysng, line 48: Have we verified that changes like this one won't break any friend assemblies that might be using it?

iNinja: No part of the new validation model has been officially released as of yet. We have not included any piece of this in SAL or MISE (other than Brent experimenting on a branch, maybe).

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Exceptions/SamlValidationError.cs
```diff
@@ -19,7 +30,11 @@ internal SamlValidationError(
         {
         }
 
-        internal override Exception GetException()
+        /// <summary>
+        /// Creates an instance of an <see cref="Exception"/> using <see cref="ValidationError"/>
+        /// </summary>
+        /// <returns>An instance of an Exception.</returns>
+        public override Exception GetException()
```
jmprieur: If this method is called several times, shouldn't the exception be cached?

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Exceptions/Saml2ValidationError.cs
```diff
@@ -19,7 +30,11 @@ internal Saml2ValidationError(
...
 
-        internal override Exception GetException()
+        /// <summary>
+        /// Creates an instance of an <see cref="Exception"/> using <see cref="ValidationError"/>
+        /// </summary>
+        /// <returns>An instance of an Exception.</returns>
+        public override Exception GetException()
         {
             if (ExceptionType == typeof(Saml2SecurityTokenReadException))
             {
```
jmprieur, line 40: same question. Don't we want to cache the exception?

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AlgorithmValidationError.cs
```diff
@@ -21,7 +33,11 @@ public AlgorithmValidationError(
...
 
-        internal override Exception GetException()
+        /// <summary>
+        /// Creates an instance of an <see cref="Exception"/> using <see cref="ValidationError"/>
+        /// </summary>
+        /// <returns>An instance of an Exception.</returns>
+        public override Exception GetException()
         {
             if (ExceptionType == typeof(SecurityTokenInvalidAlgorithmException))
             {
```
jmprieur, line 44: same remark

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AudienceValidationError.cs
```diff
@@ -28,7 +41,7 @@ public AudienceValidationError(
         /// Creates an instance of an <see cref="Exception"/> using <see cref="ValidationError"/>
         /// </summary>
         /// <returns>An instance of an Exception.</returns>
-        internal override Exception GetException()
+        public override Exception GetException()
         {
             if (ExceptionType == typeof(SecurityTokenInvalidAudienceException))
             {
```
jmprieur, line 49: same

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/IssuerSigningKeyValidationError.cs
```diff
@@ -21,7 +33,11 @@ internal IssuerSigningKeyValidationError(
...
 
-        internal override Exception GetException()
+        /// <summary>
+        /// Creates an instance of an <see cref="Exception"/> using <see cref="ValidationError"/>
+        /// </summary>
+        /// <returns>An instance of an Exception.</returns>
+        public override Exception GetException()
         {
             if (ExceptionType == typeof(SecurityTokenInvalidSigningKeyException))
             {
```
jmprieur, line 44: same (cache?)

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/IssuerValidationError.cs
```diff
@@ -21,9 +33,16 @@ internal IssuerValidationError(
...
 
-        internal override Exception GetException()
+        /// <summary>
+        /// Creates an instance of an <see cref="Exception"/> using <see cref="ValidationError"/>
+        /// </summary>
+        /// <returns>An instance of an exception.</returns>
+        public override Exception GetException()
         {
             if (ExceptionType == typeof(SecurityTokenInvalidIssuerException))
             {
```
jmprieur, line 49: same (cache?)

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/LifetimeValidationError.cs
```diff
@@ -28,7 +41,7 @@ public LifetimeValidationError(
         /// Creates an instance of an <see cref="Exception"/> using <see cref="ValidationError"/>
         /// </summary>
         /// <returns>An instance of an Exception.</returns>
-        internal override Exception GetException()
+        public override Exception GetException()
         {
             if (ExceptionType == typeof(SecurityTokenNoExpirationException))
             {
```
jmprieur, line 48: same (cache?)

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/SignatureValidationError.cs
```diff
@@ -7,8 +7,20 @@
 #nullable enable
 namespace Microsoft.IdentityModel.Tokens
 {
-    internal class SignatureValidationError : ValidationError
+    /// <summary>
+    /// Represents a signature validation error.
+    /// </summary>
+    public class SignatureValidationError : ValidationError
```
jmprieur: ValidationError needs to be public too, as well as the types of the parameters

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/SignatureValidationError.cs
```diff
@@ -21,7 +33,11 @@ public SignatureValidationError(
...
         }
 
-        internal override Exception GetException()
+        /// <summary>
+        /// Creates an instance of an <see cref="Exception"/> using <see cref="ValidationError"/>
+        /// </summary>
+        /// <returns>An instance of an exception.</returns>
+        public override Exception GetException()
         {
             var inner = InnerException ?? InnerValidationError?.GetException();
```
jmprieur, line 43: same (should the exceptions be cached?)

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/TokenTypeValidationError.cs
```diff
@@ -21,7 +33,11 @@ internal TokenTypeValidationError(
...
 
-        internal override Exception GetException()
+        /// <summary>
+        /// Creates an instance of an <see cref="Exception"/> using <see cref="ValidationError"/>
+        /// </summary>
+        /// <returns>An instance of an exception.</returns>
+        public override Exception GetException()
         {
             if (ExceptionType == typeof(SecurityTokenInvalidTypeException))
             {
```
jmprieur, line 44: same (cache?)

src/Microsoft.IdentityModel.Tokens/Validation/Results/IssuerValidationSource.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#nullable enable
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Represents the source of the validation of an issuer.
```
jmprieur: what does it mean?
can you please provide examples?
(I see: configuration, validation parameters below)
or say the the only values are the static instances?

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedSigningKeyLifetime.cs
```diff
@@ -0,0 +1,104 @@
...
+
+using System;
+
+#nullable enable
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Represents a validated signing key lifetime.
+    /// </summary>
+    internal readonly struct ValidatedSigningKeyLifetime : IEquatable<ValidatedSigningKeyLifetime>
```
jmprieur, line 12: wouldn't a ValidatedSingingKeyLifetime a kind of ValidatedLifetime ? should we inherit?

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
```diff
@@ -14,30 +14,22 @@ namespace Microsoft.IdentityModel.Tokens
...
-            SecurityToken = securityToken ?? throw new ArgumentNullException(nameof(securityToken));
-            ValidationParameters = validationParameters ?? throw new ArgumentNullException(nameof(validationParameters));
-        }
-
         /// <summary>
         /// Logs the validation result.
         /// </summary>
-        public void Log(ILogger logger)
-        {
-            Logger.TokenValidationSucceeded(
```
jmprieur, line 40: Is it done whatever the LogLevel?

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
```diff
@@ -14,30 +14,22 @@ namespace Microsoft.IdentityModel.Tokens
...
     /// Contains the results of successfully validating a <see cref="SecurityToken"/>.
     /// </summary>
-    internal class ValidatedToken
+    /// <remarks>
+    /// Creates an instance of <see cref="ValidatedToken"/>
+    /// </remarks>
+    /// <param name="securityToken">The <see cref="SecurityToken"/> that is being validated.</param>
+    /// <param name="tokenHandler">The <see cref="TokenHandler"/> that is being used to validate the token.</param>
+    /// <param name="validationParameters">The <see cref="ValidationParameters"/> to be used for validating the token.</param>
+    internal class ValidatedToken(
```
jmprieur: is it ValidatedToken? Or ValidatedTokenResult?

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
```diff
@@ -177,6 +201,9 @@ private object ClaimsIdentitySyncObj
         #endregion
 
         #region Logging
+        /// <summary>
+        /// Internal class used for logging.
+        /// </summary>
         private static class Logger
         {
             private static readonly Action<ILogger, string, string, Exception?> s_tokenValidationFailed =
```
jmprieur, line 216: no test against the log level?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
```diff
@@ -233,8 +213,7 @@ await ValidateJWSAsync(decryptedToken!, validationParameters, configuration, cal
 
             if (!validationResult.IsValid)
             {
-                StackFrame validationFailureStackFrame = StackFrames.JWEValidationFailed ??= new StackFrame(true);
-                return validationResult.UnwrapError().AddStackFrame(validationFailureStackFrame);
+                return validationResult.UnwrapError().AddCurrentStackFrame();
             }
 
             JsonWebToken jsonWebToken = (validationResult.UnwrapResult().SecurityToken as JsonWebToken)!;
```
jmprieur, line 218: line 248, can we do better (more precise?) than catching the general Exception class?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -31,49 +30,42 @@ internal ValidationResult<string> DecryptToken(
         {
             if (jwtToken == null)
             {
-                StackFrame tokenNullStackFrame = StackFrames.DecryptionTokenNull ??= new StackFrame(true);
                 return ValidationError.NullParameter(
                     nameof(jwtToken),
-                    tokenNullStackFrame);
+                    ValidationError.GetCurrentStackFrame());
```
jmprieur, line 35:
```suggestion
                    ValidationError.GetCurrentStackFrame(),
                    "The JWT token provided is null. Ensure that a valid token is passed.");
```

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -31,49 +30,42 @@ internal ValidationResult<string> DecryptToken(
...
             }
 
             if (validationParameters == null)
             {
-                StackFrame validationParametersNullStackFrame = StackFrames.DecryptionValidationParametersNull ??= new StackFrame(true);
                 return ValidationError.NullParameter(
                     nameof(validationParameters),
-                    validationParametersNullStackFrame);
+                    ValidationError.GetCurrentStackFrame());
```
jmprieur, line 42:
```suggestion
                    ValidationError.GetCurrentStackFrame()
                    "Validation parameters are missing. Provide the necessary parameters for token validation.");
```

jmprieur (approved):
LGTM
I've asked a few questions, which I'd like to be addressed, and proposed to improve the error messages to make them more actionable.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/SignatureValidationError.cs
iNinja: This was not meant to be public in this branch. Restored it to internal, since this is aimed at dev.

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedSigningKeyLifetime.cs
iNinja: Because we are using value types for this, we cannot implement inheritance between them.
I believe that overall, this is okay.

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
iNinja: This is what I consider a "success object", in the sense that it is only ever created and returned as part of a successful validation. The `ValidationResult` can be either positive or negative, but the `ValidatedToken` is just that, a token that was successfully validated.

src/Microsoft.IdentityModel.Tokens/Validation/Results/IssuerValidationSource.cs
iNinja: Updated the comment for clarity and extensibility. Please advise if you think of better wording.

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
iNinja: Added log level check.

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
iNinja: Added log level check

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
iNinja: In general, given this path is not meant to throw exceptions and we do not do it within the default validators, this is more of a safety net for the cases where a customer implements one of the delegates and decides to throw an exception.
Given that in that situation we have no way of knowing or limiting the types of exceptions that could occur, this seems like our safest approach.

In a basic scenario where no delegates are overwritten, this catch would never execute.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
iNinja: Currently, the `NullParameter` static method is a bit of a shortcut for a `ValidationError` with ValidationFailureType of `NullArgument`, and a message of `IDX10000: The parameter '{0}' cannot be a 'null' or an empty object.`.

I see two options here if we really want to add this information:
- We create a new IDX to allow the inclusion of further explanation
- We allow to override the IDX10000 message with this

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
iNinja: Addressed this in the other similar suggestion.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
brentschmaltz: Suggest 'ValidationParameters' instead of 'Validation parameters'

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
brentschmaltz: Suggest 'JsonWebToken' instead of "JWT'

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
brentschmaltz: Will not the current error message along with the stack trace will provide sufficient information to for the user?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
```diff
@@ -18,22 +17,14 @@ public partial class JsonWebTokenHandler : TokenHandler
     {
         /// <summary>
         /// Validates a token.
-        /// On a validation failure, no exception will be thrown; instead, the exception will be set in the returned TokenValidationResult.Exception property.
-        /// Callers should always check the TokenValidationResult.IsValid property to verify the validity of the result.
+        /// On a validation failure, no exception will be thrown; instead, the <see cref="ValidationError"/> will contain the information about the error that occurred.
```
brentschmaltz: suggest: "On validation failure no exception will be thrown. 'see cref="ValidationError"' will contain information pertaining to the error."

slightly fewer words.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
```diff
@@ -200,8 +183,7 @@ await ValidateJWEAsync(jsonWebToken, validationParameters, lkgConfiguration, cal
             }
 
             // If we reach this point, the token validation failed and we should return the error.
-            StackFrame stackFrame = StackFrames.TokenValidationFailed ??= new StackFrame(true);
-            return result.UnwrapError().AddStackFrame(stackFrame);
+            return result.UnwrapError().AddCurrentStackFrame();
```
brentschmaltz, line 185: If there no Error, UnWrapError() will throw an InvalidOperationException.
This may not be valuable to the user.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Exceptions/SamlValidationError.cs
```diff
@@ -7,9 +7,20 @@
 #nullable enable
 namespace Microsoft.IdentityModel.Tokens.Saml
 {
+    /// <summary>
```
brentschmaltz, line 10: I am curious why we need this class.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Exceptions/Saml2ValidationError.cs
```diff
@@ -7,9 +7,20 @@
...
-        internal Saml2ValidationError(
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Saml2ValidationError"/> class.
+        /// </summary>
+        /// <param name="messageDetail" /> contains information about the exception that is used to generate the exception message.
+        /// <param name="validationFailureType"/> is the type of validation failure that occurred.
+        /// <param name="exceptionType"/> is the type of exception that occurred.
+        /// <param name="stackFrame"/> is the stack frame where the exception occurred.
+        /// <param name="innerException"/> is the inner exception that occurred.
+        public Saml2ValidationError(
```
brentschmaltz, line 23: Curious why we need this class.
As I don't see a JwtValidationError.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AlgorithmValidationError.cs
```diff
@@ -7,8 +7,20 @@
 #nullable enable
 namespace Microsoft.IdentityModel.Tokens
 {
+    /// <summary>
+    /// Represents an algorithm validation error.
```
brentschmaltz: Suggest: When validating a SecurityToken the user can specify which algorithms are acceptable. AlgorithmValidationError contains details of why the validation of the SecurityToken failed associated with the algorithm found in the SecurityToken.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AlgorithmValidationError.cs
```diff
@@ -7,8 +7,20 @@
...
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AlgorithmValidationError"/> class.
+        /// </summary>
+        /// <param name="messageDetail" /> contains information about the exception that is used to generate the exception message.
+        /// <param name="validationFailureType"/> is the type of validation failure that occurred.
+        /// <param name="exceptionType"/> is the type of exception that occurred.
+        /// <param name="stackFrame"/> is the stack frame where the exception occurred.
+        /// <param name="invalidAlgorithm"/> is the algorithm that could not be validated.
+        /// <param name="innerException"/> is the inner exception that occurred.
```
brentschmaltz: suggest: is the inner exception that occurred, may be null.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/TokenTypeValidationError.cs
```diff
@@ -7,9 +7,21 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TokenTypeValidationError"/> class.
+        /// </summary>
+        /// <param name="messageDetail" /> contains information about the exception that is used to generate the exception message.
+        /// <param name="validationFailureType"/> is the type of validation failure that occurred.
+        /// <param name="exceptionType"/> is the type of exception that occurred.
+        /// <param name="stackFrame"/> is the stack frame where the exception occurred.
+        /// <param name="invalidTokenType"/> is the token type that could not be validated.
+        /// <param name="innerException"/> is the inner exception that occurred.
+        public TokenTypeValidationError(
```
brentschmaltz, line 25: In general we are not consistent with our default parameters.
invalidTokenType can be null, but we don't have default = null, for innerException we do have a default.

default parameters limit how class methods can be modified in the future.

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedIssuer.cs
```diff
@@ -0,0 +1,97 @@
...
+    internal readonly struct ValidatedIssuer : IEquatable<ValidatedIssuer>
+    {
+        /// <summary>
+        /// Initializes a new instance of <see cref="ValidatedIssuer"/>.
+        /// </summary>
+        /// <param name="Issuer">The issuer that was validated.</param>
+        /// <param name="ValidationSource">The source of the validation, i.e. configuration, validation parameters.</param>
+        public ValidatedIssuer(string Issuer, IssuerValidationSource ValidationSource)
+        {
+            this.Issuer = Issuer;
```
brentschmaltz: in general we don't use 'this.'

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedSigningKeyLifetime.cs
brentschmaltz: I suggest we rename this to reflect that this is validating the SigningKey.
ValidatedSigningKeyLifetime -> ValidatedSigningKey.

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
```diff
@@ -14,30 +14,21 @@ namespace Microsoft.IdentityModel.Tokens
     /// <summary>
     /// Contains the results of successfully validating a <see cref="SecurityToken"/>.
     /// </summary>
-    internal class ValidatedToken
+    /// <remarks>
+    /// Creates an instance of <see cref="ValidatedToken"/>
+    /// </remarks>
+    /// <param name="securityToken">The <see cref="SecurityToken"/> that is being validated.</param>
+    /// <param name="tokenHandler">The <see cref="TokenHandler"/> that is being used to validate the token.</param>
+    /// <param name="validationParameters">The <see cref="ValidationParameters"/> to be used for validating the token.</param>
```
brentschmaltz: We should comment that we throw if securityToken, tokenHandler, validationParameters are null.

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
```diff
@@ -46,29 +37,61 @@ public void Log(ILogger logger)
                 ValidatedSigningKey?.KeyId ?? "none",
                 ActorValidationResult is not null
             );
-        }
 
-        public SecurityToken SecurityToken { get; private set; }
+        /// <summary>
+        /// The <see cref="SecurityToken"/> that was validated.
+        /// </summary>
+        public SecurityToken SecurityToken { get; private set; } = securityToken ?? throw new ArgumentNullException(nameof(securityToken));
```
brentschmaltz: Do we need the 'private set' on these properties?

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
```diff
@@ -46,29 +37,61 @@ public void Log(ILogger logger)
...
 
+        /// <summary>
+        /// The audience that was validated, if any.
+        /// </summary>
         public string? ValidatedAudience { get; internal set; }
 
+        /// <summary>
+        /// The issuer that was validated. If present, it contains the source of the validation as well.
+        /// </summary>
         public ValidatedIssuer? ValidatedIssuer { get; internal set; }
```
brentschmaltz, line 78: Is the idea that the 'internal set' will become public so the user can set the value?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
iNinja: It is not possible to reach this line without an error.

ValidationError can only be constructed with a successful result or an error, but always one of the two is present.
The code above this line returns the success value if the result is valid, only reaching this point if it is not, and therefore holding the error.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
iNinja: We should not be assuming the stack traces will be normally seen, as we are aiming to not generate the exceptions unless there is a problem that is being diagnosed.

It may be worth allowing adding a little context along with the `X parameter cannot be null or empty` to make it easier to go without the stack trace and make that more of a last resort.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Exceptions/SamlValidationError.cs
iNinja: It was introduced as part of the intention to specialise the returned errors from the delegates, though there is a chance that the SAML and SAML2 validation errors could be left as ValidationError and keep the others for the common validations. I'll have a look.

src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Exceptions/Saml2ValidationError.cs
iNinja: Addressed on the SAML validation error comment.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AlgorithmValidationError.cs
iNinja: I'll update the wording to reflect the nullability, thanks.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/TokenTypeValidationError.cs
iNinja: These are two different cases and the default value intends to reflect that.

We want the `invalidTokenType` to be present, but because the value on the token itself may be missing, we have to allow `null` to be passed here.
The inner exception, on the other hand, is an optional parameter that can be added if present but is not required.

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedIssuer.cs
iNinja: Will update, explanation in another comment like this one.

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
iNinja: I think it was done when I accepted the suggestion to use the primary constructor.
I've reverted this to use a regular constructor and perform the validations there, adding the validation exceptions to the docs and removing the set from these properties.

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
iNinja: The ideal would be to make these properties be `{ get; init; }` to make them only settable from the two-part constructor when calling:
```
var validatedToken = new ValidatedToken(token, handler, parameters)
{
    ValidatedIssuer = someValidatedIssuer,
    ValidatedLifetime = someValidatedLifetime,
}
```

and outside of that, keep the property as read-only.

There is a small incompatibility between the `init` keyword and .NET Framework / .NET Standard that requires a workaround that is not included here as it requires its own conversation.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/AlgorithmValidationError.cs
iNinja: This kind of explanation sounds too verbose to me and more suited for a KB article / wiki.
I'll try to expand the current description to something more clear.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/TokenTypeValidationError.cs
jmprieur: but one important to have is the `cancellationToken = default)` as the last parameter of Async methods

src/Microsoft.IdentityModel.Tokens/Validation/Results/IssuerValidationSource.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <summary>
+        /// Represents the issuer validation source that has been matched with the configuration provided.
+        /// </summary>
+        public static readonly IssuerValidationSource IssuerMatchedConfiguration = new("IssuerMatchedConfiguration");
+
+        /// <summary>
+        /// Represents the issuer validation source that has been matched with the validation parameters provided.
+        /// </summary>
+        public static readonly IssuerValidationSource IssuerMatchedValidationParameters = new("IssuerMatchedValidationParameters");
+    }
```
jmprieur, line 45: please add a ToString() method that returns the Name

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedIssuer.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// Initializes a new instance of <see cref="ValidatedIssuer"/>.
+        /// </summary>
+        /// <param name="issuer">The issuer that was validated.</param>
+        /// <param name="validationSource">The source of the validation, i.e. configuration, validation parameters.</param>
+        public ValidatedIssuer(string issuer, IssuerValidationSource validationSource)
+        {
+            Issuer = issuer;
+            ValidationSource = validationSource;
+        }
+
```
jmprieur, line 24: Please add ToString() that return $"{Issuer} (from {ValidationSource})"

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedLifetime.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// Initializes a new instance of <see cref="ValidatedLifetime"/>.
+        /// </summary>
+        /// <param name="notBefore">The <see cref="DateTime"/> representing the time from which the token is considered valid.</param>
+        /// <param name="expires">The <see cref="DateTime"/> representing the token's expiration time.</param>
+        public ValidatedLifetime(DateTime? notBefore, DateTime? expires)
+        {
+            NotBefore = notBefore;
+            Expires = expires;
+        }
+
```
jmprieur, line 24: Please add a ToString() method that returns `$"[{NotBefore}, {Expires}["`

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedSigningKeyLifetime.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// <summary>
+        /// The date until which the signing key is considered valid.
+        /// </summary>
+        public DateTime? ValidTo { get; }
+
+        /// <summary>
+        /// The time the validation occurred.
+        /// </summary>
+        public DateTime? ValidationTime { get; }
+
```
jmprieur, line 41: Can we add a ToString() method that displays `$"{ValidationTime} ∊ [{ValidFrom}, {Validto}]"`

jmprieur (approved):
LTGM
Thanks @iNinja

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Exceptions/SamlValidationError.cs
iNinja: Took me a moment to remember the exact reasoning, but here it is and it is simple:

We have 2 exceptions thrown by the SAMLSecurityTokenHandler and SAML2SecurityHandler that need to be created, each of them deriving from a SamlSecurityTokenException or Saml2SecurityTokenException.

If we try to create those exceptions within `ValidationError` using the current approach, we would need to have `M.IM.Tokens` depend on the SAML and SAML2 packages.

We implement these two validation errors in order to have each create their custom exception.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/TokenTypeValidationError.cs
iNinja: I am adding the default on the main method.

jmprieur (approved):

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedIssuer.cs
```diff
@@ -92,6 +92,12 @@ public bool Equals(ValidatedIssuer other)
 
             return true;
         }
+
+        /// <summary>
+        /// The validated issuer's string representation.
+        /// </summary>
+        /// <returns>A string representing the issuer and where it was validated from.</returns>
+        public override string ToString() => $"{Issuer} (from {ValidationSource})";
```
keegan-caruso, line 100: Could we cache this so we don't create a new string each time? (similar comment for other files in this commit)

keegan-caruso (approved):
Just one question, otherwise LGTM

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedIssuer.cs
iNinja: I'll leave this out of this PR to discuss different approaches, but this is a great point.

PR #3060, opened 2024-12-16
New validation model - Public API review

iNinja:
# New Validation Model - Public API review
This PR is based off #3056, so some changes will show up here until that one is merged.

- Made classes and structures that will need to be public in Microsoft.IdentityModel.Tokens public.
- Added those to PublicAPI.Unshipped.txt for ease of review.

Note: This branch currently won't build because of some issues with the way the public API analyzers work when there is nullability annotations partially implemented. Working on a fix as well as making APIs public in JsonWebTokens and SAML/SAML2.

This PR is **not** to be merged.


iNinja:
Moved out of draft to trigger validations. Will move back.

jennyf19:
@iNinja is the idea to ship the preview off of `feature/new-validation-model` branch?

iNinja:
@jennyf19 yes, that would be best from my perspective, as we need to make many classes and methods public that we don't want to expose on dev.

I'll turn this PR back to draft, I moved it out of it to trigger the validations.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
```diff
@@ -18,55 +17,44 @@ public partial class JsonWebTokenHandler : TokenHandler
...
-        /// <exception cref="ArgumentNullException">Returned if <paramref name="validationParameters"/> is null.</exception>
-        /// <exception cref="ArgumentException">Returned if 'token.Length' is greater than <see cref="TokenHandler.MaximumTokenSizeInBytes"/>.</exception>
-        /// <exception cref="SecurityTokenMalformedException">Returned if <paramref name="token"/> is not a valid <see cref="JsonWebToken"/>, <see cref="ReadToken(string, CallContext)"/></exception>
-        /// <exception cref="SecurityTokenMalformedException">Returned if the validationParameters.TokenReader delegate is not able to parse/read the token as a valid <see cref="JsonWebToken"/>, <see cref="ReadToken(string, CallContext)"/></exception>
-        internal async Task<ValidationResult<ValidatedToken>> ValidateTokenAsync(
+        public async Task<ValidationResult<ValidatedToken>> ValidateTokenAsync(
             string token,
             ValidationParameters validationParameters,
             CallContext callContext,
             CancellationToken cancellationToken)
```
jmprieur, line 32: It's a best practice to no impose the usage of CancellationToken, by providing the default value:

```suggestion
            CancellationToken cancellationToken = default(CancellationToken))
```

jmprieur (approved):
LGTM

- We'd want to have the cancellation token in the public API have a default value `default(CancellationToken)`

Not blocking, but incremental:
- Let's have ToString() methods on the ValidatinSource, and the ValidatedXXX classes. This is nicer when debugging.
- We need to document better the overrides of ValidateTokenAsync in JSonWebTokenHandler. I'll propose something.
- We'd probably want to have code snippets in the validation Error class. I'll propose something.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
```diff
@@ -184,12 +184,23 @@ internal Exception GetException(Type exceptionType, Exception? innerException)
             return exception;
         }
 
-        internal void Log(ILogger logger)
+        /// <summary>
+        /// Logs the validation error.
+        /// </summary>
+        /// <param name="logger">The <see cref="ILogger"/> to be used for logging.</param>
+        [CLSCompliant(false)]
+        public void Log(ILogger logger)
```
jmprieur, line 192: Should this be public?
Is it meant to be used by users? 
Or maybe just frameworks on top of Wilson? In which case would we want to have an explicit interface implementation?

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
```diff
@@ -240,7 +256,7 @@ public ValidationError AddStackFrame(StackFrame stackFrame)
         /// <param name="lineNumber">The line number from which this method is called. CAptured automatically by default.</param>
         /// <param name="skipFrames">The number of stack frames to skip when capturing. Used to avoid capturing this method and get the caller instead.</param>
         /// <returns>The updated object.</returns>
-        internal ValidationError AddCurrentStackFrame([CallerFilePath] string filePath = "", [CallerLineNumber] int lineNumber = 0, int skipFrames = 1)
+        public ValidationError AddCurrentStackFrame([CallerFilePath] string filePath = "", [CallerLineNumber] int lineNumber = 0, int skipFrames = 1)
```
jmprieur, line 259: Same question. Should it be public? what is the scenario?

jmprieur: Same question for NullParameter().

iNinja: These need to be available for custom validation delegate advanced scenarios.

For the basic ones this can be completely ignored.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
iNinja: It is meant to be used by anyone consuming Wilson. There is a decent chance that would be an upper layer and not the consumer, but if someone were to call `ValidateTokenAsync` on the `TokenHandler` and receive the `ValidatedToken` or `ValidationError` in the result object, they can call this to log.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.Internal.cs
iNinja: Updating this on dev.

iNinja:
Closing this after gathering feedback.

PR #3078, opened 2025-01-07
Update common.props

ciaozhang:
Suppress warning:
##[warning]C:\Users\ContainerAdministrator\.nuget\packages\microsoft.bcl.memory\9.0.0\buildTransitive\netcoreapp2.0\Microsoft.Bcl.Memory.targets(4,5): Warning : Microsoft.Bcl.Memory 9.0.0 doesn't support net6.0 and has not been tested with it. Consider upgrading your TargetFramework to net8.0 or later. You may also set <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings> in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk.

https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1410786&view=logs&j=15dfcb1a-0989-5cf6-3160-3e181e44de87&t=32ee994e-3d32-5011-f0cb-438d28090217

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -29,10 +29,12 @@
     <NetStandardImplicitPackageVersion>$(NetStandardVersion)</NetStandardImplicitPackageVersion>
     <LangVersion>13</LangVersion>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
     <!-- Suprpess the "Microsoft.Bcl.Memory 9.0.0 doesn't support net6.0 and has not been tested with it" warning -->
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
   </PropertyGroup>
-
 
   <PropertyGroup>
     <EnablePackageValidation>false</EnablePackageValidation>
--- a/build/common.props
+++ b/build/common.props
@@ -29,6 +29,8 @@
     <NetStandardImplicitPackageVersion>$(NetStandardVersion)</NetStandardImplicitPackageVersion>
     <LangVersion>13</LangVersion>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <!-- Suprpess the "Microsoft.Bcl.Memory 9.0.0 doesn't support net6.0 and has not been tested with it" warning -->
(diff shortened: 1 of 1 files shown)


build/common.props
```diff
@@ -29,6 +29,8 @@
     <NetStandardImplicitPackageVersion>$(NetStandardVersion)</NetStandardImplicitPackageVersion>
     <LangVersion>13</LangVersion>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <!-- Suprpess the "Microsoft.Bcl.Memory 9.0.0 doesn't support net6.0 and has not been tested with it" warning -->
+    <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
```
jmprieur, line 36: Does this warning appear in all projects and for all target frameworks?
If not, I'd like us to have a more targeted suppression.

(in the affected csproj, and in a PropertyGroup guarded by a condition on the framewor,

jmprieur (changes requested):
Let's suppress this warning in a more targeted manner, otherwise we might miss future warnings

jmprieur (commented):
I'm assuming the risk is mitigated as we run all the tests in .NET 6.
Why do we reference 9.x packages ? where does this come from? Is it something we should only do in the net9.0 target framework

ciaozhang:
> I'm assuming the risk is mitigated as we run all the tests in .NET 6. Why do we reference 9.x packages ? where does this come from? Is it something we should only do in the net9.0 target framework.

it is from
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/c0eb23c1673e34c9343a5d432b523ba19ae31120#diff-fe95b71339300cd58f4eab448dc069e28ca64a4d25c3c67e847dfafcf44c31d9

jmprieur (approved):
Ok. So I see that we reference:

`Microsoft.Bcl.Memory` for the Base64Encoding.

How was the PR merged in the first place?

ciaozhang:
> How was the PR merged in the first place?
The current PR build doesn't check for warning, that is why we need onebranch PR build

jennyf19 (approved):

pmaytak:
> Ok. So I see that we reference:
> 
> `Microsoft.Bcl.Memory` for the Base64Encoding.
> 
> How was the PR merged in the first place?

The checks passed. Looks like we just don't run this ADO build on PRs.

image

PR #3084, opened 2025-01-09
Added internal virtual on TokenHandler

brentschmaltz:
Add support to AadIssuerValidator for ValidationParameters.
Removed the parameter BaseConfiguration from delegates as ValidationParameters has ConfigurationManager which should be used to obtain BaseConfiguration when needed.

ValidationParameters.PropertyBag was not instantiated.
Create ValidationParameters.PropertyBag and InstancePropertyBag on demand.
Copy logic from existing AadIssuerValidator, keeping as code the same as much as possible.
Copy logic from existing AadTokenValidationParametersExtension to AadValidationParametersExtension keeping as code the same as much as possible.
This copied code will need to be refactored into common base code.
Move CloudInstanceNameKey to constants file.

Added tests to AadTokenValidationParametersExtensionTests and MicrosoftIdentityIssuerValidatorTest that cover existing tests.


test/Microsoft.IdentityModel.TestUtils/TestUtilities.cs
```diff
@@ -35,6 +35,78 @@ public class GetSetContext
 
     public static class TestUtilities
     {
+        internal static ValidationParameters CreateFromTokenValidationParameters(TokenValidationParameters tokenValidationParameters)
```
jmprieur, line 38: Nice!
Do we want to do something for the delegates too?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.Internal.cs
```diff
@@ -0,0 +1,141 @@
...
+    public partial class AadIssuerValidator
+    {
+        /// <summary>
+        /// Validate the issuer for single and multi-tenant applications of various audiences (Work and School accounts, or Work and School accounts +
+        /// Personal accounts) and the various clouds.
+        /// </summary>
+        /// <param name="issuer">Issuer to validate (will be tenanted).</param>
+        /// <param name="securityToken">Received security token.</param>
+        /// <param name="validationParameters">The <see cref="ValidationParameters"/> to be used for validating the token.</param>
+        /// <param name="callContext"></param>
```
keegan-caruso: A few these are not filled out

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.Internal.cs
```diff
@@ -0,0 +1,141 @@
...
+                if (IsValidIssuer(validationParameters.ValidIssuers[i], tenantId, issuer))
+                    return new ValidatedIssuer(issuer!, IssuerValidationSource.IssuerMatchedConfiguration);
+            }
+
+            try
+            {
+                var issuerVersion = GetTokenIssuerVersion(securityToken);
+                var effectiveConfigurationManager = GetEffectiveConfigurationManager(issuerVersion);
+
+                string aadIssuer = null;
```
keegan-caruso:
```suggestion
                string aadIssuer;
```

This gets unconditionally wrote to below.

src/Microsoft.IdentityModel.Validators/AadValidationParametersExtension.Internal.cs
```diff
@@ -0,0 +1,254 @@
...
+        internal static void EnableEntraIdSigningKeyCloudInstanceValidation(this ValidationParameters validationParameters)
+        {
+            if (validationParameters == null)
+                throw LogHelper.LogArgumentNullException(nameof(validationParameters));
+
+            IssuerSigningKeyValidationDelegate issuerSigningKeyValidationDelegate = validationParameters.IssuerSigningKeyValidator;
+
+            IssuerSigningKeyValidationDelegate issuerSigningKeyValidationDelegate1 = (securityKey, securityToken, validationParameters, callContext) =>
+            {
+                // TODO remove GetAwaiter
```
keegan-caruso: Why can't we do this now and make the method async? Better to do this sooner I would think

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -374,7 +380,9 @@ public string NameClaimType
         /// Gets or sets the <see cref="IDictionary{TKey, TValue}"/> that contains a collection of custom key/value pairs.
         /// This allows addition of parameters that could be used in custom token validation scenarios.
         /// </summary>
-        public IDictionary<string, object> PropertyBag { get; }
+        public IDictionary<string, object> PropertyBag => _propertyBag ??
```
jmprieur, line 396: and this one is not unique to the instance? is it shared between ValidationParameters that were cloned from each other?

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.IdentityModel.Validators
     /// <summary>
     /// Generic class that validates the issuer for either JsonWebTokens or JwtSecurityTokens issued from the Microsoft identity platform (AAD).
     /// </summary>
-    public class AadIssuerValidator
```
jmprieur, line 22: How much copy/paste is there in the partial implementation vs how much can we have in common? I understand that the input (ValidationParameters, context) will be different, and that the output is an Error, not an exception thrown, but can the logic be a bit more common?

src/Microsoft.IdentityModel.Validators/AadValidationParametersExtension.Internal.cs
```diff
@@ -0,0 +1,254 @@
...
+                BaseConfiguration configuration = null;
+                if (validationParameters.ConfigurationManager != null)
+                    configuration = validationParameters.ConfigurationManager.GetBaseConfigurationAsync(System.Threading.CancellationToken.None).ConfigureAwait(false).GetAwaiter().GetResult();
```
jmprieur: Is it ok to have a sync over async here? I guess we need to configuration and need to block?

test/Microsoft.IdentityModel.TestUtils/TestUtilities.cs
jmprieur: Given the number of common fields, should there be a common base class with these?

src/Microsoft.IdentityModel.Validators/AadValidationParametersExtension.Internal.cs
```diff
@@ -0,0 +1,254 @@
...
+        }
+
+        /// <summary>
+        /// Validates the issuer signing key.
+        /// </summary>
+        /// <param name="securityKey">The <see cref="SecurityKey"/> that signed the <see cref="SecurityToken"/>.</param>
+        /// <param name="securityToken">The <see cref="SecurityToken"/> being validated, could be a JwtSecurityToken or JsonWebToken.</param>
+        /// <param name="configuration">The <see cref="BaseConfiguration"/> provided.</param>
+        /// <returns><c>true</c> if the issuer of the signing key is valid; otherwise, <c>false</c>.</returns>
+        internal static bool ValidateIssuerSigningKey(SecurityKey securityKey, SecurityToken securityToken, BaseConfiguration configuration)
```
jmprieur: This method is common with AadTokenValidationParametersExtensions (to the letter)

src/Microsoft.IdentityModel.Validators/AadValidationParametersExtension.Internal.cs
```diff
@@ -0,0 +1,254 @@
...
+
+                if (openIdConnectConfiguration.AdditionalData.TryGetValue(AadIssuerValidatorConstants.CloudInstanceNameKey, out object configurationCloudInstanceNameObjectValue))
+                {
+                    string configurationCloudInstanceName = configurationCloudInstanceNameObjectValue as string;
+                    if (string.IsNullOrWhiteSpace(configurationCloudInstanceName))
+                        return;
+
+                    if (!string.Equals(signingKeyCloudInstanceName, configurationCloudInstanceName, StringComparison.Ordinal))
+                        throw LogHelper.LogExceptionMessage(
+                            new SecurityTokenInvalidCloudInstanceException(
```
jmprieur, line 101: this is the only line that changes in this method vs in AadValidationParametersExtension

src/Microsoft.IdentityModel.Validators/AadValidationParametersExtension.Internal.cs
```diff
@@ -0,0 +1,254 @@
...
+                            {
+                                ConfigurationCloudInstanceName = configurationCloudInstanceName,
+                                SigningKeyCloudInstanceName = signingKeyCloudInstanceName,
+                                SigningKey = securityKey,
+                            });
+                }
+            }
+        }
+
+        private static JsonWebKey GetJsonWebKeyBySecurityKey(OpenIdConnectConfiguration configuration, SecurityKey securityKey)
```
jmprieur: Same method and content in both

src/Microsoft.IdentityModel.Validators/AadValidationParametersExtension.Internal.cs
```diff
@@ -0,0 +1,254 @@
...
+            foreach (JsonWebKey key in configuration.JsonWebKeySet.Keys)
+            {
+                if (key.Kid == securityKey.KeyId)
+                    return key;
+            }
+
+            return null;
+        }
+
+        private static string GetTid(SecurityToken securityToken)
```
jmprieur: Same method and content in both

src/Microsoft.IdentityModel.Validators/AadValidationParametersExtension.Internal.cs
```diff
@@ -0,0 +1,254 @@
...
+                    }
+
+                    return string.Empty;
+
+                default:
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidIssuerException(LogMessages.IDX40010));
+            }
+        }
+
+        private static void EnforceSingleClaimCaseInsensitive(IEnumerable<string> keys, string claimType)
```
jmprieur: Same method and content in both

jmprieur (commented):
I've asked a few questions.
A lot of duplicated code in AadTokenValidationParametersExtension / AadValidationParametersExtension.Internal

test/Microsoft.IdentityModel.TestUtils/TestUtilities.cs
brentschmaltz: @jmprieur perhaps in the future, we aren't there right now.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
brentschmaltz: Yes shared between all instances.

src/Microsoft.IdentityModel.Validators/AadIssuerValidator/AadIssuerValidator.cs
brentschmaltz: Certainly, that can be done. Not the goal right now. The goal is to enable upper layers to make the calls.

src/Microsoft.IdentityModel.Tokens/TokenHandler.Internal.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    public abstract partial class TokenHandler
+    {
+        internal virtual Task<ValidationResult<ValidatedToken>> ValidateTokenAsync(
+            string token,
+            ValidationParameters validationParameters,
+            CallContext callContext,
+            CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
```
jmprieur: throw new NotImplementedExeption("Make sure you override this method in the derived class");

src/Microsoft.IdentityModel.Tokens/TokenHandler.Internal.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        internal virtual Task<ValidationResult<ValidatedToken>> ValidateTokenAsync(
+            SecurityToken token,
+            ValidationParameters validationParameters,
+            CallContext callContext,
+            CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
```
jmprieur: same?

src/Microsoft.IdentityModel.Validators/AadValidationParametersExtension.Internal.cs
brentschmaltz: The delegate is not async, so i think we have to add some additional work to make this happen.

keegan-caruso (approved):

jmprieur (approved):
LGTM
Thanks @brentschmaltz

src/Microsoft.IdentityModel.Tokens/TokenHandler.Internal.cs
brentschmaltz: Addressed

Issue #3082, opened 2025-01-08
[Bug] Behavioral change in ConfigurationManager


keegan-caruso:
#2780 introduced background refresh for metadata.

This is fine for automatic refresh, for RequestRefresh, the historical behavior should not change.

PR #3087, opened 2025-01-10
RequestRefresh back to a signal for GetConfigurationAsync

keegan-caruso:
Further reverting the change to RequestRefresh from #2780 This is still too much of a behavioral change. RequestRefresh goes back to its historical function, signaling that the next call to GetConfigurationAsync should request data. This is incorporated with the background thread change to instead do the request as a blocking operation if from a RequestRefresh

Part of #3082

When reviewing it will be helpful to look at #2780

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -38,7 +38,10 @@ namespace Microsoft.IdentityModel.Protocols
         private readonly TimeProvider _timeProvider = TimeProvider.System;
 
         // If a refresh is requested, then do the refresh as a blocking operation
-        // not on a background thread.
+        // not on a background thread. RequestRefresh signals that the app is explicitly
+        // requesting a refresh, so it should be done immediately so the next
+        // call to GetConfiguration will return new configuration if the minimum
+        // refresh interval has passed.
         bool _refreshRequested;
 
         /// <summary>
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -35,6 +35,10 @@ namespace Microsoft.IdentityModel.Protocols
         private const int ConfigurationRetrieverRunning = 1;
         private int _configurationRetrieverState = ConfigurationRetrieverIdle;
 
+        // If a refresh is requested, then do the refresh as a blocking operation
+        // not on a background thread.
+        bool _refreshRequested;
+
         /// <summary>
         /// Instantiates a new <see cref="ConfigurationManager{T}"/> that manages automatic and controls refreshing on configuration data.
         /// </summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -310,15 +320,11 @@ public override async Task<BaseConfiguration> GetBaseConfigurationAsync(Cancella
...
         {
             DateTimeOffset now = DateTimeOffset.UtcNow;
-
             if (now >= DateTimeUtil.Add(_lastRequestRefresh.UtcDateTime, RefreshInterval) || _isFirstRefreshRequest)
             {
                 _isFirstRefreshRequest = false;
-                if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
-                {
-                    UpdateCurrentConfiguration();
-                    _lastRequestRefresh = now;
```
keegan-caruso, line 320: This is off, and should have been a test failure. Will fix both.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -35,6 +35,12 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
         private const int ConfigurationRetrieverRunning = 1;
         private int _configurationRetrieverState = ConfigurationRetrieverIdle;
 
+        private readonly TimeProvider _timeProvider = TimeProvider.System;
+
+        // If a refresh is requested, then do the refresh as a blocking operation
+        // not on a background thread.
```
jennyf19: nit: maybe a bit more background on the rationale behind the change would be helpful for the future & the impact on the refresh logic.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -309,16 +321,13 @@ public override async Task<BaseConfiguration> GetBaseConfigurationAsync(Cancella
         /// </summary>
         public override void RequestRefresh()
```
jennyf19, line 325: are there any notable changes in perf with this change?

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -750,6 +718,96 @@ public void TestConfigurationComparer()
             TestUtilities.AssertFailIfErrors(context);
         }
 
+        [Fact]
+        public async Task RequestRefresh_RespectsRefreshInterval()
```
jennyf19, line 722: do we need to check for extremely short or long durations on refresh interval?

jennyf19 (approved):

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: Somewhat not straight forward. Yes, but the behavior was wrong before. Before this and the last change, RequestRefresh moved to a background thread, so there was no immediate impact on performance of the calling thread to queue a refresh. This was incorrect because our code and others write to the contract that if a RequestRefresh is called, and enough time has passed, new configuration will be returned by GetConfigurationAsync. If it is on a background thread, we lose that guarantee.

GeoK (approved):

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
keegan-caruso: Added a few more scenarios, used some better names.

PR #3090, opened 2025-01-11
Make updates to syncAfter and lastRefreshRequest atomic

### LINKED ISSUE #2900: [Bug] Race condition on ConfigurationManager._syncAfter

keegan-caruso:
# Make updates to syncAfter and lastRefreshRequest atomic

## Description

syncAfter can be updated on a background thread,
lastRefreshRequest could be simultaneously updated
by different threads as there is not a lock around
the check and update.

To be able to do an atomic write, DateTimeOffset fields had
to be switched to DateTime fields. DateTime is a struct
with a single ulong field, DateTimeOffset has two fields,
one for the datetime, one for the offset. DateTime can
be treated as a long for this reason (and is in DateTime's ToBinary method).

This also means ConfigurationManager can continue to use DateTime instead of an
int, which is much more ergonomic.

See the [implementation] for when you can't use a long and need to use a ulong: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/DateTime.cs#L1316

Fixes #2900

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Net.Http;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.IdentityModel.Logging;
@@ -18,8 +19,16 @@ namespace Microsoft.IdentityModel.Protocols
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable")]
     public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class
     {
-        private DateTimeOffset _syncAfter = DateTimeOffset.MinValue;
-        private DateTimeOffset _lastRequestRefresh = DateTimeOffset.MinValue;
+        // To prevent tearing, this needs to be only updated through AtomicUpdateSyncAfter.
+        // Reads should be done through the property SyncAfter.
+        private DateTime _syncAfter = DateTime.MinValue;
+        private DateTime SyncAfter => _syncAfter;
+
+        // See comment above, this should only be updated through AtomicUpdateLastRequestRefresh,
+        // read through LastRequestRefresh.
+        private DateTime _lastRequestRefresh = DateTime.MinValue;
+        private DateTime LastRequestRefresh => _lastRequestRefresh;
+
         private bool _isFirstRefreshRequest = true;
         private readonly SemaphoreSlim _configurationNullLock = new SemaphoreSlim(1);
 
@@ -156,7 +165,7 @@ namespace Microsoft.IdentityModel.Protocols
         /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
         public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
         {
-            if (_currentConfiguration != null && _syncAfter > _timeProvider.GetUtcNow())
+            if (_currentConfiguration != null && SyncAfter > _timeProvider.GetUtcNow())
                 return _currentConfiguration;
 
             Exception fetchMetadataFailure = null;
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -300,8 +309,27 @@ private void UpdateCurrentConfiguration()
...
                 TimeSpan.FromSeconds(new Random().Next((int)AutomaticRefreshInterval.TotalSeconds / 20)));
+            AtomicUpdateSyncAfter(newSyncTime);
+        }
+
+        private void AtomicUpdateSyncAfter(DateTime syncAfter)
+        {
+            // DateTime's backing data is safe to treat as a long if the Kind is not local.
+            // _syncAfter will always be updated to a UTC time.
+            // See the implementation of ToBinary on DateTime.
+            Interlocked.Exchange(
```
jennyf19, line 322: Have we looked at the performance impact of using `Interlocked.Exchange` for atomic updates under high load? Do we have test cases which simulate high-concurrency scenarios?

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-// Ignore Spelling: Metadata Validator Retreiver
```
msbw2, line 4: Move the property rename change to a separate PR. It makes the changes in this file more difficult to review.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
msbw2: @brentschmaltz

msbw2 (dismissed):
See comments.

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
keegan-caruso: This should be reviewed commit by commit, then the change is only within that one commit

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: Updating time shouldn't be done frequently, only on fetches and RequestRefreshes. I can still get some numbers. And we still need to do stress testing.

jennyf19: sounds good. Let's not block on merging. thanks @keegan-caruso

jennyf19 (approved):
:shipit:

GeoK (approved):

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTests.cs
keegan-caruso: It is done in a separate commit to help with this.

brentschmaltz (approved):

PR #3095, opened 2025-01-16
Add additional condition before requesting update.

brentschmaltz:
It was possible that request pass the guard: SyncAfter >_timeProvider.GetUtcNow() while a Refresh was underway. Before reaching the Interlock.CompareExchange, the Refresh finishes setting _configurationRetrieverState to ConfigurationRetrieverIdle and another refresh occurs.

This was found with a stress test with 128 tasks running concurrently and calling ConfigurationManger.GetConfigurationAsync.
The change was tested and the correct number of Refreshes were requested.

_refreshRequested was placed before UpdateCurrentConfiguration was called to reduce the chance of a race condition.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -269,8 +269,8 @@ namespace Microsoft.IdentityModel.Protocols
                         try
                         {
                             TelemetryClient.IncrementConfigurationRefreshRequestCounter(
-                            MetadataAddress,
-                            TelemetryConstants.Protocols.Automatic);
+                                MetadataAddress,
+                                TelemetryConstants.Protocols.Automatic);
                         }
 #pragma warning disable CA1031 // Do not catch general exception types
                         catch
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -249,18 +249,33 @@ namespace Microsoft.IdentityModel.Protocols
                     if (_refreshRequested)
                     {
                         _refreshRequested = false;
-                        // Log as manual because RequestRefresh was called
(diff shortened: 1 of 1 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -248,22 +248,26 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
                         UpdateCurrentConfiguration();
-                        _refreshRequested = false;
                     }
-                    else
+                    else if (SyncAfter <= _timeProvider.GetUtcNow())
                     {
                         TelemetryClient.IncrementConfigurationRefreshRequestCounter(
-                            MetadataAddress,
-                            TelemetryConstants.Protocols.Automatic);
+                        MetadataAddress,
```
westin-m: [Question] Why change the indentation on these parameters to the method call? It should match the similar call in the preceding `if`

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -248,22 +248,41 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
                     }
-                    else
+                    else if (SyncAfter <= _timeProvider.GetUtcNow())
                     {
-                        TelemetryClient.IncrementConfigurationRefreshRequestCounter(
+                        try
+                        {
+                            TelemetryClient.IncrementConfigurationRefreshRequestCounter(
                             MetadataAddress,
                             TelemetryConstants.Protocols.Automatic);
```
westin-m:
```suggestion
                                MetadataAddress,
                                TelemetryConstants.Protocols.Automatic);
```
For consistent indentation

brentschmaltz: @westin-m thanks, fixed.

keegan-caruso (approved):

westin-m (approved):

PR #3077, opened 2025-01-06
Adding end to end tests and doc for the new validation

jmprieur:
# end to end tests and doc for the new validation

The goal here is to understand the public API and the developer experience.
1. Updated the class diagrams:
2. Added end to end tests for the ValidationAsync (and found potential issues to discuss)

## ValidationParameters and their delegates
image

## ValidationErrors
image

PR #3100, opened 2025-01-21
Update error messages and bring E2E tests for the new validation model from the feature branch

iNinja:
# Update error messages and bring E2E tests for the new validation model from the feature branch
- Cherry picked E2E tests added in #3077 that were merged onto the feature branch
- Added new error messages for signature validation cases that were not accurately explained in the previous
- Updated signature validation and E2E tests to use the new error messages

Part of #2711


keegan-caruso (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests_e2e.cs
```diff
@@ -0,0 +1,433 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
jmprieur, line 1: where does this come from, @iNinja ?
Didn't we already have this code somewhere?

jmprieur (approved):
LGTM
Thanks @iNinja

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests_e2e.cs
iNinja: We have it in the feature branch. I'm bringing it into dev.

I'll hold the merge until we get to speak today to make sure there is no problem having this in dev.

iNinja: The current objective is to have the feature branch **only** differ in having the new validation model classes public, along with some annotations that are needed when those classes are public.

All functionality and tests I am aiming to have them be the same between dev and the feature branch.

PR #3102, opened 2025-01-22
 KeyID should be present in exception messages logs and should not be considered PII.

### LINKED ISSUE #3104: KeyID should be present in exception messages logs and should not be considered PII
Repair item from an incident and the auth client team needed more info, which the exception message did not include but should. KeyID is not PII.

kllysng:
Fixes #3104

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -77,7 +77,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     new InvalidOperationException(
                         LogHelper.FormatInvariant(
                             TokenLogMessages.IDX10637,
-                            LogHelper.MarkAsNonPII(signingCredentials.Key == null ? "Null" : signingCredentials.Key.ToString()),
+                            LogHelper.MarkAsNonPII(signingCredentials.Key == null ? "Null" : signingCredentials.Key.KeyId),
                             LogHelper.MarkAsNonPII(signingCredentials.Algorithm))));
 
             try
@@ -114,7 +114,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     new InvalidOperationException(
                         LogHelper.FormatInvariant(
                             TokenLogMessages.IDX10637,
-                            LogHelper.MarkAsNonPII(signingCredentials.Key == null ? "Null" : signingCredentials.Key.ToString()),
+                            LogHelper.MarkAsNonPII(signingCredentials.Key == null ? "Null" : signingCredentials.Key.KeyId),
                             LogHelper.MarkAsNonPII(signingCredentials.Algorithm))));
 
             try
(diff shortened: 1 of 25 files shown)


jmprieur (approved):
LGTM. Thanks @kllysng 
cc: @iNinja FYI

jmprieur (approved):

pmaytak (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -367,7 +367,7 @@ private static ValidationError GetDecryptionError(
                 return new ValidationError(
                     new MessageDetail(
                         TokenLogMessages.IDX10603,
-                        keysAttempted.ToString(),
+                        LogHelper.MarkAsNonPII(keysAttempted.ToString()),
```
keegan-caruso, line 370: Do we want this to only be the kids?

kllysng:
Updating this PR to include more than just the one log message.

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -365,15 +365,15 @@ internal static void ValidateIssuerSecurityKey(SecurityKey securityKey, Security
             if (validationParameters.IssuerSigningKeyValidatorUsingConfiguration != null)
             {
                 if (!validationParameters.IssuerSigningKeyValidatorUsingConfiguration(securityKey, securityToken, validationParameters, configuration))
-                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSigningKeyException(LogHelper.FormatInvariant(LogMessages.IDX10232, securityKey)) { SigningKey = securityKey });
+                    throw LogHelper.LogExceptionMessage(new SecurityTokenInvalidSigningKeyException(LogHelper.FormatInvariant(LogMessages.IDX10232, LogHelper.MarkAsNonPII(securityKey.ToString()))) { SigningKey = securityKey });
```
kllysng: @GeoK double checking that logging the Kid and InternaId is acceptable? [SecurtiyKey.ToString()](https://learn.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.tokens.securitykey.tostring?view=msal-web-dotnet-latest#microsoft-identitymodel-tokens-securitykey-tostring) returns the formatted string: GetType(), KeyId: 'value', InternalId: 'value'.

kllysng: Otherwise, I can log Key.Kid + change all the log messages from `.. key: {0}..` to `.. key Kid: {0}..`

GeoK: Was the same returned previously? The difference is that now we are not treating it as PII.

kllysng: Can confirm that while debugging, logging `SecurityKey` and `SecurityKey.ToString()` both invoke .ToString() and have the same output.

kllysng: is it acceptable to also log `InternalId`?

westin-m (approved):

armandRobled:
Robledo6-patche

Luis Armando Robledo salazar

src/Microsoft.IdentityModel.Tokens/Validators.cs
kllysng: Discussed in internally offline

westin-m (approved):

PR #3105, opened 2025-01-28
Experimental interface for new validation model

iNinja:
# Experimental interface for new validation model
- Added public experimental interface to expose new validation model methods in `JsonWebTokenHandler`, `SamlSecurityTokenHandler`, and `Saml2SecurityTokenHandler`
- Sorted unshipped API txt files

Note: The `ValidateTokenAsync` methods can be made private now, but initially leaving them internal to avoid breaking existing tests and experimental consumers with internal access.

Part of #2711


jmprieur (approved):
LGTM
you probably want to update the E2E tests to show how the feature would be used?

PR #3106, opened 2025-01-28
App context switch to allow blocking or non-blocking GetConfiguration

### LINKED ISSUE #3082: [Bug] Behavioral change in ConfigurationManager

keegan-caruso:
# App context switch to allow blocking or non-blocking GetConfiguration

Stress testing showed that the current non-blocking implementation wasn't always triggered as expected, moving to an AutoResetEvent fixed this.

How Configuration is Updated Now:
Asynchronous Updates: The configuration can be updated asynchronously using background tasks. The method. The method to trigger the background task was updated to a AutoResetEvent.

Blocking vs Non-Blocking: The AppContextSwitches.UpdateConfigAsBlocking switch determines whether the configuration update should block other requests. If this switch is enabled, the method RequestRefreshBlocking is called; otherwise, RequestRefreshBackgroundThread is used.

Telemetry: The method TelemetryForUpdate handles logging the mode of update (manual or automatic) to a telemetry client.
These changes aim to improve the efficiency and responsiveness of the configuration update process by leveraging background tasks and signals.

Adds ShutdownBackgroundTask() method to ConfigurationManager to stop the background task.

Introduce ResetAppContextSwitches to have deterministic reset per test that uses an app context, this also serializes tests.

Adds 

Fixes #3082

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -18,17 +18,21 @@ namespace Microsoft.IdentityModel.Protocols
     /// </summary>
     /// <typeparam name="T">The type of <see cref="IDocumentRetriever"/>.</typeparam>
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable")]
-    public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class
+    public partial class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class
     {
-        // To prevent tearing, this needs to be only updated through AtomicUpdateSyncAfter.
-        // Reads should be done through the property SyncAfter.
         private DateTime _syncAfter = DateTime.MinValue;
-        private DateTime SyncAfter => _syncAfter;
+        private DateTime SyncAfter
+        {
+            get => _syncAfter;
+            set => AtomicUpdateDateTime(ref _syncAfter, ref value);
+        }
 
-        // See comment above, this should only be updated through AtomicUpdateLastRequestRefresh,
-        // read through LastRequestRefresh.
         private DateTime _lastRequestRefresh = DateTime.MinValue;
-        private DateTime LastRequestRefresh => _lastRequestRefresh;
+        private DateTime LastRequestRefresh
+        {
+            get => _lastRequestRefresh;
+            set => AtomicUpdateDateTime(ref _lastRequestRefresh, ref value);
+        }
 
         private bool _isFirstRefreshRequest = true;
         private readonly SemaphoreSlim _configurationNullLock = new SemaphoreSlim(1);
@@ -55,6 +59,8 @@ namespace Microsoft.IdentityModel.Protocols
         // refresh interval has passed.
         bool _refreshRequested;
 
+        private readonly AutoResetEvent _updateMetadata = new(false);
+        Task _updateMetadataTask;
 
         /// <summary>
         /// Instantiates a new <see cref="ConfigurationManager{T}"/> that manages automatic and controls refreshing on configuration data.
@@ -117,6 +123,7 @@ namespace Microsoft.IdentityModel.Protocols
             MetadataAddress = metadataAddress;
             _docRetriever = docRetriever;
             _configRetriever = configRetriever;
+            EnsureBackgroundTaskIsRunning();
         }
 
         /// <summary>
@@ -171,6 +178,14 @@ namespace Microsoft.IdentityModel.Protocols
             if (_currentConfiguration != null && SyncAfter > _timeProvider.GetUtcNow())
                 return _currentConfiguration;
 
+            if (AppContextSwitches.UpdateConfigAsBlocking)
+                return await GetConfigurationWithBlockingAsync(cancel).ConfigureAwait(false);
+            else
+                return await GetConfigurationWithBackgroundTaskUpdatesAsync(cancel).ConfigureAwait(false);
+        }
+
+        private async Task<T> GetConfigurationWithBackgroundTaskUpdatesAsync(CancellationToken cancel)
+        {
             Exception fetchMetadataFailure = null;
 
             // LOGIC
@@ -246,38 +261,10 @@ namespace Microsoft.IdentityModel.Protocols
             {
                 if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
                 {
-                    if (_refreshRequested)
+                    if (SyncAfter <= _timeProvider.GetUtcNow())
                     {
-                        _refreshRequested = false;
-
-                        try
-                        {
-                            // Log as manual because RequestRefresh was called
-                            TelemetryClient.IncrementConfigurationRefreshRequestCounter(
-                                MetadataAddress,
-                                TelemetryConstants.Protocols.Manual);
-                        }
-#pragma warning disable CA1031 // Do not catch general exception types
-                        catch
-                        { }
-#pragma warning restore CA1031 // Do not catch general exception types
-
-                        UpdateCurrentConfiguration();
-                    }
-                    else if (SyncAfter <= _timeProvider.GetUtcNow())
-                    {
-                        try
-                        {
-                            TelemetryClient.IncrementConfigurationRefreshRequestCounter(
-                                MetadataAddress,
-                                TelemetryConstants.Protocols.Automatic);
-                        }
-#pragma warning disable CA1031 // Do not catch general exception types
-                        catch
-                        { }
-#pragma warning restore CA1031 // Do not catch general exception types
-
-                        _ = Task.Run(UpdateCurrentConfiguration, CancellationToken.None);
+                        EnsureBackgroundTaskIsRunning();
+                        _updateMetadata.Set();
                     }
                     else
                     {
@@ -300,6 +287,40 @@ namespace Microsoft.IdentityModel.Protocols
                     fetchMetadataFailure));
         }
 
+        private void EnsureBackgroundTaskIsRunning()
+        {
+            if (_updateMetadataTask == null || _updateMetadataTask.Status != TaskStatus.Running)
+                _updateMetadataTask = Task.Run(UpdateCurrentConfigurationUsingSignals, CancellationToken.None);
+        }
+
+        private void TelemetryForUpdate()
+        {
+            var updateMode = _refreshRequested ? TelemetryConstants.Protocols.Manual : TelemetryConstants.Protocols.Automatic;
+
+            if (_refreshRequested)
+                _refreshRequested = false;
+
+            try
+            {
+                TelemetryClient.IncrementConfigurationRefreshRequestCounter(
+                    MetadataAddress,
+                    updateMode);
+            }
+#pragma warning disable CA1031 // Do not catch general exception types
+            catch
+            { }
+#pragma warning restore CA1031 // Do not catch general exception types
+        }
+
+        private void UpdateCurrentConfigurationUsingSignals()
+        {
+            while (true)
+            {
+                _updateMetadata.WaitOne();
+                UpdateCurrentConfiguration();
+            }
+        }
+
         /// <summary>
         /// This should be called when the configuration needs to be updated either from RequestRefresh or AutomaticRefresh
         /// The Caller should first check the state checking state using:
@@ -307,6 +328,7 @@ namespace Microsoft.IdentityModel.Protocols
         /// </summary>
         private void UpdateCurrentConfiguration()
         {
+            TelemetryForUpdate();
 #pragma warning disable CA1031 // Do not catch general exception types
             long startTimestamp = _timeProvider.GetTimestamp();
 
@@ -368,25 +390,17 @@ namespace Microsoft.IdentityModel.Protocols
             _currentConfiguration = configuration;
             var newSyncTime = DateTimeUtil.Add(_timeProvider.GetUtcNow().UtcDateTime, AutomaticRefreshInterval +
                 TimeSpan.FromSeconds(new Random().Next((int)AutomaticRefreshInterval.TotalSeconds / 20)));
-            AtomicUpdateSyncAfter(newSyncTime);
+            SyncAfter = newSyncTime;
         }
 
-        private void AtomicUpdateSyncAfter(DateTime syncAfter)
+        private static void AtomicUpdateDateTime(ref DateTime field, ref DateTime value)
         {
             // DateTime's backing data is safe to treat as a long if the Kind is not local.
-            // _syncAfter will always be updated to a UTC time.
+            // time will always be updated to a UTC time.
             // See the implementation of ToBinary on DateTime.
             Interlocked.Exchange(
-                ref Unsafe.As<DateTime, long>(ref _syncAfter),
-                Unsafe.As<DateTime, long>(ref syncAfter));
-        }
-
-        private void AtomicUpdateLastRequestRefresh(DateTime lastRequestRefresh)
-        {
-            // See the comment in AtomicUpdateSyncAfter.
-            Interlocked.Exchange(
-                ref Unsafe.As<DateTime, long>(ref _lastRequestRefresh),
-                Unsafe.As<DateTime, long>(ref lastRequestRefresh));
+                ref Unsafe.As<DateTime, long>(ref field),
+                Unsafe.As<DateTime, long>(ref value));
         }
 
         /// <summary>
@@ -409,14 +423,30 @@ namespace Microsoft.IdentityModel.Protocols
         /// </summary>
         public override void RequestRefresh()
         {
+            if (AppContextSwitches.UpdateConfigAsBlocking)
+            {
+                RequestRefreshBlocking();
+            }
+            else
+            {
+                RequestRefreshBackgroundThread();
+            }
+        }
+
+        private void RequestRefreshBackgroundThread()
+        {
             DateTime now = _timeProvider.GetUtcNow().UtcDateTime;
 
             if (now >= DateTimeUtil.Add(LastRequestRefresh, RefreshInterval) || _isFirstRefreshRequest)
             {
                 _isFirstRefreshRequest = false;
-                AtomicUpdateSyncAfter(now);
-                AtomicUpdateLastRequestRefresh(now);
-                _refreshRequested = true;
+
+                if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
+                {
+                    _refreshRequested = true;
+                    LastRequestRefresh = now;
+                    _updateMetadata.Set();
+                }
             }
         }
 
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -22,7 +22,7 @@ namespace Microsoft.IdentityModel.Protocols
     {
 #pragma warning disable IDE0044 // Add readonly modifier
 #pragma warning disable CS0649 // Unused, it gets used in tests.
-        private Action _onBackgroundTaskFinish;
+        internal Action _onBackgroundTaskFinish;
 #pragma warning restore CS0649 // Unused
 #pragma warning restore IDE0044 // Add readonly modifier
 
@@ -65,7 +65,13 @@ namespace Microsoft.IdentityModel.Protocols
         // refresh interval has passed.
         bool _refreshRequested;
 
-        private readonly AutoResetEvent _updateMetadata = new(false);
+        // Wait handle used to signal a background task to update the configuration.
+        // Handle starts unset, and AutoResetEvent.Set() sets it, this indicates that
+        // the background refresh task should immediately run.
+        private readonly AutoResetEvent _updateMetadataEvent = new(false);
+
+        // Background task that updates the configuration. Signaled with _updateMetadataEvent.
+        // Task should be started with EnsureBackgroundRefreshTaskIsRunning.
         Task _updateMetadataTask;
 
         /// <summary>
@@ -74,7 +80,7 @@ namespace Microsoft.IdentityModel.Protocols
         /// then this will not be used.
         /// Note that this does not influence <see cref="GetConfigurationAsync(CancellationToken)"/>.
         /// </summary>
-        public CancellationToken BackgroundTaskCancellationToken { get; set; }
+        public CancellationToken BackgroundRefreshTaskCancellationToken { get; set; }
 
         /// <summary>
         /// Instantiates a new <see cref="ConfigurationManager{T}"/> that manages automatic and controls refreshing on configuration data.
@@ -139,7 +145,7 @@ namespace Microsoft.IdentityModel.Protocols
             _configRetriever = configRetriever;
 
             if (!AppContextSwitches.UpdateConfigAsBlocking)
-                EnsureBackgroundTaskIsRunning();
+                EnsureBackgroundRefreshTaskIsRunning();
         }
 
         /// <summary>
@@ -279,8 +285,8 @@ namespace Microsoft.IdentityModel.Protocols
                 {
                     if (SyncAfter <= _timeProvider.GetUtcNow())
                     {
-                        EnsureBackgroundTaskIsRunning();
-                        _updateMetadata.Set();
+                        EnsureBackgroundRefreshTaskIsRunning();
+                        _updateMetadataEvent.Set();
                     }
                     else
                     {
@@ -303,9 +309,9 @@ namespace Microsoft.IdentityModel.Protocols
                     fetchMetadataFailure));
         }
 
-        private void EnsureBackgroundTaskIsRunning()
+        private void EnsureBackgroundRefreshTaskIsRunning()
         {
-            if (BackgroundTaskCancellationToken.IsCancellationRequested)
+            if (BackgroundRefreshTaskCancellationToken.IsCancellationRequested)
                 return;
 
             if (_updateMetadataTask == null || _updateMetadataTask.Status != TaskStatus.Running)
@@ -335,9 +341,9 @@ namespace Microsoft.IdentityModel.Protocols
         {
             try
             {
-                while (!BackgroundTaskCancellationToken.IsCancellationRequested)
+                while (!BackgroundRefreshTaskCancellationToken.IsCancellationRequested)
                 {
-                    if (_updateMetadata.WaitOne(500))
+                    if (_updateMetadataEvent.WaitOne(500))
                     {
                         UpdateCurrentConfiguration();
                         _onBackgroundTaskFinish?.Invoke();
@@ -473,8 +479,8 @@ namespace Microsoft.IdentityModel.Protocols
                 {
                     _refreshRequested = true;
                     LastRequestRefresh = now;
-                    EnsureBackgroundTaskIsRunning();
-                    _updateMetadata.Set();
+                    EnsureBackgroundRefreshTaskIsRunning();
+                    _updateMetadataEvent.Set();
                 }
             }
         }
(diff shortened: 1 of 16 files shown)


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+        }
+
+        private void TelemetryForUpdateBlocking()
+        {
+            string updateMode;
+
+            if (_currentConfiguration is null)
+            {
+                updateMode = TelemetryConstants.Protocols.FirstRefresh;
```
GeoK, line 139: Is there value in tracking a separate set of update modes for blocking ops?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -408,15 +429,31 @@ public override async Task<BaseConfiguration> GetBaseConfigurationAsync(Cancella
...
                 _isFirstRefreshRequest = false;
-                AtomicUpdateSyncAfter(now);
-                AtomicUpdateLastRequestRefresh(now);
-                _refreshRequested = true;
+
+                if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
+                {
+                    _refreshRequested = true;
+                    LastRequestRefresh = now;
+                    _updateMetadata.Set();
```
GeoK: On ln272 you are calling EnsureBackgroundTaskIsRunning before signaling. Should you do the same here?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -300,13 +293,49 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+#pragma warning restore CA1031 // Do not catch general exception types
+        }
+
+        private void UpdateCurrentConfigurationUsingSignals()
+        {
+            while (true)
+            {
+                _updateMetadata.WaitOne();
+                UpdateCurrentConfiguration();
+                _onBackgroundTaskFinish?.Invoke();
```
GeoK: This is a bit unfortunate, but the motivation is understandable.

keegan-caruso: I agree, I don't like have code here only for testing. But this allows us to have stable tests, without more invasive changes.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
keegan-caruso: It would help distinguish if one or the other is failing, so I think so yes.

brentschmaltz: Could we put that in the Tag?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Threading;
+using System;
+using Microsoft.IdentityModel.Logging;
+using Microsoft.IdentityModel.Protocols.Configuration;
+using Microsoft.IdentityModel.Tokens;
+using Microsoft.IdentityModel.Telemetry;
+
+namespace Microsoft.IdentityModel.Protocols
+{
+    partial class ConfigurationManager<T> where T : class
```
brentschmaltz, line 14: Does it matter if this definition is different than the other?
partial class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
```diff
@@ -0,0 +1,168 @@
...
+    {
+        private readonly SemaphoreSlim _refreshLock = new(1);
+
+        private TimeSpan _bootstrapRefreshInterval = TimeSpan.FromSeconds(1);
+
+        private DateTime _lastRefresh = DateTime.MinValue;
+        private DateTime LastRefresh
```
brentschmaltz: Do we need both properties 'LastRefresh' and 'RequestLastRefresh' ?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -117,6 +131,7 @@ public ConfigurationManager(string metadataAddress, IConfigurationRetriever<T> c
             MetadataAddress = metadataAddress;
             _docRetriever = docRetriever;
             _configRetriever = configRetriever;
+            EnsureBackgroundTaskIsRunning();
```
brentschmaltz: If the AppContextSwitch directs the runtime to use blocking mode, should we start the task here?

Will EnsureBackgroundTaskIsRunning start the task?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
keegan-caruso: This is the minimum amount needed here.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
pmaytak, line 1: Can you please add more details to the PR description - how does config manager work in general now?

Answer from copilot (if it's accurate):

> How Configuration is Updated Now:
> Asynchronous Updates: The configuration can now be updated asynchronously using background tasks. The method GetConfigurationWithBackgroundTaskUpdatesAsync is called to handle configuration retrieval with background task updates.
> Signals and Background Tasks: The method UpdateCurrentConfigurationUsingSignals runs in a loop and waits for the _updateMetadata signal to update the configuration. This method ensures that the configuration is refreshed without blocking the main thread.
> Blocking vs Non-Blocking: The AppContextSwitches.UpdateConfigAsBlocking switch determines whether the configuration update should block other requests. If this switch is enabled, the method RequestRefreshBlocking is called; otherwise, RequestRefreshBackgroundThread is used.
> Telemetry: The method TelemetryForUpdate handles logging the mode of update (manual or automatic) to a telemetry client.
> These changes aim to improve the efficiency and responsiveness of the configuration update process by leveraging background tasks and signals, reducing potential blocking of the main thread.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -18,17 +18,29 @@ namespace Microsoft.IdentityModel.Protocols
...
     /// <typeparam name="T">The type of <see cref="IDocumentRetriever"/>.</typeparam>
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable")]
-    public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class
+    public partial class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class
     {
-        // To prevent tearing, this needs to be only updated through AtomicUpdateSyncAfter.
-        // Reads should be done through the property SyncAfter.
+#pragma warning disable IDE0044 // Add readonly modifier
+#pragma warning disable CS0649 // Unused, it gets used in tests.
+        private Action _onBackgroundTaskFinish;
```
pmaytak: Can this be made internal? I think internal represent the use scope of the variable more accurately, even if it's just tests. I'm kind of not a fan of using reflection in tests to deal with private members.

src/Microsoft.IdentityModel.Protocols/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,2 @@
+Microsoft.IdentityModel.Protocols.ConfigurationManager<T>.BackgroundTaskCancellationToken.get -> System.Threading.CancellationToken
```
keegan-caruso: @jmprieur Heads up on this

keegan-caruso: We need a way to shut down the background task. This seemed like the least invasive way. We could also add a shutdown method or make the type disposable.

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -70,6 +70,17 @@ internal static class AppContextSwitches
 
         internal static bool UseRfcDefinitionOfEpkAndKid => _useRfcDefinitionOfEpkAndKid ??= (AppContext.TryGetSwitch(UseRfcDefinitionOfEpkAndKidSwitch, out bool isEnabled) && isEnabled);
 
+        /// <summary>
+        /// Enabling this switch will cause the configuration manager to block other requests to GetConfigurationAsync if a request is already in progress.
+        /// The default behavior is if a request is already in progress, the current configuration will be returned until the request is completed on
```
pmaytak:
```suggestion
        /// The default configuration refresh behavior is if a request is already in progress, the current configuration will be returned until the ongoing request is completed on
```

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
pmaytak, line 1: Please update https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/App-Context-Switches-in-IdentityModel

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -55,6 +65,16 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
         // refresh interval has passed.
         bool _refreshRequested;
 
+        private readonly AutoResetEvent _updateMetadata = new(false);
+        Task _updateMetadataTask;
```
pmaytak: nit: this class/logic is more on a complex side, opting out here for more comments

```suggestion
         // Synchronization mechanism used to ensure that only one Task is updating the configuration.
        // By default, this AutoResetEvent is unsignaled.
        // The first and the following threads will wait until AutoResetEvent.Set is called first.       
        private readonly AutoResetEvent _updateMetadata = new(false);
        // Task that updates the configuration.
        Task _updateMetadataTask;
```

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -55,6 +65,16 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
...
+
+        /// <summary>
+        /// Cancellation token to control cancelling the background task.
+        /// If 'Switch.Microsoft.IdentityModel.UpdateConfigAsBlocking' is set to true,
+        /// then this will not be used.
+        /// Note that this does not influence <see cref="GetConfigurationAsync(CancellationToken)"/>.
+        /// </summary>
+        public CancellationToken BackgroundTaskCancellationToken { get; set; }
```
pmaytak: `BackgroundRefreshTaskCancellationToken`? `BackgroundConfigRefreshTaskCancellationToken`?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -300,13 +303,60 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
                     fetchMetadataFailure));
         }
 
+        private void EnsureBackgroundTaskIsRunning()
```
pmaytak: `EnsureBackgroundRefreshTaskIsRunning`? `EnsureBackgroundConfigRefreshTaskIsRunning`?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -55,6 +65,16 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
         // refresh interval has passed.
         bool _refreshRequested;
 
+        private readonly AutoResetEvent _updateMetadata = new(false);
```
pmaytak: `_updateMetadataEvent`?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -300,13 +303,60 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+                {
+                    if (_updateMetadata.WaitOne(500))
+                    {
+                        UpdateCurrentConfiguration();
+                        _onBackgroundTaskFinish?.Invoke();
+                    }
+                }
+            }
+            catch (Exception)
+            {
```
pmaytak, line 375: Should we log something here? For ex, if the event gets into a faulted state somehow, we won't know.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -300,13 +303,60 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+        }
+
         /// <summary>
         /// This should be called when the configuration needs to be updated either from RequestRefresh or AutomaticRefresh
         /// The Caller should first check the state checking state using:
         ///   if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle).
         /// </summary>
         private void UpdateCurrentConfiguration()
         {
+            TelemetryForUpdate();
```
pmaytak, line 387: `TelemetryForUpdate` is only used here, so maybe no need for the separate method, just move the code here, so it's easier to read?

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
keegan-caruso: Yes, after this merges

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: I'd rather keep telemetry related functions separate. It isn't async so the overhead should be negligible.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: We don't have a logger here, might be a good place for another telemetry signal. @GeoK ?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
keegan-caruso: Tag was a good fit. Can read it from the environment in right where we create the tags

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: I added this, it makes sense

src/Microsoft.IdentityModel.Protocols/PublicAPI.Unshipped.txt
jmprieur: I guess having a cancelation token on RequestRefresh() would be a breaking change. Would it be the right way of doing? probably not. 
How would be the set method be used? Set a cancellation token and cancel it?

keegan-caruso: @jmprieur The most idiomatic way would be to make the type disposable. But it is little bit of a grey area as the type should be used as a singleton in almost all cases. We have to have some way to cancel the task. If you are okay with making this disposable, I am happy to make that change instead.

jmprieur: Maybe a ShutdownBackgroundTask() method?

jmprieur: or the cancellation token

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -18,17 +18,27 @@ namespace Microsoft.IdentityModel.Protocols
...
     /// <typeparam name="T">The type of <see cref="IDocumentRetriever"/>.</typeparam>
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable")]
-    public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class
+    public partial class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationManager<T> where T : class
     {
-        // To prevent tearing, this needs to be only updated through AtomicUpdateSyncAfter.
-        // Reads should be done through the property SyncAfter.
+#pragma warning disable IDE0044 // Add readonly modifier
+#pragma warning disable CS0649 // Unused, it gets used in tests.
+        internal Action _onBackgroundTaskFinish;
```
pmaytak, line 25: I suggest that internals should have the same casing as public members, since we kind of treat internals and publics similar related to breaking changes. Also seeing an underscore make it easier to think about code since that variable is private and not used anywhere else.

We aren't really consistent with this naming throughout our SDKs though (using both underscore and Pascal casing). We could update our rules to issue a warning, but that will probably require suppressions, since we wouldn't want to rename internal members. Thoughts?

```suggestion
        internal Action OnBackgroundTaskFinish;
```

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -55,6 +65,26 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
...
         bool _refreshRequested;
 
+        // Wait handle used to signal a background task to update the configuration.
+        // Handle starts unset, and AutoResetEvent.Set() sets it, this indicates that
+        // the background refresh task should immediately run.
+        private readonly AutoResetEvent _updateMetadataEvent = new(false);
+
+        // Background task that updates the configuration. Signaled with _updateMetadataEvent.
+        // Task should be started with EnsureBackgroundRefreshTaskIsRunning.
+        Task _updateMetadataTask;
```
pmaytak:
```suggestion
        private Task _updateMetadataTask;
```

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -55,6 +65,26 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
...
+
+        private readonly CancellationTokenSource _backgroundRefreshTaskCancellationTokenSource;
+
+        /// <summary>
+        /// Requests that background tasks be shutdown.
+        /// This only applies if 'Switch.Microsoft.IdentityModel.UpdateConfigAsBlocking' is set to true.
```
pmaytak:
```suggestion
        /// This only applies if 'Switch.Microsoft.IdentityModel.UpdateConfigAsBlocking' is not set or set to false.
```

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -55,6 +65,26 @@ public class ConfigurationManager<T> : BaseConfigurationManager, IConfigurationM
...
+        /// Note that this does not influence <see cref="GetConfigurationAsync(CancellationToken)"/>.
+        /// </summary>
```
pmaytak, line 87:
```suggestion
        /// Background refresh task cannot be restarted. New instance of ConfigurationManager must be created instead.
        /// </summary>
```

src/Microsoft.IdentityModel.Tokens/Telemetry/ITelemetryClient.cs
```diff
@@ -24,5 +24,9 @@ internal void IncrementConfigurationRefreshRequestCounter(
             string metadataAddress,
             string operationStatus,
             Exception exception);
+
+        internal void LogBackgroundRefreshFailure(
```
pmaytak:
```suggestion
        internal void LogBackgroundConfigurationRefreshFailure(
```

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryClient.cs
```diff
@@ -14,13 +16,19 @@ internal class TelemetryClient : ITelemetryClient
     {
         public string ClientVer = IdentityModelTelemetryUtil.ClientVer;
 
+        private KeyValuePair<string, object> _BlockingTagValue = new(
```
pmaytak:
```suggestion
        private KeyValuePair<string, object> _blockingTagValue = new(
```

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryClient.cs
```diff
@@ -57,11 +67,27 @@ public void LogConfigurationRetrievalDuration(string metadataAddress, TimeSpan o
...
-                { TelemetryConstants.ExceptionTypeTag, exception.GetType().ToString() }
+                { TelemetryConstants.ExceptionTypeTag, exception.GetType().ToString() },
+                _BlockingTagValue
             };
 
             long durationInMilliseconds = (long)operationDuration.TotalMilliseconds;
             TelemetryDataRecorder.RecordConfigurationRetrievalDurationHistogram(durationInMilliseconds, tagList);
         }
+
+        public void LogBackgroundRefreshFailure(
```
pmaytak:
```suggestion
        public void LogBackgroundConfigurationRefreshFailure(
```

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryDataRecorder.cs
```diff
@@ -26,9 +26,16 @@ internal class TelemetryDataRecorder
...
         internal const string IdentityModelConfigurationManagerCounterName = "IdentityModelConfigurationManager";
         internal const string IdentityModelConfigurationManagerCounterDescription = "Counter capturing configuration manager operations.";
-        internal static readonly Counter<long> ConfigurationManagerCounter = IdentityModelMeter.CreateCounter<long>(IdentityModelConfigurationManagerCounterName, description: IdentityModelConfigurationManagerCounterDescription);
+
+        /// <summary>
+        /// Counter to capture background refresh failures in the ConfigurationManager.
+        /// </summary>
+        internal static readonly Counter<long> BackgroundRefreshFailureCounter = IdentityModelMeter.CreateCounter<long>(BackgroundRefreshFailureCounterName, description: BackgroundRefreshFailureCounterDescription);
+        internal const string BackgroundRefreshFailureCounterName = "IdentityModelConfigurationManagerBackgroundRefreshFailure";
+        internal const string BackgroundRefreshFailureCounterDescription = "Counter capturing configuration manager background refresh failures.";
```
pmaytak: I think it'd be clearer to specify what the refresh refers to.
```suggestion
        internal static readonly Counter<long> BackgroundConfigurationRefreshFailureCounter = IdentityModelMeter.CreateCounter<long>(BackgroundConfigurationRefreshFailureCounterName, description: BackgroundConfigurationRefreshFailureCounterDescription);
        internal const string BackgroundConfigurationRefreshFailureCounterName = "IdentityModelConfigurationManagerBackgroundRefreshFailure";
        internal const string BackgroundConfigurationRefreshFailureCounterDescription = "Counter capturing configuration manager background refresh failures.";
```

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryDataRecorder.cs
```diff
@@ -47,5 +54,10 @@ internal static void IncrementConfigurationRefreshRequestCounter(in TagList tagL
         {
             ConfigurationManagerCounter.Add(1, tagList);
         }
+
+        internal static void IncrementConfigurationBackgroundRefreshFailureCounter(in TagList tagList)
```
pmaytak:
```suggestion
        internal static void IncrementBackgroundConfigurationRefreshFailureCounter(in TagList tagList)
```

src/Microsoft.IdentityModel.Protocols/PublicAPI.Unshipped.txt
keegan-caruso: ShutdownBackgroundTask fits well here.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: Ha, in this case, not having the underscore raises an error, would rather not suppress it here.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
keegan-caruso: This isn't quite true, if it stops, we restart it. I will add that to this description.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
```diff
@@ -0,0 +1,161 @@
...
+    partial class ConfigurationManager<T> where T : class
+    {
+        private readonly SemaphoreSlim _refreshLock = new(1);
+
+        private TimeSpan _bootstrapRefreshInterval = TimeSpan.FromSeconds(1);
+
+        private async Task<T> GetConfigurationWithBlockingAsync(CancellationToken cancel)
+        {
+            Exception _fetchMetadataFailure = null;
+            await _refreshLock.WaitAsync(cancel).ConfigureAwait(false);
```
GeoK, line 23: 8.0.1 has the opportunistic check on time to refresh before the lock. Is the re-ordering intentional (e.g. sync after update)?

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTelemetryTests.cs
```diff
@@ -6,20 +6,36 @@
...
-using Microsoft.IdentityModel.TestUtils;
 using Microsoft.IdentityModel.Telemetry;
 using Microsoft.IdentityModel.Telemetry.Tests;
+using Microsoft.IdentityModel.TestUtils;
+using Microsoft.IdentityModel.Tokens;
 using Xunit;
 
 namespace Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests
 {
+    [ResetAppContextSwitches]
```
GeoK, line 21: This is interesting. I wonder if there is a way to apply the behavior across all tests in the codebase as tests withing a collection are executed sequentially only within an assembly.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
keegan-caruso: It was moved to the method calling this

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/4458df8390452d53bbaac6406908bf6c6dd96e69/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs#L209

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerTelemetryTests.cs
keegan-caruso: Since it is static state, unfortunately, it at least needs to be sequential for this to be stable.

GeoK (approved):

pmaytak (approved):

PR #3107, opened 2025-01-29
New validation model: Fixed stack frames being created when already cached

iNinja:
# New validation model: Fixed stack frames being created when already cached
Addressed a bug in the `GetCurrentStackFrame` method in the `ValidationError.cs` file that caused stack frames to be created even when they were already cached.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
@@ -288,16 +288,9 @@ namespace Microsoft.IdentityModel.Tokens
             // String is allocated, but it goes out of scope immediately after the call
             string key = filePath + lineNumber;
 
-            // Try to get the stack frame from the cache,
-            // otherwise create a new one and add it to the cache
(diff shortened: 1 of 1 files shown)


src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
```diff
@@ -287,8 +287,17 @@ public static StackFrame GetCurrentStackFrame(
         {
             // String is allocated, but it goes out of scope immediately after the call
             string key = filePath + lineNumber;
-            StackFrame frame = CachedStackFrames.GetOrAdd(key, new StackFrame(skipFrames, true));
```
jmprieur, line 290: Another way would be to use the override with a delegate.

jmprieur (approved):

brentschmaltz (approved):

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
iNinja: Updated to use this instead.

jmprieur (approved):

PR #3108, opened 2025-01-30
New validation model:  Preview docs

iNinja:
# New validation model:  Preview docs
- Added extra benchmark to highlight the performance difference in failure scenarios.
- Added initial markdown document for the new validation model.
- Updated E2E tests to access the new validation method through the experimental interface

Part of #2711

### DIFF
--- a/docs/new-token-validation-model.md
+++ b/docs/new-token-validation-model.md
@@ -1,0 +1,86 @@
+# New token validation model
+The existing token validation model on `Microsoft.IdentityModel` relies on the exception based control flow dominant in C#. In some high-performance scenarios where latency is key, this can cause unintended increases when validations fail and throw.  
+
+As part of an effort to modernize aspects of the library that have become bloated over the years, we are introducing an alternative token validation model that breaks away from some of the previous behaviors and intends to offer a slimmer and more performant API.  
+
+## Key points on the new model
+- Performance increase by removing unnecessary exception throwing within the validation itself.
+- Return a result where it is clear whether the validation was successful, including relevant information depending on the case.
+  - In success scenarios, provide the validated information for audit purposes.
+  - In failure scenarios, provide enough information to identify what part of the token was invalid or could not be validated without the need for further steps, and for diagnostic purposes provide the ability to create an exception that can be thrown and observed. 
+- Do not automatically print logs as part of the validation. These can be printed from the validation result using C#'s [high-performance logging](https://learn.microsoft.com/en-us/dotnet/core/extensions/high-performance-logging) based on `ILogger`.
+- Simpler validation parameters object, as the current one has grown in complexity over the years by including multiple parameters for things like `ValidIssuer` and `ValidIssuers`, or multiple delegates for the same validation. `ValidationParameters` is being introduced to offer a simplified object that enables simplifying the default validation code as well.
+- New APIs provide nullability annotations to simplify code branches when the API can ensure no `null` will be returned from a method.
+- New APIs receive a `CancellationToken` to allow for the cancellation of running validation operations.
+
+## Embracing the Result pattern to remove exceptions on a hot path
+In order to remove the exceptions being thrown, we are introducing a `ValidationResult` type to handle the result of all validation operations.  
+This type can be thought of as a **one of** type as `ValidationResult<TResult, TError>`, though the `TError` type is initially being fixed to `ValidationError`.  
+
+The `ValidationResult` object can only be created in one of two ways:
+- As a success object, where `TResult` is provided.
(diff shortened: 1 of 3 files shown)


docs/new-token-validation-model.md
```diff
@@ -0,0 +1,86 @@
+# New token validation model
```
jennyf19, line 1: this would move to the wiki, in a page there. identityModel has the docs generally there.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests_e2e.cs
```diff
@@ -29,7 +30,7 @@ public async Task TestDefaultValidToken()
             };
             CallContext callContext = new CallContext();
 
-            ValidationResult<ValidatedToken> validationResult = await jsonWebTokenHandler.ValidateTokenAsync(token, validationParameters, callContext, default);
+            ValidationResult<ValidatedToken> validationResult = await ((IResultBasedValidation)jsonWebTokenHandler).ValidateTokenAsync(token, validationParameters, callContext, default);
```
jmprieur: Instead of having casts everywhere, can we have a private get property returning this?

docs/new-token-validation-model.md
```diff
@@ -0,0 +1,86 @@
...
+The existing token validation model on `Microsoft.IdentityModel` relies on the exception based control flow dominant in C#. In some high-performance scenarios where latency is key, this can cause unintended increases when validations fail and throw.
```
jmprieur, line 2: What does "dominant in C#" mean?

There is another point which is that where trying several authentication schemes in ASP.NET Core, if one fails and the next succeeds, an error is still logged (which is confusing for developers). In other words the logs can make the decision too early to fail.

docs/new-token-validation-model.md
```diff
@@ -0,0 +1,86 @@
...
+- Performance increase by removing unnecessary exception throwing within the validation itself.
+- Return a result where it is clear whether the validation was successful, including relevant information depending on the case.
+  - In success scenarios, provide the validated information for audit purposes.
+  - In failure scenarios, provide enough information to identify what part of the token was invalid or could not be validated without the need for further steps, and for diagnostic purposes provide the ability to create an exception that can be thrown and observed. 
+- Do not automatically print logs as part of the validation. These can be printed from the validation result using C#'s [high-performance logging](https://learn.microsoft.com/en-us/dotnet/core/extensions/high-performance-logging) based on `ILogger`.
+- Simpler validation parameters object, as the current one has grown in complexity over the years by including multiple parameters for things like `ValidIssuer` and `ValidIssuers`, or multiple delegates for the same validation. `ValidationParameters` is being introduced to offer a simplified object that enables simplifying the default validation code as well.
```
jmprieur:
```suggestion
- Simpler validation parameters object, as the current one has grown in complexity over the years by including multiple parameters for things like `ValidIssuer` and `ValidIssuers`, or multiple delegates for the same validation. `ValidationParameters` is being introduced, as an alternative to `TokenValidationParameters` to offer a simplified object that enables simplifying the default validation code as well.
```

docs/new-token-validation-model.md
```diff
@@ -0,0 +1,86 @@
...
+- New APIs provide nullability annotations to simplify code branches when the API can ensure no `null` will be returned from a method.
+- New APIs receive a `CancellationToken` to allow for the cancellation of running validation operations.
```
jmprieur:
```suggestion
- New APIs async receive a `CancellationToken` to allow for the cancellation of running validation operations, as expected
```

jmprieur (approved):
Looks great!
I proposed a few tweaks.

docs/new-token-validation-model.md
iNinja: I included it in the PR for review. Should we remove the file from the repo and just send it to the wiki?

docs/new-token-validation-model.md
iNinja: I meant to say that the exception based control flow is the most common way to perform control flow in C#.
Other languages/frameworks have options at times, but here I would dare say over 90% of the codebases follow this approach as it is the norm.

docs/new-token-validation-model.md
jennyf19: Yes please! thanks @iNinja

jennyf19:
@iNinja the package shipped last week, are you planning on merging this and updating the wiki?

jennyf19 (approved):

PR #3128, opened 2025-02-13
Add TryAllDecryptionKeys flag to whether decrypt if no key IDs match

### LINKED ISSUE #3129: [Bug] Token is decrypted using all keys if no key ID matches
Currently the precedence for token decryption keys is:
- `TokenValidationParameters.TokenDecryptionKeyResolver`, if set.
- `TokenValidationParameters.TokenDecryptionKey`, if set and key ID matches.
- `TokenValidationParameters.TokenDecryptionKeys`, if set and contains any keys where key ID matches.
- `TokenValidationParameters.TokenDecryptionKey`, `TokenValidationParameters.TokenDecryptionKeys`, and `TokenDecryptionKeys` from configuration.

Signature validation has similar logic and also has `TryAllIssuerSigningKeys` flag as a fallback.

Add `TryAllDecryptionKeys` flag, which when set, will enable validating tokens if key IDs don't match. If not set, only keys with matching key IDs will be used.

pmaytak:
Fixes #3129

* Added a new property `TryAllDecryptionKeys` to `TokenValidationParameters` to indicate whether all decryption keys should be tried during token decryption when a key is not matched to the token 'kid' or if the token 'kid' is empty. This property defaults to `true`. [[1]](diffhunk://#diff-fd6fb46108ab0d6b00c019e12ed417bd02e026042ddaa215ec2b2301fc1304faL595-R608) [[2]](diffhunk://#diff-3c7d43b6d12b5f2a95d17637606de002d7b7fcff07f51a247f2c642de3f794c2R519-R525)
* Updated the `GetContentEncryptionKeys` and `DecryptToken` methods to use the new `TryAllDecryptionKeys` property for determining when to try all decryption keys. [[1]](diffhunk://#diff-2e592db8e9e75f43c6efe0487d18431538706d98ec79c930924eb92e2a47d753L1323-R1323) [[2]](diffhunk://#diff-3f800ae7e51ac75b2620fa577baef4714bd8435b92b8ba271947e8da321407c5L124-R124)
* Also updates the new validation code path.

The precedence of decryption methods:
- `TokenValidationParameters.TokenDecryptionKeyResolver`, if set.
- `TokenValidationParameters.TokenDecryptionKey`, if set and key ID matches.
- `TokenValidationParameters.TokenDecryptionKeys`, if set and contains any keys where key ID matches.
- If `TokenValidationParameters.TryAllDecryptionKeys` is set to 'true', `TokenValidationParameters.TokenDecryptionKey`, `TokenValidationParameters.TokenDecryptionKeys`, and `TokenDecryptionKeys` from configuration.

**Note: I will add more tests if this approach is agreed on.**

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
@@ -1320,7 +1320,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             // 2. ResolveTokenDecryptionKey returned null
             // 3. ResolveTokenDecryptionKeyFromConfig returned null
             // Try all the keys. This is the degenerate case, not concerned about perf.
-            if (keys == null)
+            if (validationParameters.TryAllDecryptionKeys && keys.IsNullOrEmpty())
             {
                 keys = JwtTokenUtilities.GetAllDecryptionKeys(validationParameters);
                 if (configuration != null)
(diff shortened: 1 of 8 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
```diff
@@ -1335,60 +1335,62 @@ internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken
             // keep track of exceptions thrown, keys that were tried
             StringBuilder exceptionStrings = null;
             StringBuilder keysAttempted = null;
-            foreach (var key in keys)
+            if (keys != null)
             {
-                try
+                foreach (var key in keys)
```
jennyf19, line 1340: wondering if we can extract some of this to a helper method, which we can isolate with testing? Maybe is outside of this PR though.

jennyf19:
Any performance tests needed given the additional logic for key handling?

jennyf19 (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
pmaytak: We have tests for this `GetContentEncryptionKeys` already. I think it's fairly easy and straightforward to hit all code paths with the current code.

pmaytak:
> Any performance tests needed given the additional logic for key handling?

Don't think so. It's just one extra bool flag check.

Jimmyawesom (commented):
Everything looks great, thank

jmprieur (approved):
LGTM
Thanks @pmaytak 

There are formatting issues (IDE0055) that need to be addressed, probably in a different PRs (see at the bottom of the PR, as you have not touched that file)

Jimmyawesom:
Been trying to get a BOT in hacker out of my phone for the past year and a half and nothing has worked there after my investments please help anybody name ideas. I’m open for anybody anything

PR #3138, opened 2025-02-19
Update Severity of Logs for IDX10223

### LINKED ISSUE #2614: Change severity level for logs like "IDX10223: Lifetime validation failed. The token is expired"
### Microsoft.Identity.Web Library

Microsoft.Identity.Web

### Microsoft.Identity.Web version

2.16.1

### Web app

Sign-in users and call web APIs

### Web API

Protected web APIs (validating tokens)

### Token cache serialization

Not Applicable

### Description

Every day I got a lot of error logs like

1. `IDX10223: Lifetime validation failed. The token is expired...`
2. `IDX10205: Issuer validation failed...`

The 1st one is not an error and we want to ignore it: this means that some of our clients make a request with an expired token, then the 401 will be returned and the client goes to the refresh endpoint for the new token. So it's not an error on the server side, we can't fix it from the server side. Why is it treated as an error?

How to disable/ignore/change-severity them?

mdchennu:
#{Update Severity of Logs for IDX10223 from Exception to Warning}

## Description

**Detail**
Things left to fix: 

- [ ] Tests failing in Microsoft.IdentityModel.JsonWebTokens.Tests [ValidateToken_Async] -> due to a null reference issue (idk what the cause is yet)
- [ ] Tests failing in Microsoft.IdentityModel.Tokens.Validation.Tests.LifetimeValidationResultTests.ValidateLifetime -> due to System.InvalidOperationException : Cannot unwrap error result (cause is the current set up of test cases, thinking abt fix still)

- [X] Tests are still failing in Microsoft.IdentityModel.Tokens.Saml.Tests -> due to the thrown exception being of type ExpectedException.SecurityTokenSignatureKeyNotFoundException("IDX10512:"), but the expected exception should be SecurityTokenInvalidIssuerException (IDX10204) (my guess)

Fixes #2614 (in this specific format)

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Validators.Lifetime.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Validators.Lifetime.cs
@@ -90,16 +90,10 @@ namespace Microsoft.IdentityModel.Tokens
                     expires);
 
             if (expires.HasValue && (expires.Value < DateTimeUtil.Add(utcNow, validationParameters.ClockSkew.Negate())))
-                return new LifetimeValidationError(
-                    new MessageDetail(
-                        LogMessages.IDX10223,
-                        LogHelper.MarkAsNonPII(expires.Value),
-                        LogHelper.MarkAsNonPII(utcNow)),
-                    ValidationFailureType.LifetimeValidationFailed,
-                    typeof(SecurityTokenExpiredException),
-                    ValidationError.GetCurrentStackFrame(),
-                    notBefore,
-                    expires);
+                LogHelper.LogWarning(
+                    LogMessages.IDX10223,
+                    LogHelper.MarkAsNonPII(expires.Value),
+                    LogHelper.MarkAsNonPII(utcNow));
 
             // if it reaches here, that means lifetime of the token is valid
             return new ValidatedLifetime(notBefore, expires);
--- a/src/Microsoft.IdentityModel.Tokens/ValidatorUtilities.cs
+++ b/src/Microsoft.IdentityModel.Tokens/ValidatorUtilities.cs
@@ -42,6 +42,12 @@ namespace Microsoft.IdentityModel.Tokens
                 {
                     NotBefore = notBefore.Value
                 });
+
+            if (expires.HasValue && (expires.Value < DateTimeUtil.Add(utcNow, validationParameters.ClockSkew.Negate())))
+                if (LogHelper.IsEnabled(EventLogLevel.Warning))
+                {
+                    LogHelper.LogWarning(LogMessages.IDX10223, LogHelper.MarkAsNonPII(expires.Value), LogHelper.MarkAsNonPII(utcNow));
+                }
             // if it reaches here, that means lifetime of the token is valid
             if (LogHelper.IsEnabled(EventLogLevel.Informational))
                 LogHelper.LogInformation(LogMessages.IDX10239);
--- a/src/Microsoft.IdentityModel.Tokens/ValidatorUtilities.cs
+++ b/src/Microsoft.IdentityModel.Tokens/ValidatorUtilities.cs
@@ -42,13 +42,6 @@ namespace Microsoft.IdentityModel.Tokens
                 {
                     NotBefore = notBefore.Value
                 });
-
-            if (expires.HasValue && (expires.Value < DateTimeUtil.Add(utcNow, validationParameters.ClockSkew.Negate())))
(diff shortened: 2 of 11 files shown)


src/Microsoft.IdentityModel.Tokens/ValidatorUtilities.cs
```diff
@@ -42,13 +42,6 @@ internal static void ValidateLifetime(DateTime? notBefore, DateTime? expires, Se
                 {
                     NotBefore = notBefore.Value
                 });
-
```
kllysng, line 45: Maybe I'm not understanding the WI, but why are we moving this code to another file?

mdchennu: The way that I understood the work item was that any exception of IDX10223 should be considered a warning rather than an exception. 

So I thought that moving this code out from ValidatorUtilities (which mainly throws exceptions) would make sense, since IDX10223 will no longer trigger an exception. 

I took another look, and it does make more sense to just leave it in its original place and log a warning. I just moved it back! Thank you for reviewing Kelly :)

test/Microsoft.IdentityModel.Tokens.Saml.Tests/SamlTestData.cs
```diff
@@ -70,7 +70,9 @@ public static TheoryData<CreateTokenTheoryData> SecurityKeyNotFoundExceptionTest
                         ValidateIssuer = false,
                         ValidateAudience = false
                     },
-                    ExpectedException = ExpectedException.SecurityTokenExpiredException("IDX10223:")
+                    //not sure this is the correct exception, it probably needs to be SecurityTokenInvalidIssuerException (IDX10204) since the goal of this PR is to remove: SecurityTokenExpiredException (IDX10223)
```
pmaytak, line 73: Line 70 disables issuer validation.

pmaytak:
Looks good so far. 

The validation code path has been refactored - so there are basically two validation paths side by side now (only the current one works, the newer one is not hooked up). So you have to make sure that changes are made to both paths. For ex here too: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/6a9e82379a1611e40d65aa94496cd638d3fe1056/src/Microsoft.IdentityModel.Tokens/Validation/Validators.Lifetime.cs#L95

Another thing to think about is that changing this to not throwing an exception could be considered a [breaking change](https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules#exceptions).

mdchennu:
@kllysng the failing test i mentioned earlier today: 

**Lifetime Validation Failures** -> for both in the old and new validation paths. 

An example below:

**Test File:** Microsoft.IdentityModel.Tokens.Validation.Tests.LifetimeValidationResultTests.ValidateLifetime

**Failing Test Case Names:** 
theoryData: Invalid_Expired (#9)
theoryData: Invalid_Expired_SkewBackward (#11)

**Error Message:**
System.InvalidOperationException : Cannot unwrap error result

**Stack Trace:**
   ValidationResult`1.UnwrapResult() line 74
   LifetimeValidationResultTests.ValidateLifetime(ValidateLifetimeTheoryData theoryData) line 38
   InvokeStub_LifetimeValidationResultTests.ValidateLifetime(Object, Span`1)
   MethodBaseInvoker.InvokeWithOneArg(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)

Potential Fix:

1. Maybe the LifetimeValidation.cs case needs to throw an error or log an error so that the tests can sort through the code.

src/Microsoft.IdentityModel.Tokens/Validation/Validators.Lifetime.cs
```diff
@@ -90,16 +90,10 @@ internal static ValidationResult<ValidatedLifetime> ValidateLifetime(
...
-                        LogHelper.MarkAsNonPII(utcNow)),
-                    ValidationFailureType.LifetimeValidationFailed,
-                    typeof(SecurityTokenExpiredException),
-                    ValidationError.GetCurrentStackFrame(),
-                    notBefore,
-                    expires);
+                LogHelper.LogWarning(
+                    LogMessages.IDX10223,
+                    LogHelper.MarkAsNonPII(expires.Value),
+                    LogHelper.MarkAsNonPII(utcNow));
```
jennyf19, line 96: this doesn't seem right because previously we were throwing a `SecurityTokenExpiredException` and now we log a warning. We only want to change the places where we were logging IDX10223 as an error level and move that to warning level. See https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#configure-logging

mdchennu:
Closing PR for now as work is more complicated than expected, and requires new validation path.

PR #3147, opened 2025-02-24
Serialize array claims to string if AppContext switch is set

sruke:


sruke:
This approach has been abandoned.

PR #3149, opened 2025-02-25
Trigger metadata refresh for token decryption errors

### LINKED ISSUE #3148: [Feature Request] Trigger metadata refresh for token decryption failures
Currently if token decryption fails, we don't trigger the metadata refresh and just fail the validation.

Trigger a configuration refresh in this scenario (similar to token signature validation failures).

pmaytak:
Fixes #3148

## Summary
aims to address decryption issues by triggering a metadata refresh when token decryption fails due to the token's Key ID (Kid) not matching any of the decryption keys' IDs. It introduces changes to handle SecurityTokenEncryptionKeyNotFoundException as a recoverable exception when the current configuration contains decryption keys.

Key changes include:

- Updates to decryption logic to check for Kid matches.
- Modifications to exception handling to treat specific decryption errors as recoverable.
- Corrections and improvements in exception messages and log messages.- 

## Details:
- Updated decryption code to throw `SecurityTokenEncryptionKeyNotFoundException` when decryption fails and token's Kid doesn't match the decryption keys' IDs.
   - In certain scenarios decryption keys don't have key IDs. In this case 
- `SecurityTokenEncryptionKeyNotFoundException` is considered a recoverable exception when the current configuration contains decryption keys and in that case a metadata refresh will be triggered.

`SecurityTokenEncryptionKeyNotFoundException` already existed but it seems like it has never been used since it was created years ago and seems like the purpose was for this exact scenario.


jennyf19 (approved):
Re-setting my review, I need a bit more time to look at this.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -252,7 +252,8 @@ internal static bool IsRecoverableException(Exception exception)
         {
             return exception is SecurityTokenInvalidSignatureException
                   || exception is SecurityTokenInvalidIssuerException
-                  || exception is SecurityTokenSignatureKeyNotFoundException;
+                  || exception is SecurityTokenSignatureKeyNotFoundException
+                  || exception is SecurityTokenDecryptionFailedException;
```
jmprieur: We only want to refresh the metadata if:
- the token decryption failed because the encryption key is unknown
- **and** there are decryption keys in the openid connect metadata.

Here aren't we going to attempt the refresh the metadata each time a token encryption fails (modulo the LKG 5 mins wait)

GeoK: I don't think that this is sufficient. We should signal metadata refresh only if encryption keys are part of metadata.

pmaytak: @GeoK 
I updated the code with that check here:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/be8e3a9b93f0b01c495615b9efd22b0180a3c586/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.cs#L514

Also throwing a more specific `SecurityTokenEncryptionKeyNotFoundException`.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
```diff
@@ -245,15 +245,17 @@ internal static IEnumerable<Claim> MergeClaims(IEnumerable<Claim> claims, IEnume
...
+        /// <param name="configContainsDecryptionKeys">Whether the configuration contains decryption keys.</param>
         /// <returns><c>true</c> if the exception is certain types of exceptions otherwise, <c>false</c>.</returns>
-        internal static bool IsRecoverableException(Exception exception)
+        internal static bool IsRecoverableException(Exception exception, bool configContainsDecryptionKeys)
         {
             return exception is SecurityTokenInvalidSignatureException
                   || exception is SecurityTokenInvalidIssuerException
                   || exception is SecurityTokenSignatureKeyNotFoundException
-                  || exception is SecurityTokenDecryptionFailedException;
+                  || (exception is SecurityTokenEncryptionKeyNotFoundException && configContainsDecryptionKeys);
```
GeoK: SecurityTokenEncryptionKeyNotFoundException seems to be to narrow. SecurityTokenDecryptionFailedException looks like a better fit for the scenario.

jmprieur (commented):
@pmaytak 
You changed the public API, but didn't update the PublicApi.Unshipped.txt files (you didn't use the fixer in the IDE)

- IDX10603 is not part of the declared API in LogMessages.cs.
- IDX10907 is not part of the declared API in LogMessages.cs.
- IsRecoverableException is not part of the declared API in TokenUtilities.cs.
- IsRecoverableExceptionType is not part of the declared API in TokenUtilities.cs.

src/Microsoft.IdentityModel.Tokens/TokenUtilities.cs
jmprieur: Would you want to trigger the LKG in other conditions that KeyNotFound, @GeoK ? what would the scenario be?

pmaytak: Updated the code:
- Don't try to match on key IDs
- Revert to throwing a more general token decryption error

We still only do the refresh if the current config has keys.

src/Microsoft.IdentityModel.Tokens/Validation/Results/Details/ValidationError.cs
```diff
@@ -161,8 +161,8 @@ internal Exception CreateException(Type exceptionType, Exception? innerException
                     exception = new SecurityTokenInvalidIssuerException(MessageDetail.Message, innerException);
                 else if (exceptionType == typeof(SecurityTokenSignatureKeyNotFoundException))
                     exception = new SecurityTokenSignatureKeyNotFoundException(MessageDetail.Message, innerException);
-                else if (exceptionType == typeof(SecurityTokenDecryptionFailedException))
```
GeoK: Why remove this piece of code?

pmaytak: Trying to keep the changes in the PR minimal to the issue at hand.

GeoK (approved):

jmprieur (approved):
LGTM
Thanks @pmaytak. Great work!

PR #3152, opened 2025-03-03
Populate error messages correctly from JwtTokenUtilities.DecryptJwtToken

### LINKED ISSUE #3003: [Bug] JwtTokenDecryptionParameters's Alg and Enc is not set in JsonWebTokenHandler's token decryption, IDX10611 and IDX10619 shows null in messges
**Where is the issue?**
 * [x ] M.IM.JsonWebTokens
 
**Repro**
JwtSecurityTokenHandler DecryptToken inital the Alg and Enc
image
While JsonWebTokenHandler don't:
image

**Expected behavior**
Error message shouldshows the token's enc
**Actual behavior**
Microsoft.IdentityModel.Tokens.SecurityTokenDecryptionFailedException: IDX10619: Decryption failed. Algorithm: 'Null'. Either the Encryption Algorithm: 'Null' or none of the Security Keys are supported by the CryptoProviderFactory.

IDX10611: Decryption failed. Encryption is not supported for: Algorithm: 'Null', SecurityKey: '[PII of type 'Microsoft.IdentityModel.Tokens.X509SecurityKey' is hidden. For more details, see https://aka.ms/IdentityModel/PII.]'.

ksaaf:
# Populate error messages correctly from JwtTokenUtilities.DecryptJwtToken
This PR fixes populating error messages when JwtTokenUtilities.DecryptJwtToken is called for to decrypt JsonWebToken.

## Description

Fixes #3003

### DIFF
--- a/build/commonTest.props
+++ b/build/commonTest.props
@@ -32,6 +32,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNETTestSdkVersion)" />
+    <PackageReference Include="Moq" Version="$(MoqVersion)" />
     <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftVersion)" />
     <PackageReference Include="xunit" Version="$(XunitVersion)" />
     <PackageReference Include="xunit.runner.console" Version="$(XunitRunnerConsoleVersion)" />
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
@@ -68,15 +68,12 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     ValidationError.GetCurrentStackFrame());
             }
 
+            var decryptionParameters = CreateJwtTokenDecryptionParameters(jwtToken, result.contentEncryptionKeys);
+
             return JwtTokenUtilities.DecryptJwtToken(
                 jwtToken,
                 validationParameters,
-                new JwtTokenDecryptionParameters
-                {
-                    DecompressionFunction = JwtTokenUtilities.DecompressToken,
-                    Keys = result.contentEncryptionKeys,
-                    MaximumDeflateSize = MaximumTokenSizeInBytes
-                },
+                decryptionParameters,
                 callContext);
         }
 
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
@@ -51,13 +51,13 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     ValidationError.GetCurrentStackFrame());
             }
 
-            (IList<SecurityKey>? contentEncryptionKeys, ValidationError? validationError) result =
+            (IList<SecurityKey>? ContentEncryptionKeys, ValidationError? ValidationError) result =
                 GetContentEncryptionKeys(jwtToken, validationParameters, configuration, callContext);
 
-            if (result.validationError != null)
-                return result.validationError.AddCurrentStackFrame();
+            if (result.ValidationError != null)
+                return result.ValidationError.AddCurrentStackFrame();
 
-            if (result.contentEncryptionKeys == null || result.contentEncryptionKeys.Count == 0)
+            if (result.ContentEncryptionKeys == null || result.ContentEncryptionKeys.Count == 0)
             {
                 return new ValidationError(
                     new MessageDetail(
@@ -68,7 +68,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                     ValidationError.GetCurrentStackFrame());
             }
 
-            var decryptionParameters = CreateJwtTokenDecryptionParameters(jwtToken, result.contentEncryptionKeys);
+            var decryptionParameters = CreateJwtTokenDecryptionParameters(jwtToken, result.ContentEncryptionKeys);
 
             return JwtTokenUtilities.DecryptJwtToken(
                 jwtToken,
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -350,7 +350,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             return JwtTokenUtilities.DecryptJwtToken(jwtToken, validationParameters, decryptionParameters);
         }
 
-        internal JwtTokenDecryptionParameters CreateJwtTokenDecryptionParameters(JsonWebToken jwtToken, IEnumerable<SecurityKey> keys)
+        private JwtTokenDecryptionParameters CreateJwtTokenDecryptionParameters(JsonWebToken jwtToken, IEnumerable<SecurityKey> keys)
         {
             return new JwtTokenDecryptionParameters
             {
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
@@ -344,15 +344,28 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 throw LogHelper.LogExceptionMessage(new SecurityTokenException(LogHelper.FormatInvariant(TokenLogMessages.IDX10612)));
 
             var keys = GetContentEncryptionKeys(jwtToken, validationParameters, configuration);
-            return JwtTokenUtilities.DecryptJwtToken(
-                jwtToken,
-                validationParameters,
-                new JwtTokenDecryptionParameters
-                {
-                    DecompressionFunction = JwtTokenUtilities.DecompressToken,
-                    Keys = keys,
-                    MaximumDeflateSize = MaximumTokenSizeInBytes
-                });
+
+            var decryptionParameters = CreateJwtTokenDecryptionParameters(jwtToken, keys);
+
+            return JwtTokenUtilities.DecryptJwtToken(jwtToken, validationParameters, decryptionParameters);
+        }
+
+        internal JwtTokenDecryptionParameters CreateJwtTokenDecryptionParameters(JsonWebToken jwtToken, IEnumerable<SecurityKey> keys)
+        {
+            return new JwtTokenDecryptionParameters
+            {
+                Alg = jwtToken.Alg,
+                AuthenticationTagBytes = jwtToken.AuthenticationTagBytes,
+                CipherTextBytes = jwtToken.CipherTextBytes,
+                DecompressionFunction = JwtTokenUtilities.DecompressToken,
+                Enc = jwtToken.Enc,
+                EncodedToken = jwtToken.EncodedToken,
+                HeaderAsciiBytes = jwtToken.HeaderAsciiBytes,
+                InitializationVectorBytes = jwtToken.InitializationVectorBytes,
+                MaximumDeflateSize = MaximumTokenSizeInBytes,
+                Keys = keys,
+                Zip = jwtToken.Zip,
+            };
         }
 
         private static SecurityKey ResolveTokenDecryptionKeyFromConfig(JsonWebToken jwtToken, BaseConfiguration configuration)
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -332,6 +332,14 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             }
         }
 
+        /// <summary>
+        /// This is added for testing purpose to allow returning null.
+        /// 
+        /// When this is called outside of test(s),
+        /// <see cref="TokenValidationParameters.CryptoProviderFactory"/> can be null
+        /// but <see cref="SecurityKey.CryptoProviderFactory"/> will never be. Thus, the caller
+        /// will never receive a null <see cref="CryptoProviderFactory"/>.
+        /// </summary>
         internal static Func<TokenValidationParameters, SecurityKey, CryptoProviderFactory>
             s_getCryptoProviderFactory = (validationParameters, key) =>
         {
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
@@ -264,7 +264,7 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             string zipAlgorithm = null;
             foreach (SecurityKey key in decryptionParameters.Keys)
             {
-                var cryptoProviderFactory = s_getCryptoProviderFactory(validationParameters, key);
+                var cryptoProviderFactory = validationParameters.CryptoProviderFactory ?? key.CryptoProviderFactory;
                 if (cryptoProviderFactory == null)
                 {
                     if (LogHelper.IsEnabled(EventLogLevel.Warning))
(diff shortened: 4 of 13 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -68,15 +68,12 @@ internal ValidationResult<string> DecryptToken(
                     ValidationError.GetCurrentStackFrame());
             }
 
+            var decryptionParameters = CreateJwtTokenDecryptionParameters(jwtToken, result.contentEncryptionKeys);
```
pmaytak: nit: Not related to your changes - can you please update the casing to this proper casing for a property so it's easier to read:
`(IList<SecurityKey>? ContentEncryptionKeys, ValidationError? ValidationError) result`

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -344,15 +344,28 @@ private string DecryptToken(JsonWebToken jwtToken, TokenValidationParameters val
...
-                    Keys = keys,
-                    MaximumDeflateSize = MaximumTokenSizeInBytes
-                });
+
+            var decryptionParameters = CreateJwtTokenDecryptionParameters(jwtToken, keys);
+
+            return JwtTokenUtilities.DecryptJwtToken(jwtToken, validationParameters, decryptionParameters);
+        }
+
+        internal JwtTokenDecryptionParameters CreateJwtTokenDecryptionParameters(JsonWebToken jwtToken, IEnumerable<SecurityKey> keys)
```
pmaytak: Can be `private`. It's only used in JsonWebTokenHandler class?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -2812,6 +2814,38 @@ public static TheoryData<CreateTokenTheoryData> RoundTripJWEDirectTestCases
...
+
+                    var keysAttempted = new StringBuilder().AppendLine(validationParameters.TokenDecryptionKey.KeyId);
+                    var incompleteExceptionMessage = new MessageDetail(
+                                Tokens.LogMessages.IDX10603,
+                                LogHelper.MarkAsNonPII(keysAttempted.ToString()),
+                                string.Empty,   // Using empty since actual exception contains file paths, which are machine specific.
+                                string.Empty)   // Using empty since EncodedToken is not available and the message is getting used partially below.
+                        .Message;
+                    // Get partial messages as the actual exception message contains file paths.
+                    var partialExceptionMessage = incompleteExceptionMessage.Substring(
```
pmaytak, line 2834: Not used?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/Microsoft.IdentityModel.JsonWebTokens.Tests.csproj
```diff
@@ -12,6 +12,10 @@
     <SignAssembly>true</SignAssembly>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Moq" Version="4.16.1" />
```
pmaytak: Dependency versions can go here:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/build/dependenciesTest.props

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -365,6 +332,12 @@ internal static string DecryptJwtToken(
             }
         }
 
+        internal static Func<TokenValidationParameters, SecurityKey, CryptoProviderFactory>
```
pmaytak: I would probably have this as a static method instead. I believe Func may have extra perf impact.

Or ideally maybe keep the code as-is and test by passing CryptoProviderFactory via the validation parameters or the key.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JwtTokenUtilitiesTests.cs
```diff
@@ -208,6 +215,281 @@ public void ResolveTokenSigningKey()
...
+            TokenValidationParameters validationParameters = null;
+
+            // Act & Assert
+            var exception = Assert.Throws<ArgumentNullException>(() =>
+                JwtTokenUtilities.DecryptJwtToken(
+                    securityToken: null,
+                    validationParameters: validationParameters,
+                    decryptionParameters: null));
+
+            Assert.Equal("validationParameters", exception.ParamName);
```
pmaytak: Use `nameof(validationParameters)`. instead of literal.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JwtTokenUtilitiesTests.cs
```diff
@@ -208,6 +215,281 @@ public void ResolveTokenSigningKey()
...
+            var exception = Assert.Throws<ArgumentNullException>(() =>
+                JwtTokenUtilities.DecryptJwtToken(
+                    securityToken: null,
+                    validationParameters: validationParameters,
+                    decryptionParameters: decryptionParameters));
+
+            Assert.Equal("decryptionParameters", exception.ParamName);
+        }
+
+        [Fact(Skip = "Skip due to potential false positives it may cause running tests in parallel as it overrides JwtTokenUtilities.s_getCryptoProviderFactory.")]
```
pmaytak: Skip is not needed.

`By default, each test class is a unique test collection. Tests within the same test class will not run in parallel against each other. Let's examine a very simple test assembly, one with a single test class:`

https://xunit.net/docs/running-tests-in-parallel#:~:text=By%20default%2C%20each%20test%20class%20is%20a%20unique%20test%20collection.%20Tests%20within%20the%20same%20test%20class%20will%20not%20run%20in%20parallel%20against%20each%20other.%20Let%27s%20examine%20a%20very%20simple%20test%20assembly%2C%20one%20with%20a%20single%20test%20class%3A

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JwtTokenUtilitiesTests.cs
```diff
@@ -208,6 +215,281 @@ public void ResolveTokenSigningKey()
...
+        public void DecryptJwtToken_WhenCryptoProviderFactoryIsNull_FailsWithIDX10609AndLogsIDX10607()
```
pmaytak: nit: I think for the naming just a general `_ThrowsException` is okay. It's a bit easier to read and we can look down in asserts what specific exception is thrown.
```suggestion
        public void DecryptJwtToken_WhenCryptoProviderFactoryIsNull_ThrowsException()
```

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
ksaaf: Updated.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
ksaaf: Good catch, removed.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
ksaaf: Meant to put it into as the expected substring in ``ExpectedException``; added.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/Microsoft.IdentityModel.JsonWebTokens.Tests.csproj
ksaaf: Done.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JwtTokenUtilitiesTests.cs
ksaaf: Usually, I use the general ``_ThrowsExpcetion`` too; somehow I didn't here. Updated.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JwtTokenUtilitiesTests.cs
ksaaf: I purposely have used literal string in the ``Assert`` on lines 271, and 288 (pasted below). 

``
Assert.Equal("validationParameters", exception.ParamName);
``
``
Assert.Equal("decryptionParameters", exception.ParamName);
``

I didn't use ``nameof`` as the ``exception.ParamName`` is set to the corresponding parameter name of the ``DecryptJwtToken`` function. If those parameters' names change, this test will break and require updating the literal string.

If I use the suggested nameof(), I will set have to update the variable name in the test to match the parameter name of the ``DecryptJwtToken`` function, though not the nameof().

That said, I moved the strings to a variable. That, I think is cleaner and also does not use literal inline in ``Assert``.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JwtTokenUtilitiesTests.cs
ksaaf: Thanks for the pointer. I wasn't aware of. Removed skip.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
ksaaf: I added summary for the ``s_getCryptoProviderFactory`` to explain the need for it. 

Given ``DecryptJwtToken`` is static, in order to test ``IDX10607``, this is the only way I am aware of without a lot of refactoring. 

As for the Func's perf impact, I am running benchmarks to confirm.

``s_getCryptoProviderFactory`` is an inline statement lamba. So, the impact may be minimal.

pmaytak: We have benchmarks here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/benchmark/Microsoft.IdentityModel.Benchmarks/ValidateJWEAsyncTests.cs

Is there a reason why this can't be an `internal static method?

```
internal static CryptoProviderFactory GetCryptoProviderFactory(TokenValidationParameters validationParameters, SecurityKey key)
{
    return validationParameters.CryptoProviderFactory ?? key.CryptoProviderFactory;
}
```

ksaaf: I need to mock this out from test to return null. static function like this cannot be mocked.

ksaaf: @pmaytak - Below are the benchmark run comparisons for the relevant tests. I do see a minor improvement in speed; no change in memory. 

BTW, this static function technique is what Microsoft Fakes also uses. See the Static methods section [here](https://learn.microsoft.com/en-us/visualstudio/test/using-shims-to-isolate-your-application-from-other-assemblies-for-unit-testing?view=vs-2022&tabs=csharp#shims-for-different-kinds-of-methods).

**PR Branch**

| Method                                   | Mean     | Error    | StdDev   | P90      | P95      | P100     | Gen0   | Gen1   | Allocated |
|----------------------------------------- |---------:|---------:|---------:|---------:|---------:|---------:|-------:|-------:|----------:|
| JsonWebTokenHandler_ValidateJWEAsync     | 34.66 us | 0.370 us | 0.805 us | 35.63 us | 36.18 us | 37.30 us | 3.6621 |      - |   15.7 KB |
| JwtSecurityTokenHandler_ValidateJWEAsync | 36.99 us | 0.330 us | 0.725 us | 38.02 us | 38.48 us | 38.69 us | 5.3711 | 0.1221 |   22.8 KB |

**Dev Branch**

| Method                                   | Mean     | Error    | StdDev   | P90      | P95      | P100     | Gen0   | Gen1   | Allocated |
|----------------------------------------- |---------:|---------:|---------:|---------:|---------:|---------:|-------:|-------:|----------:|
| JsonWebTokenHandler_ValidateJWEAsync     | 35.55 us | 0.931 us | 2.003 us | 37.28 us | 38.75 us | 44.98 us | 3.6621 |      - |   15.7 KB |
| JwtSecurityTokenHandler_ValidateJWEAsync | 37.27 us | 0.491 us | 1.067 us | 38.41 us | 39.40 us | 41.69 us | 5.3711 | 0.1221 |   22.8 KB |

GeoK: Can it be tested otherwise? We shouldn't degrade perf to optimize for tests, especially on a hot path. Looking at your results, perf actually got improved but it is likely result of your other changes.

ksaaf: @GeoK - Removed that static ``GetCryptoProviderFactory`` and used reflection to set ``CryptoProviderFactory`` to null in the ``SymmetricSecurityKey``.

@pmaytak FYI.

GeoK (approved):

build/commonTest.props
```diff
@@ -32,6 +32,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNETTestSdkVersion)" />
+    <PackageReference Include="Moq" Version="$(MoqVersion)" />
```
jennyf19, line 35: Are we sure we want to add this package? https://medium.com/@michalsitek/critical-security-vulnerability-in-moq-4-20-0-ffd24739cc49

Can we use something else?

jennyf19 (changes requested):
do we need the moq package?

build/commonTest.props
ksaaf: This PR is not adding Moq. It is already used in Microsoft.IdentityModel.Abstractions.Tests before.

The referenced issue is in Moq 4.20.0 and this repository is using 4.16.1.
I also searched repos across AzureAD org and Moq has been used in multiple repositories as well.

ksaaf: Pushed a commit to lock the Moq version similar to https://github.com/Azure/azure-sdk-for-net/pull/38113.

ksaaf:
> do we need the moq package?

Pushed a commit to lock the Moq version similar to https://github.com/Azure/azure-sdk-for-net/pull/38113.

build/commonTest.props
ksaaf: @jennyf19 - I am resolving this conversation as Moq already is in use. Per conversation with @GeoK, we will work separately on removing dependency on Moq.

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -275,61 +280,28 @@ internal static string DecryptJwtToken(
 
                 try
                 {
```
jmprieur, line 277: Wasn't this code necessary?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -344,15 +344,28 @@ private string DecryptToken(JsonWebToken jwtToken, TokenValidationParameters val
...
+
+            return JwtTokenUtilities.DecryptJwtToken(jwtToken, validationParameters, decryptionParameters);
+        }
+
+        private JwtTokenDecryptionParameters CreateJwtTokenDecryptionParameters(JsonWebToken jwtToken, IEnumerable<SecurityKey> keys)
+        {
+            return new JwtTokenDecryptionParameters
+            {
+                Alg = jwtToken.Alg,
+                AuthenticationTagBytes = jwtToken.AuthenticationTagBytes,
```
jmprieur, line 358: What was the impact of this missing before?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1762,7 +1762,14 @@ protected string DecryptToken(JwtSecurityToken jwtToken, TokenValidationParamete
 
             var keys = GetContentEncryptionKeys(jwtToken, validationParameters);
 
-            return JwtTokenUtilities.DecryptJwtToken(jwtToken, validationParameters, new JwtTokenDecryptionParameters
+            var decryptionParameters = CreateJwtTokenDecryptionParameters(jwtToken, keys);
+
+            return JwtTokenUtilities.DecryptJwtToken(jwtToken, validationParameters, decryptionParameters);
+        }
+
+        private JwtTokenDecryptionParameters CreateJwtTokenDecryptionParameters(JwtSecurityToken jwtToken, IEnumerable<SecurityKey> keys)
```
jmprieur, line 1770: Isn't it the same method as in the other file?

jmprieur (commented):
Thanks @ksaaf 
I left a couple of questions

jennyf19:
@ksaaf you need to sign the CLA (see the comment from the PR)

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
jmprieur: Discussed in person: same code as below

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
jmprieur: Discussed in person: yes, but the token flavor is different so different properties

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
jmprieur: Discussed in person: same code as below

jmprieur (approved):
LGTM
Thanks @ksaaf

ksaaf:

PR #3155, opened 2025-03-05
Add in lazy client certificate obtaining action functionality

tlupes:
# Add in lazy client certificate obtaining action functionality

Adds the ability to populate the ClientCertificates property when first obtained, instead of immediately. This is important as obtaining the client certificates may trigger a renegotiation, which is not desirable if the client certificates are not to be inspected.

Fixes #{bug number} (in this specific format)

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs
+++ b/src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs
@@ -17,6 +17,8 @@ namespace Microsoft.IdentityModel.Protocols
     {
         private IDictionary<string, IEnumerable<string>> _headers = new Dictionary<string, IEnumerable<string>>(StringComparer.OrdinalIgnoreCase);
         private X509Certificate2Collection _clientCertificates;
+        private Action<X509Certificate2Collection> _lazyClientCertificates;
+        private readonly object _lazyClientCertificatesLock = new object();
 
         /// <summary>
         /// Gets or sets the http request URI. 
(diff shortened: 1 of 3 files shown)


tlupes:

src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs
```diff
@@ -51,9 +53,29 @@ public IDictionary<string, IEnumerable<string>> Headers
...
-            _clientCertificates;
+        public X509Certificate2Collection ClientCertificates
+        {
+            get
+            {
+                _clientCertificates ??=
+                    Interlocked.CompareExchange(ref _clientCertificates, [], null) ??
+                    _clientCertificates;
+
+                if (_lazyClientCertificates != null)
```
MZOLN, line 64: Would spin lock / interlocked. be better here? Wouldn't serializing threads here have a negative effect? 
How does Asp.NET ClientCertificates handle it?

I'm understanding that this lock is instance based so that the impact is minimal?

MZOLN (commented):
Left comments

src/Microsoft.IdentityModel.Protocols/HttpRequestData.cs
tlupes: The lock _is_ instance based, so it should not be a problem to have a single lock to load the certificates. It will serialize all the threads that require the certificate for this specific request, but that should be minimal (typically just one).

Spin-lock wouldn't be the right move as this invokes a user-provided function and could run for a while. Interlocked alone would also be problematic as multiple threads could access the collection while it was being populated (It just ensures the empty collection is made once but won't protect the creation).

Asp.Net essentially uses a shared Task as the synchronization primitive, the first thread(s) create the task, and subsequent tasks await the task until completion.

PR #3157, opened 2025-03-06
New Validation Model: Use experimental namespace for public classes/methods

iNinja:
# New Validation Model: Use experimental namespace for public classes/methods
- Moved all public classes from the new validation model to the Experimental namespace
- Updated namespace references
- Updated InternalAPI and PublicAPI files to reflect the namespace changes

This change is made to enable us to release a preview release of the new validation model work from dev while expressing that it is not finalised and can still change before its final release.
The PR is targetting the feature branch to make it easier to review the changes before raising a PR to dev with the result.

Part of #2711


src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ReadToken.cs
```diff
@@ -5,6 +5,8 @@
 using System.Text;
 using System.Xml;
 using Microsoft.IdentityModel.Logging;
+using Microsoft.IdentityModel.Tokens.Experimental;
+using Microsoft.IdentityModel.Tokens.Saml.Experimental;
```
brentschmaltz, line 9: I think we should have just one experimental namespace M.IM.T.Experimental or  M.IM.Experimental.

keegan-caruso (approved):
No concerns

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ReadToken.cs
brentschmaltz: Will follow up with this issue in the feature branch.

PR #3158, opened 2025-03-06
New Validation Model: Make ValidationResult's error type explicit

iNinja:
# New Validation Model: Make ValidationResult's error type explicit
- Refactored ValidationResult to expose and customise the error type instead of using ValidationError
- Updated use of ValidationResult with more specialised error types for the different validation steps

This PR builds on top of  #3157 and targets its branch until it is merged to avoid showing duplicated changes.

Part of #2711


src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidationResult.cs
```diff
@@ -11,29 +11,30 @@ namespace Microsoft.IdentityModel.Tokens.Experimental
     /// Represents a validation result that can be either valid or invalid.
     /// </summary>
     /// <typeparam name="TResult"></typeparam>
```
keegan-caruso, line 13: Can we fill out the type param descriptions?

keegan-caruso (approved):
LGTM

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -23,7 +23,7 @@ public partial class JsonWebTokenHandler : TokenHandler
         /// <param name="configuration">The <see cref="BaseConfiguration"/> to be used for validating the token.</param>
         /// <param name="callContext"></param>
         /// <returns>The decoded / cleartext contents of the JWE.</returns>
```
sruke, line 25: Consider updating the comments in the return fields here and below to indicate a valid result on success or an error on failure.

sruke (approved):

PR #3168, opened 2025-03-12
Move CNF from SHR to M.IM.Tokens

### LINKED ISSUE #3165: [Feature Request] Public API to parse and return CNF

I need a public API that can parse and return a CNF claim into a domain object.

Cnf logic does currently exist, but it is internal and it is in the SignedHttpRequest project.

The public APIs should move to a lower project, likely M.IM.Tokens. This shouldn't be a problem for cross version setups as SignedHttpRequest has low usage and is a leaf dependency with the I.M solution

keegan-caruso:
# Make CNF type public

## Description

Makes the Cnf type public and moves it to M.IM.Tokens. Don't remove any types from SHR, instead make it inherit from the new type

Fixes #3165

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Cnf.cs
+++ b/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Cnf.cs
@@ -1,88 +1,19 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System;
-using System.Text;
-using System.Text.Json;
-using System.Text.Json.Serialization;
-using Microsoft.IdentityModel.Logging;
-using Microsoft.IdentityModel.Tokens;
-using Microsoft.IdentityModel.Tokens.Json;
-
 namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
 {
     /// <summary>
     /// Represents the Cnf Claim
     /// </summary>
-    internal class Cnf
+    internal class Cnf : Tokens.Cnf
     {
-        internal const string ClassName = "Microsoft.IdentityModel.Protocols.SignedHttpRequest.Cnf";
-
-        public Cnf() { }
-
-        public Cnf(string json)
+        public Cnf() : base()
         {
-            if (string.IsNullOrEmpty(json))
-                throw LogHelper.LogArgumentNullException(nameof(json));
-
-            Utf8JsonReader reader = new(Encoding.UTF8.GetBytes(json).AsSpan());
-            if (!JsonSerializerPrimitives.IsReaderAtTokenType(ref reader, JsonTokenType.StartObject, true))
-                throw LogHelper.LogExceptionMessage(
-                    new JsonException(
-                        LogHelper.FormatInvariant(
-                        Tokens.LogMessages.IDX11023,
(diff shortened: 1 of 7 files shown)


jennyf19 (approved):

src/Microsoft.IdentityModel.Tokens/Cnf.cs
```diff
@@ -0,0 +1,108 @@
...
+using Microsoft.IdentityModel.Tokens.Json;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    /// Represents the Cnf Claim
+    /// </summary>
+    public class Cnf
+    {
+        private const string ClassName = "Microsoft.IdentityModel.Protocols.SignedHttpRequest.Cnf";
```
GeoK: Should we change this to M.IM.Tokens.Cnf?

keegan-caruso: Wasn't sure if we want to change the error message. Thoughts on accepting it as a param on a protected constructor?

GeoK: I like the idea.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/ConfirmationClaimTypes.cs
```diff
@@ -14,27 +14,27 @@ public static class ConfirmationClaimTypes
         /// <summary>
         /// https://datatracker.ietf.org/doc/html/rfc7800#section-6.1.1
         /// </summary>
-        public const string Cnf = "cnf";
+        public const string Cnf = Tokens.ConfirmationClaimTypes.Cnf;
```
brentschmaltz, line 17: Can we change these constants?

src/Microsoft.IdentityModel.Tokens/Cnf.cs
brentschmaltz: I like the idea also.
It's important to be accurate.

Another possibility is to have: protected string virtual ClassName.
OR use the fully qualified name of the type.

keegan-caruso: ctor didn't quite make sense after looking at it, but protected internal property works well here.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/Cnf.cs
```diff
@@ -1,88 +1,23 @@
...
     /// <summary>
     /// Represents the Cnf Claim
     /// </summary>
-    internal class Cnf
+    internal class Cnf : Tokens.Cnf
     {
-        internal const string ClassName = "Microsoft.IdentityModel.Protocols.SignedHttpRequest.Cnf";
-
-        public Cnf() { }
+        private readonly string _shrClassName = "Microsoft.IdentityModel.Protocols.SignedHttpRequest.Cnf";
```
msbw2:
```suggestion
        private const string _shrClassName = "Microsoft.IdentityModel.Protocols.SignedHttpRequest.Cnf";
```

GeoK (approved):

jmprieur (approved):
LGTM
Thanks @keegan-caruso

PR #3170, opened 2025-03-12
JsonWebToken exposes Header Claims

### LINKED ISSUE #3164: [Feature Request]  JsonWebToken should expose more information about headers

I need a public API to access the JWT header keys, values, and claim value types.

Currently this is only accessible through the JsonWebToken's JsonClaimSet, which is an internal class. JsonClaimSet doesn't needs to be made public.

This would improve the usability of the JsonWebToken class

Reparsing the token is unacceptably high-performance cost.

keegan-caruso:
# JsonWebToken exposes Header Claims

The JsonWebToken should provide an API that allows access to the JWT header keys, values, and claim value types, this is done already through the Claim type.

Fixes #3164

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
@@ -614,6 +614,15 @@ namespace Microsoft.IdentityModel.JsonWebTokens
         }
 
         /// <summary>
+        /// Gets an <see cref="IReadOnlyList{Claim}"/> where each claim in the JWT header { name, value } is returned as a <see cref="Claim"/>.
+        /// </summary>
+        /// <remarks>
+        /// A <see cref="Claim"/> requires each value to be represented as a string. If the value was not a string, then <see cref="Claim.Type"/> contains the json type.
+        /// <see cref="JsonClaimValueTypes"/> and <see cref="ClaimValueTypes"/> to determine the json type.
+        /// </remarks>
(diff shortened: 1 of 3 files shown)


src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -613,6 +613,15 @@ public virtual IEnumerable<Claim> Claims
...
         }
 
+        /// <summary>
+        /// Gets an <see cref="IReadOnlyList{Claim}"/> where each claim in the JWT header { name, value } is returned as a <see cref="Claim"/>.
+        /// </summary>
+        /// <remarks>
+        /// A <see cref="Claim"/> requires each value to be represented as a string. If the value was not a string, then <see cref="Claim.Type"/> contains the json type.
+        /// <see cref="JsonClaimValueTypes"/> and <see cref="ClaimValueTypes"/> to determine the json type.
+        /// </remarks>
+        public IReadOnlyList<Claim> HeaderClaims => Header.Claims(Issuer ?? ClaimsIdentity.DefaultIssuer);
```
keegan-caruso, line 623: Alternatives: Only expose header names, and then call GetHeaderValue<Claim>(key) with each key from the header names, or reparse the encoded header

sruke: For known claim types, TryGetHeaderValue(ClaimType, out string claimValue) can be used without exposing header claims. Could you explain why this change is needed?

Issue #3169, opened 2025-03-12
[Feature Request] Use System.Threading.Lock on NET9 or greater


keegan-caruso:
**Is your feature request related to a problem? Please describe.**

Where today the library uses a lock, it should use the new System.Threading.Lock type introduced in NET 9 when able.

See: https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock?view=net-9.0

JoshLozensky:
Merged the changes to the locks, a separate request was made to add some thread-safety tests for the locks which I'll do in a new PR for issue #3178

PR #3171, opened 2025-03-13
Need to change the locks

JoshLozensky:
Change all plain object locks to System.Thread.Lock objects for .NET 9 or greater.

Addresses issue #3169

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Results/TokenValidationResult.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Results/TokenValidationResult.cs
@@ -30,7 +30,11 @@ namespace Microsoft.IdentityModel.Tokens
         // reordered relative to the other operations. The rest of the objects are not because the .NET memory model
         // guarantees object writes are store releases and that reads won't be introduced.
         private volatile bool _claimsIdentityInitialized;
+#if NET9_0_OR_GREATER
+        private Lock _claimsIdentitySyncObj = new();
+#else
         private object _claimsIdentitySyncObj;
+#endif
         private ClaimsIdentity _claimsIdentity;
         private Dictionary<string, object> _claims;
         private Dictionary<string, object> _propertyBag;
@@ -196,6 +200,23 @@ namespace Microsoft.IdentityModel.Tokens
             }
         }
 
+#if NET9_0_OR_GREATER
+        /// <summary>Gets the object to use in <see cref="ClaimsIdentity"/> for double-checked locking.</summary>
+        private Lock ClaimsIdentitySyncObj
+        {
+            get
+            {
+                Lock syncObj = _claimsIdentitySyncObj;
+                if (syncObj is null)
+                {
+                    Interlocked.CompareExchange(ref _claimsIdentitySyncObj, new Lock(), null);
+                    syncObj = _claimsIdentitySyncObj;
+                }
+
+                return syncObj;
+            }
+        }
+#else
         /// <summary>Gets the object to use in <see cref="ClaimsIdentity"/> for double-checked locking.</summary>
         private object ClaimsIdentitySyncObj
         {
@@ -211,7 +232,7 @@ namespace Microsoft.IdentityModel.Tokens
                 return syncObj;
             }
         }
-
+#endif
         /// <summary>
         /// Gets or sets the <see cref="Exception"/> that occurred during validation.
         /// </summary>
--- a/src/Microsoft.IdentityModel.Tokens/Validation/Results/TokenValidationResult.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Validation/Results/TokenValidationResult.cs
@@ -31,7 +31,7 @@ namespace Microsoft.IdentityModel.Tokens
         // guarantees object writes are store releases and that reads won't be introduced.
         private volatile bool _claimsIdentityInitialized;
 #if NET9_0_OR_GREATER
-        private Lock _claimsIdentitySyncObj = new();
+        private Lock _claimsIdentitySyncObj;
 #else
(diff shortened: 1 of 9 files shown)


jennyf19:
Looking good, @JoshLozensky ...curious if we can add some thread safety tests and have you done any benchmarks on the perf gains when using the specialized type?

src/Microsoft.IdentityModel.Tokens/Validation/Results/TokenValidationResult.cs
```diff
@@ -196,6 +200,23 @@ internal ClaimsIdentity ClaimsIdentityNoLocking
...
         }
 
+#if NET9_0_OR_GREATER
+        /// <summary>Gets the object to use in <see cref="ClaimsIdentity"/> for double-checked locking.</summary>
+        private Lock ClaimsIdentitySyncObj
+        {
+            get
+            {
+                Lock syncObj = _claimsIdentitySyncObj;
+                if (syncObj is null)
```
keegan-caruso, line 210: looks like this will never be null

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
```diff
@@ -190,6 +194,23 @@ internal ClaimsIdentity ClaimsIdentityNoLocking
...
         }
 
+#if NET9_0_OR_GREATER
+        /// <summary>Gets the Lock to use in <see cref="ClaimsIdentity"/> for double-checked locking.</summary>
+        private Lock ClaimsIdentitySyncObj
+        {
+            get
+            {
+                Lock? syncObj = _claimsIdentitySyncObj;
+                if (syncObj is null)
```
keegan-caruso, line 204: Looks like this will never be null?

src/Microsoft.IdentityModel.Tokens/X509SecurityKey.cs
```diff
@@ -110,12 +116,17 @@ public AsymmetricAlgorithm PublicKey
                 return _publicKey;
             }
         }
-
+#if NET9_0_OR_GREATER
+        Lock ThisLock
```
keegan-caruso: Could use expression bodies for the new sections e.g.

Lock ThisLock => _thisLock;

src/Microsoft.IdentityModel.Tokens/Validation/Results/TokenValidationResult.cs
JoshLozensky: agreed removing the init to match initial implementation

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
JoshLozensky: I don't think so in this case. Wouldn't the compare exchange inside the if statement be the only place _claimsIdentitySyncObj is set or am I missing something here? @keegan-caruso

src/Microsoft.IdentityModel.Tokens/X509SecurityKey.cs
JoshLozensky: changed

JoshLozensky:
> Looking good, @JoshLozensky ...curious if we can add some thread safety tests and have you done any benchmarks on the perf gains when using the specialized type?

Table below shows the percentage change in performance for the given benchmark (negative is "good"). The AggregatePercentageChange is an average of all means from statistically significant tests in the corresponding benchmark. Refer to the formulas below for relevant definitions. It looks like there is a fairly consistent trend of about 3.8% faster validation with the new Lock.

Relevant formulas:

Difference = TimeNewLock - TimeOldLock

StandardError = StandardErrorNewLock + StandardErrorOldLock

IsSignificant = Difference > Standard Error

PercentageChange = Difference / TimeOldLock

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/JOSHLO~1/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/JOSHLO~1/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">



</head>

<body link="#467886" vlink="#96607D">


Benchmark | AggregatePercentageChange
-- | --
CreateJWE | 0.98%
CreateSHR | -0.85%
CreateToken | -2.57%
ValidateJWE | -3.96%
ValidateSHR | -3.75%
ValidateTokenAsync | -3.82%



</body>

</html>

JoshLozensky:
> Looking good, @JoshLozensky ...curious if we can add some thread safety tests and have you done any benchmarks on the perf gains when using the specialized type?

I could add some thread-safety tests in a different PR if you'd like but since no logic is changed in this PR I think it is out of scope @jennyf19 

[Link to test PR](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3185)

src/Microsoft.IdentityModel.Tokens/Validation/Results/ValidatedToken.cs
keegan-caruso: No longer relevant, I was looking at the previous commit that new'd the object as part of the object initialization.

keegan-caruso (approved):
thanks

jennyf19 (approved):
:shipit:

PR #3173, opened 2025-03-13
Fix ValidationParameters Inconsistencies  

### LINKED ISSUE #3131: In the new validation model, ValidationParameters differs from TokenValidationParameters.
-Need to fix: 
- In the [ValidationParameters(ValidationParameters other)](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs), the value of `TryAllIssuerSigningKeys` is not copied over.
- Also in the new model `TryAllIssuerSigningKeys` is `false` by default, but `true` by default in the current model.
- Verify that there are no other missed differences between ValidationParameters and TokenValidationParameters.
Note: some differences are intended, like there's no TokenDecryptionKey property, only the TokenDecryptionKeys collection.

omidmloo:
# Fix ValidationParameters Inconsistencies  

## Summary  
This PR fixes inconsistencies in `ValidationParameters` by ensuring properties are copied correctly from `TokenValidationParameters`.  

### Description
- The `TryAllIssuerSigningKeys` property was not being copied in the `ValidationParameters(ValidationParameters other)` constructor.  
  - **Fix:** Now properly copies `TryAllIssuerSigningKeys`.  
  - **Behavior Change:** In the new model, `TryAllIssuerSigningKeys` was `false` by default, whereas it was `true` in the current model. This has been aligned accordingly.  
  - **Fix:** Updated `ValidateActor` in `ValidationParameters` for consistency.   

Fixes #3131


src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -530,6 +531,7 @@ public TokenReplayValidationDelegate TokenReplayValidator
         /// all available keys will be tried.
         /// </summary>
         /// <remarks>Default is false.</remarks>
+        [DefaultValue(true)]
```
pmaytak, line 535: This also needs to be set in the empty constructor.

omidmloo: Done

JoshLozensky (dismissed):

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
```diff
@@ -530,6 +532,7 @@ public TokenReplayValidationDelegate TokenReplayValidator
         /// all available keys will be tried.
         /// </summary>
         /// <remarks>Default is false.</remarks>
```
sruke: Please update the remarks to state that the default value is now true.

sruke:
Few unit tests are failing that require investigation.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
omidmloo: Done.

src/Microsoft.IdentityModel.Tokens/Validation/ValidationParameters.cs
brentschmaltz: It might be better to set the value on the property as if a new ctor is created it may be missed.
It would be a good idea to set the other properties: LogTokenId, TryAllDecryptionKeys
Add the attribute [DefaultValue(...)] as some of the bool properties as missing.

``
public bool TryAllIssuerSigningKeys { get; set; } = true;
``

brentschmaltz:
@omidmloo we are working on a rather large change at the moment to get the API's for the new model completed.
We are working out of this branch, https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/tree/brentsch/ValidationError

We may want to wait until this PR is in, before additional cleanup.

Issue #3206, opened 2025-04-14
[Bug] Type attribute of SubjectConfirmationData has incorrect casing


gislikonrad:
**Which version of Microsoft.IdentityModel are you using?**
Microsoft.IdentityModel.Tokens.Saml 8.8.0

**Where is the issue?**
 * [x] M.IM.Tokens.Saml

**Is this a new or an existing app?**
The app is in production and I have upgraded to a new version of Microsoft.IdentityModel.

**Repro**
Reproduction is a bit complex since we have extended the saml security token handling and serialization to handle cases which Microsoft.IdentityModel.Tokens.Saml doesn't, for example `holder-of-key`. The reproduction code uses our package, **Solid.IdentityModel.Tokens.Saml** version 1.0.6. When using the version of **Microsoft.IdentityModel.Tokens.Saml** that is the dependency of our package (which is 6.8.0), then the SubjectConfirmationData is serialized correctly. If we update to 8.8.0, then the type attribute has changed from **Type** to **type**. This breaks our integration with ADFS, which expects the type attribute to be **Type**.

The saml assertion has been formatted for readability.

```csharp
using System.Security.Claims;
using System.Security.Cryptography;
using System.Xml.Linq;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.IdentityModel.Tokens;
using Solid.IdentityModel.Tokens;
using Solid.IdentityModel.Tokens.Saml2;

var services = new ServiceCollection()
    .AddCustomCryptoProvider(options => options.AddFullSupport())
    .BuildServiceProvider()
;

CryptoProviderFactory.Default = services.GetRequiredService<CryptoProviderFactory>();

using var signing = RSA.Create(2048);
using var encryption = RSA.Create(2048);
var proof = new byte[32];
RandomNumberGenerator.Fill(proof);
var proofKey = new SymmetricSecurityKey(proof);
var claims = new List<Claim>
{
    new (ClaimTypes.NameIdentifier, "user"),
    new (ClaimTypes.Name, "Some User Name"),
    new (ClaimTypes.Role, "user_role")
};
var identity = new ClaimsIdentity(claims, "Saml2Repro", ClaimTypes.Name, ClaimTypes.Role);
var descriptor = new RequestedSecurityTokenDescriptor()
{
    Issuer = "urn:Saml2Repro",
[...]

PR #3207, opened 2025-04-14
Cline fix for Type attribute of SubjectConfirmationData has incorrect …

jennyf19:
…casing
Possible fix for #3206  from Cline VS code extension


westin-m (approved):

gislikonrad:
Thanks for the quick response, I would like to point out that reading will still be broken. The following line will never get the correct type and the validation in the following if statement will never be run.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dff18d775f3a1492da96b0d69086d57327718a0b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2Serializer.cs#L1284

PR #3220, opened 2025-05-06
Remove Microsoft.Bcl.Memory 9.0 in Wilson

### LINKED ISSUE #3099: netstandard 2.0 build depends on Microsoft.Bcl.Memory causing conflict with Indexes provided by IndexRange
Microsoft.IdentityModel.Tokens 8.3.1

**Where is the issue?**
Microsoft.IdentityModel.Tokens 8.3.1 netstandard 2.0 build Depends on Microsoft.Bcl.Memory

Existing app, Api .Net 4.7.2, working fine with Microsoft.IdentityModel.Tokens 8.3.0

**Repro**
Using IndexRange to provide support for language features of index, especially
```c#
var aList = someArray.ToList();
string last = actions[^1];  // <<<<<
```
Also see https://github.com/bgrainger/IndexRange/issues/14

**Expected behavior**
Everything compiling without issues, and no major change since it is only a minor version change. (trying to be optimistic at times)
(Also wouldn't expect Microsoft.Bcl.Memory to break IndexRange usage, but that might be a more difficult issue to resolve.)
Change of package use would be nice to have in release notes with explanation.
Introduced in #2817

**Actual behavior**
Compile time errors
```
CS0656 Missing compiler required member 'System.Index..ctor'
CS0518 Predefined type 'System.Index' is not defined or imported
CS0518 Predefined type 'System.Range' is not defined or imported
```

**Possible solution**
Is `Microsoft.Bcl.Memory` required as a dependency in `netstandard` builds?
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/658604a398105c95fb229601ea3a91849cc6a145/src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj#L68
 `<ItemGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))">`

**Current workaround**
Block version 8.3.1 and later

### LINKED ISSUE #3224: Remove Microsoft.Bcl.Memory 9.0 in Wilson
Associated GitHub Issue in Wilson: [[Bug] fix dependent packages · Issue #3208 · AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/3208)

**Proposal Part 1**
- Microsoft.Bcl.Memory is only for pre-net9 TFM; the new APIs are available in .NET 9. Remove this dependency and restore older implementation for netstandard2.0 and lower TFMs.

mdchennu:
### Remove Microsoft.Bcl.Memory 9.0 in Wilson

## Description
This PR removes the reference to Microsoft.Bcl.Memory in Wilson, and replaces it with the older version referenced here: [Use new Base64Url API (#2817) · AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet@c0eb23c](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/c0eb23c1673e34c9343a5d432b523ba19ae31120). 

The proposal includes three steps as below. This PR covers the first step listed below. 
### Proposal:
Covered in this PR
- Microsoft.Bcl.Memory is only for pre-net9 TFM; the new APIs are available in .NET 9. Remove this dependency and restore older implementation for netstandard2.0 and lower TFMs.

Fixes [#3224](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/3224)  (in this specific format)

For Testing:
1. Updated code
2. Ran all the tests for TFMS + all of them passed
3. Create a unified build against my branch -> build was successful: https://identitydivision.visualstudio.com/DevEx/_build/results?buildId=1476634&view=results

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
@@ -3,7 +3,7 @@
 
 using System;
 using System.Buffers;
-#if NET9_0
+#if NET9_0_OR_GREATER
 using System.Buffers.Text;
 #endif
 using System.Text;
--- a/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Buffers;
+#if NET9_0
 using System.Buffers.Text;
+#endif
 using System.Text;
 using Microsoft.IdentityModel.Logging;
 
@@ -21,6 +23,8 @@ namespace Microsoft.IdentityModel.Tokens
         private const char Base64PadCharacter = '=';
         private const char Base64Character62 = '+';
         private const char Base64Character63 = '/';
+        private const char Base64UrlCharacter62 = '-';
+        private const char Base64UrlCharacter63 = '_';
 
         /// <summary>
         /// Performs base64url encoding, which differs from regular base64 encoding as follows:
@@ -98,9 +102,17 @@ namespace Microsoft.IdentityModel.Tokens
                         LogHelper.MarkAsNonPII(inArray.Length))));
 #pragma warning restore CA2208 // Instantiate argument exceptions correctly
 
+#if NET9_0
             return Base64Url.EncodeToString(inArray.AsSpan().Slice(offset, length));
+#else
+            char[] destination = new char[(inArray.Length + 2) / 3 * 4];
+            int j = Encode(inArray.AsSpan<byte>().Slice(offset, length), destination.AsSpan<char>());
+
+            return new string(destination, 0, j);
+#endif
         }
 
+#if NET9_0
         /// <summary>
         /// Populates a <see cref="Span{T}"/> with the base64url encoded representation of a <see cref="ReadOnlySpan{T}"/> of bytes.
         /// </summary>
@@ -108,7 +120,68 @@ namespace Microsoft.IdentityModel.Tokens
         /// <param name="output">The span of characters to write the encoded output.</param>
         /// <returns>The number of characters written to the output span.</returns>
         public static int Encode(ReadOnlySpan<byte> inArray, Span<char> output) => Base64Url.EncodeToChars(inArray, output);
+#else
+        /// <summary>
+        /// Populates a <see cref="Span{T}"/> with the base64url encoded representation of a <see cref="ReadOnlySpan{T}"/> of bytes.
+        /// </summary>
+        /// <param name="inArray">A read-only span of bytes to encode.</param>
+        /// <param name="output">The span of characters to write the encoded output.</param>
+        /// <returns>The number of characters written to the output span.</returns>
+        public static int Encode(ReadOnlySpan<byte> inArray, Span<char> output)
+        {
+            int lengthmod3 = inArray.Length % 3;
+            int limit = (inArray.Length - lengthmod3);
+            ReadOnlySpan<byte> table = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_"u8;
 
+            int i, j = 0;
+
+            // takes 3 bytes from inArray and insert 4 bytes into output
+            for (i = 0; i < limit; i += 3)
+            {
+                byte d0 = inArray[i];
+                byte d1 = inArray[i + 1];
+                byte d2 = inArray[i + 2];
+
+                output[j + 0] = (char)table[d0 >> 2];
+                output[j + 1] = (char)table[((d0 & 0x03) << 4) | (d1 >> 4)];
+                output[j + 2] = (char)table[((d1 & 0x0f) << 2) | (d2 >> 6)];
+                output[j + 3] = (char)table[d2 & 0x3f];
+                j += 4;
+            }
+
+            //Where we left off before
+            i = limit;
+
+            switch (lengthmod3)
+            {
+                case 2:
+                    {
+                        byte d0 = inArray[i];
+                        byte d1 = inArray[i + 1];
+
+                        output[j + 0] = (char)table[d0 >> 2];
+                        output[j + 1] = (char)table[((d0 & 0x03) << 4) | (d1 >> 4)];
+                        output[j + 2] = (char)table[(d1 & 0x0f) << 2];
+                        j += 3;
+                    }
+                    break;
+                case 1:
+                    {
+                        byte d0 = inArray[i];
+
+                        output[j + 0] = (char)table[d0 >> 2];
+                        output[j + 1] = (char)table[(d0 & 0x03) << 4];
+                        j += 2;
+                    }
+                    break;
+
+                    //default or case 0: no further operations are needed.
+            }
+
+            return j;
+        }
+
+#endif
         /// <summary>
         /// Converts the specified base64url encoded string to UTF-8 bytes.
         /// </summary>
@@ -123,6 +196,8 @@ namespace Microsoft.IdentityModel.Tokens
 #if NETCOREAPP
         [SkipLocalsInit]
 #endif
+
+#if NET9_0
         internal static byte[] Decode(ReadOnlySpan<char> strSpan)
         {
             int upperBound = Base64Url.GetMaxDecodedLength(strSpan.Length);
@@ -144,6 +219,24 @@ namespace Microsoft.IdentityModel.Tokens
                     ArrayPool<byte>.Shared.Return(rented, true);
             }
         }
+#else
+        internal static byte[] Decode(ReadOnlySpan<char> strSpan)
+        {
+            int mod = strSpan.Length % 4;
+            if (mod == 1)
+                throw LogHelper.LogExceptionMessage(new FormatException(LogHelper.FormatInvariant(LogMessages.IDX10400, strSpan.ToString())));
+
+            bool needReplace = strSpan.IndexOfAny(Base64UrlCharacter62, Base64UrlCharacter63) >= 0;
+            int decodedLength = strSpan.Length + (4 - mod) % 4;
+#if NET6_0_OR_GREATER
+            Span<byte> output = new byte[decodedLength];
+            int length = Decode(strSpan, output, needReplace, decodedLength);
+            return output.Slice(0, length).ToArray();
+#else
+            return UnsafeDecode(strSpan, needReplace, decodedLength);
+#endif
+        }
+#endif
 
 #if !NET8_0_OR_GREATER
         private static bool IsOnlyValidBase64Chars(ReadOnlySpan<char> strSpan)
--- a/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
@@ -238,32 +238,18 @@ namespace Microsoft.IdentityModel.Tokens
         }
 #endif
 
-#if !NET8_0_OR_GREATER
-        private static bool IsOnlyValidBase64Chars(ReadOnlySpan<char> strSpan)
-        {
-            foreach (char c in strSpan)
-                if (!char.IsDigit(c) && !char.IsLetter(c) && c != Base64Character62 && c != Base64Character63 && c != Base64PadCharacter)
-                    return false;
-
-            return true;
-        }
-
-#endif
 #if NETCOREAPP
         [SkipLocalsInit]
 #endif
+
+#if NET9_0
         internal static int Decode(ReadOnlySpan<char> strSpan, Span<byte> output)
         {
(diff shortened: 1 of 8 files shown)


src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -98,17 +102,86 @@ public static string Encode(byte[] inArray, int offset, int length)
                         LogHelper.MarkAsNonPII(inArray.Length))));
 #pragma warning restore CA2208 // Instantiate argument exceptions correctly
 
+#if NET9_0
             return Base64Url.EncodeToString(inArray.AsSpan().Slice(offset, length));
```
mdchennu, line 106: Should this be NET9_0_OR_GREATER? or left as NET9_0?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -144,33 +219,37 @@ internal static byte[] Decode(ReadOnlySpan<char> strSpan)
...
+            int length = Decode(strSpan, output, needReplace, decodedLength);
+            return output.Slice(0, length).ToArray();
+#else
+            return UnsafeDecode(strSpan, needReplace, decodedLength);
+#endif
         }
-
 #endif
+
 #if NETCOREAPP
```
mdchennu, line 241: What is the difference between NETCOREAPP and NET9_0? Why do we use NETCOREAPP?

mdchennu:
One more comment: I'm splitting the work into two PRs as below: 

**Covered in this PR**

Microsoft.Bcl.Memory is only for pre-net9 TFM; the new APIs are available in .NET 9. Remove this dependency and restore older implementation for netstandard2.0 and lower TFMs.

**Covered in next PR**

Remove System.Text.Json reference for net9.0 TFM
Update Microsoft.Extensions.Logging.Abstractions to 8.0.0 for .NET9.0 from 8.0.2

I was thinking this makes sense because: 
1. To separate out changes
2. reduce size of PR for ease of review

Is that alright to do or should I keep it all in one?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
pmaytak: NET9_0_OR_GREATER. Generally we want the X_OR_GREATER flags since these new APIs will be available in the future .NET versions.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
pmaytak: NETCOREAPP includes other newer frameworks, NET 10/9/8/7/6, etc. So we use NETCOREAPP for newer APIs, as opposed to older Net Framework APIs (net472, net462, etc.)

pmaytak: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation

pmaytak:
> One more comment: I'm splitting the work into two PRs as below:
> 
> **Covered in this PR**
> 
> Microsoft.Bcl.Memory is only for pre-net9 TFM; the new APIs are available in .NET 9. Remove this dependency and restore older implementation for netstandard2.0 and lower TFMs.
> 
> **Covered in next PR**
> 
> Remove System.Text.Json reference for net9.0 TFM Update Microsoft.Extensions.Logging.Abstractions to 8.0.0 for .NET9.0 from 8.0.2
> 
> I was thinking this makes sense because:
> 
> 1. To separate out changes
> 2. reduce size of PR for ease of review
> 
> Is that alright to do or should I keep it all in one?

+1 to separating the PRs.

sruke:
> > One more comment: I'm splitting the work into two PRs as below:
> > **Covered in this PR**
> > Microsoft.Bcl.Memory is only for pre-net9 TFM; the new APIs are available in .NET 9. Remove this dependency and restore older implementation for netstandard2.0 and lower TFMs.
> > **Covered in next PR**
> > Remove System.Text.Json reference for net9.0 TFM Update Microsoft.Extensions.Logging.Abstractions to 8.0.0 for .NET9.0 from 8.0.2
> > I was thinking this makes sense because:
> > 
> > 1. To separate out changes
> > 2. reduce size of PR for ease of review
> > 
> > Is that alright to do or should I keep it all in one?
> 
> +1 to separating the PRs.

Like the idea of splitting the work this way.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
mdchennu: Okay, thank you! For the places where I am using `NET9_0_OR_GREATER` should I replace to `NETCOREAPP`?

kllysng: Check [this](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#:~:text=%2C%20NET5_0%2C-,NETCOREAPP,-%2C%20NETCOREAPP3_1%2C) out. Hopefully it's highlighted correctly, otherwise you might have to scroll down to the chart under "Conditional Compliation". `#if NETCOREAPP` for us means if we're on .net 6, .net 8, or .net 9 because those are the NETCOREAPP frameworks that we support. See our supported fw's [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/bc001335c1715ad38c8bc60313609515b081486b/build/targets.props#L3). So basically, this line is saying run this code if we're on .net 6, 8, or 9, and not if we're on net462, net472, or netstandard2.0 (the other fw's that we support).

edit: sorry, I must have had this open a while ago and didn't refresh the page/see the existing convo

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -3,13 +3,36 @@
...
 using System;
 using System.Buffers;
+#if NET9_0
 using System.Buffers.Text;
+#endif
 using Microsoft.IdentityModel.Logging;
 
 namespace Microsoft.IdentityModel.Tokens
 {
     internal static class Base64UrlEncoding
```
sruke, line 18: Consider adding the following comment back for non-net9 targets.
// Base64 encode/decode implementation for as per https://tools.ietf.org/html/rfc4648#section-5.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -151,10 +182,16 @@ public static T Decode<T, TX, TY, TZ>(
             }
             finally
             {
+#if NET9_0_OR_GREATER
                 ArrayPool<byte>.Shared.Return(output, true);
```
sruke: Can ArrayPool be cleared by passing true for all targets?

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
```diff
@@ -284,5 +416,41 @@ internal static int ValidateAndGetOutputSize(ReadOnlySpan<char> strSpan, int off
             outputSize += (effectiveLength / 4) * 3;
             return outputSize;
         }
+#if !NET9_0_OR_GREATER
```
sruke, line 413: nit: Add new line

pmaytak:
We should remove the SuppressTfmSupportBuildWarnings flag and verify with the builds that we don't get the unsupported TFM warnings.

https://github.com/search?q=repo%3AAzureAD%2Fazure-activedirectory-identitymodel-extensions-for-dotnet%20Microsoft.Bcl.Memory&type=code

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
mdchennu: Okay got it! thank you Kelly!

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
mdchennu: added!

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
mdchennu: added back in with a comment about the implementation for net9.0

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoding.cs
mdchennu: omg genius. i tried and it worked wahoo. fixed.

mdchennu:
> We should remove the SuppressTfmSupportBuildWarnings flag and verify with the builds that we don't get the unsupported TFM warnings.
> 
> https://github.com/search?q=repo%3AAzureAD%2Fazure-activedirectory-identitymodel-extensions-for-dotnet%20Microsoft.Bcl.Memory&type=code

Okay sounds good. I removed the flag, and will check the builds for warnings once they are ready.

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -98,17 +102,86 @@ public static string Encode(byte[] inArray, int offset, int length)
...
         /// <returns>The number of characters written to the output span.</returns>
         public static int Encode(ReadOnlySpan<byte> inArray, Span<char> output) => Base64Url.EncodeToChars(inArray, output);
+#else
+        /// <summary>
+        /// Populates a <see cref="Span{T}"/> with the base64url encoded representation of a <see cref="ReadOnlySpan{T}"/> of bytes.
+        /// </summary>
+        /// <param name="inArray">A read-only span of bytes to encode.</param>
+        /// <param name="output">The span of characters to write the encoded output.</param>
+        /// <returns>The number of characters written to the output span.</returns>
+        public static int Encode(ReadOnlySpan<byte> inArray, Span<char> output)
```
brentschmaltz, line 130: do we need to check the size of 'output'?

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
```diff
@@ -65,10 +65,6 @@
     <PackageReference Include="Microsoft.Bcl.TimeProvider" Version="$(MicrosoftBclTimeProviderVersion)" />
   </ItemGroup>
 
-  <ItemGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))">
-    <PackageReference Include="Microsoft.Bcl.Memory" Version="$(MicrosoftBclMemory)" />
```
pmaytak, line 69: Can also remove this:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/23f88d69061ebde619d2955ae7223b39b9c409a8/build/dependencies.props#L15

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
mdchennu: I'm sorry, I think I missed updating this file earlier. All previous references to NET9_0 should now be NET9_0_OR_GREATER now.

pmaytak (approved):

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
mdchennu: I dont think we need to because the higher level Encode method, on line 108, allocates the size before calling the Encode method on line 130. 

image

This is the link within the code: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3220/files#diff-d8157f3c41ee79206c7c316a85950971914d143866bc8ca23dcfc8d3658338b3R108:~:text=%23else-,char%5B%5D,-destination%20%3D%20new

What do you think?

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
mdchennu: updated and removed!

sruke:
nit: Consider moving the preprocessor directives inside the method if all targets share the same API signature. To keep the file clean, you could introduce a static class (e.g., LegacyBaseUrlEncoder) for pre-.NET 9 implementations, either in this PR or a follow-up cleanup.

sruke (approved):

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
mdchennu: Resolving this comment as per the other PR updates

PR #3222, opened 2025-05-07
Add ReadOnlyMemory based overload for CanReadToken

### LINKED ISSUE #3221: [Feature Request] ReadOnlySpan based overload for JsonWebTokenHandler.CanReadToken
Not related to a problem, new feature to minimize allocations.

A new overload for the CanReadToken method that accepts `ReadOnlySpan<char>` similar to https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/bc001335c1715ad38c8bc60313609515b081486b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs#L129

None

**Additional context**
None

henrikwidlund:
# Add Memory based overload for CanReadToken

## Description

Added an overload for the `CanReadToken` in `JsonWebTokenHandler` and `JwtSecurityTokenHandler` with `ReadOnlyMemory<char>` to reduce allocations and adjusted the `string` based overload to call the new method.

New overload is only available on dotnet 8 and above as the span based regex methods used to match JWS/JWE isn't available in dotnet 6.

Fixes #3221


henrikwidlund:

henrikwidlund:
I did not add benchmarks, even though this is a performance improvement. The reason being that the string based API calls the span based one.

kllysng:
Thanks for your contribution @henrikwidlund . Are you planning to add unit tests?

henrikwidlund:
@kllysng I was thinking that the regular string based ones would cover it since the string based method calls the memory based one. Let me know if you want explicit tests.

henrikwidlund:
I've added tests now either way. Also added the method to `JwtSecurityTokenHandler`. Was thinking of adding the memory based overload to `JwtSecurityTokenHandler.ReadToken`, but feels out of scope.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -118,6 +118,59 @@ public IDictionary<string, string> InboundClaimTypeMap
...
+                    LogHelper.LogInformation(TokenLogMessages.IDX10209, LogHelper.MarkAsNonPII(token.Length), LogHelper.MarkAsNonPII(MaximumTokenSizeInBytes));
+
+                return false;
+            }
+
+            // Count the number of segments, which is the number of periods + 1. We can stop when we've encountered
+            // more segments than the maximum we know how to handle.
+            int segmentCount = JwtTokenUtilities.CountJwtTokenPart(token.Span, JwtConstants.MaxJwtSegmentCount + 1);
+
+            switch (segmentCount)
```
brentschmaltz: this code (switch (segmentCount)) can be shared between JsonWebTokenHandler and JwtSecurityTokenHandler.

When you think about it, most of this code could be shared between JsonWebTokenHandler and JwtSecurityTokenHandler.

brentschmaltz (changes requested):
refactor and share common code between JsonWebTokenHandler and JwtSecurityTokenHandler.

henrikwidlund:
Good points @brentschmaltz, I've applied your suggestions.

sruke (approved):
LGTM, thank you for taking the suggestions and making the contribution @henrikwidlund.

henrikwidlund:
I see that a few unit tests are failing because of accessibility. I'll make the method internal again tomorrow

henrikwidlund:
@sruke fixed the compiler error

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
henrikwidlund: @brentschmaltz please resolve if you're satisfied with the solution :)

sruke:
@henrikwidlund, a few unit tests are still failing. We’re planning to release IdentityModel today—would it be okay to include this PR in the next release (in a month) instead?

henrikwidlund:
@sruke I believe I've fixed the test (was a real bug). I leave it up to you if it's getting to tight for your deadline (although, it would be nice to have it :))

henrikwidlund:
I noticed that the PR build is still failing, but I don't understand if it's related to my code changes?

sruke:
> I noticed that the PR build is still failing, but I don't understand if it's related to my code changes?

The failure (Resource not accessible by integration) appears unrelated to recent changes. All tests passed. I’m rerunning the build to check if the issue resolves. This change wasn’t included in the previous IdentityModel release but will be part of the upcoming one.

henrikwidlund:
> > I noticed that the PR build is still failing, but I don't understand if it's related to my code changes?
> 
> 
> 
> The failure (Resource not accessible by integration) appears unrelated to recent changes. All tests passed. I’m rerunning the build to check if the issue resolves. This change wasn’t included in the previous IdentityModel release but will be part of the upcoming one.

Notice that it didn't help 😕

henrikwidlund:
@sruke @brentschmaltz ping, any ideas what the problem is and when this can be merged?

henrikwidlund:
@sruke @brentschmaltz ping again :/

src/Microsoft.IdentityModel.JsonWebTokens/JwtTokenUtilities.cs
```diff
@@ -580,30 +580,72 @@ internal static SecurityKey ResolveTokenSigningKey(string kid, string x5t, IEnum
...
+        }
+
         /// <summary>
         /// Counts the number of JWT token segments.
         /// </summary>
         /// <param name="token">The JWT token.</param>
         /// <param name="maxCount">The maximum number of segments to count up to.</param>
         /// <returns>The number of segments up to <paramref name="maxCount"/>.</returns>
-        internal static int CountJwtTokenPart(string token, int maxCount)
+        internal static int CountJwtTokenPart(ReadOnlySpan<char> token, int maxCount)
```
brentschmaltz, line 618: Why do we need this as ReadOnlySpan?

We don't like to remove the old method as that can cause issues when mixed versions are used.
We end up with a MethodNotFound exception

henrikwidlund: We need it because without it the PR would not make sense as we'd need to work with strings again. If you want I can create a method that takes string and calls this one.

brentschmaltz: @henrikwidlund i was thinking why ReadOnlySpan and not ReadOnlyMemory?

henrikwidlund: Only reasons are passing a smaller object and that that's what the method works with. If that's what you're thinking about I don't understand your comment about method not found as that would happen regardless.

torarnegustad: @brentschmaltz I believe older versions of .NET does not support regular expressions for `ReadOnlySpan`, and this library targets older versions of .NET. This is the reason `ReadOnlyMemory` is used in the `CanReadToken` method.

`ReadOnlySpan` is more efficient and what libraries should use/support whenever possible.

I agreee that supporting `string` would be useful, even if it is basically just a wrapper for the new method.

brentschmaltz: @henrikwidlund
> Only reasons are passing a smaller object and that that's what the method works with. If that's what you're thinking about I don't understand your comment about method not found as that would happen regardless.

If we remove a method in M.IM.Tokens in say, 8.81, then it is possible that M.IM.JsonWebTokens 8.80 might have a call that expected that method. This is particularly problematic for internal methods.

henrikwidlund: So what's your suggestion, add back the string based one and have that one call the span one, or close the PR?

brentschmaltz: > So what's your suggestion, add back the string based one and have that one call the span one, or close the PR?

Yes, that is what we are thinking.

henrikwidlund: Which of them? I can do both, but it would be a bit of a waste 😅

brentschmaltz: add back the string based one and have that one call the span one,

henrikwidlund: Okay, done :)

henrikwidlund:
Is this PR still wanted? It's been open for four months now. @brentschmaltz @sruke @jennyf19

pmaytak:
@henrikwidlund Sorry about the delay. I'm on holiday next week. I'll bring this up with the team, if there's available bandwidth; otherwise, I'll follow up on this after I'm back.

Issue #3225, opened 2025-05-08
Microsoft.Extensions.Logging.Abstractions update


mdchennu:
Part Two of the Proposal 
- Remove System.Text.Json reference for net9.0 TFM 
- Update Microsoft.Extensions.Logging.Abstractions to 8.0.0 for .NET9.0 from 8.0.2

Issue #3208, opened 2025-04-15
[Bug] fix dependent packages


sebastienros:
Microsoft.IdentityModel.Tokens 8.8.0

The package depends on `Microsoft.Bcl.Memory 9.0` even for TFMs lower than 9.0. Even though this package has specific targets for net8.0 and netstandard2.0 and hence "works" with these, this should not be used in LTS tfms or users will be left in support limbo when dotnet 9.0 becomes not supported anymore. AFAIK as this library is part of the dotnet runtime product it won't be patched once 9.0 os out of support. Right now there are no version of IM that have neither known vulnerability or long term support dependencies.

kevinchalet:
Note: Wilson uses that package to polyfill base64url on older TFMs so the old implementation (that wasn’t as efficient but worked fine) will need to be brought back for anything that isn’t .NET 9.0+.

sebastienros:
Some nits if anything is worked on:

- Microsoft.Identity.Web.Tokens

![Image](https://github.com/user-attachments/assets/76e0f72d-16dd-477a-b0c2-bf9288b211a3)

Ideally it should reference the lowest abstractions version necessary, there is no need to force users in newer ones that may conflict with other dependencies.

- Microsoft.Identity.Web

System.Text.Json doesn't need to be referenced in net9.0 (there is no known vulnerability, which would be a reason to do so)

keegan-caruso:
See https://github.com/AzureAD/microsoft-identity-web/pull/3332 for STJ in Identity Web.

PR #3226, opened 2025-05-08
[For Net9.0 TFM] Change MicrosoftExtensionsLoggingAbstractions from 8.0.2 to 8.0.0

mdchennu:
# [For Net9.0 TFM] Change Abstractions from 8.0.2 to 8.0.0

## Description

This PR updates Msft.Extensions.Logging.Abstractions to 8.0.0 for .Net9.0

**Covered in PR**
Update Microsoft.Extensions.Logging.Abstractions to 8.0.0 for .NET9.0 from 8.0.2
Fixes #{[3225](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/3225)} (in this specific format)

For testing:
1. Built the solution -> clean build for all .NET9 Frameworks
2. Ran the tests -> all passed

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -19,7 +19,7 @@
     <!--
       Do not bump to 9.0. complicated reasons due to some runtime incompatibilities if 
       CoreWCF compiles against an older version but then run with a newer version in a related package.
-      Also, moved to 8.0.0 due to removal of Microsoft.Bcl.Memory 9.0 in Wilson
+      Also, moved to 8.0.0 to standardize across all versions
     -->
     <MicrosoftExtensionsLoggingAbstractionsVersion>8.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
   </PropertyGroup>
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -19,9 +19,10 @@
 <PropertyGroup Condition="'$(TargetFramework)' == 'net9.0'">
     <!--
       Do not bump to 9.0. complicated reasons due to some runtime incompatibilities if 
-      CoreWCF compiles against an older version but then run with a newer version in a related package
+      CoreWCF compiles against an older version but then run with a newer version in a related package.
+      Also, moved to 8.0.0 due to removal of Microsoft.Bcl.Memory 9.0 in Wilson
     -->
-    <MicrosoftExtensionsLoggingAbstractionsVersion>8.0.2</MicrosoftExtensionsLoggingAbstractionsVersion>
+    <MicrosoftExtensionsLoggingAbstractionsVersion>8.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Microsoft.IdentityModel.Protocols.OpenIdConnect.csproj
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Microsoft.IdentityModel.Protocols.OpenIdConnect.csproj
@@ -28,8 +28,12 @@
     <ProjectReference Include="..\System.IdentityModel.Tokens.Jwt\System.IdentityModel.Tokens.Jwt.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="$(TargetFramework) == 'net6.0' or $(TargetFramework) == 'net8.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' And '$(TargetFramework)' != 'net9.0' ">
     <PackageReference Include="System.Text.Json" Version="$(SystemTextJson)" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Microsoft.IdentityModel.Protocols.OpenIdConnect.csproj
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Microsoft.IdentityModel.Protocols.OpenIdConnect.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <Import Project="..\..\build\common.props" />
 
@@ -28,7 +28,7 @@
     <ProjectReference Include="..\System.IdentityModel.Tokens.Jwt\System.IdentityModel.Tokens.Jwt.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' And !$(TargetFramework.StartsWith('net9.0'))">
+  <ItemGroup Condition="$(TargetFramework) == 'net6.0' or $(TargetFramework) == 'net8.0'">
     <PackageReference Include="System.Text.Json" Version="$(SystemTextJson)" />
   </ItemGroup>
 
--- a/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Microsoft.IdentityModel.Protocols.OpenIdConnect.csproj
+++ b/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Microsoft.IdentityModel.Protocols.OpenIdConnect.csproj
@@ -28,12 +28,8 @@
     <ProjectReference Include="..\System.IdentityModel.Tokens.Jwt\System.IdentityModel.Tokens.Jwt.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' And '$(TargetFramework)' != 'net9.0' ">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
     <PackageReference Include="System.Text.Json" Version="$(SystemTextJson)" />
-  </ItemGroup>
-
-  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
-    <FrameworkReference Include="Microsoft.AspNetCore.App" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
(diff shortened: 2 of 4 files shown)


src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
```diff
@@ -57,7 +57,7 @@
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="$(SystemDiagnosticSourceVersion)" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' And !$(TargetFramework.StartsWith('net9.0'))">
```
kllysng: Our NetCoreApp target fw's are [net6, net8, and net9](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/35caa6cb0b693188b5c5c1b9434414f1a3ba43ba/build/targets.props#L3C83-L3C104). Consider being explicit like:  
```suggestion
  <ItemGroup Condition="$(TargetFramework) == 'net6.0' or $(TargetFramework) == 'net8.0'">
```

mdchennu: thank you! I've updated!

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
```diff
@@ -57,7 +57,7 @@
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="$(SystemDiagnosticSourceVersion)" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
+  <ItemGroup Condition="$(TargetFramework) == 'net6.0' or $(TargetFramework) == 'net8.0'">
```
kllysng: Sorry I missed this in my last comment, here and elsewhere, we also need single quotes around $(TargetFramework)
```suggestion
  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0' or '$(TargetFramework)' == 'net8.0'">
```

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
mdchennu: Wait, Kelly, I tried changing the condition to   `<ItemGroup Condition="'$(TargetFramework)'== 'net6.0' or '$(TargetFramework)' == 'net8.0'">: `

Then I got a bunch of build errors for the three `csproj` files updated with the condition above. The build errors were from all TFMs and were of two categories:

1. cannot find dlls
2. `The type or namespace name 'JsonIgnore' could not be found (are you missing a using directive or an assembly reference?`

I tried researching why and found this thru copilot hah: 
`The error occurs because in .NET 9.0, even though System.Text.Json is included as part of the framework, you need to add an explicit reference to the assembly through a using directive. This is due to how JsonIgnore and other Json-related attributes are organized in .NET.`

The explicit reference I added to the csproj files was:
```
<ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
  <FrameworkReference Include="Microsoft.AspNetCore.App" />
</ItemGroup>
```

This change still caused build errors similar to above, so now I updated the reference as follows: 
```
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' And '$(TargetFramework)' != 'net9.0' ">
    <PackageReference Include="System.Text.Json" Version="$(SystemTextJson)" />
  </ItemGroup>
```

What do you think?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Microsoft.IdentityModel.Protocols.OpenIdConnect.csproj
```diff
@@ -28,10 +28,14 @@
     <ProjectReference Include="..\System.IdentityModel.Tokens.Jwt\System.IdentityModel.Tokens.Jwt.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' And '$(TargetFramework)' != 'net9.0' ">
     <PackageReference Include="System.Text.Json" Version="$(SystemTextJson)" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
brentschmaltz: asp.net takes a dependency on M.IM.Protocols.OpenIdConnect, is there a circular reference here?

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
```diff
@@ -57,7 +57,7 @@
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="$(SystemDiagnosticSourceVersion)" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' And '$(TargetFramework)' != 'net9.0' ">
```
brentschmaltz: How will M.IM.Tokens use System.Text.Json?
For example, JsonSerializerPrimitives references Utf8JsonReader.

brentschmaltz (changes requested):
couple of comments about circular reference and System.Text.Json used by M.IM.Tokens.

build/dependencies.props
```diff
@@ -19,9 +19,10 @@
 <PropertyGroup Condition="'$(TargetFramework)' == 'net9.0'">
     <!--
       Do not bump to 9.0. complicated reasons due to some runtime incompatibilities if 
-      CoreWCF compiles against an older version but then run with a newer version in a related package
+      CoreWCF compiles against an older version but then run with a newer version in a related package.
+      Also, moved to 8.0.0 due to removal of Microsoft.Bcl.Memory 9.0 in Wilson
```
pmaytak: What does this mean? Why is Bcl.Memory removal relevant to this change?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Microsoft.IdentityModel.Protocols.OpenIdConnect.csproj
pmaytak: Why is this needed?

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Microsoft.IdentityModel.Protocols.OpenIdConnect.csproj
```diff
@@ -28,10 +28,14 @@
     <ProjectReference Include="..\System.IdentityModel.Tokens.Jwt\System.IdentityModel.Tokens.Jwt.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' And '$(TargetFramework)' != 'net9.0' ">
```
pmaytak: .NETCoreApp already includes net9.0.

pmaytak (changes requested):

pmaytak:
I don't think we need to do anything extra related to System.Text.Json. It's already only used for non-.NET Core TFMs.

https://www.nuget.org/packages/Microsoft.IdentityModel.Protocols.OpenIdConnect/8.9.0#dependencies-body-tab

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
kllysng: Hi Mallika! I pulled your branch and tried things out. Sorry, I misread the initial line and interpreted it as `$(TargetFramework) == NetCoreApp` instead of `!=`. 

You can remove the second half of the statement because any fw that's != NetCoreApp won't start with net9.0 and I've verified it locally

mdchennu:
> I don't think we need to do anything extra related to System.Text.Json. It's already only used for non-.NET Core TFMs.
> 
> https://www.nuget.org/packages/Microsoft.IdentityModel.Protocols.OpenIdConnect/8.9.0#dependencies-body-tab

Okay, so I was looking at the comments in the Bug description and it mentioned to make the following two changes (which I did in the PR):
image

Bug description: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3238459

Are both changes (highlighted in green) still necessary? @sruke @pmaytak

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
mdchennu: All good! thank you! I removed it

mdchennu:
> couple of comments about circular reference and System.Text.Json used by M.IM.Tokens.

Thank you for the comments! I've removed the circular reference, and removed the change I added in for usage of System.Text.Json. I'm going to ask Sruthi and Peter about if this change is still required (I think I'm having a hard time understanding why we need to make these changes for Bcl.Memory removal), and then follow up on this PR.

pmaytak:
> > I don't think we need to do anything extra related to System.Text.Json. It's already only used for non-.NET Core TFMs.
> > https://www.nuget.org/packages/Microsoft.IdentityModel.Protocols.OpenIdConnect/8.9.0#dependencies-body-tab
> 
> Okay, so I was looking at the comments in the Bug description and it mentioned to make the following two changes (which I did in the PR): image
> 
> Bug description: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3238459
> 
> Are both changes (highlighted in green) still necessary? @sruke @pmaytak

Related to System.Text.Json, the work item may just be incorrect. I took a look at a few Wilson packages on Nuget and none reference STJ for newer frameworks (correctly).

pmaytak (approved):

build/dependencies.props
mdchennu: Okay, I spoke to Sruthi! Bcl.Memory is not relevant to this change. It was included in the bug description because the customer left this feedback in the same github issue:

image

Link: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/3208

I think this change should still go in, but I'll remove the comments that relate it to Bcl.Memory work. What do you think? @sruke @pmaytak

sruke: Makes sense to me.

pmaytak (approved):

sruke (approved):

brentschmaltz (approved):

kllysng (approved):

PR #3230, opened 2025-05-09
Move suppression of RS006 to csproj.

brentschmaltz:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/tree/iinglese/use-experimental-namespace-for-new-validation-model

This above branch was merged with dev then moved to a new branch, with a couple of small changes:
1. Results are specific with ValidationResult<SecurityToken, ValidationError>specifying the Error type.
2. The suppression of RS006 is moved to the csproj file to reduce the number of file changes
3. Fix Benchmarks build issue.
4. SignatureValidator returns  ValidationResult<SecurityKey, SignatureValidationError>


src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ValidateToken.Internal.cs
```diff
@@ -137,12 +139,12 @@ internal override async Task<ValidationResult<ValidatedToken>> ValidateTokenAsyn
                 }
             }
 
-            ValidationResult<SecurityKey> signatureValidationResult = ValidateSignature(samlToken, validationParameters, callContext);
+            ValidationResult<SecurityKey, ValidationError> signatureValidationResult = ValidateSignature(samlToken, validationParameters, callContext);
```
sruke: Should this be SignatureValidationError?

src/Microsoft.IdentityModel.Tokens/Experimental/IResultBasedValidation.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// Callers should always check the ValidationResult.IsValid property to verify the validity of the result.
+        /// </summary>
+        /// <param name="token">The token to be validated.</param>
+        /// <param name="validationParameters">The <see cref="ValidationParameters"/> to be used for validating the token.</param>
+        /// <param name="callContext">A <see cref="CallContext"/> that contains call information.</param>
+        /// <returns>A <see cref="ValidationResult{TResult, TError}"/> with either a <see cref="ValidatedToken"/> if the token was validated or an <see cref="ValidationError"/> with the failure information and exception otherwise.</returns>
+        public Task<ValidationResult<ValidatedToken, ValidationError>> ValidateTokenAsync(
+            string token,
+            ValidationParameters validationParameters,
+            CallContext callContext);
```
sruke, line 27: This async API doesn’t accept a cancellation token.
Is it being introduced for backward compatibility with the existing model?

sruke (approved):
Approved with two questions.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ValidateToken.Internal.cs
brentschmaltz: @sruke I agree, would you agree that SignatureValidationDelegate should return ValidationResult<SecurityKey, SignatureValidationError>?

src/Microsoft.IdentityModel.Tokens/Experimental/IResultBasedValidation.cs
brentschmaltz: @sruke @jmprieur there was some discussion about this where we wanted an API without the cancellation token.
I believe the POR is to add an additional api with a cancellation token, is this understanding correct?

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ValidateToken.Internal.cs
sruke: Yes! That makes sense!

brentschmaltz: @sruke i made the change.

sruke (approved):

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.ValidateToken.Internal.cs
sruke: LGTM! Thank you.

src/Microsoft.IdentityModel.Tokens/Experimental/IResultBasedValidation.cs
sruke: Considering there are two public async APIs, one that accepts a cancellation token and one that doesn't (likely as a convenience for those developers who don't need cancellation) and internally we set CancellationToken.None. I don't have a strong preference here and will resolve the conversation.

keegan-caruso (approved):

PR #3239, opened 2025-05-14
Reduce memory consumption by avoiding adding items when queue is full.

brentschmaltz:
Do not add items if cache is full or compacting.
Under heavy load, stress tests showed that the LRUCache was unable to keep up a change was made to avoid adding to the cache if we detect it was full or compacting.

A change to start TaskEventQ on startup instead of starting and stopping improves performance and simplifies the code.
We were not checking is a SignatureProvider was actually added to the cache. This is important as if the SignatureProvider is not added, the cache will not be able to properly apply the dispose logic and kernel objects could pile up.


src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -157,20 +132,14 @@ internal EventBasedLRUCache(
         /// Stop the event queue task.
         /// This is provided mainly for users who have unit tests that check for running task(s) to stop the task at the end of each test.
         /// </summary>
```
keegan-caruso, line 137: update the summary?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -157,20 +132,14 @@ internal EventBasedLRUCache(
...
-        /// Stop the event queue task immediately if it is running. This allows the task/thread to terminate gracefully.
-        /// Currently there is no unmanaged resource, if any is added in the future it should be disposed of in this method.
-        /// </summary>
-        private void StopEventQueueTaskImmediately() => _shouldStopImmediately = true;
+        internal void StopEventQueueTaskImmediately() => _stopEventQueueTask = true;
 
         private void AddActionToEventQueue(Action action)
         {
-            _eventQueue.Enqueue(action);
+            if (_eventQTask == null || _eventQTask.Status != TaskStatus.Running)
```
keegan-caruso: Is this thread safe?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -414,22 +371,9 @@ protected int CalculateNewCacheSize()
...
-        /// <summary>
-        /// This is the method that determines the end time for the event queue task.
-        /// The goal is to be able to track the incoming events and predict how long the task should run in order to
-        /// avoid a long running task and reduce the overhead costs of restarting tasks.
-        /// For example, maybe we can track the last three events' time and set the _eventQueueRunDurationInSeconds = 2 * average_time_between_events.
-        /// Note: tasks are based on thread pool so the overhead should not be huge but we should still try to minimize it.
-        /// </summary>
-        /// <returns>the time when the event queue task should end</returns>
-        private DateTime SetTaskEndTime()
+        public bool SetValue(TKey key, TValue value)
```
sruke: Consider naming it TrySetValue to indicate the operation could succeed/fail

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: @keegan-caruso i think the comment is still true. Help me understand what you are thinking.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: @sruke Good idea.
I kept the older methods as we had shipped them.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -444,11 +396,15 @@ public bool SetValue(TKey key, TValue value, DateTime expirationTime)
             if (_removeExpiredValues && expirationTime < DateTime.UtcNow)
                 return false;
 
+            if (Interlocked.CompareExchange(ref _compactValuesState, ActionQueuedOrRunning, ActionQueuedOrRunning) == ActionQueuedOrRunning)
```
keegan-caruso, line 401: Can you help me understand this?

Why do we use ActionQueuedOrRunning for both the value and the comparand?

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -50,34 +50,22 @@ internal class EventBasedLRUCache<TKey, TValue>
...
         private const int ActionNotQueued = 0; // compaction action not in the event queue
         private const int ActionQueuedOrRunning = 1; // compaction action in the event queue or currently in progress
+
         private int _compactValuesState = ActionNotQueued;
         private int _removeExpiredValuesState = ActionNotQueued;
         private int _processCompactedValuesState = ActionNotQueued;
+        private readonly Task _eventQTask;
+        private int _eventQueuePollingInterval = 50; // in milliseconds
+        // for testing purpose only to verify the task count
+        private int _taskCount;
```
keegan-caruso: this doesn't look relevant anymore.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -50,34 +50,22 @@ internal class EventBasedLRUCache<TKey, TValue>
...
-        private int _eventQueueTaskState = EventQueueTaskStopped;
-
         private const int ActionNotQueued = 0; // compaction action not in the event queue
         private const int ActionQueuedOrRunning = 1; // compaction action in the event queue or currently in progress
+
         private int _compactValuesState = ActionNotQueued;
         private int _removeExpiredValuesState = ActionNotQueued;
         private int _processCompactedValuesState = ActionNotQueued;
+        private readonly Task _eventQTask;
+        private int _eventQueuePollingInterval = 50; // in milliseconds
```
keegan-caruso, line 63:
```suggestion
        private readonly int _eventQueuePollingInterval = 50; // in milliseconds
```

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -134,9 +111,8 @@ internal EventBasedLRUCache(
             _removeExpiredValues = removeExpiredValues;
             _compactIntervalInSeconds = compactIntervalInSeconds;
             _timeForNextExpiredValuesRemoval = DateTime.UtcNow.AddSeconds(_removeExpiredValuesIntervalInSeconds);
-            _timeForNextCompaction = DateTime.UtcNow.AddSeconds(_compactIntervalInSeconds);
-            _eventQueueTaskStopTime = DateTime.UtcNow;
             _maintainLRU = maintainLRU;
+            _eventQTask = Task.Run(EventQueueTaskAction);
```
keegan-caruso, line 117: There are a few comments that need updated, for example, line 160 talks about processing until the queue is empty.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: I agree, I will set it to 1.
There is an internal TaskCount method that we should keep.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: We just want to check if we are compacting without changing state.
This code seemed to meet that requirement.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
brentschmaltz: Tried to address all comments.

keegan-caruso (approved):
Looks right, there are a few failing tests that need fixed, otherwise LGTM

keegan-caruso:
Prefer #3374

PR #3250, opened 2025-05-29
Add expected Base64UrlEncoder.Decode overload for NET6 and 8

### LINKED ISSUE #3251: [Bug] MissingMethodException for Base64UrlEncoder.Decode
Base64UrlEncoder class was updated to remove a dependency on Microsoft.Bcl.Memory on non-NET9 frameworks. The return type of `Decode(ReadOnlySpan<char> strSpan, Span<byte> output)` was changed from int to void.

The issue is a NET6 app using IdentityModel.JsonWebTokens 8.9.0 expects this method to have an int return type; but it's void in IdentityModel.Tokens 8.10.0.

pmaytak:
Fixes #3251 

- Adds an overload for `internal static int Decode(ReadOnlySpan<char> strSpan, Span<byte> output)` with `int` return type for NET 6 and NET 8.
- Adds API tests.

| **Version**     | **TFM** | **Method return value** | **Uses new Base64 API** |
|------------------|---------|--------------------------|--------------------------|
| <= 8.3.0         | All     | void                     | No                       |
| 8.3.1 - 8.9.0    | All     | int                      | Yes                      |
| 8.10.0 - 8.11.0  | NET9    | int                      | Yes                      |
|                  | ELSE    | void                     | No                       |
| >= 8.12.0        | NET9    | int                      | Yes                      |
|                  | NET6, 8 | int                      | No                       |
|                  | ELSE    | void                     | No                       |

Working notes which can be disregarded (just didn't want to lose them):

> **<= 8.3.0**
> internal static void Decode(ReadOnlySpan<char> strSpan, Span<byte> output)
> 
> NET6+
> private static int Decode(ReadOnlySpan<char> strSpan, Span<byte> output, bool needReplace, int decodedLength)
> ELSE
> private static void Decode(ReadOnlySpan<char> strSpan, Span<byte> output, bool needReplace, int decodedLength)
> 
> **8.3.1 - 8.9.0**
> internal static int Decode(ReadOnlySpan<char> strSpan, Span<byte> output)
> 
> NETCOREAPP
> private static int Decode(ReadOnlySpan<char> strSpan, Span<byte> output, int decodedLength)
> ELSE
> private static int Decode(ReadOnlySpan<char> strSpan, Span<byte> output, int decodedLength)
> 
> **8.10.0 - 8.11.0**
> NET9
> int Decode(ReadOnlySpan<char> strSpan, Span<byte> output)
> ELSE
> void Decode(ReadOnlySpan<char> strSpan, Span<byte> output)
> 
> NET9
> private static int Decode(ReadOnlySpan<char> strSpan, Span<byte> output, int decodedLength)
> NET6, NET8
> private static int Decode(ReadOnlySpan<char> strSpan, Span<byte> output, bool needReplace, int decodedLength)
> ELSE
> private static void Decode(ReadOnlySpan<char> strSpan, Span<byte> output, bool needReplace, int decodedLength)


jennyf19 (approved):

test/Microsoft.IdentityModel.Tokens.Tests/Base64UrlEncodingTests.cs
```diff
@@ -343,5 +344,61 @@ public void EncodeDecode_InvalidParameters_ThrowsExceptionTests()
             Assert.Throws<ArgumentNullException>(static () => Base64UrlEncoding.Decode<object, object>("abc", 0, 0, null, null));
             Assert.Throws<ArgumentNullException>(static () => Base64UrlEncoding.Decode<object, object, object, object>(null, 0, 0, null, null, null, null));
         }
+
+        [Fact]
```
keegan-caruso, line 348: Do we need tests for any specific targets?

mdchennu (approved):
I'm really sorry about missing the different in returns for the Decode method, Peter. Thank you for noticing and fixing it. The change makes sense to me! I noticed a unit test (unrelated) was failing, so re-ran the job.

test/Microsoft.IdentityModel.Tokens.Tests/Base64UrlEncodingTests.cs
pmaytak: Just the one for Decode method below, I think, for NET6+/ELSE

src/Microsoft.IdentityModel.Tokens/Base64UrlEncoder.cs
```diff
@@ -259,6 +259,17 @@ internal static int Decode(ReadOnlySpan<char> strSpan, Span<byte> output)
 
             return Decode(strSpan, output, decodedLength);
         }
+#elif NET6_0_OR_GREATER
+        internal static int Decode(ReadOnlySpan<char> strSpan, Span<byte> output)
```
keegan-caruso, line 263: I would think we need to adjust the API files for this

keegan-caruso: It can be a follow up

keegan-caruso (approved):

PR #3265, opened 2025-06-16
Provide support to Log using ILogger

brentschmaltz:
This work creates an entry point for use of ILogger for capturing logs.

It sets a pattern that will be expanded to include additional logs, but first we want to establish an acceptable model.
This PR focuses on Validation of Lifetime, Issuer and Audience and will be expanded with other PR's.


src/Microsoft.IdentityModel.Tokens/Extrensions/TokenValidationParametersExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public static class TokenValidationParametersExtensions
+    {
+#pragma warning disable CS8601 // Possible null reference assignment.
+        internal static string s_callContextKey = typeof(TokenValidationParameters).FullName + ".CallContext";
+#pragma warning restore CS8601 // Possible null reference assignment.
```
keegan-caruso:
```suggestion
        internal static string s_callContextKey = typeof(TokenValidationParameters).FullName! + ".CallContext";
```

src/Microsoft.IdentityModel.Tokens/Extrensions/TokenValidationParametersExtensions.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.IdentityModel.Tokens
+{
+    /// <summary>
+    ///
```
keegan-caruso: fill out

src/Microsoft.IdentityModel.Tokens/Extrensions/TokenValidationParametersExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+#pragma warning disable CS8601 // Possible null reference assignment.
+        internal static string s_callContextKey = typeof(TokenValidationParameters).FullName + ".CallContext";
+#pragma warning restore CS8601 // Possible null reference assignment.
+
+        /// <summary>
+        /// Adds the provided <see cref="ILogger"/> to the <see cref="TokenValidationParameters"/>.
+        /// </summary>
+        /// <param name="validationParameters">The <see cref="TokenValidationParameters"/> instance.</param>
+        /// <param name="logger">The <see cref="ILogger"/> to use.</param>
+#pragma warning disable CS3001 // Argument type is not CLS-compliant
```
keegan-caruso: I think we should just mark this extension type as not cls compliant

src/Microsoft.IdentityModel.Tokens/Extrensions/TokenValidationParametersExtensions.cs
keegan-caruso: There are a few other in this file as well

src/Microsoft.IdentityModel.Tokens/CallContext.cs
```diff
@@ -24,5 +25,37 @@ public CallContext() : base()
         public CallContext(Guid activityId) : base(activityId)
         {
         }
+#pragma warning disable CS3001 // ILogger is not CLSCompliant
+#pragma warning disable CS3003 // ILogger is not CLSCompliant
+        /// <summary>
+        /// Instantiates a new <see cref="CallContext"/> with an activity identifier.
+        /// </summary>
+        /// <param name="logger"><see cref="ILogger"/> to record logs.</param>
+        /// <param name="activityId">activity id to include in logs.</param>
```
pmaytak:
```suggestion
        /// <param name="activityId">Activity ID to include in logs.</param>
```

src/Microsoft.IdentityModel.Tokens/CallContext.cs
```diff
@@ -24,5 +25,37 @@ public CallContext() : base()
...
+        {
+        }
+
+        /// <summary>
+        /// Instantiates a new <see cref="CallContext"/> with a collelation id.
+        /// </summary>
+        /// <param name="logger"><see cref="ILogger"/> to record logs.</param>
+        /// <param name="correlationId">correlation id to include in logs.</param>
+        /// <exception cref="ArgumentNullException">if <paramref name="logger"/> is null.</exception>
+        public CallContext(ILogger logger, string correlationId) : base(logger, correlationId)
```
pmaytak, line 46: Do correlationId and activityId supposed to mean the same thing, just different type or are they different semantically? If they are the same, then the parameter names should be the same (change this to activityId to match the existing constructor). If different, clarify the difference in the comment.

src/Microsoft.IdentityModel.Logging/LoggerContext.cs
```diff
@@ -28,6 +29,55 @@ public LoggerContext(Guid activityId)
...
+        /// </summary>
+        public ILogger Logger { get; }
+
+#pragma warning restore CS3001 // ILogger is not CLSCompliant
+#pragma warning restore CS3003 // ILogger is not CLSCompliant
+
+        /// <summary>
+        /// Gets or set a <see cref="string"/> that will be logged.
+        /// </summary>
+        public string CorrelationId { get; set; }
```
pmaytak, line 80: Similar to the other comment - it's not clear how ActivityId and CorrelationId are similar/different. Can we reuse ActivityId and call ToString on it when needed? Or if we'd rather have the string property, obsolete the ActivityId and use CorrelationId everywhere.

test/Microsoft.IdentityModel.Logging.Tests/PIITests.cs
```diff
@@ -24,6 +25,7 @@ public class PIITests
         [Fact]
         public void LogOpenIdConnectProtocolExceptions()
         {
+            IdentityModelEventSource.ShowPII = false;
```
pmaytak: nit: this can also be set in the test class constructor which is called before each test method invocation.

src/Microsoft.IdentityModel.Tokens/CallContext.cs
brentschmaltz: The activityId is a GUID and in a public release, so we can't change it.
I added the correlationId as a string as this is much more flexible.

test/Microsoft.IdentityModel.Logging.Tests/PIITests.cs
brentschmaltz: Oh, didn't know that, will make that change.

src/Microsoft.IdentityModel.Logging/Microsoft.IdentityModel.Logging.csproj
```diff
@@ -36,6 +36,10 @@
     <ProjectReference Include="..\Microsoft.IdentityModel.Abstractions\Microsoft.IdentityModel.Abstractions.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggingAbstractionsVersion)" />
```
pmaytak, line 40: @GeoK This is a concern of yours then?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.cs
```diff
@@ -480,6 +502,8 @@ internal async ValueTask<TokenValidationResult> ValidateTokenAsync(
             TokenValidationParameters validationParameters)
         {
             BaseConfiguration currentConfiguration = null;
+            CallContext callContext = TokenValidationParametersExtensions.GetCallContext(validationParameters);
```
pmaytak: Is this line needed?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -269,7 +281,38 @@ public static Exception LogExceptionMessage(EventLevel eventLevel, Exception exc
...
+        /// <param name="exception">The exception to log.</param>
+        /// <param name="loggerContext">The <see cref="LoggerContext"/> contains information useful for logging and debugging.</param>
+        public static Exception LogExceptionMessage(Exception exception, LoggerContext loggerContext)
+        {
+            if (exception == null)
+                return null;
+
+            LogExceptionMessage(exception);
+
+            if (loggerContext == null || loggerContext.Logger == null)
```
pmaytak: nit (and in the few other places in the class):
```suggestion
            if (loggerContext?.Logger == null)
```

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -269,7 +281,38 @@ public static Exception LogExceptionMessage(EventLevel eventLevel, Exception exc
...
+        /// Logs an exception using the event source logger.
+        /// </summary>
+        /// <param name="exception">The exception to log.</param>
+        /// <param name="loggerContext">The <see cref="LoggerContext"/> contains information useful for logging and debugging.</param>
+        public static Exception LogExceptionMessage(Exception exception, LoggerContext loggerContext)
+        {
+            if (exception == null)
+                return null;
+
+            LogExceptionMessage(exception);
```
pmaytak, line 319: This method logs using the EventSourceLogger. Should the Log methods that accept LoggerContext only log to ILogger? Otherwise, the method comments need to be updated that this also logs to ILogger.

src/Microsoft.IdentityModel.Tokens/Validators.cs
```diff
@@ -279,38 +301,61 @@ internal static async ValueTask<string> ValidateIssuerAsync(
...
+                    return validationParameters.IssuerValidatorUsingConfiguration(issuer, securityToken, validationParameters, configuration);
 
-            if (validationParameters.IssuerValidatorUsingConfiguration != null)
-                return validationParameters.IssuerValidatorUsingConfiguration(issuer, securityToken, validationParameters, configuration);
+                if (validationParameters.IssuerValidator != null)
+                    return validationParameters.IssuerValidator(issuer, securityToken, validationParameters);
+            }
+            catch (Exception exception)
+            {
+                if (LogHelper.IsEnabled(EventLogLevel.Error) || (callContext?.Logger?.IsEnabled(LogLevel.Error) ?? false))
```
pmaytak, line 359: Can this be extracted into LogHelper which will check if both loggers are set?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
brentschmaltz: I think it makes sense to log wherever users have enabled the logger.
This is the pattern that IdentityLogger uses.

src/Microsoft.IdentityModel.Logging/LoggerContext.cs
```diff
@@ -28,9 +29,60 @@ public LoggerContext(Guid activityId)
             ActivityId = activityId;
         }
 
+#pragma warning disable CS3001 // ILogger is not CLSCompliant
```
keegan-caruso, line 32: It would be better to mark these with

[CLSCompliant(false)]

src/Microsoft.IdentityModel.Logging/LoggerContext.cs
```diff
@@ -28,9 +29,60 @@ public LoggerContext(Guid activityId)
...
+        /// <summary>
+        /// Instantiates a new <see cref="LoggerContext"/> with a correlation identifier.
+        /// </summary>
+        /// <param name="logger"><see cref="ILogger"/> to record logs.</param>
+        /// <param name="correlationId">correlation id to include in logs.</param>
+        /// <exception cref="ArgumentNullException">if <paramref name="logger"/> is null.</exception>
+        public LoggerContext(ILogger logger, string correlationId)
+        {
+            Logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            CorrelationId = correlationId;
```
keegan-caruso, line 65: We could check if the correlation id is a guid, if it is, then set the activity id with it

test/Microsoft.IdentityModel.Logging.Tests/ILoggerTests.cs
```diff
@@ -0,0 +1,1103 @@
...
+                foreach (string logValue in theoryData.LogValues)
+                {
+                    foreach (var logEntry in theoryData.Logger.GetLogEntries())
+                    {
+                        if (logEntry.State.ToString().Contains(logValue))
+                            logCount++;
+                    }
+                }
+
+                Assert.True(logCount == theoryData.LogValues.Count,
```
pmaytak, line 227: nit: Use Assert.Equal for these.

pmaytak (approved):

saurabhsathe-ms:
Closed this pull request as I am actively working on this on a different branch!

PR #3268, opened 2025-06-20
Base classes for OperationResult and OperationError

brentschmaltz:
Presents a common model for a general OperationResult and OperationError.

### DIFF
--- a/src/Microsoft.IdentityModel.Abstractions/Experimental/OperationResult.cs
+++ b/src/Microsoft.IdentityModel.Abstractions/Experimental/OperationResult.cs
@@ -1,0 +1,79 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#if NETCOREAPP
+using System.Diagnostics.CodeAnalysis;
+#endif
+
+#nullable enable
+namespace Microsoft.IdentityModel.Abstractions.Experimental
+{
+    /// <summary>
+    /// Represents the result of an operation that can either succeed with a valid result or failed with a specific error.
+    /// </summary>
+    /// <typeparam name="TResult">The type of the successful result value.</typeparam>
+    /// <typeparam name="TError">The type of the error.</typeparam>
+    /// <remarks>
+    /// This struct implements a discriminated union pattern for operation results.
+    /// It ensures that either a valid result or an error is available, but never both.
+    /// This pattern helps with clear error handling and propagation.
+    /// </remarks>
+#pragma warning disable RS0016 // Add public types and members to the declared API
+    public readonly record struct OperationResult<TResult, TError>
+#pragma warning restore RS0016 // Add public types and members to the declared API
+            where TResult : class
+            where TError : OperationError
+    {
+        /// <summary>
+        /// Creates a new instance of <see cref="OperationResult{TResult, TError}"/> indicating a successful operation
+        /// and containing an object of the associated type.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


src/Microsoft.IdentityModel.Abstractions/Experimental/OperationResult.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// <typeparam name="TError">The type of the error.</typeparam>
+    /// <remarks>
+    /// This struct implements a discriminated union pattern for operation results.
+    /// It ensures that either a valid result or an error is available, but never both.
+    /// This pattern helps with clear error handling and propagation.
+    /// </remarks>
+#pragma warning disable RS0016 // Add public types and members to the declared API
+    public readonly record struct OperationResult<TResult, TError>
+#pragma warning restore RS0016 // Add public types and members to the declared API
+            where TResult : class
```
keegan-caruso, line 24: Do we need TResult to have a class constraint? Using a value type should be a valid scenario.

src/Microsoft.IdentityModel.Abstractions/Experimental/OperationResult.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// </summary>
+        /// <value>
+        /// <c>true</c> if the operation succeeded and <see cref="Result"/> contains a valid value;
+        /// <c>false</c> if the operation failed and <see cref="Error"/> contains the error.
+        /// </value>
+#if NETCOREAPP
+        [MemberNotNullWhen(true, nameof(Result))]
+        [MemberNotNullWhen(false, nameof(Error))]
+#endif
+        public readonly bool IsValid { get; }
```
keegan-caruso, line 62: Succeeded? Not sure this is the best name

src/Microsoft.IdentityModel.Abstractions/Experimental/OperationResult.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// </summary>
+    /// <typeparam name="TResult">The type of the successful result value.</typeparam>
+    /// <typeparam name="TError">The type of the error.</typeparam>
+    /// <remarks>
+    /// This struct implements a discriminated union pattern for operation results.
+    /// It ensures that either a valid result or an error is available, but never both.
+    /// This pattern helps with clear error handling and propagation.
+    /// </remarks>
+#pragma warning disable RS0016 // Add public types and members to the declared API
+    public readonly record struct OperationResult<TResult, TError>
```
keegan-caruso, line 22: I'm okay with the name

jmprieur:
Shouldn't this be in Microsoft.Identity.Abstractions?
The acquisition part would also benefit from these.

src/Microsoft.IdentityModel.Abstractions/Experimental/OperationResult.cs
brentschmaltz: @keegan-caruso a value type, such as int we cannot check for null.
We could return default, but that would be 0 for int.

brentschmaltz:
Moved to Microsoft.Identity.Abstractions

src/Microsoft.IdentityModel.Abstractions/Experimental/OperationResult.cs
keegan-caruso: @brentschmaltz - With this model, we would not need to check the TResult for null. It should be sufficient to check `IsValid`. If true, then we know that TResult is set and should be used.

PR #3281, opened 2025-07-18
Allow usage of relative URIs in AuthnContextClassRef

### LINKED ISSUE #3279: [Feature Request] Allow relaxed parsing of URIs in SAML2
I'm an AAI Admin at a german university and I'm currently in the process of bringing "BundID" into our AAI.  
BundID is a federated login system to allow login via European Issued Identity Cards, is mostly a SAML2, but has some unique quirks. One of them being "AuthnContextClassRef" not being an absolute URI.

While this works well with Keycloak and (most likely) Shibboleth IdP, we're at a loss with Microsoft.IdentityModel, since there's an active check for `IsAbsolute` in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blame/7699e533557a32fb3b74dc440c10a61c9b90df1b/src/Microsoft.IdentityModel.Tokens.Saml/Saml2/Saml2AuthenticationContext.cs#L81-L82.
From a standards view-point this might be the correct thing to do, but from a library user standpoint, this might be a little bit too strict.

So I'd like to propose a relaxation of the check by either using an configurable switch OR by removing the check alltogether - I'm aware this isn't 100% up to SAML2-Core, but it's at least interoperable with other SAML2 products.

Unfortunately my only alternatives are either: take the SAML Response, Decrypt it, validate signature by hand (that won't happen) or ditch the usage of Microsoft.IdentityModel and go down another route with e.g. Keycloak.

glatzert:
# Allow usage of relative URIs in AuthnContextClassRef

Introduces a AppContext switch, that allows to cirumvent a condition that checks for `Uri.IsAbsoulte` on `Saml2AuthenticationContext.ClassReference`

Fixes #3279 

@brentschmaltz

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
@@ -98,7 +98,9 @@ namespace Microsoft.IdentityModel.Tokens
         private static bool? _useCapitalizedXMLTypeAttr;
         internal static bool UseCapitalizedXMLTypeAttr => _useCapitalizedXMLTypeAttr ??= (AppContext.TryGetSwitch(UseCapitalizedXMLTypeAttrSwitch, out bool useCapitalizedXMLTypeAttr) && useCapitalizedXMLTypeAttr);
 
-
+        /// <summary>
+        /// When enabled, allows using relative URIs in SAML2 authentication context.
+        /// </summary>
         internal const string AllowRelativeUrisInSaml2AuthnContextSwitch = "Switch.Microsoft.IdentityModel.AllowRelativeUrisInSaml2AuthnContext";
         private static bool? _allowRelativeUrisInSaml2AuthnContext;
         internal static bool AllowRelativeUrisInSaml2AuthnContext => _allowRelativeUrisInSaml2AuthnContext ??= (AppContext.TryGetSwitch(AllowRelativeUrisInSaml2AuthnContextSwitch, out bool allowRelativeUris) && allowRelativeUris);
--- a/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
@@ -98,6 +98,12 @@ namespace Microsoft.IdentityModel.Tokens
         private static bool? _useCapitalizedXMLTypeAttr;
         internal static bool UseCapitalizedXMLTypeAttr => _useCapitalizedXMLTypeAttr ??= (AppContext.TryGetSwitch(UseCapitalizedXMLTypeAttrSwitch, out bool useCapitalizedXMLTypeAttr) && useCapitalizedXMLTypeAttr);
 
+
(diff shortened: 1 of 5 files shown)


test/Microsoft.IdentityModel.Tokens.Saml.Tests/Saml2AuthenticationContextCollectionDefinition.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
brentschmaltz, line 1: Do we need this class?

brentschmaltz: I think its ok

brentschmaltz (approved):

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -98,6 +98,12 @@ internal static class AppContextSwitches
         private static bool? _useCapitalizedXMLTypeAttr;
         internal static bool UseCapitalizedXMLTypeAttr => _useCapitalizedXMLTypeAttr ??= (AppContext.TryGetSwitch(UseCapitalizedXMLTypeAttrSwitch, out bool useCapitalizedXMLTypeAttr) && useCapitalizedXMLTypeAttr);
 
+
+        internal const string AllowRelativeUrisInSaml2AuthnContextSwitch = "Switch.Microsoft.IdentityModel.AllowRelativeUrisInSaml2AuthnContext";
```
pmaytak:
```suggestion
        /// <summary>
        /// When enabled, allows using relative URIs in SAML2 authentication context.
        /// </summary>
        internal const string AllowRelativeUrisInSaml2AuthnContextSwitch = "Switch.Microsoft.IdentityModel.AllowRelativeUrisInSaml2AuthnContext";
```

pmaytak (approved):

pmaytak:
Once this is released, we'll need to update: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/App-Context-Switches-in-IdentityModel

glatzert:
I updated the suggestions and rebased as per Githubs request - I hope I did not break anything in your workflow by doing that 🙈 

I was sick last week, so I needed a liitle bit.
Nevertheless I found one thing about the test cases - if they run in Visual Studio simultanously for each .net version, _some_ will fail - I think the xunit runner does not isolate the AppContext against each other in that case. Running the tests for a single .net version will work as expected.

pmaytak (approved):

glatzert:
Can I do something about the (probably) shared AppContext in the different dotnet versions or did i miss something in older .net runtimes?

glatzert:
After being out-of-office for quite some time now, I wanted to follow up about this change - anything that can be done from my side to make it happen?

PR #3287, opened 2025-08-05
Building with NET 10 preview

### LINKED ISSUE #3176: [Feature Request] Update to the latest .NET 10 Preview
.NET 10 can be downloaded from [here](https://dotnet.microsoft.com/en-us/download/dotnet/10.0). You will need the latest Visual Studio Preview to run the project with this version.

To-do:
- Add the latest .NET 10 Preview to targets for this project.
- Resolve any new analyzer warnings and errors.
- Confirm all tests pass.
- Update the build templates to use this .NET version.
- In some projects and build pipelines, we have `TargetNet9` conditional build variable. Rename this to `TargetNetNext` so it can be reused for subsequent updates more easily.
- Check [.NET 10 breaking changes](https://learn.microsoft.com/en-us/dotnet/core/compatibility/10.0) for anything that may affect our code.

JoshLozensky:
Wilson.sln builds, tests pass, and pack succeeds locally, running in NET 10.0 and in other versions

For local runs the TargetNetNext var needs to be set to True in the targets.props and targetsTest.props or run with:
-p:TargetNetNext=True


armandRobled:
ACTIVE

Enviado desde mi Samsung Mobile de Telcel
Obtener Outlook para Android<https://aka.ms/AAb9ysg>

build/dependencies.props
```diff
@@ -15,6 +15,13 @@
     <SystemTextJson>8.0.5</SystemTextJson>
   </PropertyGroup>
 
+<PropertyGroup Condition="'$(TargetFramework)' == 'net10.0'">
+    <!--
+      Unknown if 10.0 has the same problem as 9 where there are some runtime incompatibilities if
```
brentschmaltz: The issue is we want to use a LTS target, 9.0 is not long term, which is the reason CoreWCF asked us to stick with 8.0.

Is 10 LTS?

JoshLozensky: It is LTS, so I changed it to be 10.0

build/dependenciesTest.props
```diff
@@ -26,6 +24,9 @@
   <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))">
     <XunitRunnerVisualStudioVersion>3.0.0-pre.49</XunitRunnerVisualStudioVersion>
   </PropertyGroup>
+  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net10.0'))">
```
pmaytak: This shouldn't be needed as the previous check is true for .NET 10.

build/commonTest.props
```diff
@@ -20,7 +20,7 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.1' OR '$(TargetFramework)' == 'net6.0' Or '$(TargetFramework)' == 'net8.0' Or '$(TargetFramework)' == 'net9.0'">
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.1' OR '$(TargetFramework)' == 'net6.0' Or '$(TargetFramework)' == 'net8.0' Or '$(TargetFramework)' == 'net9.0' Or '$(TargetFramework)' == 'net10.0'">
```
pmaytak:
```suggestion
  <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
```

.github/workflows/dotnetcore.yml
```diff
@@ -39,12 +36,17 @@ jobs:
       with:
         global-json-file: ./global.json
 
+    - name: Setup latest .NET 10 preview SDK
+      uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: '10.0.x-preview'
```
keegan-caruso:
```suggestion
        dotnet-version: '10.0.x'
        dotnet-quality: 'preview'
```

See: https://github.com/actions/setup-dotnet

build/commonTest.props
keegan-caruso: Agree on the target framework identifier, but this should be a property group

build/template-Build-run-tests-sign.yml
```diff
@@ -33,6 +33,12 @@ steps:
     version: 8.x
     installationPath: $(Agent.ToolsDirectory)/dotnet
 
+- task: UseDotNet@2
```
keegan-caruso, line 36: Use version 4 and need to include preview quality

src/Microsoft.IdentityModel.LoggingExtensions/Microsoft.IdentityModel.LoggingExtensions.csproj
```diff
@@ -9,6 +9,7 @@
     <PackageId>Microsoft.IdentityModel.LoggingExtensions</PackageId>
     <PackageTags>.NET;Windows;Authentication;Identity;Extensions;Logging</PackageTags>
     <TargetFrameworks>netstandard2.0;net9.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(TargetNetNext)'== 'True'">$(TargetFrameworks);</TargetFrameworks>
```
keegan-caruso: Was this supposed to add net10?

JoshLozensky: I don't think so because this should still be getting Net10 in the $(TargetFrameworks) set in the targets.props file

.github/workflows/dotnetcore.yml
```diff
@@ -17,9 +17,6 @@ on:
     branches:
       - dev
 
-env:
```
jennyf19: we don't need this anymore?

.github/workflows/dotnetcore.yml
jennyf19: In the past, we've had to specifically define the version of the preview

build/commonTest.props
JoshLozensky: I just removed NET_CORE entirely and use NETCOREAPP in the code instead

build/dependenciesTest.props
JoshLozensky: removed

jennyf19:
Exit criteria would be: 
- Test: With `TargetNetNext=True`, does `dotnet build` succeed everywhere, and do test projects run for `net10.0`?
- release builds with .net10.0 and without.
- confirm that the net10.0 API files are validated in CI when the opt-in is active

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectMessageTests.cs
```diff
@@ -288,9 +288,9 @@ public void OidcCreateAuthenticationRequestUrl(string testId, OpenIdConnectMessa
 #elif NET9_0
             if (!message.SkuTelemetryValue.Equals("ID_NET9_0"))
                 context.Diffs.Add($"{message.SkuTelemetryValue} != ID_NET9_0");
-#elif NET_CORE
-            if (!message.SkuTelemetryValue.Equals("ID_NETSTANDARD2_0"))
-                context.Diffs.Add($"{message.SkuTelemetryValue} != ID_NETSTANDARD2_0");
+#elif NET10_0
```
jennyf19, line 291: do we need the netstandard2_0?

src/Microsoft.IdentityModel.LoggingExtensions/Microsoft.IdentityModel.LoggingExtensions.csproj
keegan-caruso: I think the line directly above will override that value

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectMessageTests.cs
```diff
@@ -560,7 +560,10 @@ public void OidcCreateLogoutRequestUrl(string testId, OpenIdConnectMessage messa
 #elif NET9_0
             if (!message.SkuTelemetryValue.Equals("ID_NET9_0"))
                 context.Diffs.Add($"{message.SkuTelemetryValue} != ID_NET9_0");
-#elif NET_CORE
+#elif NET10_0
+            if (!message.SkuTelemetryValue.Equals("ID_NET10_0"))
+                context.Diffs.Add($"{message.SkuTelemetryValue} != ID_NET10_0");
+#elif NETCOREAPP
```
jennyf19: the change from `NET_CORE → NETCOREAPP` ...if net10.0 is only conditionally added, will these still be exercised/tested

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
```diff
@@ -632,12 +627,7 @@ public static SecurityKey DefaultRsaSecurityKey2
...
                 var cert = CertificateHelper.LoadX509Certificate(certData);
-                AsymmetricAlgorithm publicKey;
-#if NET462 || NET472 || NET_CORE
-                publicKey = RSACertificateExtensions.GetRSAPublicKey(cert);
-#else
-                publicKey = cert.PublicKey.Key;
-#endif
+                AsymmetricAlgorithm publicKey = RSACertificateExtensions.GetRSAPublicKey(cert);
```
jennyf19, line 630: why was this removed?

src/Microsoft.IdentityModel.LoggingExtensions/Microsoft.IdentityModel.LoggingExtensions.csproj
JoshLozensky: I'm seeing a net10.0 folder in the bin that has dlls in it, also the build fails without the code below added in dependencies.props as further evidence that it is built in net10.

`<PropertyGroup Condition="'$(TargetFramework)' == 'net10.0'">
      <MicrosoftExtensionsLoggingAbstractionsVersion>10.0.0-preview.*</MicrosoftExtensionsLoggingAbstractionsVersion>
</PropertyGroup>`

.github/workflows/dotnetcore.yml
JoshLozensky: at the link Keegan included it says preview versions are included by default and it does get installed... ¯\_(ツ)_/¯

test/Microsoft.IdentityModel.TestUtils/KeyingMaterial.cs
JoshLozensky: the if was inclusive of all versions we support so we weren't using the else, therefore we didn't need to pre-define publicKey for scope

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectMessageTests.cs
JoshLozensky: no I removed it

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/OpenIdConnectMessageTests.cs
JoshLozensky: The only part missed would be the NET10_0 condition if NET 10 isn't run the rest should still work the same. Also, removed the NETCOREAPP case as all versions we run the test in are already explicitly defined.

.github/workflows/dotnetcore.yml
JoshLozensky: hmmm, I thought this was something I added when I removed it.... guess not, I'll put it back

src/Microsoft.IdentityModel.LoggingExtensions/Microsoft.IdentityModel.LoggingExtensions.csproj
JoshLozensky: added the net 10 as indicated

pmaytak (approved):

build/targets.props
```diff
@@ -1,7 +1,12 @@
...
+  <PropertyGroup>
+    <!-- This must be set to True to build with NET 10.0 locally or set it in a CLI arg-->
+    <TargetNetNext>False</TargetNetNext>
+  </PropertyGroup>
+  
   <PropertyGroup>
     <SrcTargets Condition="'$(LocalBuild)' != 'True'">net462;net472;netstandard2.0;net6.0;net8.0;net9.0</SrcTargets>
     <SrcTargets Condition="'$(LocalBuild)' == 'True'">netstandard2.0;net8.0;net9.0</SrcTargets>
-    <SrcTargets Condition="'$(TargetNetNext)'== 'True'">$(SrcTargets);</SrcTargets>
+    <SrcTargets Condition="'$(TargetNetNext)'== 'True'">$(SrcTargets);net10.0</SrcTargets>
```
mdchennu, line 10: Why can't the net10.0 be added to line 9? and then we leave line 10 unchanged? (asking so i know/not as feedback!)

test/System.IdentityModel.Tokens.Jwt.Tests/CreateAndValidateTokens.cs
```diff
@@ -611,8 +611,6 @@ public static TheoryData<JwtTheoryData> RoundTripTokensTheoryData()
                 ValidationParameters = Default.SymmetricEncryptSignTokenValidationParameters
             });
 
-#if NET462 || NET_CORE
-            // RsaPss is not supported on .NET < 4.6
             var rsaPssSigningCredentials = new SigningCredentials(Default.AsymmetricSigningKey, SecurityAlgorithms.RsaSsaPssSha256);
```
mdchennu, line 614: How can this if be removed? What happens with  .Net472 here?

build/targets.props
JoshLozensky: Since NET 10 is still in preview, we only want to build with it by choice, if we put it on line 9 anyone who tries to build the repo locally without first installing the NET 10 preview SDK would not be able to build it.

test/System.IdentityModel.Tokens.Jwt.Tests/CreateAndValidateTokens.cs
JoshLozensky: Good Q you are correct that NET 472 is not included currently and this will include it, this is intentional. The reason stated in the notes for this if condition is that RsaPass isn't available below NET 4.6 but we don't support versions below NET 4.6 any longer like we did when this was written so there is no need for it. So this code should have been updated to include 472 but never was. 

Additionally, since this is a test file there is no potential for direct customer impact and the change is tested :).

test/Microsoft.IdentityModel.Xml.Tests/EnvelopedSignatureWriterTests.cs
```diff
@@ -112,7 +112,7 @@ public static TheoryData<EnvelopedSignatureTheoryData> CreateSignatureWithoutSpe
                 theoryData.Add(CreateSignatureTestCase(KeyingMaterial.RsaSecurityKey_2048, SecurityAlgorithms.RsaSha512, null, SecurityAlgorithms.RsaSha512 + "_DigestNULL"));
                 theoryData.Add(CreateSignatureTestCase(KeyingMaterial.RsaSecurityKey_2048, SecurityAlgorithms.RsaSha512Signature, null, SecurityAlgorithms.RsaSha512Signature + "_DigestNULL"));
 
-#if NET_CORE
+#if NETCOREAPP
```
jmprieur: Really. can you elaborate please @JoshLozensky ?

JoshLozensky: In this [earlier review comment](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3287#discussion_r2255929104) it was suggested that we set our custom NET_CORE property to '$(TargetFrameworkIdentifier)' == '.NETCoreApp' It seemed more clear to just use NETCOREAPP directly rather than making a property that equals it. Is there a reason we shouldn't do this @jmprieur?

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestE2ETests.cs
```diff
@@ -137,7 +137,7 @@ public static TheoryData<RoundtripSignedHttpRequestTheoryData> RoundtripTheoryDa
                 tvpWrongIssuerSigningKey.IssuerSigningKey = KeyingMaterial.RsaSecurityKey2;
                 var ecdsaSigningCredentials = new SigningCredentials(KeyingMaterial.Ecdsa256Key, SecurityAlgorithms.EcdsaSha256);
 
-#if NET_CORE
+#if NETCOREAPP
```
keegan-caruso: I know we typically use this, but the more modern symbol is just `NET`, which is roughly targets NET 5.0 or greater

keegan-caruso: Should be an easy find and replace 😄

global.json
```diff
@@ -1,7 +1,8 @@
 {
     "sdk": {
```
keegan-caruso, line 2: Doesn't this mean we would always build with NET 10 SDK? Or 11 when it is in early preview next year? That doesn't seem right. We should keep rollForward as latestFeature, then use the yaml tasks to add NET 10 SDK when the variable is set.

build/commonTest.props
```diff
@@ -52,8 +56,13 @@
...
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net462' or '$(TargetFramework)' == 'net472'">
-    <!-- Suprpess the "Microsoft.Extensions.TimeProvider.Testing 9.0.0 doesn't support net462 and has not been tested with it" warning -->
+    <!-- Suppress the "Microsoft.Extensions.TimeProvider.Testing 9.0.0 doesn't support net462 and has not been tested with it" warning -->
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
   </PropertyGroup>
-  
+
+  <PropertyGroup Condition ="'$(TargetFramework)' == 'net6.0'">
+    <!-- Suppress the "Microsoft.NET.Test.Sdk doesn't support net6.0 and has not been tested with it" error-->
```
keegan-caruso: Can you help me understand this? We didn't increase the version of the Microsoft.NET.Test.Sdk package in this change did we?

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestE2ETests.cs
JoshLozensky: replaced

global.json
JoshLozensky: Only if the NET 10 SDK is installed, which should only happen in the build pipeline if we set TargetNetNext to True. If we don't set it this way it will refuse to use the NET 10 SDK even when present.

build/commonTest.props
JoshLozensky: I think it is a transitive ref somewhere

JoshLozensky: I tried doing so at one point and changed it back, forgot to remove this. Took it out

build/dependenciesTest.props
```diff
@@ -23,14 +23,25 @@
...
   <!-- MicrosoftExtensionsTimeProviderTesting 8.x has a 6.0.0 target, 9.x does not. -->
   <PropertyGroup Condition="'$(TargetFramework)' != 'net6.0'">
     <MicrosoftExtensionsTimeProviderTestingVersion>9.0.0</MicrosoftExtensionsTimeProviderTestingVersion>
   </PropertyGroup>
+  
   <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
     <MicrosoftExtensionsTimeProviderTestingVersion>8.10.0</MicrosoftExtensionsTimeProviderTestingVersion>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net10.0'">
```
pmaytak, line 44: Why is this added? System.Text,Json dependency comes within the .NET framework.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestCreationTests.cs
```diff
@@ -910,7 +910,7 @@ public static TheoryData<CreateSignedHttpRequestTheoryData> CreateCnfClaimTheory
                         SigningCredentials = new SigningCredentials(KeyingMaterial.X509SecurityKey1, SecurityAlgorithms.RsaSha256, SecurityAlgorithms.Sha256),
                         ExpectedClaimValue = $@"{{""cnf"":{{""{ConfirmationClaimTypes.Jwk}"":{{""{JsonWebKeyParameterNames.Kid}"":""{Base64UrlEncoder.Encode(rsaJwkFromX509Key.ComputeJwkThumbprint())}"",""{JsonWebKeyParameterNames.E}"":""{rsaJwkFromX509Key.E}"",""{JsonWebKeyParameterNames.Kty}"":""{JsonWebAlgorithmsKeyTypes.RSA}"",""{JsonWebKeyParameterNames.N}"":""{rsaJwkFromX509Key.N}""}}}}}}"
                     },
-#if NET472 || NET_CORE
+#if NET472 || NETCOREAPP
```
pmaytak: nit? And in other places
```suggestion
#if NET472 || NET
```

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
```diff
@@ -1356,7 +1356,7 @@ public static TheoryData<JWEDecompressionTheoryData> JweDecompressSizeTheoryData
 
             TheoryData<JWEDecompressionTheoryData> theoryData = new TheoryData<JWEDecompressionTheoryData>();
 
-#if NETCOREAPP2_1
+#if NET2_1
```
pmaytak: Is this a valid flag? I don't see it here: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation

build/dependenciesTest.props
JoshLozensky: There were some transitive deps on it from other packages that triggered CVE alerts. This explicitly declares a secure version.

test/System.IdentityModel.Tokens.Jwt.Tests/JwtSecurityTokenHandlerTests.cs
JoshLozensky: fair, changed it back.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestCreationTests.cs
JoshLozensky: done

build/dependenciesTest.props
JoshLozensky: It is defined above as 8.0.5 for addressing a CVE, so I'm changing it back to using the NET 10 version here

pmaytak (approved):

mdchennu (approved):

korekuztomz-web (commented):
/

PR #3332, opened 2025-10-02
Align pipeline build with test phase by passing TargetNetNext to build & pack (fixes missing net10.0 test dll)

alexholub113:
[Pipelines - Run 20250930.2](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1539960&view=results)

The pipeline was failing during the VSTest phase with: The test source file ...\net10.0\Microsoft.IdentityModel.Logging.Tests.dll was not found. Root cause: runTests.ps1 (or related props) enables the next .NET target (net10.0) via TargetNetNext, causing test discovery to look for net10.0 test assemblies. The earlier solution build step in template-Build-run-tests-sign.yml did not pass /p:TargetNetNext, so net10.0 binaries were never built. VSTest then failed when attempting to execute the non-existent net10.0 test DLL.

### DIFF
--- a/build/template-Build-run-tests-sign.yml
+++ b/build/template-Build-run-tests-sign.yml
@@ -1,9 +1,9 @@
 # template-Build-run-tests-sign.yml
 
 parameters:
-  - name: TargetNetNext
-    type: string
-    default: 'True'
+    - name: TargetNetNext
+      type: string
+      default: "True"
 
 steps:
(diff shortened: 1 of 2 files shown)


keegan-caruso:
@alexholub113 - Can you pls rebase

keegan-caruso:
closing in favor of #3363 , I pass the variable via an env var.

PR #3334, opened 2025-10-06
Enable Long Paths in GitHub Actions workflow

keegan-caruso:
Added a step to enable long paths in Windows.


keegan-caruso:
Appears to not be needed

PR #3335, opened 2025-10-06
Don't target NET 10 SDK for dotnetcore.yml

keegan-caruso:
# Don't target NET 10 SDK for dotnetcore.yml

## Description

Currently targeting net10 preview is breaking the build


jmprieur (approved):
LGTM

kllysng (approved):

PR #3373, opened 2025-11-26
Fix DateTime claim value type recognition for ISO8601 strings with and without trailing zeros

jaywang-cn:
## Background
Currently, the method `GetStringClaimValueType` returns `ClaimValueTypes.DateTime` only if the input string matches exactly the output of `DateTime.ToUniversalTime().ToString("o", ...)`. 
This leads to inconsistent behavior for ISO8601 UTC strings with or without trailing zeros, e.g.:
- "2025-11-26T07:18:45.1234560Z" → DateTime
- "2025-11-26T07:18:45.123456Z"  → String

## Fix
With this change, all strings that can be successfully parsed as a valid DateTime will return `ClaimValueTypes.DateTime`, regardless of trailing zeroes.

## Impact
This fix ensures that semantically equivalent ISO8601 UTC time strings are consistently recognized as DateTime claims.

## Tests
- Added/Updated tests for ISO8601 datetime strings with and without trailing zeros.


jaywang-cn:
@microsoft-github-policy-service agree company="WiseTech Global"

jaywang-cn:
@microsoft-github-policy-service agree company="WiseTech Global"

jaywang-cn:
@pmaytak

PR #3374, opened 2025-12-01
Reduce memory consumption by avoiding adding items when queue is full

keegan-caruso:
Based on #3239

Do not add items if cache is full or compacting.
Under heavy load, stress tests showed that the LRUCache was unable to keep up a change was made to avoid adding to the cache if we detect it was full or compacting.

A change to start TaskEventQ on startup instead of starting and stopping improves performance and simplifies the code.
We were not checking is a SignatureProvider was actually added to the cache. This is important as if the SignatureProvider is not added, the cache will not be able to properly apply the dispose logic and kernel objects could pile up.


pmaytak:
@copilot Generate summary of the changes and separate the source and test changes.

brentschmaltz:
Nice to see this one get a time slice :-)

bgavrilMS (approved):

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -998,276 +995,7 @@ public async Task ReferenceCountingTest_MultiThreaded()
             TestUtilities.AssertFailIfErrors(context);
         }
 
-        /// <summary>
```
bgavrilMS, line 1001: Is there enough coverage without these tests? Maybe work item should have some acceptance test to give some confidence.

bgavrilMS (commented):
Not sure about test coverage

brentschmaltz:
@bgavrilMS there were some perf / stress tests that exposed this issue.

keegan-caruso:
Closing, suggested path would be to use #3376

PR #3376, opened 2025-12-04
Variation on 3374

keegan-caruso:
See #3374 for context

This PR instead keeps the existing implementation and by a feature flag uses the new strategy.


src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -98,6 +95,15 @@ internal static class AppContextSwitches
...
         internal static bool UseCapitalizedXMLTypeAttr => _useCapitalizedXMLTypeAttr ??= (AppContext.TryGetSwitch(UseCapitalizedXMLTypeAttrSwitch, out bool useCapitalizedXMLTypeAttr) && useCapitalizedXMLTypeAttr);
 
+
+        /// <summary>
+        /// Enabling this switch will cause the crypto provider to use the caching behavior prior to 8.16.0
+        /// </summary>
+        internal const string UseLegacyCryptoProviderCache = "Switch.Microsoft.IdentityModel.UseLegacyCryptoProviderCache";
+        private static bool? _useLegacyCryptoProviderCache;
+
+        internal static bool UseLegacyCryptoProviderCacheFlag => _useLegacyCryptoProviderCache ??= (AppContext.TryGetSwitch(UseLegacyCryptoProviderCache, out bool useLegacy) && useLegacy);
```
GeoK, line 105: Perhaps we should make this feature opt-in rather than opt-out until sufficiently tested.

src/Microsoft.IdentityModel.Tokens/EventBasedLRUCache.cs
```diff
@@ -163,21 +204,33 @@ internal EventBasedLRUCache(
...
         private void AddActionToEventQueue(Action action)
         {
+            Debug.Assert(!UseMinimalCache);
             _eventQueue.Enqueue(action);
 
             // start the event queue task if it is not running
             StartEventQueueTaskIfNotRunning();
         }
 
+        private bool UseMinimalCache => !AppContextSwitches.UseLegacyCryptoProviderCacheFlag && !_maintainLRU;
```
GeoK, line 224: This could be easy to miss going forward. I don't see any code paths setting _maintainLRU to true. Perhaps it would be cleaner to remove LRU related codepaths and depend only on the flag.

brentschmaltz: @GeoK i would agree with you, the LRU part was never used and is internal.
We found that the LRU part really slowed things down, i mean really slowed things down.
Not to mention complicating the code.

keegan-caruso: This sounds like a better solution, thanks for the suggestion.

keegan-caruso: Correct on _maintainLRU, it is only used by internals/friends.

src/Microsoft.IdentityModel.Tokens/EventBasedCache.cs
```diff
@@ -0,0 +1,340 @@
...
+                if (LogHelper.IsEnabled(EventLogLevel.Warning))
+                    LogHelper.LogWarning(LogHelper.FormatInvariant(LogMessages.IDX10900, ex));
+            }
+        }
+
+        /// <summary>
+        /// Remove all expired cache items from the _map ONLY.
+        /// The enumerator returned from the dictionary is safe to use concurrently with reads and writes to the dictionary, according to the MS document.
+        /// </summary>
+        /// <returns>Number of items removed.</returns>
```
keegan-caruso, line 143:
```suggestion
```

keegan-caruso:
I will not be working on this. This will be picked up by someone else on the team.

PR #3410, opened 2026-02-04
Add telemetry around signature validation

iNinja:
This pull request introduces telemetry tracking for JWT and SAML/2 signature validation in the `JsonWebTokenHandler` and related classes. It refactors several methods to non-static to support instance-level telemetry. The changes enable detailed telemetry reporting on signature validation outcomes, such as missing keys, unsupported algorithms, and verification failures.

**Telemetry integration for JWT and SAML signature validation:**

* Added telemetry reporting to all major signature validation paths in `JsonWebTokenHandler`, `SamlSecurityTokenHandler`, and `Saml2SecurityTokenHandler`, capturing events like signing key not found, algorithm not supported, signature provider creation failure, and signature verification failure or success. This is accomplished by calling the new `RecordSignatureValidationTelemetry` helper at appropriate points.
* Refactored `ValidateSignature`, `ValidateSignatureUsingAllKeys`, and `ValidateSignatureWithKey` methods from static to instance methods, allowing access to the instance telemetry client (`_telemetryClient`).
* Updated the signature validation logic to pass the telemetry client through the call stack, ensuring telemetry is recorded for all signature validation attempts, including when using all keys or a specific key.

These changes lay the groundwork for comprehensive telemetry on signature validation, improving observability and diagnostics for token validation scenarios.

# New Telemetry Counter
## Signature Validation Counter
The counter tracks **5 dimensions** (tags):

| Tag Name | Tag Key | Purpose | Value Source |
|----------|---------|---------|--------------|
| **Library Version** | `IdentityModelVersion` | Library version tracking | `IdentityModelTelemetryUtil.ClientVer` |
| **Algorithm** | `Algorithm` | Signature algorithm from token header | Token `alg` claim |
| **Key Algorithm** | `KeyAlgorithm` | Key type and size identifier | `CryptoTelemetry.GetKeyAlgorithmId(SecurityKey)` |
| **Issuer** | `Issuer` | Token issuer (allowlisted hosts only) | `CryptoTelemetry.GetTrackedIssuerOrOther(issuer)` |
| **Error** | `Error` | Error type or success indicator | `TelemetryConstants.SignatureValidationErrors.*` |

#### 1. IdentityModelVersion
- **Cardinality**: Low (~5-10 active versions)
- **Values**: Semantic versions (e.g., `"7.3.1"`, `"8.0.0"`)
- **Typical Production**: 1-3 versions during rollouts
- **Lifecycle**: Reduces over time as deployments converge

#### 2. Algorithm
- **Cardinality**: ~12-15
- **Possible Values** (from JWT/SAML specs):

| Algorithm Family | Algorithms | Count |
|-----------------|------------|-------|
| **RSA (PKCS#1 v1.5)** | `RS256`, `RS384`, `RS512` | 3 |
| **RSA-PSS** | `PS256`, `PS384`, `PS512` | 3 |
| **ECDSA** | `ES256`, `ES384`, `ES512`, `ES256K` | 4 |
| **HMAC** | `HS256`, `HS384`, `HS512` | 3 |
| **Legacy/Rare** | `none`, custom algorithms | ~2-5 |

- **Typical Production**: 2-5 algorithms (commonly `RS256`, `ES256`, `PS256`)
- **Bounded**: JWT/JOSE specs define a finite set of algorithms

#### 3. KeyAlgorithm
- **Cardinality**: ~11-13
- **Implementation**: `CryptoTelemetry.GetKeyAlgorithmId(SecurityKey)` returns predefined constants
- **Possible Values**:

| Key Type | Key Algorithm IDs | Description |
|----------|------------------|-------------|
| **RSA** | `RSA-2048`, `RSA-3072`, `RSA-4096`, `RSA-UNKNOWN` | RSA public/private keys |
| **ECDSA** | `ECDSA-P256`, `ECDSA-P384`, `ECDSA-P521`, `ECDSA-UNKNOWN` | Elliptic curve keys |
| **Symmetric** | `SYM-128`, `SYM-192`, `SYM-256`, `SYM-384`, `SYM-512`, `SYM-UNKNOWN` | HMAC shared secrets |
| **Special** | `NO-KEY`, `UNKNOWN` | Missing key or unsupported key type |

- **Typical Production**: 3-6 key types (commonly `RSA-2048`, `RSA-4096`, `ECDSA-P256`)
- **Bounded**: Fixed set of industry-standard key sizes

#### 4. Issuer
- **Cardinality**: Low (~5-20 tracked hosts)
- **Implementation**: `CryptoTelemetry.GetTrackedIssuerOrOther(issuer)` with allowlist-based filtering
- **Behavior**:
  - Extracts host from issuer URI (e.g., `https://login.microsoftonline.com/tenant/` → `login.microsoftonline.com`)
  - Returns host if in `CryptoTelemetry.TrackedIssuers` allowlist
  - Returns `"other"` for all non-allowlisted issuers
- **Typical Values**:
  - `login.microsoftonline.com` (Microsoft Entra ID)
  - `accounts.google.com` (Google)
  - `appleid.apple.com` (Apple)
  - `other` (catch-all for non-tracked issuers)
- **Cardinality Control**: Allowlist prevents unbounded growth from arbitrary issuers
- **Default**: Empty allowlist (all issuers reported as `"other"`)

#### 5. Error
- **Cardinality**: 6 fixed values
- **Implementation**: `TelemetryConstants.SignatureValidationErrors.*`
- **Possible Values**:

| Error Type | Constant | Meaning |
|-----------|----------|---------|
| **Success** | `None` | Signature validation succeeded |
| **Verification Failed** | `SignatureVerificationFailed` | Signature invalid (key present, crypto works, but signature doesn't match) |
| **Algorithm Not Supported** | `AlgorithmNotSupported` | Algorithm not supported by key or crypto provider |
| **Provider Creation Failed** | `SignatureProviderCreationFailed` | Crypto provider could not create signature provider |
| **Signing Key Not Found** | `SigningKeyNotFound` | No signing key was found or resolved |
| **Other** | `Other` | Other errors not covered by specific categories |

- **Typical Production**: 2-4 error types observed (commonly `None`, `SignatureVerificationFailed`, `SigningKeyNotFound`)
- **Bounded**: Fixed set of error constants to prevent cardinality explosion

### Total Cardinality Calculation

```
Total Combinations = IdentityModelVersion × Algorithm × KeyAlgorithm × Issuer × Error
                   = 5 × 15 × 13 × 20 × 6
                   = 117,000 theoretical maximum (with full issuer allowlist)
```

**Production Reality (Empty Issuer Allowlist - Default)**:
```
Active Versions × Active Algorithms × Active Key Types × Issuer ("other" only) × Active Errors
= 2 × 4 × 4 × 1 × 3
= 96 typical active time series
```

**Production Reality (With 5 Tracked Issuers)**:
```
Active Versions × Active Algorithms × Active Key Types × (Tracked Issuers + "other") × Active Errors
= 2 × 4 × 4 × 6 × 3
= 576 typical active time series
```

**Upper Bound Estimate**: 
- **Default (no issuer tracking)**: ~150-250 time series
- **With issuer tracking (5-10 issuers)**: ~1,000-2,000 time series

**Note**: The issuer dimension is strictly controlled via the `CryptoTelemetry.TrackedIssuers` allowlist, preventing unbounded cardinality growth.

### Cardinality Assessment

✅ **Low-Medium Cardinality** - Safe for production at scale

**Rationale**:
1. ✅ Issuer dimension is strictly allowlist-controlled (default: all issuers → `"other"`)
2. ✅ Error dimension is a fixed enumeration (6 values)
3. ✅ Algorithm set is finite and standardized
4. ✅ Key sizes are industry-standard values (not arbitrary)
5. ✅ Library versions naturally consolidate over time

# Benchmarks
## For loop 100_000 calls + 10 OperationsPerInvoke
| Method                                                              | Mean     | Error   | StdDev  | Median   | P90      | P95      | P100     | Ratio | RatioSD | Gen0      | Allocated | Alloc Ratio |
|---------------------------------------------------------------------|----------|---------|---------|----------|----------|----------|----------|-------|---------|-----------|-----------|-------------|
| JsonWebTokenHandler_ValidateTokenAsync_TelemetryDisabled             | 180.9 ms | 1.81 ms | 3.77 ms | 179.5 ms | 185.9 ms | 188.3 ms | 193.1 ms | 1.00  | 0.03    | 4300.0000 | 38.68 MB  | 1.00        |
| JsonWebTokenHandler_ValidateTokenAsync_TelemetryEnabledNoTracking    | 179.5 ms | 4.20 ms | 9.13 ms | 175.2 ms | 190.0 ms | 195.0 ms | 225.0 ms | 0.99  | 0.05    | 4300.0000 | 38.68 MB  | 1.00        |
| JsonWebTokenHandler_ValidateTokenAsync_TelemetryEnabledWithTracking  | 178.2 ms | 2.31 ms | 5.12 ms | 176.4 ms | 185.9 ms | 188.2 ms | 191.7 ms | 0.99  | 0.03    | 4300.0000 | 38.68 MB  | 1.00        |

## No for loop, 100_000 OperationsPerInvoke
| Method                                                              | Mean      | Error     | StdDev    | P90       | P95      | P100     | Ratio | RatioSD | Allocated | Alloc Ratio |
|--------------------------------------------------------------------|----------:|----------:|----------:|----------:|---------:|---------:|------:|--------:|----------:|------------:|
| JsonWebTokenHandler_ValidateTokenAsync_TelemetryDisabled            | 0.8680 ns | 0.0362 ns | 0.0779 ns | 0.9705 ns | 1.000 ns | 1.064 ns |  1.01 |    0.13 |         - |          NA |
| JsonWebTokenHandler_ValidateTokenAsync_TelemetryEnabledNoTracking   | 1.0351 ns | 0.1411 ns | 0.2882 ns | 1.3120 ns | 1.621 ns | 2.207 ns |  1.20 |    0.35 |         - |          NA |
| JsonWebTokenHandler_ValidateTokenAsync_TelemetryEnabledWithTracking | 1.0730 ns | 0.1187 ns | 0.2556 ns | 1.4330 ns | 1.605 ns | 1.770 ns |  1.25 |    0.31 |         - |          NA |


src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1850,9 +1929,10 @@ internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JwtSecurityToken jwtT
                         var unwrappedKey = kwp.UnwrapKey(Base64UrlEncoder.DecodeBytes(jwtToken.RawEncryptedKey));
                         unwrappedKeys.Add(new SymmetricSecurityKey(unwrappedKey));
                     }
-                    else
-#endif
+                    else if (key.CryptoProviderFactory.IsSupportedAlgorithm(jwtToken.Header.Alg, key))
+#else
                     if (key.CryptoProviderFactory.IsSupportedAlgorithm(jwtToken.Header.Alg, key))
+#endif
```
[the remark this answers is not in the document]
iNinja: The behaviour remains the same. The only difference is when the conditional is true, it no longer causes an indentation issue due to the `if` on the `else` branch being on the same level.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
```diff
@@ -0,0 +1,321 @@
...
+using System.Buffers;
+#endif
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Telemetry;
+
+/// <summary>
+/// Provides utility methods for telemetry and analytics purposes.
+/// </summary>
+public static class CryptoTelemetry
```
GeoK, line 15: Can we have CryptoTelemetry as an internal class?

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
```diff
@@ -0,0 +1,321 @@
...
+    /// <para>Issuer hosts not in this list will be reported as "other" to prevent cardinality explosion in telemetry systems.</para>
+    /// <para>When set to null or empty, all issuers will be reported as "other".</para>
+    /// <para>Example: new[] { "login.microsoftonline.com", "accounts.google.com" }</para>
+    /// </remarks>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Performance", "CA1819:Properties should not return arrays", Justification = "Simple configuration API for setting allowlist")]
+    public static string[] TrackedIssuers
+    {
+        get
+        {
+            lock (_trackedIssuersLock)
```
GeoK: Lock on every read seems excessive for this (read-heavy) scenario. ConcurrentDictionary provides lock free reads.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
```diff
@@ -0,0 +1,321 @@
...
+            return OtherIssuersLabel;
+
+        // Check if this issuer is in the tracked list
+        lock (_trackedIssuersLock)
+        {
+            return _trackedIssuers.Contains(host) ? host : OtherIssuersLabel;
+        }
+    }
+
+    private static string ExtractHostFromIssuerCore(string issuer)
```
GeoK: Seems like a relatively heavy operation to have on the hot path, particularly for multi-tenant applications. Cache could fill-up quickly and we'd run the same op per each validation.
As correctness is not imperative, would a naive op like Contains string help achieve the same outcome without the performance penalty?

tehran19r (approved):

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
iNinja: After discussing, I left the class public with one method to configure the telemetry and made everything else internal/private.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
iNinja: Replaced the lock with a `volatile` array.
Given that the array is static and only replaced in full this gives us thread safety and the worst case scenario upon a race condition is that a thread reads an outdated list of whitelisted issuers. The issuers are stored on a local variable before using so there's no risk of iterating after changing.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
iNinja: The operation looks like more than what it actually is, given that we are working with spans for the most part.
The cache filling up in multi-tenant applications is a very real scenario. It could be resolved by keeping the cache disabled.

iNinja: I have removed the caching for now and will revisit with a proper LRU cache in the future.

benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncTests.cs
```diff
@@ -193,4 +194,95 @@ public async Task<List<Claim>> JsonWebTokenHandler_ValidateTokenAsyncWithVP_Crea
             return claims.ToList();
         }
     }
+
+    // ===== Telemetry Impact Benchmarks =====
+
+    public enum TelemetryConfiguration
```
pmaytak: nit: not needed, remove?

benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncTests.cs
```diff
@@ -193,4 +194,95 @@ public async Task<List<Claim>> JsonWebTokenHandler_ValidateTokenAsyncWithVP_Crea
...
+        }
+
+        [IterationSetup(Target = nameof(JsonWebTokenHandler_ValidateTokenAsync_TelemetryDisabled))]
+        public void Setup_TelemetryDisabled()
+        {
+            CryptoTelemetry.EnableSignatureValidationTelemetry(false, null);
+        }
+
+        [BenchmarkCategory("ValidateTokenAsync_TelemetryImpact"), Benchmark(Baseline = true)]
+        public async Task<TokenValidationResult> JsonWebTokenHandler_ValidateTokenAsync_TelemetryDisabled()
```
pmaytak, line 239: Can also set [OperationsPerInvoke](https://github.com/dotnet/performance/blob/main/docs/microbenchmark-design-guidelines.md#operationsperinvoke) to IterationCount which will normalize the final result.

benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncTests.cs
```diff
@@ -193,4 +194,95 @@ public async Task<List<Claim>> JsonWebTokenHandler_ValidateTokenAsyncWithVP_Crea
...
+    public enum TelemetryConfiguration
+    {
+        Disabled,
+        EnabledNoTracking,
+        EnabledWithTracking,
+        EnabledWithCaching
+    }
+
+    [GroupBenchmarksBy(BenchmarkLogicalGroupRule.ByCategory)]
+    public class ValidateTokenAsyncTests_TelemetryImpact
```
pmaytak, line 204: nit: add a small comment what tracking and caching means/does.

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -229,6 +236,12 @@ private static ValidationResult<SecurityKey, ValidationError> ValidateSignatureW
             CryptoProviderFactory cryptoProviderFactory = validationParameters.CryptoProviderFactory ?? key.CryptoProviderFactory;
             if (!cryptoProviderFactory.IsSupportedAlgorithm(jsonWebToken.Alg, key))
             {
+                RecordSignatureValidationTelemetry(
```
GeoK, line 239: This seems inconsistent with SAML case where alg not supported is not reported.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
```diff
@@ -0,0 +1,287 @@
...
+    /// <returns>The issuer host if in the allowlist, otherwise "other".</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static string GetTrackedIssuerOrOther(string issuer)
+    {
+        string[] trackedIssuers = TrackedIssuers;
+
+        if (trackedIssuers.Length == 0 || string.IsNullOrWhiteSpace(issuer))
+            return OtherIssuersLabel;
+
+        string host = ExtractHostFromIssuer(issuer);
```
GeoK: I still think we could simplify issuer matching, improve perf and reduce allocations on the hot path.

Idea would be to replace host extraction with a simple `Contains` check, something like:

   ```csharp
   foreach (var tracked in trackedIssuers)
   {
       if (issuer.Contains(tracked, StringComparison.Ordinal))
           return tracked;
   }
   return OtherIssuersLabel;
   ```

  Example: tracked = "login.microsoftonline.com" matches issuer "https://login.microsoftonline.com/tenant-id/v2.0"

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
GeoK: added another comment on the same topic here: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3410#discussion_r2789904368

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1238,6 +1238,23 @@ public override string WriteToken(SecurityToken token)
...
+            string errorType,
+            SecurityToken securityToken,
+            string algorithm,
+            SecurityKey key)
+        {
+            if (CryptoTelemetry.RecordSignatureValidationTelemetry)
+            {
+                telemetryClient.IncrementSignatureValidationCounter(
+                    errorType,
+                    securityToken.Issuer,
```
GeoK, line 1252: Any chance for nullref ex here? Could securityToken be null in any case at this point?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -1247,21 +1264,76 @@ public override string WriteToken(SecurityToken token)
...
+                    throw LogHelper.LogExceptionMessage(
+                        new InvalidOperationException(
+                            LogHelper.FormatInvariant(TokenLogMessages.IDX10636, LogHelper.MarkAsNonPII(key == null ? "Null" : key.KeyId), LogHelper.MarkAsNonPII(algorithm))));
+                }
+            }
+            catch (Exception)
+            {
+                RecordSignatureValidationTelemetry(
+                    telemetryClient,
+                    TelemetryConstants.SignatureValidationErrors.SignatureProviderCreationFailed,
```
GeoK, line 1306: nit: empty line

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
iNinja: I am running benchmarks on this before making a change.

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
iNinja: None. All validation paths fail before reaching the signature validation if the token is null.

benchmark/Microsoft.IdentityModel.Benchmarks/ValidateTokenAsyncTests.cs
iNinja: I wasn't aware of that. I'll update to avoid the for loop, thanks.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
```diff
@@ -0,0 +1,287 @@
...
+
+namespace Microsoft.IdentityModel.Telemetry;
+
+/// <summary>
+/// Provides utility methods for telemetry and analytics purposes.
+/// </summary>
+public static class CryptoTelemetry
+{
+    internal static bool RecordSignatureValidationTelemetry { get; set; }
+    private static volatile string[] _trackedIssuersArray = Array.Empty<string>();
```
pmaytak, line 18: Any performance issues with making this volatile?

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
```diff
@@ -0,0 +1,287 @@
...
+
+        if (string.IsNullOrWhiteSpace(host))
+            return OtherIssuersLabel;
+
+        return trackedIssuers.Contains(host, StringComparer.OrdinalIgnoreCase)
+            ? host
+            : OtherIssuersLabel;
+    }
+
+    internal static string ExtractHostFromIssuer(string issuer)
```
pmaytak: What about parsing into a Uri class and getting the host from there?

src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs
```diff
@@ -37,7 +37,7 @@ public class JwtSecurityTokenHandler : SecurityTokenHandler
         private static string _shortClaimType = _namespace + "/ShortTypeName";
         private bool _mapInboundClaims = DefaultMapInboundClaims;
 
-        internal Microsoft.IdentityModel.Telemetry.ITelemetryClient TelemetryClient = new TelemetryClient();
+        internal Microsoft.IdentityModel.Telemetry.ITelemetryClient _telemetryClient = new TelemetryClient();
```
pmaytak: Should stay `TelemetryClient` because it's internal, not private.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
```diff
@@ -0,0 +1,287 @@
...
+                    .ToArray();
+            }
+        }
+    }
+
+#if NET8_0_OR_GREATER
+    private static readonly SearchValues<char> _hostDelimiters = SearchValues.Create(['/', '?', ':']);
+#endif
+
+    private static class KeyAlgorithmIds
```
pmaytak, line 61: Can https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/8ffd5b0493c07af0a5ced6410690b83d08c4e1a3/src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs be reused?

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
```diff
@@ -0,0 +1,287 @@
...
+
+    /// <summary>
+    /// Gets the issuer host for telemetry, returning "other" if not in the tracked issuers allowlist.
+    /// </summary>
+    /// <param name="issuer">The full issuer URI.</param>
+    /// <returns>The issuer host if in the allowlist, otherwise "other".</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static string GetTrackedIssuerOrOther(string issuer)
+    {
+        string[] trackedIssuers = TrackedIssuers;
```
pmaytak: Why copy over the variable?
Can TrackedIssuers be a set?

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
```diff
@@ -0,0 +1,287 @@
...
+                384 => KeyAlgorithmIds.Symmetric384,
+                512 => KeyAlgorithmIds.Symmetric512,
+                _ => KeyAlgorithmIds.SymmetricUnknown
+            },
+            _ => KeyAlgorithmIds.Unknown
+        };
+    }
+
+    /// <summary>
+    /// Gets the issuer host for telemetry, returning "other" if not in the tracked issuers allowlist.
```
pmaytak, line 173: How many tracked issuers do we expect? What is the purpose, why not log all issuers?

test/Microsoft.IdentityModel.Tokens.Tests/Telemetry/CryptoTelemetryTests.cs
```diff
@@ -0,0 +1,423 @@
...
+        CryptoTelemetry.EnableSignatureValidationTelemetry(false, null);
+
+        // Act
+        var result = CryptoTelemetry.ExtractHostFromIssuer(issuer);
+
+        // Assert
+        Assert.Equal(expectedHost, result);
+    }
+
+    public static TheoryData<string, string> ExtractHostFromIssuerTestData => new()
```
pmaytak: Missing (or worth adding) cases
1) null input (contract-defining)
Right now you only test "", not null.

null → expected: "" / null / exception (whatever the method’s intended behavior is)
2) Whitespace-only and trimming
You don’t test leading/trailing whitespace.

"   "
" https://example.com "
If the method trims, assert it; if it doesn’t, this may reveal a bug.
3) Upper/lower case normalization
Hosts are case-insensitive; many implementations normalize to lowercase.

"HTTPS://EXAMPLE.COM/Path" → "example.com" (or "EXAMPLE.COM" if you preserve case)
4) Fragments (#...)
You test query strings but not fragments.

"https://example.com/path#frag" → "example.com"
5) User info in authority (user:pass@host)
This can break naive parsers.

"https://user:pass@example.com/path" → "example.com"
6) IPv4 / IPv6 hosts
Common omission; also stresses parsing around brackets.

"https://127.0.0.1/path" → "127.0.0.1"
"https://[::1]/" → "::1" (or "[::1]" depending on implementation)
7) Trailing dot / unusual-but-valid host forms
"https://example.com./" → "example.com." or "example.com" (define expected)
8) Non-HTTP schemes (if allowed/should be rejected)
You currently only cover http/https and schemeless.

"ftp://example.com/resource" → expected?
"file:///C:/temp" (no host) → "" expected?
If the method is only meant for web issuers, these should assert "" (or exception), to prevent accidental acceptance.

9) Clearly malformed authority cases beyond the ones you have
You have "https://" and "://example.com", but not other malformed shapes:

"https:///path-only" (triple slash, empty host) → ""
"http://:8080" (missing host, has port) → ""
"http://?q=1" (no host) → ""
10) Issuer-specific variants (likely relevant for this repo)
You have login.microsoftonline.com/tenant... and /common/..., but Azure AD issuers commonly show up as:

"https://sts.windows.net/{tenantid}/" → "sts.windows.net"
Sovereign clouds:
"https://login.microsoftonline.us/{tenant}/v2.0"
"https://login.partner.microsoftonline.cn/{tenant}/v2.0" (Only if the method is intended to be generic; still good regression coverage.)

src/Microsoft.IdentityModel.Tokens/Telemetry/ITelemetryClient.cs
```diff
@@ -35,6 +36,19 @@ internal void LogBackgroundConfigurationRefreshFailure(
...
 
+        /// <summary>
+        /// Increments the signature validation counter with algorithm and key size details.
+        /// </summary>
+        /// <param name="errorType">The error type constant from <see cref="TelemetryConstants.SignatureValidationErrors"/>. Use <see cref="TelemetryConstants.SignatureValidationErrors.None"/> for successful validations.</param>
+        /// <param name="issuer">The token issuer.</param>
+        /// <param name="algorithm">The signature algorithm used (e.g., RS256, ES256, HS256).</param>
+        /// <param name="key">The security key used for signature validation.</param>
+        internal void IncrementSignatureValidationCounter(
+            string errorType,
```
pmaytak, line 47: ErrorType/exception is usually last for consistency.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/SamlSecurityTokenHandler.cs
```diff
@@ -30,6 +31,8 @@ public partial class SamlSecurityTokenHandler : SecurityTokenHandler
         private IEqualityComparer<SamlSubject> _samlSubjectEqualityComparer = new SamlSubjectEqualityComparer();
         private SamlSerializer _serializer = new SamlSerializer();
 
+        internal Telemetry.ITelemetryClient _telemetryClient = new Telemetry.TelemetryClient();
```
pmaytak:
```suggestion
        internal Telemetry.ITelemetryClient TelemetryClient = new Telemetry.TelemetryClient();
```

test/Microsoft.IdentityModel.Tokens.Tests/Telemetry/CryptoTelemetryTests.cs
```diff
@@ -0,0 +1,423 @@
...
+        // Act
+        var result = CryptoTelemetry.GetKeyAlgorithmId(key);
+
+        // Assert
+        Assert.Equal(expectedId, result);
+    }
+#endif
+
+    [Fact]
+    public void GetKeyAlgorithmId_X509SecurityKey_Rsa2048_ReturnsCorrectId()
```
pmaytak: Can these be combined into inline data?

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
iNinja: Addressing privately, but the purpose of not tracking every issuer is to keep telemetry cardinality known and avoid exploding it.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
iNinja: Originally, it was meant to handle the situation where the volatile variable is written to between the check for count and the iteration.
After moving to the simpler tracked issuer comparison, it is not needed. I'll remove it.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
iNinja: Not in this scenario. Here we are combining the data we track per key type to keep it as a single field for telemetry.

test/Microsoft.IdentityModel.Tokens.Tests/Telemetry/CryptoTelemetryTests.cs
iNinja: Updating.

src/Microsoft.IdentityModel.Tokens/Telemetry/ITelemetryClient.cs
iNinja: In this scenario, to avoid using two tags for `success` and `errorType`, we are using error type `None` for successes.
That was the main reason to bring it forward.

test/Microsoft.IdentityModel.Tokens.Tests/Telemetry/CryptoTelemetryTests.cs
iNinja: This is great. I've updated the tests based on the updated implementation for the host matching which is now simpler and doesn't look into a lot of these things.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
iNinja: That incurs allocations so we went with the current approach.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
iNinja: No, this is more performant than using a read/write lock, and the use case is not affected by the potential race conditions which would lead to reading stale data if it is modified between the two reads.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
iNinja: Updated to use this approach.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
```diff
@@ -182,106 +174,17 @@ private static string GetJsonWebKeyAlgorithmId(JsonWebKey jwk)
...
-            // Path found - use it as host end (query would come after path if present).
-            hostEnd = pathStart;
-        }
-        else
+        foreach (string trackedHost in TrackedIssuers)
         {
-            // No path - check for query parameter.
-            int queryStart = issuerSpan.IndexOf('?');
-            hostEnd = queryStart >= 0 ? queryStart : issuerSpan.Length;
+            if (issuer.IndexOf(trackedHost, StringComparison.OrdinalIgnoreCase) >= 0)
```
GeoK: nit: consider Ordinal for perf. We shouldn't expect authorities to change _iss_ casing.
_iss_ is case sensitive https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1

GeoK (approved):

pmaytak (approved):

PR #3423, opened 2026-02-16
Emit warning for package version mismatch

cpp11nullptr:
## Description

Add build task which detects whether Microsoft.IdentityModel.* or System.IdentityModel.* packages, which were resolved for a given project, have version mismatch. This warning aims to reduce runtime failures caused by version skew.

It's opt-out task which can be disabled by following project setting:

```xml
<PropertyGroup>
  <DisableIdentityModelVersionMismatchCheck>true</DisableIdentityModelVersionMismatchCheck>
</PropertyGroup>
```

## Validation

Validation was done by building a project which introduces version mismatch (`9.0.0` there is example of released version with that build task):

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="8.0.2" />
    <PackageReference Include="Microsoft.IdentityModel.Tokens" Version="9.0.0" />
  </ItemGroup>
</Project>
```

Build output:

```cmd
> dotnet build
   
Restore complete (0.6s)
  IdentityModelVersionMismatch net8.0 succeeded with 1 warning(s) (1.3s) → bin\Debug\net8.0\IdentityModelVersionMismatch.dll
    C:\Users\username\.nuget\packages\microsoft.identitymodel.tokens\9.0.0\buildTransitive\Microsoft.IdentityModel.Tokens.targets(35,5): warning IDMODEL001: 

      All Microsoft.IdentityModel.* and System.IdentityModel.* packages must have the same version to avoid hard-to-diagnose runtime errors.
      The following packages were resolved with different versions:

        - Microsoft.IdentityModel.Abstractions 9.0.0
        - Microsoft.IdentityModel.JsonWebTokens 7.1.2 <-- upgrade to 9.0.0
        - Microsoft.IdentityModel.Logging 9.0.0
        - Microsoft.IdentityModel.Protocols 7.1.2 <-- upgrade to 9.0.0
        - Microsoft.IdentityModel.Protocols.OpenIdConnect 7.1.2 <-- upgrade to 9.0.0
        - Microsoft.IdentityModel.Tokens 9.0.0
        - System.IdentityModel.Tokens.Jwt 7.1.2 <-- upgrade to 9.0.0

      To fix this, add explicit PackageReference entries for the packages with lower versions to upgrade them to 9.0.0:

        <ItemGroup>
          <PackageReference Include="Microsoft.IdentityModel.JsonWebTokens" Version="9.0.0" />
          <PackageReference Include="Microsoft.IdentityModel.Protocols" Version="9.0.0" />
          <PackageReference Include="Microsoft.IdentityModel.Protocols.OpenIdConnect" Version="9.0.0" />
          <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="9.0.0" />
        </ItemGroup>
```

Actual resolved versions:

```
> dotnet list package --include-transitive

Restore complete (0.6s)
                                                                                                                                                                             
Build succeeded in 0.7s
Project 'IdentityModelVersionMismatch' has the following package references
   [net8.0]: 
   Top-level Package                                    Requested                 Resolved
   > Microsoft.AspNetCore.Authentication.JwtBearer      8.0.2                     8.0.2
   > Microsoft.IdentityModel.Tokens                     9.0.0   9.0.0

   Transitive Package                                           Resolved
   > Microsoft.Extensions.DependencyInjection.Abstractions      8.0.0
   > Microsoft.Extensions.Logging.Abstractions                  8.0.0
   > Microsoft.IdentityModel.Abstractions                       9.0.0
   > Microsoft.IdentityModel.JsonWebTokens                      7.1.2
   > Microsoft.IdentityModel.Logging                            9.0.0
   > Microsoft.IdentityModel.Protocols                          7.1.2
   > Microsoft.IdentityModel.Protocols.OpenIdConnect            7.1.2
   > System.IdentityModel.Tokens.Jwt                            7.1.2
```

Applying proposed package references resolves the warning.

### DIFF
--- a/src/Common/Microsoft.IdentityModel.VersionMismatch.targets
+++ b/src/Common/Microsoft.IdentityModel.VersionMismatch.targets
@@ -1,0 +1,57 @@
+<!--
+  Microsoft.IdentityModel Version Mismatch Detection
+
+  This MSBuild targets file is shipped inside every Microsoft.IdentityModel.* and
+  System.IdentityModel.* NuGet package under the buildTransitive/ folder. It runs
+  after NuGet package resolution and warns when the resolved package versions are
+  not all the same — a common source of hard-to-diagnose runtime failures.
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!--
+    Allow consumers to opt out of the check by setting
+    DisableIdentityModelVersionMismatchCheck=true in their project file.
+  -->
+  <PropertyGroup Condition="'$(_IdentityModelVersionCheckImported)' == ''">
+    <_IdentityModelVersionCheckImported>true</_IdentityModelVersionCheckImported>
+    <DisableIdentityModelVersionMismatchCheck Condition="'$(DisableIdentityModelVersionMismatchCheck)' == ''">false</DisableIdentityModelVersionMismatchCheck>
+  </PropertyGroup>
+
+  <!--
+    The target hooks into ResolvePackageAssets. There is used @(RuntimeCopyLocalItems)
+    where each item carries NuGetPackageId and NuGetPackageVersion metadata,
+    which allow detecting all resolved IdentityModel packages and their versions.
+    The guard condition '_IdentityModelVersionCheckExecuted' prevents running more
+    than once when multiple IdentityModel packages are referenced.
+  -->
+  <Target Name="_CheckIdentityModelVersions"
+          AfterTargets="ResolvePackageAssets"
+          Condition="'$(DisableIdentityModelVersionMismatchCheck)' != 'true' and '$(DesignTimeBuild)' != 'true' and '$(_IdentityModelVersionCheckExecuted)' != 'true'">
+
+    <PropertyGroup>
+      <_IdentityModelVersionCheckExecuted>true</_IdentityModelVersionCheckExecuted>
+    </PropertyGroup>
+
+    <_IdentityModelVersionMismatchTask ResolvedPackages="@(RuntimeCopyLocalItems)" />
+  </Target>
+
+  <!--
+    Inline task: the C# implementation is kept in a separate file for easier
+    maintenance and testing. RoslynCodeTaskFactory compiles it on first use.
+  -->
+  <UsingTask TaskName="_IdentityModelVersionMismatchTask"
+             TaskFactory="RoslynCodeTaskFactory"
+             AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.Core.dll">
+    <ParameterGroup>
+      <ResolvedPackages ParameterType="Microsoft.Build.Framework.ITaskItem[]" Required="true" />
+    </ParameterGroup>
+    <Task>
+      <Using Namespace="System" />
+      <Using Namespace="System.Collections.Generic" />
+      <Using Namespace="System.Linq" />
+      <Code Type="Fragment" Language="cs"
+            Source="$(MSBuildThisFileDirectory)Microsoft.IdentityModel.VersionMismatchTask.cs" />
+    </Task>
+  </UsingTask>
+
+</Project>
--- a/src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
+++ b/src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
@@ -5,14 +5,6 @@
 // MSBuild inline task (<Code Type="Fragment">). It inspects the resolved NuGet
 // package graph for Microsoft.IdentityModel.* / System.IdentityModel.* packages
 // and emits a build warning when multiple different versions are detected.
-//
-// Because this is a code fragment, the code below is placed directly inside the
-// Execute() method body. Early "return true" statements are valid and indicate
-// successful task completion (no warning needed). At the end of a fragment,
-// MSBuild auto-appends "return !Log.HasLoggedErrors;" so no explicit final
-// return is required.
-//
-// See: https://learn.microsoft.com/visualstudio/msbuild/msbuild-roslyncodetaskfactory
 
 // Explicit list of Microsoft.IdentityModel / System.IdentityModel packages
 // shipped from this repository (matching the project directories under src/).
--- a/src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
+++ b/src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
@@ -150,4 +150,4 @@ sb.AppendLine("Note: Disabling this check is not recommended. Mismatched package
 sb.AppendLine("runtime errors such as TypeLoadException, MissingMethodException, or other hard-to-diagnose failures.");
 sb.AppendLine();
 
-Log.LogWarning(null, "IDMODEL001", null, null, 0, 0, 0, 0, sb.ToString());
+Log.LogWarning(null, "IDX00001", null, null, 0, 0, 0, 0, sb.ToString());
--- a/src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
+++ b/src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
@@ -2,14 +2,41 @@
 // Licensed under the MIT License.
 
 // This file is compiled at build-time by RoslynCodeTaskFactory and runs as an
-// MSBuild inline task. It inspects the resolved NuGet package graph for
-// Microsoft.IdentityModel.* / System.IdentityModel.* packages and emits a
-// build warning when multiple different versions are detected.
+// MSBuild inline task (<Code Type="Fragment">). It inspects the resolved NuGet
+// package graph for Microsoft.IdentityModel.* / System.IdentityModel.* packages
+// and emits a build warning when multiple different versions are detected.
+//
+// Because this is a code fragment, the code below is placed directly inside the
+// Execute() method body. Early "return true" statements are valid and indicate
+// successful task completion (no warning needed). At the end of a fragment,
+// MSBuild auto-appends "return !Log.HasLoggedErrors;" so no explicit final
+// return is required.
+//
+// See: https://learn.microsoft.com/visualstudio/msbuild/msbuild-roslyncodetaskfactory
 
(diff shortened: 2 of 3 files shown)


src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
```diff
@@ -0,0 +1,126 @@
...
+
+foreach (var pkgId in packagesToUpgrade)
+{
+    sb.AppendLine("    <PackageReference Include=\"" + pkgId + "\" Version=\"" + maxVersion + "\" />");
+}
+
+sb.AppendLine("  </ItemGroup>");
+sb.AppendLine();
+
+Log.LogWarning(null, "IDMODEL001", null, null, 0, 0, 0, 0, sb.ToString());
```
GeoK: Is a return statement (true) required for code fragments?

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
```diff
@@ -0,0 +1,126 @@
...
+sb.AppendLine();
+sb.AppendLine("  <ItemGroup>");
+
+foreach (var pkgId in packagesToUpgrade)
+{
+    sb.AppendLine("    <PackageReference Include=\"" + pkgId + "\" Version=\"" + maxVersion + "\" />");
+}
+
+sb.AppendLine("  </ItemGroup>");
+sb.AppendLine();
```
GeoK, line 142: It seems like appending a line would result in multiple lines prefixed with the same error code, for example:

 `warning IDMODEL001: All Microsoft.IdentityModel.* and System.IdentityModel.* packages must have the same version to avoid hard-to-diagnose runtime errors.`
`warning IDMODEL001: warning IDMODEL001: The following packages were resolved with different versions:`

Was that indented behavior or having a single message broken into multiple lines?

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
GeoK: Thoughts on changing this to an error?

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
```diff
@@ -0,0 +1,126 @@
...
+    sb.AppendLine("  - " + pkg.Key + " " + pkg.Value + marker);
+
+    if (!StringComparer.OrdinalIgnoreCase.Equals(pkg.Value, maxVersion))
+    {
+        packagesToUpgrade.Add(pkg.Key);
+    }
+}
+
+sb.AppendLine();
+sb.Append("To fix this, add explicit PackageReference entries for the packages with lower versions to upgrade them to " + maxVersion + ":");
```
GeoK, line 131: Consider documenting an explicit switch that would disable this check and consequences of doing so.

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
```diff
@@ -0,0 +1,126 @@
...
+    {
+        var packageId = item.GetMetadata("NuGetPackageId");
+        var version = item.GetMetadata("NuGetPackageVersion");
+
+        if (string.IsNullOrEmpty(packageId) || string.IsNullOrEmpty(version))
+        {
+            continue;
+        }
+
+        if (packageId.StartsWith("Microsoft.IdentityModel.", StringComparison.OrdinalIgnoreCase) ||
```
GeoK: Consider excluding some packages or creating a map of existing packages. One example that could mess up the logic is ADAL (
Microsoft.IdentityModel.Clients.ActiveDirectory). It is deprecated but it is still being used.

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
cpp11nullptr: The intention is to have a single warning message (one per build regardless how many Microsoft.IdentityModel packages are consumed by a project), as multiline string for a better readability.

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
cpp11nullptr: Good point, done.

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
cpp11nullptr: Added, thanks.

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
cpp11nullptr: I'd keep it as warning for some time to prevent blocking if something went wrong. The recommended practice is to track warnings as errors anyway.

cpp11nullptr: No, it's implicitly set to success at the end so can be omitted.

GeoK (approved):

pmaytak:
> <Project Sdk="Microsoft.NET.Sdk">
>   <PropertyGroup>
>     <OutputType>Exe</OutputType>
>     <TargetFramework>net8.0</TargetFramework>
>   </PropertyGroup>
>   <ItemGroup>
>     <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="8.0.2" />
>     <PackageReference Include="Microsoft.IdentityModel.Tokens" Version="9.0.0" />
>   </ItemGroup>
> </Project>

The reverse scenario is also covered? (Referencing a lower version explicitly, ex. M.IM.Tokens 7.1.0, and a higher version transitive dependencies).

src/Common/Microsoft.IdentityModel.VersionMismatch.targets
```diff
@@ -0,0 +1,57 @@
...
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!--
+    Allow consumers to opt out of the check by setting
+    DisableIdentityModelVersionMismatchCheck=true in their project file.
+  -->
+  <PropertyGroup Condition="'$(_IdentityModelVersionCheckImported)' == ''">
+    <_IdentityModelVersionCheckImported>true</_IdentityModelVersionCheckImported>
+    <DisableIdentityModelVersionMismatchCheck Condition="'$(DisableIdentityModelVersionMismatchCheck)' == ''">false</DisableIdentityModelVersionMismatchCheck>
```
pmaytak, line 17: Just thinking, since teams have treat warnings as errors enabled often, this would be a breaking change. In IdentityModel 9.x, we could set this to false by default; but in 8.x, disable by default? Mismatched versions don't always cause issues; so this is an edge case.

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
```diff
@@ -0,0 +1,153 @@
...
+sb.AppendLine();
+sb.AppendLine("  <PropertyGroup>");
+sb.AppendLine("    <DisableIdentityModelVersionMismatchCheck>true</DisableIdentityModelVersionMismatchCheck>");
+sb.AppendLine("  </PropertyGroup>");
+sb.AppendLine();
+sb.AppendLine("Note: Disabling this check is not recommended. Mismatched package versions can cause unexpected");
+sb.AppendLine("runtime errors such as TypeLoadException, MissingMethodException, or other hard-to-diagnose failures.");
+sb.AppendLine();
+
+Log.LogWarning(null, "IDMODEL001", null, null, 0, 0, 0, 0, sb.ToString());
```
pmaytak: IdentityModel uses "IDX" prefix. Can we use it for consistency?

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
```diff
@@ -0,0 +1,153 @@
...
+
+// Determine the highest version for the upgrade suggestion.
+var maxVersion = identityModelPackages
+    .Select(item => item.Value)
+    .OrderByDescending(item =>
+    {
+        var version = parseVersion(item);
+        return version;
+    }, Comparer<Tuple<System.Version, bool, string>>.Default)
+    .First();
```
pmaytak, line 102: Nit: Would something like this work:

```csharp
var maxVersion = identityModelPackages
    .Select(p => p.Value)
    .MaxBy(v => parseVersion(v), Comparer<Tuple<Version, bool, string>>.Default);
```

cpp11nullptr:
> The reverse scenario is also covered? (Referencing a lower version explicitly, ex. M.IM.Tokens 7.1.0, and a higher version transitive dependencies).

Yes, it detects a version mismatch in both direct and transitive dependencies with suggesting to use the highest resolved version as target one.

src/Common/Microsoft.IdentityModel.VersionMismatch.targets
cpp11nullptr: I'd still advocate to provide a visibility of version mismatch via opt-out warning - introduced runtime issues may be frustrating, especially if it occurs in edge cases long after production deployment, so it's better to catch it earlier.

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
cpp11nullptr: It seems that [.MaxBy()](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.maxby) is supported only from .NET 6, so I'll stick with `.OrderByDescending()` for a better supportability.

src/Common/Microsoft.IdentityModel.VersionMismatchTask.cs
cpp11nullptr: Changes to use IDX prefix.

bgavrilMS (approved):

pmaytak (approved):

PR #3435, opened 2026-03-18
[IdentityModel 8.x] Downgrade MicrosoftExtensionsLoggingAbstractionsVersion to 8.0.0 on .NET 10

pmaytak:
Some users are unable to reference Microsoft.Extensions.* 10x.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -22,7 +22,10 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net10.0'">
-    <MicrosoftExtensionsLoggingAbstractionsVersion>10.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
+    <!--
+      Keep at 8.x for compatibility for upstream users
+    -->
+    <MicrosoftExtensionsLoggingAbstractionsVersion>8.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net9.0'">


build/dependencies.props
```diff
@@ -22,7 +22,10 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net10.0'">
-    <MicrosoftExtensionsLoggingAbstractionsVersion>10.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
+    <!--
```
bgavrilMS, line 25: nit:

Move this property in the PropertyGroup at line 3 and delete this block as well as the net9 block at line 31 and the net8 block at line 40. Also add a comment like:

bgavrilMS (approved):

cherrianathomas17-art (approved):

build/dependencies.props
pmaytak: I'll keep this PR to minimal changes and we can reorganize in a separate one.

mdchennu (approved):

pvlst:
Can you explain the downgrade? When somebody's project is targeting .NET 10, what would be the serious reason for referencing .NET 8 library?

bgavrilMS:
> Can you explain the downgrade? When somebody's project is targeting .NET 10, what would be the serious reason for referencing .NET 8 library?

It's not a .NET 8 library. It's version 8 of the Microsoft.Extension.* libraries.  This stratety was suggested by the .NET team: starting with version 8, there is ensured backwards compatibility. So essentially the version 8 of Microosft.Extensions.* acts like an interface - it declares the minimal contract between this SDK and Microsoft.Extensions (ME), while allowing apps top upgrade the implementation of ME to whatever they desire. Note that .NET releases multiple versions of ME libraries, i.e. 8, 9, 10 etc. 

With the current strategy, this SDK would release: net9 with ME version 9, net10 with ME version 10 etc. This can lead to diamond dependency conflicts.

PR #3436, opened 2026-03-18
[IdentityModel 9.x] Downgrade MicrosoftExtensionsLoggingAbstractionsVersion to 8.0.0 on .NET 10

pmaytak:
Some users are unable to reference Microsoft.Extensions.* 10x.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -22,7 +22,10 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net10.0'">
-    <MicrosoftExtensionsLoggingAbstractionsVersion>10.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
+    <!--
+      Keep at 8.x for compatibility for upstream users
+    -->
+    <MicrosoftExtensionsLoggingAbstractionsVersion>8.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


bgavrilMS (approved):

build/dependencies.props
```diff
@@ -22,7 +22,10 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net10.0'">
-    <MicrosoftExtensionsLoggingAbstractionsVersion>10.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
+    <!--
+      Keep at 8.x for compatibility for upstream users
+    -->
+    <MicrosoftExtensionsLoggingAbstractionsVersion>8.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
```
bgavrilMS, line 28: nit: 

Move this property in the PropertyGroup at line 3 and delete this block as well as the net9 block at line 31 and the net8 block at line 40. Also add a comment like: 

`<!-- Design decision: Pin Microosft.Extensions.* libraries to version 8 starting with net8, as per recommendation from .NET team. Microsoft.Extensions.* version 9, 10 etc. are expected to be fully backwards cmpatible with version 8-->`

pmaytak: I'll keep this PR to minimal changes and we can reorganize in a separate one.

mdchennu (approved):

PR #3443, opened 2026-04-10
Add DPoP (RFC 9449) proof creation and server-side validation

MZOLN:
## Summary

PBI 3549154: Implement dPoP validation primitives in Wilson
 
Adds `Microsoft.IdentityModel.Dpop` — a new package implementing DPoP (Demonstrating Proof-of-Possession) per [RFC 9449](https://www.rfc-editor.org/rfc/rfc9449.html). 

Provides both client-side proof creation and server-side proof validation with no `System.Net.Http` dependency.
 
 ## What's included
 
 **Core types:**
 - `DPoPProof` — client-side: creates signed DPoP proof JWTs with `htm`, `htu`, `iat`, `jti`, `ath`, and optional `nonce` claims
 - `DPoPProofValidator` — server-side: validates DPoP proofs per RFC 9449 §4.3 (all 12 steps), returns structured `DPoPValidationResult`
 - `DPoPValidationOptions` — configurable algorithms, lifetime, clock skew, nonce, and `IJtiReplayCache` plug-in
 - `DPoPValidationResult` — sealed result type with factory methods (`Success`, `Failed`, `NonceRequired`)
 
 **Interface:** `IJtiReplayCache` — plug-in point for jti-based replay detection
 
 ## Design decisions
 
  - **Replay protection is layered:** `iat` freshness (MUST per RFC) is always enforced; `nonce` (MAY) and `jti` tracking ("can") are opt-in via `ExpectedNonce` and `JtiReplayCache`
 - **`ath` uses ASCII encoding** with strict `EncoderFallback.ExceptionFallback` per RFC 9449 §4.2
 - **JWK thumbprint** strips `Alg`, `Kid`, `Use` per RFC 7638 §3.2 (only required members)
 - **`DPoPProofValidator` is virtual** 
  
 ## Multi-TFM support
 
 Targets `$(SrcTargets)` from `common.props`. Uses `#if` guards matching Wilson conventions:
 - `#if NET6_0_OR_GREATER` — `SHA256.HashData`, `RandomNumberGenerator.Fill`
 - `#if NET` — `CryptographicOperations.FixedTimeEquals`
 - `#if NET8_0_OR_GREATER` — `TimeProvider`
 
  - **`JsonWebKey.RepresentAsAsymmetricPublicJwkForDpop()`**: RSA branch was including `Alg`, `Kid`, `Use` in the JWK — these must be excluded per RFC 7638 §3.2 for 
correct thumbprint computation
 
 ## Tests
 
 47 tests covering all RFC §4.3 validation steps, edge cases (expired proof, wrong algorithm, symmetric key rejection, private key detection, replay cache, RSA + EC keys,
 nonce handling). Tests excluded on `net462` (`#if !NET462`) due to missing `ECCurve`/`RSA.Create(int)` APIs.
 
 ## Code quality
 
 - `PublicAPI.Shipped.txt` / `PublicAPI.Unshipped.txt` with per-TFM files (no RS0016 suppression)
 - `GlobalSuppressions.cs` with targeted member-level suppressions (CA1031, CA2227, CA1805)
 - `CancellationToken` forwarded end-to-end through `ValidateTokenAsync`
 - XML doc comments on all public members

PBI 3549154: Implement dPoP validation primitives in Wilson


src/Microsoft.IdentityModel.Dpop/Microsoft.IdentityModel.Dpop.csproj
```diff
@@ -0,0 +1,40 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <AssemblyName>Microsoft.IdentityModel.Dpop</AssemblyName>
```
bgavrilMS, line 6: Why does dPOP require a new project while SHR POP and MTLS POP did not?

src/Microsoft.IdentityModel.Dpop/DPoPConstants.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    /// <summary>
+    /// The DPoP HTTP WWW-Authenticate challenge parameter.
+    /// </summary>
+    public const string DPoPChallengeParameter = "DPoP";
+
+    /// <summary>
+    /// The DPoP token type.
+    /// </summary>
+    public const string DPoPTokenType = "DPoP";
```
bgavrilMS, line 18: nit: you can use 1 constant for DPoPHeaderName , DPoPChallengeParameter  and DPoPTokenType . These shold all be the same, irrespective of protocol.

test/Microsoft.IdentityModel.Dpop.Tests/DpopTests.cs
```diff
@@ -0,0 +1,233 @@
...
+#endif
+        }
+
+        const string KeyCloakRealmAndClient = "http://localhost:8080/realms/dpop/";
+
+        /// <summary>
+        /// This test is an integration test that requires a running Keycloak server.
+        /// </summary>
+        [Fact(Skip = "integration test")]
+        public async Task KeyCloak_IntegrationTest_RequestDpopBoundAT()
```
bgavrilMS: nit: Should this be a developer app instead? We are never going to productize this integration test. Instead, EntraID tests should be used.

test/Microsoft.IdentityModel.Dpop.Tests/DpopTests.cs
```diff
@@ -0,0 +1,233 @@
...
+
+            var options = new DPoPProofOptions
+            {
+                SigningCredentials = signingCredentials,
+                ValidationParameters = new()
+            };
+
+            var dpopProof = new DPoPProof(options);
+            // provide the code from the authorization code flow.
+            string dpopToken = await GetDPoPToken(httpClient, dpopProof, code: "");
```
bgavrilMS: nit: why don't use ROPC instead?

src/Microsoft.IdentityModel.Dpop/DPoPProof.cs
```diff
@@ -0,0 +1,151 @@
...
+
+    /// <summary>
+    /// Creates a DPoP proof JWT for the specified HTTP request.
+    /// </summary>
+    /// <param name="httpMethod">The HTTP method of the request.</param>
+    /// <param name="uri">The URI of the request.</param>
+    /// <param name="accessToken">The access token to bind this proof to.</param>
+    /// <param name="serverProvidedNonce">Optional server provided nonce to include in the proof.</param>
+    /// <returns>A JWT string representing the DPoP proof.</returns>
+    public string CreateProof(string httpMethod, Uri uri, string accessToken = null, string serverProvidedNonce = null)
```
bgavrilMS: Why does this SDK provide a method for creating DPOP proofs? This is a client-only operation.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    /// <summary>
+    /// Validates that the <c>cnf.jkt</c> claim in an access token matches the DPoP proof's
+    /// public key thumbprint, using constant-time comparison to prevent timing attacks.
+    /// </summary>
+    /// <param name="accessTokenCnfJkt">The <c>jkt</c> value from the access token's <c>cnf</c> claim.</param>
+    /// <param name="dpopProofJwkThumbprint">The computed JWK SHA-256 thumbprint from the DPoP proof.</param>
+    /// <returns><see langword="true"/> if the values match; otherwise <see langword="false"/>.</returns>
+    public static bool ValidateCnfJktBinding(string accessTokenCnfJkt, string dpopProofJwkThumbprint)
```
bgavrilMS: Why is this public API?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,423 @@
...
+    {
+        if (string.IsNullOrEmpty(accessTokenCnfJkt) || string.IsNullOrEmpty(dpopProofJwkThumbprint))
+            return false;
+
+#if NET
+        return System.Security.Cryptography.CryptographicOperations.FixedTimeEquals(
+            Encoding.UTF8.GetBytes(accessTokenCnfJkt),
+            Encoding.UTF8.GetBytes(dpopProofJwkThumbprint));
+#else
+        var a = Encoding.UTF8.GetBytes(accessTokenCnfJkt);
```
bgavrilMS: Should this be a helper method? Or maybe you want to have some sort of poly-fil ?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    /// <summary>
+    /// Computes the <c>ath</c> (access token hash) per RFC 9449 §4.2:
+    /// base64url-encoded SHA-256 hash of the ASCII-encoded access token.
+    /// </summary>
+    /// <param name="accessToken">The raw access token string.</param>
+    /// <returns>The base64url-encoded SHA-256 hash.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="accessToken"/> is null.</exception>
+    public static string ComputeAccessTokenHash(string accessToken)
```
bgavrilMS: No need for this to be public

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    /// <summary>
+    /// Checks whether a <see cref="JsonWebKey"/> contains private key material.
+    /// Per RFC 9449 §4.3 step 7, the DPoP proof's JWK must be a public key only.
+    /// </summary>
+    /// <param name="jwk">The JSON Web Key to check.</param>
+    /// <returns><see langword="true"/> if private key parameters are present; otherwise <see langword="false"/>.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="jwk"/> is null.</exception>
+    public static bool ContainsPrivateKeyMaterial(JsonWebKey jwk)
```
bgavrilMS: not public

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,423 @@
...
+               !string.IsNullOrEmpty(jwk.QI);
+    }
+
+    /// <summary>
+    /// Computes the base64url-encoded SHA-256 JWK thumbprint per RFC 7638.
+    /// </summary>
+    /// <param name="jwk">The JSON Web Key.</param>
+    /// <returns>The base64url-encoded thumbprint.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="jwk"/> is null.</exception>
+    public static string ComputeJwkThumbprint(JsonWebKey jwk)
```
bgavrilMS: not public. Pls review all methods here and make private or internal. I will stop commenting on this aspect.

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+/// Options for server-side DPoP proof validation per RFC 9449 §4.3.
+/// </summary>
+public class DPoPValidationOptions
+{
+    /// <summary>
+    /// Gets or sets the set of allowed signing algorithms for DPoP proofs.
+    /// Only asymmetric algorithms are permitted; symmetric (HMAC) and "none" are always rejected.
+    /// Default: ES256, PS256, RS256.
+    /// </summary>
+    public ISet<string> ProofAllowedSigningAlgorithms { get; set; }
```
bgavrilMS: nit: naming is hard to read, maybe `AllowedSigningAlogrithmsForProof`

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// Default: null.
+    /// </summary>
+    public string ExpectedNonce { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the <c>ath</c> (access token hash) claim is required.
+    /// When true and an access token is provided, the proof must contain a valid <c>ath</c> claim.
+    /// Default: true.
+    /// </summary>
+    public bool RequireAccessTokenHash { get; set; } = true;
```
bgavrilMS: Seems like turning this to false weakens the security posture. Do we need to add this option?

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// Default: 300 (5 minutes).
+    /// </summary>
+    public int ProofMaxLifetimeInSeconds { get; set; } = 300;
+
+    /// <summary>
+    /// Gets or sets the clock skew tolerance in seconds.
+    /// Applied both for proofs that are slightly too old and for proofs issued slightly in the future.
+    /// Default: 300 (5 minutes).
+    /// </summary>
+    public int ClockSkewInSeconds { get; set; } = 300;
```
bgavrilMS, line 33: Is this used only for the proof? Maybe name it as such and add a comment on how to configure the access token clock skew.

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// </summary>
+    public IJtiReplayCache JtiReplayCache { get; set; }
+
+#if NET8_0_OR_GREATER
+    /// <summary>
+    /// Gets or sets the time provider used for timestamp validation.
+    /// Override for unit testing.
+    /// Default: <see cref="TimeProvider.System"/>.
+    /// </summary>
+    public TimeProvider TimeProvider { get; set; } = TimeProvider.System;
```
bgavrilMS, line 57: Is this added for testing purposes only? Should it be internal?

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    /// <summary>
+    /// Gets or sets the optional replay cache for <c>jti</c> (JWT ID) replay detection.
+    /// When set, each proof's jti is checked against the cache to prevent reuse.
+    /// When null, jti replay detection is skipped.
+    /// Default: null.
+    /// </summary>
+    public IJtiReplayCache JtiReplayCache { get; set; }
+
+#if NET8_0_OR_GREATER
```
bgavrilMS, line 51: There is quite a bit of .NET and .NET FWK different code. I'd think we want to run all tests on both .NET and .NET FWK; As such, it's worth having a polyfil or some abstraction for time that allows testing on both?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,423 @@
...
+    {
+        if (string.IsNullOrEmpty(accessToken) || string.IsNullOrEmpty(dpopProofJwt))
+            return false;
+
+        try
+        {
+            var handler = new JsonWebTokenHandler();
+
+            // Extract cnf.jkt from the access token payload
+            var accessJwt = handler.ReadJsonWebToken(accessToken);
```
bgavrilMS: This seems sub-optimal. We parse the access token once to validate issuer, audience etc. And you now parse it again to extract the cnf and ath claims.

src/Microsoft.IdentityModel.Dpop/DPoPProof.cs
```diff
@@ -0,0 +1,151 @@
...
+using System.Text.Json.Nodes;
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Dpop;
+
+/// <summary>
+/// This class creates and validates DPoP proof JWTs according to RFC 9449.
+/// </summary>
+public class DPoPProof
```
bgavrilMS: There is a lot of public API (dPoPProof, dPopProofOptions) that have to do exclusively with creation of proofs. But creation of dPOP proofs is a client operation only. Why do we need this here? Can we move it to the test classes only?

bgavrilMS: If there is a strong requirement for having creation helpers here, I would recommend better naming like dPoPProofCreator and dPOPProofCreatorOptions

bgavrilMS (changes requested):
Too much public API. Creation APIs for proofs are not needed .

test/Microsoft.IdentityModel.Dpop.Tests/DpopTests.cs
```diff
@@ -0,0 +1,233 @@
...
+using System.Security.Cryptography;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Dpop;
+using Microsoft.IdentityModel.Tokens;
+using Xunit;
+
+namespace Microsoft.IdentityModel.Protocols.OAuth2.Tests
+{
+    public class DPoPTests
```
bgavrilMS, line 17: None of these tests cover or show an E2E - validation of both dPOP access token and it's associated proof.

test/Microsoft.IdentityModel.Dpop.Tests/DpopTests.cs
```diff
@@ -0,0 +1,233 @@
...
+            // provide the code from the authorization code flow.
+            string dpopToken = await GetDPoPToken(httpClient, dpopProof, code: "");
+
+            var boundRequest = new HttpRequestMessage(HttpMethod.Post, "http://localhost:8080/someapi");
+            string proof = dpopProof.CreateProof(boundRequest.Method.Method, boundRequest.RequestUri, dpopToken);
+            boundRequest.Headers.Add(DPoPConstants.DPoPHeaderName, proof);
+            boundRequest.Headers.Authorization = new AuthenticationHeaderValue(DPoPConstants.DPoPTokenType, dpopToken);
+
+            // Make a request to a resource server...
+            var response = await httpClient.SendAsync(boundRequest);
```
bgavrilMS: how do you it fails?

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
MZOLN: It fits the usage -> I think it will make more sense when seen from the MISE perspective.

src/Microsoft.IdentityModel.Dpop/Microsoft.IdentityModel.Dpop.csproj
MZOLN: I personally like this distinction, I do not have a strong argument to keep it, but this sort of separation makes sense to me and keeps things clean.

src/Microsoft.IdentityModel.Dpop/DPoPProof.cs
MZOLN: I think it's a good point let me revisit if this could be in the sample instead.

If it's needed we can always bring it back into the product in the future.

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
MZOLN: Let me look more into this, the thinking was to expose .net frameworks constructs regarding time operations as platform supports them, in case a server is using them.

src/Microsoft.IdentityModel.Dpop/Microsoft.IdentityModel.Dpop.csproj
MZOLN: Microsoft.IdentityModel.Protocols.SignedHttpRequest would be a prior art. I'd leave it separated. I will resolve it for now, please revive if you see fit.

src/Microsoft.IdentityModel.Dpop/Microsoft.IdentityModel.Dpop.csproj
```diff
@@ -0,0 +1,40 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Include="..\..\README.md">
+      <Pack>True</Pack>
+      <PackagePath>\</PackagePath>
+    </None>
+  </ItemGroup>
+
+  <ItemGroup>
```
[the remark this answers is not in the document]
MZOLN: The root PublicAPI.Shipped.txt and PublicAPI.Unshipped.txt files are automatically included by Microsoft.CodeAnalysis.PublicApiAnalyzers 3.3+ (we use 4.14.0) without
needing explicit <AdditionalFiles> entries — this is documented behavior. 

The root PublicAPI.Unshipped.txt contains the full common API surface (67 symbols). The per-TFM
files added via <AdditionalFiles Include="PublicAPI/$(TargetFramework)/..." /> serve as deltas — they only contain TFM-specific additions (e.g., TimeProvider properties
on net8.0+, empty on netstandard2.0). 

This is the same pattern used by Microsoft.IdentityModel.Tokens and Microsoft.IdentityModel.Protocols. The build is clean with zero RS0016 warnings across all TFMs, confirming the analyzer resolves both root + per-TFM files correctly.

src/Microsoft.IdentityModel.Dpop/DPoPConstants.cs
bgavrilMS: Discussed offline. Intent is clearer with different constants.

src/Microsoft.IdentityModel.Dpop/DPoPProof.cs
MZOLN: Refactored out. It seems better even though it creates duplications.

src/Microsoft.IdentityModel.Dpop/DPopProofOptions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Dpop;
+
+/// <summary>
+/// Contains options for creating and validating DPoP proofs according to RFC 9449.
+/// </summary>
+public class DPoPProofOptions
```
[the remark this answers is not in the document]
MZOLN: It's fine it is matching the existing pattern

src/Microsoft.IdentityModel.Dpop/DPopProofOptions.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// </summary>
+    public SigningCredentials SigningCredentials { get; set; }
+
+    /// <summary>
+    /// Gets or sets the token validation parameters for validating DPoP proofs.
+    /// When null, default validation parameters are used.
+    /// </summary>
+    public TokenValidationParameters ValidationParameters { get; set; }
+
+    /// <summary>
```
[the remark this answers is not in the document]
MZOLN: It's used in MISE

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
MZOLN: Refactored from MISE side. It looks much cleaner now, thank you

test/Microsoft.IdentityModel.Dpop.Tests/DpopTests.cs
MZOLN: Correct. Those tests are in MISE.

src/Microsoft.IdentityModel.Dpop/DPoPConstants.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    /// <summary>
+    /// The required <c>typ</c> header value for DPoP proof JWTs.
+    /// </summary>
+    public const string DPoPProofTokenType = "dpop+jwt";
+
+    /// <summary>
+    /// The DPoP nonce HTTP header name.
+    /// </summary>
+    public const string DPoPNonceHeaderName = "DPoP-Nonce";
```
robezirg, line 28: are these case sensitive? curious why some are all lowercase and some aren't

src/Microsoft.IdentityModel.Dpop/DPoPConstants.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    /// <summary>
+    /// The key identifier for the DPoP proof confirmation claim in the access token.
+    /// </summary>
+    public const string ConfirmationClaimDPoPKeyId = "jkt";
+
+    /// <summary>
+    /// The DPoP confirmation claim name.
+    /// </summary>
+    public const string ConfirmationClaimType = "cnf";
```
robezirg: shouldn't this be either in claim types or an existing claim type?

src/Microsoft.IdentityModel.Dpop/DPoPErrorCodes.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.IdentityModel.Dpop;
+
+/// <summary>
+/// RFC 9449 error codes for DPoP validation failures.
+/// </summary>
+public static class DPoPErrorCodes
+{
+    /// <summary>The DPoP proof is invalid or malformed.</summary>
+    public const string InvalidDPoPProof = "invalid_dpop_proof";
```
robezirg: these are all rfc? nothing we made up?

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <summary>
+    /// Gets or sets the time provider used for timestamp validation.
+    /// Override for unit testing.
+    /// Default: <see cref="TimeProvider.System"/>.
+    /// </summary>
+    public TimeProvider TimeProvider { get; set; } = TimeProvider.System;
```
robezirg, line 57: do we want to allow the end user to do this?

test/Microsoft.IdentityModel.Dpop.Tests/DpopTests.cs
MZOLN: Added some tests here.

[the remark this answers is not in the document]
MZOLN: Fixed ty

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
MZOLN: TimeProvider is a public .NET 8 platform API (System.TimeProvider) specifically designed for time abstraction and testability. It's the standard pattern across the .NET ecosystem - HttpClient, DistributedCache, IHostedService all expose TimeProvider on their options. 

Making it internal would be inconsistent with both the .NET platform convention and Wilson's own TokenValidationParameters which exposes similar time-related configuration. It's not a testing-only concern — consumers may legitimately need
 it for clock synchronization in distributed systems.

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
MZOLN: The #if conditional compilation pattern follows existing Wilson conventions throughout the codebase (see Utility.cs for SHA256.HashData, CryptographicOperations.FixedTimeEquals, RandomNumberGenerator.Fill). 

Wilson intentionally uses conditional compilation rather than polyfills for performance - polyfills
 add runtime indirection while #if gives the optimal codepath per TFM with zero overhead. Tests run on both .NET (net8.0+) and .NET Framework (net472, net462) .

For TimeProvider specifically, on netstandard2.0/net472 it falls back to DateTimeOffset.UtcNow - the property is simply not exposed on those TFMs (guarded by #if NET8_0_OR_GREATER), so there's nothing to polyfill.

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
MZOLN: I understand the concern. This would be set to false for token aquisition. 

The default is true (ath required), which is the secure default for resource servers. 
Setting it to false is only appropriate for token endpoint proof validation. 

MISE always sets it to true for resource protection.

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
MZOLN: I think that this is implied from dpop context. Similarly to renaming ProofAllowedSingingAlgorithms to AllowedSigningAlgorithms it makes it more consistent, especially if we consider usage from the perspective of MISE options. I would leave it as is, please let's discuss further if you disagree.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: It's used by MISE for validating binding between the token and the proof.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Poly-fil causes indirection. Wilson avoids poly-fil.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: It's not needed, let me remove. Good catch.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,423 @@
...
+        var hash = SHA256.HashData(tokenBytes);
+#else
+        byte[] hash;
+        using (var sha256 = SHA256.Create())
+        {
+            hash = sha256.ComputeHash(tokenBytes);
+        }
+#endif
+        return Base64UrlEncoder.Encode(hash);
+    }
```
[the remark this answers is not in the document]
MZOLN: I recall fixing this before, let me check.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Removed.

test/Microsoft.IdentityModel.Dpop.Tests/DpopTests.cs
MZOLN: Removed

src/Microsoft.IdentityModel.Dpop/DPoPProof.cs
MZOLN: Moved creation helper to tests, which lead to duplication but removes this concern out of the product.

test/Microsoft.IdentityModel.Dpop.Tests/DpopTests.cs
MZOLN: Removed

test/Microsoft.IdentityModel.Dpop.Tests/DpopTests.cs
```diff
@@ -0,0 +1,233 @@
...
+        }
```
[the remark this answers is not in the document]
MZOLN: Removed

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Moved to internal

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Resolved

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,423 @@
...
+
+        try
+        {
+            return await ValidateCoreAsync(dpopProofJwt, httpMethod, requestUri, accessToken, options, cancellationToken)
+                .ConfigureAwait(false);
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
+        {
+            return DPoPValidationResult.Failed(
+                $"DPoP proof validation error: {ex.Message}",
```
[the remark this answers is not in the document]
MZOLN: This follows Wilson's existing pattern - both JsonWebTokenHandler.ValidateTokenAsync and 
 SignedHttpRequestHandler.ValidateSignedHttpRequestAsync return the full exception object on their result types ( TokenValidationResult.Exception, SignedHttpRequestValidationResult.Exception). 
 
 The Error string on DPoPValidationResult is consumed server-side for logging/diagnostics, not returned to HTTP clients. MISE's DPopResponseHeaderCreator constructs its own sanitized WWW-Authenticate challenge headers using DPoPErrorCodes constants -> it never surfaces result.Error to the client. That said, if we want to be extra cautious, we could use LogHelper.FormatInvariant to log the detail and return a generic string — but that would diverge from how SHR/JWT  handle it.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Found a method in Wilson that does exactly this. Switching to that method instead.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+        string httpMethod,
+        Uri requestUri,
+        string accessToken,
+        DPoPValidationOptions options,
+        CancellationToken cancellationToken = default)
+    {
+        _ = dpopProofJwt ?? throw new ArgumentNullException(nameof(dpopProofJwt));
+        _ = httpMethod ?? throw new ArgumentNullException(nameof(httpMethod));
+        _ = requestUri ?? throw new ArgumentNullException(nameof(requestUri));
+        _ = options ?? throw new ArgumentNullException(nameof(options));
```
robezirg, line 57: how come access token is not null checked?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+
+        if (string.IsNullOrWhiteSpace(dpopProofJwt))
```
robezirg, line 59: ambiguous logic, should we return failed when the dpopProofJwt or arg exception?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+
+        try
+        {
+            return await ValidateCoreAsync(dpopProofJwt, httpMethod, requestUri, accessToken, options, cancellationToken)
+                .ConfigureAwait(false);
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
+        {
+            return DPoPValidationResult.Failed(
+                $"DPoP proof validation error: {ex.Message}",
```
robezirg: is this standard practice in Willson to wrap the whole internal ex message and lose the stack trace?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+                DPoPErrorCodes.InvalidDPoPProof);
```
robezirg: having empty dpopProofJwt and exception happening during validation seems like different scenarios, but the end user error code is the same. is this intentional?

robezirg: this can happen for many different reasons right? InvalidDPoPProof seems like an assumption here

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        // Validate alg is asymmetric and in allowed set
+        var alg = proofToken.Alg;
+        if (string.IsNullOrEmpty(alg) ||
+            string.Equals(alg, "none", StringComparison.OrdinalIgnoreCase))
+        {
+            return DPoPValidationResult.Failed(
+                "DPoP proof algorithm must not be 'none'.",
```
robezirg: none or empty

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        if (alg.StartsWith("HS", StringComparison.OrdinalIgnoreCase))
+        {
+            return DPoPValidationResult.Failed(
+                "DPoP proof must use an asymmetric algorithm.",
+                DPoPErrorCodes.InvalidDPoPProof);
+        }
+
+        if (options.AllowedSigningAlgorithms.Count > 0 &&
```
robezirg: what if count is 0?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+            return DPoPValidationResult.Failed(
+                $"DPoP proof algorithm '{alg}' is not in the allowed set.",
+                DPoPErrorCodes.InvalidDPoPProof);
+        }
+
+        // Extract JWK from header, verify no private key present
+        if (!proofToken.TryGetHeaderValue("jwk", out object jwkObj) || jwkObj == null)
+        {
+            return DPoPValidationResult.Failed(
+                "DPoP proof is missing the 'jwk' header parameter.",
```
robezirg: do we not assign error codes for these cases? would be easier for end user to identify the more specific error

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+        {
+            return DPoPValidationResult.Failed(
+                $"DPoP proof contains an invalid 'jwk' header: {ex.Message}",
+                DPoPErrorCodes.InvalidDPoPProof);
+        }
+
+        if (ContainsPrivateKeyMaterial(jwk))
+        {
+            return DPoPValidationResult.Failed(
+                "DPoP proof JWK must not contain private key material.",
```
robezirg: nit: jwk for error code consistency

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+
+namespace Microsoft.IdentityModel.Dpop;
+
+/// <summary>
+/// Validates DPoP proof JWTs on the server side per RFC 9449 §4.3.
+/// </summary>
+/// <remarks>
+/// This class is stateless and thread-safe. A single instance may be reused across requests.
+/// </remarks>
+public class DPoPProofValidator
```
robezirg, line 20: so there is no common interface that these protocols are sharing?

test/Microsoft.IdentityModel.Dpop.Tests/DPoPE2ETests.cs
```diff
@@ -0,0 +1,249 @@
...
+        var thumbprint = DPoPProofValidator.ComputeJwkThumbprint(proofJwk);
+
+        // Build the cnf claim as a JSON string (Wilson handles it via JsonElement)
+        var cnfJson = $"{{\"jkt\":\"{thumbprint}\"}}";
+
+        // Parse cnf into JsonElement for proper embedding
+        using var cnfDoc = System.Text.Json.JsonDocument.Parse(cnfJson);
+        var cnfElement = cnfDoc.RootElement.Clone();
+
+        var handler = new JsonWebTokenHandler();
```
robezirg, line 51: should this not be a singleton? is there state that goes into this?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+
+    private static async Task<DPoPValidationResult> ValidateCoreAsync(
+        string dpopProofJwt,
+        string httpMethod,
+        Uri requestUri,
+        string accessToken,
+        DPoPValidationOptions options,
+        CancellationToken cancellationToken)
+    {
+        var handler = new JsonWebTokenHandler();
```
robezirg: can this not be made static? is there state that handler holds? (i.E. cant be reused or is not thread safe)

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+            ValidateAudience = false,
+            ValidateLifetime = false,
+            IssuerSigningKey = jwk,
+            ValidAlgorithms = new string[] { alg },
+        };
+
+        var signatureResult = await handler
+            .ValidateTokenAsync(proofToken, validationParams, cancellationToken)
+            .ConfigureAwait(false);
+
```
robezirg, line 223: should ct be checked here for cancellation ?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+        if (!signatureResult.IsValid)
+        {
+            return DPoPValidationResult.Failed(
+                "DPoP proof signature validation failed.",
+                DPoPErrorCodes.InvalidDPoPProof);
+        }
+
+        // Validate htm matches HTTP method
+        if (!proofToken.TryGetPayloadValue(DPoPClaimTypes.Htm, out string htmValue) ||
```
robezirg: are these not cheaper checks to make before the sig validation? what should be the order of ops for the most optimal execution?

robezirg: you should check that htm exists first before you compare to the http method

robezirg: by this i mean it should produce a different error if the claim is missing, like you did for others, such as htu

also as with prev comments, from devex perspective, would you rather check that the claims are missing first or check that claim x is not missing and the logic is correct.

i.e. with this logic, if htm claim is not missing but does not match, howevre the htu is missing, you will see an error that htm claims does not match, but should the end user first want to know that htu is missing over htm not matching? 

overall this is a nit, just wanna make sure this choice is explicit and is okay with you

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+
+        if (now - issuedAt > maxAge)
+        {
+            return DPoPValidationResult.Failed(
+                "DPoP proof has expired.",
+                DPoPErrorCodes.InvalidDPoPProof);
+        }
+
+        // Reject proofs issued in the future beyond clock skew
+        if (issuedAt - now > TimeSpan.FromSeconds(options.ClockSkewInSeconds))
```
robezirg, line 273: so clock skew works in both directions? past and future?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        // Replay protection
+        if (options.JtiReplayCache != null)
+        {
+            var jtiExpiration = issuedAt.Add(maxAge);
+            bool added = await options.JtiReplayCache
+                .TryAddAsync(jtiValue, jtiExpiration, cancellationToken)
+                .ConfigureAwait(false);
+
```
robezirg, line 294: possible point of checking the ct here as well

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+            if (!added)
+            {
+                return DPoPValidationResult.Failed(
+                    "DPoP proof 'jti' has already been used (replay detected).",
+                    DPoPErrorCodes.InvalidDPoPProof);
+            }
+        }
+
+        // Validate nonce if expected
+        if (options.ExpectedNonce != null)
```
robezirg, line 302: can it be empty?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+        {
+            if (!proofToken.TryGetPayloadValue(DPoPClaimTypes.Nonce, out string nonceValue) ||
+                string.IsNullOrEmpty(nonceValue))
+            {
+                return DPoPValidationResult.NonceRequired();
```
robezirg, line 312: why does nonce get special treatment here vs other claims

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+            }
+
+            if (!string.Equals(options.ExpectedNonce, nonceValue, StringComparison.Ordinal))
+            {
+                return DPoPValidationResult.NonceRequired();
```
robezirg: seems like incorrect messaging here, the result should be that invalid nonce

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+            }
+        }
+
+        // Validate ath (access token hash)
+        if (!string.IsNullOrEmpty(accessToken) && options.RequireAccessTokenHash)
```
robezirg: so AT being null is a legit scenario?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+        {
+            if (!proofToken.TryGetPayloadValue(DPoPClaimTypes.Ath, out string athValue) ||
+                string.IsNullOrEmpty(athValue))
+            {
+                return DPoPValidationResult.Failed(
```
robezirg: nit: I would create a helper method and just plugin the claim name rather than repeating this 10 times with different claim names

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    /// <summary>
+    /// Validates that the <c>cnf.jkt</c> claim in an access token matches the DPoP proof's
+    /// public key thumbprint, using constant-time comparison to prevent timing attacks.
+    /// </summary>
+    /// <param name="accessTokenCnfJkt">The <c>jkt</c> value from the access token's <c>cnf</c> claim.</param>
+    /// <param name="dpopProofJwkThumbprint">The computed JWK SHA-256 thumbprint from the DPoP proof.</param>
+    /// <returns><see langword="true"/> if the values match; otherwise <see langword="false"/>.</returns>
+    public static bool ValidateCnfJktBinding(string accessTokenCnfJkt, string dpopProofJwkThumbprint)
```
robezirg: why is this a specific public API?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,372 @@
...
+    /// <returns>The base64url-encoded SHA-256 hash.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="accessToken"/> is null.</exception>
+    internal static string ComputeAccessTokenHash(string accessToken)
+    {
+        _ = accessToken ?? throw new ArgumentNullException(nameof(accessToken));
+
+        // Use strict ASCII encoding per RFC 9449 §4.2 — reject non-ASCII tokens
+        // rather than silently replacing with '?' (which would produce an incorrect hash).
+        var tokenBytes = Encoding.GetEncoding(
+            "us-ascii",
```
robezirg: no cosnt for this?

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
```diff
@@ -1092,6 +1093,11 @@ public static void WriteObject(ref Utf8JsonWriter writer, string key, object obj
                 writer.WritePropertyName(key);
                 j.WriteTo(writer);
             }
+            else if (obj is JsonNode jn)
```
robezirg, line 1096: curious why we never needed this before

src/Microsoft.IdentityModel.Tokens/PublicAPI/net10.0/PublicAPI.Unshipped.txt
```diff
@@ -1 +1 @@
-~Microsoft.IdentityModel.Tokens.CaseSensitiveClaimsIdentity.SecurityToken.set -> void
\ No newline at end of file
+~Microsoft.IdentityModel.Tokens.CaseSensitiveClaimsIdentity.SecurityToken.set -> void
```
robezirg, line 1: what has changed here?

src/Microsoft.IdentityModel.Tokens/Microsoft.IdentityModel.Tokens.csproj
```diff
@@ -10,6 +10,7 @@
     <PackageTags>.NET;Windows;Authentication;Identity;SecurityTokens;Cryptographic operations;Signing;Verifying Signatures;Encryption</PackageTags>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageReadmeFile>README.md</PackageReadmeFile>
+    <NoWarn>$(NoWarn)</NoWarn>
```
robezirg: ?

MZOLN: Yeah this is moot now, we could revert it. Would make sense.

src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs
MZOLN: We never had this case before -> no existing code constructs a JSON object and passes it as a header claim value. DPoP proofs require the jwk header to be an embedded JSON object (not a string). 

The alternative is string concatenation ($"{{\"kty\":\"{kty}\",...") which is fragile and risks malformed JSON. JsonObject + WriteTo is the type-safe approach, consistent with the existing JsonElement branch right above it.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: It's needed for validation -> MISE part explains it.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: we could absolutely check it.

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
MZOLN: Yes that's a standard in newer .NET platforms. Bogdan called it out as well, I've placed a more exhaustive reasoning there.

src/Microsoft.IdentityModel.Dpop/DPoPConstants.cs
MZOLN: Http header names are case insensitive, however this particular casing matches what RFC 9449 §7.1 uses in its examples, so it's the conventional form.

src/Microsoft.IdentityModel.Dpop/DPoPConstants.cs
MZOLN: Good call - ConfirmationClaimTypes.Cnf already exists in Microsoft.IdentityModel.Tokens (which we reference). Will remove the duplicate from DPoPConstants and use the existing one.

src/Microsoft.IdentityModel.Dpop/DPoPErrorCodes.cs
MZOLN: RFC 9449 §7.1 says the server SHOULD use invalid_token in WWW-Authenticate responses, which we do (MISE PR).

invalid_dpop_proof is our internal error code on DPoPValidationResult. MISE maps it to invalid_token in the
actual HTTP response (see DPopResponseHeaderCreator line 124). 

The distinct internal code lets servers differentiate "AT failed" from "proof failed" without parsing descriptions.

test/Microsoft.IdentityModel.Dpop.Tests/DPoPE2ETests.cs
MZOLN: I think it shouldn't make much difference in a test utility function?

src/Microsoft.IdentityModel.Tokens/PublicAPI/net10.0/PublicAPI.Unshipped.txt
MZOLN: Whitespace-only change -> likely a trailing space or line ending normalization. No functional difference.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: The ArgumentNullException guard above makes the null path dead code, but .NET doesn't offer a string.IsWhitespace() without the null check -> wouldnt IsNullOrWhiteSpace be the idiomatic API?

The intent is to return a structured Failed result for empty/whitespace input (malformed, not a programming error) rather than throw. This is consistent with how the rest of ValidateAsync returns Failed for invalid proofs instead of throwing.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: No it's a miss. The DPoPValidationResult should have an exception property instead.

MZOLN: Fixed. Passed to MISE, which passes it through Now it's consistent. Thanks

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Yes, intentional -> every failure from this class is a proof failure by definition. The sub-reason is in Error and Exception. Honestly, the error code is not strictly necessary -> Wilson's prior art (TokenValidationResult, SignedHttpRequestValidationResult) doesn't have error codes, just IsValid + Exception. 

The only flow-control distinction we need is IsNonceRequired (which is a separate bool). 

The error code is a convenience for logging/diagnostics but could be removed if you would prefer to stick strictly to Wilson's existing pattern.

src/Microsoft.IdentityModel.Dpop/DPoPErrorCodes.cs
robezirg: I think we should call it out in comments then, because the reader might think that this is defined in RFC

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
robezirg: understood, that was my guess as well, but wanted to make sure this is deliberate

test/Microsoft.IdentityModel.Dpop.Tests/DPoPE2ETests.cs
robezirg: sorry this is residual comment, i meant in the prod code

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,334 @@
...
+    /// <paramref name="requestUri"/>, or <paramref name="options"/> is null.
+    /// </exception>
+    /// <exception cref="ArgumentException">
+    /// Thrown when <paramref name="requestUri"/> is not an absolute URI.
+    /// </exception>
+    public virtual async Task<DPoPValidationResult> ValidateAsync(
+        string dpopProofJwt,
+        string httpMethod,
+        Uri requestUri,
+        string accessToken,
```
bgavrilMS, line 47: I find this API a bit misleading. It takes in an accessToken, so I'd expect this API will do everything: 

- validate ath
- validate the binding 

But it will only validate the ath, not the binding. For the binding, you need to use extra API ValidateCnfJktBinding. 

I can see how ppl will forget about ValidateCnfJktBinding and will have bad implementations.

How about a ValidateAPI that does everything, and instead of accessToken it takes in the parsed token (SecurityTicket or whatever the object is). 

And then add `ValidateWithoutBindingAsync(... string accessTokenHash)` and `ValidateCnfJktBinding` as separaet APIs? Maybe in a differnt namespace (e.g. extensiblity)

src/Microsoft.IdentityModel.Dpop/Microsoft.IdentityModel.Dpop.csproj
```diff
@@ -0,0 +1,44 @@
...
+    <AdditionalFiles Include="PublicAPI/$(TargetFramework)/PublicAPI.Shipped.txt" />
+    <AdditionalFiles Include="PublicAPI/$(TargetFramework)/PublicAPI.Unshipped.txt" />
+    <AdditionalFiles Include="InternalAPI.Shipped.txt" />
```
bgavrilMS, line 33: nit: we should really stop adding friend assemblies except for tests. Ideally new projects don't have this.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Yes, it was, in original PR iteration I wanted to support scenario where library would be used in TA flows, but after discussing this with Bogdan, I'm cutting this scope and reducing to validation flows only.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: We need a control value to be able to emit a header in case the nonce is not present. It is to support the nonce flow.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Nope, not in Wilson.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Thank you @bgavrilMS  for taking the time to discuss this extensively and alleviating my concerns with regard to usage of this library in token acquisition flows. 

As agreed focusing only on validation. This is simplifying the API significantly and I might be able to reduce the public API surface even more.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Collapsed it into existing validation flow.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Actually since it's always a base64url encoded jwt here it will always be ascii and we can just follow wilson here and not listen to copilot. Let me revert it back to net framework ASCII encoding object and add a comment as to why it's safe.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: As long as we dont mutate the state it should be safe to have it as static. We only deserialize the token in this code so it should be fine. Modified.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Then contains will return false.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: CancellationToken is already forwarded into ValidateTokenAsync which should throw OperationCanceledException if cancelled during signature validation. 

The outer ValidateAsync has cancellationToken.ThrowIfCancellationRequested() at entry, and the catch (Exception ex) when (ex is not OperationCanceledException) filter lets cancellation propagate naturally. No additional check needed here.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: If operation is cancelled the async method should throw. The pattern is to check for cancellation at the beginning of the method not at every turn. What we have should suffice.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Good point on ordering. RFC 9449 §4.3 specifies a numbered sequence where signature validation (step 6) comes before claim checks (steps 8-12). We follow the RFC order.

That said, the RFC doesn't mandate this order -> reordering for performance (cheap string checks first, expensive crypto last) is valid. However, checking claims on an unsigned proof means we'd report specific claim errors for potentially forged proofs, which could leak validation logic to attackers (which in this case would probably not be so dramatic). Signature-first is the defensive choice -> reject forged proofs early with a generic error before revealing what specific claims are wrong. 

Fair point as well on devex. Currently TryGetPayloadValue returns false for both "missing" and "wrong type" -> the combined || produces one error message. Splitting into separate "claim missing" vs "claim mismatch" checks is more informative. On the ordering question (report all missing claims before mismatches) — that would require a two-pass validation which adds complexity for marginal devex gain. The current sequential approach matches SHR's pattern: each claim is fully validated in order, and the first failure is reported. Happy to split the missing vs mismatch messages if the team wants more granular errors, but I'd keep the sequential order.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Yes clock skew is bidirectional. A proof from a client whose clock is slightly ahead (future iat) shouldn't be rejected, just like a proof from a slightly behind clock.

This matches Wilson's  TokenValidationParameters.ClockSkew behavior which also applies symmetrically to nbf and exp.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Answered this below.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Good catch empty string is ambiguous. Handled it with explict additional error message, in case the expected nonce was not set by the server for some reason.

src/Microsoft.IdentityModel.Dpop/Microsoft.IdentityModel.Dpop.csproj
MZOLN: I agree with the sentiment. The InternalAPI files are required by the PublicApiAnalyzer to track internals exposed via InternalsVisibleTo (which we have for the test project only, which your comment agrees with?).

If we remove InternalsVisibleTo and make the 3 helper methods private, the InternalAPI files go away too. 
The tradeoff is losing direct unit test coverage of ComputeAccessTokenHash, ComputeJwkThumbprint, and ContainsPrivateKeyMaterial -> they'd only be tested indirectly through ValidateAsync, which would be cleaner in my mind but more costly. Happy to go either way.

bgavrilMS: It's ok for the test project to access internals for testing purposes. But there is no need to track the internal api for that.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -307,28 +316,30 @@ private static async Task<DPoPValidationResult> ValidateCoreAsync(
...
             {
-                return DPoPValidationResult.NonceRequired();
+                return DPoPValidationResult.NonceValidationFailed();
             }
         }
 
-        // Validate ath (access token hash)
-        if (!string.IsNullOrEmpty(accessToken) && options.RequireAccessTokenHash)
+        // Validate ath (access token hash) — always required since accessToken is required
+        if (!proofToken.TryGetPayloadValue(DPoPClaimTypes.Ath, out string athValue) ||
```
bgavrilMS, line 322: But you are now making the ath claim mandatory for proofs. Is this intended? The RFC states it is optional?

Maybe it's better to validate the ath only if it present?

bgavrilMS (approved):

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Yes, intentional.

It's optional only for token acquisition it's always needed for token validation.

RFC 9449 §4.2 says ath is REQUIRED when the DPoP proof is used "in conjunction with the presentation of an access token."

Since we cut away support for TA flows, accessToken is always a non-null parameter, we're always in that scenario. The ath-optional case only applies at the token endpoint where no access token exists yet, which is TA which we explicitly cut yesterday.

src/Microsoft.IdentityModel.Dpop/Microsoft.IdentityModel.Dpop.csproj
MZOLN: The tool throws warnings otherwise.

bgavrilMS (approved):

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,346 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Validates that the <c>cnf.jkt</c> claim in an access token matches the DPoP proof's
+    /// public key thumbprint, using constant-time comparison to prevent timing attacks.
+    /// </summary>
+    private static bool ValidateCnfJktBinding(string accessTokenCnfJkt, string dpopProofJwkThumbprint)
+    {
+        if (string.IsNullOrEmpty(accessTokenCnfJkt) || string.IsNullOrEmpty(dpopProofJwkThumbprint))
```
robezirg: do we wanna guard on white space?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
robezirg: it will not run since you have a short circuit on it being bigger than 0

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
robezirg: my point is that at entry it is less likely to be expired / cancelled then on exit. is there such check?

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
robezirg: this pattern is the only one doing this, all the other claims check if it exists and then checks the value separately

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: It's actually guaranteed by the caller upstream so the null check only made sense here, however since we collapsed this method to "private" we could now just remove the indirection and make this clearer.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Added explicit checks

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: There are pros and cons to the approaches. Let's take this in a different conversation. Discussed offline.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Added explicit check.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,344 @@
...
+        {
+            return DPoPValidationResult.Failed("DPoP proof algorithm must not be 'none'.");
+        }
+
+        if (SupportedAlgorithms.IsSupportedSymmetricAlgorithm(alg))
+        {
+            return DPoPValidationResult.Failed("DPoP proof must use an asymmetric algorithm.");
+        }
+
+        if (options.AllowedSigningAlgorithms == null || options.AllowedSigningAlgorithms.Count <= 0)
```
robezirg, line 179: is there a check to see if it is assymetric? 

nit: but would be more precise to say is it !assymetric then throw error, because maybe at some point in time, !symmetric might not mean asymmetric. maybe there is another kind you know. Not a blocker, but food for thought

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
```diff
@@ -0,0 +1,344 @@
...
+            }
+
+            if (!string.Equals(options.ExpectedNonce, nonceValue, StringComparison.Ordinal))
+            {
+                return DPoPValidationResult.NonceValidationFailed();
+            }
+        }
+
+        // Validate ath (access token hash) — always required since accessToken is required
+        if (!proofToken.TryGetPayloadValue(DPoPClaimTypes.Ath, out string athValue) ||
```
robezirg, line 322: as per the other comment, i want to understand how the acquisition flow going to use this class if at all.

robezirg (approved):
approving with some questions

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: The RFC states 
"It MUST NOT be none or an identifier for a symmetric algorithm (Message Authentication Code (MAC))."

Thus it is an explicit implementation of that.

Our class currently only has IsSymmetric(...) thus negation implies asymmetric.
I think it's low risk, as it is still controlled by the developer.

src/Microsoft.IdentityModel.Dpop/DPoPProofValidator.cs
MZOLN: Discussed offline. It is possible to extend it without breaking changes.

PR #3444, opened 2026-04-14
L2 cache bypass

westin-m:
Addresses [Bug 3542293](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3542293): RequestRefresh not bypassing L2 Cache

Introduce a new interface that provides context to the event handler implementation, allowing it to optionally bypass a cache lookup. 

**This will need to be merged and released before [the corresponding MISE fix](https://identitydivision.visualstudio.com/DevEx/_git/MISE/pullrequest/23268) can be merged.**


src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationEventHandlerContextAware.cs
```diff
@@ -0,0 +1,34 @@
...
+    Task<ConfigurationEventHandlerResult<T>> BeforeRetrieveAsync(string metadataAddress, ConfigurationRetrievalContext context, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Called after a configuration has been successfully retrieved in a fire-and-forget manner.
+    /// </summary>
+    /// <param name="metadataAddress">The metadata endpoint address.</param>
+    /// <param name="configuration">The retrieved configuration.</param>
+    /// <param name="context">The context for the configuration retrieval operation, providing additional information and control.</param>
+    /// <param name="cancellationToken">A cancellation token to observe while waiting for the task to complete.</param>
+    Task AfterUpdateAsync(string metadataAddress, T configuration, ConfigurationRetrievalContext context, CancellationToken cancellationToken = default);
```
robezirg, line 33: I don't think we need to add this now, we can always add as needed? idk I can understand why you did this as the class name suggest context aware

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
robezirg, line 481: how about this case?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -493,7 +495,20 @@ private async Task<ConfigurationEventHandlerResult<T>> HandleBeforeRetrieveAsync
...
             try
             {
-                var handlerResult = await ConfigurationEventHandler.BeforeRetrieveAsync(MetadataAddress, cancellationToken).ConfigureAwait(false);
+                ConfigurationEventHandlerResult<T> handlerResult;
+                if (ConfigurationEventHandler is IConfigurationEventHandlerContextAware<T> contextAware)
+                {
+                    var context = new ConfigurationRetrievalContext() { BypassCache = _isRefreshRequested };
+                    handlerResult = await contextAware.BeforeRetrieveAsync(
+                        MetadataAddress, context, cancellationToken).ConfigureAwait(false);
+                    _isRefreshRequested = false;
```
robezirg: is there any issue we can run into with this logic?

1. we only set it to false in this one path
2. is there potential of concurrency issues? 

if refresh is requested, is there a reason not to set bypass to true? why do we need to keep instance var?

src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationEventHandlerContextAware.cs
westin-m: I agree that we don't need it now. It's taken out with a comment acknowledging that we can add it later

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
westin-m: Didn't have it for the poc, but I've updated the blocking path now

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
westin-m: The instance variable would introduce concurrency issues, so the updated changes use a parameter instead.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -523,6 +538,7 @@ private async Task<ConfigurationEventHandlerResult<T>> HandleBeforeRetrieveAsync
                     return handlerResult;
                 }
             }
+#pragma warning disable CA1031 // Do not catch general exception types
```
robezirg: why is this needed?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -331,7 +331,10 @@ private async Task<T> GetConfigurationNonBlockingAsync(CancellationToken cancel)
         /// The Caller should first check the state checking state using:
         ///   if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle).
         /// </summary>
-        private async Task UpdateCurrentConfigurationAsync()
+        private async Task UpdateCurrentConfigurationAsync() =>
+            await UpdateCurrentConfigurationAsync(bypassCache: false).ConfigureAwait(false);
```
robezirg: no need to await

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -481,19 +484,31 @@ private void RequestRefreshBackgroundThread()
...
                 {
-                    _ = Task.Run(UpdateCurrentConfigurationAsync, CancellationToken.None);
+                    _ = Task.Run(() => UpdateCurrentConfigurationAsync(bypassCache: true), CancellationToken.None);
                     _lastRequestRefresh = now;
                 }
             }
         }
 
-        private async Task<ConfigurationEventHandlerResult<T>> HandleBeforeRetrieveAsync(CancellationToken cancellationToken = default)
+        private async Task<ConfigurationEventHandlerResult<T>> HandleBeforeRetrieveAsync(bool bypassCache, CancellationToken cancellationToken = default)
```
robezirg: is there value of passing the context into this method? I think passing bool is fine, but just asking the question outloud

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
robezirg: in fact you don't even need this, you can default it to false in the method

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -331,7 +331,10 @@ private async Task<T> GetConfigurationNonBlockingAsync(CancellationToken cancel)
         /// The Caller should first check the state checking state using:
         ///   if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle).
         /// </summary>
-        private async Task UpdateCurrentConfigurationAsync()
+        private async Task UpdateCurrentConfigurationAsync() =>
+            await UpdateCurrentConfigurationAsync(bypassCache: false).ConfigureAwait(false);
+
+        private async Task UpdateCurrentConfigurationAsync(bool bypassCache)
```
robezirg, line 342: is this only in the non-blocking scenario?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
westin-m: I dug and found that it's suppressed elsewhere, I just need to add the new interface to that suppression. The reasoning is because an interface implementation could throw any type of exception.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
westin-m: I'm not sure I understand your question. Analogous changes were made to the blocking config manager

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
westin-m: Probably the same amount of value that we get from eventually making the context anyway, so I will pass the context to this method

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -481,19 +482,30 @@ private void RequestRefreshBackgroundThread()
                 _isFirstRefreshRequest = false;
                 if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
                 {
-                    _ = Task.Run(UpdateCurrentConfigurationAsync, CancellationToken.None);
+                    _ = Task.Run(() => UpdateCurrentConfigurationAsync(bypassCache: true), CancellationToken.None);
```
robezirg: these task runs are now closures that basically are re-allocating memory every time. 

we can create things like

private static Task UpdateCurrentConfigurationWithBypass()
    => UpdateCurrentConfigurationAsync(bypassCache: true);

so that caching is possible.

westin-m: Makes sense. Thank you for calling that out

src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationEventHandlerContextAware.cs
pmaytak: How can we add it later; we won't be able to change the interface?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -231,7 +234,7 @@ private async Task<T> GetConfigurationNonBlockingAsync(CancellationToken cancel)
                     // If provided configuration is valid, skip regular retriaval process and update current configuration.
                     if (ConfigurationEventHandler != null)
                     {
-                        var configurationRetrieved = await HandleBeforeRetrieveAsync(cancel).ConfigureAwait(false);
+                        var configurationRetrieved = await HandleBeforeRetrieveAsync(new ConfigurationRetrievalContext(), cancel).ConfigureAwait(false);
```
pmaytak:
```suggestion
                        var configurationRetrieved = await HandleBeforeRetrieveAsync(new ConfigurationRetrievalContext { BypassCache = false }, cancel).ConfigureAwait(false);
```

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerEventHandlerContextAwareTests.cs
```diff
@@ -0,0 +1,238 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Time.Testing;
+using Microsoft.IdentityModel.TestUtils;
+using Microsoft.IdentityModel.Tokens;
```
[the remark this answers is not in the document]
westin-m: that package is used for the appcontextswitches and removing it creates a compilation error

src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationEventHandlerContextAware.cs
robezirg: is adding methods to a protocol a breaking change?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -205,6 +205,9 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
                 return await GetConfigurationNonBlockingAsync(cancel).ConfigureAwait(false);
         }
 
+        private Task UpdateCurrentConfigurationWithBypassAsync()
```
robezirg: I missed the fact that the method UpdateCurrentConfigurationAsync is not static itself, so i think this doesn't achieve what I wanted.

you can do this by having this instead. 
private readonly Func<Task> _updateWithBypass;
private readonly Func<Task> _updateWithoutBypass;

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -205,6 +205,9 @@ public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
...
+            => UpdateCurrentConfigurationAsync(bypassCache: false);
```
robezirg: this is incorrect should be true

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -308,7 +311,7 @@ private async Task<T> GetConfigurationNonBlockingAsync(CancellationToken cancel)
                         TelemetryConstants.Protocols.Automatic,
                         TelemetryConstants.Protocols.ConfigurationSourceUnknown);
 
-                    _ = Task.Run(UpdateCurrentConfigurationAsync, CancellationToken.None);
+                    _ = Task.Run(UpdateCurrentConfigurationWithBypassAsync, CancellationToken.None);
```
robezirg: this should be without bypass

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -481,19 +485,30 @@ private void RequestRefreshBackgroundThread()
                 _isFirstRefreshRequest = false;
                 if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
                 {
-                    _ = Task.Run(UpdateCurrentConfigurationAsync, CancellationToken.None);
+                    _ = Task.Run(() => UpdateCurrentConfigurationAsync(bypassCache: true), CancellationToken.None);
```
robezirg: you should update this as well with cached versions

src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationEventHandlerContextAware.cs
robezirg: my two cents here is this:
 If adding an interface method will become a breaking change in the future, we can do either or
1. say it is okay that AfterUpdateAsync doesn't get the context. Context might be more relevant to the before update.
2. add it to the AfterUpdateAsync as well (we need to make sure that we can persist the context between calling the before and after) - I am fine with this option

westin-m: I will go with option 2 and add the AfterUpdateAsync with context. We are able to pass the same context between before and after, so I will build that in now as well

src/Microsoft.IdentityModel.Protocols/PublicAPI.Unshipped.txt
```diff
@@ -2,10 +2,17 @@ Microsoft.IdentityModel.Protocols.Configuration.ConfigurationEventHandlerResult<
...
+Microsoft.IdentityModel.Protocols.Configuration.IConfigurationEventHandlerContextAware<T>
+Microsoft.IdentityModel.Protocols.Configuration.IConfigurationEventHandlerContextAware<T>.AfterUpdateAsync(string metadataAddress, T configuration, Microsoft.IdentityModel.Protocols.Configuration.ConfigurationRetrievalContext context, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) -> System.Threading.Tasks.Task
+Microsoft.IdentityModel.Protocols.Configuration.IConfigurationEventHandlerContextAware<T>.BeforeRetrieveAsync(string metadataAddress, Microsoft.IdentityModel.Protocols.Configuration.ConfigurationRetrievalContext context, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) -> System.Threading.Tasks.Task<Microsoft.IdentityModel.Protocols.Configuration.ConfigurationEventHandlerResult<T>>
```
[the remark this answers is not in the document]
westin-m: @copilot+claude-opus-4.7 that is incorrect. The API check passes currently. Entries in the per-TFM files are only needed when APIs differ between target frameworks. This new API is the same across all targets, so it would be redundant and unnecessary to put it in all those files as well.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
```diff
@@ -32,13 +32,15 @@ private async Task<T> GetConfigurationWithBlockingAsync(CancellationToken cancel
             {
                 if (_syncAfter <= TimeProvider.GetUtcNow())
                 {
+                    var retrievalContext = new ConfigurationRetrievalContext { BypassCache = _refreshRequested };
```
robezirg, line 35: you capture the _refreshRequested here, however there is a chance that it is updated in other places, like line 51, i think this would create a bug

westin-m: We would want the after operation to have the same context that the before operation did. Changing the bool after it's been used to build the context won't affect the context

robezirg (approved):

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerEventHandlerContextAwareTests.cs
pmaytak, line 1: From copilot:

> Based on my analysis of the test file and the production code, here are the missing test cases:
> Blocking mode tests (none exist)
> All current tests use the non-blocking (default) path. There are no tests for the blocking path (UpdateConfigAsBlocking), which also has context-aware logic:
> 1.	Blocking: BypassCache is false on first retrieval
> 2.	Blocking: BypassCache is true after RequestRefresh
> 3.	Blocking: BypassCache resets to false after consumed
> 4.	Blocking: Context-aware BeforeRetrieveAsync called instead of base
> 5.	Blocking: Context-aware AfterUpdateAsync called instead of base
> 
> Handler returns configuration (short-circuit path)
> Current tests never set ConfigurationToReturn, so the handler always returns NoResult and config comes from the endpoint. Missing:
> 6.	BeforeRetrieveAsync returns config → BypassCache=false propagated correctly
> 7.	BeforeRetrieveAsync returns config after RequestRefresh → BypassCache=true propagated correctly
> 8.	AfterUpdateAsync receives correct context when handler provided the config (not the retriever)
> 
> Error/edge cases
> 9.	Context-aware BeforeRetrieveAsync throws → falls back to endpoint, context still correct
> 10.	Context-aware AfterUpdateAsync throws → exception logged, no crash (fire-and-forget)
> 11.	Base handler AfterUpdateAsync still called (not context-aware overload) for non-context-aware handler — only BeforeRetrieveAsync(string, CancellationToken) is verified in BaseHandler_StillWorks_WithoutContextAwareInterface(), but AfterUpdateAsync(string, OpenIdConnectConfiguration, CancellationToken) is not checked.
> 
> Automatic refresh with AfterUpdateAsync
> 12.	AfterUpdateAsync BypassCache=false on automatic refresh (only first retrieval and RequestRefresh are tested for AfterUpdateAsync(string, OpenIdConnectConfiguration, CancellationToken))
> 
> Multiple RequestRefresh calls
> 13.	Multiple rapid RequestRefresh calls → only one bypass=true refresh fires (validates the Interlocked.CompareExchange singleton behavior with context)

pmaytak (approved):

test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerEventHandlerContextAwareTests.cs
westin-m: added the tests, thanks for calling these out

pmaytak (approved):

PR #3479, opened 2026-05-08
Add ML-DSA (FIPS 204) post-quantum signature support

iNinja:
# Add ML-DSA (FIPS 204) post-quantum signature support
This PR adds ML-DSA (Module-Lattice-Based Digital Signature Algorithm, FIPS 204) support to Microsoft.IdentityModel.Tokens, enabling post-quantum digital signatures across the JOSE pipeline — JWS signing/verification, JWK key representation, X.509 certificate key extraction, and JWK conversion.

## Standards

| Standard | Status | Usage |
|----------|--------|-------|
| [FIPS 204](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf) | Final | ML-DSA algorithm (key gen, sign, verify) |
| [RFC 9881](https://datatracker.ietf.org/doc/rfc9881/) | Final (Oct 2025) | X.509 algorithm OIDs for ML-DSA-44/65/87 |
| [draft-ietf-cose-dilithium-11](https://datatracker.ietf.org/doc/draft-ietf-cose-dilithium/) (RFC 9964 pending) | AUTH48 — content frozen | JWK key type (`kty=AKP`), parameter names (`pub`, `priv`), JWS algorithm identifiers, thumbprint computation |

### New types
- **`MlDsaSecurityKey`** — `AsymmetricSecurityKey` subclass wrapping `MLDsa`; supports key size, private key detection, JWK thumbprint computation
- **`MlDsaAdapter`** (internal) — creates `MLDsa` from JWK parameters with pub/priv consistency validation

### Modified types
- **`AsymmetricAdapter`** — ML-DSA sign/verify delegates (byte[], offset, span), X.509 ML-DSA routing
- **`AsymmetricSignatureProvider`** — ML-DSA key size maps, hash algorithm guard pattern (`TryGetHashAlgorithmName`)
- **`SupportedAlgorithms`** — ML-DSA algorithm collection, algorithm/key parameter set enforcement
- **`JsonWebKey`** — `Pub`/`Priv` properties, AKP key size, `HasPrivateKey`, thumbprint computation
- **`JsonWebKeySerializer`** — read/write `pub`/`priv` parameters
- **`JsonWebKeyConverter`** — `ConvertFromMlDsaSecurityKey`, `TryConvertToMlDsaSecurityKey`, X.509→JWK conversion for ML-DSA certificates (x5c and extractKeyMaterial modes)
- **`X509SecurityKey`** — ML-DSA OID detection (RFC 9881), `MlDsaPublicKey`/`MlDsaPrivateKey` properties with .NET 6 platform compatibility handling
- **`CryptoTelemetry`** — ML-DSA key algorithm identifiers
- **`SecurityAlgorithms`** — `MlDsa44`, `MlDsa65`, `MlDsa87` constants
- **`JsonWebAlgorithmsKeyTypes`** — `Akp` constant
- **`JsonWebKeyParameterNames`** — `Pub`, `Priv` constants with UTF8 equivalents

### Dependencies
- **`Microsoft.Bcl.Cryptography` ≥ 10.0.2** — provides `MLDsa` type on all Wilson TFMs
- **`System.Memory`** upgraded 4.5.5 → 4.6.3 (transitive requirement)

## Design decisions

- **No conditional compilation** — `MLDsa` is available on all TFMs via the BCL compatibility package
- **No `[Experimental]` on Wilson types** — only internal call sites that invoke .NET experimental X.509 extension methods use `#pragma SYSLIB5006` suppression
- **Algorithm/key enforcement** — `IsSupportedAlgorithm` validates that the ML-DSA key's parameter set matches the requested JOSE algorithm (prevents algorithm confusion)
- **Pub/priv consistency validation** — JWK import verifies the claimed `pub` matches the key derived from `priv` seed
- **`kty=AKP` with mandatory `alg`** — per draft-ietf-cose-dilithium; routing checks both `kty` and `alg`, never `kty` alone
- **JWK thumbprint** — canonical form `{"alg","kty","pub"}` in lexicographic order
- **Private key format** — 32-byte seed only (not expanded key), zeroed after use in `finally` blocks
- **X.509 key extraction** — `MlDsaPrivateKey` catches `PlatformNotSupportedException` on .NET 6 where private key extraction is not supported

## Test coverage (93 tests)

| Category | Tests |
|----------|-------|
| Key construction & properties | 5 |
| JWK thumbprint (deterministic, pub/priv match, cross-key, X509 consistency) | 7 |
| JWK round-trip (SecurityKey ↔ JWK ↔ JSON) | 6 |
| X.509 key size, thumbprint, JWK conversion | 15 |
| JWK negative tests (missing alg, missing pub, invalid alg, pub/priv mismatch) | 4 |
| Algorithm/key mismatch enforcement (MlDsaSecurityKey, JsonWebKey, X509SecurityKey) | 15 |
| Public-key-only signing/verification | 2 |
| Signature correctness (sign/verify round-trip, tampering, cross-key) | 7 |
| E2E JWT create/validate (JsonWebTokenHandler, JwtSecurityTokenHandler, experimental ValidationParameters, JWK keys) | 12 |
| E2E JWT with X.509 (sign with X509SecurityKey, cross-key-type validation) | 6 |
| Existing asymmetric signature tests | 523 (unchanged, no regressions) |

Tests run on: **net462, net472, net6.0, net8.0, net9.0, net10.0**

## Known limitations

1. **X.509 ML-DSA private key on .NET 6** — `GetMLDsaPrivateKey()` throws `PlatformNotSupportedException`. Verification works; signing requires `MlDsaSecurityKey` with a standalone key.
2. **X.509 certificate creation** — `CertificateRequest(string, MLDsa)` is only available on .NET 10. Test certificates are pre-generated PFX embedded as base64 constants.
3. **JOSE draft status** — draft-ietf-cose-dilithium is in AUTH48 (RFC 9964 pending). Content is frozen; algorithm identifiers and wire formats are final.
4. **BCL package net6.0 warnings** — suppressed via targeted `Directory.Build.targets` overrides. Remove when net6.0 is dropped from `SrcTargets`.


build/dependencies.props
```diff
@@ -3,13 +3,14 @@
...
     <BannedApiAnalyzersVersion>4.14.0</BannedApiAnalyzersVersion>
+    <MicrosoftBclCryptographyVersion>10.0.2</MicrosoftBclCryptographyVersion>
     <MicrosoftBclTimeProviderVersion>8.0.1</MicrosoftBclTimeProviderVersion>
     <MicrosoftCSharpVersion>4.5.0</MicrosoftCSharpVersion>
     <MicrosoftSourceLinkGitHubVersion>1.0.0</MicrosoftSourceLinkGitHubVersion>
     <NetStandardVersion>2.0.3</NetStandardVersion>
     <NewtonsoftVersion>13.0.3</NewtonsoftVersion>
     <SystemDiagnosticSourceVersion>6.0.2</SystemDiagnosticSourceVersion>
-    <SystemMemoryVersion>4.5.5</SystemMemoryVersion>
+    <SystemMemoryVersion>4.6.3</SystemMemoryVersion>
```
iNinja, line 13: This is being bumped as a requirement from Microsoft.Bcl.Cryptography.

src/Microsoft.IdentityModel.Tokens/X509SecurityKey.cs
```diff
@@ -171,6 +273,23 @@ public override PrivateKeyStatus PrivateKeyStatus
...
+                    // we return Unknown rather than DoesNotExist because the private key may exist
+                    // in the certificate but the platform cannot extract it.
+                    // Note: AsymmetricSignatureProvider only blocks signing when PrivateKeyStatus ==
+                    // DoesNotExist, so Unknown will pass the constructor guard. However, the adapter
+                    // will fail with a clear InvalidOperationException (IDX10723) when it attempts
+                    // to access the null MlDsaPrivateKey during signing initialization.
+                    MLDsa mlDsaPrivateKey = MlDsaPrivateKey;
+
+                    if (_mlDsaPrivateKeyUnsupported)
+                        return PrivateKeyStatus.Unknown;
```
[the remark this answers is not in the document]
iNinja: This was deliberately changed from `DoesNotExist` to `Unknown` based on an earlier review comment — `DoesNotExist` asserts the key definitively doesn't exist, but on unsupported platforms the key may be present in the certificate; we just can't extract it. The `Unknown` status is semantically accurate per the enum contract. The adapter's null guard (IDX10723: "Unable to extract the private key from the X.509 certificate for algorithm '{0}' (Key: '{1}'). Private key extraction may not be supported on this platform.") provides a clear, actionable failure at signing initialization. Returning `DoesNotExist` would silently report incorrect information to callers who check `PrivateKeyStatus` for diagnostic purposes.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
```diff
@@ -319,6 +319,26 @@ public string Kid
 #endif
         public string Y { get; set; }
 
+        /// <summary>
+        /// Gets or sets the 'pub' (AKP public key).
```
MZOLN, line 323: Do we have a spec for those new properties anywhere? Is it following an RFC?

src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs
```diff
@@ -97,14 +99,34 @@ public static JsonWebKey ConvertFromX509SecurityKey(X509SecurityKey key)
             if (key == null)
                 throw LogHelper.LogArgumentNullException(nameof(key));
 
+            // ML-DSA certificates: PublicKey (AsymmetricAlgorithm) is null because MLDsa
+            // does not inherit from AsymmetricAlgorithm. Route via MlDsaPublicKey instead.
```
MZOLN, line 103: Why not?

src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs
```diff
@@ -136,6 +158,43 @@ public static JsonWebKey ConvertFromX509SecurityKey(X509SecurityKey key, bool re
             if (!representAsRsaKey)
                 return ConvertFromX509SecurityKey(key);
 
+            // ML-DSA: extract key material directly from the X509SecurityKey's cached MLDsa
+            // instance rather than wrapping in a temporary MlDsaSecurityKey. This avoids an
+            // unnecessary allocation and keeps the data flow explicit — we only need the raw
```
MZOLN, line 163: Is it following or breaking existing design?

build/dependencies.props
MZOLN: I guess this drives the requirement upwards for entire dependency tree?

iNinja: Yes. `Microsoft.Bcl.Cryptography` 10.0.2 is a new dependency that provides the `MLDsa` API on downlevel TFMs (net462, net472, net6.0, net8.0, net9.0). Without it we can only support ML-DSA on .NET 10+. `System.Memory` 4.5.5 → 4.6.3 is a transitive requirement from `Microsoft.Bcl.Cryptography`. Both will flow to consumers of the NuGet packages.

src/Microsoft.IdentityModel.Tokens/JsonWebKey.cs
iNinja: The `pub` and `priv` parameters follow the AKP (Algorithm Key Pair) key type definition from [draft-ietf-cose-dilithium](https://datatracker.ietf.org/doc/draft-ietf-cose-dilithium/) (which covers ML-DSA JWK representation). The `kty: "AKP"` value, along with the `pub`, `priv`, and the required `alg` parameter, are defined there. The draft has been assigned RFC 9964 (https://www.rfc-editor.org/auth48/rfc9964) pending publication.

src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs
iNinja: `MLDsa` is a new .NET 10 API that does not inherit from `AsymmetricAlgorithm` due to a .NET runtime design decision. Because of this, `X509SecurityKey.PublicKey` (typed as `AsymmetricAlgorithm`) is null for ML-DSA certificates, so we can't match it in the switch. The early-return guard (`key.MlDsaPublicKey != null`) routes ML-DSA certificates before we reach the switch. I've added a clarifying comment on the switch itself

src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs
iNinja: This was an intentional break for performance reasons. Wrapping in a temporary `MlDsaSecurityKey` would allocate an extra object just to call `ConvertFromMlDsaSecurityKey`, which does the same byte export. If we want to keep consistency, we can route through `ConvertFromMlDsaSecurityKey(new MlDsaSecurityKey(...))` to match the RSA/ECDsa approach at the cost of an extra allocation.

src/Microsoft.IdentityModel.Tokens/X509SecurityKey.cs
```diff
@@ -145,6 +166,81 @@ public AsymmetricAlgorithm PublicKey
...
+                    {
+                        if (!_mlDsaPrivateKeyInitialized)
+                        {
+                            try
+                            {
+#pragma warning disable SYSLIB5006 // GetMLDsaPrivateKey is experimental
+                                _mlDsaPrivateKey = Certificate.GetMLDsaPrivateKey();
+#pragma warning restore SYSLIB5006
+                            }
+                            catch (PlatformNotSupportedException)
```
westin-m, line 190: Is acquiring the lock and waiting for the method to throw the best way to detect an unsupported platform?

iNinja: Good question. The platform detection here is specifically about X.509 private key extraction, not ML-DSA support in general. MLDsa.IsSupported tells us whether the algorithm is available, but GetMLDsaPrivateKey() can still throw PlatformNotSupportedException on older TFMs (e.g. .NET 6 via Microsoft.Bcl.Cryptography) where the algorithm works but the X.509 integration doesn't.

There isn't a separate capability check for "can extract private key from this cert type", so catching the exception on first access and caching the result via _mlDsaPrivateKeyUnsupported is the most reliable approach. The lock is already there for thread-safe lazy initialization (same pattern as the RSA/ECDsa properties above), so the cost of catching once is negligible.

Open to suggestions if you see a cleaner way to probe for this.

westin-m: That makes sense and I do not have any suggestions to make it cleaner. Thank you

westin-m (approved):

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -176,6 +187,23 @@ private void InitializeUsingEcdsaSecurityKey(ECDsaSecurityKey ecdsaSecurityKey)
...
         }
 
+        private void InitializeUsingMlDsaSecurityKey(MlDsaSecurityKey mlDsaSecurityKey)
+        {
+            InitializeUsingMlDsa(mlDsaSecurityKey.MLDsa);
+        }
+
+        private void InitializeUsingMlDsa(MLDsa mlDsa)
+        {
+            MLDsa = mlDsa;
```
MZOLN: This could be used multithreaded and potentially use shared state.

RSA and ECDsa are marked thread safe and we potentially clone state, not sure about this algo and potentially symcrypt.

Specifically verify that MLDsa.SignData / VerifyData are safe for concurrent invocation on a single instance.
We could contact the owning team and add a stress test (e.g. 100 parallel tasks signing different payloads with one provider) so we'd catch a future regression in BCL.

src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs
```diff
@@ -259,6 +318,37 @@ public static bool TryConvertToSecurityKey(JsonWebKey webKey, out SecurityKey ke
...
+
+                    // AKP JWKs with x5c contain a certificate — convert to X509SecurityKey.
+                    // The certificate must contain an ML-DSA key matching the claimed alg.
+                    if (webKey.X5c != null && webKey.X5c.Count > 0)
+                    {
+                        if (!TryConvertToX509SecurityKey(webKey, out key))
+                            return false;
+
+                        if (key is not X509SecurityKey x509Key
+                            || x509Key.MlDsaPublicKey == null
```
MZOLN, line 339: We should call Dispose on the MlDsaPublicKey component to release the underlying SymCrypt handle in the code below. Otherwise we leak native handles to GC finalization on rejection.

src/Microsoft.IdentityModel.Tokens/MlDsaSecurityKey.cs
```diff
@@ -0,0 +1,149 @@
...
+using System;
+using System.Security.Cryptography;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Tokens;
+
+/// <summary>
+/// Represents an ML-DSA security key.
+/// </summary>
+public class MlDsaSecurityKey : AsymmetricSecurityKey
```
MZOLN: Shouldnt this be IDisposable?

We would be leaking handles here.

MZOLN (changes requested):

src/Microsoft.IdentityModel.Tokens/MlDsaSecurityKey.cs
cpp11nullptr: Added disposing for scenario when underlying ML-DSA key is owned by `MlDsaSecurityKey` instance.

src/Microsoft.IdentityModel.Tokens/JsonWebKeyConverter.cs
cpp11nullptr: Good point, fixed.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
bartonjs: > RSA and ECDsa are marked thread safe

No, they're not.  The OS doesn't guarantee it, so .NET doesn't guarantee it.

> We could ... add a stress test

That'll probably pass.  The method is "probably" thread-safe.  But it's not guaranteed.  Do with that what you will :smile:.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -342,6 +430,51 @@ private byte[] SignUsingOffsetECDsa(byte[] bytes, int offset, int count)
...
+            // MLDsa.SignData requires destination to be exactly SignatureSizeInBytes.
+            MLDsa.SignData(data, destination.Slice(0, signatureSize), context: default);
+            bytesWritten = signatureSize;
+            return true;
+        }
+#endif
+
+        private byte[] SignUsingOffsetMlDsa(byte[] bytes, int offset, int count)
+        {
+#if NET6_0_OR_GREATER
```
bartonjs: Why the \#if? Microsoft.Bcl.Cryptography requries System.Memory, so this code should compile and work just fine on older TFMs.  And even the .NET Framework-compatible "slow span" is better than making a copy.

For RSA and ECDsa the problem would be that .NET Framework doesn't have span-reading methods, but MLDsa has them in all TFMs.

src/Microsoft.IdentityModel.Tokens/MlDsaSecurityKey.cs
```diff
@@ -0,0 +1,180 @@
...
+                {
+                    // Cannot determine private key status (e.g., platform limitation).
+                    // Do not cache — let PrivateKeyStatus determine independently.
+                    return false;
+                }
+            }
+
+            return _hasPrivateKey.Value;
+        }
+    }
```
bartonjs: Why isn't this just


```suggestion
    public override bool HasPrivateKey => PrivateKeyStatus == PrivateKeyStatus.Exists;
```

rather than duplicating the export logic?

src/Microsoft.IdentityModel.Tokens/MlDsaSecurityKey.cs
```diff
@@ -0,0 +1,180 @@
...
+    /// </returns>
+    public override PrivateKeyStatus PrivateKeyStatus
+    {
+        get
+        {
+            if (_hasPrivateKey == null)
+            {
+                try
+                {
+                    byte[] seed = MLDsa.ExportMLDsaPrivateSeed();
```
bartonjs: It might not apply to your particular domain, but a key imported from MLDsaPrivateKey is a signing key, but can't export the seed (because it doesn't know it).

So ExportMLDsaPrivateKey is a more reliable test... but doesn't work with non-exportable keys.

Rather than moving the private key information around, why not just try to sign something?

src/Microsoft.IdentityModel.Tokens/MlDsaSecurityKey.cs
cpp11nullptr: Good point, done.

src/Microsoft.IdentityModel.Tokens/MlDsaSecurityKey.cs
cpp11nullptr: Thanks, it was changed accordingly.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
cpp11nullptr: Done.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
cpp11nullptr: It was changed so adapter would have its own copy of ML-DSA key pair or shared one under lock (in case if neither seed nor private key is exportable, e.g., for HSM case).

MZOLN (approved):

brentschmaltz:
Nice job in getting this in :-)

PR #3483, opened 2026-05-12
Add TryValidateSignature and SignatureValidatorWithToken delegate

iNinja:
# Add TryValidateSignature and SignatureValidatorWithToken delegate

## Summary

This PR adds two public APIs to support scenarios where a `SignatureValidatorUsingConfiguration` delegate needs to call back into the handler's own signature validation logic for algorithms it does not handle directly.

### `JsonWebTokenHandler.TryValidateSignature`

A new public instance method that validates a JWT signature against a single specified key:

```csharp
public bool TryValidateSignature(
    JsonWebToken jsonWebToken,
    SecurityKey key,
    TokenValidationParameters validationParameters)
```

- Validates algorithm, creates a `SignatureProvider`, and verifies the signature
- Uses the handler's telemetry infrastructure
- The caller is responsible for key resolution — this method validates a single key
- Throws on null arguments, algorithm validation failure, or inability to create a signature provider
- Returns `false` if the signature does not match

### `SignatureValidatorWithToken` delegate

A new delegate on `TokenValidationParameters` that receives the already-parsed `SecurityToken` instead of a raw string:

```csharp
public delegate SecurityToken SignatureValidatorWithToken(
    SecurityToken token,
    BaseConfiguration configuration);
```

```csharp
// On TokenValidationParameters:
public SignatureValidatorWithToken SignatureValidatorWithToken { get; set; }
```

- Takes priority over `SignatureValidatorUsingConfiguration` and `SignatureValidator` when set
- Avoids re-parsing the token string, since the handler has already parsed it
- Receives the resolved `BaseConfiguration` (OIDC metadata), enabling the delegate to access `configuration.JsonWebKeySet.Keys` for key resolution

### Configuration passthrough fix

The existing `ValidateSignatureUsingDelegates` method was not passing the resolved `BaseConfiguration` to the signature validation delegates (hardcoded to `null` with a `TODO` comment). This has been fixed — all three delegates now receive the configuration that was already resolved upstream.

## Usage example

```csharp

tvp.SignatureValidatorWithToken = (token, tvp, configuration) =>
{
    var jwt = (JsonWebToken)token;
    
    if (IsCustomAlgorithm(jwt.Alg))
    {
        // Handle custom algorithm directly
        VerifyWithCustomProvider(jwt, configuration);
    }
    else
    {
        // Fall back to the handler for standard algorithms
        var key = ResolveKey(jwt, configuration);
        if (!handler.TryValidateSignature(jwt, key, tvp))
            throw new SecurityTokenInvalidSignatureException("Signature validation failed.");
        jwt.SigningKey = key;
    }
    
    return jwt;
};
```

## Tests

8 tests covering:
- Valid RSA signature → returns true
- Wrong key → returns false
- ECDSA key → returns true
- Algorithm mismatch (ECDSA key vs RSA token) → returns false
- Null token/key/validationParameters → throws ArgumentNullException
- End-to-end delegate pattern using `SignatureValidatorWithToken` + `TryValidateSignature`


iNinja:
Abandoning in favour of a simpler approach.

PR #3489, opened 2026-05-14
Add CacheCustomProviders property to CryptoProviderFactory

iNinja:
## Summary

Adds an opt-in property to enable caching of `SignatureProvider` instances returned by `ICryptoProvider.Create()`. Currently, providers created by a `CustomCryptoProvider` bypass the `CryptoProviderCache` entirely, meaning a new provider is created on every signature validation. This can be expensive when provider construction involves key materialisation or native resource allocation.

## New API

```csharp
public bool CacheCustomProviders { get; set; } // default: false
```

When set to `true` (and `CacheSignatureProviders` is also `true`), the `SignatureProvider` returned by `ICryptoProvider.Create()` is routed through the existing `CryptoProviderCache`. Subsequent calls with the same key and algorithm return the cached provider, avoiding repeated construction.

## Behaviour

- **Default (`false`)**: No change from current behaviour. Custom providers are created and released on every call.
- **`true`**: After `Create()` returns a `SignatureProvider`, the cache is checked for an existing equivalent (handling concurrent creation races). If not cached, the provider is added via `TryAdd`. On cache hit, the cached provider is returned and the duplicate is released.

The cache key is `{KeyType}-{InternalId}-{Algorithm}-{ProviderType}`, consistent with built-in providers.

## Race condition handling

When two threads concurrently create providers for the same key/algorithm:
1. Both call `Create()` and get new provider instances
2. One wins the `TryAdd` to cache
3. The other finds the cached provider via `TryGetSignatureProvider`, increments its ref count, releases the duplicate (`Release()` + `ICryptoProvider.Release()`), and returns the cached one

## Copy constructor

The property is copied in `CryptoProviderFactory(CryptoProviderFactory other)` to ensure it propagates correctly when the factory is cloned.

## No breaking changes

Existing behaviour is unchanged when `CacheCustomProviders` is `false` (the default). The `CustomCryptoProvider` early-return path in `CreateSignatureProvider` only reaches the cache logic when the new property is explicitly enabled.


MZOLN:
Could we add some tests? And also some tests for the race conditions?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -608,6 +621,28 @@ private SignatureProvider CreateSignatureProvider(
                                 LogHelper.MarkAsNonPII(key.KeyId),
                                 LogHelper.MarkAsNonPII(typeof(SignatureProvider)))));
 
+                if (CacheCustomProviders && CacheSignatureProviders && cacheProvider)
```
MZOLN, line 642: I'm not fully clear on this, if caching is optional, why is this logic gated?
It would be good if we could see some tests explaining the logic behind it.

iNinja: The caching is gated on three conditions: CacheCustomProviders, CacheSignatureProviders, and cacheProvider.

CacheSignatureProviders and cacheProvider are pre-existing flags that both need to be enabled for the caching to happen. In order to preserve backwards compatibility we added CacheCustomProviders defaulting to false, that when set to true will cache the custom signature providers as long as the original gates allow it.

Added tests for each of these cases, plus a concurrency test with 20 threads verifying they all get the same cached instance.

westin-m (approved):

MZOLN (approved):

cpp11nullptr:
@copilot resolve the merge conflicts in this pull request

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -115,6 +117,18 @@ public CryptoProviderFactory(CryptoProviderFactory other)
...
+        /// Gets or sets a bool controlling if <see cref="SignatureProvider"/> instances created by <see cref="CustomCryptoProvider"/>
+        /// should be cached. Default is <see langword="false"/>.
+        /// </summary>
+        /// <remarks>
+        /// When <see langword="true"/>, signature providers returned by <see cref="ICryptoProvider.Create(string, object[])"/>
+        /// are cached using the same <see cref="CryptoProviderCache"/> used for built-in providers. This avoids
+        /// repeated provider creation and key materialisation on every signature validation.
+        /// </remarks>
+        [DefaultValue(false)]
+        public bool CacheCustomProviders { get; set; }
```
pmaytak, line 130: In what cases would we set it to true? In our higher level SDKs?

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
pmaytak, line 1: >  1. 🐛 ReleaseSignatureProvider will call CustomCryptoProvider.Release() on cached providers
> 
>   Line 901–903: When a cached custom provider is released, ReleaseSignatureProvider still calls CustomCryptoProvider.Release(signatureProvider) unconditionally (because IsSupportedAlgorithm returns true). This means the custom crypto provider may dispose/release the provider even though it's still in the cache and referenced by other callers.
> 
>   You need a guard like:
> 
>    if (CustomCryptoProvider != null && CustomCryptoProvider.IsSupportedAlgorithm(signatureProvider.Algorithm))
>    {
>        if (!signatureProvider.IsCached)
>            CustomCryptoProvider.Release(signatureProvider);
>    }

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
cpp11nullptr: The primary scenario is higher-level SDKs (e.g. MSAL, Microsoft.Identity.Web) or any application that plugs in a CustomCryptoProvider for algorithms like ML-DSA / PQC. Without caching, every ValidateSignature call goes through ICryptoProvider.Create() again - allocating a new provider and re-materialising keys. Setting CacheCustomProviders = true opts custom providers into the same InMemoryCryptoProviderCache path that built-in providers already use, so repeated validations with the same key+algorithm hit the cache instead.

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
cpp11nullptr: Thanks, done.

cpp11nullptr (approved):

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
```diff
@@ -598,6 +612,23 @@ private SignatureProvider CreateSignatureProvider(
             SignatureProvider signatureProvider;
             if (CustomCryptoProvider != null && CustomCryptoProvider.IsSupportedAlgorithm(algorithm, key, willCreateSignatures))
             {
+                if (CacheCustomProviders && CacheSignatureProviders && cacheProvider)
+                {
+                    // Try cache lookup first using the remembered provider type from a previous Create call.
+                    string cacheTypeKey = CustomCryptoProvider.GetType().ToString() + "-" + algorithm;
```
pmaytak, line 618: > 2. ⚠️ _customProviderTypeCache is static but CacheCustomProviders is per-instance
> 
> The _customProviderTypeCIsache dictionary is static, so type mappings leak across all CryptoProviderFactory instances. If one factory uses CustomCryptoProviderA for algorithm X and another uses CustomCryptoProviderB for the same algorithm, the cached type from the first pollutes the second. The key includes CustomCryptoProvider.GetType() which helps if they're different types, but if two instances share the same ICryptoProvider type with different behavior, this is a problem.
> 
> Consider making it instance-scoped or including a provider identity in the key.

> Yes, exactly. The scenario you're describing is real. Here's how it plays out:
> 
>  1. FactoryA has CustomCryptoProvider = providerA (type MyCryptoProvider)
>  2. FactoryB has CustomCryptoProvider = providerB (also type MyCryptoProvider, different instance/config)
>  3. Both compute the same cacheTypeKey: "MyCryptoProvider-RS256"
>  4. providerA.Create(...) runs first, caches the provider type in the static _customProviderTypeCache
>  5. providerB now hits the type cache, gets the same providerType string, and uses it for a CryptoProviderCache lookup
> 
> Now — the CryptoProviderCache lookup itself is per-instance (each factory has its own InMemoryCryptoProviderCache), so FactoryB won't return FactoryA's cached SignatureProvider. That part is safe.
> 
> The real issue is subtler: the _customProviderTypeCache assumes that a given ICryptoProvider type + algorithm always produces the same SignatureProvider type. If providerA.Create() returns CustomRsaSignatureProvider but providerB.Create() returns DifferentRsaSignatureProvider — same ICryptoProvider type, different output type — then FactoryB will look up the wrong provider type in its cache, miss, fall through to Create(), and still work correctly. So it's a missed cache hit rather than a wrong cache hit.
> 
> The worst case I can construct: if both produce the same SignatureProvider type but with different internal state (e.g., different key material), and both factories share the same CryptoProviderCache (which doesn't happen with the copy constructor since line 93 creates a new one — but could happen if someone sets CryptoProviderCache manually to a shared instance), then you'd get a cross-contamination bug.
> 
> So practically: the static type cache is a smell but not a bug in the common case, because the per-instance CryptoProviderCache isolates the actual providers. The risk is edge-case and requires deliberate sharing of the cache instance. Worth documenting the assumption though — or just making _customProviderTypeCache instance-scoped to eliminate the question entirely.

> Make it instance. The static buys almost nothing — it's just a small string→string map that avoids one redundant Create() call per algorithm per factory lifetime. That's negligible savings for the shared-state complexity it introduces.
> 
> Instance-level means:
> 
>  - No cross-factory pollution possible
>  - Simpler correctness reasoning
>  - Trivial memory cost (handful of entries)
>  - Each factory "warms up" independently — one extra Create() per algorithm on first access, then cached
> 
> The alternative (changing the key to include something instance-specific) defeats the purpose of sharing it statically in the first place. If you're keying per-instance, just put it on the instance.
>

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -1271,6 +1271,272 @@ private static IList<SignatureProvider> CreateVerifyingProviders(CryptoProviderF
...
+            string internalId = signingKey.InternalId;
+
+            Assert.True(internalId.Length > 0,
+                $"Key InternalId should not be empty, got: '{internalId}'");
+            Assert.True(provider1.IsCached,
+                $"Provider should be marked as cached. IsCached={provider1.IsCached}");
+
+            // Try to retrieve directly from the cache to isolate the issue
+            // Wait briefly for the async event queue
+            System.Threading.Thread.Sleep(200);
```
pmaytak: nit:

>  5. 🧹 Test cleanup: Thread.Sleep(200) in test
> 
>   The Thread.Sleep(200) in CacheCustomProviders_WhenEnabled_CachesProviderFromCustomCryptoProvider is a code smell — TryAdd to InMemoryCryptoProviderCache is synchronous, so the sleep shouldn't be needed. If it was added to debug a problem, remove it now that the test passes.
>

src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs
pmaytak: This also fixes this:

> 3. ⚠️ Race condition: double-create before type is cached
> 
> On first access with N concurrent threads, all N will miss the _customProviderTypeCache lookup and call CustomCryptoProvider.Create(...). Each then calls TryAdd to the cache — only one wins, the rest are created but never cached (and never released via CustomCryptoProvider.Release). The providers that lose the TryAdd race leak.

test/Microsoft.IdentityModel.Tokens.Tests/CryptoProviderFactoryTests.cs
```diff
@@ -1271,6 +1271,343 @@ private static IList<SignatureProvider> CreateVerifyingProviders(CryptoProviderF
...
+
```
pmaytak:
```suggestion
```

> 🧹 Test cleanup: Thread.Sleep(200) in test
> The Thread.Sleep(200) in CacheCustomProviders_WhenEnabled_CachesProviderFromCustomCryptoProvider is a code smell — TryAdd to InMemoryCryptoProviderCache is synchronous, so the sleep shouldn't be needed. If it was added to debug a problem, remove it now that the test passes.

pmaytak (approved):

cpp11nullptr (approved):

PR #3490, opened 2026-05-15
Bound telemetry algorithm dimension cardinality

iNinja:
Constrains the algorithm dimension in crypto telemetry to a known allowlist, replacing unbounded user-supplied values with a fixed placeholder. This prevents high-cardinality dimension explosion in telemetry pipelines.

## Changes
- Algorithm values are validated against a known set before being emitted as telemetry dimensions
- Unknown algorithms are recorded as a constant placeholder value
- Updated telemetry tests


pmaytak (approved):

cpp11nullptr (approved):

adcosta118:
Can's we expose to the developers a fixed enum including unknown, instead relying on them providing arbitrary telemetry labels?

src/Microsoft.IdentityModel.Tokens/Telemetry/TelemetryConstants.cs
```diff
@@ -128,6 +128,20 @@ public static class Protocols
             public const string ConfigurationSourceUnknown = "Unknown";
         }
 
+        /// <summary>
+        /// Algorithm family constants for telemetry. Raw algorithm strings are mapped to these
+        /// bounded values to prevent cardinality explosion in metric dimensions.
+        /// </summary>
+        public static class AlgorithmFamilies
+        {
+            public const string RSA = "RSA";
```
adcosta118, line 137: Should we be a bit more granular to allow distinction with other RSA algos (e.g.: RSA SHA 512).

iNinja: We decided to constraint the algorithm families to prevent cardinality explosion. We track both the family and the key. It gives us a good picture of what is currently being used.

PR #3492, opened 2026-05-15
Defensive copy of collections in ValidationParameters

iNinja:
Ensures the copy constructor produces independent collection instances so that modifications to the copy do not affect the original.
Also avoids growing the source collection when merging configuration keys during content encryption key resolution.


pmaytak (approved):

cpp11nullptr (approved):

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -121,10 +121,7 @@ internal ValidationResult<string, ValidationError> DecryptToken(
 
                     if (keys != null)
                     {
-                        if (keys is List<SecurityKey> keysList)
-                            keysList.AddRange(configurationKeys);
-                        else
-                            keys = keys.Concat(configurationKeys).ToList();
+                        keys = keys.Concat(configurationKeys).ToList();
```
adcosta118, line 124: I suppose the previous version was an optimization to reduce allocations? Do we want to revert it?

adcosta118 (approved):

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
iNinja: The previous AddRange path mutated the caller's list in-place, which is what this PR aims to prevent. The extra allocation is intentional to create the defensive copy.

PR #3494, opened 2026-05-15
Disable automatic redirects on default HttpClient for JKU retrieval

iNinja:
Configures the default internal HttpClient to not follow HTTP redirects when retrieving JKU keys.
Consumers who require different behaviour can supply their own client via `SignedHttpRequestValidationParameters.HttpClientProvider`.


pmaytak (approved):

cpp11nullptr (approved):

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -36,14 +36,21 @@ public class SignedHttpRequestHandler
...
         private readonly Uri _baseUriHelper = new Uri("http://localhost", UriKind.Absolute);
-        internal readonly HttpClient _defaultHttpClient = new HttpClient();
+
+        // Redirects are disabled on the default client. Consumers who need
+        // different behaviour can supply their own client via
+        // SignedHttpRequestValidationParameters.HttpClientProvider.
+        internal readonly HttpClient _defaultHttpClient = new HttpClient(
+            new HttpClientHandler { AllowAutoRedirect = false })
+        {
+            Timeout = TimeSpan.FromSeconds(10)
```
adcosta118, line 46: Is this the same value we get from `new HttpClient()`, or is it an intentional change?

iNinja: I believe it is the same.

PR #3504, opened 2026-05-27
DPoP: fail-closed when replay protection isn't configured; add FailureType categorization

MZOLN:
- Add DPoPValidationOptions.ReplayProtectionHandledExternally and fail closed when both ExpectedNonce and JtiReplayCache are null without it
 - Introduce DPoPValidationFailureType and DPoPValidationError
 - Improve code docs


src/Microsoft.IdentityModel.Dpop/DpopValidationError.cs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.IdentityModel.Dpop;
```
pmaytak, line 6: nit casing is a different in namespace, project name, "Dpop", "DPoP", in the case this should be consistent before release.

src/Microsoft.IdentityModel.Dpop/DPoPValidationOptions.cs
```diff
@@ -96,19 +96,67 @@ public int ClockSkewInSeconds
     /// <summary>
     /// Gets or sets the expected nonce value (RP-provided).
     /// When non-null, the DPoP proof must contain a matching <c>nonce</c> claim.
-    /// When null, nonce validation is skipped.
-    /// Default: null.
     /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This is one of the two replay-protection mechanisms surfaced by Wilson. When
```
pmaytak: Use IdentityModel instead of Wilson.

bgavrilMS (approved):

pmaytak (approved):

PR #3506, opened 2026-05-28
Apply RFC 3986 section 6.2.2 normalization to DPoP htu comparison

MZOLN:


bgavrilMS (approved):

src/Microsoft.IdentityModel.Dpop/PopUriComparer.cs
```diff
@@ -0,0 +1,76 @@
...
+/// <para>
+/// Per RFC 3986 §6.2.2.1, the scheme and host components are compared case-insensitively
+/// (and host is additionally subject to IDN normalization). The port is normalized to the
+/// scheme's default when omitted (§3.2.3). The path component is compared
+/// <strong>case-sensitively</strong> (§3.3 reserves case as significant). Percent-encoded
+/// triplets of unreserved characters are decoded for comparison (§6.2.2.2). Query and
+/// fragment are excluded per RFC 9449 §4.3.
+/// </para>
+/// </remarks>
+internal static class PopUriComparer
```
pmaytak: Should this be DpopUriComprarer since it applies to Dpop htu and not all Pop?

test/Microsoft.IdentityModel.Dpop.Tests/DpopProofValidatorTests.cs
```diff
@@ -521,6 +521,40 @@ public async Task ValidateAsync_HtuIgnoresQueryString_Succeeds()
             Assert.True(result.IsValid);
         }
 
+        // RFC 9449 §4.3 + RFC 3986 §6.2.2: htu comparison must normalize scheme/host
+        // case-insensitively but keep path comparison case-sensitive. Each case below pins
+        // a specific contract; see PopUriComparer for the full contract documentation.
+        [Theory]
```
pmaytak, line 527: Any of there test cases worth adding?

>  Missing/worth adding:
> 
>    1. Query string ignored — RFC 9449 §4.3 explicitly excludes query. A case like ("https://host/path?q=1", "https://host/path", true) would pin this.
>    2. HTTP default port (80) — Only HTTPS/443 is tested. ("http://host:80/path", "http://host/path", true) ensures the non-default-scheme port works too.
>    3. Dot-segment removal — ("https://host/a/../b", "https://host/b", true) proves path normalization per §6.2.2.3.
>    4. Percent-encoding of reserved chars stays encoded — ("https://host/%2F", "https://host//", false) ensures %2F (reserved /) is NOT decoded, which is the flip side of case 9.
>    5. Empty path equivalence — ("https://host", "https://host/", true) — both should normalize to /.

pmaytak (approved):

test/Microsoft.IdentityModel.Dpop.Tests/DpopProofValidatorTests.cs
MZOLN: Sure let's add them all, even though we're not supposed to support http.

src/Microsoft.IdentityModel.Dpop/PopUriComparer.cs
MZOLN: Sure Im ok with that.

test/Microsoft.IdentityModel.Dpop.Tests/DpopProofValidatorTests.cs
MZOLN: Added

src/Microsoft.IdentityModel.Dpop/PopUriComparer.cs
MZOLN: Renamed to UriComparer, but we will move it to a better location as need appears.

pmaytak (approved):

kllysng (approved):

PR #3502, opened 2026-05-23
Coalesce concurrent fetches

MZOLN:
Coalesce concurrent fetches on cold run.
Expose async handler for issuer validation.

PR #3516, opened 2026-06-02
Fix race condition in ConfigurationManager refresh state (#3142)

### LINKED ISSUE #3142: [Bug] Ensure updates to state are done atomically in ConfigurationManager
Note that to get help, you need to run the latest version. 
Microsoft.IdentityModel 8.5.0

**Where is the issue?**
 * [ ] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [x] M.IM.Protocols
 * [ ] M.IM.TestExtensions
 * [ ] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [ ] S.IM.Tokens.Jwt
 * Other (please describe)

**Expected behavior**

Ensure updates to state are done atomically in ConfigurationManager. 

See #3090, these changes showed a viable way to handle this. It would need small changes to work with 8.5.0

**Actual behavior**

_syncAfter and _lastRequestRefresh do not have guaranteed atomic updates.

dipeshchaudhari3255:
# Fix race condition in ConfigurationManager refresh state (#3142)

* [x] You've read the Contributor Guide and Code of Conduct.
* [ ] You've included unit or integration tests for your change, where applicable.
* [x] You've included inline docs for your change, where applicable.
* [ ] If any gains or losses in performance are possible, you've included benchmarks for your changes.
* [x] There's an open issue for the PR that you are making.

Summary of the changes

Fix race condition caused by non-atomic refresh timestamp access.

## Description

Issue #3142 reports a thread-safety concern around concurrent access to refresh timing state in `ConfigurationManager`.

The fields `_syncAfter` and `_lastRequestRefresh` were used by multiple threads to determine whether a configuration refresh was required. These values were accessed concurrently without atomic read/write operations, which could result in inconsistent observations under heavy concurrency and lead to duplicate metadata retrievals.

### Changes

* Store `_syncAfter` as `long` ticks.
* Store `_lastRequestRefresh` as `long` ticks.
* Replace direct reads with `Interlocked.Read`.
* Replace direct writes with `Interlocked.Exchange`.
* Add inline comments documenting the thread-safety rationale and the relationship to Issue #3142.

### Validation

The issue was reproduced using a test application that generated 50 concurrent refresh requests.

Before the fix:

* Request count: 50
* Actual metadata HTTP calls: 2

After the fix:

* Request count: 50
* Actual metadata HTTP calls: 1

This confirms that concurrent refresh requests no longer trigger multiple simultaneous metadata retrievals.

Fixes #3142

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -21,9 +21,11 @@ namespace Microsoft.IdentityModel.Protocols
     {
         internal Action _onBackgroundTaskFinish;
 
-        private DateTimeOffset _syncAfter = DateTimeOffset.MinValue;
-        private DateTimeOffset _lastRequestRefresh = DateTimeOffset.MinValue;
-        private bool _isFirstRefreshRequest = true;
+        // Stored as ticks so reads/writes can be performed atomically via Interlocked.
+        // Prevents torn reads/writes under concurrent access (Issue #3142).
+        private long _syncAfter = DateTimeOffset.MinValue.Ticks;
+        private long _lastRequestRefresh = DateTimeOffset.MinValue.Ticks;
+        private int _isFirstRefreshRequest = 1;
         private readonly SemaphoreSlim _configurationNullLock = new SemaphoreSlim(1);
 
         private readonly IDocumentRetriever _docRetriever;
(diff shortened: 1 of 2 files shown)


MZOLN:
This seems to overlap with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3502

dipeshchaudhari3255:
@MZOLN
Thanks for pointing that out.

I reviewed #3502  and it appears to focus on coalescing concurrent callers during the initial configuration retrieval path when `_currentConfiguration` is null by introducing a shared in-flight fetch task.

This PR addresses Issue #3142 specifically by making access to `_syncAfter` and `_lastRequestRefresh` atomic using `Interlocked.Read` and `Interlocked.Exchange`. The goal is to prevent concurrent callers from observing inconsistent refresh timestamp state during refresh scheduling and triggering duplicate metadata refreshes.

The two changes appear related in that they both reduce duplicate metadata retrievals, but they address different code paths and different concurrency issues.

PR #3526, opened 2026-06-16
Make Signed HTTP Request `p` claim (path) comparison case-sensitive (RFC 3986 section 3.3)

debchoudhury-id4s:
## What

Make the Signed HTTP Request `p` (path) claim comparison case-sensitive, per RFC 3986 section 3.3. A new opt-out AppContext switch, `Switch.Microsoft.IdentityModel.SignedHttpRequest.UseCaseSensitivePClaimComparison`, restores the old behavior.

## Why

1. URI paths are case-sensitive, but the `p` claim was compared case-insensitively, so a proof for `/Admin/resource` was accepted against a request to `/admin/resource`. 
2. The fix also makes percent-encoding hex case significant (`%2f` no longer equals `%2F`). This matches DPoP, which already compares paths case-sensitively.

## Compatibility

1. 9.x (`dev`): case-sensitive by default; set the switch to `false` to opt out.
3. 8.x backport: defaults to `false` (opt-in), so it stays non-breaking.
4. Only callers whose signed path casing differs from the request path are affected; identical casing still matches.
5. The `m` (method) and `u` (host) claims are unchanged (case-insensitive by spec).

## Tests 

1. cover case mismatches in both directions, `%2f` vs `%2F`, and both switch states.


src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -822,7 +822,8 @@ internal virtual void ValidatePClaim(JsonWebToken signedHttpRequest, SignedHttpR
             pClaimValue = pClaimValue.Trim('/');
             var expectedPClaimValue = httpRequestUri.AbsolutePath.Trim('/');
 
-            if (!string.Equals(expectedPClaimValue, pClaimValue, StringComparison.OrdinalIgnoreCase))
+            // URI path components are case-sensitive per RFC 3986 section 3.3, so the comparison must be ordinal (case-sensitive).
+            if (!string.Equals(expectedPClaimValue, pClaimValue, StringComparison.Ordinal))
```
robezirg: this is a breaking change. how will this affect existing users?

kllysng: @robezirg was this fix not the object of the bug? I think it's reasonable as part of the breaking changes going into dev, aka 9x.

kllysng (approved):

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: resolving with latest revision that incorporates flag approach

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 See the [releases](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/releases) for details on bug fixes and added features.
 
+9.0.0
```
kllysng, line 3: So cool that you found this proactively. Consider renaming to something like:

```suggestion
Planned for 9.0.0
```
this is because 9.0.0 isn't out yet.

kllysng (approved):

CHANGELOG.md
debchoudhury-id4s: Thanks @kllysng for the tip. Curious if I add `Planned for`, wouldn't it require a follow up PR for the removal later? Is this a general approach for such Changelog updates around upcoming major version plans? If not a hard req, would consider keeping as is.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
...
+====
+## Bug Fixes
+- Make Signed HTTP Request `p` (path) claim comparison case-sensitive per RFC 3986 section 3.3, behind an AppContext switch. Previously `/Admin/resource` matched `/admin/resource`. See [PR #3526](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3526).
```
robezirg: I don't think this is a bug fix, this is an option to make it case sensitive as per rfc

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -106,6 +106,20 @@ internal static class AppContextSwitches
         private static bool? _successValidationLogsAsInformation;
         internal static bool SuccessValidationLogsAsInformation => _successValidationLogsAsInformation ??= (AppContext.TryGetSwitch(SuccessValidationLogsAsInformationSwitch, out bool successLogsAsInfo) && successLogsAsInfo);
 
+        /// <summary>
+        /// Controls whether the Signed HTTP Request 'p' (path) claim is compared case-sensitively (ordinal) per RFC 3986 section 3.3.
+        /// On this line the comparison is case-sensitive by default; set the switch to <c>false</c> to restore the legacy case-insensitive (OrdinalIgnoreCase) comparison.
+        /// </summary>
+        internal const string UseCaseSensitivePClaimComparisonSwitch = "Switch.Microsoft.IdentityModel.SignedHttpRequest.UseCaseSensitivePClaimComparison";
```
robezirg, line 113: i don't think this is the correct path. I am not in favor of app context swithch, this should be more of shr validation setting

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
```diff
@@ -337,6 +337,45 @@ public void ValidatePClaim(ValidateSignedHttpRequestTheoryData theoryData)
             TestUtilities.AssertFailIfErrors(context);
         }
 
+        [Fact]
+        [ResetAppContextSwitches]
```
pmaytak, line 342: Add the same attribute on line 322

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
pmaytak: Context switches are more of a fallback from the new default behavior; parameters/options flag is more for something that's a public API where all flag settings are valid.

debchoudhury-id4s: @robezirg & @pmaytak  :  we already ship secure-by-default switches the same way. Please check [UseClaimsIdentityType](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/168714d244b2d8f27b1d5b769c3480ccb0a87da0/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs#L30) (case-sensitive identity by default) and [DontFailOnMissingTid](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/168714d244b2d8f27b1d5b769c3480ccb0a87da0/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs#L39)

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
debchoudhury-id4s: Done

CHANGELOG.md
debchoudhury-id4s: reworded to lean on RFC guidance

pmaytak (approved):

MZOLN (approved):

pmaytak (approved):

dlingam-msft (approved):

RojaEnnam (approved):

debchoudhury-id4s:
Superseded by #3539 — recreated as an in-org (non-fork) PR so the required build-validation pipeline runs (it does not run on PRs from forks). Closing this one in favor of #3539.

PR #3530, opened 2026-06-22
Add JsonWebToken header-replacement constructor to avoid re-parsing payload

### LINKED ISSUE #3529: Add `JsonWebToken` constructor to reuse a parsed payload when only the header changes
## Summary

Add a public `JsonWebToken` constructor that produces a new token by replacing only the encoded header of an existing `JsonWebToken`, reusing its already-parsed payload (and signature) instead of re-parsing the full token from a string.

## Motivation and goals

Scenarios that rewrite **only the JWT header** and then need a `JsonWebToken` for downstream signature validation currently have no way to avoid re-parsing the payload. The typical pattern is:

```csharp
// change a header value (e.g. hash a nonce), then reassemble "header.payload.signature":
string newEncodedHeader = TransformHeader(source);
string reassembled = newEncodedHeader + "." + source.EncodedPayload + "." + source.EncodedSignature;
var jwt = handler.ReadJsonWebToken(reassembled); // re-parses header AND payload
```

The payload bytes are byte-for-byte identical to the source token, yet `ReadJsonWebToken` rebuilds the entire payload claim set (`CreatePayloadClaimSet` -> dictionary + per-claim objects). On large tokens this is the dominant cost of the rewrite.

Concrete motivating case: **Protected Forwarded Token (PFT)** validation in Microsoft Identity Service Essentials (MISE), used by the Microsoft Graph gateway. PFT validation hashes a nonce in the header, then must hand a `JsonWebToken` to signature validation. Today that re-parses the whole payload per request.

**Isolated microbenchmark** (Wilson `Microsoft.IdentityModel.Benchmarks`, net8.0, MediumRun, MemoryDiagnoser) measuring just the header-replacement operation: reassemble-and-reparse vs the proposed reuse-payload constructor.

| Token size | Re-parse | Reuse-payload ctor | Time | Allocation |
|---|---|---|---|---|
| Minimal ~0.7 KB | 2,549 ns / 2.63 KB | 601 ns / 2.26 KB | **-76% (4.2x faster)** | **-14%** |
| Small ~1.9 KB (+50 claims) | 16,758 ns / 11.25 KB | 818 ns / 4.81 KB | **-95% (20x faster)** | **-57%** |
| Typical ~4 KB (+125 claims) | 47,911 ns / 24.30 KB | 1,079 ns / 8.85 KB | **-98% (44x faster)** | **-64%** |

The divergence is the point: re-parse cost grows steeply with payload (2.5 us -> 47.9 us), while the reuse-payload path barely moves (601 ns -> 1,079 ns) because only the small header is re-parsed. At a realistic ~4 KB Graph token the constructor is ~44x faster and allocates 64% less.

**End-to-end** (MISE Bearer-PFT validation, realistic ~4 KB Graph user token; BenchmarkDotNet LongRun, .NET 9.0.16, N ~= 280, StdErr <= 0.13%) shows the real-world impact after the fixed cost of signature verification and the validation pipeline:

| Metric | Re-parse (`ReadJsonWebToken`) | Reuse-payload ctor | Improvement |
|---|---|---|---|
| Mean latency | 75.55 us | 62.35 us | **-17.5%** |
| Allocated | 59.29 KB | 46.33 KB | **-21.9%** |

The win scales with payload size, because the eliminated work is the payload parse while signature verification and the rest of the pipeline are fixed cost.

Goals:
- Provide an allocation- and CPU-efficient way to construct a header-swapped `JsonWebToken` that reuses an existing parsed payload.
- Keep it purely additive and backward compatible.
- Preserve byte-identical token semantics: the resulting `EncodedToken` and signature must be exactly what the reassemble-and-reparse path produces.

## In scope

- A public constructor on `JsonWebToken` taking an existing `JsonWebToken` plus a replacement encoded (Base64Url) header.
- JWS (signed) source tokens: the common case for header rewrites.
- Reusing the source token's parsed payload claim set and its encoded signature segment.

## Out of scope

- JWE (encrypted) source tokens: the plaintext payload is not available to reuse; these should be rejected.
- Rewriting the payload (only the header changes in the target scenarios).
- `JwtSecurityToken` (legacy handler): the motivating consumers use `JsonWebToken`.
- A `ReadOnlySpan<char>` header overload: possible additive follow-up if profiling warrants; not required initially.

## Risks / unknowns

- **Misuse:** a caller could pass a header whose `alg`/`kid` no longer matches what the signature was computed over. This constructor does not validate, same "decoded, not validated" contract as existing `JsonWebToken` constructors; validation remains the handler's job. Documentation should state this clearly.
- **JWE guard:** must fail fast with a clear message when the source is encrypted (payload reuse is invalid).
- **API review:** this is new public surface; `PublicAPI.Unshipped.txt` updated accordingly.

## Examples

```csharp
// Existing parsed token (already validated-or-not, just decoded).
JsonWebToken source = handler.ReadJsonWebToken(incoming);

// Rewrite only the header (e.g., replace a nonce value), Base64Url-encode it.
string newEncodedHeader = BuildTransformedHeader(source);

// New: reuse the parsed payload + signature; only the header is decoded/parsed.
var transformed = new JsonWebToken(source, newEncodedHeader);

// Equivalent today (but re-parses the whole payload):
// var transformed = handler.ReadJsonWebToken(newEncodedHeader + "." + source.EncodedPayload + "." + source.EncodedSignature);

// Downstream signature validation is unchanged and still passes.
```

shankar-shubh:
# Add JsonWebToken header-replacement constructor to avoid re-parsing payload

Add a JsonWebToken constructor that reuses a parsed payload on header replacement

## Description

Adds a public constructor `JsonWebToken(JsonWebToken jsonWebToken, string encodedHeader)` that builds a token by replacing the header of an existing JWS while reusing its already-parsed payload and signature. The token is formed as `encodedHeader.EncodedPayload.EncodedSignature`; only the replacement header is decoded and parsed, so the payload claim set is reused instead of being parsed a second time.

This supports header-only transforms (for example PFT nonce hashing) where callers previously had to reassemble the token string and fully re-read it via `ReadJsonWebToken`, incurring a redundant payload parse and allocation.

Encrypted (JWE) source tokens are rejected with a new `IDX14117` message because their plaintext payload is not available for reuse.

### Changes

- New public constructor on `JsonWebToken` (additive; existing parse paths unchanged).
- New `IDX14117` log message for the JWE-source rejection.
- `PublicAPI.Unshipped.txt` updated for the new public API.
- Unit tests: header replacement, payload/signature reuse, equivalence with a reparsed token, signature validation, null/empty guards, and the JWE rejection path.
- `JsonWebTokenHeaderReplacementBenchmarks` isolating the new constructor against reassemble-and-reparse.

### Benchmarks

Isolated microbenchmark (`Microsoft.IdentityModel.Benchmarks`, net8.0, MediumRun, MemoryDiagnoser) — reassemble-and-reparse vs. the new constructor:

| Token size | Re-parse | Reuse-payload ctor | Time | Allocation |
|---|---|---|---|---|
| Minimal ~0.7 KB | 2,549 ns / 2.63 KB | 601 ns / 2.26 KB | -76% (4.2x) | -14% |
| Small ~1.9 KB | 16,758 ns / 11.25 KB | 818 ns / 4.81 KB | -95% (20x) | -57% |
| Typical ~4 KB | 47,911 ns / 24.30 KB | 1,079 ns / 8.85 KB | -98% (44x) | -64% |

The eliminated payload parse scales with payload size, so the win grows with the token.

Fixes #3529


shankar-shubh:

iNinja:
Nice perf win. One thought on the public surface: the new ctor reuses the original signature with a swapped header. That's exactly right for the PFT nonce-hashing case the description mentions (the new header reconstructs the originally-signed bytes), but as a general public ctor it lets any caller swap a header and silently keep an invalid signature.

Could we expose this through `JsonWebTokenHandler` instead, e.g.:

    public virtual JsonWebToken ReplaceTokenHeader(JsonWebToken jsonWebToken, string encodedHeader)

Same payload-reuse perf, but it sits next to `ReadJsonWebToken`, reads as an intentional transform rather than a general-purpose ctor, and is mockable. The XML doc should note the signature is preserved as-is and only validates when the header reconstructs the originally-signed form (as in PFT). Suggest dropping the ctor from the public API and adding the handler method.

shankar-shubh:
@iNinja Have updated the PR as per the comment.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -233,6 +233,48 @@ public JsonWebToken(
...
+                throw LogHelper.LogExceptionMessage(new ArgumentException(LogMessages.IDX14117, nameof(jsonWebToken)));
+
+            string encodedPayload = jsonWebToken.EncodedPayload;
+            string encodedSignature = jsonWebToken.EncodedSignature;
+            string encodedToken = encodedHeader + "." + encodedPayload + "." + encodedSignature;
+
+            Dot1 = encodedHeader.Length;
+            Dot2 = Dot1 + 1 + encodedPayload.Length;
+            Dot3 = -1;
+            IsSigned = encodedSignature.Length > 0;
```
MZOLN, line 289: This would be faster and would reduce allocations further

int totalLength = encodedHeader.Length + 1 + encodedPayload.Length + 1 + encodedSignature.Length;
string encodedToken = string.Create(
    totalLength,
    (encodedHeader, encodedPayload, encodedSignature),
    static (span, state) =>
    {
        var (header, payload, signature) = state;
        header.AsSpan().CopyTo(span);
        int pos = header.Length;
        span[pos++] = '.';
        payload.AsSpan().CopyTo(span.Slice(pos));
        pos += payload.Length;
        span[pos++] = '.';
        signature.AsSpan().CopyTo(span.Slice(pos));
    });

MZOLN (commented):
Left some comments

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
shankar-shubh: I considered using Span but string.Create is only available post .NET Core 2.1 and Wilson supports older frameworks. 

So this will need to be behind guards if we want to use Spans, let me know if that is fine.

shankar-shubh: I see string.Create is guarded by NET6_0_OR_GREATER in the repo, so have updated as per suggestion and included the TFM guard.

MZOLN (approved):

iNinja (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebToken.cs
```diff
@@ -233,6 +233,67 @@ public JsonWebToken(
...
+        /// The already parsed payload of <paramref name="jsonWebToken"/> is reused, so the payload is not decoded or parsed again; only the replacement header is decoded and parsed.
+        /// The signature is reused verbatim, so it only remains valid when <paramref name="encodedHeader"/> reconstructs the originally signed header.
+        /// <para>
+        /// The contents of the returned <see cref="JsonWebToken"/> have not been validated, the JSON Web Token is simply decoded. Validation can be accomplished using the validation methods in <see cref="JsonWebTokenHandler"/>
+        /// </para>
+        /// </remarks>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="jsonWebToken"/> is null.</exception>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="encodedHeader"/> is null or empty.</exception>
+        /// <exception cref="ArgumentException">Thrown if <paramref name="jsonWebToken"/> is an encrypted token (JWE), as its payload cannot be reused.</exception>
```
bgavrilMS, line 252: nit: it's not clear why JWEs aren't supported

bgavrilMS (approved):

MZOLN (approved):

rishiprakash-microsoft:
We can close this one? as @MZOLN merged the changes to dev branch - fork ones were stuck on Wilson-PR check.

PR #3538, opened 2026-07-08
POC: Synchronous cache-hit validation fast path; make ValidateTokenAsync non-async (#3459)

### LINKED ISSUE #3459: POC Inconsistent API surface regarding sync vs async methods
Demonstrate a consistent validation API (clear sync vs async) without regressing functionality

•	Refactor internal call graph to avoid unnecessary async state machines in CPU-bound paths
•	Update call sites in a small sample app/test harness to validate 

acceptance criteria
•	Benchmark shows fewer allocations vs. baseline
•	API review sign-off on rule consistency

debchoudhury-id4s:
## Overview

POC for #3459. On the configuration **cache-hit** path, `JsonWebTokenHandler.ValidateTokenAsync` does not need to run an async state machine. This PR adds a synchronous cache-hit fast path and — importantly — makes `ValidateTokenAsync` itself non-`async` so the saving is actually observable.

The shipped `TokenValidationParameters` path is unchanged, and behaviour on the cache-miss path is unchanged.

> **Scope note.** Per @westin-m's comment on #3459 (2026-07-14), a public synchronous `Validate`/`ValidateToken` contract is reserved for @Toidi357's synchronous-validation project and is **out of scope** for this POC. The public synchronous `ValidateToken` overloads are nevertheless still present in this branch, deliberately, so the cost of a public synchronous contract could be measured directly rather than estimated (point **B** below). **They are ready to be removed on request** — the internal fast path and every measurement below are unaffected by their removal. See "Relationship to the synchronous-validation project".

## Why `ValidateTokenAsync` had to become non-`async`

If the fast path is placed inside a method that stays `async`, the C# compiler still emits the async state-machine box and a `Task`, so the allocation saving disappears and the benchmark shows nothing. Declaring the wrapper non-`async` and returning `Task.FromResult(...)` on the fast path is what makes the gain real.

## Changes

- `BaseConfigurationManager.TryGetCurrentConfiguration(out BaseConfiguration?)` — synchronous configuration peek. Base returns `false`; `ConfigurationManager<T>` overrides it and reuses the async fast-path predicate so the two cannot drift.
- `ISynchronousIssuerValidator`, `Validators.ValidateIssuer` / `ValidateIssuerInternal` — synchronous issuer validation against the peeked configuration, implemented by `DefaultIssuerValidator`. This also removes the repeated configuration lookup during issuer validation.
- `JsonWebTokenHandler.ValidateTokenAsync` (both overloads) — **converted from `async` to non-`async` `Task`-returning methods**. The original asynchronous body moved to `ValidateTokenCoreAsync`; `AddStackFrameOnFailureAsync` handles the deferred path.
- `TryValidateTokenSynchronously(...)` — single shared implementation of the cache-hit fast path, used by both `ValidateTokenAsync` and the public synchronous `ValidateToken`. Removed roughly 70 lines of duplicated logic.
- Token replay guard — when `ValidationParameters.TokenReplayCache` is set, the fast path declines and the call routes to the asynchronous path. `ValidateTokenReplayInternal` mutates the cache, so running it speculatively and then falling back would record the token twice and report a spurious replay.
- Benchmarks — cached-configuration benchmark pair in `ValidateTokenAsyncTests` so the async and sync cache-hit paths can be compared directly.
- `PublicAPI.Unshipped.txt`, `CHANGELOG.md`, and `ValidateTokenSyncTests` (sync-vs-async golden master plus cache-hit and lifecycle cases).

## Results

Isolated cache-hit benchmark, net9.0, BenchmarkDotNet 0.13.12, MediumRun, .NET 9.0.18. Only the final validation call and the library build differ between points; token, parameters, handler and configuration priming are identical.

| Point | What it is | Mean | StdDev | Allocated | Alloc Δ vs A |
| --- | --- | ---: | ---: | ---: | ---: |
| **A** | async baseline (`ValidateTokenAsync`, before this PR) | 27.74 µs | 0.166 µs | 7.32 KB | — |
| **B′** | fast path under the non-`async` `ValidateTokenAsync` (**this PR**) | 27.31 µs | 0.192 µs | **6.70 KB** | **−8.47 %** |
| **B** | public synchronous `ValidateToken` | 27.97 µs | 1.102 µs | 6.61 KB | −9.70 % |

**This is an allocation result, not a latency result.** The latency spread between the three points is smaller than B's own standard deviation, which is about six times B′'s. An earlier draft of the perf document claimed a 1.68 % latency improvement; that figure was not reproduced and has been withdrawn.

A separate deterministic harness (`GC.GetAllocatedBytesForCurrentThread`, 300 000 iterations, one scenario per process) independently measures **−640 B/op (−18.5 %)** for A→B′ and **−88 B/op** for B′→B, agreeing with BenchmarkDotNet to within a few bytes.

### Where the saving comes from

An allocation trace (`GCAllocationTick`, GC keyword, Verbose; sampled bytes covered 100.6 % of measured allocation) attributes essentially the entire A→B′ delta to four `Task<T>` rows — which is how the runtime reports `AsyncStateMachineBox<TResult, TStateMachine>`, since it derives from `Task<TResult>`:

| Type | A | B′ | B | Δ A→B′ |
| --- | ---: | ---: | ---: | ---: |
| `Task<BaseConfiguration>` | 219.6 | 0 | 0 | **−219.6** |
| `Task<ValidationResult<ValidatedIssuer, ValidationError>>` | 201.8 | 0 | 0 | **−201.8** |
| `Task<OpenIdConnectConfiguration>` | 160.2 | 0 | 0 | **−160.2** |
| `Task<ValidationResult<ValidatedToken, ValidationError>>` | 181.2 | 84.2 | 0 | **−97.0** |

(sampled bytes per operation; every other type moves less than the ±25 B sampling noise floor)

Those are exactly the state machines for configuration retrieval and asynchronous issuer validation — the two behaviours the fast path replaces. B′ still allocates one `Task` because it returns `Task.FromResult(...)`; the public synchronous `ValidateToken` removes that last 88 B. **So B′ already captures 640 of the 728 bytes a fully public synchronous API would capture — about 88 %.**

### Cache-miss → async fetch → sync cache-hit lifecycle

Starting from an empty configuration cache with a counting retriever: one cold validation, then 200 000 warm ones. Cold and warm reported separately, never averaged.

| Build | Retrievals after cold | Warm mean | Warm alloc/op | Total retrievals |
| --- | ---: | ---: | ---: | ---: |
| A | 1 | 23.257 µs | 3 456.04 B | 1 |
| B′ | 1 | 22.844 µs | 2 816.03 B | 1 |

The retriever is invoked exactly once; every warm call is served from the cache with no further retrieval. Covered by `ValidateTokenAsync_CacheMissThenCacheHits_RetrievesConfigurationOnce`.

## Relationship to the synchronous-validation project

| This PR (#3459) | @Toidi357's synchronous-validation project |
| --- | --- |
| Internal synchronous cache-hit fast path under `ValidateTokenAsync` | Public fully-synchronous `ValidateToken` contract |
| Synchronous configuration peek (`TryGetCurrentConfiguration`) | Synchronous behaviour on cache **miss** via true sync retrieval (`HttpClient.Send`) |
| Removing the repeated configuration lookup during issuer validation | L2 `HandleBeforeRetrieveAsync` sync replacement, including sync Redis |
| Cache-hit / MISE evidence, profiler attribution, lifecycle evidence | Dedicated metadata runner, snapshot publication, refresh coordination |
| | ThreadPool-starvation / deadlock proof |

**Known limitation.** The public synchronous `ValidateToken` still blocks (`GetAwaiter().GetResult()`) when the fast path declines — cache miss, recoverable failure needing refresh/LKG, or a configured replay cache. That sync-over-async is exactly what the synchronous-validation project resolves, and it is not addressed here. The internal fast path under `ValidateTokenAsync` has no such problem: it simply returns the pending `Task`.

## Validation

- Build clean (0 warnings, 0 errors) on all target frameworks.
- `Microsoft.IdentityModel.JsonWebTokens.Tests` — 679 passed, 1 skipped (pre-existing), 0 failed.
- `Microsoft.IdentityModel.Tokens.Tests` — 2 427 passed, 0 failed.
- `ValidateTokenSyncTests` — 22 passed, 0 failed.

Full methodology, verbatim BenchmarkDotNet output, profiler histograms and MISE results are in the perf document linked in the comments.


Toidi357:
I think this is related to my intern project. I'm also working on making a synchronous result-based validation method `ValidateToken`, but doesn't have any asynchronous fallback. Async fallback creates sync-over-async (when you do `.GetAwaiter().GetResult()` issues which I'm hoping to avoid by adding methods to `HttpDocumentRetriever`, `ConfigurationManager`, and other related classes

debchoudhury-id4s:
perf results are in [Perf Comparison synchronous validation POC.docx](https://microsoft-my.sharepoint.com/:w:/p/debchoudhury/cQou5BIUNnK0QbeMJu9vApIFEgUDmM12z7lK4LiHlViPw07dHw?isSPOFile=1&xsdata=MDV8MDJ8fDdhYWJjOTZkN2JkZDQ4ZDRmNzJmMDhkZWUxMDg0ODVmfDcyZjk4OGJmODZmMTQxYWY5MWFiMmQ3Y2QwMTFkYjQ3fDB8MHw2MzkxOTU2MjI4MjM0MjU4Mzl8VW5rbm93bnxWR1ZoYlhOVFpXTjFjbWwwZVZObGNuWnBZMlY4ZXlKRFFTSTZJbFJsWVcxelgwRlVVRk5sY25acFkyVmZVMUJQVEU5R0lpd2lWaUk2SWpBdU1DNHdNREF3SWl3aVVDSTZJbGRwYmpNeUlpd2lRVTRpT2lKUGRHaGxjaUlzSWxkVUlqb3hNWDA9fDF8TDJOb1lYUnpMekU1T20xbFpYUnBibWRmV1ZkSmQxcHFTbXBhVkVsMFQxUlZNazFUTURCYWJWVXhURmRGTkZwVVkzUlpla0pxVDFSSmVGbFVWWHBOZWxac1FIUm9jbVZoWkM1Mk1pOXRaWE56WVdkbGN5OHhOemd6T1RZMU16a3dNVEkxfDZkODNhNGVlN2UxZTQ2MmJmNzJmMDhkZWUxMDg0ODVmfDQwZDQwY2JhY2UzZTQzMTJiMDYxMjFiZDhhZWU5NDUw&sdata=RDdUTzVRVWFZTUpPV2N0bU9LVXcxd3dTbUltVjdScjMzNURYK1JFL0Npdz0%3D&ovuser=72f988bf-86f1-41af-91ab-2d7cd011db47%2Cdebchoudhury%40microsoft.com). 

TL;DR the POC is safe to be merged (not saying we should NOW). Check section 2 which does a sync vs async comparison showing how the POC reduces latency and improves perf

PR #3547, opened 2026-07-15
Reinstate Microsoft.IdentityModel.Protocols.WsTrust as a supported 8.x package

### LINKED ISSUE #3463: Deprecation of Microsoft.IdentityModel.Protocols.WsTrust when System.ServiceModel.Federation needs this
The nuget package Microsoft.IdentityModel.Protocols.WsTrust is needed by System.ServiceModel.Federation. As part of the original work when Microsoft.IdentityModel.Protocols.WsTrust was created, there was an expectation that the implementation would end up in the main development branch. Issue #1623 was opened by someone in the community asking when this would happen. The issue was closed with the recommendation to use Microsoft.IdentityModel.Protocols.WsFederation instead. Here is a summary of the findings using Copilot stating the differences between the two libraries and what's missing:

## Summary

The .NET WCF client repo ([dotnet/wcf](https://github.com/dotnet/wcf)) ships **`System.ServiceModel.Federation`**, which depends on **`Microsoft.IdentityModel.Protocols.WsTrust`** (last released v6.8.0, now deprecated). The package was used to build, send, and parse WS-Trust 1.3 / WS-Trust 1.4 / WS-Trust Feb 2005 active-profile **`RequestSecurityToken` (RST)** and **`RequestSecurityTokenResponse` (RSTR)** SOAP messages exchanged with a Secure Token Service (STS).

The suggested replacement, **[`Microsoft.IdentityModel.Protocols.WsFederation`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/tree/dev/src/Microsoft.IdentityModel.Protocols.WsFederation)** **does not provide any of the functionality WCF actually consumes**. It is a fundamentally different protocol surface:

| Concern | `Microsoft.IdentityModel.Protocols.WsTrust` (deprecated) | `Microsoft.IdentityModel.Protocols.WsFederation` (proposed replacement) |
|---|---|---|
| Profile | **WS-Trust active profile** (SOAP RST/RSTR) | WS-Federation **passive/browser** profile (HTTP query string / form POST) + federation **metadata** parsing |
| Wire format | `XmlDictionaryReader` / `XmlDictionaryWriter` reading/writing SOAP body envelopes | `application/x-www-form-urlencoded` query strings / form parameters; XML metadata documents |
| Object model for RST/RSTR | Yes – full hierarchy | **Missing entirely** – only string constants in `Microsoft.IdentityModel.Xml.WsTrustConstants[_1_3 / _1_4 / _2005]` |
| Serializer for RST/RSTR | `WsTrustSerializer.WriteRequest(XmlDictionaryWriter, …)` and `ReadResponse(XmlDictionaryReader)` | **None** |

The WsFederation package's only XML-format support is [`WsFederationMetadataSerializer`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMetadataSerializer.cs), which parses **WS-Federation metadata** documents (entity descriptors, role descriptors, signing keys). That is unrelated to the active WS-Trust `RequestSecurityToken` SOAP exchange WCF needs.

The WsFederation package's [`WsFederationMessage`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Protocols.WsFederation/WsFederationMessage.cs) derives from `AuthenticationProtocolMessage` and is built around `FromQueryString` / `FromUri` / `BuildRedirectUrl` / `BuildFormPost` for the **browser-redirect** sign-in flow. It cannot construct or parse a SOAP `<wst:RequestSecurityToken>` body.

Code search of the [AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet) repository (`dev` branch) confirms that **no class named `WsTrustRequest`, `WsTrustResponse`, or `WsTrustSerializer` exists outside the [`brentsch/wstrust`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/tree/brentsch/wstrust) feature branch.** That feature branch is the source of the deprecated package; nothing equivalent has been merged forward into `dev`.

## What WCF actually consumes from `Microsoft.IdentityModel.Protocols.WsTrust` (v6.8.0)

Below is the complete API surface used by `System.ServiceModel.Federation`. Each entry must be replaced one-for-one (or by an equivalent capability) for the migration to be feasible. Citations point to WCF source under [`src/System.ServiceModel.Federation/src/`](https://github.com/dotnet/wcf/tree/main/src/System.ServiceModel.Federation/src).

### 1. Object model — request side

Used in `WSTrustChannelSecurityTokenProvider.CreateWsTrustRequest()` and `WSTrustChannel`.

| Type / member (deprecated namespace) | Used as | Equivalent in `Microsoft.IdentityModel.Protocols.WsFederation`? |
|---|---|---|
| `Microsoft.IdentityModel.Protocols.WsTrust.WsTrustRequest` (ctor takes the `Issue` action) | The mutable RST message | **Missing** |
| `WsTrustRequest.WsTrustVersion`, `RequestType`, `KeyType`, `KeySizeInBits`, `TokenType`, `Context`, `ComputedKeyAlgorithm`, `Entropy`, `Claims`, `AppliesTo`, `AdditionalXmlElements` | All used to populate the RST body | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsTrust.WsTrustVersion` (enum: `TrustFeb2005`, `Trust13`, `Trust14`) | Selects the wire version | **Missing** (only namespace string constants exist) |
| `Microsoft.IdentityModel.Protocols.WsTrust.WsTrustActions` (instance per version with `.Issue`, `.IssueRequest`, `.Cancel`, `.CancelRequest`, `.Renew`, `.RenewRequest`, `.Validate`, `.ValidateRequest`) | Computes WS-A action | **Missing** (only `WsTrustConstants_1_3.Actions.Issue` etc. as raw strings — no Cancel/Renew/Validate constants and no version-keyed lookup object) |
| `Microsoft.IdentityModel.Protocols.WsTrust.WsSerializationContext` (`new WsSerializationContext(WsTrustVersion)`; `.TrustVersion`, `.TrustActions`, `.TrustKeyTypes`) | Per-version constants container used throughout | **Missing** |
| `WsSerializationContext.TrustKeyTypes.Symmetric / Bearer / PublicKey / PSHA1` | Canonical key type URIs per version | **Missing** (raw string constants are not exposed) |
| `Microsoft.IdentityModel.Protocols.WsTrust.Entropy` (ctor takes `BinarySecret`; `BinarySecret`, `ProtectedKey` props) | Carries client entropy | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsTrust.BinarySecret` (`Data` byte array) | Entropy payload | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsPolicy.AppliesTo` (ctor takes `EndpointReference`) | Target service URL | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsAddressing.EndpointReference` (ctor takes string) | WS-A EPR for `AppliesTo` | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsFed.Claims` (ctor takes dialect + `IList<ClaimType>`; `Dialect`, `ClaimTypes`) | Requested claims block | **Missing** (no equivalent class even in the suggested replacement) |
| `Microsoft.IdentityModel.Protocols.WsFed.ClaimType` (`Uri`, `Value`, `IsOptional`) | Single requested claim | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsTrust.Psha1KeyGenerator.FillRandomBytes(byte[])` and `ComputeCombinedKey(byte[], byte[], int)` | Generates client entropy and computes PSHA1 combined key per WS-Trust §4.4.3 | **Missing** |

### 2. Object model — response side

Used in `WSTrustUtilities.CreateGenericXmlSecurityToken` / `GetProofToken` and in `WSTrustChannelSecurityTokenProvider.IsWsTrustResponseExpired`.

| Type / member | Used as | Equivalent in WsFederation pkg? |
|---|---|---|
| `Microsoft.IdentityModel.Protocols.WsTrust.WsTrustResponse` (`RequestSecurityTokenResponseCollection`) | Top-level deserialized RSTR(C) | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsTrust.RequestSecurityTokenResponse` (`KeyType`, `KeySizeInBits`, `RequestedSecurityToken`, `RequestedProofToken`, `Entropy`, `AttachedReference`, `UnattachedReference`, `Lifetime`) | Single RSTR | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsTrust.RequestedSecurityToken.TokenElement` (XmlElement) | The actual SAML/SAML2 token XML to wrap as `GenericXmlSecurityToken` | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsTrust.RequestedProofToken` (`BinarySecret`, `ComputedKeyAlgorithm`, `EncryptedKey`) | Proof key delivery | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsTrust.Lifetime` (`Created`, `Expires`) | Token validity window for caching/expiration | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsSecurity.SecurityTokenReference` | Attached / unattached `<o:SecurityTokenReference>` | **Missing** |
| `Microsoft.IdentityModel.Protocols.WsSecurity.WsSecuritySerializer.CreateXmlElement(SecurityTokenReference)` | Materializes STR as `XmlElement` to feed `GenericXmlSecurityKeyIdentifierClause` | **Missing** |

### 3. Serializer — the central capability

Used in `WSTrustRequestBodyWriter.OnWriteBodyContents`, `WSTrustChannel.IssueAsync`, and `WSTrustChannelSecurityTokenProvider.GetTokenCore` / `GetTokenAsyncCore`.

```csharp
// Write
new WsTrustSerializer().WriteRequest(XmlDictionaryWriter, WsTrustVersion, WsTrustRequest);

// Read
WsTrustResponse response = new WsTrustSerializer().ReadResponse(XmlDictionaryReader);
```

This is the **single most important missing capability**. It performs the entire RST and RSTR(C) XML serialization/deserialization across all three trust versions, including:

- `<wst:RequestSecurityToken>` envelope with `Context`, `RequestType`, `TokenType`, `KeyType`, `KeySize`, `AppliesTo`, `Claims`, `Entropy`, `BinarySecret`, `ComputedKeyAlgorithm`, plus arbitrary `AdditionalXmlElements` extension points
- `<wst:RequestSecurityTokenResponseCollection>` / `<wst:RequestSecurityTokenResponse>` reading: `RequestedSecurityToken` (extracts the embedded SAML/SAML2 `XmlElement`), `RequestedProofToken` (binary or computed-key), server `Entropy`, `Lifetime` (`xsd:dateTime` `Created`/`Expires`), `RequestedAttachedReference` and `RequestedUnattachedReference` containing `SecurityTokenReference`
- WS-Trust 1.3 (OASIS `200512`), WS-Trust 1.4 (OASIS `200802`), and WS-Trust Feb 2005 (`xmlsoap.org/2005/02`) namespaces and element/attribute differences

**Replacement availability:** `Microsoft.IdentityModel.Protocols.WsFederation` exposes **no serializer at all** for these messages. `WsFederationMetadataSerializer` reads/writes federation metadata XML, not WS-Trust messages. `WsFederationMessage` (de)serializes browser-redirect query strings, not SOAP bodies.

### 4. What the WsFederation package *does* provide (for completeness)

Per [`PublicAPI.Shipped.txt`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Protocols.WsFederation/PublicAPI.Shipped.txt) on `dev`:

- `WsFederationConfiguration`, `WsFederationConfigurationRetriever`, `WsFederationConfigurationValidator`, `SecurityTokenServiceTypeRoleDescriptor` — federation **metadata** discovery and validation
- `WsFederationMessage` (extends `AuthenticationProtocolMessage`) — passive/browser profile message with query-string parameters (`wa`, `wtrealm`, `wctx`, `wresult`, etc.) and `BuildRedirectUrl` / `BuildFormPost`
- `WsFederationMetadataSerializer` — read/write the WS-Federation **metadata** document
- Constant tables: `WsFederationConstants`, plus `Microsoft.IdentityModel.Xml.WsTrustConstants`, `WsTrustConstants_1_3`, `WsTrustConstants_1_4`, `WsTrustConstants_2005`, `WsAddressingConstants`, `WsPolicyConstants`, `WsUtilityConstants`
- `WsFederationException`

None of these provide an object model or serializer for `<wst:RequestSecurityToken>` / `<wst:RequestSecurityTokenResponse>`.

## Conclusion / requested action

To unblock migration of `System.ServiceModel.Federation` off the deprecated `Microsoft.IdentityModel.Protocols.WsTrust`, an actively-maintained replacement must provide, at minimum:

1. **A WS-Trust active-profile object model** equivalent to:
   - `WsTrustRequest`, `WsTrustResponse`, `RequestSecurityTokenResponse`
   - `Entropy`, `BinarySecret`, `RequestedProofToken`, `RequestedSecurityToken`, `Lifetime`, `AppliesTo`, `EndpointReference`, `Claims`, `ClaimType`, `SecurityTokenReference`
   - `WsTrustVersion` enum and a `WsSerializationContext` (or equivalent) that exposes per-version `TrustKeyTypes` (Symmetric/Bearer/PublicKey/PSHA1) and `TrustActions` (Issue/Cancel/Renew/Validate plus their `…Request` action URIs).
2. **A serializer** equivalent to `WsTrustSerializer.WriteRequest(XmlDictionaryWriter, WsTrustVersion, WsTrustRequest)` and `WsTrustSerializer.ReadResponse(XmlDictionaryReader)`, supporting **WS-Trust 1.3, 1.4, and Feb 2005**, and supporting `XmlDictionaryReader`/`XmlDictionaryWriter` (so it can be plugged directly into a WCF `Message` body via `BodyWriter.OnWriteBodyContents` and `Message.GetReaderAtBodyContents`).
3. **`Psha1KeyGenerator`** (or an equivalent helper) implementing PSHA1 combined-key derivation per WS-Trust 1.3 §4.4.3 and a cryptographically-strong random byte filler.
4. Continued support for `<wsse:SecurityTokenReference>` element materialization (today: `WsSecuritySerializer.CreateXmlElement(SecurityTokenReference)`), since WCF wraps the result in `GenericXmlSecurityKeyIdentifierClause`.

`Microsoft.IdentityModel.Protocols.WsFederation` covers a different protocol (passive/browser SSO + federation metadata) and supplies none of the above. As shipped on `dev`, it is **not** a functional substitute for `Microsoft.IdentityModel.Protocols.WsTrust` for `System.ServiceModel.Federation`'s use case.  

Alternatively, migrate the Microsoft.IdentityModel.Protocols.WsTrust implementation to the dev branch and release it along with the other packages with a current version number.

iNinja:
## Summary

Reinstates `Microsoft.IdentityModel.Protocols.WsTrust` as a supported package in the Wilson 8.x line. The package was deprecated in 6.x, while `System.ServiceModel.Federation` still depends on it and has no replacement path. See #3463.

This forward-port preserves the 6.8 public surface and the behaviour required by WCF, adds current target frameworks and build integration, and addresses the parser, serializer, and resource-handling defects found during review.

## Review guide

The implementation evolved through review. Please assess the final diff rather than relying on intermediate commit states.

Suggested review order:

1. **Forward-port, build, and API surface**
   - Package and test project integration
   - Target frameworks
   - Public and internal API baselines
2. **AOT and compatibility corrections**
   - SAML source preservation without reflection
   - WS-Trust constants and key-generation corrections
   - Existing 6.8 compatibility boundaries
3. **Parser and container boundaries**
   - Reader progress
   - Empty elements
   - Unknown and extension content
   - Request and response attributes
4. **Token, reference, and extension handling**
   - `OnBehalfOf`, `ProofEncryption`, `UseKey`, and requested tokens
   - XML Encryption proof-token round-tripping
   - Opt-in `ActAs`
5. **XML resource limits**
   - Parse-scoped cumulative budgets
   - Bounded streaming before DOM materialisation
   - Skipped and unknown subtree handling
6. **Regression and interoperability coverage**
   - Six target frameworks
   - WCF unit tests
   - Live independent STS scenarios

## Forward-port

The WS-Trust source was ported from the original 6.8 codebase to `dev8x`.

| Area | Change |
|---|---|
| Target frameworks | `net462`, `net472`, `netstandard2.0`, `net6.0`, `net8.0`, `net9.0`, `net10.0` |
| Build integration | Added the package and test projects to the solution and package build |
| Cryptography | Replaced obsolete `RNGCryptoServiceProvider` usage |
| API tracking | Added public and internal API baselines |
| AOT and trimming | Removed reflection used to access captured SAML XML and enabled direct internal access from `Microsoft.IdentityModel.Tokens.Saml` |
| XML handling | Disabled external XML resolution and bounded buffered XML processing |

## Parser and serializer corrections

The updated serializer guarantees progress when reading unknown or empty elements and keeps each reader within its own wrapper. This covers unknown proof-token content, endpoint-reference extensions, duplicate `AppliesTo` local names, non-empty request extensions, empty context items, empty request and response containers, and reference wrappers.

Request and response extension attributes now preserve their names, namespaces, and values without duplicating typed `Context` or namespace declarations. Response context and open content round-trip through the public model.

Typed WS-* elements are recognised by exact QName. Wrong-namespace lookalikes remain non-fatal and are captured as extensions or skipped according to the containing element.

Token, security-token-reference, and preserved raw XML forms now share the same read and write paths for `OnBehalfOf`, `ProofEncryption`, `UseKey`, and `RequestedSecurityToken`. Existing `TokenElement` precedence and empty `ProofEncryption` read behaviour are preserved. Static `UseKey` and `ProofEncryption` APIs also accept explicit token-handler collections.

An additive `WriteRequest` overload can opt in to serialising the existing `ActAs` property. The existing overload remains unchanged and continues to omit `ActAs` for compatibility. The opt-in path writes the WS-Trust 1.4 `200802` extension inside the selected core request namespace, preserves SAML source XML, supports security-token references and custom handlers, and writes `ActAs` before `AdditionalContext`.

Parsed XML Encryption `EncryptedKey` content can be retained and round-tripped. A manually constructed empty `EncryptedKey`, protected entropy, or another in-scope token state without a serialisable representation fails before any parent XML is written rather than producing an empty wrapper.

Empty `BinarySecret` elements retain their `Type` with an empty data array. Additional XML attributes and `SecurityTokenReference.Usage` now round-trip.

## XML resource limits

Buffered and streamed XML processing now uses a shared parse-scoped budget. Limits are applied before internal buffering and cover depth, buffered bytes, text and Base64 content, element counts, attribute counts, names, and cumulative extension content.

Library-owned XML copying reads text and attribute values in chunks before materialising DOM content. Unknown, skipped, endpoint-reference, and token subtrees all use this bounded path. Quota failures use `IDX15026`.

The current internal defaults allow up to 4 MB of buffered or textual XML content and 4,096 elements or attributes within one parse.

## Compatibility

This is a compatibility-first forward-port of a legacy protocol implementation. Where schema purity and established 6.8 wire behaviour differ, this PR preserves the deployed behaviour unless a change is required to address a concrete safety, correctness, or interoperability failure. Broader protocol changes remain separate and require consumer-specific validation.

The existing public API remains available. Duplicate singleton handling, `TokenElement` precedence, extension placement, and the broad 6.8 properties that are outside the validated WCF scenarios remain unchanged.

The existing `WriteRequest` overload preserves the 6.8 `ActAs` behaviour. Callers must use the new overload to include it. Other broader serializer-completeness work remains separate.

The intentional behaviour changes are limited to states that previously appeared to serialise successfully while producing unusable empty content, malformed input that cannot safely make progress, wrong-namespace content that was incorrectly classified as typed content, and messages exceeding the new resource limits.

## Validation

| Check | Result |
|---|---|
| WS-Trust tests on `net462` | 183 passed |
| WS-Trust tests on `net472` | 183 passed |
| WS-Trust tests on `net6.0` | 183 passed |
| WS-Trust tests on `net8.0` | 183 passed |
| WS-Trust tests on `net9.0` | 183 passed |
| WS-Trust tests on `net10.0` | 183 passed |
| Total WS-Trust test executions | 1,098 passed, 0 failed, 0 skipped |
| `System.ServiceModel.Federation` unit tests | 3 passed, 0 failed |
| WCF federation integration tests against the independent net471 STS | 10 passed, 0 failed, 0 skipped |
| Package dependency alignment | All required `Microsoft.IdentityModel` packages resolved to the same validation version |

The WCF integration matrix covers WS-Trust February 2005 and WS-Trust 1.3, Text and MTOM encodings, and secure-conversation enabled and disabled.

Closes #3463.


src/Microsoft.IdentityModel.Protocols.WsTrust/WsFed/WsFedSerializer.cs
```diff
@@ -0,0 +1,205 @@
...
+            //      Optional="true/false">
+            //   <auth:Value>
+            //      77a6fa04-0454-4d08-8761-2a840e281399
+            //   </auth:Value>
+            // </auth:ClaimType>
+
+            XmlAttributeHolder[] attributes = XmlAttributeHolder.ReadAttributes(reader);
+            string uri = XmlAttributeHolder.GetAttribute(attributes, WsFedAttributes.Uri, @namespace);
+            if (string.IsNullOrEmpty(uri))
+                throw LogHelper.LogExceptionMessage(new XmlReadException(LogHelper.FormatInvariant(WsTrust.LogMessages.IDX15013, WsFedElements.ContextItem, WsFedAttributes.Name)));
```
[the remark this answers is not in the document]
iNinja: Fixed. Error message now references ClaimType and Uri.

src/Microsoft.IdentityModel.Protocols.WsTrust/Psha1KeyGenerator.cs
```diff
@@ -0,0 +1,254 @@
...
+
+        static int s_minKeySizeInBits = 16 * 8; // 16 Bytes - 128 bits.
+        static int s_maxKeySizeInBits = (16 * 1024) * 8; // 16 K
+
+        /// <summary>
+        /// Computes the session key based on PSHA1 algorithm.
+        /// </summary>
+        /// <param name="requestorEntropy">The entropy from the requestor side.</param>
+        /// <param name="issuerEntropy">The entropy from the token issuer side.</param>
+        /// <param name="keySizeInBits">The desired key size in bits.</param>
```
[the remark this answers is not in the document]
iNinja: Fixed. Aligned doc param order with method signature.

test/Microsoft.IdentityModel.Protocols.WsTrust.Tests/RequestSecurityTokenResponseTests.cs
```diff
@@ -0,0 +1,400 @@
...
+            }
+            catch (Exception ex)
+            {
+                theoryData.ExpectedException.ProcessException(ex, context);
+            }
+
+            TestUtilities.AssertFailIfErrors(context);
+        }
+
+        public static TheoryData<WsTrustTheoryData> ReadRequestSeurityTokenResponseTestCases
```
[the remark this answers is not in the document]
iNinja: Fixed. Renamed to ReadRequestedSecurityTokenResponse.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsSecurity/WsSecuritySerializer.cs
```diff
@@ -0,0 +1,172 @@
...
+            //  <wsse:SecurityTokenReference wsu:Id="...",
+            //                               wsse:TokenType="...",
+            //                               wsse:Usage="...">
+            //      ...
+            //  </wsse:SecurityTokenReference>
+
+            XmlAttributeHolder[] xmlAttributes = XmlAttributeHolder.ReadAttributes(reader);
+            var securityTokenReference = new SecurityTokenReference();
+
+            string id = XmlAttributeHolder.GetAttribute(xmlAttributes, WsUtilityAttributes.Id, WsSecurityConstants.WsSecurity10.Namespace);
```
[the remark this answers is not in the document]
iNinja: Fixed. Using WsUtilityConstants.WsUtility10.Namespace.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustBinarySecrectTypes.cs
```diff
@@ -0,0 +1,90 @@
...
+    public class WsTrust13BinarySecretTypes : WsTrustBinarySecretTypes
+    {
+        /// <summary>
+        /// Instantiates WS-Trust 1.3 BinarySecretTypes.
+        /// </summary>
+        public WsTrust13BinarySecretTypes()
+        {
+            AsymmetricKey = "http://docs.oasis-open.org/ws-sx/ws-trust/200512/AsymmetricKey";
+            Nonce = "http://docs.oasis-open.org/ws-sx/ws-trust/200512/Nonce";
+            SymmetricKey = "http://docs.oasis-open.org/ws-sx/ws-trust/200512/SymmetricKey"; ;
```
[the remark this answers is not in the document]
iNinja: Fixed. Removed.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsSecurity/WsSecuritySerializer.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            writer.WriteStartElement(WsSecurityConstants.WsSecurity10.Prefix, WsSecurityElements.SecurityTokenReference, WsSecurityConstants.WsSecurity10.Namespace);
+
+            // For Saml2 tokens, the 'TokenType' was defined in must be in wsse1.1 namespace
+            if (!string.IsNullOrEmpty(securityTokenReference.TokenType))
+                writer.WriteAttributeString(WsSecurityAttributes.TokenType, WsSecurityConstants.WsSecurity11.Namespace, securityTokenReference.TokenType);
+
+            if (!string.IsNullOrEmpty(securityTokenReference.Id))
+                writer.WriteAttributeString(WsUtilityAttributes.Id, securityTokenReference.Id);
+
```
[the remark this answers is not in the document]
iNinja: Fixed. Writing as wsu:Id with WsUtilityConstants.WsUtility10 prefix and namespace.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsSecurity/WsSecuritySerializer.cs
```diff
@@ -0,0 +1,172 @@
...
+            //                      ValueType="..."
+            //                      EncodingType="...">
+            //      ...
+            //  </wsse:KeyIdentifier>
+
+            writer.WriteStartElement(WsSecurityConstants.WsSecurity10.Prefix, WsSecurityElements.KeyIdentifier, WsSecurityConstants.WsSecurity10.Namespace);
+
+            if (!string.IsNullOrEmpty(keyIdentifier.Id))
+                writer.WriteAttributeString(WsUtilityAttributes.Id, keyIdentifier.Id);
+
```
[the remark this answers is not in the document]
iNinja: Fixed. Writing as wsu:Id with WsUtilityConstants.WsUtility10 prefix and namespace.

src/Microsoft.IdentityModel.Protocols.WsTrust/Psha1KeyGenerator.cs
```diff
@@ -0,0 +1,254 @@
...
+        /// <exception cref="ArgumentException">When keySizeInBits is not a whole number of bytes.</exception>
+        internal static byte[] GenerateSymmetricKey(int keySizeInBits, byte[] senderEntropy, out byte[] receiverEntropy)
+        {
+            if (senderEntropy == null)
+                throw LogHelper.LogArgumentNullException(nameof(senderEntropy));
+
+            int keySizeInBytes = ValidateKeySizeInBytes(keySizeInBits);
+            receiverEntropy = new byte[keySizeInBytes];
+            _random.GetNonZeroBytes(receiverEntropy);
+            return ComputeCombinedKey(senderEntropy, receiverEntropy, keySizeInBits);
```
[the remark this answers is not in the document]
iNinja: Fixed. Swapped to ComputeCombinedKey(receiverEntropy, senderEntropy, keySizeInBits) so issuer entropy is the first argument.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustBinarySecrectTypes.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// Constants: WS-Trust 1.4 BinarySecretTypes.
+    /// </summary>
+    public class WsTrust14BinarySecretTypes : WsTrustBinarySecretTypes
+    {
+        /// <summary>
+        /// Instantiates WS-Trust 1.4 BinarySecretTypes.
+        /// </summary>
+        public WsTrust14BinarySecretTypes()
+        {
+            AsymmetricKey = "http://docs.oasis-open.org/ws-sx/ws-trust/200512/Bearer";
```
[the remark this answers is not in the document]
iNinja: Fixed. Corrected to `http://docs.oasis-open.org/ws-sx/ws-trust/200802/AsymmetricKey`.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustResponse.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// Instantiates a new <see cref="WsTrustResponse"/> with a <see cref="RequestSecurityTokenResponse"/>.
+        /// </summary>
+        /// <param name="requestSecurityTokenResponse">the response to add to the collection.</param>
+        /// <exception cref="ArgumentNullException">thrown if <paramref name="requestSecurityTokenResponse"/> is null.</exception>
+        public WsTrustResponse(RequestSecurityTokenResponse requestSecurityTokenResponse)
+        {
+            if (requestSecurityTokenResponse == null)
+                LogHelper.LogArgumentNullException(nameof(requestSecurityTokenResponse));
+
+            RequestSecurityTokenResponseCollection.Add(requestSecurityTokenResponse);
```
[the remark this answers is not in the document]
iNinja: Fixed. Added missing throw.

test/Microsoft.IdentityModel.Protocols.WsTrust.Tests/EntropyTests.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        [Theory, MemberData(nameof(WriteEntropyTestCases))]
+        public void WriteEntropy(WsTrustTheoryData theoryData)
+        {
+            var context = TestUtilities.WriteHeader($"{this}.WriteEntropy", theoryData);
+            try
+            {
+                WsTrustSerializer.WriteEntropy(theoryData.Writer, theoryData.WsSerializationContext, theoryData.Entropy);
+                //IdentityComparer.AreEqual(entropy, theoryData.Entropy, context);
```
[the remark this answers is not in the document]
iNinja: Fixed. Removed.

test/Microsoft.IdentityModel.Protocols.WsTrust.Tests/WsDefaults.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public static SecurityTokenReference SecurityTokenReference => new SecurityTokenReference {
+            KeyIdentifier = KeyIdentifier,
+            TokenType = "http://docs.oasis-open.org/wss/oasis-wss-saml-token-profile-1.1#SAMLV2.0",
+
+        };
+    }
+}
+
+#pragma warning restore CS3016 // Arrays as attribute arguments is nEndoot CLS-compliant
```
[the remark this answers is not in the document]
iNinja: Fixed.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsSecurity/WsSecurityConstants.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        /// <summary>
+        /// Gets constants for WS-Security 1.0.
+        /// </summary>
+        public static WsSecurity10Constants WsSecurity10 => new WsSecurity10Constants();
+
+        /// <summary>
+        /// Gets constants for WS-Security 1.1.
+        /// </summary>
+        public static WsSecurity11Constants WsSecurity11 => new WsSecurity11Constants();
```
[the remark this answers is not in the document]
iNinja: Fixed. Both properties now return cached static instances.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsSecurity/WsSecuritySerializer.cs
```diff
@@ -0,0 +1,172 @@
...
+            //                      ValueType="..."
+            //                      EncodingType="...">
+            //      ...
+            //  </wsse:KeyIdentifier>
+
+            bool isEmptyElement = reader.IsEmptyElement;
+            var xmlAttributes = XmlAttributeHolder.ReadAttributes(reader);
+
+            var keyIdentifier = new KeyIdentifier();
+            string id = XmlAttributeHolder.GetAttribute(xmlAttributes, WsUtilityAttributes.Id, WsSecurityConstants.WsSecurity10.Namespace);
```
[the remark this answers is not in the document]
iNinja: Fixed. Using WsUtilityConstants.WsUtility10.Namespace.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustBinarySecrectTypes.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// Constants: WS-Trust 1.4 BinarySecretTypes.
+    /// </summary>
+    public class WsTrust14BinarySecretTypes : WsTrustBinarySecretTypes
+    {
+        /// <summary>
+        /// Instantiates WS-Trust 1.4 BinarySecretTypes.
+        /// </summary>
+        public WsTrust14BinarySecretTypes()
+        {
+            AsymmetricKey = "http://docs.oasis-open.org/ws-sx/ws-trust/200802/AsymmetricKey";
```
adcosta118: 🤖 *AI-generated review comment — please verify before acting.*

Replacing the obvious .../200512/Bearer bug with .../200802/AsymmetricKey fixes the wrong value, but it leaves WsTrust14BinarySecretTypes internally inconsistent: AsymmetricKey now uses the 200802 namespace while Nonce (line 86) and SymmetricKey (line 87) remain on 200512.

Per the WS-Trust 1.4 (200802) spec the defined BinarySecret Type URIs are .../200802/SymmetricKey and .../200802/Nonce — and 1.4 does **not** define an AsymmetricKey BinarySecret type at all. So this change is likely wrong in two ways: Nonce/SymmetricKey should probably move to 200802, and .../200802/AsymmetricKey is not a spec-defined value. Please double-check the intended URIs (and whether any interop/round-trip test actually asserts these strings).

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustSerializer.cs
```diff
@@ -0,0 +1,1650 @@
...
+                    byte[] data = reader.ReadContentAsBase64();
+                    if (data != null)
+                        binarySecret.Data = data;
+
+                    reader.ReadEndElement();
+                }
+                else
+                {
+                    reader.Read();
+                    return null;
```
adcosta118: 🤖 *AI-generated review comment — please verify before acting.*

Behavioral change worth confirming: an empty <t:BinarySecret Type="..."/> now returns 
ull, silently discarding the EncodingType/Type attribute that was just parsed on lines 67-70. Both callers (ReadEntropy ~L183, proof-token reader ~L643) assign the result straight into .BinarySecret, so 
ull is tolerated — but an empty BinarySecret carrying only a Type attribute is now indistinguishable from a missing one. If that attribute-only-but-empty case is intended to be dropped, consider a brief comment; otherwise returning the constructed inarySecret (with EncodingType set, Data null) may be more faithful.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustSerializer.cs
```diff
@@ -0,0 +1,1650 @@
...
+            XmlTokenStream xmlTokenStream = null;
+            if (assertion is Saml2Assertion saml2)
+                xmlTokenStream = saml2.XmlTokenStream;
+            else if (assertion is SamlAssertion saml)
+                xmlTokenStream = saml.XmlTokenStream;
+
+            if (xmlTokenStream == null)
+                return false;
+
+            xmlTokenStream.WriteTo(writer);
```
adcosta118, line 1283: 🤖 *AI-generated review comment — please verify before acting.*

The reflection-to-direct-access refactor also drops the old 	ry/catch (Exception) that returned alse on any failure. Now if XmlTokenStream.WriteTo(writer) throws, the exception propagates instead of causing TryWriteSourceData to return alse and fall back to the SecurityTokenHandlers loop in WriteOnBehalfOf. This is arguably better (no silent swallow), but it is a behavior change on the write path — please confirm the lost fallback is acceptable.

src/Microsoft.IdentityModel.Protocols.WsTrust/Protocols/WsUtils.cs
```diff
@@ -0,0 +1,226 @@
...
+        /// <see cref="XmlDictionaryReaderQuotas"/> with a bounded <see cref="XmlDictionaryReaderQuotas.MaxDepth"/>
+        /// (32) used when materializing buffered XML. All other limits stay at their maximum so large but
+        /// shallow WS-Trust messages are unaffected. Keeps element nesting within a sane bound, consistent
+        /// with the reader quotas used elsewhere in the stack.
+        /// </summary>
+        internal static readonly XmlDictionaryReaderQuotas BoundedReaderQuotas = new XmlDictionaryReaderQuotas
+        {
+            MaxArrayLength = int.MaxValue,
+            MaxBytesPerRead = int.MaxValue,
+            MaxDepth = 32,
```
adcosta118, line 34: 🤖 *AI-generated review comment — please verify before acting.*

MaxDepth = 32 matches the WCF/XmlDictionaryReaderQuotas framework default, so this is a reasonable bound. One thing to sanity-check: this quota (and the DepthLimitingXmlReader cap) is applied to ReadUnknownElement, ReadEndpointReference and buffered re-parsing, which can carry signed/encrypted SAML assertions with fairly deep XML (nested KeyInfo/EncryptedData/ds:Signature). Please confirm 32 levels is comfortably above the deepest legitimate token you expect, so valid OnBehalfOf/extension payloads aren't rejected with IDX15025.

src/Microsoft.IdentityModel.Protocols.WsTrust/Psha1KeyGenerator.cs
```diff
@@ -0,0 +1,254 @@
...
+        /// <exception cref="ArgumentException">When keySizeInBits is not a whole number of bytes.</exception>
+        internal static byte[] GenerateSymmetricKey(int keySizeInBits, byte[] senderEntropy, out byte[] receiverEntropy)
+        {
+            if (senderEntropy == null)
+                throw LogHelper.LogArgumentNullException(nameof(senderEntropy));
+
+            int keySizeInBytes = ValidateKeySizeInBytes(keySizeInBits);
+            receiverEntropy = new byte[keySizeInBytes];
+            _random.GetNonZeroBytes(receiverEntropy);
+            return ComputeCombinedKey(receiverEntropy, senderEntropy, keySizeInBits);
```
adcosta118: 🤖 *AI-generated review comment — please verify before acting.*

Confirming this is an intentional correctness fix (not a regression): ComputeCombinedKey(issuerEntropy, requestorEntropy, ...) uses 
equestorEntropy as the P_SHA1 secret and issuerEntropy as the seed (key = P_SHA1(EntREQ, EntRES)). Since 
eceiverEntropy here is the issuer/responder entropy and senderEntropy is the requestor entropy, ComputeCombinedKey(receiverEntropy, senderEntropy, ...) now maps the roles correctly — the previous order was reversed. Note this does change the derived key bytes vs. the prior (buggy) behavior, so anything that pinned the old output will differ.

adcosta118 (commented):
Automated review of the post-port improvement commits (69ec2ac0, 7378709f, 96d8e841); the forward-port commit 89aa6244 was excluded per request. Findings are inline. 🤖 AI-generated — please verify.

src/Microsoft.IdentityModel.Protocols.WsTrust/Protocols/WsUtils.cs
```diff
@@ -155,4 +171,56 @@ internal static void CheckReaderOnEntry(XmlReader reader, string element, WsSeri
...
     }
+
+    /// <summary>
+    /// Wraps an <see cref="XmlReader"/> and throws <see cref="System.Xml.XmlException"/> if element
+    /// nesting depth exceeds the configured maximum. Does not buffer and does not own (dispose)
+    /// the inner reader.
+    /// </summary>
+    internal sealed class DepthLimitingXmlReader : XmlReader
+    {
+        private readonly XmlReader _inner;
```
adcosta118, line 521: DepthLimitingXmlReader already extends XmlReader; do we need to encapsulate an additional instance?

src/Microsoft.IdentityModel.Protocols.WsTrust/Protocols/WsUtils.cs
```diff
@@ -155,4 +171,56 @@ internal static void CheckReaderOnEntry(XmlReader reader, string element, WsSeri
...
+        internal DepthLimitingXmlReader(XmlReader inner, int maxDepth)
+        {
+            _inner = inner ?? throw LogHelper.LogArgumentNullException(nameof(inner));
+            _maxDepth = maxDepth;
+        }
+
+        public override bool Read()
+        {
+            bool result = _inner.Read();
+            if (_inner.Depth > _maxDepth)
```
adcosta118, line 537: I don't see where we are increasing the depth, is it inside Xmlreader? If so, the check should also be there no?

src/Microsoft.IdentityModel.Protocols.WsTrust/WsSecurity/WsSecuritySerializer.cs
```diff
@@ -36,12 +36,13 @@ public static XmlElement CreateXmlElement(SecurityTokenReference securityTokenRe
                     stream.Seek(0, SeekOrigin.Begin);
                     var dom = new XmlDocument
                     {
-                        PreserveWhitespace = true
+                        PreserveWhitespace = true,
+                        XmlResolver = null
                     };
 
-                    using (var textReader = new XmlTextReader(stream) { DtdProcessing = DtdProcessing.Prohibit })
+                    using (var dictReader = XmlDictionaryReader.CreateTextReader(stream, XmlDictionaryReaderQuotas.Max))
```
adcosta118: Should we use BoundedReaderQuotas here also? If so, maybe we should extract the one you have nested inside WsUtils.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustSerializer.cs
```diff
@@ -869,14 +861,18 @@ private WsTrustResponse ReadResponse(XmlDictionaryReader reader, WsSerialization
...
         private static void ReadUnknownElement(XmlDictionaryReader reader, WsTrustRequest trustRequest)
         {
+            // Capture the (open-content / extension) unknown element into a DOM. The subtree is read
+            // through a streaming DepthLimitingXmlReader that caps element nesting depth; it is a
+            // pass-through on typical input and only signals when the configured depth is exceeded.
             bool isEmptyElement = reader.IsEmptyElement;
-            var doc = new XmlDocument();
-            doc.Load(reader.ReadSubtree());
+            var doc = new XmlDocument { XmlResolver = null };
+            using (var depthLimitingReader = new DepthLimitingXmlReader(reader.ReadSubtree(), WsUtils.BoundedReaderQuotas.MaxDepth))
```
adcosta118: Does reader.ReadSubtree() actually do any read? If so isn't it possible to get a stack overflow before DepthLimitingXmlReader is able to do anything about it?

src/Microsoft.IdentityModel.Protocols.WsTrust/Psha1KeyGenerator.cs
```diff
@@ -166,7 +166,7 @@ internal static byte[] GenerateSymmetricKey(int keySizeInBits, byte[] senderEntr
             int keySizeInBytes = ValidateKeySizeInBytes(keySizeInBits);
             receiverEntropy = new byte[keySizeInBytes];
             _random.GetNonZeroBytes(receiverEntropy);
-            return ComputeCombinedKey(senderEntropy, receiverEntropy, keySizeInBits);
+            return ComputeCombinedKey(receiverEntropy, senderEntropy, keySizeInBits);
```
adcosta118: Was this an existing bug?

src/Microsoft.IdentityModel.Protocols.WsTrust/Psha1KeyGenerator.cs
```diff
@@ -20,8 +20,8 @@ public static class Psha1KeyGenerator
         /// <summary>
         /// Computes the session key based on PSHA1 algorithm.
         /// </summary>
-        /// <param name="requestorEntropy">The entropy from the requestor side.</param>
         /// <param name="issuerEntropy">The entropy from the token issuer side.</param>
+        /// <param name="requestorEntropy">The entropy from the requestor side.</param>
```
adcosta118, line 24: We have
- in the comments: issuerEntropy and requestorEntropy
- in the code: senderEntropy and receiverEntropy

Are they the same pairs? If so it would be less confusing if we used the same name

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustSerializer.cs
iNinja: Fixed. Empty elements now return the BinarySecret object with EncodingType set. The previous null return was discarding the Type attribute.

src/Microsoft.IdentityModel.Protocols.WsTrust/Psha1KeyGenerator.cs
iNinja: Pre-existing bug carried from the 6.8.0 source. GenerateSymmetricKey is internal and has no callers in this package, so there is nothing to migrate.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustSerializer.cs
iNinja: ReadSubtree() returns a new reader positioned at the current element without advancing into child content. The first actual read happens when XmlDocument.Load calls Read() on the subtree reader, at which point the depth check applies.

src/Microsoft.IdentityModel.Protocols.WsTrust/Protocols/WsUtils.cs
iNinja: XmlReader.Depth is a built-in property of the underlying XML parser tracking element nesting. We read it after every Read() and throw if it exceeds the limit.

src/Microsoft.IdentityModel.Protocols.WsTrust/Protocols/WsUtils.cs
iNinja: The depth limit only applies to unknown open-content extension elements and additional EndpointReference elements. The main token path, including RequestedSecurityToken and any SAML assertions inside it, is read directly without going through DepthLimitingXmlReader.

src/Microsoft.IdentityModel.Protocols.WsTrust/Psha1KeyGenerator.cs
iNinja: Fixed. Renamed to requestorEntropy and issuerEntropy to match ComputeCombinedKey.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustBinarySecrectTypes.cs
iNinja: Fixed. Corrected all three 1.4 URIs to use the 200802 namespace (AsymmetricKey, Nonce, SymmetricKey).

src/Microsoft.IdentityModel.Protocols.WsTrust/Protocols/WsUtils.cs
iNinja: XmlReader is abstract with no useful default implementations. The delegation pattern is the standard approach for partial reader wrappers in Wilson (see DelegatingXmlDictionaryReader). All abstract members need to be overridden regardless.

src/Microsoft.IdentityModel.Protocols.WsTrust/Psha1KeyGenerator.cs
iNinja: Yes, present in the original origin/wstrust source and carried forward in the port. Fixed in the previous commit once GHCP identified it.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsSecurity/WsSecuritySerializer.cs
iNinja: Fixed. Using WsUtils.BoundedReaderQuotas for consistency.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustSerializer.cs
iNinja: The catch existed to handle reflection failures when locating XmlTokenStream. With InternalsVisibleTo and direct property access that failure path no longer exists, so there is nothing to catch. Exceptions propagate as expected.

src/Microsoft.IdentityModel.Protocols.WsTrust/WsTrustBinarySecrectTypes.cs
adcosta118: 🤖 *AI-generated review comment — please verify before acting.*

Reopening with a correction to my earlier guidance: the authoritative WS-Trust 1.4 specification defines all three predefined `BinarySecret/@Type` values under the **200512** URI namespace, even though the WS-Trust 1.4 element namespace itself is `200802`:

- `http://docs.oasis-open.org/ws-sx/ws-trust/200512/AsymmetricKey`
- `http://docs.oasis-open.org/ws-sx/ws-trust/200512/SymmetricKey`
- `http://docs.oasis-open.org/ws-sx/ws-trust/200512/Nonce`

The current change to `200802/*` therefore makes all three constants non-standard. Please restore the three 1.4 `BinarySecret` type constants to the `200512/*` values and add exact-value assertions so this distinction is preserved. Apologies for the incorrect direction in the original comment.

adcosta118: 🤖 *AI-generated review comment — please verify before acting.*

Tracking the still-open issue from the corrected follow-up in the resolved thread: WS-Trust 1.4 defines `BinarySecret/@Type` as `.../200512/AsymmetricKey`, `.../200512/SymmetricKey`, and `.../200512/Nonce`; the current `200802/*` constants are non-standard. Please restore all three exact `200512/*` values and cover them with exact-value tests.

Original corrected follow-up: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3547#discussion_r3735027452

iNinja: Addressed.

fre-berhane (approved):

bgavrilMS (approved):

PR #3554, opened 2026-07-16
Fix SanitizeQueryParams to split on first '=' for consistent query parameter parsing

iNinja:
`SanitizeQueryParams` split each query parameter token on `=` and discarded any token that did not produce exactly two parts. This caused parameters whose values contain `=` (e.g. base64-padded strings like `sig=YWJjZA==`) to be silently dropped, and meant that malformed-looking parameters were not visible to the `AcceptUnsignedQueryParameters` validation gate.

This change replaces the split-on-all-`=` approach with a split on the first `=` only (`IndexOf('=')`), aligning Wilson's query-parameter canonicalization with how standard HTTP parsers (e.g. `HttpUtility.ParseQueryString`, ASP.NET `Request.Query`) handle the same input.

- Parameters with `=`-containing values (base64, encoded data) are now parsed and hashed correctly.
- Parameters that were previously invisible to the `AcceptUnsignedQueryParameters=false` gate are now visible; the existing `IDX23029` rejection path handles them.
- The `ValidQ6` creation test was updated to use a URL without the edge case that the old behavior was implicitly depending on.
- Two new validation tests cover the base64-padded case and the previously-bypassed unsigned-extra case.


test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
```diff
@@ -850,6 +850,28 @@ public static TheoryData<ValidateSignedHttpRequestTheoryData> ValidateQClaimTheo
...
+                        HttpRequestUri = new Uri("https://www.contoso.com/path1?queryParam1=value1&admin=true=1"),
+                        SignedHttpRequestToken = SignedHttpRequestTestUtils.ReplaceOrAddPropertyAndCreateDefaultSignedHttpRequest(new JProperty(SignedHttpRequestClaimTypes.Q, JArray.Parse($"[[\"queryParam1\"],\"{SignedHttpRequestTestUtils.CalculateBase64UrlEncodedHash("queryParam1=value1")}\"]"))),
+                        ExpectedException = new ExpectedException(typeof(SignedHttpRequestInvalidQClaimException), "IDX23029"),
+                        SignedHttpRequestValidationParameters = new SignedHttpRequestValidationParameters()
+                        {
+                            AcceptUnsignedQueryParameters = false,
+                        },
+                        TestId = "InvalidMalformedQueryParamBypassRejected",
+                    },
+                    new ValidateSignedHttpRequestTheoryData
```
cpp11nullptr, line 990: Can we add test case for `name1=value1&name2=&name3=value3`?

cpp11nullptr (approved):

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
iNinja: Added `ValidEmptyValueQueryParam` covering `name1=value1&name2=&name3=value3`.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1345,20 +1345,20 @@ private static Dictionary<string, string> SanitizeQueryParams(Uri httpRequestUri
             var queryParamKeyValuePairs = queryString.Split('&');
             foreach (var queryParamValuePair in queryParamKeyValuePairs)
             {
-                var queryParamKeyValuePairArray = queryParamValuePair.Split('=');
-                if (queryParamKeyValuePairArray.Length == 2)
+                // Split on the first '=' only so that values containing '=' (e.g. base64-padded
+                // strings like "sig=YWJjZA==") are parsed consistently with how resource servers
+                // parse them, rather than being silently discarded.
+                int separatorIndex = queryParamValuePair.IndexOf('=');
+                if (separatorIndex > 0)
```
adcosta118: 🤖 *AI-generated review comment — please verify before acting.*

**Interop / canonicalization change (non-blocking).** This changes how the `q` hash is computed for two input shapes that were previously dropped: values containing `=` and empty values (`name=`). Since both the client (create) and resource server (validate) share this method, both sides must run a version containing this change or the computed `q` hashes will disagree, causing validation failures during a rolling upgrade. Worth calling out in release notes so operators upgrade both sides together.

**Residual gap (non-blocking).** A bare parameter with no `=` (e.g. `?admin`) still yields `separatorIndex == -1` and is silently skipped, so it stays invisible to the `AcceptUnsignedQueryParameters=false` gate — the same class of bypass this PR closes for `admin=true=1`. If closing that is in scope, consider treating a no-`=` token consistently with how resource servers parse it.

adcosta118 (commented):
🤖 AI-generated review — please verify before acting. The split-on-first-'=' logic is correct and the added validation tests cover the key cases. Two non-blocking notes inline.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1345,20 +1345,20 @@ private static Dictionary<string, string> SanitizeQueryParams(Uri httpRequestUri
...
-                            repeatedQueryParams.Add(queryParamName);
-                        else if (!string.IsNullOrEmpty(queryParamValue))
-                            sanitizedQueryParams.Add(queryParamName, queryParamValue);
-                    }
+                    var queryParamName = queryParamValuePair.Substring(0, separatorIndex);
+                    var queryParamValue = queryParamValuePair.Substring(separatorIndex + 1);
+                    // if sanitizedQueryParams already contains the query parameter name it means that the queryParamName is repeated.
+                    // in that case queryParamName should not be added, and the existing entry in sanitizedQueryParams should be removed.
+                    if (sanitizedQueryParams.ContainsKey(queryParamName))
+                        repeatedQueryParams.Add(queryParamName);
```
adcosta118: It seems this was already implemented like that, but don't we care about the value in that case? When multiple query params with the same key are present, this usually means it's a collection.

I wonder whether there isn't a URL decoding library we can use here instead doing it manually.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestCreationTests.cs
```diff
@@ -536,7 +536,7 @@ public static TheoryData<CreateSignedHttpRequestTheoryData> CreateQClaimTheoryDa
                     {
                         ExpectedClaim = SignedHttpRequestClaimTypes.Q,
                         ExpectedClaimValue = $"{{\"q\":[[\"queryParam1\"],\"{SignedHttpRequestTestUtils.CalculateBase64UrlEncodedHash("queryParam1=value1")}\"]}}",
-                        HttpRequestUri = new Uri("https://www.contoso.com/path1?&queryParam1=value1&query=Param2=value2")
```
adcosta118, line 539: Is a `=` character actually allowed in queryParam values?
I think correct value will be URIComponent encoded, e.g.: `query=Param2%3Dvalue2`.

So this whole PR seems to be to not require the values to be correctly encoded?

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
```diff
@@ -850,6 +850,36 @@ public static TheoryData<ValidateSignedHttpRequestTheoryData> ValidateQClaimTheo
...
```
adcosta118, line 990: Could you please also add a tests for
- key starts with =
- key ends with =
- value starts with =
- value ends ends with =

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
iNinja: Added `Switch.Microsoft.IdentityModel.SignedHttpRequest.UseLegacyQueryParamParsing` as a rolling-upgrade escape hatch — restores the old drop behavior so operators can unblock while updating both sides. Bare `?flag` params are now also covered by the same fix (included as `key=\\`), and we log IDX23039 before IDX23029 when a previously-dropped param is the cause of rejection.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
iNinja: The repeated-name behavior is spec-mandated (SHR draft §7.5) — when a param name appears more than once there's no authoritative value, so it's excluded from coverage. On the library question, `HttpUtility.ParseQueryString` brings a `System.Web` dependency that's not available on all TFMs we target, so the manual approach is intentional.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestCreationTests.cs
iNinja: Yes, `=` is allowed unencoded in query param values per RFC 3986 — it's only a delimiter at the key/value boundary. The old `Split('=').Length == 2` requirement was overly strict: `sig=YWJjZA==` (base64 padding) is a valid param that was being silently dropped, which is a correctness bug independent of the security fix.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
iNinja: Added: leading `=` (no name, dropped), value starting with `=` (`key==rest` parsed as name=`key`, value=`=rest`), value ending with `=` (`key=value=`), and a `[Fact]` covering the legacy switch. Empty value was already covered by `ValidEmptyValueQueryParam`.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1345,21 +1349,52 @@ private static Dictionary<string, string> SanitizeQueryParams(Uri httpRequestUri
...
-                var queryParamKeyValuePairArray = queryParamValuePair.Split('=');
-                if (queryParamKeyValuePairArray.Length == 2)
+                // Split on the first '=' only so that values containing '=' (e.g. base64-padded
+                // strings like "sig=YWJjZA==") are parsed consistently with how resource servers
+                // parse them, rather than being silently discarded.
+                // Set Switch.Microsoft.IdentityModel.SignedHttpRequest.UseLegacyQueryParamParsing=true
+                // to restore the old drop-on-multi-equals behavior during a rolling upgrade.
+                int separatorIndex = queryParamValuePair.IndexOf('=');
+                string queryParamName, queryParamValue;
+                if (AppContextSwitches.UseLegacyQueryParamParsing)
```
pmaytak, line 1359: > The legacy switch does not faithfully restore pre-8.19 behavior for empty-value params. ( SignedHttpRequestHandler.cs:1359-1365 ,  AppContextSwitches.cs:116 )
> 
> Original code only added a param when the value was non-empty ( else if (!string.IsNullOrEmpty(queryParamValue)) ), so  name2=  was dropped. The new legacy branch splits on length==2 and adds  name2  with an empty value — it does not drop it:
> 
> ```csharp
> var splitQueryParam = queryParamValuePair.Split('=');
> if (splitQueryParam.Length != 2) continue;
> queryParamName = splitQueryParam[0];
> queryParamValue = splitQueryParam[1];   // "" for "name2="
> // ... later added unconditionally
> ```
> Consequence: during the exact rolling-upgrade scenario the switch exists for, a request signed by an old client (which excluded  name2=  from the  q  names and hash) validated by a new RP with the switch on will now include  name2 , causing a spurious  IDX23029  (or hash mismatch) — the regression the switch is meant to prevent. Suggest: in the legacy branch,  continue  when  queryParamValue  is empty, so it matches the old drop behavior.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1330,11 +1333,12 @@ private static bool IsJkuUriInListOfAllowedDomains(string jkuSetUrl, SignedHttpR
         /// Sanitizes the query params to comply with the specification.
         /// </summary>
         /// <remarks>https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#section-7.5</remarks>
-        private static Dictionary<string, string> SanitizeQueryParams(Uri httpRequestUri)
+        private static Dictionary<string, string> SanitizeQueryParams(Uri httpRequestUri, out bool hadPreviouslyDroppedParam)
         {
             // Remove repeated query params according to the spec: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#section-7-5
             // "If a header or query parameter is repeated on either the outgoing request from the client or the
             // incoming request to the protected resource, that query parameter or header name MUST NOT be covered by the hash and signature."
+            hadPreviouslyDroppedParam = false;
```
pmaytak, line 1341: > Diagnostics — minor
>
> •  FYI:   hadPreviouslyDroppedParam  is not set for the empty-value case ( name2= ), which was also previously dropped. So  IDX23039  may not fire in a case where the new parsing actually changed behavior. ( SignedHttpRequestHandler.cs:1367-1382 )
> •  FYI:  The  IDX23039  warning lists the remaining unsigned params, which may differ from the param that was actually "previously dropped" (the dropped one may be covered and removed). Best-effort diagnostic — fine, just slightly imprecise wording vs. reality.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
```diff
@@ -746,6 +746,32 @@ public void ValidateQClaim(ValidateSignedHttpRequestTheoryData theoryData)
             TestUtilities.AssertFailIfErrors(context);
         }
 
+        [Fact]
+        [ResetAppContextSwitches]
+        public void ValidateQClaim_LegacySwitch_DropsBase64PaddedParam()
```
pmaytak, line 753: >  add (a) a legacy-mode test for an empty-value param — it would lock in intent and surface the bug above; (b) an assertion that  IDX23039  is emitted on the bypass-rejection path; (c) creation-side coverage that a multi- =  param now lands in the  q  claim (currently the one creation test was just simplified to avoid the case).

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
iNinja: Fixed, added a test for it.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
iNinja: Fixed, added a test for it.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
iNinja: Added all three.

iNinja:
Superseded by https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3588. The replacement preserves existing q/h claim canonicalization and limits the change to strict validation of request elements that cannot be covered.

PR #3555, opened 2026-07-16
Validate key-management alg against algorithm policy during JWE decryption

iNinja:
Enforce `ValidAlgorithms` / `AlgorithmValidator` on the JWE key-management `alg` header in all three decrypt paths (modern TVP, legacy TVP, experimental VP).

Previously `alg` was gated only by `IsSupportedAlgorithm` (a capability check). An algorithm excluded from the caller's allowlist — notably `RSA1_5` — was still accepted for CEK unwrap. The gap was uniform across all key-management algorithms.

**Fix:** call `Validators.ValidateAlgorithm` / `ValidateAlgorithmInternal` after the ECDH-ES/dir early-return and before the per-key unwrap loop, in:
- `JsonWebTokenHandler.GetContentEncryptionKeys` (TVP path)
- `JwtSecurityTokenHandler.GetContentEncryptionKeys` (legacy TVP path)
- Experimental `JsonWebTokenHandler.GetContentEncryptionKeys` (VP path)

**Behavior change:** when the alg is rejected by policy, callers now receive `SecurityTokenInvalidAlgorithmException` (fail-fast, before any unwrap attempt) rather than `SecurityTokenDecryptionFailedException` after all keys are tried. Two existing tests updated accordingly.

**Tests added:** two new theory cases in `JsonWebTokenHandler.DecryptToken.UnwrapBehaviourTests.cs` covering `RSA1_5` excluded by `ValidAlgorithms` (must reject) and RSA-OAEP included (must succeed).

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
@@ -137,6 +137,11 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             if (keys is null)
                 return (keys, null); // Cannot iterate over null.
 
+            ValidationResult<string, ValidationError> algResult =
+                Validators.ValidateAlgorithmInternal(jwtToken.Alg, jwtToken, validationParameters, callContext ?? new CallContext());
+            if (!algResult.Succeeded)
+                return (null, algResult.Error!.AddCurrentStackFrame());
+
             var unwrappedKeys = new List<SecurityKey>();
 
             // Pre-generate a placeholder key used as a fallback when an unwrap call
--- a/src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 using System;
@@ -134,13 +134,24 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 || jwtToken.Alg.Equals(SecurityAlgorithms.EcdhEs, StringComparison.Ordinal))
                 return (keys, null);
 
+            // Validate key-management alg before checking keys, so a policy-excluded alg surfaces
+            // as SecurityTokenInvalidAlgorithmException regardless of whether keys were resolved.
+            if (!AppContextSwitches.SkipKeyManagementAlgorithmValidation)
+            {
+                ValidationResult<string, ValidationError> algResult =
+                    Validators.ValidateAlgorithmInternal(jwtToken.Alg, jwtToken, validationParameters, callContext ?? new CallContext());
+                if (!algResult.Succeeded)
+                {
+                    if (LogHelper.IsEnabled(EventLogLevel.Warning))
+                        LogHelper.LogWarning(TokenLogMessages.IDX10726,
+                            LogHelper.MarkAsNonPII(jwtToken.Alg),
+                            LogHelper.MarkAsNonPII(AppContextSwitches.SkipKeyManagementAlgorithmValidationSwitch));
+                    return (null, algResult.Error!.AddCurrentStackFrame());
+                }
+            }
+
             if (keys is null)
                 return (keys, null); // Cannot iterate over null.
-
-            ValidationResult<string, ValidationError> algResult =
-                Validators.ValidateAlgorithmInternal(jwtToken.Alg, jwtToken, validationParameters, callContext ?? new CallContext());
-            if (!algResult.Succeeded)
-                return (null, algResult.Error!.AddCurrentStackFrame());
 
             var unwrappedKeys = new List<SecurityKey>();
 
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
@@ -1332,6 +1332,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 || jwtToken.Alg.Equals(SecurityAlgorithms.EcdhEs, StringComparison.Ordinal))
                 return keys;
 
+            Validators.ValidateAlgorithm(jwtToken.Alg, null, jwtToken, validationParameters);
+
             var unwrappedKeys = new List<SecurityKey>();
 
             // Pre-generate a placeholder key used as a fallback when an unwrap call
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
@@ -1332,7 +1332,8 @@ namespace Microsoft.IdentityModel.JsonWebTokens
                 || jwtToken.Alg.Equals(SecurityAlgorithms.EcdhEs, StringComparison.Ordinal))
                 return keys;
 
-            Validators.ValidateAlgorithm(jwtToken.Alg, null, jwtToken, validationParameters);
+            if (!AppContextSwitches.SkipKeyManagementAlgorithmValidation)
+                JwtTokenUtilities.ValidateKeyManagementAlgorithm(jwtToken.Alg, jwtToken, validationParameters);
 
             var unwrappedKeys = new List<SecurityKey>();
 
--- a/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
+++ b/src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
@@ -113,6 +113,18 @@ namespace Microsoft.IdentityModel.Tokens
         internal static bool UseCaseSensitivePClaimComparison => _useCaseSensitivePClaimComparison ??= (AppContext.TryGetSwitch(UseCaseSensitivePClaimComparisonSwitch, out bool useCaseSensitivePClaimComparison) ? useCaseSensitivePClaimComparison : UseCaseSensitivePClaimComparisonDefault);
 
         /// <summary>
+        /// When enabled, restores the pre-8.19.x behavior where the JWE key-management <c>alg</c> header
+        /// was not validated against the caller's algorithm policy (<c>ValidAlgorithms</c> /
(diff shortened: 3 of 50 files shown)


cpp11nullptr (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
```diff
@@ -1332,6 +1332,8 @@ internal IEnumerable<SecurityKey> GetContentEncryptionKeys(JsonWebToken jwtToken
                 || jwtToken.Alg.Equals(SecurityAlgorithms.EcdhEs, StringComparison.Ordinal))
                 return keys;
 
+            Validators.ValidateAlgorithm(jwtToken.Alg, null, jwtToken, validationParameters);
```
adcosta118: 🤖 *AI-generated review comment — please verify before acting.*

**Breaking-change note for opt-in callers (non-blocking).** This correctly closes the gap where the key-management `alg` bypassed the caller's allowlist. But callers who already set `ValidAlgorithms` typically list only signing + content-encryption (`enc`) algorithms; they will now also need to include the key-management `alg` (e.g. `RSA-OAEP`) or JWE decryption starts failing with `IDX10696`. Please make this explicit in the release notes / migration guidance so existing allowlists are updated proactively.

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -137,6 +137,11 @@ internal ValidationResult<string, ValidationError> DecryptToken(
             if (keys is null)
                 return (keys, null); // Cannot iterate over null.
 
+            ValidationResult<string, ValidationError> algResult =
```
adcosta118: 🤖 *AI-generated review comment — please verify before acting.*

**Minor ordering inconsistency (non-blocking).** Here the alg policy check runs *after* the `if (keys is null) return` guard, so a token with a policy-excluded `alg` but no resolvable decryption key returns as a plain null-keys/decryption failure rather than an invalid-algorithm error. In the throwing paths (`JsonWebTokenHandler.CreateToken.cs` / `JwtSecurityTokenHandler.cs`) `ValidateAlgorithm` runs even when keys is null, so the surfaced error differs across paths. Harmless for security (both reject), but worth aligning if you want consistent error surfacing.

adcosta118 (commented):
🤖 AI-generated review — please verify before acting. Verified all three decrypt paths place the alg check after the dir/ECDH-ES early return and before the unwrap loop; ValidateAlgorithm throws SecurityTokenInvalidAlgorithmException (IDX10696 for ValidAlgorithms, IDX10697 for AlgorithmValidator), matching the updated test expectations. The fix is sound and non-breaking for default configs (no-op when both ValidAlgorithms and AlgorithmValidator are null). Two non-blocking notes inline.

adcosta118 (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
iNinja: Added `Switch.Microsoft.IdentityModel.SkipKeyManagementAlgorithmValidation` as a temporary escape hatch for callers whose `ValidAlgorithms` list doesn't yet include the key-management alg. We also log IDX10726 before rethrowing IDX10696/IDX10697 so the guidance is visible in the exception flow.

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
iNinja: Fixed — the alg check now runs before the `keys is null` guard in the experimental path, consistent with the other two paths.

westin-m (approved):

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -134,6 +134,22 @@ internal ValidationResult<string, ValidationError> DecryptToken(
                 || jwtToken.Alg.Equals(SecurityAlgorithms.EcdhEs, StringComparison.Ordinal))
                 return (keys, null);
 
+            // Validate key-management alg before checking keys, so a policy-excluded alg surfaces
+            // as SecurityTokenInvalidAlgorithmException regardless of whether keys were resolved.
+            if (!AppContextSwitches.SkipKeyManagementAlgorithmValidation)
```
pmaytak, line 139: > The main thing to flag:  ValidAlgorithms  is a single list on  TokenValidationParameters  shared by signature and algorithm validation. Any caller who set  ValidAlgorithms  to their signing algorithms only — and also decrypts JWEs — will now fail decryption because the key-management alg isn't on that list. That's the intended fix, but it's a real breaking change with potentially broad blast radius. Two asks: (1) run the full JWE decryption/validation suite (not just the new tests) to surface any existing tests that set  ValidAlgorithms  without the key-mgmt alg; (2) make sure this ships with a clear CHANGELOG/release note, since the escape hatch only helps people who know it exists.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
```diff
@@ -3337,7 +3337,7 @@ public static TheoryData<JwtTheoryData> ValidateJweTestCases
                             TokenDecryptionKey = KeyingMaterial.DefaultX509Key_2048,
                             AlgorithmValidator = ValidationDelegates.AlgorithmValidatorBuilder(false)
                         },
-                        ExpectedException = new ExpectedException(typeof(SecurityTokenDecryptionFailedException), "IDX10697:"),
+                        ExpectedException = new ExpectedException(typeof(SecurityTokenInvalidAlgorithmException), "IDX10697:"),
```
pmaytak, line 3340: The previous exception type was retryable. This type shouldn't be. Need to verify if it is so.

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -112,6 +112,18 @@ internal static class AppContextSwitches
 
         internal static bool UseCaseSensitivePClaimComparison => _useCaseSensitivePClaimComparison ??= (AppContext.TryGetSwitch(UseCaseSensitivePClaimComparisonSwitch, out bool useCaseSensitivePClaimComparison) ? useCaseSensitivePClaimComparison : UseCaseSensitivePClaimComparisonDefault);
 
+        /// <summary>
+        /// When enabled, restores the pre-8.19.x behavior where the JWE key-management <c>alg</c> header
+        /// was not validated against the caller's algorithm policy (<c>ValidAlgorithms</c> /
+        /// <c>AlgorithmValidator</c>). Callers who set <c>ValidAlgorithms</c> but did not include their
+        /// key-management algorithm (e.g. <c>RSA-OAEP</c>) can set this switch while updating their allowlist.
+        /// </summary>
+        internal const string SkipKeyManagementAlgorithmValidationSwitch = "Switch.Microsoft.IdentityModel.SkipKeyManagementAlgorithmValidation";
```
saurabhsathe-ms, line 121: SkipKeyManagementAlgorithmValidation  has no test, yet it's the sole back-compat mitigation — and it guards the exact scenario that had zero prior coverage: a  ValidAlgorithms  list that omits the key-management alg during JWE decrypt. The existing JWE cases ( JsonWebTokenHandlerTests.cs:3334/3368 ) both include the key-mgmt alg, which is why nothing broke. Can we please add a test for switch on → excluded alg still decrypts (the inverse of  KeyManagementAlgExcludedByPolicy... ). Otherwise, a broken/inverted switch ships undetected.

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
```diff
@@ -134,6 +134,22 @@ internal ValidationResult<string, ValidationError> DecryptToken(
...
+            {
+                ValidationResult<string, ValidationError> algResult =
+                    Validators.ValidateAlgorithmInternal(jwtToken.Alg, jwtToken, validationParameters, callContext ?? new CallContext());
```
saurabhsathe-ms, line 142: The new alg check on this experimental path isn't covered by the two added tests — those go through  ValidateTokenAsync  and the modern  GetContentEncryptionKeys , which throws via  Validators.ValidateAlgorithm . This path is different: it calls  ValidateAlgorithmInternal  and returns a  ValidationResult / AlgorithmValidationError  instead of throwing, so the  !Succeeded  branch and the error return here never run in any test.  JsonWebTokenHandler.DecryptTokenTests.cs  already exercises this exact  DecryptToken  overload but has no  ValidAlgorithms / AlgorithmValidator  case — could you add one there that builds a JWE with a key-mgmt alg like  RSA1_5 , sets  ValidAlgorithms  to a list that excludes it, and asserts the result fails with an  AlgorithmValidationError ? A matching positive case (alg included → succeeds) would round it out and mirror the modern pair

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
iNinja: Full suite passes after rebase onto latest dev8x. CHANGELOG entry added explaining the behavioural change and providing guidance.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandlerTests.cs
iNinja: `SecurityTokenInvalidAlgorithmException` is not in `IsRecoverableException` and will not trigger a `ConfigurationManager` retry, unlike `SecurityTokenDecryptionFailedException` which is retried when the config provides decryption keys. This is intentional — refreshed metadata will not resolve an algorithm policy mismatch. Called out in the CHANGELOG.

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
iNinja: Added `SkipKeyManagementAlgorithmValidation_ExcludedAlgStillDecrypts` covering switch on with an excluded alg.

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.DecryptToken.cs
iNinja: Added two tests in `DecryptionKeysTests.cs` using the experimental VP path directly: excluded alg surfaces a `ValidationError`, allowed alg decrypts successfully.

PR #3558, opened 2026-07-22
Add IgnoreCaseWhenValidatingAudience flag to audience validation

cpp11nullptr:
Introduces a new opt-in TokenValidationParameters.IgnoreCaseWhenValidatingAudience flag (mirrored on the experimental ValidationParameters) that, when enabled, makes audience comparison case-insensitive (OrdinalIgnoreCase). The default remains false, preserving the existing case-sensitive (ordinal) behavior.

AudiencesMatch and AudiencesMatchIgnoringTrailingSlash now select the StringComparison based on the flag, so both exact and trailing-slash-tolerant comparisons honor it. This complements the existing IgnoreTrailingSlashWhenValidatingAudience flag.

Adds PublicAPI.Unshipped.txt entries and unit tests covering case-insensitive matches (exact and combined with trailing-slash tolerance).

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Experimental/Validation/ValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Experimental/Validation/ValidationParameters.cs
@@ -75,6 +75,7 @@ namespace Microsoft.IdentityModel.Tokens.Experimental
             DebugId = other.DebugId;
             IncludeTokenOnFailedValidation = other.IncludeTokenOnFailedValidation;
             IgnoreTrailingSlashWhenValidatingAudience = other.IgnoreTrailingSlashWhenValidatingAudience;
+            IgnoreCaseWhenValidatingAudience = other.IgnoreCaseWhenValidatingAudience;
             SignatureKeyResolver = other.SignatureKeyResolver;
             _signingKeys = other.SigningKeys;
             SignatureKeyValidator = other.SignatureKeyValidator;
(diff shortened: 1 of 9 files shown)


robezirg (approved):

src/Microsoft.IdentityModel.Tokens/Experimental/Validation/ValidationParameters.cs
```diff
@@ -265,6 +266,13 @@ public virtual ClaimsIdentity CreateClaimsIdentity(SecurityToken securityToken,
         [DefaultValue(true)]
         public bool IgnoreTrailingSlashWhenValidatingAudience { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets a boolean that controls if case is ignored when validating the audience.
+        /// The default is <c>false</c>, meaning audience comparison is case-sensitive (ordinal).
+        /// </summary>
+        [DefaultValue(false)]
```
MZOLN, line 273: The [DefaultValue] attribution brings no functional value.

cpp11nullptr: As I can see there are both defined `[DefaultValue]` (including the one with **false** value) and omitted ones for properties this class but I'll keep it at least to be consistent with related `IgnoreTrailingSlashWhenValidatingAudience` setting.

MZOLN (approved):

Issue #1840, opened 2022-04-25
[Bug] ClaimsIdentity Actor not serialized into JWTs


gustavdw:
**Which version of Microsoft.IdentityModel are you using?**
Note that to get help, you need to run the latest version. 
Im using OpenIddict 3.1.1

**Where is the issue?**
 * [x ] M.IM.JsonWebTokens
 * [ ] M.IM.KeyVaultExtensions
 * [ ] M.IM.Logging
 * [ ] M.IM.ManagedKeyVaultSecurityKey
 * [ ] M.IM.Protocols
 * [x ] M.IM.Protocols.OpenIdConnect
 * [ ] M.IM.Protocols.SignedHttpRequest
 * [ ] M.IM.Protocols.WsFederation
 * [ ] M.IM.TestExtensions
 * [ ] M.IM.Tokens
 * [ ] M.IM.Tokens.Saml
 * [ ] M.IM.Validators
 * [ ] M.IM.Xml
 * [ ] S.IM.Tokens.Jwt
 * Other (please describe)

**Is this a new or an existing app?**
The app is in production and I haven't upgraded Microsoft.IdentityModel.*, but started seeing this issue

**Repro**
Using OpenIddict, see issue here: https://github.com/openiddict/openiddict-core/issues/1426 

**Expected behavior**
The ClaimsIdentity Actor is serialized into the JWT (as an json object with a subset of claims attached to the actor claims identity)

**Actual behavior**
The ClaimsIdentity Actor is ignored

**Possible solution**
None

brentschmaltz:
@gustavdw as @kevinchalet pointed out, IdentityModel is not adding ClaimsIdentity.Actor to the JWT.
This is a bug.

kevinchalet:
Great, thanks for taking a look, @brentschmaltz.

Note: it's possible this is not the only place that needs an update for `ClaimsIdentity.Actor` to work correctly: the reading part seems to use `ClaimTypes.Actor` and not the `actort` claim the previous IM versions used to use, so I think it won't work correctly when claims mapping is disabled:

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/799a162276cb699c6634f12077ab7af0646cf710/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L699-L710

[...]

PR #3560, opened 2026-07-23
Add actor (act) claim serialization to JsonWebTokenHandler

saurabhsathe-ms:
# Enable Actor Claim Support in Wilson - Serialization

- [x] If any gains or losses in performance are possible, you've included benchmarks for your changes.
- [x] There's an open issue for the PR that you are making.

## Why

`JwtSecurityTokenHandler` has always serialized `ClaimsIdentity.Actor` into a JWT, but the
faster `JsonWebTokenHandler` never did — it processed `Subject.Claims` and ignored
`Subject.Actor` entirely. This PR closes that gap on the **write path**, emitting the
RFC 8693 `act` claim. Deserialization follows in stacked PR #3571.

Fixes part of #1840.

## What changed

- `JsonWebTokenHandler` now serializes the actor as the RFC 8693 **`act`** JSON object,
  sourced from `Subject.Actor` or a `ClaimsIdentity` under `Claims["act"]`. The `act` object
  contains **identity claims only** — the handler never injects `exp`/`nbf`/`iat` into it.
- New public  JsonWebTokenHandler.MaxActorChainLength  (default  1 , min  1 ) bounds how many nested actor levels are materialized as JSON objects; deeper chains degrade to a JSON-text string. This degradation is still bounded by the library's global JSON depth limit of 64 (shared by all serialized JSON): an actor nested beyond depth 64 fails fast with  IDX10815  rather than emitting pathological JSON. 
- New public **`JwtRegisteredClaimNames.Act = "act"`** (mirrored into
  `System.IdentityModel.Tokens.Jwt`).
- The `act` member is **single-sourced**: a raw `act` value (a non-`ClaimsIdentity`
  `Claims["act"]`, or a literal `act` claim on `Subject.Claims`) never produces a duplicate
  `act` — the structural actor wins.
- New message `IDX14317` (`MaxActorChainLength` must be ≥ 1).

## Notes

- Scope: writes the modern RFC 8693 `act` (JSON object); it does **not** write the legacy
  `actort` (unsigned-JWT string) that `JwtSecurityTokenHandler` emits.
- No shipped internal API **signatures** changed. `WriteJwsPayload` and `AddSubjectClaims`
  bodies were updated only to keep `act` single-sourced.

Full design rationale, precedence rules, performance analysis, and test coverage live in the
actor-claim README:
[`src/Microsoft.IdentityModel.JsonWebTokens/ActorClaim.md`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/ssathe/serializeActorClaim/src/Microsoft.IdentityModel.JsonWebTokens/ActorClaim.md).


test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimSerializationTests.cs
```diff
@@ -0,0 +1,598 @@
...
+                    Audience = "https://api.example.com",
+                    Expires = DateTime.UtcNow.AddHours(1),
+                    SigningCredentials = Default.AsymmetricSigningCredentials,
+                };
+                var token = tokenHandler.CreateToken(tokenDescriptor);
+                JsonWebToken decodedToken = tokenHandler.ReadJsonWebToken(token);
+
+                // Verify actor claim exists in the token
+                Assert.True(decodedToken.Payload.HasClaim("act"), "JWT token should contain 'act' claim");
+
```
Toidi357, line 427: duplicated lines here

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimSerializationTests.cs
```diff
@@ -0,0 +1,598 @@
...
+    public class ActClaimTestsCollection { }
+
+    [Collection("ActClaimTests")]
+    public class ActClaimSerializationTests : IDisposable
+    {
+        // Reset the process-wide MaxActorChainLength after every test for isolation.
+        public void Dispose() => JsonWebTokenHandler.MaxActorChainLength = 1;
+
+        [Fact]
+        public void MaxActorChainLength_DefaultIsOne()
```
Toidi357, line 25: do we really need this test?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimSerializationTests.cs
```diff
@@ -0,0 +1,598 @@
...
+            var token = handler.CreateToken(tokenDescriptor);
+            var decoded = handler.ReadJsonWebToken(token);
+
+            // The top-level payload still gets default temporal claims (feature unchanged there).
+            Assert.True(decoded.Payload.HasClaim("exp"), "top-level payload should contain 'exp'");
+            Assert.True(decoded.Payload.HasClaim("nbf"), "top-level payload should contain 'nbf'");
+            Assert.True(decoded.Payload.HasClaim("iat"), "top-level payload should contain 'iat'");
+
+            // The act object contains identity claims only.
+            var act = decoded.Payload.GetValue<JsonElement>("act");
```
Toidi357, line 76: maybe this test should try to add a "exp" or other claim to the "act", then ensure that it either throws or gets ignored? because right now the Assert.False's will definitely pass since you are only adding "sub" (line 46)

Toidi357 (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimSerializationTests.cs
saurabhsathe-ms: Yes I would want to have this as default so that any modifications to the default value is known to break exisiing behavior

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimSerializationTests.cs
saurabhsathe-ms: the exp/nbf these claims are added by the code in Wilson. Hence this test ensures they are not added by default by the code. I have updated the tests. Thank you for your review

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimSerializationTests.cs
saurabhsathe-ms: Removed the duplicate lines. Thank you for the review @Toidi357

benchmark/Microsoft.IdentityModel.Benchmarks/ActorClaimBenchmarks.cs
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Security.Claims;
+using BenchmarkDotNet.Attributes;
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.IdentityModel.Benchmarks
```
westin-m: nit: file-scoped namespaces in new files

westin-m (approved):

benchmark/Microsoft.IdentityModel.Benchmarks/ActorClaimBenchmarks.cs
saurabhsathe-ms: Done!

debchoudhury-id4s:
Is the PR description intentionally this detailed? Is it aimed as a replacement for some design documentation customers need to be aware of?

[Nit]: consider making the PR description more succinct, simple while moving design decisions to a README if it makes sense

saurabhsathe-ms:
> Is the PR description intentionally this detailed? Is it aimed as a replacement for some design documentation customers need to be aware of?
> 
> [Nit]: consider making the PR description more succinct, simple while moving design decisions to a README if it makes sense

Thank you so much for reviewing this @debchoudhury-id4s. I have kept it detailed only for this particular PR to help the reviewers get any context they need from any perspective! I can make it shorter but just for this particular PR I prefer it to be longer and detailed. We do have external customers who read this for more context!

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
```diff
@@ -1072,6 +1115,122 @@ internal static byte[] WriteJweHeader(SecurityTokenDescriptor tokenDescriptor)
...
+
+            writer.WritePropertyName(ActClaimType);
+            WriteActorValue(ref writer, actor, remainingActorLevels);
+        }
+
+        // Writes a single actor as the value of an "act" member: a nested JSON object while within the
+        // configured chain length, otherwise the remaining subtree as a JSON-text string.
+        private static void WriteActorValue(ref Utf8JsonWriter writer, ClaimsIdentity actor, int remainingActorLevels)
+        {
+            if (remainingActorLevels > 0)
```
debchoudhury-id4s, line 1165: Is this the recursive termination check?  Is there a need to factor in the `MaxActorChainLength` instead of using the `int.MaxValue`? Also have you added sufficient unit tests to guard against cyclic loops?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
```diff
@@ -668,11 +703,17 @@ internal static void WriteJwsPayload(
             // Duplicates are resolved according to the following priority:
             // SecurityTokenDescriptor.{Audience/Audiences, Issuer, Expires, IssuedAt, NotBefore}, SecurityTokenDescriptor.Claims, SecurityTokenDescriptor.Subject.Claims
             // SecurityTokenDescriptor.Claims are KeyValuePairs<string,object>, whereas SecurityTokenDescriptor.Subject.Claims are System.Security.Claims.Claim and are processed differently.
+            bool isActorFound = false;
 
             if (tokenDescriptor.Claims != null && tokenDescriptor.Claims.Count > 0)
             {
                 foreach (KeyValuePair<string, object> kvp in tokenDescriptor.Claims)
                 {
+                    if (kvp.Key.Equals(ActClaimType, StringComparison.Ordinal) && kvp.Value is ClaimsIdentity)
```
JoshLozensky: Is it possible for a key to have the ActClaimType but not have the value as a ClaimsIdentity? If so, wouldn't we run the risk of recording "act" multiple times as this if wouldn't be hit but an act key would be recorded at line 795?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
```diff
@@ -27,6 +27,41 @@ public partial class JsonWebTokenHandler : TokenHandler
     {
         private static readonly SecurityTokenDescriptor s_emptyTokenDescriptor = new();
 
+        // The actor is always serialized under the RFC 8693 "act" claim. The claim name
+        // is intentionally fixed (not configurable) to avoid actor claims shadowing
+        // registered claims such as iss/aud/exp.
+        internal const string ActClaimType = "act";
```
JoshLozensky: Why not record act as JwtRegisteredClaimNames.Act?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
saurabhsathe-ms: Thank you for your review @JoshLozensky . Yes, if  Claims["act"]  held a non- ClaimsIdentity  value and  Subject.Actor  was set, the loop wrote  act  verbatim and  WriteActor  also emitted  act  from  Subject.Actor , producing a duplicate key ( Utf8JsonWriter  doesn't dedupe). Fixed: the  Claims   act  key now takes precedence and exactly one  act  is written. Added a regression test asserting a single top-level  act  member. Thanks!

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
saurabhsathe-ms: Great suggestion. Added a public  Act = "act"  (RFC 8693), mirrored it into  System.IdentityModel.Tokens.Jwt , and  ActClaimType  now derives from it. Updated  PublicAPI.Unshipped.txt  in both packages

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
saurabhsathe-ms: Thanks @debchoudhury-id4s — three quick answers:

1. Is this the recursive termination check?
Not quite — the actual stop condition is  actor.Actor is null  (the end of the delegation chain). The  remainingActorLevels > 0  check only decides object vs. JSON-string for a given level; it doesn't end the recursion. So recursion always terminates when the chain ends.

2. Factor in  MaxActorChainLength  instead of  int.MaxValue ?
 int.MaxValue  is intentional. Past  MaxActorChainLength  we degrade the remaining actor subtree to a JSON-text string, losslessly — we still write every level, just as a string instead of nested objects. Passing  MaxActorChainLength  there would re-truncate and silently drop actor levels, which is the opposite of the intended "don't drop caller data" behavior.  int.MaxValue  simply means "expand whatever remains, in full."

3. Tests to guard against cyclic loops?
A cycle can never reach the serializer:  ClaimsIdentity.Actor 's setter throws  InvalidOperationException  on any circular reference (self or mutual) via its  IsCircular  check, so the  Actor  chain is always finite and acyclic — the  int.MaxValue  recursion is guaranteed to terminate. Added a test documenting this (self + mutual cycle):  ActorChain_CyclicClaimsIdentityActor_IsRejectedByClaimsIdentity .

debchoudhury-id4s: Would the current code still recursively serialize every remaining actor into the JSON string after writing one object level when `MaxActorChainLength = 1`, eventually throwing `IDX10815`? 

Could you verify this hypothesis with the following unit test?

  ```
     JsonWebTokenHandler.MaxActorChainLength = 1;
     var subject = new ClaimsIdentity(new[] { new Claim("sub", "subject") });
     ClaimsIdentity current = subject;

     for (int i = 0; i < 70; i++)
     {
         var actor = new ClaimsIdentity(new[] { new Claim("sub", $"actor-{i}") });
         current.Actor = actor;
         current = actor;
     }

     var exception = Assert.Throws<InvalidOperationException>(() =>
         new JsonWebTokenHandler().CreateToken(new SecurityTokenDescriptor
         {
             Subject = subject,
             SigningCredentials = Default.AsymmetricSigningCredentials,
         }));

     Assert.Contains("IDX10815", exception.Message);
  ```

JoshLozensky (approved):

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.CreateToken.cs
saurabhsathe-ms: Confirmed, it isn't actually unbounded. The degraded-string expansion still flows through Wilson's global JSON writer guard ( JsonSerializerPrimitives , max depth 64), so a delegation chain beyond ~64 levels fails fast with  IDX10815  — the same depth cap Wilson enforces on every token it writes. So deep chains get a clear, consistent error instead of runaway output, and 64-deep delegation is far past any real case (RFC 8693 uses only the top actor; default  MaxActorChainLength  is 1). Here is the test: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/ssathe/serializeActorClaim/test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimSerializationTests.cs#L245-L276

debchoudhury-id4s: if exception gets thrown doesn't it contradict the message [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3560/changes#diff-0a9cacb78cbfe0809b7e3156a419ee4ecff64b21b94241d79594bbd718edbe2eR104) _"Beyond the limit:** the remaining actor subtree **degrades to a JSON-text string** (never a throw)"_

debchoudhury-id4s: discussed on chat to correct the messaging

debchoudhury-id4s (approved):
approved with suggestions

debchoudhury-id4s (approved):

PR #3561, opened 2026-07-23
Handle percent-encoding hex case in SHR p-claim validation

debchoudhury-id4s:
## Problem

Signed HTTP Request `p` claim validation rejects equivalent paths when percent-encoded octets use different hexadecimal letter casing. For example, `/foo%2Fbar` does not match `/foo%2fbar`, even though RFC 3986 defines those triplets as equivalent.

Literal path letters must remain case-sensitive by default. Encoded reserved characters must also remain distinct from literal delimiters, so `%2F` must not match `/`.

## Requirements

1. Treat hexadecimal letters inside valid percent-encoded triplets as case-insensitive during `p` claim validation.
2. Preserve literal path-letter behavior controlled by `UseCaseSensitivePClaimComparison`.
3. Keep encoded characters distinct from literal delimiters and do not URL-decode values.
4. Leave malformed percent sequences unchanged.
5. Keep `p` claim creation output unchanged.

## Code Changes

1. `ValidatePClaim` trims the claim and request path with `ReadOnlySpan<char>` to avoid allocating comparison strings.
1. The trimmed spans are first compared directly with the configured `StringComparison`.
1. After an ordinal mismatch, `ArePClaimValuesEqual` traverses both spans and ignores hexadecimal letter casing only inside aligned, valid `%XX` triplets. Values are not URL-decoded.
1. If validation fails, `IDX23011` reports the original trimmed values without changing their percent-encoding casing.

### Functional results

| Input | Flag status | Result |
|---|---|---|
| Request `/foo%2Fbar`; claim `/foo%2fbar` | Enabled | Pass |
| Request `/foo%2fbar`; claim `/foo%2Fbar` | Enabled | Pass |
| Request `/foo%2Fbar%ABbaz`; claim `/foo%2fbar%abbaz` | Enabled | Pass |
| Request `/Foo%2Fbar`; claim `/foo%2fbar` | Enabled | Fail — `IDX23011` |
| Request `/Foo%2Fbar`; claim `/foo%2fbar` | Disabled | Pass |
| Request `/foo%2Fbar`; claim `/foo/bar` | Enabled | Fail — `IDX23011` |
| Request `/foo/bar`; claim `/foo%2Fbar` | Enabled | Fail — `IDX23011` |
| Request `/foo%252Fbar`; claim `/foo%2Fbar` | Enabled | Fail — `IDX23011` |
| Request `/foo%25zz`; malformed claim `/foo%zz` | Enabled | Fail — `IDX23011` |

All 58 targeted `ValidatePClaim` tests pass on .NET 8.

#### Performance benchmark

- **A** = `dev` & **B** = `debchoudhury/shr-p-claim-percent-encoding`
- BenchmarkDotNet 0.13.12, .NET 8.0.29, X64 RyuJIT, MediumRun with 4 launches, 10 warmups, and 15 measured iterations.
- B:
  - allocates 0 bytes in every measured scenario (A allocates 144–240 bytes for the two trimmed strings)
  - is faster in 22 of the 24 comparison combinations.

| Scenario | Comparison | A mean | B mean | B/A | A allocated | B allocated |
|---|---|---:|---:|---:|---:|---:|
| Plain match | Ordinal | 53.75 ns | 7.99 ns | 0.15x | 224 B | 0 B |
| Plain match | OrdinalIgnoreCase | 64.43 ns | 20.76 ns | 0.32x | 224 B | 0 B |
| Plain mismatch | Ordinal | 55.37 ns | 44.88 ns | 0.81x | 224 B | 0 B |
| Plain mismatch | OrdinalIgnoreCase | 64.55 ns | 20.54 ns | 0.32x | 224 B | 0 B |
| Different length | Ordinal | 48.72 ns | 7.25 ns | 0.15x | 208 B | 0 B |
| Different length | OrdinalIgnoreCase | 49.08 ns | 6.19 ns | 0.13x | 208 B | 0 B |
| Uppercase triplets | Ordinal | 56.17 ns | 8.14 ns | 0.14x | 240 B | 0 B |
| Uppercase triplets | OrdinalIgnoreCase | 60.00 ns | 18.48 ns | 0.31x | 240 B | 0 B |
| Lowercase triplets on one side | Ordinal | 53.69 ns | 56.81 ns | 1.06x | 240 B | 0 B |
| Lowercase triplets on one side | OrdinalIgnoreCase | 65.46 ns | 18.75 ns | 0.29x | 240 B | 0 B |
| Lowercase triplets on both sides | Ordinal | 57.66 ns | 8.09 ns | 0.14x | 240 B | 0 B |
| Lowercase triplets on both sides | OrdinalIgnoreCase | 62.95 ns | 18.57 ns | 0.29x | 240 B | 0 B |
| Multiple mixed-case triplets | Ordinal | 56.03 ns | 56.87 ns | 1.02x | 240 B | 0 B |
| Multiple mixed-case triplets | OrdinalIgnoreCase | 64.59 ns | 18.47 ns | 0.29x | 240 B | 0 B |
| Different valid triplets | Ordinal | 52.41 ns | 33.80 ns | 0.64x | 240 B | 0 B |
| Different valid triplets | OrdinalIgnoreCase | 58.46 ns | 15.48 ns | 0.26x | 240 B | 0 B |
| Encoded slash versus literal slash | Ordinal | 53.52 ns | 7.13 ns | 0.13x | 232 B | 0 B |
| Encoded slash versus literal slash | OrdinalIgnoreCase | 46.25 ns | 6.12 ns | 0.13x | 232 B | 0 B |
| Double-encoded versus single-encoded slash | Ordinal | 52.85 ns | 6.93 ns | 0.13x | 240 B | 0 B |
| Double-encoded versus single-encoded slash | OrdinalIgnoreCase | 46.76 ns | 6.09 ns | 0.13x | 240 B | 0 B |
| Malformed triplet mismatch | Ordinal | 50.92 ns | 40.78 ns | 0.80x | 240 B | 0 B |
| Malformed triplet mismatch | OrdinalIgnoreCase | 63.01 ns | 21.70 ns | 0.34x | 240 B | 0 B |
| Truncated triplet mismatch | Ordinal | 39.75 ns | 31.66 ns | 0.80x | 144 B | 0 B |
| Truncated triplet mismatch | OrdinalIgnoreCase | 43.88 ns | 16.28 ns | 0.37x | 144 B | 0 B |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ See the [releases](https://github.com/AzureAD/azure-activedirectory-identitymode
 ====
 ## Bug Fixes
 - Align Signed HTTP Request `p` (path) claim comparison with RFC 3986 section 3.3 by comparing the path case-sensitively; previously `/Admin/resource` matched `/admin/resource`. An AppContext opt out is available during transition. See [PR #3526](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3526).
+- Normalize hexadecimal letter casing inside percent-encoded triplets during Signed HTTP Request `p` claim validation. Literal path-letter comparison remains controlled by `UseCaseSensitivePClaimComparison`, and `p` claim creation output is unchanged.
 
 7.7.2
 ====
--- a/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
+++ b/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
@@ -10,7 +10,6 @@ using System.Security.Cryptography;
 using System.Text;
 using System.Text.Encodings.Web;
 using System.Text.Json;
-using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.IdentityModel.Abstractions;
@@ -27,8 +26,6 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
     /// <remarks>The handler implementation is based on 'A Method for Signing HTTP Requests for OAuth' specification.</remarks>
     public class SignedHttpRequestHandler
     {
-        private const string PercentEncodedTripletPattern = "%[0-9A-Fa-f]{2}";
-
         // https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#section-3.2
         // "Encodes the name and value of the header as "name: value" and appends it to the string buffer separated by a newline "\n" character."
         private readonly string _newlineSeparator = "\n";
@@ -821,40 +818,26 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
             if (!signedHttpRequest.TryGetPayloadValue(SignedHttpRequestClaimTypes.P, out string pClaimValue) || pClaimValue == null)
                 throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidPClaimException(LogHelper.FormatInvariant(LogMessages.IDX23003, LogHelper.MarkAsNonPII(SignedHttpRequestClaimTypes.P))));
 
-            // relax comparison by trimming start and ending forward slashes
-            pClaimValue = pClaimValue.Trim('/');
-            var expectedPClaimValue = httpRequestUri.AbsolutePath.Trim('/');
+            // Relax comparison by trimming start and ending forward slashes without allocating new strings.
+            ReadOnlySpan<char> pClaimValueSpan = pClaimValue.AsSpan().Trim('/');
+            ReadOnlySpan<char> expectedPClaimValueSpan = httpRequestUri.AbsolutePath.AsSpan().Trim('/');
 
             // URI path components are case-sensitive per RFC 3986 section 3.3, so the comparison is ordinal (case-sensitive) by default.
             // The AppContext switch allows a consumer to restore the legacy case-insensitive comparison during transition.
             var comparison = AppContextSwitches.UseCaseSensitivePClaimComparison ? StringComparison.Ordinal : StringComparison.OrdinalIgnoreCase;
-            // Fast path: identical values such as "/path" avoid normalization.
-            if (!string.Equals(expectedPClaimValue, pClaimValue, comparison))
-            {
-                // Only equal-length ordinal values with '%' on both sides can differ solely by hex case, such as "%2F" and "%2f".
-                bool mayMatchAfterNormalization =
-                    comparison == StringComparison.Ordinal &&
-                    expectedPClaimValue.Length == pClaimValue.Length &&
-                    expectedPClaimValue.IndexOf('%') >= 0 &&
-                    pClaimValue.IndexOf('%') >= 0;
+            if (expectedPClaimValueSpan.Equals(pClaimValueSpan, comparison))
+                return;
 
-                // A failure cannot be caused only by percent-triplet hex casing because values such as
-                // "foo%2Fbar" and "foo%2fbar" compare equal after normalization. Normalize both
-                // values before inserting them into the IDX23011 error message. For example,
-                // "foo%2fbar" and "goo%2Fbar" are logged as "foo%2Fbar" and "goo%2Fbar".
-                // This removes the irrelevant "%2f" versus "%2F" difference and makes the real
-                // "f" versus "g" mismatch clear.
-                expectedPClaimValue = NormalizePercentEncodingHexCase(expectedPClaimValue);
-                pClaimValue = NormalizePercentEncodingHexCase(pClaimValue);
+            // RFC 3986 requires an additional check because ordinal comparison distinguishes equivalent percent-encoded hexadecimal casing.
+            if (comparison == StringComparison.Ordinal &&
+                ArePClaimValuesEqual(expectedPClaimValueSpan, pClaimValueSpan))
+                return;
 
-                if (mayMatchAfterNormalization &&
-                    string.Equals(expectedPClaimValue, pClaimValue, StringComparison.Ordinal))
-                {
-                    return;
-                }
-
-                throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidPClaimException(LogHelper.FormatInvariant(LogMessages.IDX23011, LogHelper.MarkAsNonPII(SignedHttpRequestClaimTypes.P), expectedPClaimValue, pClaimValue)));
-            }
+            throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidPClaimException(LogHelper.FormatInvariant(
+                LogMessages.IDX23011,
+                LogHelper.MarkAsNonPII(SignedHttpRequestClaimTypes.P),
+                expectedPClaimValueSpan.ToString(),
+                pClaimValueSpan.ToString())));
         }
 
         /// <summary>
@@ -1312,16 +1295,50 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
             return Base64UrlEncoder.Encode(hashedBytes);
         }
 
-        // Uppercases hexadecimal letters only within valid percent-encoded triplets.
-        // It does not URL-decode values or modify malformed sequences.
-        // For example, "foo%2fbar%zz" becomes "foo%2Fbar%zz".
-        private static string NormalizePercentEncodingHexCase(string value)
+        // Compares p-claim path values ordinally while ignoring hexadecimal letter casing
+        // only within aligned, valid percent-encoded triplets. The triplets are not decoded:
+        // "foo%2Fbar" equals "foo%2fbar", but "foo%2Fbar" does not equal "foo/bar",
+        // and malformed "foo%2Gbar" does not equal "foo%2gbar".
+        private static bool ArePClaimValuesEqual(ReadOnlySpan<char> expectedValue, ReadOnlySpan<char> actualValue)
         {
-            return Regex.Replace(
-                value,
-                PercentEncodedTripletPattern,
-                static match => match.Value.ToUpperInvariant(),
-                RegexOptions.CultureInvariant);
+            if (expectedValue.Length != actualValue.Length)
+                return false;
+
+            for (int i = 0; i < expectedValue.Length; i++)
+            {
+                if (expectedValue[i] == '%' &&
+                    actualValue[i] == '%' &&
+                    i + 2 < expectedValue.Length)
+                {
+                    char expectedFirstHexCharacter = expectedValue[i + 1];
+                    char expectedSecondHexCharacter = expectedValue[i + 2];
+                    char actualFirstHexCharacter = actualValue[i + 1];
+                    char actualSecondHexCharacter = actualValue[i + 2];
+
+                    if (Uri.IsHexDigit(expectedFirstHexCharacter) &&
+                        Uri.IsHexDigit(expectedSecondHexCharacter) &&
+                        Uri.IsHexDigit(actualFirstHexCharacter) &&
+                        Uri.IsHexDigit(actualSecondHexCharacter))
+                    {
+                        // RFC 3986 section 6.2.2.1 defines uppercase hexadecimal letters as the canonical form for percent-encoded triplets.
+                        if (char.ToUpperInvariant(expectedFirstHexCharacter) != char.ToUpperInvariant(actualFirstHexCharacter) ||
+                            char.ToUpperInvariant(expectedSecondHexCharacter) != char.ToUpperInvariant(actualSecondHexCharacter))
+                        {
+                            return false;
+                        }
+
+                        i += 2;
+                        continue;
+                    }
+                }
+
+                if (expectedValue[i] != actualValue[i])
+                {
+                    return false;
+                }
+            }
+
+            return true;
         }
 
         /// <summary>
--- a/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
+++ b/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
@@ -1286,67 +1286,35 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
 
         private static string NormalizePercentEncodingHexCase(string value)
         {
-            for (int i = 0; i < value.Length - 2; i++)
+            char[] normalizedValue = null;
+            for (int i = 0; i + 2 < value.Length; i++)
             {
                 if (value[i] != '%')
                     continue;
 
                 char firstHexCharacter = value[i + 1];
                 char secondHexCharacter = value[i + 2];
-                bool firstCharacterIsHex =
-                    (firstHexCharacter >= '0' && firstHexCharacter <= '9') ||
-                    (firstHexCharacter >= 'A' && firstHexCharacter <= 'F') ||
-                    (firstHexCharacter >= 'a' && firstHexCharacter <= 'f');
-                bool secondCharacterIsHex =
-                    (secondHexCharacter >= '0' && secondHexCharacter <= '9') ||
-                    (secondHexCharacter >= 'A' && secondHexCharacter <= 'F') ||
-                    (secondHexCharacter >= 'a' && secondHexCharacter <= 'f');
-
-                if (!firstCharacterIsHex || !secondCharacterIsHex)
+                if (!Uri.IsHexDigit(firstHexCharacter) || !Uri.IsHexDigit(secondHexCharacter))
                     continue;
 
-                bool firstCharacterNeedsNormalization = firstHexCharacter >= 'a' && firstHexCharacter <= 'f';
-                bool secondCharacterNeedsNormalization = secondHexCharacter >= 'a' && secondHexCharacter <= 'f';
-                if (!firstCharacterNeedsNormalization && !secondCharacterNeedsNormalization)
+                if (IsLowerHex(firstHexCharacter) || IsLowerHex(secondHexCharacter))
                 {
-                    i += 2;
-                    continue;
+                    normalizedValue ??= value.ToCharArray();
+
+                    if (IsLowerHex(firstHexCharacter))
+                        normalizedValue[i + 1] = (char)(firstHexCharacter - ('a' - 'A'));
+
+                    if (IsLowerHex(secondHexCharacter))
+                        normalizedValue[i + 2] = (char)(secondHexCharacter - ('a' - 'A'));
                 }
 
-                char[] normalizedValue = value.ToCharArray();
-                for (int j = i; j < normalizedValue.Length - 2; j++)
-                {
-                    if (normalizedValue[j] != '%')
-                        continue;
-
-                    firstHexCharacter = normalizedValue[j + 1];
-                    secondHexCharacter = normalizedValue[j + 2];
-                    firstCharacterIsHex =
-                        (firstHexCharacter >= '0' && firstHexCharacter <= '9') ||
-                        (firstHexCharacter >= 'A' && firstHexCharacter <= 'F') ||
-                        (firstHexCharacter >= 'a' && firstHexCharacter <= 'f');
-                    secondCharacterIsHex =
-                        (secondHexCharacter >= '0' && secondHexCharacter <= '9') ||
-                        (secondHexCharacter >= 'A' && secondHexCharacter <= 'F') ||
-                        (secondHexCharacter >= 'a' && secondHexCharacter <= 'f');
-
-                    if (!firstCharacterIsHex || !secondCharacterIsHex)
-                        continue;
-
-                    if (firstHexCharacter >= 'a' && firstHexCharacter <= 'f')
-                        normalizedValue[j + 1] = (char)(firstHexCharacter - ('a' - 'A'));
-
-                    if (secondHexCharacter >= 'a' && secondHexCharacter <= 'f')
-                        normalizedValue[j + 2] = (char)(secondHexCharacter - ('a' - 'A'));
-
-                    j += 2;
-                }
-
-                return new string(normalizedValue);
+                i += 2;
             }
 
-            return value;
+            return normalizedValue == null ? value : new string(normalizedValue);
         }
+
+        private static bool IsLowerHex(char character) => character >= 'a' && character <= 'f';
 
         /// <summary>
         /// Ensures that the <paramref name="uri"/> is <see cref="UriKind.Absolute"/>.
--- a/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
+++ b/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
@@ -827,6 +827,9 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
             var comparison = AppContextSwitches.UseCaseSensitivePClaimComparison ? StringComparison.Ordinal : StringComparison.OrdinalIgnoreCase;
             if (!ArePClaimValuesEqual(expectedPClaimValue, pClaimValue, comparison))
             {
+                // ArePClaimValuesEqual already treats percent-triplet hex digits as case-insensitive, so a failure here is never caused by hex casing alone.
+                // Normalize both values to the canonical uppercase form only for the IDX23011 message, so it does not surface a hex-case difference the comparison deliberately ignored.
+                // Example: request path "/Foo%2Fbar" versus claim "/foo%2fbar" fails on the literal "F" vs "f"; without normalization the message would also show "%2F" vs "%2f" and point at a non-cause.
                 expectedPClaimValue = NormalizePercentEncodingHexCase(expectedPClaimValue);
                 pClaimValue = NormalizePercentEncodingHexCase(pClaimValue);
                 throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidPClaimException(LogHelper.FormatInvariant(LogMessages.IDX23011, LogHelper.MarkAsNonPII(SignedHttpRequestClaimTypes.P), expectedPClaimValue, pClaimValue)));
--- a/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
+++ b/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
@@ -1297,14 +1297,14 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
                 if (!Uri.IsHexDigit(firstHexCharacter) || !Uri.IsHexDigit(secondHexCharacter))
                     continue;
 
-                if (IsLowerHex(firstHexCharacter) || IsLowerHex(secondHexCharacter))
+                if (char.IsLower(firstHexCharacter) || char.IsLower(secondHexCharacter))
                 {
                     normalizedValue ??= value.ToCharArray();
 
-                    if (IsLowerHex(firstHexCharacter))
+                    if (char.IsLower(firstHexCharacter))
                         normalizedValue[i + 1] = (char)(firstHexCharacter - ('a' - 'A'));
 
-                    if (IsLowerHex(secondHexCharacter))
+                    if (char.IsLower(secondHexCharacter))
                         normalizedValue[i + 2] = (char)(secondHexCharacter - ('a' - 'A'));
                 }
 
@@ -1313,8 +1313,6 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
 
             return normalizedValue == null ? value : new string(normalizedValue);
         }
-
-        private static bool IsLowerHex(char character) => character >= 'a' && character <= 'f';
 
         /// <summary>
         /// Ensures that the <paramref name="uri"/> is <see cref="UriKind.Absolute"/>.
--- a/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
+++ b/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
@@ -819,14 +819,18 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
                 throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidPClaimException(LogHelper.FormatInvariant(LogMessages.IDX23003, LogHelper.MarkAsNonPII(SignedHttpRequestClaimTypes.P))));
 
             // relax comparison by trimming start and ending forward slashes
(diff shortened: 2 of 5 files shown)


src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -819,8 +819,8 @@ internal virtual void ValidatePClaim(JsonWebToken signedHttpRequest, SignedHttpR
                 throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidPClaimException(LogHelper.FormatInvariant(LogMessages.IDX23003, LogHelper.MarkAsNonPII(SignedHttpRequestClaimTypes.P))));
 
             // relax comparison by trimming start and ending forward slashes
-            pClaimValue = pClaimValue.Trim('/');
-            var expectedPClaimValue = httpRequestUri.AbsolutePath.Trim('/');
+            pClaimValue = NormalizePercentEncodingHexCase(pClaimValue.Trim('/'));
+            var expectedPClaimValue = NormalizePercentEncodingHexCase(httpRequestUri.AbsolutePath.Trim('/'));
 
             // URI path components are case-sensitive per RFC 3986 section 3.3, so the comparison is ordinal (case-sensitive) by default.
```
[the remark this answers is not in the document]
debchoudhury-id4s: Keeping the current behavior. `ValidatePClaim` already reports the transformed values used for comparison: on `dev`, both values are trimmed before being passed to `IDX23011`. This change adds percent-triplet normalization to the same pipeline. Reporting pre-normalized values would show values that were not compared and add irrelevant `%2F`/`%2f` differences to the error.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
```diff
@@ -463,18 +487,113 @@ public static TheoryData<ValidateSignedHttpRequestTheoryData> ValidatePClaimTheo
...
                         SignedHttpRequestToken = SignedHttpRequestTestUtils.ReplaceOrAddPropertyAndCreateDefaultSignedHttpRequest(new JProperty(SignedHttpRequestClaimTypes.P, "/foo%2fbar")),
-                        TestId = "ValidPHexCaseMatch",
+                        TestId = "ValidPHexCaseMatchLower",
                     },
                     new ValidateSignedHttpRequestTheoryData
                     {
-                        // %2F (upper) and %2f (lower) differ under the ordinal comparison and must be rejected; they would fold under OrdinalIgnoreCase.
                         HttpRequestUri = new Uri("https://www.contoso.com/foo%2Fbar"),
                         SignedHttpRequestToken = SignedHttpRequestTestUtils.ReplaceOrAddPropertyAndCreateDefaultSignedHttpRequest(new JProperty(SignedHttpRequestClaimTypes.P, "/foo%2fbar")),
+                        TestId = "ValidPHexCaseMismatchUpperRequest",
```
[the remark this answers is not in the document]
debchoudhury-id4s: Keeping these names. `Valid` states the expected successful result, while `HexCaseMismatch` describes the input condition. `UpperRequest` and `LowerRequest` distinguish the two directions. The outcome is clear and the names preserve useful test detail.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1284,6 +1284,36 @@ private static string CalculateBase64UrlEncodedHash(byte[] bytes)
...
+                char firstHexCharacter = value[i + 1];
+                char secondHexCharacter = value[i + 2];
+                if (!Uri.IsHexDigit(firstHexCharacter) || !Uri.IsHexDigit(secondHexCharacter))
+                    continue;
+
+                if (char.IsLower(firstHexCharacter) || char.IsLower(secondHexCharacter))
+                {
+                    normalizedValue ??= value.ToCharArray();
+
+                    if (char.IsLower(firstHexCharacter))
```
robezirg: how efficient is this check?

robezirg: do we know what is the more common scenario? upper or lower?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -819,8 +819,8 @@ internal virtual void ValidatePClaim(JsonWebToken signedHttpRequest, SignedHttpR
                 throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidPClaimException(LogHelper.FormatInvariant(LogMessages.IDX23003, LogHelper.MarkAsNonPII(SignedHttpRequestClaimTypes.P))));
 
             // relax comparison by trimming start and ending forward slashes
-            pClaimValue = pClaimValue.Trim('/');
-            var expectedPClaimValue = httpRequestUri.AbsolutePath.Trim('/');
+            pClaimValue = NormalizePercentEncodingHexCase(pClaimValue.Trim('/'));
```
robezirg: would it be more efficient on hot path to do the comparison on char array rather than converting it back to string? seems like extra allocation that is not needed.

in fact, i think we can make the comparison faster by just manually doing the string comparison for both. Walk both strings char by char and do the comparison manually

debchoudhury-id4s: Incorporated in latest commit. I had originally adopted that approach for code clarity and reuse as the normalization (involving string alloc) is needed on fail path .

However, your suggestion does give a minor perf improvement for the hot path, hence incorporated it

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: This **no longer** runs on the validation **hot path**. Check latest update which has introduced a  a single left-to-right pass over the two strings. `NormalizePercentEncodingHexCase` runs only inside the failure branch

debchoudhury-id4s: Direction does not affect the hot path anymore (check this commit compared to earlier), Just as FYI [RFC 3986 section 6.2.2.1](https://www.rfc-editor.org/rfc/rfc3986.html#section-6.2.2.1) defines uppercase as the canonical form for percent-encoded triplets.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1284,6 +1291,83 @@ private static string CalculateBase64UrlEncodedHash(byte[] bytes)
...
+            {
+                if (value[i] != '%')
+                    continue;
+
+                char firstHexCharacter = value[i + 1];
+                char secondHexCharacter = value[i + 2];
+                if (!Uri.IsHexDigit(firstHexCharacter) || !Uri.IsHexDigit(secondHexCharacter))
+                    continue;
+
+                if (char.IsLower(firstHexCharacter) || char.IsLower(secondHexCharacter))
```
dlingam-msft: Suggestion to use ToUpperInvariant here too so both helpers do it the same way

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1284,6 +1291,83 @@ private static string CalculateBase64UrlEncodedHash(byte[] bytes)
...
+                i += 2;
+                comparisonStartIndex = i + 1;
+            }
+
+            // Compare the remaining literal path segment after the final triplet.
+            int remainingSegmentLength = expectedValue.Length - comparisonStartIndex;
+            return string.Compare(expectedValue, comparisonStartIndex, actualValue, comparisonStartIndex, remainingSegmentLength, comparison) == 0;
+        }
+
+        private static string NormalizePercentEncodingHexCase(string value)
```
dlingam-msft: ArePClaimValuesEqual  and  NormalizePercentEncodingHexCase  both walk the string looking for  %XX  triplets with almost identical logic. That's fine, but leave a short note that the two loops must stay in sync if one changes.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
```diff
@@ -342,39 +342,124 @@ public void ValidatePClaim(ValidateSignedHttpRequestTheoryData theoryData)
...
+        [InlineData("/caf%C3%A9", "/caf%c3%a9", true, true)]
+        [InlineData("/caf%C3%A9", "/caf%c3%a9", false, true)]
+        [InlineData("/Path", "/path", true, false)]
+        [InlineData("/Path", "/path", false, true)]
+        [InlineData("/path", "/paths", true, false)]
+        [InlineData("/foo%25", "/foo%", true, false)]
+        [InlineData("/foo%25X", "/foo%X", false, false)]
+        [InlineData("/foo%252G", "/foo%2G", true, false)]
+        [ResetAppContextSwitches]
+        public void ValidatePClaim_ComparisonMatrix(string requestPath, string pClaim, bool useCaseSensitiveComparison, bool expectedValid)
```
dlingam-msft, line 407: This standalone  [Theory] mostly repeats cases already in  ValidatePClaimTheoryData , and uses a different style ( Record.Exception ) than the rest of the file. Prefer moving these rows into the existing theory data so the tests stay consistent.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
```diff
@@ -342,39 +342,124 @@ public void ValidatePClaim(ValidateSignedHttpRequestTheoryData theoryData)
...
+                HttpRequestUri = new Uri("https://www.contoso.com/Expected%2fPath"),
+                SignedHttpRequestToken = SignedHttpRequestTestUtils.ReplaceOrAddPropertyAndCreateDefaultSignedHttpRequest(new JProperty(SignedHttpRequestClaimTypes.P, "/Actual%2aPath")),
+            };
+            var signedHttpRequestValidationContext = theoryData.BuildSignedHttpRequestValidationContext();
+
+            // Act
+            var exception = Assert.Throws<SignedHttpRequestInvalidPClaimException>(() => handler.ValidatePClaim(theoryData.SignedHttpRequestToken, signedHttpRequestValidationContext));
+
+            // Assert
+            Assert.Contains("Expected%2FPath", exception.Message);
```
dlingam-msft: The assertion only checks the uppercase form is present. It would still pass if lowercase leaked too.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
```diff
@@ -463,18 +548,113 @@ public static TheoryData<ValidateSignedHttpRequestTheoryData> ValidatePClaimTheo
...
+                        SignedHttpRequestToken = SignedHttpRequestTestUtils.ReplaceOrAddPropertyAndCreateDefaultSignedHttpRequest(new JProperty(SignedHttpRequestClaimTypes.P, "/foo%")),
+                        ExpectedException = new ExpectedException(typeof(SignedHttpRequestInvalidPClaimException), "IDX23011"),
+                        TestId = "InvalidPTrailingPercent",
+                    },
+                    new ValidateSignedHttpRequestTheoryData
+                    {
+                        HttpRequestUri = new Uri("https://www.contoso.com/foo%252"),
+                        SignedHttpRequestToken = SignedHttpRequestTestUtils.ReplaceOrAddPropertyAndCreateDefaultSignedHttpRequest(new JProperty(SignedHttpRequestClaimTypes.P, "/foo%2")),
+                        ExpectedException = new ExpectedException(typeof(SignedHttpRequestInvalidPClaimException), "IDX23011"),
+                        TestId = "InvalidPTruncatedTriplet",
```
dlingam-msft, line 621: It's good to add equal-length failing cases (claim ending in  %  or  %2 ) so the last two chars can't be a triplet.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1284,6 +1291,83 @@ private static string CalculateBase64UrlEncodedHash(byte[] bytes)
             return Base64UrlEncoder.Encode(hashedBytes);
         }
 
+        private static bool ArePClaimValuesEqual(string expectedValue, string actualValue, StringComparison comparison)
```
dlingam-msft: Add a fast path at the top  for the common case with no percent-encoding, so plain paths keep using the vectorized string intrinsic instead of the per-char scan

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -825,8 +825,15 @@ internal virtual void ValidatePClaim(JsonWebToken signedHttpRequest, SignedHttpR
             // URI path components are case-sensitive per RFC 3986 section 3.3, so the comparison is ordinal (case-sensitive) by default.
             // The AppContext switch allows a consumer to restore the legacy case-insensitive comparison during transition.
             var comparison = AppContextSwitches.UseCaseSensitivePClaimComparison ? StringComparison.Ordinal : StringComparison.OrdinalIgnoreCase;
-            if (!string.Equals(expectedPClaimValue, pClaimValue, comparison))
+            if (!ArePClaimValuesEqual(expectedPClaimValue, pClaimValue, comparison))
```
debchoudhury-id4s: From @MZOLN : _"couldn't we just url decode before comparison? I'd rather us use less code here as the risk is introduction of a bug in url decoding. This could expose us to extra bugs just to do smth that everyone is doing to avoid a single string alloc? Isn't there one in like a util library from .net?_

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: Addressed in latest commit. Both manual helpers were removed. The single regex replacement path now uses `match.Value.ToUpperInvariant()`

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: Addressed in latest commit.There are no duplicate loops to keep synchronized. Match / Comparison is now defined by the regex pattern.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
debchoudhury-id4s: Added in latest commit

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: The custom per-character comparer was removed.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: replacing the custom URI-processing loops with a standard-library regex operation. General URL decoding is not used because decoding `%2F` produces `/` and would incorrectly make an encoded reserved slash equal to a literal path separator.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
debchoudhury-id4s: The test now also asserts that `Expected%2fPath` and `Actual%2aPath` are absent from the exception message.

test/Microsoft.IdentityModel.Protocols.SignedHttpRequest.Tests/SignedHttpRequestValidationTests.cs
debchoudhury-id4s: I kept the compact comparison matrix because it sets the AppContext switch per row and covers both strict and compatibility behavior

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: In the latest commit ([`09a9bf5`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3561/commits/09a9bf5af487a87d9489ff283201bb527d5bed43)) I dropped the custom single-pass comparer, so replying to keep this from going stale. 

The check is now `string.Equals` fast path ([L832](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/09a9bf5af487a87d9489ff283201bb527d5bed43/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs#L832)), which handles the common exact-match case; the only extra cost is a `Regex.Replace` uppercase pass ([L1310-L1317](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/09a9bf5af487a87d9489ff283201bb527d5bed43/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs#L1310-L1317) that runs only on a mismatch, never on the success path.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: In the latest commit ([`09a9bf5`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3561/commits/09a9bf5af487a87d9489ff283201bb527d5bed43)) I reverted the manual char-by-char walk and went back to `string.Equals` + a `Regex.Replace` normalization, so flagging that my earlier "incorporated it" note is now stale. 

I intentionally picked this middle ground for simplicity and **fewer new lines** : `ArePClaimValuesEqual` is removed and the hot path is the intrinsic ordinal compare ([L832](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/09a9bf5af487a87d9489ff283201bb527d5bed43/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs#L832)).

The string allocation is now on the mismatch branch only ([L842-L843](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/09a9bf5af487a87d9489ff283201bb527d5bed43/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs#L842-L843)).

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: Still agnostic in the latest commit ([`09a9bf5`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3561/commits/09a9bf5af487a87d9489ff283201bb527d5bed43)), updating since my earlier answer referenced the removed char-scan path. 

The hot path is a plain `string.Equals` ([L832](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/09a9bf5af487a87d9489ff283201bb527d5bed43/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs#L832)) and on a mismatch both sides are canonicalized to uppercase regardless of direction ([L842-L843](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/09a9bf5af487a87d9489ff283201bb527d5bed43/src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs#L842-L843)), so casing frequency doesn't matter for correctness.

Uppercase is just the canonical form per [RFC 3986 §6.2.2.1](https://www.rfc-editor.org/rfc/rfc3986.html#section-6.2.2.1).

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ See the [releases](https://github.com/AzureAD/azure-activedirectory-identitymode
 ====
 ## Bug Fixes
 - Align Signed HTTP Request `p` (path) claim comparison with RFC 3986 section 3.3 by comparing the path case-sensitively; previously `/Admin/resource` matched `/admin/resource`. An AppContext opt out is available during transition. See [PR #3526](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3526).
+- Normalize hexadecimal letter casing inside percent-encoded triplets during Signed HTTP Request `p` claim validation. Literal path-letter comparison remains controlled by `UseCaseSensitivePClaimComparison`, and `p` claim creation output is unchanged.
```
bgavrilMS, line 7: Are you adding to an already released version? I don't know the process for Wilson, but generally the changelog is changed after the release.

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -26,6 +27,8 @@ namespace Microsoft.IdentityModel.Protocols.SignedHttpRequest
     /// <remarks>The handler implementation is based on 'A Method for Signing HTTP Requests for OAuth' specification.</remarks>
     public class SignedHttpRequestHandler
     {
+        private const string _percentEncodedTripletPattern = "%[0-9A-Fa-f]{2}";
```
bgavrilMS: nit: we generally try to follow the .NET team's naming scheme. I believe for constants, even private ones, it's UpperCase i.e. PercentEncodedTripletPattern

bgavrilMS (approved):

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -825,8 +828,28 @@ internal virtual void ValidatePClaim(JsonWebToken signedHttpRequest, SignedHttpR
...
+            {
+                // Only equal-length ordinal values with '%' on both sides can differ solely by hex case, such as "%2F" and "%2f".
+                bool mayMatchAfterNormalization =
+                    comparison == StringComparison.Ordinal &&
+                    expectedPClaimValue.Length == pClaimValue.Length &&
+                    expectedPClaimValue.IndexOf('%') >= 0 &&
+                    pClaimValue.IndexOf('%') >= 0;
+
+                // Canonicalize valid triplets for comparison and IDX23011, such as "%2f" to "%2F".
+                expectedPClaimValue = NormalizePercentEncodingHexCase(expectedPClaimValue);
```
bgavrilMS: Shouldn't you avoid the normalization operation if `mayMatchAfterNormalization ==false` ?

debchoudhury-id4s: No its still needed for the error message case

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: Sure. Can change to uppercase

CHANGELOG.md
debchoudhury-id4s: I added it to sit alongside the related work. It will anyways be needed hence decided to add to changelog to avoid work later on (or miss) during release

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1284,6 +1307,15 @@ private static string CalculateBase64UrlEncodedHash(byte[] bytes)
             return Base64UrlEncoder.Encode(hashedBytes);
         }
 
+        private static string NormalizePercentEncodingHexCase(string value)
```
robezirg: I am not sold on the regex here, especially if it is not compiled, and just seems like there is really no need for this. I also don';t see why we need to normalize for logging. Imo we can just do a single pass, and it will be cleanest and cheapest option 

we can do something like this:

```
var comparison = AppContextSwitches.UseCaseSensitivePClaimComparison
    ? StringComparison.Ordinal : StringComparison.OrdinalIgnoreCase;

if (string.Equals(expectedPClaimValue, pClaimValue, comparison))
    return;

if (comparison == StringComparison.Ordinal && EqualsIgnoringPercentEncodingHexCase(expectedPClaimValue, pClaimValue))
    return;

throw LogHelper.LogExceptionMessage(new SignedHttpRequestInvalidPClaimException(
    LogHelper.FormatInvariant(LogMessages.IDX23011, LogHelper.MarkAsNonPII(SignedHttpRequestClaimTypes.P), expectedPClaimValue, pClaimValue)));
```

```
private static bool EqualsIgnoringPercentEncodingHexCase(string expected, string actual)
{
    if (expected.Length != actual.Length)
        return false;

    for (int i = 0; i < expected.Length; i++)
    {
        if (Uri.IsHexEncoding(expected, i) && Uri.IsHexEncoding(actual, i))
        {
            // Both triplets are ASCII hex, so setting bit 5 folds 'A'-'F' to 'a'-'f' and leaves '0'-'9' unchanged.
            if ((expected[i + 1] | 0x20) != (actual[i + 1] | 0x20) ||
                (expected[i + 2] | 0x20) != (actual[i + 2] | 0x20))
                return false;

            i += 2;
            continue;
        }

        if (expected[i] != actual[i])
            return false;
    }

    return true;
}
```

debchoudhury-id4s: > we can do something like this:

 @robezirg Your suggestion is the custom code I had used in prev commit. The clear callout on that was to avoid such custom newlines IF simpler library friendly option is available. 

This current approach avoids the custom code approach as well as not incur perf overhead in majority cases (check PR description)

> I also don';t see why we need to normalize for logging

@robezirg We do need the normalization in logging. Please check the comments inline in the src file itself.

debchoudhury-id4s: @robezirg I can add more comments above the normalization / logging code lines on why its needed

debchoudhury-id4s: Added comments inline

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: Addressed in latest commit

dlingam-msft (approved):

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: Check latest commit in which custom traversal is introduced and logging normalization is removed

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: Check latest commit in which custom traversal is re-introduced and logging normalization is removed

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1284,6 +1295,52 @@ private static string CalculateBase64UrlEncodedHash(byte[] bytes)
...
+                    {
+                        // RFC 3986 section 6.2.2.1 defines uppercase hexadecimal letters as the canonical form for percent-encoded triplets.
+                        if (char.ToUpperInvariant(expectedFirstHexCharacter) != char.ToUpperInvariant(actualFirstHexCharacter) ||
+                            char.ToUpperInvariant(expectedSecondHexCharacter) != char.ToUpperInvariant(actualSecondHexCharacter))
+                        {
+                            return false;
+                        }
+
+                        i += 2;
+                        continue;
```
robezirg, line 1331: what is the point of this continue?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
```diff
@@ -1284,6 +1295,52 @@ private static string CalculateBase64UrlEncodedHash(byte[] bytes)
...
+                    }
+                }
+
+                if (expectedValue[i] != actualValue[i])
```
robezirg, line 1335: this should be the first check no?

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: handle the next pair of hex triplets (ex: `%2F` & `%2f`) instead of simply moving on to the next pair of chars (ex: `F` & `f`)

src/Microsoft.IdentityModel.Protocols.SignedHttpRequest/SignedHttpRequestHandler.cs
debchoudhury-id4s: No. The `%XX` (triplet) check must run first. Otherwise, `%` passes the literal check. and then one character at a time check would fail `F` and `f` . 

This check is needed **after** the triplet check to compare the literal characters (ex: like `B` & `b` while comparing `%2FBar` and `%2fbar`)

robezirg (approved):

MZOLN (approved):

PR #3566, opened 2026-07-28
add interface definitions to support sync validation pipeline

Toidi357:
# Sync Validation Pipeline PR #1

Additional of all interfaces needed to support sync validation pipeline.


src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationRetrieverSync.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <typeparam name="T">The type of the configuration metadata.</typeparam>
+    public interface IConfigurationRetrieverSync<T>
+    {
+        /// <summary>
+        /// Retrieves a populated configuration given an address and an <see cref="IDocumentRetriever"/>.
+        /// </summary>
+        /// <param name="address">Address of the discovery document.</param>
+        /// <param name="retriever">The <see cref="IDocumentRetriever"/> to use to read the discovery document.</param>
+        /// <param name="cancel">A cancellation token that can be used by other objects or threads to receive notice of cancellation. <see cref="CancellationToken"/>.</param>
+        T GetConfigurationSync(string address, IDocumentRetriever retriever, CancellationToken cancel);
```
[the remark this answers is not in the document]
Toidi357: fixed in PR2

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -98,6 +101,124 @@ public HttpDocumentRetriever(HttpClient httpClient)
...
 
+#if NET5_0_OR_GREATER
+        /// <summary>
+        /// Returns a string converted from remote document when completed, by using the provided address.
+        /// </summary>
+        /// <param name="address">Location of document</param>
+        /// <param name="cancel">A cancellation token that can be used by other objects or threads to receive notice of cancellation. <see cref="CancellationToken"/></param>
+        /// <returns>Document as a string</returns>
+        public string GetDocumentSync(string address, CancellationToken cancel)
+        {
```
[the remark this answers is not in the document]
Toidi357: fixed in PR2

src/Microsoft.IdentityModel.Protocols/Configuration/FileDocumentRetriever.cs
```diff
@@ -45,5 +45,31 @@ public async Task<string> GetDocumentAsync(string address, CancellationToken can
...
+
+        /// <summary>
+        /// Reads a document synchronously using <see cref="FileStream"/>.
+        /// </summary>
+        /// <param name="address">Fully qualified path to a file.</param>
+        /// <param name="cancel"><see cref="CancellationToken"/> not used.</param>
+        /// <returns>UTF8 decoding of bytes in the file.</returns>
+        /// <exception cref="ArgumentNullException">If address is null or whitespace.</exception>
+        /// <exception cref="IOException">With inner exception containing the original exception.</exception>
+        public string GetDocumentSync(string address, CancellationToken cancel)
```
RojaEnnam: The sync methods need not have the Sync suffix. Please rename all these methods GetDocumentSync -> GetDocument

src/Microsoft.IdentityModel.Protocols/Configuration/ISyncDocumentRetriever.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License.
+
+using System.Threading;
+
+namespace Microsoft.IdentityModel.Protocols
+{
+    /// <summary>
+    /// Synchronous interface that defines a document retriever that returns the document as a string.
+    /// </summary>
+    public interface ISyncDocumentRetriever
```
RojaEnnam: can we keep this consistent? sync as a suffix similar to other interfaces?

RojaEnnam (approved):

src/Microsoft.IdentityModel.Protocols/Configuration/FileDocumentRetriever.cs
Toidi357: just pushed a commit to fix this

src/Microsoft.IdentityModel.Protocols/Configuration/ISyncDocumentRetriever.cs
Toidi357: just pushed a commit to fix this

src/Microsoft.IdentityModel.Protocols/Configuration/FileDocumentRetriever.cs
```diff
@@ -45,5 +45,31 @@ public async Task<string> GetDocumentAsync(string address, CancellationToken can
                 throw LogHelper.LogExceptionMessage(new IOException(LogHelper.FormatInvariant(LogMessages.IDX20804, address), ex));
             }
         }
+
+        /// <summary>
+        /// Reads a document synchronously using <see cref="FileStream"/>.
+        /// </summary>
+        /// <param name="address">Fully qualified path to a file.</param>
+        /// <param name="cancel"><see cref="CancellationToken"/> not used.</param>
```
westin-m, line 53: Why is there a cancellation token on a synchronous operation?

src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationEventHandlerSync.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Threading;
+
+namespace Microsoft.IdentityModel.Protocols.Configuration
+{
```
westin-m, line 13: nit: please be consistent and use file-scoped namespaces in new files

src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationEventHandlerContextAwareSync.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System.Threading;
+
+namespace Microsoft.IdentityModel.Protocols.Configuration;
+
+/// <summary>
+/// Defines synchronous, context-aware event handlers for configuration retrieval and update operations.
+/// </summary>
+/// <typeparam name="T">The type of configuration.</typeparam>
+public interface IConfigurationEventHandlerContextAwareSync<T> : IConfigurationEventHandlerSync<T> where T : class
```
westin-m: Why is there not an AfterRetrieve method on these?

src/Microsoft.IdentityModel.Protocols/Configuration/FileDocumentRetriever.cs
Toidi357: On the HTTP synchronous send, cancellation token can actually be used [see here](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.send?view=net-10.0)

this is built into the interface `IDocumentRetrieverSync` which both `HttpDocumentRetriever` and `FileDocumentRetriever` implement

src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationEventHandlerSync.cs
Toidi357: 👍

src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationEventHandlerContextAwareSync.cs
Toidi357: Added `Task AfterUpdateAsync` to `IConfigurationEventHandlerSync` and `IConfigurationEventHandlerContextAwareSync`. 

`AfterUpdateAsync` is called in a fire-and-forget matter, so AfterRetrieve does not need to be synchronous

RojaEnnam: Doesn't seem right to have an Async method on a Sync interface, why can't we make it sync?

westin-m (approved):

src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationEventHandlerContextAwareSync.cs
Toidi357: changed it to sync

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -98,6 +101,124 @@ public HttpDocumentRetriever(HttpClient httpClient)
...
+                        LogHelper.FormatInvariant(
+                            LogMessages.IDX20804,
+                            LogHelper.MarkAsNonPII(address)),
+                        ex));
+            }
+
+            throw LogHelper.LogExceptionMessage(unsuccessfulHttpResponseException);
+        }
+
+        private HttpResponseMessage SendAndRetryOnNetworkError(HttpClient httpClient, Uri uri, CancellationToken cancel, out string responseContent)
```
MZOLN, line 172: This should be encapsulated in a separate component wiht it's own interface and potentially a strategy to be able to inject new modes of operation.

The actual mode we want to allow should be picked with a runtime configuration setting (some sort of an option structure?

We cannot arbitrarily decide for users what strategy is correct for them, so we need to provide injection points and ability for them to pick from our default implementations through configuration.

src/Microsoft.IdentityModel.Protocols/Configuration/HttpDocumentRetriever.cs
```diff
@@ -98,6 +101,124 @@ public HttpDocumentRetriever(HttpClient httpClient)
...
+            Exception unsuccessfulHttpResponseException;
+            HttpResponseMessage response;
+            try
+            {
+                if (LogHelper.IsEnabled(EventLogLevel.Verbose))
+                    LogHelper.LogVerbose(LogMessages.IDX20805, LogHelper.MarkAsNonPII(address));
+
+                var httpClient = _httpClient ?? _defaultHttpClient;
+                var uri = new Uri(address, UriKind.RelativeOrAbsolute);
+                response = SendAndRetryOnNetworkError(httpClient, uri, cancel, out string responseContent);
```
MZOLN, line 133: This requires sperate encapsulation/interface and probably a strategy that  will be driven by options provided by the caller, to pick the one they want.

Callers need to have ability to provide their own ways of handling this, as they might want to be able to inject different resiliency strategies. 

We cannot arbitrary decide which one is the correct one as each service is different. There will be different resiliency profiles depending on the service.

MZOLN (changes requested):

PR #3567, opened 2026-07-28
Add synchronous ConfigurationManager tests and supporting sync APIs

Toidi357:
See more recent comments


src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -340,6 +372,128 @@ private async Task<T> GetConfigurationNonBlockingAsync(CancellationToken cancel)
...
+                                TelemetryConstants.Protocols.ConfigurationSourceHandler);
+
+                            UpdateConfiguration(configurationRetrieved.Configuration, configurationRetrieved.RetrievalTime, retrievalContext);
+                            return _currentConfiguration;
+                        }
+                    }
+
+                    // Don't use the individual CT here, this is a shared operation that shouldn't be affected by an individual's cancellation.
+                    // The transport should have its own timeouts, etc.
+                    T configuration = ((IConfigurationRetrieverSync<T>)_configRetriever).GetConfigurationSync(
```
RojaEnnam: Any custom or third-party retriever, and any future built-in one that does not implement the sync interface will hit the InvalidCastException  on the sync path.

Can we add another constructor that takes in sync config retriever and document retriever and use that?

Also, consider throwing here if it is not a sync retriever.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -340,6 +372,128 @@ private async Task<T> GetConfigurationNonBlockingAsync(CancellationToken cancel)
...
+            else
+            {
+                if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
+                {
+                    TelemetryClient.IncrementConfigurationRefreshRequestCounter(
+                        MetadataAddress,
+                        TelemetryConstants.Protocols.Automatic,
+                        TelemetryConstants.Protocols.ConfigurationSourceUnknown);
+
+                    _ = Task.Run(_updateCurrentConfigurationWithoutBypassAsync, CancellationToken.None);
```
RojaEnnam: Even on the sync path, the background refresh is dispatched to the async pipeline ( _configRetriever.GetConfigurationAsync  / async  _docRetriever ). This works today only because the built-in retrievers implement both interfaces and the async fields are populated. Once a sync-only retriever is injected, this will break. We'll need a sync background-refresh path so the sync pipeline is self-contained.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -470,6 +624,18 @@ public override async Task<BaseConfiguration> GetBaseConfigurationAsync(Cancella
             return obj as BaseConfiguration;
         }
 
+        /// <summary>
+        /// Obtains an updated version of Configuration.
+        /// </summary>
+        /// <param name="cancel">CancellationToken</param>
+        /// <returns>Configuration of type BaseConfiguration.</returns>
+        /// <remarks>If the time since the last call is less than <see cref="BaseConfigurationManager.AutomaticRefreshInterval"/> then <see cref="IConfigurationRetriever{T}.GetConfigurationAsync"/> is not called and the current Configuration is returned.</remarks>
```
RojaEnnam: still referencing async

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -340,6 +372,128 @@ private async Task<T> GetConfigurationNonBlockingAsync(CancellationToken cancel)
                     fetchMetadataFailure));
         }
 
+        private T GetConfigurationNonBlockingSync(CancellationToken cancel)
```
RojaEnnam: Most of this method is identical to  GetConfigurationNonBlockingAsync  , only the retrieval call differs (sync vs async). Could we extract the non- await  pieces into shared private helpers that both call? Specifically the validator check, the success telemetry increment, and the return stale or throw IDX20803". That shrinks each copy to just the lock + fetch +  UpdateConfiguration  skeleton and reduces the risk of the two drifting.

Toidi357:
### Large refactoring changes to `ConfigurationManager`

Old: `ConfigurationManager.cs` and `ConfigurationManager_Blocking.cs`
New: 
- `ConfigurationManager_Constructors.cs`: holds all constructors and `CreateSync` factory methods
- `ConfigurationManager_Async.cs`: holds all methods related to async retrieval
- `ConfigurationManager_Sync.cs`: holds all methods related to sync retrieval
- `ConfigurationManager.cs`: houses shared properties (like locks), has helper methods that both sync and async retrieval paths call (mostly error handling and telemetry logging helper functions)

### About `CreateSync`
Adding `CreateSync` methods to `ConfigurationManager` (used like `ConfigurationManager.CreateSync(...)`, returns instance of `ConfigurationManager` ) allows for instantiating `ConfigurationManager`'s with _only_ synchronous retrievals supported

No regressions, can still instantiate with `new ConfigurationManager(...)`

#### Behavior matrix

If you have a `retriever` that supports both `I...Retriever` and `I...RetrieverSync`, you can use either standard constructors or `CreateSync` factory methods; they behave the same. Additionally both `GetConfigurationAsync` and `GetConfigurationSync` works. The provided `OpenIdConnectConfigurationRetriever`/`WsFederationConfigurationRetriever` and `HttpDocumentRetriever`/`FileDocumentRetriever` classes implement both async and sync methods, so it doesn't matter how you instantiate them

If you have a `retriever` that supports only `I...Retriever` (only async supported), only standard constructors work, and `GetConfigurationSync` will throw an exception

If you have a `retriever` that supports only `I...RetrieverSync` (only sync supported), only `CreateSync` factories work, and `GetConfigurationAsync` will throw an exception

All these cases are exercised in `ConfigurationManagerSyncAsyncConstructorsTests.cs`, located in `...Protocols.Tests/` package

saurabhsathe-ms:
**RequestRefresh()  can silently no-op with a custom async-only retriever**
The sync pipeline is enabled by best-effort casts:
_configRetrieverSync = configRetriever as IConfigurationRetrieverSync<T>; // null for any pre-existing custom retriever
 IConfigurationRetrieverSync<T>  is new, so every custom retriever already in the wild implements async only — and we can't audit customer types.

**Repro:** custom async-only retriever +  ConfigurationEventHandlerSync  set.  RequestRefresh()  routes to the sync lane → handler returns  NoResult  →  UpdateCurrentConfigurationSync  hits  _configRetrieverSync == null  → throws  IDX20814 , caught and logged inside the fire-and-forget  Task.Run . Caller sees nothing; the manual refresh never happens (breaks the unknown- kid  → force-refresh → retry pattern).
Not asking you to throw per-refresh — swallowing transient failures is right. The issue is routing to a lane the retrievers structurally can't complete, which looks identical to a transient failure.

**Ask:** Could you confirm how  RequestRefresh  is meant to pick the sync vs. async lane?

1. Should it route by construction mode (async ctor → async,  CreateSync  → sync)? That guarantees the chosen lane's retrievers are always wired, so the silent no-op can't happen. Today a  CreateSync  manager with dual-interface retrievers routes to async instead.
2. If a sync handler must force the sync lane, should mis-wiring (sync lane but  _configRetrieverSync  null) fail fast at construction ( NotSupportedException , zero hot-path cost) rather than be swallowed per-call?

**Suggestion:** Track how the manager was constructed with an explicit flag (e.g.  _isSyncManager , set in the ctor /  CreateSync ) and route on that instead of inferring from the  as -cast retriever fields. Since each construction path guarantees its own retrievers are wired, the chosen lane can always complete — which removes the silent-no-op entirely without restricting retrievers from implementing both interfaces.

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Async.cs
```diff
@@ -0,0 +1,375 @@
...
+    /// <para>
+    /// By using the app context switch you choose what works best for you when there is a signing key update:
+    /// either block requests from being validated until the new key is retrieved, or allow requests to be validated
+    /// with the current key until the new key is retrieved. If blocking, a service receiving high concurrent request
+    /// may experience thread starvation.
+    /// </para>
+    /// </remarks>
+    public virtual async Task<T> GetConfigurationAsync(CancellationToken cancel)
+    {
+        if (_currentConfiguration != null && _syncAfter > TimeProvider.GetUtcNow())
```
debchoudhury-id4s, line 61: Do you need to perform null checks on retrievers as well before returning the configuration ?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
```diff
@@ -56,393 +58,131 @@ public partial class ConfigurationManager<T> : BaseConfigurationManager, IConfig
...
+                    TelemetryConstants.Protocols.Manual,
+                    TelemetryConstants.Protocols.ConfigurationSourceUnknown);
 
-            MetadataAddress = metadataAddress;
-            _docRetriever = docRetriever;
-            _configRetriever = configRetriever;
+                _isFirstRefreshRequest = false;
+                if (Interlocked.CompareExchange(ref _configurationRetrieverState, ConfigurationRetrieverRunning, ConfigurationRetrieverIdle) == ConfigurationRetrieverIdle)
+                {
+                    if (ConfigurationEventHandlerSync != null || _configRetrieverAsync == null || _docRetrieverAsync == null)
```
debchoudhury-id4s: Any checks needed against `_configRetrieverSync` and `_docRetrieverSync` ?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Async.cs
Toidi357: Checks on retrievers are done in GetConfigurationNonBlockingAsync/GetConfigurationWithBlockingAsync which are in the next lines after this snippet

Right now how it's setup is that you *can* call GetConfigurationAsync even if you're using the sync route if and only if configuration is already cached, and it will throw an exception if a metadata fetch is actually performed without an async retriever. do you think we should gate the entire GetConfigurationAsync?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
Toidi357: Yes, this is related to Saurabh's comment as well

Toidi357:
@saurabhsathe-ms @debchoudhury-id4s to respond to your comments:

The decision to choose sync vs. async can be based on whether you instantiated the "async way" with new ConfigurationManager() or the "sync way" with ConfigurationManager.CreateSync() and not ConfigurationManagerEventHandler; to do this a property was added to "remember" which way you constructed it. This means we wouldn't have dual retriever support: if you pass in a retriever that supports async and sync, you can call either one of GetConfigurationSync/Async and they both work, but with this more restrictive gating, you can only use one of sync or async

So RequestRefresh() and all GetCnofigurationAsync/Sync would gate on this property, and not null checks on retrievers. 

If someone instantiates with async only retrievers, but for some reason decides to set a ConfigurationEventHandler**Sync**, nothing would happen since everything would go async-only and it wouldn't look at ConfiguartionEventHandlerSync

how does this sound?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Async.cs
Toidi357: ok on second thought, it's sort of a makes-sense behavior that GetConfigurationAsync throws if you only have sync set, no matter whether the cache is set or not, i'm gating the public entrypoints (GetConfigurationAsync/Sync) with null checks on retrievers

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManagerSync_Blocking.cs
```diff
@@ -0,0 +1,373 @@
...
+
+                        UpdateConfiguration(configurationRetrieved.Configuration, configurationRetrieved.RetrievalTime, retrievalContext);
+                        return _currentConfiguration;
+                    }
+                }
+
+                T configuration = _configRetrieverSync.GetConfigurationSync(
+                    MetadataAddress,
+                    _docRetrieverSync,
+                    CancellationToken.None);
```
debchoudhury-id4s, line 117: Can this lead to / propagate `OperationCanceledException` ?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManagerSync_Blocking.cs
```diff
@@ -0,0 +1,373 @@
...
+
+                        // replicate the behavior of successful retrieval from endpoint
+                        if (configurationRetrieved != null && configurationRetrieved.Configuration != null)
+                        {
+                            TelemetryForUpdateBlocking(TelemetryConstants.Protocols.ConfigurationSourceHandler);
+
+                            if (_refreshRequested)
+                                _refreshRequested = false;
+
+                            UpdateConfiguration(configurationRetrieved.Configuration, configurationRetrieved.RetrievalTime, retrievalContext);
```
debchoudhury-id4s, line 188: Does this need to reset `_lastRequestRefresh` so the next `RequestRefresh()` respects `RefreshInterval`?

debchoudhury-id4s:
### Could this be split into self-contained PRs? Something along below lines:
 
1. first move the constructors and async logic into without behavior changes; 
2. then add the core manager and tests; 
3. then add `OpenID Connect` and `WS-Federation` sync retrieval
4. Finally remaining handlers / telemetry 

### Not asking to split this PR but please keep in mind for subsequent PRs

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManagerSync_Blocking.cs
Toidi357: Don't think so, since we pass CancellationToken.None

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManagerSync_Blocking.cs
Toidi357: Yes, this is actually a wider bug within ConfigurationManager, fixed it with a commit here and also see https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3586

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
westin-m, line 1: Why is this being deleted?

Toidi357: Instead of splitting files by blocking/non-blocking, I propose it's better to split by sync/async; none of the original code/functions are being deleted, just being moved around

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationRetriever.cs
```diff
@@ -85,5 +90,48 @@ public static async Task<OpenIdConnectConfiguration> GetAsync(string address, ID
...
         }
+
+        /// <summary>
+        /// Retrieves a populated <see cref="OpenIdConnectConfiguration"/> given an address and an <see cref="IDocumentRetrieverSync"/>.
+        /// </summary>
+        /// <param name="address">address of the discovery document.</param>
+        /// <param name="retriever">the <see cref="IDocumentRetrieverSync"/> to use to read the discovery document</param>
+        /// <param name="cancel"><see cref="CancellationToken"/>.</param>
+        /// <returns>A populated <see cref="OpenIdConnectConfiguration"/> instance.</returns>
+        public static OpenIdConnectConfiguration GetSync(string address, IDocumentRetrieverSync retriever, CancellationToken cancel)
```
MZOLN: Wouldnt it be better if this was an isolated class?

src/Microsoft.IdentityModel.Protocols.WsFederation/Configuration/WsFederationConfigurationRetriever.cs
```diff
@@ -83,5 +89,30 @@ public static async Task<WsFederationConfiguration> GetAsync(string address, IDo
...
+        /// <summary>
+        /// Retrieves a populated <see cref="WsFederationConfiguration"/> given an address and an <see cref="IDocumentRetrieverSync"/>.
+        /// </summary>
+        /// <param name="address">address of the metadata document.</param>
+        /// <param name="retriever">the <see cref="IDocumentRetrieverSync"/> to use to read the metadata document</param>
+        /// <param name="cancel"><see cref="CancellationToken"/>.</param>
+        /// <returns>A populated <see cref="WsFederationConfiguration"/> instance.</returns>
+        /// <exception cref="ArgumentNullException">if <paramref name="address"/> is null or empty.</exception>
+        /// <exception cref="ArgumentNullException">if <paramref name="retriever"/> is null.</exception>
+        public static WsFederationConfiguration GetSync(string address, IDocumentRetrieverSync retriever, CancellationToken cancel)
```
MZOLN: Same

src/Microsoft.IdentityModel.Protocols/Configuration/IConfigurationRetrieverSync.cs
```diff
@@ -12,10 +12,10 @@ namespace Microsoft.IdentityModel.Protocols;
...
     /// <summary>
-    /// Retrieves a populated configuration given an address and an <see cref="IDocumentRetriever"/>.
+    /// Retrieves a populated configuration given an address and an <see cref="IDocumentRetrieverSync"/>.
     /// </summary>
     /// <param name="address">Address of the discovery document.</param>
-    /// <param name="retriever">The <see cref="IDocumentRetriever"/> to use to read the discovery document.</param>
+    /// <param name="retriever">The <see cref="IDocumentRetrieverSync"/> to use to read the discovery document.</param>
     /// <param name="cancel">A cancellation token that can be used by other objects or threads to receive notice of cancellation. <see cref="CancellationToken"/>.</param>
-    T GetConfigurationSync(string address, IDocumentRetriever retriever, CancellationToken cancel);
+    T GetConfigurationSync(string address, IDocumentRetrieverSync retriever, CancellationToken cancel);
```
MZOLN, line 20: Sync should have a suffix, only async

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationRetriever.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.IdentityModel.Protocols.OpenIdConnect
     /// <summary>
     /// Retrieves a populated <see cref="OpenIdConnectConfiguration"/> given an address.
     /// </summary>
-    public class OpenIdConnectConfigurationRetriever : IConfigurationRetriever<OpenIdConnectConfiguration>
+    public class OpenIdConnectConfigurationRetriever : IConfigurationRetriever<OpenIdConnectConfiguration>, IConfigurationRetrieverSync<OpenIdConnectConfiguration>
```
MZOLN: I'm not sure that I'm a fan of this pattern.

MZOLN (changes requested):

src/Microsoft.IdentityModel.Protocols.OpenIdConnect/Configuration/OpenIdConnectConfigurationRetriever.cs
MZOLN: The interface usage here is concerning, because the implementation is static. I'm not sure what it benefits from and how to drive it forward currently.

Toidi357:
This PR adds a synchronous `ConfigurationManagerSync` class that is entirely independent of the existing `ConfigurationManager` class. We add a new base abstract class `BaseConfigurationManagerSync` to allow compatibility once we integrate this with the `JsonWebTokenHandler` in PR4. These new classes are internal only.

Additionally, `OpenIdConnectConfigurationRetriever` and `WsFederationConfigurationRetriever` have been duplicated to entirely separate sync classes. Their sync counterparts are internal only.

This PR adds a `TokenValidationParameters.ConfigurationManagerSync` internal property

PR #3571, opened 2026-07-29
# Deserialize Actor claim act

saurabhsathe-ms:
# Enable Actor Claim Support in Wilson - Deserialization

- [x] If any gains or losses in performance are possible, you've included benchmarks for your changes.
- [x] There's an open issue for the PR that you are making.

## Summary of the changes (Less than 80 chars)

Deserialize the RFC 8693 `act` claim into `ClaimsIdentity.Actor` in JsonWebTokenHandler

## Why

This is the **read-path** companion to the serialization PR (#3560, merged). #3560 taught
`JsonWebTokenHandler` to *write* the RFC 8693 `act` claim; this PR teaches it to *read* the actor
back into `ClaimsIdentity.Actor`, so a token written by this handler round-trips.

Fixes part of #1840.

## What changed

- On validation, `JsonWebTokenHandler` populates `ClaimsIdentity.Actor`. The RFC 8693 **`act`**
  (JSON object) claim **takes precedence** whenever present; when absent, the legacy **`actort`**
  (an unsigned nested-JWT string) is expanded for **read back-compatibility** — this handler writes
  `act`, never `actort`. Both apply on the `MapInboundClaims` on/off paths; `act` is detected via
  the **raw** `"act"` type, so inbound claim mapping can't hide it. Having both is not an error:
  `act` wins and nothing is thrown.
- **Nested actors:** `act` objects are expanded up to the process-wide
  `JsonWebTokenHandler.MaxActorChainLength`; deeper levels are **kept as a claim** (silent). The
  `actort` chain follows the classic behavior and is **not** bounded by `MaxActorChainLength`.
- **Degrade, don't throw.** A non-object `act` logs a warning (`IDX14316`) and is kept as a claim;
  a failing custom `ActClaimRetriever` logs a warning (`IDX14314`, PII-scrubbed) and leaves `Actor`
  null. **Nothing in the actor read path fails token validation** (except a null
  `TokenValidationParameters`).
- **Issuer.** `act`-derived actor claims carry the **outer token's validated issuer** on
  `Claim.Issuer` (the `act` object is asserted by the outer token). `NameClaimType` /
  `RoleClaimType` / `AuthenticationType` are left at their defaults — a documented, non-RFC choice.
- New public **`TokenValidationParameters.ActClaimRetriever`** — a delegate that, when supplied,
  fully owns `act` actor construction from the raw `JsonElement`.
- New messages: `IDX14314` (retriever failure — logged as a warning) and `IDX14316`
  (*"The 'act' claim must be a JSON object."* — logged as a warning).

## Notes

- **Depth limit is 64** — the reverse of serialization's `IDX10815`-on-write: a token whose `act`
  nests beyond the JSON reader's max depth of 64 fails to **read** with `IDX14101` wrapping a
  depth-64 `JsonReaderException`, before actor logic runs. The default `act` expansion is bounded by
  `MaxActorChainLength` and never recurses unboundedly.
- Extensibility: override `CreateClaimsIdentity` to customize `actort`-derived actors, or set
  `ActClaimRetriever` for `act`.
- No shipped internal API **signatures** changed; no public API beyond
  `TokenValidationParameters.ActClaimRetriever` (and its delegate type).
- Read/validate benchmarks added (`ActorClaimReadBenchmarks`).

Full design rationale, precedence rules, performance analysis, and test coverage live in the
actor-claim README:
[`src/Microsoft.IdentityModel.JsonWebTokens/ActorClaim.md`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/ssathe/deserializeActorClaim/src/Microsoft.IdentityModel.JsonWebTokens/ActorClaim.md).


Toidi357:
Suggestion to combine both your actor claim benchmark files into just one, you'd be able to re-use some code
Or keep it separate but make the names more descriptive like `ActorClaimSerializationTests` + `ActorClaimDeserializationTests` vs `ActorClaimBenchmarks` + `ActorClaimReadBenchmarks`

westin-m (approved):

src/Microsoft.IdentityModel.JsonWebTokens/LogMessages.cs
```diff
@@ -52,6 +52,8 @@ internal static class LogMessages
         internal const string IDX14310 = "IDX14310: JWE authentication tag is missing.";
         internal const string IDX14311 = "IDX14311: Unable to decode the authentication tag as a Base64Url encoded string.";
         internal const string IDX14312 = "IDX14312: Unable to decode the cipher text as a Base64Url encoded string.";
```
debchoudhury-id4s, line 54: Are `IDX14313` and `IDX14315` reserved for follow up PRs or any other use case ?

src/Microsoft.IdentityModel.JsonWebTokens/LogMessages.cs
```diff
@@ -52,6 +52,8 @@ internal static class LogMessages
...
+        internal const string IDX14314 = "IDX14314: Unable to deserialize act claim. Exception faced while using custom delegate to deserialize act claim. Nested exception is :{0}";
```
debchoudhury-id4s: Does this need to have the messaging / quote 'act' the way `IDX14316` does below?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimDeserializationTests.cs
```diff
@@ -0,0 +1,1298 @@
...
+using System.Diagnostics.Tracing;
+using System.Linq;
+using System.Security.Claims;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.TestUtils;
+using Microsoft.IdentityModel.Tokens;
+using Xunit;
+
+namespace Microsoft.IdentityModel.JsonWebTokens.Tests.ActClaimTests
```
debchoudhury-id4s, line 17: [Nit] this looks like a large file. any way to reduce lines of code by extracting out reusable helpers?

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimDeserializationTests.cs
```diff
@@ -0,0 +1,1298 @@
...
+                Claims = new Dictionary<string, object>
+                {
+                    { "act", actorIdentity}
+                },
+            };
+            string token = handler.CreateToken(tokenDescriptor);
+            handler.MapInboundClaims = true;
+
+            // Validate token
+            var validationParameters = new TokenValidationParameters
```
debchoudhury-id4s: Would it make sense to extract this setup into a private helper and use wherever applicable? Should reduce number of lines overall

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimDeserializationTests.cs
```diff
@@ -0,0 +1,1298 @@
...
+            actor.AddClaim(new Claim("sub", "actor-subject-id"));
+            actor.AddClaim(new Claim("name", "Actor Name"));
+            actor.Actor = nestedActor;
+
+            var mainIdentity = new CaseSensitiveClaimsIdentity("Bearer");
+            mainIdentity.AddClaim(new Claim("sub", "main-subject-id"));
+            mainIdentity.AddClaim(new Claim("name", "Main User"));
+
+            var handler = new JsonWebTokenHandler();
+            var tokenDescriptor = new SecurityTokenDescriptor
```
debchoudhury-id4s, line 459: consider extracting this setup into a private helper. should reduce lot of duplicate code lines

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -644,5 +631,171 @@ private static TokenValidationResult ReadToken(string token, TokenValidationPara
...
+        /// <param name="issuer">The issuer for the claims.</param>
+        /// <param name="currentDepth">The current recursion depth for nested actor processing.</param>
+        /// <returns>A ClaimsIdentity containing claims from the JsonElement.</returns>
+        internal static ClaimsIdentity CreateActorClaimsIdentityFromJsonElement(
+            JsonElement jsonElement,
+            TokenValidationParameters tokenValidationParameters,
+            string issuer = null,
+            int currentDepth = 0)
+        {
+            if (tokenValidationParameters is null)
```
debchoudhury-id4s: Is there any use of this `tokenValidationParameters` besides this null check? Can it be dropped from the method parameter list ?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -644,5 +631,171 @@ private static TokenValidationResult ReadToken(string token, TokenValidationPara
...
+        // read path stays back-compatible with tokens produced by JwtSecurityTokenHandler. This handler only
+        // reads "actort" - it never writes it. Having both "act" and "actort" is not an error: "act" simply
+        // wins and no exception is thrown.
+        private ClaimsIdentity ResolveActorClaimsIdentity(JsonWebToken jwtToken, TokenValidationParameters validationParameters, string issuer)
+        {
+            // "act" (RFC 8693) takes precedence whenever the claim is present, in ANY form, and always
+            // suppresses the legacy "actort". TryGetPayloadValue<JsonElement> succeeds only for JSON
+            // objects/arrays, so the common object case is a single lookup here (object -> expanded;
+            // array -> the helper warns IDX14316 and yields null).
+            if (jwtToken.TryGetPayloadValue<JsonElement>(ActClaimType, out JsonElement actClaim))
```
debchoudhury-id4s, line 646: Is a `HasPayloadClaim(ActClaimType)` needed before calling `TryGetPayloadValue<JsonElement>` ?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -215,28 +216,19 @@ private ClaimsIdentity CreateClaimsIdentityWithMapping(JsonWebToken jwtToken, To
...
                 bool wasMapped = _inboundClaimTypeMap.TryGetValue(jwtClaim.Type, out string claimType);
 
                 if (!wasMapped)
                     claimType = jwtClaim.Type;
 
-                if (claimType == ClaimTypes.Actor)
-                {
-                    if (identity.Actor != null)
-                        throw LogHelper.LogExceptionMessage(new InvalidOperationException(LogHelper.FormatInvariant(
-                                    LogMessages.IDX14112,
```
debchoudhury-id4s, line 229: Is it intentional to drop the `IDX14112` messaging? If yes, should similar changes be made in [Experimental/JsonWebTokenHandler.ClaimsIdentity.cs](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/49c64d135d22059006fac9834e623d92668f335d/src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.ClaimsIdentity.cs#L79) or unrelated ?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -644,5 +631,171 @@ private static TokenValidationResult ReadToken(string token, TokenValidationPara
...
+                        }
+
+                        // Within the limit but not a JSON object: keep as a claim and warn.
+                        LogHelper.LogWarning(LogMessages.IDX14316);
+                    }
+
+                    // Beyond the limit (silent) or a within-limit non-object (warned above): retain the
+                    // "act" value verbatim as a claim.
+                    var actClaim = JsonClaimSet.CreateClaimFromJsonElement(claimType, issuer, value);
+                    if (actClaim is not null)
```
debchoudhury-id4s: below code block is repeated multiple times

```
var xClaim = JsonClaimSet.CreateClaimFromJsonElement(claimType, issuer, value);
if (xClaimis not null)
     identity.AddClaim(xClaim);
```

Would it make sense to extract into `AddClaimIfNotNull(identity, claimType, issuer, value)` and reuse it?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
```diff
@@ -644,5 +631,171 @@ private static TokenValidationResult ReadToken(string token, TokenValidationPara
...
+                        identity.AddClaim(actClaim);
+
+                    continue;
+                }
+
+                // For all other claims, create and add them
```
debchoudhury-id4s: Any similarity with [JsonClaimSet.cs#L96-L111](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/49c64d135d22059006fac9834e623d92668f335d/src/Microsoft.IdentityModel.JsonWebTokens/Json/JsonClaimSet.cs#L96-L111)? Any potential helper re-use oppurtunity?

src/Microsoft.IdentityModel.JsonWebTokens/LogMessages.cs
saurabhsathe-ms: Not reserved — just unused numbers from iteration. I've renumbered this PR's two new messages contiguously after the shipped  IDX14312 :  IDX14314→IDX14313  (retriever failure) and  IDX14316→IDX14314  ( 'act'  must be a JSON object). The remaining  14315/14316  gap sits before  IDX14317 , which is from the already-merged serialization PR #3560, so I left that numbering alone to avoid cross-PR churn. I will create a followup PR to fix this!

src/Microsoft.IdentityModel.JsonWebTokens/LogMessages.cs
saurabhsathe-ms: Good catch — updated the retriever-failure message (now  IDX14313 ) to quote  'act'  consistently with  IDX14314 , and tidied the  is :{0}  →  is: {0}  spacing.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimDeserializationTests.cs
saurabhsathe-ms: Done — extracted two private helpers:  ActorValidationParameters(configure)  (the standard non-validating TVP, was repeated 16×) and  BuildIdentity(authType, params (type,value)[])  (repeated identity construction, 28×). Net −135 lines.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimDeserializationTests.cs
saurabhsathe-ms: Extracted  ActorValidationParameters(configure)  and  BuildIdentity(...)  and applied them here and across the file — removes the duplicated TVP/identity setup.

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActClaimDeserializationTests.cs
saurabhsathe-ms: Same — this setup now uses the shared  ActorValidationParameters(configure)  /  BuildIdentity(...)  helpers. Left the empty  new TokenValidationParameters()  sites and the varied  SecurityTokenDescriptor  blocks inline for readability.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
saurabhsathe-ms: Done — dropped it. It was only a defensive null-guard threaded through recursion; the real null-guard and the  ActClaimRetriever  invocation both live one level up in  CreateActorClaimsIdentity  (which still validates it), so no protection is lost on the validation path. Removed the redundant inner guard, updated the caller + recursive call + test call sites, and removed the now-moot  _NullValidationParameters_  test.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
saurabhsathe-ms: No pre-check needed —  TryGetPayloadValue<JsonElement>  already returns  false  when the claim is absent. The  HasPayloadClaim  below it is intentional and load-bearing: it's in the fall-through to distinguish a primitive  act  present (warn  IDX14314 , suppress  actort ) from  act  absent (fall through to  actort ). Adding a pre-check before  TryGetPayloadValue  would just be a redundant dictionary probe on the common path, so I've left it as-is.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
saurabhsathe-ms: Yes, intentional. Actor resolution moved out of the per-claim loop into a single  ResolveActorClaimsIdentity  call reading the dictionary-backed payload, so duplicate  act / actort  members collapse before we read them and the "second Actor claim" condition  IDX14112  guarded can no longer occur — the throw was dead code. This also matches the PR's "degrade, don't throw" actor design. The Experimental ( ValidationParameters ) pipeline still has the old in-loop  actort  +  IDX14112  logic on both the serialization and deserialization sides; aligning it belongs with porting  act  support to that pipeline, which we'll address in a separate PR.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
saurabhsathe-ms: Done — extracted  AddClaimIfNotNull(identity, claimType, issuer, value)  and reused it at all three sites (act-retain, array element, single value). It's a tiny static method the JIT inlines, so no added call overhead.

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs
saurabhsathe-ms: The array-vs-scalar shape does mirror  JsonClaimSet.CreateClaimFromObject  (L96–111), but that block materializes into a  List<Claim>  inside a private method, so it isn't cleanly shareable without changing that method's signature. I kept the dispatch inline here (now built on the shared  AddClaimIfNotNull ) rather than wrapping it in a loop-containing helper — a loop blocks JIT inlining and this is on the actor read path, so I avoided adding a non-inlinable call. Net: dedupes the repeated null-check without call overhead.

debchoudhury-id4s (approved):

test/Microsoft.IdentityModel.JsonWebTokens.Tests/ActClaimTests/ActEdgeCaseTests.cs
```diff
@@ -0,0 +1,243 @@
...
+
+namespace Microsoft.IdentityModel.JsonWebTokens.Tests.ActClaimTests
+{
+    /// <summary>
+    /// Edge-case tests describing the CORRECT expected behavior of the "act" claim.
+    /// Three of these currently FAIL against the implementation in PR #3571; the
+    /// canonical-delegation test passes and is included as the contrast case.
+    /// </summary>
+    [Collection("ActClaimTests")]
+    public class ActEdgeCaseTests : IDisposable
```
JoshLozensky: I wrote this file to show the bug but it should be combined with the ActClaimDeserializationTests file, EdgeCase is not a very informative title.

Otherwise LGTM :)

JoshLozensky (approved):

PR #3576, opened 2026-07-29
Add synchronous JsonWebTokenHandler.ValidateToken pipeline

Toidi357:
This is PR number 4 of the synchronous token validation pipeline process. 

This PR adds the entrypoints `JsonWebTokenHandler.ValidateToken` to mirror existing `....ValidateTokenAsync` behavior. `ValidateToken` is gated to .NET 5+, as the `HttpClient.Send` synchronous fetch method doesn't exist until after .NET 5. 

Setting the new `AppContextSwitches.PreserveLegacySyncBehavior` to true makes `ValidateToken` do the legacy sync-over-async behavior

Full performance benchmark runs are shown [here](https://microsoft-my.sharepoint.com/:w:/p/t-lukevin/cQpVdsiLDsOkQYO3TM0nm2zZEgUCdLn1rMrwXiiP9rxMrIYW8A)

### DIFF
--- a/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.cs
+++ b/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.cs
@@ -129,6 +129,128 @@ namespace Microsoft.IdentityModel.JsonWebTokens
             }
         }
 
+        internal TokenValidationResult ValidateJWESync(
+            JsonWebToken jwtToken,
+            TokenValidationParameters validationParameters,
+            BaseConfiguration configuration)
+        {
+            return ValidateJWESync(jwtToken, validationParameters, configuration, CancellationToken.None);
+        }
+
+        internal TokenValidationResult ValidateJWESync(
+            JsonWebToken jwtToken,
+            TokenValidationParameters validationParameters,
+            BaseConfiguration configuration,
+            CancellationToken cancellationToken)
+        {
+            try
+            {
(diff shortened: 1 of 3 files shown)


RojaEnnam (approved):

src/Microsoft.IdentityModel.Tokens/Experimental/IResultBasedValidationSync.cs
```diff
@@ -0,0 +1,75 @@
...
+
+#nullable enable
+using System.Threading;
+
+namespace Microsoft.IdentityModel.Tokens.Experimental
+{
+    /// <summary>
+    /// Experimental interface that provides synchronous result based token validation instead of throwing exceptions.
+    /// </summary>
+    public interface IResultBasedValidationSync
```
MZOLN: Why is this needed? Couldnt we use existing interfaces?

src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.ValidateToken.cs
```diff
@@ -475,7 +597,41 @@ private static JsonWebToken ValidateSignatureUsingDelegates(JsonWebToken jsonWeb
         [Obsolete("`JsonWebTokens.ValidateToken(string, TokenValidationParameters)` has been deprecated and will be removed in a future release. Use `JsonWebTokens.ValidateTokenAsync(string, TokenValidationParameters)` instead. For more information, see https://aka.ms/IdentityModel/7-breaking-changes", false)]
```
Toidi357, line 597: we would remove this line and make a note about the AppContextSwitch.PreserveLegacySyncBehavior

PR #3581, opened 2026-08-03
Log IDX10650 decryption-tag failures below Error during multi-key decryption

### LINKED ISSUE #3583: [Bug] Reduce IDX10650 log severity during successful multi-key JWE decryption
Description

When multiple token decryption keys are configured, Wilson may try each key until one successfully decrypts the JWE. Every unsuccessful attempt currently logs  IDX10650  at Error level, even when a later key succeeds and token validation completes successfully.

This can produce many Error-level logs per request during normal key rollover and trigger false alerts.

Expected behavior

• Log intermediate  IDX10650  failures below Error level.
• Continue trying the remaining decryption keys.
• If every key fails, retain the final aggregated  IDX10603  or  IDX10609  Error and fail validation.

Proposed fix

Log  IDX10650  from  AuthenticatedEncryptionProvider.DecryptWithAesCbc  at Informational level. The exception should still be thrown so the decryption loop can collect it and include its details in the final aggregate failure.

This is not an API or runtime breaking change; it changes only the severity of intermediate diagnostic events.

RojaEnnam:
A per-key JWE authentication-tag failure (IDX10650) is expected noise while JsonWebTokenHandler probes multiple candidate decryption keys (try-all when it cannot resolve one by kid/x5t). Logging each attempt at Error produced many Error-level logs during otherwise-successful token validation, which broke partner alerting and could not be filtered.

Log the per-key failure at Informational instead. It is suppressed by default (default LogLevel is Warning) and remains available for diagnostics. The terminal all-keys-failed case is still reported at Error via IDX10603/IDX10609, which aggregates each attempted key's exception detail, so nothing is lost.

Adds a regression test asserting IDX10650 is not logged at Error but is logged at Informational.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
+++ b/src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Diagnostics.Tracing;
 using System.IO;
 using System.Security.Cryptography;
 using Microsoft.IdentityModel.Logging;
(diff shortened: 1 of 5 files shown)


src/Microsoft.IdentityModel.Tokens/Encryption/AuthenticatedEncryptionProvider.cs
```diff
@@ -180,7 +181,12 @@ private byte[] DecryptWithAesCbc(byte[] ciphertext, byte[] authenticatedData, by
...
-                throw LogHelper.LogExceptionMessage(new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(LogMessages.IDX10650, Base64UrlEncoder.Encode(authenticatedData), Base64UrlEncoder.Encode(iv), Base64UrlEncoder.Encode(authenticationTag))));
+                // Verifying the authentication tag can legitimately fail while probing multiple candidate
+                // decryption keys (JsonWebTokenHandler tries all keys when it cannot resolve one by kid/x5t).
+                // Log this per-key failure below Error so it is not surfaced as noise during otherwise-successful
+                // decryption; the terminal all-keys-failed case is still reported at Error via IDX10603/IDX10609,
+                // which aggregates each attempted key's exception detail.
+                throw LogHelper.LogExceptionMessage(EventLevel.Informational, new SecurityTokenDecryptionFailedException(LogHelper.FormatInvariant(LogMessages.IDX10650, Base64UrlEncoder.Encode(authenticatedData), Base64UrlEncoder.Encode(iv), Base64UrlEncoder.Encode(authenticationTag))));
```
dlingam-msft, line 189: This returns IDX10603/10609 as a  ValidationError  without calling  LogHelper , so this per-key IDX10650 was its only Error-level EventSource event on a failed decrypt. Lowering it to Informational means all-keys-failed decryption on the  ValidationParameters  path now emits zero Error events. Please validate.

RojaEnnam: Good catch. Added logging for IDX10603/10609

Toidi357 (approved):

pmaytak (approved):

westin-m (approved):

JoshLozensky (approved):

PR #3584, opened 2026-08-04
Propagate caller cancellation when retrieving OIDC configuration

cpp11nullptr:
## Problem

`ConfigurationManager<T>.GetConfigurationAsync(cancel)` dropped the caller's `CancellationToken` to `CancellationToken.None` for the shared initial retrieval, so the token had no effect. A caller that supplied a cancelable token (e.g. one with a deadline) could never abort a slow or hung metadata fetch - it would hang until the underlying request completed on its own, regardless of the token.

## Fix

Separate *who is waiting* from *the shared work*:

- Concurrent callers are coalesced onto a single shared `_configurationTask` under a lightweight lock (behind a lock-free fast path; no I/O held under the lock).
- Callers await via `WaitForConfigurationAsync(task, cancel)`, which honors each caller's token (`Task.WhenAny` + `CancellationToken.Register`) and then `ThrowIfCancellationRequested()`.
- The shared fetch still runs with `CancellationToken.None`, so it outlives any individual caller — one caller cancelling never disrupts the fetch or other awaiters.
- `HttpDocumentRetriever` now re-throws only caller-requested cancellation; transport/`HttpClient` timeouts are still wrapped as `IOException`.

Not a breaking change to the shared-fetch contract; both blocking and non-blocking paths are covered.

## Tests

- Caller cancellation does not cancel the shared retrieval (blocking + non-blocking).
- The originating caller cancels promptly while a concurrent co-awaiter still gets the config.
- Shared fetch is invoked with a non-cancelable token.
- `HttpDocumentRetriever` re-throws caller cancellation instead of wrapping as `IOException`.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs
@@ -25,11 +25,13 @@ namespace Microsoft.IdentityModel.Protocols
         private DateTimeOffset _lastRequestRefresh = DateTimeOffset.MinValue;
         private bool _isFirstRefreshRequest = true;
         private readonly SemaphoreSlim _configurationNullLock = new SemaphoreSlim(1);
+        private readonly object _configurationTaskLock = new object();
 
         private readonly IDocumentRetriever _docRetriever;
         private readonly IConfigurationRetriever<T> _configRetriever;
         private readonly IConfigurationValidator<T> _configValidator;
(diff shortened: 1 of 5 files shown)


MZOLN:
Similar to this: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3502/changes

We cannot take this change this way, we have to consider breaking behaviors. There will be a consolidated approach to how we fetch this.

PR #3586, opened 2026-08-10
Fix Inconsistent Manual Request Refresh Throttling

Toidi357:
# Fix Inconsistent Manual Request Refresh Throttling

The field `_lastRequestRefresh` is used to throttle manual `RequestRefresh()` requests to the `BaseConfigurationManager.MinimumRefreshInterval` property (default is 1 second)

The documentation for this property states `Minimum time interval (1 second) that must pass before calling RequestRefresh() to obtain new configuration.`

Currently, in blocking mode,  `_lastRequestRefresh`  was updated after successful configuration retrieval rather than when a manual refresh request was accepted.

This caused inconsistent behavior depending on how configuration was retrieved:
• Successful retrieval through the regular configuration retriever updated ` _lastRequestRefresh` .
• Successful retrieval through `IConfigurationEventHandler`  returned before updating  `_lastRequestRefresh`, noticed [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3567#discussion_r3741313381)
• Automatic and initial retrievals through the regular retriever could update the field even though no manual refresh was requested.
• Consequently, repeated manual requests could bypass throttling when an event handler supplied the configuration.

This PR moves updates to `_lastRequestRefresh` to `RequestRefresh()` call sites instead of in the configuration retrieval paths, so automatic retrievals do not affect request throttling.

### DIFF
--- a/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
+++ b/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
@@ -74,8 +74,6 @@ namespace Microsoft.IdentityModel.Protocols
                                 throw LogHelper.LogExceptionMessage(new InvalidConfigurationException(LogHelper.FormatInvariant(LogMessages.IDX20810, result.ErrorMessage)));
                         }
 
-                        _lastRequestRefresh = TimeProvider.GetUtcNow().UtcDateTime;
-
                         TelemetryForUpdateBlocking(TelemetryConstants.Protocols.ConfigurationSourceRetriever);
 
                         if (_refreshRequested)
@@ -157,12 +155,13 @@ namespace Microsoft.IdentityModel.Protocols
 
         private void RequestRefreshBlocking()
         {
-            DateTime now = TimeProvider.GetUtcNow().UtcDateTime;
+            DateTimeOffset now = TimeProvider.GetUtcNow();
 
             if (now >= DateTimeUtil.Add(_lastRequestRefresh.UtcDateTime, RefreshInterval) || _isFirstRefreshRequest)
             {
                 _refreshRequested = true;
                 _syncAfter = now;
+                _lastRequestRefresh = now;
                 _isFirstRefreshRequest = false;
             }
         }
--- a/test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerBlockingEventHandlerTests.cs
+++ b/test/Microsoft.IdentityModel.Protocols.OpenIdConnect.Tests/ConfigurationManagerBlockingEventHandlerTests.cs
@@ -495,6 +495,52 @@ public class ConfigurationManagerBlockingEventHandlerTests
     }
 
     [Fact]
+    public async Task Blocking_RequestRefresh_IsThrottledFromAcceptedRequest()
+    {
+        // Arrange
+        AppContext.SetSwitch(AppContextSwitches.UpdateConfigAsBlockingSwitch, true);
+
+        var timeProvider = new FakeTimeProvider();
+        var mockEventHandler = new MockConfigurationEventHandlerContextAware
+        {
+            ConfigurationToReturn = new OpenIdConnectConfiguration { Issuer = "https://test.issuer.com" },
+            RetrievalTimeToReturn = timeProvider.GetUtcNow()
+        };
+        var configurationManager = new ConfigurationManager<OpenIdConnectConfiguration>(
+            "OpenIdConnectMetadata.json",
+            new OpenIdConnectConfigurationRetriever(),
+            new FileDocumentRetriever())
+        {
+            ConfigurationEventHandler = mockEventHandler,
+            TimeProvider = timeProvider
+        };
+
+        await configurationManager.GetConfigurationAsync();
+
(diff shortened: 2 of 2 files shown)


westin-m:
I don't think that this is a bug. Right now, `_lastRequestRefresh` is the last time the configuration was actually refreshed. Moving it would break that invariant. If we change it to update that field when the request is made, a fetch that fails (network error or invalid config) would start the `RefreshInterval` window anyway and block manual refresh requests for that time while no config was updated. 

Moreover, `_lastRequestRefresh` is only written to inside a critical section while holding `_refreshLock`. If we move it and multiple threads have access to it, they can read a stale value and pass the throttle check at the same time. We would have no confidence that the timestamp is atomic with the rest of the config-retrieval state.

Toidi357:
So the main issue now is inconsistent behavior between code and comments. The goal of `MinimumRefreshInterval` and `DefaultRefreshInterval` is stated as `time interval that must pass for RequestRefresh to obtain a new configuration`, which is essentially manual `RequestRefresh` throttling. 

Right now, the code is setting `_lastRequestRefresh` inconsistently. In non-blocking mode, `_lastRequestRefresh` is set after a retrieval task is successfully queued inside `RequestRefreshBackground` (`ConfigurationManager.cs:507`), and in blocking mode, `_lastRequestRefresh` is set after successful retrieval (`ConfigurationManager_Blocking.cs:77`). Both these have possible issues similar to what you described:
- In blocking mode, we can have an automatic refresh that sets `_lastRequestRefresh`, then right after that, the server rotates/invalidates keys, forcing a `RequestRefresh`, but we are still within the `RefreshInterval` window and cannot retrieve updated metadata
- In non-blocking mode, we can have a `RequestRefresh` called which sets `_lastRequestRefresh`, but this retrieval can error out, but the `RefreshInterval` window is reset and will block manual refresh requests for that time while no config was updated

My PR makes blocking mode behave like non-blocking mode, `_lastRequestRefresh` is set to when `RequestRefresh` is called, not when configuration is successfully retrieved. This has its issues of not being able to update configs when necessary, but I think can be mitigated in other ways, maybe just set `DefaultRefreshInterval` to something lower than 5 minutes?

debchoudhury-id4s:
Thanks @Toidi357 for picking this. Pasting my questions from the [earlier PR](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3567#discussion_r3741313381) . 

@westin-m , @RojaEnnam  Please let me know if my understanding is wrong

1. Shouldn't `_lastRequestRefresh` throttle `RequestRefresh` only ([`DefaultRefreshInterval`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/BaseConfigurationManager.cs#L52-L54)), like non-blocking already does on accept ([here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager.cs#L504-L507))?
1. Wouldn't setting `_lastRequestRefresh` in [`RequestRefreshBlocking`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/kevin/fix-blocking-refresh-throttling/src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs#L160-L164) fix both problems: (1) after a handler returns config, another `RequestRefresh` could fire too soon because nothing stamped the timer ([#3567](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3567#discussion_r3741313381)), and (2) after a normal automatic metadata fetch, a real manual `RequestRefresh` could be blocked too long because the timer was stamped on fetch success instead of on the manual request?

src/Microsoft.IdentityModel.Protocols/Configuration/ConfigurationManager_Blocking.cs
```diff
@@ -74,8 +74,6 @@ private async Task<T> GetConfigurationWithBlockingAsync(CancellationToken cancel
                                 throw LogHelper.LogExceptionMessage(new InvalidConfigurationException(LogHelper.FormatInvariant(LogMessages.IDX20810, result.ErrorMessage)));
                         }
 
-                        _lastRequestRefresh = TimeProvider.GetUtcNow().UtcDateTime;
```
debchoudhury-id4s, line 77: If this line is removed, wouldn't a later manual `RequestRefresh` no longer get blocked by an automatic fetch?

Toidi357: my idea is same as yours, move this line to `RequestRefreshBlocking` so a manual RequestRefresh is not blocked by automatic fetch, but then that runs into a situation Westin says:

> if we change it to update that field when the request is made, a fetch that fails (network error or invalid config) would start the RefreshInterval window anyway and block manual refresh requests for that time while no config was updated.

debchoudhury-id4s: > runs into a situation

cc: @RojaEnnam  and @westin-m to please check this thread

Issue #3351, opened 2025-10-14
[Feature Request] Move Tests from feature/new-validation-model to dev


keegan-caruso:
**Is your feature request related to a problem? Please describe.**

The tests added here are high value: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3077/files

They should be moved to dev branch

PR #3589, opened 2026-08-11
Add validation tests

westin-m:
Addresses #3351 

Add additional test cases for new validation model that cover a wider range of scenarios. 

Create internal setters on ValidationParameters to enable such tests.


src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -32,11 +32,11 @@ internal static class LogMessages
         // token validation
         public const string IDX10204 = "IDX10204: Unable to validate issuer. validationParameters.ValidIssuer is null or whitespace AND validationParameters.ValidIssuers is null or empty.";
         public const string IDX10205 = "IDX10205: Issuer validation failed. Issuer: '{0}'. Did not match: validationParameters.ValidIssuer: '{1}' or validationParameters.ValidIssuers: '{2}' or validationParameters.ConfigurationManager.CurrentConfiguration.Issuer: '{3}'. For more details, see https://aka.ms/IdentityModel/issuer-validation. ";
-        public const string IDX10206 = "IDX10206: Unable to validate audience. The 'audiences' parameter is empty.";
```
Toidi357, line 35: TVP path also throws IDX10206 right now, see [Validators.cs](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/Validators.cs) Line 357, I think it should be left as a generic error without the "in ValidationParameters" part

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
```diff
@@ -32,11 +32,11 @@ internal static class LogMessages
...
+        public const string IDX10206 = "IDX10206: Unable to validate audience. The 'audiences' parameter is empty in ValidationParameters.";
         public const string IDX10207 = "IDX10207: Unable to validate audience. The 'audiences' parameter is null.";
         public const string IDX10208 = "IDX10208: Unable to validate audience. validationParameters.ValidAudience is null or whitespace and validationParameters.ValidAudiences is null.";
         public const string IDX10209 = "IDX10209: Token has length: '{0}' which is larger than the MaximumTokenSizeInBytes: '{1}'.";
-        public const string IDX10211 = "IDX10211: Issuer validation failed. Unable to validate issuer. The 'issuer' parameter is null or whitespace.";
+        public const string IDX10211 = "IDX10211: Issuer validation failed. Unable to validate issuer. The 'issuer' parameter is null or whitespace in ValidationParameters.";
```
Toidi357: same here

src/Microsoft.IdentityModel.Tokens/LogMessages.cs
westin-m: Fair enough. done

src/Microsoft.IdentityModel.Tokens/Experimental/Validation/ValidationParameters.cs
```diff
@@ -323,6 +323,10 @@ public IList<SecurityKey> SigningKeys
                     Interlocked.CompareExchange(ref _signingKeys, [], null) ??
                     _signingKeys;
             }
+            internal set
```
saurabhsathe-ms, line 326: The new `internal set` accessors write the backing field directly, which skips the `Interlocked.CompareExchange` the getters use:

```csharp
public IList<SecurityKey> SigningKeys
{
    get =>
        _signingKeys ??
        Interlocked.CompareExchange(ref _signingKeys, [], null) ??
        _signingKeys;
    internal set { _signingKeys = value; }   // plain, unsynchronized write
}
```

Two things follow from that. The getter was written to be safe under concurrent access, and an unsynchronized setter on the same field takes that guarantee away. And because the setter accepts `null`, `vp.ValidIssuers = null` followed by a read returns `[]` rather than `null` — the assignment is silently dropped.

Since these setters exist to let the new tests use object initializers, could we avoid the API change entirely? `ValidationUtils.CreateValidationParameters(...)` already does this, and four tests in this same file still use it. Adding optional `audiences` / `issuers` / `signingKeys` parameters would cover the new scenarios, keep initialization behind the existing `CompareExchange`, and make the test file internally consistent.

If object-initializer syntax is what we want here, `internal init` would be the safer accessor — it's assignable only during construction, so there's no shared-instance mutation to reason about. (Worth checking the `IsExternalInit` shim is available on the `net462` / `netstandard2.0` targets first.)

If post-construction assignment is genuinely intended, then at minimum: `Interlocked.Exchange(ref _signingKeys, value ?? []);` — that gives the write a barrier consistent with the getter and preserves the never-null invariant. Worth a `<remarks>` noting the setter is for configuration before the instance is shared.

westin-m: good call, no need for internal setters

test/Microsoft.IdentityModel.JsonWebTokens.Tests/JsonWebTokenHandler.ValidateTokenAsyncTests_e2e.cs
```diff
@@ -8,94 +8,371 @@
...
+    public async Task TestDefaultValidToken()
+    {
+        string token = testTokenCreator.CreateDefaultValidToken();
+        ValidationParameters validationParameters = ValidationUtils.CreateValidationParameters(
+            audiences: ["http://Default.Audience.com"],
+            issuers: ["http://Default.Issuer.com"],
+            signingKeys: [KeyingMaterial.JsonWebKeyRsa256SigningCredentials.Key]);
+        CallContext callContext = new CallContext();
+
+        ValidationResult<ValidatedToken, ValidationError> validationResult = await jsonWebTokenHandler.ValidateTokenAsync(token, validationParameters, callContext, default);
```
saurabhsathe-ms: The new tests invoke  jsonWebTokenHandler.ValidateTokenAsync(...)  on the concrete type, whereas the existing tests ( TestNotYetValidToken  etc.) go through  resultBasedJsonWebTokenHandler  (the  IResultBasedValidation  cast). Both bind to the same internal method, but for consistency I'd route all of them through  resultBasedJsonWebTokenHandler  (the established convention) — or switch all to the concrete call and keep it uniform.

saurabhsathe-ms (approved):

Toidi357 (approved):
much needed :)

PR #2123, opened 2023-06-26
remove all net45 code paths

westin-m:

PR #3594, opened 2026-08-13
Restore asymmetric key-size floor enforcement in AsymmetricSignatureProvider

neha-bhargava:
## Summary
Restores the minimum asymmetric key-size floor enforcement in `AsymmetricSignatureProvider`.

`ValidateAsymmetricSecurityKeySize` was only ever evaluated from the **net45-only** `CreateAsymmetricAdapter` (via the lazy `_keySizeIsValid`). When the net45 code paths were removed, that evaluation went with them, leaving `ValidKeySize()` / `ValidateAsymmetricSecurityKeySize` with no live callers. As a result, RSA/ECDSA/PSS keys **below the library's documented floors** (e.g. RS256 signing < 2048 bits, verifying < 1024 bits) were silently accepted.

## Change
- Call `ValidateAsymmetricSecurityKeySize(Key, Algorithm, WillCreateSignatures)` at the end of the shared 3-arg constructor. All other constructors chain to it via `: this(...)`, so every construction path is covered.
- This is eager (fail-fast) validation that matches the constructor's already-documented `ArgumentOutOfRangeException` contract, and simultaneously re-enables the ECDSA and PSS floor maps.
- No public API change.

## Behavioral impact
Keys below the documented floors now throw `ArgumentOutOfRangeException` (IDX10630 signing / IDX10631 verifying) at provider construction, as documented. Compliant keys (>= floor) are unaffected. Real IdP/signing keys (>= 2048-bit) are unaffected.

## Tests
- New `ConstructorEnforcesAsymmetricKeySizeFloor` covering below-floor signing (IDX10630), below-floor verifying (IDX10631), and compliant 2048-bit acceptance.
- Updated two existing tests that incidentally used sub-floor keys to compliant 2048-bit keys (`ShouldCacheSignatureProvider` empty-kid caching; `Valid_SignatureValidationResult_Success_X5tMatches`).
- Full `Microsoft.IdentityModel.*` test suite green on net8.0.


bgavrilMS:
Functionally ok, but this is potentially a breaking change

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
```diff
@@ -148,6 +148,8 @@ public AsymmetricSignatureProvider(
             _asymmetricAdapterObjectPool = new DisposableObjectPool<AsymmetricAdapter>(
                 CreateAsymmetricAdapter,
                 _cryptoProviderFactory.SignatureProviderObjectPoolCacheSize);
+
+            ValidateAsymmetricSecurityKeySizeCore(Key, Algorithm, WillCreateSignatures);
```
iNinja: This bypasses the public virtual `ValidateAsymmetricSecurityKeySize` method, so derived providers can no longer apply custom or stricter validation during normal signing and verification. `ValidKeySize()` still dispatches through the virtual method, but production code no longer calls it.

Could we preserve the mandatory base floor while continuing to invoke the virtual hook after construction—perhaps once, immediately before the first adapter is created? Please also add a derived-provider test demonstrating the intended override behaviour.

test/Microsoft.IdentityModel.Tokens.Tests/AsymmetricSignatureTests.cs
```diff
@@ -47,6 +47,61 @@ public void UnsupportedRSATypes()
             TestUtilities.AssertFailIfErrors(context);
         }
 
+        // Guards that the asymmetric key-size floor is enforced at construction time.
```
iNinja, line 50: Could this test cover the other affected paths as well? In particular, RSA-PSS, ECDSA, `JsonWebKey`, and an overridden `ValidateAsymmetricSecurityKeySize`. The existing direct-call tests would not detect that the virtual method has become disconnected from production operations.

iNinja:
One compatibility point to consider: before #2123, this validation was performed only by the `NET45` adapter path. The modern .NET and .NET Standard paths did not invoke it, so this introduces new constructor-time failures for some existing applications rather than solely restoring previously observed behaviour across all target frameworks.

I think this behavioural change should be called out clearly in the release notes. It may also be worth evaluating whether a temporary opt-out mechanism is appropriate for applications that need time to rotate legacy keys.

test/Microsoft.IdentityModel.Tokens.Tests/AsymmetricSignatureTests.cs
neha-bhargava: Done

neha-bhargava: Done

src/Microsoft.IdentityModel.Tokens/AsymmetricSignatureProvider.cs
neha-bhargava: Good catch. Done

bgavrilMS (approved):

iNinja (approved):

PR #3596, opened 2026-08-18
Add Composite ML-DSA support (internal, gated)

iNinja:
# Add Composite ML-DSA support (internal, gated)

## Summary

Adds internal support for composite post-quantum signatures (Composite ML-DSA) for JOSE, based on `draft-ietf-jose-pq-composite-sigs`. All new types and algorithm identifiers are `internal` — this PR adds no public API.

Algorithms: `ML-DSA-44-ES256`, `ML-DSA-65-ES256`, `ML-DSA-87-ES384`. EdDSA-based composites are deferred (Wilson has no EdDSA support).

## Implementation

Uses the .NET `System.Security.Cryptography.CompositeMLDsa` API (`[Experimental]`). Composite keys are treated as opaque units — the .NET API handles the combiner internally. Follows the same patterns as the existing ML-DSA implementation.

New files: `CompositeMLDsaSecurityKey`, `CompositeMLDsaAdapter`.

## Feature gate

Composite ML-DSA validation is gated behind:

```
Microsoft.IdentityModel.CompositeMLDsa.EnableDraft (default: false)
```

The switch must be explicitly enabled via `AppContext.SetSwitch(...)`. Composite signing is implicitly gated by `CompositeMLDsaSecurityKey` being `internal`.

The gate exists because `draft-ietf-jose-pq-composite-sigs` is pre-RFC and the .NET `CompositeMLDsa` API is `[Experimental]`. When the spec reaches RFC and `[Experimental]` is removed, a follow-up PR will promote the types to `public`, add `PublicAPI.Unshipped.txt` entries, and remove the gate.

## Tests

31 tests: key construction, sign/verify round-trips, JWS create/validate, concurrent signing, and AppContextSwitch behaviour (including a regression test verifying that keys constructed while the switch was on are correctly rejected after it is turned off).

Tests are skipped on platforms where `CompositeMLDsa.IsSupported` returns `false`.

## Tools

`Tools/GenerateCompositeMLDsaVectorTests/` fetches the spec reference vectors and generates a test file. Not wired into the test run; retained for use once the spec and .NET API stabilise.


src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
```diff
@@ -98,6 +98,23 @@ internal static class AppContextSwitches
...
+        /// <remarks>
+        /// Unlike other switches, this property is not cached — it reads the current
+        /// AppContext switch value on every call. This allows the switch to be toggled
+        /// dynamically at runtime (e.g. by a remote feature flag) without requiring a
+        /// process restart.
+        /// </remarks>
+        internal const string EnableCompositeMLDsaDraftSwitch = "Microsoft.IdentityModel.CompositeMLDsa.EnableDraft";
+
+        internal static bool EnableCompositeMLDsaDraft =>
+            AppContext.TryGetSwitch(EnableCompositeMLDsaDraftSwitch, out bool enabled) && enabled;
```
fre-berhane, line 116: This dynamic gate can be bypassed once a verification provider has been cached.  CryptoProviderFactory.CreateSignatureProvider  checks  CryptoProviderCache.TryGetSignatureProvider(...)  before  IsSupportedAlgorithm(...) , so a provider created while the switch is on is returned and continues validating composite signatures after the switch is turned off. The existing test only checks an already-created key and does not exercise the default provider cache. Please enforce the gate before cache lookup—or evict/disable cached composite providers when off—and add a regression test that validates once with caching enabled, disables the switch, then validates again.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -193,6 +207,35 @@ private void InitializeUsingMlDsaSecurityKey(MlDsaSecurityKey mlDsaSecurityKey,
...
+
+            if (clone is not null)
+            {
+                CompositeMLDsa = clone;
+                _disposeCryptoOperators = true;
+            }
+            else
+            {
+                CompositeMLDsa = compositeMLDsa;
+                _compositeMLDsaSyncLock = new object();
```
fre-berhane: The non-exportable-key fallback does not serialize access across the object pool. Each  AsymmetricAdapter  sharing the same source key creates its own lock object here, so two pooled adapters can enter  CompositeMLDsa.SignData  or  VerifyData  concurrently on the same instance despite the stated fallback guarantee. The synchronization object must be shared per source key—or the source itself must be locked—and the fallback path should have a concurrent regression test.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
```diff
@@ -130,6 +135,13 @@ internal static string GetKeyAlgorithmId(SecurityKey key)
                 _ => KeyAlgorithmIds.MlDsaUnknown
             },
 
+#pragma warning disable SYSLIB5006
```
cpp11nullptr: Can we add a comment what's that SYSLIB5006 is about?

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -148,6 +157,9 @@ protected virtual void Dispose(bool disposing)
 
                         if (MLDsa != null)
                             MLDsa.Dispose();
+
+                        if (CompositeMLDsa != null)
+                            CompositeMLDsa.Dispose();
```
cpp11nullptr, line 165: nit: will CompositeMLDsa?.Dispose() work there? (similar to other cases, but I realize that it was already there)

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
```diff
@@ -193,6 +207,35 @@ private void InitializeUsingMlDsaSecurityKey(MlDsaSecurityKey mlDsaSecurityKey,
...
         }
 
+        private void InitializeUsingCompositeMLDsaSecurityKey(CompositeMLDsaSecurityKey compositeMLDsaSecurityKey, bool requirePrivateKey)
+        {
+            InitializeUsingCompositeMLDsa(compositeMLDsaSecurityKey.CompositeMLDsa, requirePrivateKey);
+        }
+
+        private void InitializeUsingCompositeMLDsa(CompositeMLDsa compositeMLDsa, bool includePrivateKey)
+        {
+            CompositeMLDsa clone = CompositeMLDsaAdapter.CloneCompositeMLDsa(compositeMLDsa, includePrivateKey);
```
cpp11nullptr: The cloning is required because it's not possible to share compositeMLDsa, is my understanding correct?

test/Microsoft.IdentityModel.TestUtils/CompositeMLDsaConditionalAttributes.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// A <see cref="TheoryAttribute"/> that skips the test when the specified Composite ML-DSA
+    /// algorithm is not supported on the current platform.
+    /// </summary>
+    public sealed class CompositeMLDsaTheoryAttribute : TheoryAttribute
+    {
+        public CompositeMLDsaTheoryAttribute(string algorithm)
+        {
+            if (!CompositeMLDsa.IsSupported)
+            {
+                Skip = "Composite ML-DSA is not supported on this platform.";
```
cpp11nullptr: In case if support of ML-DSA is driven by software environment as well, perhaps, "Composite ML-DSA is not supported" (with omitting "platform") would be less confusing.

Tools/GenerateCompositeMLDsaVectorTests/Program.cs
```diff
@@ -0,0 +1,171 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cpp11nullptr: I'd advocate to avoid autogenerated code there unless there is strong reason to do it - it'll help to debug such tests in future.

src/Microsoft.IdentityModel.Tokens/AppContextSwitches.cs
iNinja: Addressed. The built-in Composite ML-DSA gate is now checked before the default signature-provider cache lookup, so a provider cached while the draft feature was enabled cannot be reacquired after the switch is disabled. Custom crypto-provider precedence is preserved. I also added a regression test that caches and releases a verifying provider, disables the switch, and confirms that requesting the same provider fails rather than returning the cached instance.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
iNinja: Addressed. Adapters that reference the same non-exportable CompositeMLDsa instance now obtain a shared synchronization object from a ConditionalWeakTable. Cloning and key export are also performed under that shared lock, preventing concurrent export attempts while pooled adapters are being created. Once cloning succeeds, each adapter continues independently without the lock. I added concurrent signing and verification regressions that force multiple pooled adapters to contend and verify that both export and cryptographic operations are serialized.

src/Microsoft.IdentityModel.Tokens/Telemetry/CryptoTelemetry.cs
iNinja: Addressed. The suppression now explains that CompositeMLDsa is marked experimental under SYSLIB5006.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
iNinja: Addressed. CompositeMLDsa?.Dispose() would be functionally equivalent here. I kept the explicit null check because it matches the adjacent ECDsa, ML-DSA, and RSA disposal code in this method and avoids changing unrelated existing style.

src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs
iNinja: Addressed. Sharing is possible when access is synchronized, but cloning is preferred for exportable keys. AsymmetricSignatureProvider maintains a pool of adapters, and CompositeMLDsa instance methods are not documented as safe for concurrent use. Giving each pooled adapter an independent clone avoids concurrent access and lock contention, and lets the adapter own and dispose its clone without disposing the caller's key. For non-exportable keys, cloning is not possible, so the implementation now shares the original instance with synchronization common to every adapter referencing that instance.

test/Microsoft.IdentityModel.TestUtils/CompositeMLDsaConditionalAttributes.cs
iNinja: Addressed. I changed the XML documentation and skip messages to say that Composite ML-DSA, or the requested algorithm, is not supported in the current environment. This covers runtime and native cryptographic-provider availability as well as the operating platform.

Tools/GenerateCompositeMLDsaVectorTests/Program.cs
iNinja: Addressed. I removed the generator project. The draft and its reference vectors can continue to evolve, and the generator followed a mutable source while no generated test file was committed or executed by CI. If cross-implementation vectors are added later, they should be pinned to an immutable draft revision and consumed by ordinary tests without a code-generation step.

cpp11nullptr (approved):

fre-berhane (approved):

PR #3607, opened 2026-08-25
Add support for ILogger via CallContext

westin-m:
Enable logging through ILogger. 

This is a precursor to using high performance logging with LoggerMessage


test/Microsoft.IdentityModel.Logging.Tests/PIITests.cs
```diff
@@ -193,6 +199,10 @@ public void LogExceptionAsArgument()
             LogHelper.LogExceptionMessage(EventLevel.Error, new ArgumentException(LogHelper.FormatInvariant("Main exception 1: {0}", new SecurityTokenCompressionFailedException("custom inner exception"))));
             LogHelper.LogExceptionMessage(EventLevel.Error, new ArgumentException(LogHelper.FormatInvariant("Main exception 2: {0}", new InvalidOperationException("system exception"))));
 
+            Console.WriteLine($"TraceBuffer: '{listener.TraceBuffer}'");
```
Toidi357: why are there Console.WriteLine's inside unit tests

westin-m: forgot to clean them up after trying something

Toidi357 (approved):

src/Microsoft.IdentityModel.Logging/LogHelper.cs
JoshLozensky: Should this be null or empty to avoid formatting when there are zero args?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -523,27 +676,78 @@ private static LogEntry WriteEntry(EventLogLevel eventLogLevel, Exception innerE
 
             message = args == null ? message : FormatInvariant(message, args);
```
JoshLozensky: Should this check be null or empty to avoid formatting when there are 0 args? I noticed at least a couple of calls to WriteEntry no longer explicitly define args as null (lines 304 and 491) so I believe they'd instead evaluate to an empty array and trigger this. Added a similar comment to a check at 481

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -289,7 +301,38 @@ public static Exception LogExceptionMessage(EventLevel eventLevel, Exception exc
...
+
+            if (!loggerContext.Logger.IsEnabled(LogLevel.Error))
+                return exception;
+
+            LogEntry entry = WriteEntry(
+                EventLogLevel.Error,
+                exception.InnerException,
+                exception.Message,
+                loggerContext.CorrelationId ?? loggerContext.ActivityId.ToString(),
+                exception.Message,
```
JoshLozensky: why is exception.Message passed twice here?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -333,7 +428,43 @@ public static void LogWarning(string message, params object[] args)
...
+
+        /// <summary>
+        /// Logs at the warning level to the listeners that have been enabled.
+        /// </summary>
+        /// <param name="message">The log message.</param>
+        /// <param name="loggerContext">A <see cref="LoggerContext"/> that contains logging control including <see cref="ILogger"/>.</param>
+        /// <param name="args">An object array that contains zero or more objects to format.</param>
+        public static void LogWarning(string message, LoggerContext loggerContext, params object[] args)
+        {
+            LogWarning(message, args);
```
JoshLozensky, line 451: Why are we calling log warning no matter what and then additionally calling Log below if there is a logger context, wouldn't that log twice?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
```diff
@@ -333,7 +428,43 @@ public static void LogWarning(string message, params object[] args)
...
+
+            if (loggerContext?.Logger == null)
+                return;
+
+            if (!loggerContext.Logger.IsEnabled(LogLevel.Warning))
+                return;
+
+            LogEntry entry = WriteEntry(
```
JoshLozensky, line 459: Why do we throw away the args in a lot of these WriteEntry calls?

src/Microsoft.IdentityModel.Logging/LogHelper.cs
westin-m: we should probably check for empty as well. I'll add that

src/Microsoft.IdentityModel.Logging/LogHelper.cs
westin-m: It should not be, I'm removing it

src/Microsoft.IdentityModel.Logging/LogHelper.cs
westin-m: That LogWarning sends logs to IdentityModelEventSource and IIdentityLogger, while the rest of this method writes to ILogger. Because the current logging writes to both of the existing options, I figure that it is in that same spirit to check if any of the three are configured. Customers would get multiple logs only if they enable all the different loggers.

src/Microsoft.IdentityModel.Logging/LogHelper.cs
westin-m: We shouldn't, that was left over from putting things together

westin-m:
Closing as the logging story requires further planning

PR #3609, opened 2026-08-26
Add high-performance logging to XML

Toidi357:
Add CallContext overloads to DSigSerializer methods and add `XmlLoggerExtensions.cs`, the extension for high performance logging

Only one message is currently being logged inside Xml package: `IDX30300: KeyInfo skipped unknown element: '{0}'.`


src/Microsoft.IdentityModel.Xml/XmlLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+
+#if !NET6_0_OR_GREATER
+using System;
+#endif
+using Microsoft.Extensions.Logging;
+using Microsoft.IdentityModel.Logging;
+
+namespace Microsoft.IdentityModel.Xml
+{
+    internal static partial class XmlLogger
```
westin-m: All of the classes with these log messages should be named `LoggerMessage`

src/Microsoft.IdentityModel.Xml/XmlLoggerExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+                    LogHelper.FormatInvariant("{0}", element));
+            }
+        }
+
+#if NET6_0_OR_GREATER
+        [LoggerMessage(
+            EventId = 30300,
+            Level = LogLevel.Warning,
+            Message = "IDX30300: KeyInfo skipped unknown element: '{Element}'.")]
+        private static partial void UnknownKeyInfoElementSkippedCore(ILogger logger, string element);
```
westin-m: Because the `Level` is set in the attribute, the source generated code includes a check for the log level. With that, there are currently three `logger.IsEnabled` checks happening each time.

src/Microsoft.IdentityModel.Xml/XmlLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        internal static void UnknownKeyInfoElementSkipped(this ILogger logger, string element)
```
westin-m: There is no need for internal methods that expose the logger methods. They should be `public` rather than `private`.

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -278,6 +344,23 @@ protected virtual bool TryReadRetrievalMethod(XmlReader reader, out string metho
         /// <param name="reader">A <see cref="XmlReader"/> positioned on a <see cref="XmlSignatureConstants.Elements.X509Data"/> element.</param>
         /// <param name="data">The parsed <see cref="XmlSignatureConstants.Elements.X509Data"/> element.</param>
         protected virtual bool TryReadX509Data(XmlReader reader, out X509Data data)
+            => TryReadX509DataCore(reader, out data, null);
```
westin-m, line 347: won't calling this always result in throwing a null argument exception?

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
```diff
@@ -171,6 +195,48 @@ protected virtual bool TryReadKeyInfoType(XmlReader reader, ref KeyInfo keyInfo)
             return true;
         }
 
+        /// <summary>
+        /// Attempts to read the key info type which is the child of the &lt;KeyInfo> element.
+        /// </summary>
+        /// <param name="reader">A <see cref="XmlReader"/> positioned on a child of a <see cref="XmlSignatureConstants.Elements.KeyInfo"/> element.</param>
+        /// <param name="keyInfo">The <see cref="KeyInfo"/> object to populate.</param>
+        /// <param name="callContext">The context used to log information about this call.</param>
+        protected virtual bool TryReadKeyInfoType(XmlReader reader, ref KeyInfo keyInfo, CallContext callContext)
```
westin-m, line 204: reading key info type is outside of the scope of adding logging. Why is this included?

src/Microsoft.IdentityModel.Xml/XmlLogger.cs
Toidi357: So something like 
```
namespace Microsoft.IdentityModel.XML
{
   internal static partial class LoggerMessage
```
??
because there already is `LoggerMessage` in line 25 and defined in Microsoft.Extensions.Logging package

src/Microsoft.IdentityModel.Xml/XmlLoggerExtensions.cs
Toidi357: yup, removed one IsEnabled check in `XmlLogger.cs`, though we still need 2 checks, one in `XmlLogger.cs` as the attribute and one in `DsigSerializer` at line 436, Agency's reasoning is:
```
The source generated check happens after method arguments are evaluated.

Without the check at line 436, this executes even when warning logging is disabled:

reader.ReadOuterXml()
LogHelper.FormatInvariant(...)

That materializes and formats the entire skipped XML element. The outer check instead calls the cheaper  reader.Skip()  when disabled. The generated check remains a defensive check inside the logging method.```

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
Toidi357: No, `TryReadX509DataCore` doesn't throw exception if `CallContext` (the third argument) is null, it will just ignore the `CallContext` and use `LogHelper`

src/Microsoft.IdentityModel.Xml/XmlLogger.cs
Toidi357: done

src/Microsoft.IdentityModel.Xml/XmlLogger.cs
westin-m: My bad, I meant `LoggerExtensions`. example: https://identitydivision.visualstudio.com/DevEx/_git/MISE?path=/src/Core/Logging/CoreLoggerExtensions.cs

src/Microsoft.IdentityModel.Xml/DsigSerializer.cs
Toidi357: `TryReadKeyInfoType` calls `TryReadX509Data` which calls `TryReadX509DataCore` which can possibly log a warning, so `CallContext` also needs to be added into this function as well

src/Microsoft.IdentityModel.Xml/InternalAPI.Unshipped.txt
westin-m, line 1: Since we will have logging methods that differ between target frameworks, we should [split the API files](https://github.com/dotnet/roslyn/blob/main/src/RoslynAnalyzers/PublicApiAnalyzers/PublicApiAnalyzers.Help.md#conditional-api-differences). It's what we do in other packages already

Toidi357: There actually is no difference in public API across the TFMs, the only difference between the TFMs is private, which doesn't get tracked in the InternalAPI

PR #3616, opened 2026-09-01
Refactor PropertyBag assignment

JoshLozensky:
Preserve the source dictionary's  IEqualityComparer  when  TokenValidationParameters.Clone()  deep-copies  PropertyBag , matching the 8.x implementation, so a bag created with a custom comparer (for example  StringComparer.OrdinalIgnoreCase ) no longer silently reverts to the default comparer on the clone.

### DIFF
--- a/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
+++ b/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs
@@ -73,7 +73,12 @@ namespace Microsoft.IdentityModel.Tokens
             LogValidationExceptions = other.LogValidationExceptions;
             NameClaimType = other.NameClaimType;
             NameClaimTypeRetriever = other.NameClaimTypeRetriever;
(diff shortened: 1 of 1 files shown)


Toidi357 (approved):
I did some more digging into this bug. The root issue is that `PropertyBag` is typed as `IDictionary` not `Dictionary`, and therefore, you cannot truly do a shallow copy of `IDictionary`. An example would be that someone sets (for some reason) `PropertyBag` to be of type `ConcurrentDictionary`, but after `Clone()`, it would just become normal `Dictionary`

I highly doubt this would pose any sort of actual, practical problem though

JoshLozensky:
> of `IDictionary`. An example would be that someone sets (for some reason) `PropertyBag` to be of type `ConcurrentDictio`

We also often prefer a deep copy so we don't risk changing or hitting issues with dict values set by the app using the library.

westin-m (approved):

PR #3622, opened 2026-09-04
Remove netcoreapp2.2 from Windows test target matrix on dev5x

saurabhsathe-ms:
# Remove netcoreapp2.2 from Windows test matrix on dev5x

## Description

Why

The  RunThirdPartyLibrariesTests  job in the release build fails on  dev5x . Its test projects target  netcoreapp2.2 , which pins  RuntimeFrameworkVersion 2.2.3 . The .NET Core 2.2 runtime (EOL Dec 2019) is no longer present on the 1ES build image ( MwWilson1EsHostedPool ), so the test host can't launch and aborts the build. Wilson code and the pipeline are unchanged — only the agent image changed. ( dev  already tests only on net6.0–net10.0.)

What

Drop  netcoreapp2.2  from the Windows  TestTargets  in  build/targetsTest.props :

- <TestTargets>net452;net461;netcoreapp2.2</TestTargets>
+ <TestTargets>net452;net461</TestTargets>

Not a breaking change —  netcoreapp2.2  is test-only; shipped package TFMs ( SrcTargets ) are untouched.

How

• Windows test runs now use  net452;net461 , both of which run on the .NET Framework present on the agent.
•  TestOnlyCoreTargets  and the  netcoreapp2.2  conditional blocks are intentionally left in place — the non-Windows build ( build.sh  /  WilsonUnix.sln ) still relies on them.

### DIFF
--- a/build/targetsTest.props
+++ b/build/targetsTest.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <TestTargets>net452;net461;netcoreapp2.2</TestTargets>
+    <TestTargets>net452;net461</TestTargets>
     <TestOnlyCoreTargets>netcoreapp2.2</TestOnlyCoreTargets>
   </PropertyGroup>
 </Project>
--- a/build/targetsTest.props
+++ b/build/targetsTest.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <TestTargets>net452;net461</TestTargets>
-    <TestOnlyCoreTargets>netcoreapp2.2</TestOnlyCoreTargets>
+    <TestTargets>net452;net461;net10.0</TestTargets>
+    <TestOnlyCoreTargets>net10.0</TestOnlyCoreTargets>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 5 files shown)


build/targetsTest.props
```diff
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <TestTargets>net452;net461;netcoreapp2.2</TestTargets>
```
bgavrilMS, line 3: Can we test with net10 instead? The test should cover supported frameworks, so it's ok to drop netcoreapp2.2 and test with the lowest supported .net version? (I guess 9 is also ok)

build/targetsTest.props
```diff
@@ -1,6 +1,6 @@
...
+    <TestTargets>net452;net461</TestTargets>
     <TestOnlyCoreTargets>netcoreapp2.2</TestOnlyCoreTargets>
```
bgavrilMS: Doesn't this need an upgrade?

build/targetsTest.props
fre-berhane: Updated to Net10.0.

bgavrilMS (approved):

pmaytak (approved):

PR #3630, opened 2026-09-10
Update signing key resolving

cpp11nullptr:


neha-bhargava (approved):

bgavrilMS (approved):
Should not be breaking. If CX uses their own key, they cannot expect default key to **also** be used for verification.

PR #3633, opened 2026-09-11
Align experimental XML and SAML validation results

### LINKED ISSUE #3450: Align Xml & SAML experimental validators on ValidationResult<T, ValidationError>
The XML package currently uses ValidationError? (nullable) as a return type. Convert to the proper discriminated-union result type and clean up adjacent SAML TODOs.

Files:

- src/Microsoft.IdentityModel.Xml/Experimental/Signature.Experimental.cs
- src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
- src/Microsoft.IdentityModel.Xml/Experimental/SignedInfo.Experimental.cs
- src/Microsoft.IdentityModel.Tokens.Saml/Experimental/SamlTokenUtilities.Experimental.cs

Specific TODOs to resolve:

- SignedInfo.Experimental.cs:24 — change return to ValidationResult<SecurityKey, ValidationError>
- Signature.Experimental.cs:58 — align IDX10511 parameters with JsonWebTokenHandler
- SamlTokenUtilities.Experimental.cs:217 — reclassify failure as CryptoProviderFactory failure (not AlgorithmValidationFailure)

Update all callers in Tokens.Saml experimental handlers accordingly.

debchoudhury-id4s:
## Problem  

Experimental XML validators use null to signal success. This differs from the result-based token validation path. Some digest failures throw or become generic signature errors. SAML fallback validation also reverses the signature and canonical byte arrays.  Fixes #3450.  

## Requirements  

1. Align experimental XML verification with the result-based validation contract. 
1. Keep SAML and SAML2 callers consistent. 
1. Distinguish crypto-factory failures from algorithm-policy failures. 
1. Preserve public XML behavior and existing signature diagnostics. 
1. Handle malformed digests and valid fallback signing keys correctly.  

## Code Changes  

1. Experimental XML verification now returns explicit results. A successful reference check returns the reference. Successful signature and signed-info checks return the signing key. SAML callers check the result status before using the key.  

1. Expected digest failures return `ReferenceDigestValidationFailed`. The error keeps the original exception. Factory support failures use a dedicated failure type. The existing `IDX10520` diagnostic stays unchanged.  

1. The fallback call now passes its byte arrays in the correct order. Named arguments make that order clear. The remaining XML factory null checks use `is null`.  

1. These are breaking changes to the experimental internal `Verify` signatures. All affected callers are updated. This aligns them with the existing result-based validation contract. Public XML overloads keep their signatures and throwing behavior.  

1. Before, `DigestValue = "not-valid-base64!!!"` threw a `FormatException`. It now returns a digest failure with that exception attached. A signed SAML token without `KeyInfo` can now validate with a matching fallback key when `TryAllSigningKeys` is enabled.  

## Testing

| Area | Result |
| --- | --- |
| Added coverage | Covers result payloads, digest errors, fallback keys, factory failures, and resource release. Both SAML versions are covered. |
| Focused tests | 82 passed on .NET 8. |
| Clean builds | Six projects passed across all declared frameworks. There were no warnings or errors. |
| Full affected suites on .NET 6/8/9/10 | 15,889 passed. None failed. One existing test is skipped on each framework. |
| net462/net472 runtime tests | Not run. Local strong-name verification blocks the delay-signed assemblies. Both targets compile. |

## Telemetry

No metrics or tags are added. The `IdentityModelSignatureValidation` counter stays in use. Successful validation records `None`. Digest failures record `SignatureVerificationFailed`. Existing factory-error telemetry tags stay unchanged.

Example counter tags:

```json
[
  { "Error": "None" },
  { "Error": "SignatureVerificationFailed" }
]

### CODE REVIEW

File: src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs
Comment (saurabhsathe-ms) [inline src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs:232]: Not beacuse of your changes but I think its worth addressing here. The null-provider result  is masked by the unconditional release in  finally (on line 302) .  CryptoProviderFactory.ReleaseSignatureProvider(null)  throws  ArgumentNullException ; the new test passes only because  TrackingCryptoProviderFactory  overrides release and ignores null. A custom factory that overrides only  CreateForVerifying  reproduces this for both SAML and SAML2. Please guard the release with  if (signatureProvider is not null) , as  Signature.Experimental.cs  already does, and adjust the test so it exercises the base release behavior.

**Review by saurabhsathe-ms — APPROVED** [review]

File: src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs [hunk 1]
```diff
@@ -20,27 +20,54 @@ public partial class Reference : DSigElement
...
-            if (!Utility.AreEqual(ComputeDigest(cryptoProviderFactory), Convert.FromBase64String(DigestValue)))
+            byte[] digest;
+            byte[] expectedDigest;
+            try
+            {
+                digest = ComputeDigest(cryptoProviderFactory);
+                expectedDigest = Convert.FromBase64String(DigestValue);
+            }
+            catch (Exception ex) when (ex is XmlException or System.Xml.XmlException or CryptographicException
+                or ArgumentException or FormatException or InvalidOperationException or NotSupportedException)
```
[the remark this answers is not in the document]
debchoudhury-id4s: Thanks. FormatException does not derive from ArgumentException; it derives from SystemException. Convert.FromBase64String throws FormatException, so this arm is required. It is not redundant and does not fail the warnings-as-errors build.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs
debchoudhury-id4s: Thanks. Guarded the SAML and JWT finally blocks so we only release a non-null provider. The tracking factory now always calls the base release. The null-provider test asserts the release count stays at 0.

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.ValidateSignature.cs
```diff
@@ -353,7 +353,8 @@ private ValidationResult<SecurityKey, ValidationError> ValidateSignatureWithKey(
             }
             finally
             {
-                cryptoProviderFactory.ReleaseSignatureProvider(signatureProvider);
+                if (signatureProvider is not null)
+                    cryptoProviderFactory.ReleaseSignatureProvider(signatureProvider);
```
MZOLN, line 357: Isn't crypto provider doing the null check? Would we have a test for that?

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs
```diff
@@ -167,8 +167,8 @@ internal static ValidationResult<SecurityKey, ValidationError> ValidateSignature
                 }
                 else if (validationParameters.TryAllSigningKeys)
                     return ValidateSignatureUsingAllKeys(
-                        canonicalBytes,
-                        signatureValueBytes,
+                        signatureValueBytes: signatureValueBytes,
```
MZOLN, line 170: unnecessary change.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs
```diff
@@ -257,28 +255,30 @@ private static ValidationResult<SecurityKey, ValidationError> ValidateSignatureW
...
-                if (result == null)
+                ValidationResult<SecurityKey, ValidationError> signedInfoResult =
+                    signature.SignedInfo.Verify(key, cryptoProviderFactory, callContext);
+
+                if (!signedInfoResult.Succeeded)
                 {
                     RecordSignatureValidationTelemetry(
                         telemetryClient,
-                        TelemetryConstants.SignatureValidationErrors.None,
+                        TelemetryConstants.SignatureValidationErrors.SignatureVerificationFailed,
```
MZOLN, line 265: Are we sure that this flip actually contained a bug and was needed? Does this error help us?

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
```diff
@@ -20,27 +20,54 @@ public partial class Reference : DSigElement
...
 
-            if (!Utility.AreEqual(ComputeDigest(cryptoProviderFactory), Convert.FromBase64String(DigestValue)))
+            byte[] digest;
+            byte[] expectedDigest;
+            try
+            {
+                digest = ComputeDigest(cryptoProviderFactory);
+                expectedDigest = Convert.FromBase64String(DigestValue);
+            }
+            catch (Exception ex) when (ex is XmlException or System.Xml.XmlException or CryptographicException
```
MZOLN, line 50: And what happens to the other exceptions? They get propagated up? Is this what we intend?

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
```diff
@@ -20,27 +20,54 @@ public partial class Reference : DSigElement
...
+            if (!Utility.AreEqual(digest, expectedDigest))
                 return new SignatureValidationError(
                     new MessageDetail(
                         LogMessages.IDX30201,
                         Uri ?? Id),
                     SignatureValidationFailure.ReferenceDigestValidationFailed,
                     ValidationError.GetCurrentStackFrame());
 
-            return null;
+            return this;
```
MZOLN, line 70: This seems like a drastic change are we sure it's correct?

src/Microsoft.IdentityModel.Xml/Experimental/Signature.Experimental.cs
```diff
@@ -62,30 +70,27 @@ public partial class Signature : DSigElement
                     if (!signatureProvider.Verify(memoryStream.ToArray(), Convert.FromBase64String(SignatureValue)))
                     {
                         StringBuilder keyAttempted = new StringBuilder().Append(key.ToString()).Append(", KeyId: ").AppendLine(key.KeyId);
-                        validationError = new SignatureValidationError(
+                        return new SignatureValidationError(
```
MZOLN, line 73: I thought that we are going for results?

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.ValidateSignature.cs
debchoudhury-id4s: crypto provider does a [throw LogHelper.LogArgumentNullException(nameof(signatureProvider))](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/c5afa90d4143f9f4049993979d22c13068016345/src/Microsoft.IdentityModel.Tokens/CryptoProviderFactory.cs#L898-L899) `if (signatureProvider == null)` . hence handling it here to align with the requirements

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs
debchoudhury-id4s: this is to align with the requirement [change return to ValidationResult<SecurityKey, ValidationError>](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/3450) that was previously aligned already. 

[`SignedInfo.Verify` now returns `ValidationResult<SecurityKey, ValidationError>`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/c5afa90d4143f9f4049993979d22c13068016345/src/Microsoft.IdentityModel.Xml/Experimental/SignedInfo.Experimental.cs#L21-L29).

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
debchoudhury-id4s: Yes, other exceptions are propagated. That is how I understood the requirements in [#3450](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/3450)

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
debchoudhury-id4s: The [old method returned `ValidationError?` and used null for success](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/f9e389b98857195cc82c73463b3c6b03731ccf30/src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs#L23-L43). 

The [new method returns `ValidationResult<Reference, ValidationError>`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/c5afa90d4143f9f4049993979d22c13068016345/src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs#L23-L37), so it needs a value when validation succeeds. The current `Reference` is that value.

src/Microsoft.IdentityModel.Xml/Experimental/Signature.Experimental.cs
debchoudhury-id4s: This is still result-based. [`Signature.Verify` returns `SignatureValidationError` when signature verification fails](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/c5afa90d4143f9f4049993979d22c13068016345/src/Microsoft.IdentityModel.Xml/Experimental/Signature.Experimental.cs#L70-L77)

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs
debchoudhury-id4s: Here the ordering (of args) needed to be changed to align with signature of [`ValidationResult<SecurityKey, ValidationError> ValidateSignatureUsingAllKey`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/c5afa90d4143f9f4049993979d22c13068016345/src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs#L311-L312)

debchoudhury-id4s: while making that change (given the same type `byte[]`) the named argument pattern was used to emphasize

src/Microsoft.IdentityModel.JsonWebTokens/Experimental/JsonWebTokenHandler.ValidateSignature.cs
MZOLN: Hah I hope that we will fix that at some point at the core of the issue.

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs
MZOLN: it's just not needed churn that's all

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs
MZOLN: But is it the same error code? before it was None now it is SingatureVerificationFailed seems like that change has an impact beyond return structure change

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
MZOLN: The method flips from returning null to a current instance and then adds a try/catch with some exceptions caught and others allowed to slip through changing the potential error surface and behavior of this method 

This doesn't sound like a simple result refactoring equivalent. All I'm asking is it intentional and are we sure that we don't want to cover all the cases?

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
MZOLN: Sounds breaking, are we sure that there wont be side effects here? Did we cover for them?

MZOLN (commented):
Replied

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
robezirg: breaking for new path only or from perspective of business logic? This is a new path so breaking changes are okay compared to TVP, unless logially this is incorrect

debchoudhury-id4s: Yes this is in experimental path only

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
debchoudhury-id4s: Got your point about the `selective catch`. We should either convert all relevant exceptions consistently or preserve the existing handling. 

I can remove this catch and restore the [original exception behavior](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/f9e389b98857195cc82c73463b3c6b03731ccf30/src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs#L30-L41)

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs
westin-m: Why not check for `result.Error == null` and keep the same telemetry? I understand changing this code to align with the new return type of `Verify`, but I do not understand flipping the success and failure cases. This seems like an overcomplicated way to update the call site to handle the new return type.

src/Microsoft.IdentityModel.Tokens/Experimental/Validation/ValidationFailureType.cs
```diff
@@ -34,6 +34,18 @@ private class NullArgumentFailure : ValidationFailureType { internal NullArgumen
...
+        /// The selected crypto provider factory does not support the requested signature algorithm with the supplied key.
+        /// </summary>
+        public static readonly ValidationFailureType CryptoProviderFactoryDoesNotSupportAlgorithm =
+            new CryptoProviderFactoryDoesNotSupportAlgorithmFailure(
+                "CryptoProviderFactoryDoesNotSupportAlgorithm");
+
+        private class CryptoProviderFactoryDoesNotSupportAlgorithmFailure : ValidationFailureType
+        {
+            internal CryptoProviderFactoryDoesNotSupportAlgorithmFailure(string name) : base(name) { }
+        }
```
westin-m:
```suggestion
        public static readonly ValidationFailureType CryptoProviderFactoryDoesNotSupportAlgorithm = new CryptoProviderFactoryDoesNotSupportAlgorithmFailure("CryptoProviderFactoryDoesNotSupportAlgorithm");
        private class CryptoProviderFactoryDoesNotSupportAlgorithmFailure : ValidationFailureType { internal CryptoProviderFactoryDoesNotSupportAlgorithmFailure(string name) : base(name) { } }
```

Please adhere to the established style to maintain readability.

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
```diff
@@ -20,14 +20,17 @@ public partial class Reference : DSigElement
         /// </summary>
         /// <param name="cryptoProviderFactory">supplies the <see cref="HashAlgorithm"/>.</param>
         /// <param name="callContext"> contextual information for diagnostics.</param>
-        /// <exception cref="ArgumentNullException">if <paramref name="cryptoProviderFactory"/> is null.</exception>
-        internal ValidationError? Verify(
+        /// <returns>
+        /// A <see cref="ValidationResult{TResult, TError}"/> containing this <see cref="Reference"/> if the digest is valid;
+        /// otherwise, a <see cref="ValidationError"/>.
+        /// </returns>
+        internal ValidationResult<Reference, ValidationError> Verify(
```
westin-m: Why return `Reference` instead of doing the same thing as in `SignedInfo` and adding the `SecurityKey` as a parameter to be returned?

src/Microsoft.IdentityModel.Xml/Experimental/Signature.Experimental.cs
```diff
@@ -62,30 +70,27 @@ public partial class Signature : DSigElement
...
             finally
             {
                 if (signatureProvider is not null)
                     cryptoProviderFactory.ReleaseSignatureProvider(signatureProvider);
             }
-
-            if (validationError is not null)
-                return validationError;
-
-            return null; // no error
```
westin-m: Introducing more exit points throughout the method body make it harder to read and analyze them. I do not see any benefit from reorganizing it this way, especially since it seems to be beyond the scope of updating call sites to handle a different return type.

src/Microsoft.IdentityModel.Xml/Experimental/SignedInfo.Experimental.cs
```diff
@@ -13,35 +13,41 @@ public partial class SignedInfo : DSigElement
...
+                ValidationResult<Reference, ValidationError> referenceResult =
+                    References[i].Verify(cryptoProviderFactory, callContext);
 
-                if (validationError is not null)
-                {
-                    validationError.AddCurrentStackFrame();
-                    break;
-                }
+                if (!referenceResult.Succeeded)
+                    return referenceResult.Error!.AddCurrentStackFrame();
```
westin-m: Same question here; why invert from checking for a non-null error to a non-success? This way, there is no guarantee that `Error` is non-null, and the method call for the stack frame could fail. 

Also, why change from `break` to `return`?

westin-m (changes requested):

src/Microsoft.IdentityModel.Tokens.Saml/Saml/Experimental/SamlTokenUtilities.Experimental.cs
debchoudhury-id4s: Addressed

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
debchoudhury-id4s: Addressed

src/Microsoft.IdentityModel.Tokens/Experimental/Validation/ValidationFailureType.cs
debchoudhury-id4s: Addressed

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
debchoudhury-id4s: Returning the `Reference` on success avoids otherwise unnecessary additional parameter. Putting the code flow here

```
SAML or SAML2 experimental token handler
  -> SamlTokenUtilities.ValidateSignature(...)
  -> ValidateSignatureWithKey(...)
  -> signatureProvider.Verify(canonicalBytes, signatureBytes)
  -> SignedInfo.Verify(key, cryptoProviderFactory, callContext)
  -> Reference.Verify(cryptoProviderFactory, callContext)
     for each reference, stop at the first failure
```

src/Microsoft.IdentityModel.Xml/Experimental/Signature.Experimental.cs
debchoudhury-id4s: Removed the new early returns from the main verification block.

src/Microsoft.IdentityModel.Xml/Experimental/SignedInfo.Experimental.cs
debchoudhury-id4s: Addressed

src/Microsoft.IdentityModel.Xml/Experimental/Reference.Experimental.cs
debchoudhury-id4s: `SecurityKey` would also not break anything and work

